PR #143, opened 2020-08-11
Add documentation around catalogers, UI elements, and the event bus

wagoodman:
- Adds godoc comments to describe high level functionality around catalogers, UI elements, and the event bus
- Adds terse descriptions for low-level functions
- Refactors cataloging to eliminate the singleton

### DIFF
--- a/internal/bus/bus.go
+++ b/internal/bus/bus.go
@@ -1,6 +1,6 @@
 /*
 Package bus provides access to a singleton instance of an event bus (provided by the calling application). The event bus
-is intended to allow for the syft library to publish events for which library consumers can subscribe to. These events
+is intended to allow for the syft library to publish events which library consumers can subscribe to. These events
 can provide static information, but also have an object as a payload for which the consumer can poll for updates.
 This is akin to a logger, except instead of only allowing strings to be logged, rich objects that can be interacted with.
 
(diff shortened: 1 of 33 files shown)


syft/cataloger/cataloger.go
```diff
@@ -1,16 +1,48 @@
...
 	SelectFiles(scope.FileResolver) []file.Reference
-	// NOTE: one of the errors which is returned is "IterationNeeded", which indicates to the driver to
-	// continue with another Select/Catalog pass
+	// Catalog is given the file contents and should return any discovered Packages after analyzing the contents.
 	Catalog(map[file.Reference]string) ([]pkg.Package, error)
+	// TODO: add "IterationNeeded" error to indicate to the driver to continue with another Select/Catalog pass
+	// TODO: we should consider refactoring to return a set of io.Readers instead of the full contents themselves (allow for optional buffering).
+}
+
+// All returns a slice of all locally defined catalogers (defined in child packages)
```
alfredodeza, line 36: Seems like you are using a dot at the end of every godoc string as a convention. This is a nit, but it would be nice to fully normalize to one or the other

wagoodman: righto, attempting to keep them as sentences (unless it is a struct field comment). I'll adjust.

internal/bus/bus.go
```diff
@@ -1,17 +1,33 @@
+/*
+Package bus provides access to a singleton instance of an event bus (provided by the calling application). The event bus
+is intended to allow for the syft library to publish events for which library consumers can subscribe to. These events
```
luhring, line 3: (nitpick on grammar)

>events for which library consumers can subscribe to.

to either of these:

_events to which library consumers can subscribe._

_events which library consumers can subscribe to._

PR #145, opened 2020-08-12
Add demo gif + slack channel links

wagoodman:
- Adds a demo gif to the README to show basic behavior
- Adds the slack links to the issue template selection, just as in https://github.com/anchore/grype/pull/123


README.md
```diff
@@ -7,12 +7,12 @@
 
 A CLI tool and go library for generating a Software Bill of Materials (SBOM) from container images and filesystems.
 
-[//]: # (TODO: add example gif here)
+![Demo](.github/assets/syft-demo.gif)
```
alfredodeza, line 10: Can we have these types of assets hosted elsewhere? This will add more weight to the repo without being part of the functionality. Perhaps the asset that is already public on toolbox.anchore.io ? Or if those are not possible, then host this one file over there?

wagoodman: I could move this over to S3, yes, and I see your point that the more blobs we add to git the longer it will take for folks to clone the repo, etc. I also feel that images that support documentation should live in-repo with the documentation itself. Though this is a trivial demo gif, I feel if we begin adding more images to support documentation that we would want those assets version controlled, to live with the document they are being rendered in.

README.md
```diff
@@ -7,12 +7,12 @@
...
 
 **Features**
 - Catalog container images and filesystems to discover packages and libraries.
 - Supports packages and libraries from various ecosystems (APK, DEB, RPM, Ruby Bundles, Python Wheel/Egg/requirements.txt, JavaScript NPM/Yarn, Java JAR/EAR/WAR, Jenkins plugins JPI/HPI, Go modules)
-- OS distribution detection (supports alpine, busybox, centos/redhat, debian/ubuntu flavored distributions)
+- Linux distribution identification (supports Alpine, BusyBox, CentOS/RedHat, Debian/Ubuntu flavored distributions)
```
luhring, line 15: I like this clarification 👏

README.md
alfredodeza: The thing is that git is not very good with binaries. Using an image (or a gif in this case) is going to be problematic when we need to update with more or different images. Git will still make you clone any other gif/jpeg in the history. Since @luhring  already approved it, I am +0, but I wouldn’t recommend adding it.

PR #150, opened 2020-08-17
Add a `setup.py` cataloger

### LINKED ISSUE #59: Add `setup.py` analyzer
Should be a best effort on parsing `setup.py` for dependencies

For example this is an extract from Pytest's `setup.py` file:

```python
requires = [
    'chardet>=3.0.2,<4',
    'idna>=2.5,<3',
    'urllib3>=1.21.1,<1.26,!=1.25.0,!=1.25.1',
    'certifi>=2017.4.17'

]
test_requirements = [
    'pytest-httpbin==0.0.7',
    'pytest-cov',
    'pytest-mock',
    'pytest-xdist',
    'PySocks>=1.5.6, !=1.5.7',
    'pytest>=3'
]
```

In this case, `grype` should be able to _only_ pick up: `'pytest-httpbin==0.0.7'`

alfredodeza:
It will go into `setup.py` files, but only capture dependencies that are pinned (using the `==` operator).

Closes #59

### DIFF
--- a/syft/cataloger/python/parse_setup.go
+++ b/syft/cataloger/python/parse_setup.go
@@ -1,0 +1,50 @@
+package python
+
+import (
+	"bufio"
+	"io"
+	"regexp"
+	"strings"
+
+	"github.com/anchore/syft/syft/cataloger/common"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+// integrity check
+var _ common.ParserFn = parseSetup
+
+// match examples:
+//		'pathlib3==2.2.0;python_version<"3.6"'  --> match(name=pathlib3 version=2.2.0)
+//		 "mypy==v0.770",                        --> match(name=mypy version=v0.770)
+//		" mypy2 == v0.770", ' mypy3== v0.770',  --> match(name=mypy2 version=v0.770), match(name=mypy3, version=v0.770)
+var pinnedDependency = regexp.MustCompile(`['"]\W?(\w+\W?==\W?[\w\.]*)`)
+
+func parseSetup(_ string, reader io.Reader) ([]pkg.Package, error) {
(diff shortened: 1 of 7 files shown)


syft/cataloger/python/parse_setup.go
```diff
@@ -0,0 +1,46 @@
...
+	"strings"
+
+	"github.com/anchore/syft/syft/cataloger/common"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+// integrity check
+var _ common.ParserFn = parseSetup
+
+var pinnedDependency = regexp.MustCompile(`['"]\W?(\w+\W?==\W?[\w\.]*)`)
```
wagoodman, line 16: nit: would be useful to add a few example matches in comments to illustrate to devs what this is trying to match on (similar to https://github.com/anchore/syft/blob/ac9141e2ada35bcbc07f29f8942d894956ea4b28/syft/cataloger/java/archive_filename.go#L13-L16)

wagoodman, line 16: nit: there is an opportunity to capture more environment marker cases with `['"]\s*([\w]+\s*==\s*[\w\.]*)\s*[;'"]` which would find `ipaddress==1.0` in:
```
setup(
    ...
    install_requires=[
        'six',
        'humanize',
    ],
    extras_require={
        ':python_version == "2.7"': [
            'ipaddress==1.0',
        ],
    },
)
``` 
Not necessary to incorporate in this PR.

alfredodeza: Great suggestion, I'll add those

alfredodeza: This is now improved!

PR #168, opened 2020-09-03
Expand matching of requirements.txt file to include any prefixes or suffixes

### LINKED ISSUE #167: Expand matching of `requirements.txt`
Currently, grype matches on `requirements.txt` files for that file only, and I think it is valid to match on more combinations of `requirements.txt` like:

* `requirements-dev.txt`
* `requirements-prod.txt`
* `test-requirements.txt`

This grep found those variations in (my) local repositories that would benefit from getting inspected. This will probably need only the globs to be updated for Python.

dakaneye:
Fixes #167 


syft/cataloger/python/cataloger.go
```diff
@@ -20,7 +20,7 @@ func New() *Cataloger {
 	globParsers := map[string]common.ParserFn{
 		"**/*egg-info/PKG-INFO":  parseEggMetadata,
 		"**/*dist-info/METADATA": parseWheelMetadata,
-		"**/requirements.txt":    parseRequirementsTxt,
+		"**/*requirements*.txt":  parseRequirementsTxt,
```
alfredodeza, line 23: this is exactly right!

alfredodeza (approved):

test/integration/pkg_cases.go
```diff
@@ -82,6 +82,12 @@ var cases = []struct {
 		pkgLanguage: pkg.Python,
 		pkgInfo: map[string]string{
 			"flask": "4.0.0",
+			"python-dateutil": "2.8.1",
```
alfredodeza, line 85: I think this file is causing the lint check to fail. There is a `make lint-fix` that should get this solved though

dakaneye: Ahh, spacing was off.

wagoodman (approved):

PR #170, opened 2020-09-11
Add support for Arch (which doesnt return a version) and Open Suse Leap

### LINKED ISSUE #48: Add suse distro identification
Should be able to identify a suse image and determine the distro version (populating a `distro.Distro` object from `distro.Identify()`)

### LINKED ISSUE #49: Add arch distro identification
Should be able to identify an arch image and determine the distro version (populating a `distro.Distro` object from `distro.Identify()`)

dakaneye:
Fixes #48 
Fixes #49 


syft/distro/identify.go
```diff
@@ -84,7 +85,8 @@ func assemble(name, version string) *Distro {
 	distroType, ok := IDMapping[name]
 
 	// Both distro and version must be present
-	if len(name) == 0 || len(version) == 0 {
+	if len(name) == 0 {
+		log.Debugf("Name (%s) not found", name)
```
wagoodman, line 89: nit: logging isn't necessary here

syft/distro/distro.go
```diff
@@ -6,6 +6,8 @@ import (
 	hashiVer "github.com/hashicorp/go-version"
 )
 
+const VersionUnknown = "(version unknown)"
```
wagoodman, line 9: I'd vote removing this and leaving the version field blank.

wagoodman (commented):
The `MajorVersion` and `String` functions should be updated to account for when the version is not included.

syft/distro/distro.go
```diff
@@ -20,9 +22,15 @@ func NewUnknownDistro() Distro {
 }
 
 func NewDistro(t Type, ver string) (Distro, error) {
-	verObj, err := hashiVer.NewVersion(ver)
-	if err != nil {
-		return Distro{}, fmt.Errorf("could not create distro version: %w", err)
+	var verObj *hashiVer.Version
```
luhring, line 25: This isn't a required change at all, just a subjective suggestion for making the implementation easier to follow and slightly less nested (and consequently, it cuts down on variables that are declared without defined values). **Feel free to discard!** 😄 

Assuming we accept @wagoodman 's suggestion of using the zero value for `RawVersion` in this case (which I agree with), we can detect a lack of version early on and return a `Distro` accordingly.

```go
if ver == "" {
	return Distro{Type: t}, nil
}

verObj, err := hashiVer.NewVersion(ver)
if err != nil {
	return Distro{}, fmt.Errorf("could not create distro version: %w", err)
}

return Distro{
	Type:       t,
	Version:    verObj,
	RawVersion: ver,
}, nil
```

syft/distro/identify.go
dakaneye: I used it for some local debugging but wasn't sure if it'd be useful later on for troubleshooting! Will remove.

syft/distro/distro.go
dakaneye: great feedback! This is normally how I prefer to program things but I'm still getting my feet wet w/ go, so I missed this! It's funny how new contexts can be distracting this way. Updated to follow this format!

luhring (approved):
Nice!

PR #174, opened 2020-09-14
remove accessing tag when not available

praveenmylavarapu:
Thank you very much for the Syft developers and Anchore for this awesome project! I would love to see the next beta release so as to be able to use CycloneDX format...

Signed-off-by: praveenmylavarapu <saipraveenmylavarapu@gmail.com>


alfredodeza:
Thanks @praveenmylavarapu ! I think we should soon see a release

wagoodman (approved):
Thanks for the bug fix!

PR #175, opened 2020-09-15
Change kebab case to camelCase in json keys throughout app

dakaneye:
@wagoodman @luhring I went through and did a find/replace:
```
json:(.+?)\-(.)(.*?)\`
```

but I'm not sure if this is sufficient. If any of these structs is stored in persistent storage (i.e. the SQLLite db) we may need to update there too.

### DIFF
--- a/schema/json/schema.json
+++ b/schema/json/schema.json
@@ -4,7 +4,7 @@
         "artifacts": {
             "items": {
                 "properties": {
-                    "found-by": {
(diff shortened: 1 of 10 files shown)


dakaneye:
@wagoodman I will do a pass on grype to make sure it handles this change! The only other integration I know of is the VSCode one, but I'm sure there are others. Is there a list you can give me?

PR #178, opened 2020-09-23
Add OCI support + use URI schemes

### LINKED ISSUE #177: Update CLI syntax for input source specification
See [the Slack discussion](https://anchore.slack.com/archives/C0131B70Q7M/p1600796387138500) for more context and motivation for this change.

Schemes (e.g. "dir", "docker") should be delimited from the rest of the URI by a colon (`:`), not by a colon and two slashes (`://`). (It's okay if an individual scheme requires two slashes to begin the scheme-specific portion of the URI.)

For example, if users want to scan the current directory, then instead of typing:

```
$ syft dir://.
```

Users would type:

```
$ syft dir:.
```

In addition, if we can detect that an input string that lacks a scheme specification (e.g. no `dir:`) is most likely to be a filesystem path as opposed to another specification, we could allow an even easier syntax for an implicit specification of input source:

```
$ syft .
```

wagoodman:
Adds the ability to catalog OCI images from tar archive or directory:
```
# explicitly
syft oci-archive:path/to.tar
syft oci-dir:path/to/dir

# or implicitly
syft path/to/tar/or/dir
```

Also hinted in the example above, this PR incorporates changing the URL approach for specifying the source transport to a URI oriented approach (see https://github.com/anchore/stereoscope/pull/41 PR description for more details). TL;DR is `://` is being replaced with `:` to fall in line with the URI RFC.

Lastly, migrates to using the stereoscope fixture utils instead of the functions from `go-testutils`.

Closes #177


go.mod
```diff
@@ -35,10 +36,9 @@ require (
...
-	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208 // indirect
-	golang.org/x/sys v0.0.0-20200610111108-226ff32320da // indirect
-	google.golang.org/genproto v0.0.0-20200615140333-fd031eab31e7 // indirect
-	gopkg.in/ini.v1 v1.57.0 // indirect
 	gopkg.in/yaml.v2 v2.3.0
 )
+
+replace github.com/anchore/stereoscope => ../stereoscope
+
+replace github.com/anchore/go-testutils => ../go-testutils
```
luhring, line 44: Q: Did we decide yesterday not to push `replace` lines?

wagoodman: good question: are we saying not to push replace lines? or not push pinned commits that aren't on main? (or both?)

I feel that there is a lot of good content ready for review and there is 0-risk of merging a replace line since it cannot pass tests (unlike a pinned commit not on main).

I'd vote allowing pushing of replace lines but not pins of commits that are not on main.

cc: @alfredodeza

luhring (approved):

go.mod
wagoodman: from an offline conversation: I'll refrain from adding replace lines in commit log in the future

PR #203, opened 2020-10-07
Add gemspec support

### LINKED ISSUE #197: Support cataloging `gemspec` files
**What would you like to be added**: Cataloging Ruby packages installed by `gem` - which requires parsing of `gemspec` files.

**Why is this needed**: To support parity with Anchore Engine that does parse gemspec, specifically related to this issue https://github.com/anchore/anchore-engine/issues/629

**Additional context**: A `gemspec` parser will need to be put in place, similar to this:

```go

// New returns a new Bundler cataloger object.
func New() *Cataloger {
	globParsers := map[string]common.ParserFn{
		"**/Gemfile.lock": parseGemfileLockEntries,
		".*specifications.*\.gemspec$": parseGemSpecEntries,
	}
```

wagoodman:
- Adds gemspec support to the bundler cataloger; Closes #197 
- Renames the bundler package to the ruby package / Renames the bundler type to Gem
- Splits the ruby cataloger into two sets, one used for images the other for directories
- Reduces the cataloger code by returning the generic cataloger where possible


syft/cataloger/cataloger.go
```diff
@@ -33,16 +33,30 @@ type Cataloger interface {
...
+	}
+}
+
+// DirectoryCatalogers returns a slice of locally implemented catalogers that are fit for detecting packages from index files (and select installations)
+func DirectoryCatalogers() []Cataloger {
+	return []Cataloger{
+		ruby.NewGemFileLockCataloger(),
+		python.NewPythonCataloger(),         // TODO: split and replace me
+		javascript.NewJavascriptCataloger(), // TODO: split and replace me
+		deb.NewDpkgdbCataloger(),
```
alfredodeza, line 56: If for `apk` the package name is `apkdb`, shouldn't this be `dpkgdb` instead of `deb`? I guess this is implying there are multiple different ways of inspecting `deb` packages and we are only using `dpkgdb` here. 
The new consistency makes sense for the constructors, but I'm not sure what the idea is with the packages

syft/cataloger/ruby/parse_gemspec.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/anchore/syft/syft/pkg"
+)
+
+// integrity check
+var _ common.ParserFn = parseGemFileLockEntries
+
+type postProcessor func(string) []string
+
+var patterns = map[string]*regexp.Regexp{
+	// match example:       name = "railties".freeze   --->   railties
```
alfredodeza, line 23: These are great, thanks for adding the examples

syft/cataloger/cataloger.go
wagoodman: good point --there is an inconsistency here that only was partially adressed. There was an attempt to make the package names more generic to the category, while allowing for more specific catalogers. How about:

- `deb.NewDpkgdbCataloger()`
- `apk.NewApkdbCataloger()`
- `rpm.NewrpmdbCataloger()`

This falls more in line with the `python`, `ruby`, `javascript` package names, which encapsulate an ecosystem, and at the same time doesn't prevent us from having more catalogers within the package (e.g. both `rpm.NewRpmCataloger()` and `rpm.NewRpmdbCataloger()`).

syft/cataloger/ruby/parse_gemspec.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/anchore/syft/syft/cataloger/common"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+// integrity check
+var _ common.ParserFn = parseGemFileLockEntries
+
+type postProcessor func(string) []string
+
+var patterns = map[string]*regexp.Regexp{
```
luhring, line 22: I believe all instances of `{1}` in these regex patterns can be safely removed (since I don't think they're needed)

luhring (approved):
Looking great! Love all the cleanup.

This will really pave the way when we evolve the flow of catalogers and resolvers in the coming months.

PR #204, opened 2020-10-07
Fix acceptance tests & add notification upon failures

wagoodman:
Acceptance tests do not occur within the PR checks, which means it is possible for a passing PR to fail after being merged to main (when acceptance tests run). This PR fixes the existing failures and adds notifications to the toolbox channel when a failure is seen.


luhring (approved):
Nice!

PR #207, opened 2020-10-08
Add acceptance test to cover gem cataloger

wagoodman:
The existing gemspec cataloger for image scanning was looking within the `specification/` dir instead of the `specifications/` dir. This PR adjusts the cataloger and adds acceptance tests against a `rails` image to compare syft results to inline scan. To do this, additional `content-*` report parsing was added to the existing inline scan wrapper/compare script to interpret additional results output by inline scanner.


test/inline-compare/compare.py
```diff
@@ -33,8 +33,10 @@ def __init__(self, image, report_dir="./"):
 
     def packages(self):
         python_packages, python_metadata = self._python_packages()
-        os_pacakges, os_metadata = self._os_packages()
-        return python_packages | os_pacakges, {**python_metadata, **os_metadata}
+        npm_packages, npm_metadata = self._npm_packages()
```
Toure, line 36: Shouldn't this npm_packages work be separated into its own PR to keep traceability clean?

test/inline-compare/compare.py
```diff
@@ -67,6 +69,30 @@ def _python_packages(self):
...
+        for entry in self._enumerate_section(
+                report="content-gem", section="content"
+        ):
+            package = Package(name=entry["package"], type=entry["type"].lower(),)
+            packages.add(package)
+            metadata[package.type][package] = Metadata(version=entry["version"])
+
+        return packages, metadata
+
+    def _npm_packages(self):
```
Toure, line 84: Same.

test/inline-compare/compare.py
wagoodman: I can split the other two into another PR 👍

test/inline-compare/compare.py
wagoodman: ditto, this'll make it in the npm package.json cataloger PR coming up

Toure (approved):

PR #232, opened 2020-10-21
Upgrade cataloger interface to use full resolver

### LINKED ISSUE #206: Cataloger 2.0 Interface
Today the Cataloger Interface in syft requires a "request all files" in advance and returns all findings for all catalogers to every cataloger. The API should be able to allow for requests of individual files and return only the contents that cataloger has asked for.

This will become particularly valuable when catalogers' logic evolves to request files from the underlying source in multiple, serial steps (i.e. where the cataloger doesn't know about all required files upfront).

**AC**

- Syft catalogers should not need to register files (including patterns) of interest with the resolver upfront.
- There should be no discernible increase in run time duration with the new implementation.

**Steps to Test**

- Verify that catalogers aren't being asked to provide any hints about files they required ahead of when they need the file(s) in question. (This change should be reflected most prominently in the `Cataloger` interface, with the removal of the `SelectFiles` method.)
- Benchmark a given cataloging workload with the previous implementation and with the new implementation; the latter should take less or approximately equal time to run.

**Notes for the developer**

In the `Cataloger` interface, the `Catalog` can take as a parameter just a `Resolver`, and cataloger implementations can use this resolver to request file reference and file contents on an "as needed" basis.

Consider an idiomatic concurrency solution for this problem, where cataloger requests for file references and contents can be made across a channel, and resolvers can read such requests from channels as they appear. A single cataloger's needs shouldn't block other catalogers from making progress more than is absolutely necessary.

wagoodman:
Upgrades the cataloger interface from:
```golang
type Cataloger interface {
	SelectFiles(scope.FileResolver) []file.Reference
	Catalog(map[file.Reference]string) ([]pkg.Package, error)
}
```

to:
```golang
type Cataloger interface {
	Catalog(resolver scope.Resolver) ([]pkg.Package, error)
}
```
(note: excluded `Name` function in the above snippet for clarity)

In this way the catalogers can reconcile multiple file sources as needed without requiring external interaction with the caller (the `lib.go` `Catalog()` function). In this way state for a cataloger (no matter how complicated it may become) can be constrained to the stack of the `Catalog` function on the cataloger.

A notable downside with this approach is that catalog content requests to stereoscope are no longer guaranteed to be well-ordered/optimized (since catalogers now control theses calls instead of the central `Catalog` function in `lib.go`). This could result in a performance hit, however, we can improve this overtime.

Additionally expands the `ContentResolver` interface to include a content request helper for a single file contents:

```golang
type ContentResolver interface {
	FileContentsByRef(ref file.Reference) (string, error)   // <--- new!
	MultipleFileContentsByRef(f ...file.Reference) (map[file.Reference]string, error)
}
```

Closes #206

### DIFF
--- a/syft/cataloger/catalog.go
+++ b/syft/cataloger/catalog.go
@@ -43,15 +43,16 @@ func Catalog(resolver scope.Resolver, catalogers ...Cataloger) (*pkg.Catalog, er
 	// perform analysis, accumulating errors for each failed analysis
 	var errs error
 	for _, theCataloger := range catalogers {
-		// TODO: check for multiple rounds of analyses by Iterate error
 		packages, err := theCataloger.Catalog(resolver)
 		if err != nil {
 			errs = multierror.Append(errs, err)
 			continue
 		}
 
-		log.Debugf("cataloger '%s' discovered '%d' packages", theCataloger.Name(), len(packages))
-		packagesDiscovered.N += int64(len(packages))
+		catalogedPackages := len(packages)
+
+		log.Debugf("cataloger '%s' discovered '%d' packages", theCataloger.Name(), catalogedPackages)
+		packagesDiscovered.N += int64(catalogedPackages)
 
 		for _, p := range packages {
 			catalog.Add(p)
--- a/syft/cataloger/catalog.go
+++ b/syft/cataloger/catalog.go
@@ -1,7 +1,6 @@
 package cataloger
 
 import (
-	"github.com/anchore/stereoscope/pkg/file"
 	"github.com/anchore/syft/internal/bus"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/event"
@@ -39,35 +38,19 @@ func newMonitor() (*progress.Manual, *progress.Manual) {
 // request.
 func Catalog(resolver scope.Resolver, catalogers ...Cataloger) (*pkg.Catalog, error) {
 	catalog := pkg.NewCatalog()
-	fileSelection := make([]file.Reference, 0)
-
 	filesProcessed, packagesDiscovered := newMonitor()
-
-	// ask catalogers for files to extract from the image tar
-	for _, a := range catalogers {
-		fileSelection = append(fileSelection, a.SelectFiles(resolver)...)
-		log.Debugf("cataloger '%s' selected '%d' files", a.Name(), len(fileSelection))
-		filesProcessed.N += int64(len(fileSelection))
-	}
-
-	// fetch contents for requested selection by catalogers
-	// TODO: we should consider refactoring to return a set of io.Readers instead of the full contents themselves (allow for optional buffering).
-	contents, err := resolver.MultipleFileContentsByRef(fileSelection...)
-	if err != nil {
-		return nil, err
-	}
 
 	// perform analysis, accumulating errors for each failed analysis
 	var errs error
-	for _, a := range catalogers {
+	for _, theCataloger := range catalogers {
 		// TODO: check for multiple rounds of analyses by Iterate error
-		packages, err := a.Catalog(contents)
+		packages, err := theCataloger.Catalog(resolver)
 		if err != nil {
 			errs = multierror.Append(errs, err)
 			continue
 		}
 
-		log.Debugf("cataloger '%s' discovered '%d' packages", a.Name(), len(packages))
+		log.Debugf("cataloger '%s' discovered '%d' packages", theCataloger.Name(), len(packages))
 		packagesDiscovered.N += int64(len(packages))
 
 		for _, p := range packages {
--- a/syft/cataloger/cataloger.go
+++ b/syft/cataloger/cataloger.go
@@ -6,7 +6,6 @@ catalogers defined in child packages as well as the interface definition to impl
 package cataloger
 
 import (
-	"github.com/anchore/stereoscope/pkg/file"
 	"github.com/anchore/syft/syft/cataloger/apkdb"
 	"github.com/anchore/syft/syft/cataloger/deb"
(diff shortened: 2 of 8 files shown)


syft/cataloger/catalog.go
```diff
@@ -39,35 +38,18 @@ func newMonitor() (*progress.Manual, *progress.Manual) {
...
-		packages, err := a.Catalog(contents)
+	for _, theCataloger := range catalogers {
+		packages, err := theCataloger.Catalog(resolver)
 		if err != nil {
 			errs = multierror.Append(errs, err)
 			continue
 		}
 
-		log.Debugf("cataloger '%s' discovered '%d' packages", a.Name(), len(packages))
+		log.Debugf("cataloger '%s' discovered '%d' packages", theCataloger.Name(), len(packages))
```
luhring, line 52: optional nit: `len(packages)` could be captured as a variable since it's used in multiple places

syft/cataloger/cataloger.go
```diff
@@ -25,12 +24,9 @@ import (
...
 	Name() string
-	// SelectFiles discovers and returns specific files that the cataloger would like to inspect the contents of.
-	SelectFiles(scope.FileResolver) []file.Reference
-	// Catalog is given the file contents and should return any discovered Packages after analyzing the contents.
-	Catalog(map[file.Reference]string) ([]pkg.Package, error)
-	// TODO: add "IterationNeeded" error to indicate to the driver to continue with another Select/Catalog pass
-	// TODO: we should consider refactoring to return a set of io.Readers instead of the full contents themselves (allow for optional buffering).
+	// Catalog is given an object to resolve file references and content, this function should return any discovered Packages after analyzing the catalog source.
+	Catalog(resolver scope.Resolver) ([]pkg.Package, error)
+	// TODO: we should consider refactoring to return a set of io.Readers or file.Openers instead of the full contents themselves (allow for optional buffering).
```
luhring, line 29: Just clarifying — is this TODO for this `Catalog` interface, or for the `Resolver` interface?

syft/scope/resolvers/directory_resolver.go
```diff
@@ -82,9 +82,20 @@ func (s DirectoryResolver) MultipleFileContentsByRef(f ...file.Reference) (map[f
 		contents, err := fileContents(fileRef.Path)
 
 		if err != nil {
-			return refContents, fmt.Errorf("could not read contents of file: %s", fileRef.Path)
+			return nil, fmt.Errorf("could not read contents of file: %s", fileRef.Path)
```
alfredodeza, line 85: not sure I follow this change, mind expanding why is this required?

syft/cataloger/common/generic_cataloger.go
```diff
@@ -34,73 +34,82 @@ func NewGenericCataloger(pathParsers map[string]ParserFn, globParsers map[string
...
-	a.selectedFiles = make([]file.Reference, 0)
-	a.parsers = make(map[file.Reference]ParserFn)
+func (c *GenericCataloger) clear() {
+	c.selectedFiles = make([]file.Reference, 0)
+	c.parsers = make(map[file.Reference]ParserFn)
+}
+
+func (c *GenericCataloger) Catalog(resolver scope.Resolver) ([]pkg.Package, error) {
+	fileSelection := c.selectFiles(resolver)
+	contents, err := resolver.MultipleFileContentsByRef(fileSelection...)
```
luhring, line 57: This is probably the wrong spot for this comment — but I'd suggest making the parameter for `MultipleFileContentsByRef` not variadic, if it's true that its argument is always a slice that's needing to be expanded with `...`

syft/cataloger/common/generic_cataloger.go
```diff
@@ -34,73 +34,82 @@ func NewGenericCataloger(pathParsers map[string]ParserFn, globParsers map[string
...
 	}
 
-	return a.selectedFiles
+	return c.selectedFiles
 }
 
 // Catalog takes a set of file contents and uses any configured parser functions to resolve and return discovered packages
-func (a *GenericCataloger) Catalog(contents map[file.Reference]string) ([]pkg.Package, error) {
-	defer a.clear()
+func (c *GenericCataloger) catalog(contents map[file.Reference]string) ([]pkg.Package, error) {
```
luhring, line 92: nit: matching method casing with doc comment casing

luhring (approved):
This is a great step forward! Very excited by this. I came up with only nit pick items — please consider, but I don't want to block this work.

syft/scope/resolvers/directory_resolver.go
wagoodman: not required, just an improvement along the way. When returning errors it's good practice to not return partial information, but instead blank or nil values.

syft/cataloger/cataloger.go
wagoodman: you're right, I should move this to the resolver now instead

syft/cataloger/common/generic_cataloger.go
wagoodman: I don't think there is a performance impact here and allowing for variadic args enables not needing to construct a list when providing multiple arguments.

luhring: Yeah, this was more of an ergonomics comment, there's certainly no performance impact. It's okay to keep this as is for now, but let's stay aware of ergonomic indicators as we move forward — e.g., if callers never end up needing to construct slices anyway (because they already have the data in slice form) then the choice to make this variadic is actually adding work to consumers

syft/scope/resolvers/directory_resolver.go
alfredodeza: Aha! good to know, I tend to get confused about being strict on what is returned and using `nil`

Issue #230, opened 2020-10-20
errors+failures parsing `package.json` files


alfredodeza:
**What happened**: syft failed to parse a container with NPM modules installed.

**What you expected to happen**: Correct marshaling of the Go struct is done

**How to reproduce it (as minimally and precisely as possible)**:

```
$ syft -o json anchore/test_images:npm
```

**Anything else we need to know?**: Here is a subset of the errors:

```
[0003] ERROR cataloger 'javascript-package-cataloger' failed to parse entries (reference={id:3194 Path:/usr/lib/node_modules/npm/node_modules/execa/package.json}): failed to parse package.json file: json: cannot unmarshal object into Go struct field PackageJSON.author of type string
[0003] ERROR cataloger 'javascript-package-cataloger' failed to parse entries (reference={id:4863 Path:/usr/lib/node_modules/npm/node_modules/pacote/node_modules/minipass/package.json}): failed to parse package.json file: json: cannot unmarshal object into Go struct field PackageJSON.author of type string
[0003] ERROR cataloger 'javascript-package-cataloger' failed to parse entries (reference={id:3948 Path:/usr/lib/node_modules/npm/node_modules/json-schema/package.json}): failed to parse package.json file: json: cannot unmarshal object into Go struct field PackageJSON.author of type string
[0003] ERROR cataloger 'javascript-package-cataloger' failed to parse entries (reference={id:2527 Path:/usr/lib/node_modules/npm/node_modules/byte-size/package.json}): failed to parse package.json file: json: cannot unmarshal object into Go struct field PackageJSON.author of type string
[0003] ERROR cataloger 'javascript-package-cataloger' failed to parse entries (reference={id:5963 Path:/usr/lib/node_modules/npm/node_modules/validate-npm-package-license/package.json}): failed to parse package.json file: json: cannot unmarshal object into Go struct field PackageJSON.author of type string
[0003] ERROR cataloger 'javascript-package-cataloger' failed to parse entries (reference={id:2789 Path:/usr/lib/node_modules/npm/node_modules/config-chain/package.json}): failed to parse package.json file: json: cannot unmarshal object into Go struct field PackageJSON.author of type string
[0003] ERROR cataloger 'javascript-package-cataloger' failed to parse entries (reference={id:4936 Path:/usr/lib/node_modules/npm/node_modules/path-parse/package.json}): failed to parse package.json file: json: cannot unmarshal object into Go struct field PackageJSON.author of type string
```

[...]

PR #233, opened 2020-10-21
Javascript parser fix author

Toure:
Included changes will now accept the two supported formats for javascript packages.json which are:

author: <string> 
author: <map>

This will in turn be normalized into a string to match the current expected format for anchore-engine.

closes: #230


go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/anchore/go-testutils v0.0.0-20200925183923-d5f45b0d3c04
 	github.com/anchore/go-version v1.2.2-0.20200701162849-18adb9c92b9b
 	github.com/anchore/stereoscope v0.0.0-20200925184903-c82da54e98fe
+	github.com/apex/log v1.3.0
```
alfredodeza, line 11: what happened here?

syft/cataloger/javascript/parse_package_json.go
```diff
@@ -16,14 +21,61 @@ var _ common.ParserFn = parsePackageLock
...
+var authorPattern = regexp.MustCompile(`^\s*(?P<name>[^<(]*)(\s+<(?P<email>.*)>)?(\s\((?P<url>.*)\))?\s*$`)
+
+func (a *Author) UnmarshalJSON(b []byte) error {
+	var authorStr string
+	var fields map[string]string
+	var author Author
+
+	if err := json.Unmarshal(b, &authorStr); err != nil {
+		// string parsing did not work, assume a map was given
+		// for more information: https://docs.npmjs.com/files/package.json#people-fields-author-contributors
```
alfredodeza, line 49: thanks for linking this

alfredodeza:
There are a couple of extra things that need to be fixed/improved here that I am finding:

`sourcepkg` is not quite right:

```
test_npms.py:67: in test_has_sourcepkg
    assert loaded['sourcepkg'] == metadata['sourcepkg']
E   AssertionError: assert 'https://gith.../hughsk/from2' == 'git://github...hsk/from2.git'
E     - git://github.com/hughsk/from2.git
E     ? ^^                           ----
E     + https://github.com/hughsk/from2
E     ? ^ +++
```

So it looks like someone/something is replacing `https` for `git`. And this is another interesting example:

```
test_npms.py:67: in test_has_sourcepkg
    assert loaded['sourcepkg'] == metadata['sourcepkg']
E   AssertionError: assert 'https://gith...s-cidr#readme' == 'git+https://...d/is-cidr.git'
E     - git+https://github.com/silverwind/is-cidr.git
E     ? ----                                     ^^^^
E     + https://github.com/silverwind/is-cidr#readme
E     ?                                      ^^^^^^^
```

The `#readme` gets removed and a `.git` gets appended.

Some licenses are not getting detected either. Here is what `syft` catches:

```
{'latest': '5.2.0',
 'lics': [],
 'name': 'safe-buffer',
 'origins': [],
 'sourcepkg': 'https://github.com/feross/safe-buffer',
 'versions': ['5.2.0']}
```

And here is what Anchore Engine has:

```
{'latest': None,
 'lics': ['MIT'],
 'name': 'safe-buffer',
 'origins': ['Feross Aboukhadijeh (http://feross.org)'],
 'sourcepkg': 'git://github.com/feross/safe-buffer.git',
 'versions': ['5.2.0']}
```

The `origins` is also not always present, not sure why. Here is `syft`:

```
{'latest': '3.0.0',
 'lics': [],
 'name': 'is-cidr',
 'origins': [],
 'sourcepkg': 'https://github.com/silverwind/is-cidr#readme',
 'versions': ['3.0.0']}
```

And here is Engine:

```
{'latest': None,
 'lics': ['BSD-2-Clause'],
 'name': 'is-cidr',
 'origins': ['silverwind'],
 'sourcepkg': 'git+https://github.com/silverwind/is-cidr.git',
 'versions': ['3.0.0']}
```

go.mod
Toure: Thanks for catching that, I imported log originally to see if I could provide some additional information on a read failure.

wagoodman (commented):
I think once https://github.com/anchore/syft/pull/233/files#r509558326 is addressed this is ready to get merged! Since this is fixing some javascript behavior it would be good to get this in now and cut a patch release.

internal/parse.go
```diff
@@ -0,0 +1,15 @@
+package internal
+
+import "regexp"
+
+// MatchCaptureGroups takes a regular expression and string and returns all of the named capture group results in a map.
+func MatchCaptureGroups(regEx *regexp.Regexp, str string) map[string]string {
+	match := regEx.FindStringSubmatch(str)
+	results := make(map[string]string)
+	for i, name := range regEx.SubexpNames() {
+		if i > 0 && i <= len(match) {
```
luhring, line 10: This is **_super_** small 😄  — for my own clarity — can `i <= len(match)` just be `i < len(match)`, if `i` never gets incremented up to the number of items in the slice?

luhring: And actually — is the part of the condition after `&&` even necessary? `range` handles the incrementing of `i` for this loop and the exiting of the loop when all slice items have been returned

wagoodman: This is a good catch.

`i>0` is not necessary, correct. 

The second term is needed; note that this is a range over `regEx.SubexpNames()` which result is mostly independent than that returned from `regEx.FindStringSubmatch(str)`... the only dependence is that the order of matches corresponds to the order of the names, and it is important to only return the **first** instance of each (so we ignore multiple instances of the capture group names).

@Toure since this is migrating existing code I'd vote you probably don't need to fix this in this PR. I can follow up with a PR to add tests for this function and clean up a bit.

wagoodman (approved):

alfredodeza (approved):
These changes pass functional tests on my end. Thanks for the fixes @Toure !

PR #234, opened 2020-10-22
Split python cataloger by image/directory scanning + add more metadata

### LINKED ISSUE #205: Enhance python cataloger to be image/directory aware.
In order to reduce the number of false positives when scanning images while also keeping the value of grype-vscode [directory] scans, it would be useful when scanning directories to key off of index-like files (e.g. requirements.txt) and installation metadata (e.g. egg, wheel, setup.py) when scanning images for the python cataloger.

**AC**
- Dependencies described within `requirements.txt`-like files are not shown during image scans (but are shown during directory scans)
- All other python catalog sources are reported for both image and directory scans. 

**Steps to Test**
- When scanning a directory:
  - ensure there are examples of a `requirements.txt`, `setup.py`, `egg` metadata installations, and `wheel` metadata installations in the directory being scanned.
  - ensure the scan finds all python examples.
- When scanning an image: 
  - ensure there are examples of a `requirements.txt`, `setup.py`, `egg` metadata installations, and `wheel` metadata installations in the image being scanned.
  - ensure the scan finds all python examples EXCEPT any requirements.txt references.

**Notes for Developer**
Should mirror the gemspec cataloger additions made in https://github.com/anchore/syft/pull/203

wagoodman:
This PR splits out the python cataloger to be dir / image aware, running python "package" catalogers for both image and directory scans while running "index" catalogers only for directory scans.

Additionally this PR adds more metadata to the package for all packages discovered by the "python package" cataloger. The set of changes needed to achieve this are as follows:
- Use the newly upgraded cataloger and extend it further to be able to find files by path relative to other file references. This is necessary since `FileResolvers` provide `file.References` for a given path or glob, however, when multiple values are returned it is not possible to correlate them with other fetched `file.References`. For instance, when looking for wheel `METADATA` and `RECORD` files, it is possible to return multiple instances of each, all originating from different layers --which is not useful when attempting to create a `pkg.Pacakge` that is valid. For this reason the `RelativeFileByPath` was added to fetch files relative to a given `file.Reference` and path, ensuring that the returned reference originates from the same layer or lower (squashed representation).
- Provide a new python package cataloger using the new cataloger interface that pieces together metadata from the original wheel/egg metadata files as well as the `RECORD` and `top_level.txt` files.
- Changes the directory resolver to only modify given paths when they are absolute references (which can be safely assumed to be relative to the root of the scope path) and leave paths that are not absolute paths alone. This allows for catalogers to craft new paths to be fetched from already fetched `file.Reference`'s paths.

Closes #205
Partially addresses https://github.com/anchore/anchore-engine/issues/654


syft/cataloger/python/package_cataloger.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+type PackageCataloger struct{}
+
+// NewPythonPackageCataloger returns a new cataloger for python packages within egg or wheel installation directories.
+func NewPythonPackageCataloger() *PackageCataloger {
+	return &PackageCataloger{}
+}
+
+func (c *PackageCataloger) Name() string {
```
luhring, line 28: We might want to add doc comments now since we're creating new methods. This should be pretty easy — largely copy and paste from the interface

syft/cataloger/python/package_cataloger.go
```diff
@@ -0,0 +1,173 @@
...
+	return "python-package-cataloger"
+}
+
+func (c *PackageCataloger) Catalog(resolver scope.Resolver) ([]pkg.Package, error) {
+	// nolint:prealloc
```
luhring, line 33: 👍  I like this. This rule is not very useful much of the time, IMHO. We may consider disabling globally it at some point.

luhring (approved):
LGTM! I'd only suggest adding doc comments as we go, it'll make our future work easier

syft/cataloger/python/package_cataloger.go
wagoodman: indeed! Feel free to put in the golangci-lint config update

PR #236, opened 2020-10-23
Bootstrap cached dependencies and CI dependencies separately for acceptance tests

wagoodman:
This was missed during the migration from CircleCI to GHA, specifically, the pipeline was depending only on cache to load the `.tmp` dir instead of invoking `make bootstrap` to explicitly populate it on a cache miss.


luhring (approved):
The change looks good.

But I could see where the make target names are slightly confusing. I had to look at the Makefile to figure out what the relative purpose of `bootstrap` and `ci-bootstrap` are. Even after refreshing my memory, it's still a bit awkward feeling (for example, that they don't overlap, and that `bootstrap` adds project dependencies whereas `ci-bootstrap` doesn't. Again, this isn't a comment on your changes at all. But the fact that things can be overlooked in our CI files heightened my sensitivity to this sort of issue.

PR #239, opened 2020-10-23
Adds additional globs in catalogers to handle the root dir instead of just subdirs

### LINKED ISSUE #238: Java cataloger doesn't scan the top-level directory during glob matching for pom and nested archives
**What happened**: Invoke syft on a directory that has a jar, in that jar at the top level is another jar. See anchore/grype#182 All of the nested jars in the winstone.jar embedded in that jar were not cataloged.

**What you expected to happen**: The winstone.jar should be inspected but is not

**How to reproduce it (as minimally and precisely as possible)**: See anchore/grype#182

**Environment**:
- Output of `syft version`: 0.3.0

zhill:
Primary use was java, but found other catalogers needed the same change as well.

Fixes #238

### DIFF
--- a/syft/cataloger/golang/cataloger.go
+++ b/syft/cataloger/golang/cataloger.go
@@ -11,6 +11,7 @@ import (
 func NewGoModCataloger() *common.GenericCataloger {
 	globParsers := map[string]common.ParserFn{
 		"**/go.mod": parseGoMod,
+		"go.mod":    parseGoMod,
 	}
 
 	return common.NewGenericCataloger(nil, globParsers, "go-cataloger")
--- a/syft/cataloger/java/archive_parser.go
+++ b/syft/cataloger/java/archive_parser.go
@@ -17,6 +17,11 @@ import (
 var _ common.ParserFn = parseJavaArchive
 
 var archiveFormatGlobs = []string{
+	"*.jar",
+	"*.war",
+	"*.ear",
+	"*.jpi",
+	"*.hpi",
 	"**/*.jar",
 	"**/*.war",
 	"**/*.ear",
--- a/syft/cataloger/javascript/cataloger.go
+++ b/syft/cataloger/javascript/cataloger.go
@@ -11,6 +11,7 @@ import (
 func NewJavascriptPackageCataloger() *common.GenericCataloger {
 	globParsers := map[string]common.ParserFn{
 		"**/package.json": parsePackageJSON,
+		"package.json":    parsePackageJSON,
 	}
 
 	return common.NewGenericCataloger(nil, globParsers, "javascript-package-cataloger")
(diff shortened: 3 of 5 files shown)


luhring (commented):
This is a good find. Here's my two cents on the implementation details:

1. Outside of the Java package, I'm not seeing the added globs (e.g. `package.json`) lead to finding packages that weren't being found otherwise. If you're seeing a difference, can you share your steps to reproduce?

1. We should have tests that assert that our catalogers are picking up packages like these.

1. It seems like the root cause of the `**/*.jar`-style globs not matching root-level files within Java archives is because names of archive entries surfaced [using `zipReader.Reader.File`](https://github.com/anchore/syft/blob/main/internal/file/zip_file_manifest.go#L54-L56) never begin with a slash (`/`). `**/*.jar` means "zero or more characters, then a slash, then zero or more characters, then '.jar' — thus an entry named "winstone.jar" wouldn't match this glob pattern. In general, the pattern `**/*.jar` _is correct_, because the absolute paths we're matching against always begin with a slash — but for ZIP archive entries, this statement doesn't hold true.

Item 3 also relates to some larger issues with our glob strategy...

- Should we amend found archive entry names to begin with a slash?
- Syft's glob handling differs between its [internal glob implementation](https://github.com/anchore/syft/blob/main/internal/file/glob_match.go#L4) we use for image and archive matching and the [library ("doublestar")](https://github.com/bmatcuk/doublestar) we use for filesystem matching. Most significantly, the two implementations handle a single star (`*`) differently.
- Should Syft be searching within any given archive (e.g. Java archive) for all other kinds of packages? (E.g., find JavaScript dependencies within a web application that's found within a `.war` file.)

In short, I think the implemented change to the patterns in `archiveFormatGlobs` works — but I'd also recommend we address the fragility in our glob approach somewhat soon, lest we have more issues finding files.

zhill:
@luhring yes, I'm fine with a better solution, this just seemed the lowest-risk to resolve the issue directly. I had not seen similar issues in the other catalogers but I have't looked particularly hard. If this is only an issue with how the paths are prepared for the nested detection, then normalizing the paths is probably a better solution to align the use cases.

Re: general archive inspection: Ideally, yes we'd inspect all found archives (jar or otherwise) and do full artifact inspection within those just like an other FS dir. i can file that as another feature/issue to track independently of this change.

luhring:
@zhill That makes sense. If you want to submit the issue re full cataloger inspection of nested archives, that'd be great!

I'll put in a PR to syft that addresses the immediate problem by normalizing the archive entry discovery process rather than by adding more globs, and we can close this PR — that is, unless you object.

zhill:
I'm good with a normalization fix, @luhring. To help track the issue I'll leave this open until the new PR is ready so we can reference this PR from that to keep the discussion connected and then close this

PR #247, opened 2020-10-30
General Java cataloger enhancements

wagoodman:
- Adds `versionAreaPattern` to the java cataloger and attempts to remove version and extra patterns found as opposed to extract the name (since it is assumed that the name is always at the front of the string)
- Does a better job at pairing up the same java packages discovered separately from pom.properties and manifest information, minimizing what appears to be uniquely separate packages with similar names but are really duplicate packages.
- Splits the java manifest by sections, specifically `main` (the first/default section that may not have a `name`) and `namedSections` (all remaining sections). This more closely follows the MANIFEST.MF specification.
- Adds `alpine:3.12.0`, `anchore/test_images:java`, `anchore/test_images:py38`, `anchore/anchore-engine:v0.8.2`, and `jenkins/jenkins:2.249.2-lts-jdk11` images to the compare script in the acceptance tests. The goal was to add java oriented images, however, the `anchore/test_image:java` image is alpine based, so a workaround for compare testing needed to be made to account for anchore-engine's stripping of the release values from the version field. This opened up the ability to add a base alpine image. Previous PRs should have put in the python example image, but were blocked from doing so for similar reasons. The `jenkins` image has proven to be an excellent java image example overall while triaging OSS issues.

Partially addresses https://github.com/anchore/anchore-engine/issues/681 (tasks 2 & 4)


syft/cataloger/java/archive_filename.go
```diff
@@ -10,11 +10,20 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
+// match on versions and anything after the version. This is used to isolate the name from the version.
+// match examples:
```
luhring, line 14: Super helpful

syft/cataloger/java/archive_parser.go
```diff
@@ -156,6 +160,7 @@ func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
 
 // discoverPkgsFromPomProperties parses Maven POM properties for a given parent package, returning all listed Java packages found and
 // associating each discovered package to the given parent package.
+// nolint:funlen,gocognit
```
luhring, line 163: This isn't a request for change, but a note for discussion at some point in the future — I think we might be getting into that stage of the code base's life where functions are slowly entering the large/complex territory. I'd love to see us discuss this as a group at some point and think about ways to extract functions and keep the code as readily understandable as possible

luhring (approved):
This is a lot of great work, very nice!

syft/cataloger/java/archive_parser.go
wagoodman: My default stance when I run into these points is to break it up into smaller helper functions (which I've had to a few times now)... this one seemed rather non-obvious in how to do so, so punted until the new pattern is more obvious. This time I added more comments to help compensate.

luhring: The comments are great. Definitely better than not having them. If you want to talk through this function, just as an exercise, let me know. But to me, it's not something that should block this PR

PR #249, opened 2020-11-04
Resolve security warning for macOS users

luhring:
- Adds signing and notarization to the release process to replace the Darwin `.tar.gz` asset with a `.dmg`asset
- Moves several pipeline steps from a Linux runner to a macOS runner and adapts the pipeline logic accordingly
- Makes minor improvements and adds clarifying comments to various release-related files

This implements https://github.com/anchore/project-toolbox/issues/27 for _syft_. Other released executables will need similar changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,10 @@ help:
 ci-bootstrap:
 	DEBIAN_FRONTEND=noninteractive sudo apt update && sudo -E apt install -y bc jq libxml2-utils
 
+.PHONY:
+ci-bootstrap-mac:
+	github_changelog_generator --version || sudo gem install github_changelog_generator
+
 .PHONY: bootstrap
 bootstrap: ## Download and install all go dependencies (+ prep tooling in the ./tmp dir)
 	$(call title,Bootstrapping dependencies)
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -241,8 +244,7 @@ changelog-release:
 
 .PHONY: changelog-unreleased
 changelog-unreleased: ## show the current changelog that will be produced on the next release (note: requires GITHUB_TOKEN set)
-	@docker run -it --rm  \
-		-v "$(shell pwd)":/usr/local/src/your-app ferrarimarco/github-changelog-generator \
+	@github_changelog_generator \
```
wagoodman, line 247: the `changelog-unreleased` is a target meant for local developers to be able to (optionally) use to see what the changelog would be before cutting a release. By leveraging the docker image dev's don't need to install the right version of ruby, install the gem, and get pathing correct. With the new change that would be required.

(The `changelog-release` target is OK, since this env is setup with `ci-bootstrap-mac`)

wagoodman (approved):
Very nice addition 🚀

Makefile
luhring: Sounds good, reverting this.

PR #250, opened 2020-11-05
Add distro information to JSON presenter

### LINKED ISSUE #169: Add identified distro and version to output for JSON presenter
Add identified distro and version to the JSON output at least, and optionally the text output as well for docker:// analyzed images.

**Why is this needed**:
Identification of the distro is useful for other downstream consumers of the syft SBoM for correlating with other dbs or making decisions on acceptability if distro control is important to the user.

alfredodeza:
Includes distro information that gets parsed in the very beginning of the scanning process down to the JSON presenter for images.

Closes #169 

Sample output (adds `distro` within `target`):

```
$ go run main.go -o json busybox:latest
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged image      [0 packages]

{
 "artifacts": [],
 "source": {
  "type": "image",
  "target": {
   "layers": [
    {
     "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
     "digest": "sha256:d2421964bad195c959ba147ad21626ccddc73a4f2638664ad1c07bd9df48a675",
     "size": 1231733
    }
   ],
   "size": 1231733,
   "digest": "sha256:f0b02e9d092d905d0d87a8455a1ae3e9bb47b4aa3dc125125ca5cd10d6441c9f",
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "tags": [
    "busybox:latest"
   ],
   "distro": {
    "name": "busybox",
    "version": "1.32.0"
   }
  }
 }
}
```


alfredodeza:
I was fully expecting the JSON schema to bark at me and that didn't happen... going through the `syft/schema/README.md` it appears that this is because the schema gets automatically generated?

syft/presenter/json/image.go
```diff
@@ -16,14 +20,23 @@ type Layer struct {
 	Size      int64  `json:"size"`
 }
 
-func NewImage(src scope.ImageSource) *Image {
+type Distro struct {
```
Toure, line 23: Just a nit, but would architecture information be a useful include field?

alfredodeza: That is not available information from the current parsers that provide distro information. In most cases it is looking into `/etc/os-release` which doesn't provide that. For example, this is Debian 8:

```
PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=debian
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support"
```

wagoodman (commented):
It seems like the distro information belongs as a separate top-level item instead of being nested under "source":

```json
{
  "artifacts": [],
  "distro": {
    "name": "busybox",
    "version": "1.32.0"
   },
  "source": {}
}
```

That is, "source" is all about in a raw sense what the image that was cataloged is, outside of any detections made. The other sections are about what was detected from analyzing the contents (in "artifacts").

syft/presenter/json/image.go
Toure: ack.

alfredodeza:
@wagoodman with the latest changes, this is how this looks now:

```json
{
 "artifacts": [],
 "source": {
  "type": "image",
  "target": {
   "layers": [
    {
     "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
     "digest": "sha256:d2421964bad195c959ba147ad21626ccddc73a4f2638664ad1c07bd9df48a675",
     "size": 1231733
    }
   ],
   "size": 1231733,
   "digest": "sha256:f0b02e9d092d905d0d87a8455a1ae3e9bb47b4aa3dc125125ca5cd10d6441c9f",
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "tags": [
    "busybox:latest"
   ]
  }
 },
 "distro": {
  "name": "busybox",
  "version": "1.32.0"
 }
}
```

wagoodman (approved):
Looks great!

PR #257, opened 2020-11-10
Enhance dpkg support

wagoodman:
- Add file list for each deb package, based on the `*.md5sum` dpkg file.
- Add license support for dpkg, derived from `/usr/share/doc/*/copyright` files.
- Add `Maintainer` field
- Adds `Licenses` onto the `Artifact` type for the json presenter

Partially addresses https://github.com/anchore/anchore-engine/issues/682


syft/cataloger/deb/cataloger.go
```diff
@@ -4,14 +4,193 @@ Package dpkg provides a concrete Cataloger implementation for Debian package DB
...
+func NewDpkgdbCataloger() *Cataloger {
+	return &Cataloger{}
+}
+
+// Name returns a string that uniquely describes a cataloger
+func (c *Cataloger) Name() string {
+	return "dpkgdb-cataloger"
+}
+
+// Catalog is given an object to resolve file references and content, this function returns any discovered Packages after analyzing rpm db installation.
```
luhring, line 35: "rpm"

syft/cataloger/deb/cataloger.go
```diff
@@ -4,14 +4,193 @@ Package dpkg provides a concrete Cataloger implementation for Debian package DB
...
+				if ref, ok := copyrightRefsByName[p.Name]; ok {
+					p.Source = append(p.Source, ref)
+				}
+			}
+		}
+	}
+	return pkgs, nil
+}
+
+func (c *Cataloger) fetchMd5Contents(resolver scope.Resolver, dbRef file.Reference, pkgs []pkg.Package) (map[string]io.Reader, map[string]file.Reference, error) {
```
luhring, line 96: It looks like this can be made into just a helper method (like `md5Key`) since it doesn't have a need for its receiver (`c`)

syft/cataloger/deb/cataloger.go
```diff
@@ -4,14 +4,193 @@ Package dpkg provides a concrete Cataloger implementation for Debian package DB
...
+	for ref, contents := range md5ContentsByRef {
+		name := nameByRef[ref]
+		contentsByName[name] = strings.NewReader(contents)
+		refsByName[name] = ref
+	}
+
+	return contentsByName, refsByName, nil
+}
+
+func (c *Cataloger) fetchCopyrightContents(resolver scope.Resolver, dbRef file.Reference, pkgs []pkg.Package) (map[string]io.Reader, map[string]file.Reference, error) {
```
luhring, line 149: It looks like this can be made into just a helper method (like `md5Key`) since it doesn't have a need for its receiver (`c`)

syft/cataloger/deb/cataloger.go
```diff
@@ -4,14 +4,193 @@ Package dpkg provides a concrete Cataloger implementation for Debian package DB
...
+			return nil, fmt.Errorf("unable to find dpkg md5 contents: %w", err)
+		}
+
+		copyrightContentsByName, copyrightRefsByName, err := c.fetchCopyrightContents(resolver, dbRef, pkgs)
+		if err != nil {
+			return nil, fmt.Errorf("unable to find dpkg copyright contents: %w", err)
+		}
+
+		for i := range pkgs {
+			p := &pkgs[i]
```
luhring, line 65: If the only use of `i` is to initially select the individual package, I'd suggest changing the for statement to `for _, p := range pkgs { ...`

syft/cataloger/deb/cataloger.go
```diff
@@ -4,14 +4,193 @@ Package dpkg provides a concrete Cataloger implementation for Debian package DB
...
+
+	var md5FileMatches []file.Reference
+	var nameByRef = make(map[file.Reference]string)
+	parentPath, err := dbRef.Path.ParentPath()
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to find parent of path=%+v: %w", dbRef.Path, err)
+	}
+	for i, p := range pkgs {
+		// look for /var/lib/dpkg/info/NAME:ARCH.md5sums
+		name := md5Key(&pkgs[i])
```
luhring, line 107: Could this be simplified to `for _, p := range pkgs`, and just use `p` instead of `pkgs[i]` (since you already have `p` defined)?

syft/cataloger/deb/cataloger.go
```diff
@@ -4,14 +4,193 @@ Package dpkg provides a concrete Cataloger implementation for Debian package DB
...
+	for ref, contents := range copyrightContentsByRef {
+		name := nameByRef[ref]
+		contentsByName[name] = strings.NewReader(contents)
+		refsByName[name] = ref
+	}
+
+	return contentsByName, refsByName, nil
+}
+
+func md5Key(p *pkg.Package) string {
```
luhring, line 188: Why does `p` need to be a pointer type here?

syft/cataloger/deb/parse_copyright.go
```diff
@@ -0,0 +1,38 @@
...
+				// this is a multi-license summary, ignore this as other recurrent license lines should cover this
+				continue
+			}
+			if candidate != "" && strings.ToLower(candidate) != "none" {
+				findings.Add(candidate)
+			}
+		}
+	}
+
+	results := findings.ToSlice()
```
luhring, line 33: Non-blocking: It looks like most consumers of `ToSlice` end up needing to sort the returned slice immediately after.. this sort could be something worth moving into the implementation of `ToSlice`

syft/cataloger/deb/parse_copyright_test.go
```diff
@@ -0,0 +1,58 @@
...
+			}()
+
+			actual := parseLicensesFromCopyright(file)
+
+			if len(actual) != len(test.expected) {
+				for _, a := range actual {
+					t.Logf("   %+v", a)
+				}
+				t.Fatalf("unexpected package count: %d!=%d", len(actual), len(test.expected))
+			}
```
luhring, line 49: Not necessarily for this PR.... I'm seeing us (not just you) repeating this kind of assertion logic a lot in our code. We do the same `len` test and then call `deep.Equal`. I think we should soon consider finding a way to be more productive (and thus less verbose) as we write these tests.

syft/cataloger/deb/parse_dpkg_status.go
```diff
@@ -2,55 +2,61 @@ package deb
...
+				MetadataType: pkg.DpkgMetadataType,
+				Metadata:     entry,
+			})
+		}
 	}
 
 	return packages, nil
 }
 
+// nolint:funlen,gocognit
```
luhring, line 47: It could be there are ways to factor out meaningful sections of this function as new functions, which would help with reader cognition and function length. For example, the body of the `default` case that begins on line 89

luhring (changes requested):
A lot of great work here! There are a few places it be good to look at, and let me know if it'd be easier to hop on a call

syft/cataloger/deb/parse_copyright_test.go
wagoodman: @alfredodeza pointed this out in a previous review as well. I do get the impending feeling of a new testing strategy in our future.

syft/cataloger/deb/cataloger.go
wagoodman: You can't mutate items retrieved from a range, as they are copies. Instead we are getting a reference to the data to mutate the item in the collection. Alternatively we could get a copy from the collection, mutate the copy, and overwrite the index in the collection with the mutated copy.

syft/cataloger/deb/cataloger.go
wagoodman: I can remove the pointer, it's technically not needed

syft/cataloger/deb/cataloger.go
wagoodman: You can't mutate items retrieved from a range, as they are copies. Instead we are getting a reference to the data to mutate the item in the collection. Alternatively we could get a copy from the collection, mutate the copy, and overwrite the index in the collection with the mutated copy.

syft/cataloger/deb/parse_copyright.go
wagoodman: possibly, sounds like a good improvement --I'll leave that for a separate PR

wagoodman:
@luhring updated to incorporate the suggested changes (leaving https://github.com/anchore/syft/pull/257#discussion_r521393902 for later)

syft/cataloger/deb/cataloger.go
luhring: Gotcha! I missed that we weren't doing something explicit with `p` at the end. Normally I'd recommend modifying (or creating anew) a copy of the element, and then make any modifications in a single, final spot. But this is okay for this PR. 👍

luhring (approved):

Issue #259, opened 2020-11-10
Cataloging python packages errors out if "top_level.txt" entry is missing


wagoodman:
**What happened**:
As [reported by Wagde Zabit](https://anchorecommunity.slack.com/archives/C4QV1RX7Y/p1605029393066600)
```
[0250] ERROR failed to catalog: 1 error occurred:
	* unable to catalog python package=/usr/local/lib/python3.8/site-packages/ptyprocess-0.6.0.dist-info/METADATA: missing python package top_level.txt (package="/usr/local/lib/python3.8/site-packages/ptyprocess-0.6.0.dist-info/METADATA"
```
**What you expected to happen**:
Cataloging should have continued to conclusion with results instead of stopping 

**Environment**:
This was discovered with grype v0.3.0

**Notes**
This is where the problem is https://github.com/anchore/syft/blob/de2e6a13b8138eb751ee87033985dd7e6c39c4fe/syft/cataloger/python/package_cataloger.go#L125-L127 , and should probably only be a log statement instead of returning an error (a best-effort strategy).

wagde-orca:
is it fixed in grype or not yet?

luhring:
@wagde-orca We're actually working on that right this second! We fixed one more thing in Syft, and then will update Grype to include the fix.

luhring:
@wagde-orca Hi there, we just released [grype v0.4.0](https://github.com/anchore/grype/releases/tag/v0.4.0) which should resolve this issue. Let us know if you have any trouble!

PR #260, opened 2020-11-10
Add warning log level

wagoodman:
- Adds `Error` log level helper (mirrors existing pattern of other levels)
- Adds `Warn*` log level helpers
- Replaces a few `log.Errorf` uses that should be `log.Warnf`
- Fixes https://github.com/anchore/syft/issues/259


luhring (approved):
This is great! A much needed step as we evolve our error handling...

PR #263, opened 2020-11-12
Add source to packages found by RPMdb cataloger

wagoodman:
Today we should be attaching the file source to the packages discovered --this PR does this for the RPMDB cataloger.


luhring (approved):
😸

PR #266, opened 2020-11-16
Add JSON document import

wagoodman:
- Adds a helper function to lib.go for parsing the json output document to generate native syft objects (`CatalogFromJSON`) + adds an integration test to cover creating json and interpreting the same json to ensure a comparison continues to work.
- Renames `Scope` to `Source` + renames the `scope` package to `source`
- Moves all implementations in the `resolvers` package to the `source` package
- Migrates all source metadata to a new `source.Metadata` struct and removes the generation of this metadata from the JSON presenter
- Adds more documentation commenting (seemingly randomly... but progress is progress)


syft/cataloger/catalog.go
```diff
@@ -54,6 +54,8 @@ func Catalog(resolver scope.Resolver, catalogers ...Cataloger) (*pkg.Catalog, er
 		log.Debugf("cataloger '%s' discovered '%d' packages", theCataloger.Name(), catalogedPackages)
 		packagesDiscovered.N += int64(catalogedPackages)
 
+		// helper function to add synthesized information...
```
luhring, line 57: What does this comment mean?

syft/lib.go
```diff
@@ -17,23 +17,26 @@ Similar to the cataloging process, Linux distribution identification is also per
...
+	"github.com/anchore/syft/syft/source"
 	"github.com/wagoodman/go-partybus"
 )
 
-// Catalog the given image from a particular perspective (e.g. squashed scope, all-layers scope). Returns the discovered
-// set of packages, the identified Linux distribution, and the scope object used to wrap the data source.
-func Catalog(userInput string, scoptOpt scope.Option) (*pkg.Catalog, *scope.Scope, *distro.Distro, error) {
+// Catalog the given image from a particular perspective (e.g. squashed source, all-layers source). Returns the discovered
+// set of packages, the identified Linux distribution, and the source object used to wrap the data source.
+func Catalog(userInput string, scoptOpt source.Scope) (*pkg.Catalog, *source.Source, *distro.Distro, error) {
```
luhring, line 37: `scoptOpt` — I think we should rename this parameter

syft/lib.go
```diff
@@ -61,24 +64,54 @@ func IdentifyDistro(s scope.Scope) distro.Distro {
...
 		catalogers = cataloger.DirectoryCatalogers()
 	default:
-		return nil, fmt.Errorf("unable to determine cataloger set from scheme=%+v", s.Scheme)
+		return nil, fmt.Errorf("unable to determine cataloger set from scheme=%+v", s.Metadata.Scheme)
 	}
 
 	return cataloger.Catalog(s.Resolver, catalogers...)
 }
 
+// CatalogFromJSON takes an existing syft report and generates catalog primitives.
```
luhring, line 85: Nit: The intended meaning of "catalog primitives" isn't obvious. This function returns a few different things, one of which is `*pkg.Catalog`, and other values as well. It might be worthwhile to spell out what's being returned, or, just state generically that this function deserializes a syft document into native syft objects.

syft/lib.go
```diff
@@ -61,24 +64,54 @@ func IdentifyDistro(s scope.Scope) distro.Distro {
...
+func CatalogFromJSON(reader io.Reader) (*pkg.Catalog, *distro.Distro, source.Metadata, error) {
```
luhring, line 86: Nit: It's not obvious why `source.Metadata` is returned without a pointer, but the same is not true for `*distro.Distro`. Both seem like reasonably lightweight structs, so I might suggest making `*distro.Distro` not a pointer type, and returning the zero-value struct on the "error" return paths (just like what's done for `source.Metadata{}`).

syft/presenter/cyclonedx/bom-extension.go
```diff
@@ -35,15 +34,34 @@ type BdComponent struct {
...
 		},
 	}
+
+	switch srcMetadata.Scheme {
+	case source.ImageScheme:
+		descriptor.Component = &BdComponent{
+			Component: Component{
+				Type:    "container",
+				Name:    srcMetadata.ImageMetadata.UserInput,
+				Version: srcMetadata.ImageMetadata.Digest,
```
luhring, line 54: Not a request for change, but the thought just occurred to me — is it worth checking any other CycloneDX tools to see how they handle this exact case?

syft/presenter/json/artifact.go
```diff
@@ -1,33 +1,135 @@
...
 
 type Artifact struct {
-	Name      string      `json:"name"`
-	Version   string      `json:"version"`
-	Type      string      `json:"type"`
-	FoundBy   []string    `json:"foundBy"`
-	Locations Locations   `json:"locations,omitempty"`
-	Licenses  []string    `json:"licenses"`
-	Metadata  interface{} `json:"metadata,omitempty"`
+	artifactBasicMetadata
```
luhring, line 12: I think it would be good to add a comment on the `Artifact` struct to explain why it exists, to help readers file it away in the right spot in their mental model.

syft/source/directory_resolver.go
```diff
@@ -7,11 +7,12 @@ import (
 	"path"
 	"path/filepath"
 
-	"github.com/anchore/stereoscope/pkg/file"
 	"github.com/anchore/syft/internal/log"
 	"github.com/bmatcuk/doublestar"
 )
 
+var _ Resolver = &DirectoryResolver{}
```
luhring, line 14: Not a necessary change, but just a note that there's also an allocation-free variant of that check that the Go folks recommend: `var _ Resolver = (*DirectoryResolver)(nil)`

luhring (commented):
This is big step forward! 🎉  Just a few comments for review

syft/cataloger/catalog.go
wagoodman: I added this as commentary on a screen share, meant to delete

syft/lib.go
wagoodman: I'll minimize the pointer usage (and while I'm at it order the return values consistently)

syft/source/directory_resolver.go
wagoodman: I always forget this one... good call

syft/presenter/cyclonedx/bom-extension.go
wagoodman: yes

luhring (approved):
🎉

PR #267, opened 2020-11-17
Improve Java version recognition

### LINKED ISSUE #255: Incorrect version parsing from certain java package names in syft 5.0 and newer
**What happened**:

Multiple "versions" are detected in the package name if numbers are used in package names that also contain hyphens. (theory, not verified fact)

Users like grype receive an array of versions where a single version is expected.

A number of internal artifacts trigger this condition, but the following opensource one also does and is likely a good test candidate

This appears to happen in syft 0.5.0 or newer, 0.4.1 does not exhibit this behavior.

```
[0000] ERROR discovered multiple name-version pairings from "spring-data-r2dbc-1.1.0.RELEASE.jar": [map[version:r2] map[version:1.1.0]]
NAME               VERSION        TYPE         
spring-data-r2dbc  1.1.0.RELEASE  java-archive  
```

Same issue via grype

`ERROR discovered multiple name-version pairings from "BOOT-INF/lib/spring-data-r2dbc-1.1.0.RELEASE.jar": [map[version:r2] map[version:1.1.0]] from-lib=syft`

Location of dependency in question, should it be of interest: https://search.maven.org/artifact/org.springframework.data/spring-data-r2dbc

**Environment**:
- Output of `syft version`:

```
Application:   syft
Version:       0.5.1
BuildDate:     2020-11-04T20:55:55Z
GitCommit:     ecfc471ce51a2a84ec144263592c223816241c7a
GitTreeState:  clean
Platform:      linux/amd64
GoVersion:     go1.14.10
Compiler:      gc
```

```
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
```

luhring:
- Closes #255 
- Reworks the regex-based recognition of package name and version from a Java archive name
- Updates test cases to include #255's scenario
- Adjusts existing tests to expect (what seems like) more correct version values
- Cleans up a bit of ancillary test logic

### DIFF
--- a/syft/cataloger/java/archive_filename.go
+++ b/syft/cataloger/java/archive_filename.go
@@ -10,45 +10,100 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
-// match on versions and anything after the version. This is used to isolate the name from the version.
-// match examples:
-// wagon-webdav-1.0.2-rc1-hudson.jar      --->     -1.0.2-rc1-hudson.jar
-// windows-remote-command-1.0.jar         --->     -1.0.jar
-// wstx-asl-1-2.jar                       --->     -1-2.jar
-// guava-rc0.jar                          --->     -rc0.jar
-var versionAreaPattern = regexp.MustCompile(`-(?P<version>(\d+\.)?(\d+\.)?(r?c?\d+)(-[a-zA-Z0-9\-.]+)*)(?P<remaining>.*)$`)
-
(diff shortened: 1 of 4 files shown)


wagoodman (approved):
very nice update

Issue #268, opened 2020-11-17
Add the Common Platform Enumeration (CPE) for the CycloneDX output


geertvanheusden:
**What would you like to be added**:
The Common Platform Enumeration (CPE) which is currently missing in the CycloneDX output

**Why is this needed**:
It makes the current CycloneDX report unusable for tools like Dependency Track which uses the CPE to scan for vulnerabilities.

_Dependency Track related message:_
> The internal analyzer evaluates components against an internal vulnerability database derived from the National Vulnerability Database and VulnDB (if enabled). This analyzer makes use of the Common Platform Enumeration (CPE) defined in components. Components with a valid CPE will be evaluated with this analyzer.

wagoodman:
Given that https://github.com/anchore/syft/pull/271 has been merged, looks like this can get picked up now.

Developer note: don't use [this](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.2.xsd#L304-L311) use [this instead](https://github.com/CycloneDX/specification/blob/master/schema/bom-1.2.xsd#L606-L616)

wagoodman:
On second look, I misinterpreted the XML schema; though the type definition still exists in the schema, the usage within components has been explicitly listed as deprecated (https://github.com/CycloneDX/specification/blob/master/schema/bom-1.2.xsd#L307). 

We could add the CPE, however, it seems potentially ill-advised since Syft generates multiple potential CPEs from the underlying raw data, where as the cyclonedx spec allows for a single value --this could be problematic since it would be a guess.

@geertvanheusden do you know if Dependency Track is planning to keep pulling CPEs from CycloneDX documents in the future (or what they may be using in it's place)?

geertvanheusden:
I just did an additional check and apparently applications with a valid PURL are checked against an internal analyser (NVD) as well. I might have misinterpreted the documentation or it is outdated/incorrect. 
@stevespringett maybe you can elaborate on the correct and current behaviour?

stevespringett:
@geertvanheusden I assume you're referring to Dependency-Track? If so DT has a few different analyzers including an internal analyzer (uses CPE for NVD and VulnDB), NPM Advisories (PURL), and OSS Index (PURL).

Findings from OSS Index may be unique (not in the NVD) or may be CVEs. In DT 3.8 and previous, it was unclear what analyzer was attributed to finding a vulnerability on a component. In DT 4.0, attribution is now specifically supported along with a reference URL to the external source of the finding.

[...]

PR #271, opened 2020-11-19
Move package PURL and CPEs to Package definition

### LINKED ISSUE #269: Migrate CPE generation to syft (and generally output package identities)
Today we generate CPEs within grype, we should migrate this to syft and output this information in JSON and Cyclone presenters

wagoodman:
Today we use CPEs in grype during the matching process, however, this is really a SBOM field being ephemerally created and thrown away --this could be useful in other aspects SBOM-wise.

This PR does the following:
- Adds `CPEs` and `PURL` to the `pkg.Package` struct.
- Does post-processing of packages after the catalogers but before being added to the catalog.
- The above point makes the `syft.Catalog` step dependent on distro, so distro identification has been moved under this call (and removed as an independent action from `lib.go`)
- When `distro.Identify()` cannot identify a distro it now returns `nil`, being consistent with other properties that have the same behavior.
- Unrelated: oraclelinux distro testing was missed previously, so was added here

Notes:
- this PR opens up future work to include CPEs into presenters (https://github.com/anchore/syft/issues/268), but in this PR has only been exposed in the JSON presenter.
- this PR decided to do a "minimal" approach and allow a central spot (within `syft.cataloger`) to be responsible for accumulating purls and CPEs and attaching them to a package object, however, there is probably a better way to do this in the future. One outcome of this was to start considering individual catalogers to be only responsible for discovering and raising up "raw" information from the entity being cataloged, and interpretations of this data is considered "out of scope" for the catalogers and should be attached to the package downstream of the individual catalogers.

Closes #269

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ RESET := $(shell tput -T linux sgr0)
 TITLE := $(BOLD)$(PURPLE)
 SUCCESS := $(BOLD)$(GREEN)
 # the quality gate lower threshold for unit test total % coverage (by function statements)
-COVERAGE_THRESHOLD := 72
+COVERAGE_THRESHOLD := 70
 
 ## Build variables
 DISTDIR=./dist
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/bmatcuk/doublestar v1.3.3
 	github.com/docker/docker v17.12.0-ce-rc1.0.20200309214505-aa6a9891b09c+incompatible
 	github.com/dustin/go-humanize v1.0.0
+	github.com/facebookincubator/nvdtools v0.1.4
 	github.com/go-test/deep v1.0.7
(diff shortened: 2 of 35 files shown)


luhring (approved):
This looks great! The only thing I'm not clear on is all the places that `Distro` was changed to use a pointer... IIRC, this was just recently changed from a pointer type to a non-pointer type. Not a blocking issue for this PR.

wagoodman:
fair question. In https://github.com/anchore/syft/pull/266 the change from pointer to non-pointer of downstream usages of distro.Distro was to eliminate the need for the pointer since the caller should be checking the `distro.Type` explicitly (and there should never be an empty distro). This PR changes the semantics such that `distro.Identify()` can indicate it could not identify the distro via a nil pointer (changing the underlying semantics to exists-or-not-exists, and changing downstream references to need the pointer type, where earlier these downstream pointers were not needed).

Issue #38, opened 2020-06-03
Upload SBOM results to Anchore Engine


wagoodman:
There should be an option to allow for results to be submitted to an Anchore Engine instance via CLI options and configuration.

**Acceptance criteria:**
1. Syft can be invoked with configuration that specifies a means of reaching and authenticating with a deployed instance of Anchore Engine.
1. After Syft exits, the cataloged packages can be observed directly from the instance of Anchore Engine (such as via `anchore-cli image content ...`).

**Steps to test:**
1. Select a container image.
1. Run Syft, passing in as arguments: a reference the container image, the URL to the instance of Anchore Engine, and authentication credentials.
1. Run `anchore-cli image content <image> <content-type>` for various content types to observe that Anchore Engine correctly received and stores Syft's cataloged findings.

**Developer notes:**
1. For context on the inspiration for this functionality and some preliminary implementation thoughts, be sure to check out the [Connecting Toolbox & Enterprise document](https://docs.google.com/document/d/14qoaIN34ZgFJP7zJK6F3XgpRiAihk4psmV9iIG_Ub2c/edit#heading=h.x3vs56j5175n)
1. To be ironed out: what's the expected behavior for uploading analysis for an image (tag) whose analysis has already been stored in Anchore Engine?
1. To be ironed out: schema version vs syft version for validating data shape on upload?

PR #276, opened 2020-11-20
Generate json schema from struct definitions

wagoodman:
Today we are generating our (messy) json schema from a set of examples derived from integration test cases. This process results in a JSON schema where type semantics are lost and merged into a homogenous blob of individual field definitions (instead of a collection of definitions with a set of parameters for each type).

This PR replaces the JSON schema (and the mechanism for generating it) with one that keeps the semantics intact. Specifically, by using [`github.com/alecthomas/jsonschema`](https://github.com/alecthomas/jsonschema) we can now generate the json schema directly from our struct definitions.

There is one caveat: this json schema does not include definitions for anything underneath weakly typed `interface{}` fields, such as `pkg.Package.Metadata`. This is pretty notable, but something that can be improved on in the future (either by tweaking this automated approach, or by replacing it with a manual approach).

Partially addresses #38


test/integration/document_import_test.go
```diff
@@ -82,6 +82,11 @@ func TestCatalogFromJSON(t *testing.T) {
 				}
 
 				for _, d := range deep.Equal(a, e) {
+					// ignore errors for empty collections vs nil for select fields
+					// TODO: this is brittle, but not dangerously so. We should still find a better way to do this.
```
luhring, line 86: It's unclear to me why this is here, it'd be good to talk this through and/or add more explanation in the comment

Makefile
```diff
@@ -153,17 +153,9 @@ fixtures:
...
-		--rm \
-		-v $(shell pwd)/schema/json:/work \
-		-w /work \
-		python:3.8 \
-			bash -x -c "\
-				pip install -r requirements.txt && \
-				python generate.py \
-			"
+generate-json-schema:  ## Generate a new json schema
+	pushd schema/json
```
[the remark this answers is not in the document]
wagoodman: I can change it to cd... pushd is just a habit

test/integration/document_import_test.go
wagoodman: we have cases where the licenses are still nil slices on the package object, something that the presenter now normalizes to an empty slice instead during presentation, but technically they are not equivalent when doing comparison testing for the serialization loop (make package -> get json -> import json -> get packages -> compare packages against original packages). This ignores a select case of the described behavior for a single field.

luhring: Okay, interesting. So this test is effectively testing several processes — normal cataloging, JSON presenters, and the new document important logic. Just so I can better understand, with regard to the document important logic, what are we verifying that wouldn't be caught with tests whose only notable function call was `CatalogFromJSON`? It seems like we could have a JSON document test fixture, and then assert that the returned values (e.g. distro, source metadata, catalog packages) are what we expect.

Btw, these could easily be notes for enhancement after this PR. I'm just trying to better understand what's going on in this code. And I think I'm coming to the realization that the empty slice vs. nil slice issue is more of a symptom than a root issue.

For now, I think it's fine to explain in the comment that this "exception" to the equality assertion is due to everything that's happening in the "serialization loop" (like you describe above), specifically that this test scope introduces the need to account for data changes in the JSON presenter logic.

luhring (approved):

wagoodman:
@luhring re: [comments](https://github.com/anchore/syft/pull/276#discussion_r528240430) . 

> It seems like we could have a JSON document test fixture...

Yes. However, this would be brittle in another way. 

As the shape of the JSON document changes, we'll need to update the JSON test fixture continually. We do this for the JSON presenter since the **unit** tests should be ensuring values are correct (which we do with a diff of the json values and depend on folks updating the snapshot with a scrutinizing eye). We already have a set of [curated test cases](https://github.com/anchore/syft/blob/main/test/integration/pkg_cases_test.go) which are [guaranteed to cover all package types](https://github.com/anchore/syft/blob/main/test/integration/pkg_coverage_test.go#L94-L99), which provides a greater variety of data testing coverage than what we do in the unit tests for the JSON presenters... using these cases along with the real JSON presenter output as test input casts a much wider net for being alerted to potential import problems that may arise.

> the empty slice vs. nil slice issue is more of a symptom than a root issue...

Ack. I came to the same conclusion, and didn't want to go to the trouble of fixing that behavior since it technically isn't "wrong" and would take more time.

test/integration/document_import_test.go
wagoodman: > It seems like we could have a JSON document test fixture...

Yes. However, this would be brittle in another way. 

As the shape of the JSON document changes, we'll need to update the JSON test fixture continually. We do this for the JSON presenter since the **unit** tests should be ensuring values are correct (which we do with a diff of the json values and depend on folks updating the snapshot with a scrutinizing eye). We already have a set of [curated test cases](https://github.com/anchore/syft/blob/main/test/integration/pkg_cases_test.go) which are [guaranteed to cover all package types](https://github.com/anchore/syft/blob/main/test/integration/pkg_coverage_test.go#L94-L99), which provides a greater variety of data testing coverage than what we do in the unit tests for the JSON presenters... using these cases along with the real JSON presenter output as test input casts a much wider net for being alerted to potential import problems that may arise.

> the empty slice vs. nil slice issue is more of a symptom than a root issue...

Ack. I came to the same conclusion, and didn't want to go to the trouble of fixing that behavior since it technically isn't "wrong" and would take more time.

PR #279, opened 2020-12-01
Include CPEs with elements from POM GroupId fields

wagoodman:
This will help solve some of the grype false negatives for java packages highlighted in https://github.com/anchore/grype/issues/209 , where we could be guessing CPEs to search with (in the grype matching process) with parsed fields from the `GroupID` POM property. 

Specifically `org.sonatype.nexus` --> `cpe:2.3:a:sonatype:nexus:*:*:*:*:*:*:*:*`.

### DIFF
--- a/syft/cataloger/cpe.go
+++ b/syft/cataloger/cpe.go
@@ -2,6 +2,7 @@ package cataloger
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/syft/pkg"
@@ -31,6 +32,7 @@ func generatePackageCPEs(p pkg.Package) []pkg.CPE {
 
 				// add a new entry...
 				candidateCpe := wfn.NewAttributesWithAny()
+				candidateCpe.Part = "a"
 				candidateCpe.Product = product
 				candidateCpe.Vendor = vendor
 				candidateCpe.Version = p.Version
@@ -45,8 +47,6 @@ func generatePackageCPEs(p pkg.Package) []pkg.CPE {
 }
 
 func candidateTargetSoftwareAttrs(p pkg.Package) []string {
-	// TODO: expand with package metadata (from type assert)
-
 	// TODO: would be great to allow these to be overridden by user data/config
 	var targetSw []string
 	switch p.Language {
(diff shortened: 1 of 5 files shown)


syft/cataloger/cpe.go
```diff
@@ -68,14 +68,43 @@ func candidateTargetSoftwareAttrs(p pkg.Package) []string {
...
 func candidateVendors(p pkg.Package) []string {
-	// TODO: expand with package metadata (from type assert)
-	vendors := []string{p.Name}
-	if p.Language == pkg.Python {
+	vendors := candidateProducts(p)
+	switch p.Language {
+	case pkg.Python:
 		vendors = append(vendors, fmt.Sprintf("python-%s", p.Name))
+	case pkg.Java:
+		if p.MetadataType == pkg.JavaMetadataType {
```
luhring, line 76: Suggestion: I noticed this case block goes to 4 levels of indentation — perhaps this could be factored out into a function or two that describes the behavior, e.g. `vendorFromPomProperties` or something.

syft/cataloger/cpe.go
```diff
@@ -68,14 +68,43 @@ func candidateTargetSoftwareAttrs(p pkg.Package) []string {
...
 	}
 	return vendors
 }
 
 func candidateProducts(p pkg.Package) []string {
-	return []string{p.Name}
+	var products = []string{p.Name}
+	switch p.Language {
+	case pkg.Java:
+		if p.MetadataType == pkg.JavaMetadataType {
```
luhring, line 95: Hey again 😄 Now that I'm seeing this second instance of the same conditional blocks, I think we really should consider factoring this out, and perhaps in a way that elegantly exposes the fields of interest (i.e. vendor and product)

syft/cataloger/cpe_test.go
```diff
@@ -119,19 +159,24 @@ func TestGenerate(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			actual := generatePackageCPEs(test.p)
 
-			if len(actual) != len(test.expected) {
-				for _, e := range actual {
-					t.Errorf("   unexpected entry: %+v", e.BindToFmtString())
-				}
-				t.Fatalf("unexpected number of entries: %d", len(actual))
+			expectedCpeSet := set.NewStringSet(test.expected...)
```
luhring, line 162: How are you liking this set package so far? Do you see this as being something we could consider using in place of our other internal string set implementations?

luhring (approved):
Nice enhancement! Check out the suggestions, and let me know if you want to discuss

syft/cataloger/cpe_test.go
wagoodman: 1000% want to replace ours with this

wagoodman:
@luhring  re: https://github.com/anchore/syft/pull/279#discussion_r533415393 and https://github.com/anchore/syft/pull/279#discussion_r533414292 I noticed the same things and also considered breaking into smaller parts or changing to a different abstraction --but ultimately decided not to... yet. The pattern of breaking into smaller functions would get pretty sprawled/ugly over time as well (swapping ugly for ugly in some senses). I also see that we could be making better decisions about CPEs if each field was not done in isolation (not only for the duplicate parsing that you pointed out, but also in considering these two fields together during parsing)... this hints to a different approach than whats here now (a somewhat orthogonal approach data-path-wise).

I decided to leave these as is for now, knowing it will get refactored later.

Issue #280, opened 2020-11-30
Cannot parse field from line: '======='


luoyimu1:
**What happened**:

[0012] ERROR failed to catalog: 1 error occurred:
* unable to catalog python package=/usr/lib/python3/dist-packages/filelock-3.0.10.egg-info/PKG-INFO: cannot parse field from line: '======='

**What you expected to happen**:
this should not be an error..
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
Dockerfile:
FROM python:3.9-slim AS base

ARG PIP_NO_CACHE_DIR=1

RUN apt-get update && apt-get install -y \
    git \
 && rm -rf /var/lib/apt/lists/*

RUN apt-get update && apt-get install -y \
    build-essential \
    python3-dev \
    libldap2-dev \
    libsasl2-dev \
    ldap-utils \
    tox \
    lcov \
    valgrind \
 && rm -rf /var/lib/apt/lists/*

RUN pip install -U pip pipenv

ADD Pipfile* /tmp/
WORKDIR /tmp
RUN pipenv install --system --ignore-pipfile

ADD src/ /app/

WORKDIR /app
RUN useradd -m -u 777 mrdocker \
    && chown mrdocker:mrdocker /app

##----------------------------------------------------------------------

FROM base AS runtime

USER mrdocker

LABEL maintainer="Security"

ENTRYPOINT ["python3", "main.py"]

**Environment**:
- Output of `grype version`:0.5.0
- OS (e.g: `cat /etc/os-release` or similar): Linux

luhring:
>Separately, the PKG-INFO file in 3.0.12 has another characteristic that Syft might not be handling correctly — specifically, the contents of a markdown file (the packages README.md) is appended to the end of the file without a specifying field. We're looking into the spec to determine if this is correct or also a malformation.

It turns out **Syft does handle this case correctly**. (Syft parsing should end for a given file when it encounters a new blank line, and it does).

luoyimu1:
Hey @luhring,
Thanks a lot for the info:)
From the docker log i can see the initial installed version is 3.0.10 which caused the issue here, and it updates to 3.0.12 during "pip install -U pip pipenv" stage, but grype still catch up with the 3.0.10 PKG-INFO (I couldn't find 3.0.12 PKG-INFO in the filesystem either).
Do you have any suggestions on how we could possibly fix it?
Cheers,

PR #281, opened 2020-12-01
Clarify comment on parsing new lines and factor out function

luhring:
- Clarifies comment on why an empty line in a `METADATA` or `PKG_INFO` file means that we cease parsing, with a reference to the spec
- Factors out field continuation logic into new function (I wasn't originally intending to do this, but the above comment clarification pushed the function's length past the linter's threshold, and it seemed like a good time to do some small cleanup.)

(This was investigated originally due to https://github.com/anchore/syft/issues/280.)

Signed-off-by: Dan Luhring <dan.luhring@anchore.com>

### DIFF
--- a/syft/cataloger/python/parse_wheel_egg_metadata.go
+++ b/syft/cataloger/python/parse_wheel_egg_metadata.go
@@ -23,29 +23,28 @@ func parseWheelOrEggMetadata(path string, reader io.Reader) (pkg.PythonPackageMe
 		line := scanner.Text()
 		line = strings.TrimRight(line, "\n")
 
-		// empty line indicates end of entry
+		// An empty line means we are done parsing (either because there's no more data,
+		// or because a description follows as specified in
+		// https://packaging.python.org/specifications/core-metadata/#description;
+		// and at this time, we're not interested in the description).
 		if len(line) == 0 {
-			// if the entry has not started, keep parsing lines
-			if len(fields) == 0 {
-				continue
+			if len(fields) > 0 {
+				break
 			}
-			break
+
+			// however, if the field parsing has not started yet, keep scanning lines
+			continue
 		}
 
 		switch {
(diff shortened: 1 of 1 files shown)


alfredodeza (approved):

syft/cataloger/python/parse_wheel_egg_metadata.go
```diff
@@ -71,8 +70,23 @@ func parseWheelOrEggMetadata(path string, reader io.Reader) (pkg.PythonPackageMe
...
-	sitePackagesRoot := filepath.Clean(filepath.Join(filepath.Dir(path), ".."))
-	metadata.SitePackagesRootPath = sitePackagesRoot
+	metadata.SitePackagesRootPath = filepath.Clean(filepath.Join(filepath.Dir(path), ".."))
 
 	return metadata, nil
 }
+
+// handleFieldBodyContinuation returns the updated value for the specified field after processing the specified line.
+// If the continuation cannot be processed, it returns an error.
+func handleFieldBodyContinuation(key, line string, fields map[string]string) (string, error) {
```
wagoodman, line 80: I take it the linter rules nudged this above the function length due to added comments?...

wagoodman (approved):

syft/cataloger/python/parse_wheel_egg_metadata.go
luhring: @wagoodman Yep, exactly — see the PR description

wagoodman: if only there was a spot in a PR where I could read about context... 🤔 ...GREAT SCOTT!! When did they add PR descriptions??!! (...my bad)

luhring: great scott

PR #282, opened 2020-12-02
Corrected syft cyclonedx generated bom to adhere to the specifications found in CycloneDX 1.2

### LINKED ISSUE #270: The extension part (metadata) in the output is not according to the CycloneDX specifications
**What happened**:

The CycloneDX output format is invalid according to the specs. Version 1.2 of the CycloneDX specification should use the `<metadata>` tag instead of the extension. The snippet (current output of syft) below demonstrates this. 

<?xml version="1.0" encoding="UTF-8"?>
<bom xmlns="http://cyclonedx.org/schema/bom/1.2" xmlns:bd="http://cyclonedx.org/schema/ext/bom-descriptor/1.0" version="1" serialNumber="urn:uuid:68e4a7c5-dc89-482d-9fd3-223bd8002b2e">
  <components>
 ...
 </components>
 <bd:metadata>
    <bd:timestamp>2020-11-18T10:11:12+01:00</bd:timestamp>
    <bd:tool>
      <bd:vendor>anchore</bd:vendor>
      <bd:name>syft</bd:name>
      <bd:version>0.8.0</bd:version>
    </bd:tool>
    <bd:component type="container">
      <name>image-xyz</name>
      <version>sha256:cbdee75f57018631712246e91054b89fc6ec3f13215cf00a749256e80bd67ddd</version>
    </bd:component>
  </bd:metadata>
</bom>

The output should have a valid structure according to the specs. More info and an example can be found over here:
* https://cyclonedx.org/ext/bom-descriptor/
* https://cyclonedx.org/use-cases/#packaging-and-distribution

By using the CycloneDX output format with syft on an image. 

I noticed this issue while importing the CycloneDX BOM into Dependency Track.
A ticket was incorrectly reported over here https://github.com/DependencyTrack/dependency-track/issues/841.

Toure:
Currently, Syft does not generate a cyclonedx sbom which follows the specifications found in [https://cyclonedx.org/use-cases/#packaging-and-distribution](CycloneDX Packaging and Distribution).
This change brings the generated output up to date according to the described specification. 

Closes #270


wagoodman (approved):
looks good!

alfredodeza:
The fixes look OK to me, but issue #270 raises problems that I don't see addressed with these changes. How did this pass validation before? I'm not seeing what validates that the actual problem is fixed

Toure:
> The fixes look OK to me, but issue #270 raises problems that I don't see addressed with these changes. How did this pass validation before? I'm not seeing what validates that the actual problem is fixed

So the pass validation worked due to the output was of the correct shape for bom-descriptor 1.0 [https://github.com/CycloneDX/specification/blob/master/schema/ext/bom-descriptor-1.0.xsd](CycloneDX bom-descriptor 1.0)

alfredodeza (approved):

PR #285, opened 2020-12-03
Add enterprise upload capability

### LINKED ISSUE #38: Upload SBOM results to Anchore Engine

wagoodman:
Allows for a user to upload the SBOM results to enterprise:
```
syft <image> -h <enterprise host> -u <username> -p <password>
```

These options can also be provided via configuration or environment variables:

```
export SYFT_ANCHORE_HOSTNAME=https://somehost
export SYFT_ANCHORE_USERNAME=joe
export SYFT_ANCHORE_PASSWORD=foobar

syft <image>
# analysis + upload
```

If done this way, the user can keep the config / env vars set and still control if upload is enabled:

```
export SYFT_ANCHORE_HOSTNAME=https://somehost
...
export SYFT_ANCHORE_UPLOAD_ENABLED=false

syft <image>
# analysis and no upload
```

Additionally a user can optionally provide a dockerfile as well:

```
syft <image> -h <host> --dockerfile <path to dockerfile>
```

What's remaining:
- [x] ETUI and logging output needs to be added to reflect the upload process / status / result
- [x] Possibly add a json schema version indicated to enterprise during upload (and handle gracefully the "deny" path here)
- [x] Should we upload docker history when available if no dockerfile is available? This requires coordination with engine on the data shape here (or how to interpret otherwise) --the config endpoint was added
- [x] README update with the new options --only config options were added, further instructions will be added closer to the Enterprise release.

Closes #38


luhring:
One small comment on UX — would it be possible to change `-h` to `-H` (or something) for the host argument? I think a good number of folks assume they can rely on `-h` for help (and this is currently true for our tools).

wagoodman:
@luhring good call, let's change this to `-H`

cmd/root.go
```diff
@@ -106,6 +121,55 @@ func startWorker(userInput string) <-chan error {
...
+		}
+	}
+
+	var scheme string
+	var hostname = appConfig.Anchore.Host
+	urlFields := strings.Split(hostname, "://")
+	if len(urlFields) > 1 {
+		scheme = urlFields[0]
+		hostname = urlFields[1]
+	}
```
luhring, line 155: (There's also built-in functionality for parsing URLs in the package `net/url` in the standard library)

internal/anchore/import.go
```diff
@@ -0,0 +1,128 @@
...
+			Stager:       progress.Stager(stage),
+			Progressable: prog,
+		}),
+	})
+
+	return stage, prog
+}
+
+// nolint:funlen
+func (c *Client) Import(ctx context.Context, imageMetadata image.Metadata, s source.Metadata, catalog *pkg.Catalog, d *distro.Distro, dockerfile []byte) error {
```
luhring, line 42: For later — several of these values (such as `image.Metadata`, `source.Metadata`, `pkg.Catalog`, and `distro.Distro`) seem to appear together often, both as function parameters and as return values. It may be worthwhile to merge these values into a single data structure (like a `SyftResult` object or something), which would simplify several function signatures, as well as return lines (I've noticed several return lines like `return nil, nil, nil, fmt.Errorf(...)`)

internal/anchore/import.go
```diff
@@ -0,0 +1,128 @@
...
+		}{
+			Stager:       progress.Stager(stage),
+			Progressable: prog,
+		}),
+	})
+
+	return stage, prog
+}
+
+// nolint:funlen
```
luhring, line 41: This is a good use of this `nolint`. I remember the discussions around the approach to importing each element. I think this calls attention to the length of the function, which may help motivate us (as a team) toward a more elegant approach going forward (in line with what you and I were thinking).

luhring (commented):
@wagoodman Looks good, no objections to the merge! A few things to discuss if you want.

Issue #228, opened 2020-10-16
Docker image


jeff-cook:
**What would you like to be added**:

A published Docker image to run syft.

**Why is this needed**:

With a published Docker image it would simplify usage in most CI environments. Without a published Docker image syft is installed every time it runs within a CI environment.

**Additional context**:

davidkarlsen:
I see the PR was merged, but no tags available at: https://hub.docker.com/r/anchore/syft ?

luhring:
Hi @davidkarlsen — one related PR was merged, but the primary PR has not been merged (#310). Long story short, this required more work than it seemed at first glance, and we ended up needing to make substantial changes to our release pipeline. (More details in that PR description.)

We're hoping to pick this back up shortly and begin publishing images to Docker Hub.

PR #288, opened 2020-12-09
Add the ability to run syft from a scratch image.

Toure:
This change will allow endusers or CI to run syft from a
minimum image which will simplify CI deployment.

closes: #228 
Signed-off-by: Toure Dunnon <toure.dunnon@anchore.com>


Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN apk --no-cache add ca-certificates
+
+# create empty directory for scratch image cache.
+RUN mkdir -p /tmp-syft
+
+# reduce container image to scratch size.
+FROM scratch
+
+# Copy directories and files needed to execute syft.
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
luhring, line 13: What do we use these certs for?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY --from=build /tmp-syft /tmp
```
luhring, line 14: One option here would be to create the `/tmp` directory using `WORKDIR`. There's an argument to be made that this is a bit of a hack. It'd also simplify the Dockerfile and possibly remove the need for the multi-stage approach altogether. Example:

```Dockerfile
FROM scratch

WORKDIR /tmp  # used only to create this directory
WORKDIR /

COPY ./syft /

ENTRYPOINT [ "/syft" ]
```

luhring (commented):
Nice work! I'd also suggest that adding documentation for how to use this Docker image is a "must have" for this addition, since it may not be obvious how to use it (e.g. that most workflows would require access to the Docker daemon).

Dockerfile
wagoodman: when syft checks for updates it does so over https, these certs are needed for verifying this trust.

wagoodman:
ditto on @luhring s comment for documentation, just a small example or two in the readme will go a long way.

Dockerfile
wagoodman: IMHO since the multi-stage approach is still required the `COPY` approach is OK (and is generally more explicit)

luhring: Agreed

README.md
```diff
@@ -69,6 +69,16 @@ brew tap anchore/syft
 brew install syft
 ```
 
+**Container Image**
+
+Example of how to use Syft scratch image:
```
wagoodman, line 74:
```suggestion
Example of how to use Syft with docker:
```

README.md
```diff
@@ -69,6 +69,16 @@ brew tap anchore/syft
...
+
+
+```bash
+docker pull anchore/syft:latest
```
wagoodman, line 78:
```suggestion

```
technically docker will attempt to pull the image, so saves example space

README.md
```diff
@@ -69,6 +69,16 @@ brew tap anchore/syft
...
+docker run -v /var/run/docker.sock:/var/run/docker.sock anchore/syft -o json alpine:latest
```
wagoodman, line 79:
```suggestion
docker run -v /var/run/docker.sock:/var/run/docker.sock anchore/syft:latest -o json alpine:latest
```

wagoodman: also, should we throw in `-t` to get the ETUI output? or leave as is for logger output? (open question)

README.md
```diff
@@ -69,6 +69,16 @@ brew tap anchore/syft
 brew install syft
 ```
 
+**Container Image**
```
wagoodman, line 72:
```suggestion
**Docker**
```

README.md
wagoodman: (in this way we just use the bolded title/section to hint at the installation mode)

README.md
alfredodeza: Why do we want to say Docker instead of container image?

README.md
Toure: So I added the "-t" or Allocate a pseudo TTY option and the terminal hung, which is odd.

README.md
Toure: I think container image is a better term as this will allow us to include other engines in the future.

wagoodman: actually, on further consideration, I retract this...

README.md
wagoodman: gotcha, we'll leave this as is 👍  thanks for confirming

wagoodman (approved):
looks great!

PR #289, opened 2020-12-11
Sort generated CPEs by specificity

luhring:
- Sorts generated CPEs by specificity (measured by number of CPE fields that **aren't** set to `*`)
- Makes progress toward the implementation mentioned in @wagoodman's [comment](https://github.com/anchore/syft/issues/268#issuecomment-741829842) on #268

### DIFF
--- a/syft/cataloger/cpe.go
+++ b/syft/cataloger/cpe.go
@@ -2,6 +2,7 @@ package cataloger
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


wagoodman (approved):
I think this is worth while getting in, I think we can use this as a start to improve the CPE generation process and CPE priority as well.

PR #290, opened 2020-12-11
Improve performance of the python cataloger

wagoodman:
Batches the file content requests for the python cataloger facilitated by a new `source.ContentRequester` object. For images with several python packages installed this sped up the cataloging time by a non-trivial amount (e.g. for `anchore/anchore-engine` the start-to-finish time dropped from 20 seconds to 3.5 seconds, a ~450% improvement).


syft/cataloger/python/package_cataloger.go
```diff
@@ -32,33 +38,87 @@ func (c *PackageCataloger) Name() string {
...
+
+	// find all primary record paths
+	matches, err := resolver.FilesByGlob(eggMetadataGlob, wheelMetadataGlob)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find files by glob: %w", err)
+	}
+	metadataLocations = append(metadataLocations, matches...)
+
+	// for every primary record path, craft all secondary record paths and build a request object to gather all file contents for each record
+	var requester = source.NewContentRequester()
```
luhring, line 71: I noticed this introduces both the `var x = ...` and `x := ...` approaches, is there a reason for the differentiation?

syft/cataloger/python/package_cataloger.go
```diff
@@ -18,6 +18,12 @@ const (
 	wheelMetadataGlob = "**/*dist-info/METADATA"
 )
 
+type pythonPackageData struct {
```
luhring, line 21: This piece of data seems to be referred to as an "entry" ubiquitously in the code — it may be worthwhile to rename the type to include "entry" for clarity as it's being used

syft/cataloger/python/package_cataloger.go
```diff
@@ -32,33 +38,87 @@ func (c *PackageCataloger) Name() string {
...
+	}
+	metadataLocations = append(metadataLocations, matches...)
+
+	// for every primary record path, craft all secondary record paths and build a request object to gather all file contents for each record
+	var requester = source.NewContentRequester()
+	var entries = make([]*pythonPackageData, len(metadataLocations))
+	for i, metadataLocation := range metadataLocations {
+		// we've been given a file reference to a specific wheel METADATA file. note: this may be for a directory
+		// or for an image... for an image the METADATA file may be present within multiple layers, so it is important
+		// to reconcile the RECORD path to the same layer (or the next adjacent lower layer). The same is true with
```
luhring, line 76: "or the next adjacent lower layer"

I'm trying to make sure I understand what this means. This is the first layer to include the path as you look at the current layer and each next lower layer?

syft/cataloger/python/package_cataloger.go
```diff
@@ -32,33 +38,87 @@ func (c *PackageCataloger) Name() string {
...
+
+		if topLevelLocation != nil {
+			entry.TopPackage = &source.FileData{
+				Location: *topLevelLocation,
+			}
+			requester.Add(entry.TopPackage)
 		}
+
+		// keep the entry for processing later
+		entries[i] = entry
```
luhring, line 112: This logic in the body of the `for` loop could potentially be factored out into a dedicated function

syft/source/content_requester.go
```diff
@@ -0,0 +1,48 @@
+package source
+
+import "sync"
+
+type ContentRequester struct {
+	request map[Location][]*FileData
+	lock    sync.Mutex
+}
+
+func NewContentRequester(data ...*FileData) *ContentRequester {
```
luhring, line 10: It looks like arguments aren't being supplied to this function. Do you see an immediate case where they will be? Would it make sense to defer the idea of taking parameters until they're needed?

syft/source/content_requester.go
```diff
@@ -0,0 +1,48 @@
...
+	requester := &ContentRequester{
+		request: make(map[Location][]*FileData),
+	}
+	for _, d := range data {
+		requester.Add(d)
+	}
+	return requester
+}
+
+func (b *ContentRequester) Add(data *FileData) {
```
luhring, line 20: It might make sense to have this not be a pointer. This would reduce the number of possible states in the system. It also looks like consumers of this function are having to take a reference to data they already have just to be able to call this function.

luhring (commented):
Amazing performance improvement! Let me know if you want to chat through any suggestions.

syft/cataloger/python/package_cataloger.go
wagoodman: no, I can make this more consistent

syft/cataloger/python/package_cataloger.go
wagoodman: this is in regards to using `RelativeFileByPath` which provides the "closest" path relative to the given file (with respect to the current layer or lower layers). I can adjust the existing comment.

syft/cataloger/python/package_cataloger.go
wagoodman: yes, I can make a helper function called `getPythonPackageEntry` which would improve readability a bit

syft/source/content_requester.go
wagoodman: Convenience for allowing something like `NewContentRequester(a_slice...).Execute(resolver)`, though this form is not used currently.

syft/source/content_requester.go
wagoodman: it is important that this remains a pointer since this requester is writing back results to the given `FileData` object upon `Execute()` (this approach will not work otherwise).

syft/source/content_requester.go
luhring: 👍 (We discussed keeping this for now, and seeing if we begin to use this code in the near future.)

syft/source/content_requester.go
luhring: 👍 (We discussed keeping this for now, and keeping an eye out for an approach that can avoid writing back to a given pointer.)

luhring (approved):
Looking very good!!

(just needs a rebase)

PR #298, opened 2020-12-18
Update dpkg license to only include single-word entries

wagoodman:
Deb package processing should only be capturing single-value licenses (not full license texts). This moves this processing in the right direction, however, the underlying data may still be mis-using the field.


luhring (approved):

dakaneye (approved):
LGTM

PR #301, opened 2020-12-22
catalogers: Python runtime is not a Python package itself, ignore it

### LINKED ISSUE #300: Python binary is not a Python package
The Python 2.7 runtime defines itself on an `egg-info` file which Syft properly catalogs and reports back as a Python _package_ even though it is the runtime. This happens because the `egg-info` is the one of the sources of truth for cataloging.

The fix should be special casing `Python` so that it can get omitted from the Python catalog.

Screen Shot 2020-12-22 at 9 27 38 AM

alfredodeza:
Ignores if the Python runtime is found as a package. Adds one test to verify this is the case as well as a fixture

Closes #300

### DIFF
--- a/syft/cataloger/python/package_cataloger.go
+++ b/syft/cataloger/python/package_cataloger.go
@@ -91,6 +91,12 @@ func (c *PackageCataloger) catalogEggOrWheel(entry *packageEntry) (*pkg.Package,
 		return nil, err
 	}
 
+	// This can happen for Python 2.7 where it is reported from an egg-info, but Python is
+	// the actual runtime, it isn't a "package". The special-casing here allows to skip it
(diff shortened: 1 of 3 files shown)


syft/cataloger/python/package_cataloger.go
```diff
@@ -91,6 +91,12 @@ func (c *PackageCataloger) catalogEggOrWheel(entry *packageEntry) (*pkg.Package,
 		return nil, err
 	}
 
+	// This can happen for Python 2.7 where it is reported from an egg-info, but Python is
```
luhring, line 94: I *think* it's all Python 2.x — in case it helps to clarify here

luhring (approved):

PR #303, opened 2020-12-22
Handle site packages based on which egg file is parsed

luhring:
- Accounts for the possibility of the parsed file's parent directory being `site-packages` itself (a possibility that was introduced in https://github.com/anchore/syft/pull/296) as opposed to an `...egg-info` directory
- Adds a new resolver: `MockResolver` — this can be used when you need a `source.Resolver` but you want to explicitly only use a set of test fixture files

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -31,7 +31,6 @@ linters:
     - misspell
     - nakedret
     - nolintlint
-    - prealloc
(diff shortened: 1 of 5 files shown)


alfredodeza (approved):
This looks good but would love at least one test if possible

wagoodman (approved):

PR #304, opened 2021-01-04
Fix symlink resolutions for constituent paths

wagoodman:
Pulls in link resolution fix from https://github.com/anchore/stereoscope/pull/54


luhring (approved):
LGTM

PR #312, opened 2021-01-22
Omit empty packages

### LINKED ISSUE #311: Partial package.json files lead to empty packages in output
**What happened**:

Some `package.json` files contain very little information, and specifically, they don't have a `name` or `version` field. This is seems to be considered allowable for unpublished packages. (See related: https://github.com/microsoft/rushstack/issues/2070).

For example, here's an entire `package.json` file we found:

```json
{
  "sideEffects": false,
  "module": "../../esm/fp/isSaturday/index.js",
  "typings": "../../typings.d.ts"
}
```

This results in Syft decoding the JSON to a `PackageJSON` without error, which results in an empty and unusable package being returned in the end.

We should enforce that if a parsed `PackageJSON` is missing the `name` or `version` fields, it should not be included in the final output. This doesn't need to be an error, but we should consider expressing this in a more detailed level of log output.

```
syft docker.io/anchore/enterprise-ui:latest -o json | jq '[ .artifacts[] | select(.name == "" or .version == "") ]'
```

**Environment**:
- Output of `syft version`:

```
Application:   syft
Version:       0.12.2
BuildDate:     2021-01-06T02:43:56Z
GitCommit:     0f6288881bf8db287e07576bf8dd067184745759
GitTreeState:  clean
Platform:      darwin/amd64
GoVersion:     go1.14.13
Compiler:      gc
```

luhring:
Closes #311 

This PR changes the parsing of `package.json` such that files that don't include the minimally required fields (currently `name` and `version`) will not be surfaced as packages. A `DEBUG` level message is logged in this case.

### DIFF
--- a/syft/cataloger/javascript/parse_package_json.go
+++ b/syft/cataloger/javascript/parse_package_json.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/anchore/syft/internal/log"
 	"io"
 	"regexp"
 
@@ -172,6 +173,12 @@ func parsePackageJSON(_ string, reader io.Reader) ([]pkg.Package, error) {
 			return nil, fmt.Errorf("failed to parse package.json file: %w", err)
 		}
 
+		if !p.hasMinimumRequiredValues() {
+			log.Debug("encountered package.json file without the minimum number of field values required for" +
+				" consideration as a package")
+			return nil, nil
+		}
+
 		licenses, err := licensesFromJSON(p)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse package.json file: %w", err)
@@ -194,4 +201,8 @@ func parsePackageJSON(_ string, reader io.Reader) ([]pkg.Package, error) {
 	}
 
 	return packages, nil
+}
+
+func (p PackageJSON) hasMinimumRequiredValues() bool {
+	return p.Name != "" && p.Version != ""
 }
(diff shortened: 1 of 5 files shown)


syft/cataloger/javascript/parse_package_json.go
```diff
@@ -172,6 +173,12 @@ func parsePackageJSON(_ string, reader io.Reader) ([]pkg.Package, error) {
 			return nil, fmt.Errorf("failed to parse package.json file: %w", err)
 		}
 
+		if !p.hasMinimumRequiredValues() {
+			log.Debug("encountered package.json file without the minimum number of field values required for" +
```
alfredodeza, line 177: which fields? (I know, but a user will probably not)

syft/cataloger/javascript/parse_package_json_test.go
```diff
@@ -142,3 +142,20 @@ func TestParsePackageJSON(t *testing.T) {
...
+		t.Fatalf("failed to open fixture: %+v", err)
+	}
+
+	actual, err := parsePackageJSON("", fixture)
+	if err != nil {
+		t.Fatalf("failed to parse package-lock.json: %+v", err)
+	}
+
+	if len(actual) != 0 {
+		t.Errorf("no packages should've been returned")
```
alfredodeza, line 159:
```suggestion
		t.Errorf("no packages should've been returned, but got %d",  len(actual))
```

test/integration/test-fixtures/image-large-apk-data/Dockerfile
```diff
@@ -1,2 +1,5 @@
-FROM alpine:latest
-RUN apk add tzdata vim alpine-sdk
+FROM alpine@sha256:d9a7354e3845ea8466bb00b22224d9116b183e594527fb5b6c3d30bc01a20378
```
alfredodeza, line 1: yassss

alfredodeza (changes requested):

alfredodeza (approved):

PR #317, opened 2021-02-01
Implement CPE matching using dictionary.

glebiller:
### Context
The current implementation provide a "naive" list of CPEs by generating all possible pairs based on detected vendors & products. This method leads to a lot of false positive detection specifically when running on java dependencies.

### New proposed CPE identification method
The complete list of up-to-date [CPE Dictionary](https://nvd.nist.gov/products/cpe) is available on the NVD website. This new implementation is checking for updates from the CPE data feed and create a local [Bleve index](http://blevesearch.com/) containing all possible (vendor|product) pair.

On package the search index is queried using all candidates vendors & products detected on the package metadata.
Additionally some specific vendors can be specified in the config file.

### Specific vendors & product 📦 
Some CPE are not completely matching the metadata, one big example is the Spring Framework.

For instance the CPE for Spring Boot is `cpe:2.3:a:pivotal_software:spring_boot:2.4.2:*:*:*:*:java:*:*`, however the "pivotal_software" company is never mentioned in the metadata of the JAR archive.

To solve this issue, I added a set of specific vendors & products that can be customised through configuration file:
```
  specific-vendors:
    - match: ^spring-.*$
      term: pivotal_software
      boost: 4
```

### Added commands 📜 

- syft cpe check: checking for updates
- syft cpe delete: delete local dictionary
- syft cpe import: import CPE archive (useful for offline environment)
- syft cpe status: information of the local dictionary
- syft cpe update: update the local dictionary with latest web archive

### Improvement required 🚧 

More testing is required for OS packages to detect correct vendors & products to improve matching.

### Next goal

Adding the correct CPE matching to syft will allow grype to have a simpler algorithm for vulnerability detection that should lead to fewer false positive.

It would also be possible to restrict the number of CPE to a single entry as a product should have one and only one matching CPE.


luhring:
Hi @Kissy — first off, this is very cool! A _HUGE_ "thank you" for creating this!

I've been reviewing this over the past week, and we've chatted a bit about this internally. Here are some high-level thoughts so far...

1. It's clear that a lot of investment went into aligning this code with the patterns that already exist in our tools. This makes it much easier to integrate. We really appreciate this. 💯 

2. This PR actually ties into a project we've been talking about internally — moving Syft's CPE generation to its own project. At a minimum, this would be a library that Syft leverages as it identifies packages, so that much of today's data flow would remain the same. As you've seen, there would likely be multiple strategies for generating CPEs, and some may be more involved than others (for example, downloading external data like the CPE dictionary). This project could potentially expose a CLI, where you could feed Syft output (or even another SBOM format) into the tool, and the tool would return CPEs for each package. This idea might seem like a curveball, but we'd love to collaborate with you on this if you're interested.

3. Another item on our wishlist has been a statistical evaluation harness for assessing changes like this. If the goal is to reduce false positives, for example, we should be able to prove that. And we should be able to verify that other problems haven't been introduced at the same time.

4. Along those lines, I'm concerned about a potentially large increase in false negatives due to the CPE dictionary missing CPEs for many packages being identified. Since both the CPE dictionary and the actual NVD data is sourced from various (and usually, non-overlapping) places, there's no inherent guarantee that the CPE dictionary entries help us with downstream vulnerability matching (as far as I'm aware). I'm interested in your take on this.

5. The user experience for Syft and Grype is very important to us. We're very interested in the tools staying performant. On my machine, the CPE dictionary indexing step usually takes _over 10 seconds_, which felt very different than what I'm used to with Syft. We're wondering about finding a way for end users not to be required to run the indexing themselves. With Grype, we've created a nightly process of building a database that Grype users can fetch and then use immediately. Similarly, it could help keep runtime down if the CPE generation consumes data that's immediately ready to use. Anchore could create and host such an artifact, but we could additionally allow a path for users that doesn't depend on Anchore or its infrastructure.

Those are our big picture thoughts on this PR. We'd love to get your thoughts, and we can figure out next steps after that.

Thanks again for this submission! It's very exciting.

Kissy:
Hey @luhring, thank you for your detailed feedback. Here are more though on the various points :)

1. I really enjoyed the architecture of each projects, it was a pleasure to follow everything :)

2. I totally agree with it. Matching CPE is quite a complex task and seems to be totally different based on the package type (Java vs RPM/Deb for instance seems to have a totally different nomenclature for vendor & project naming in CPE. Moving this task into a separate project could make sense. 
I think this point will also be tied to the 5, matching a CPE should be a simple job enough if the CPE Dictionary is pre-processed correctly.

3. That was one of my main point, this will requires a large amount of tests to be sure that everything is matched correctly: both false positive and missed identification.
I started this PR because we were very interested to implement grype (and the admission controller for K8s). Unfortunately few tests on our projects (Java mostly) turned out that we had dozens of false positive / missed CPE compared to our actual implementation of OWASP dependency-check. Even a plain Oracle Linux image with Tomcat installed already reported false positive. This was a no-go for the company to migrate to grype and we will stick to OWASP Dependency checks for Java.
The real benefit of grype is to detect vulnerabilities not only in Java but on the whole container image! 

4. Looking at the Anchore vulnerability feed (https://ancho.re/v1/service/feeds/), only nvd & nvd2 are relying on CPEs because only the NVD Vulnerability Feed is using CPE to match packages.
The _NVD CPE Dictionary_ is the source of truth. All vulnerabilities in the _NVD Vulnerability Feed_ are matching a CPE that is listed in the _NVD CPE Dictionary_. If a package does not have known vulnerabilities according to NVD, then it is safer to omit the CPE to avoid false positive as the _NVD Vulnerability Feed_ cannot contain an entry for that package. However, this is true only with an up-to-date CPE Dictionary and can be inaccurate later in time, therefor saving the result of the CPE to test it at a later time could lead to a missed vulnerability. 
Having an empty CPE does not mean that we will be have false negative if the CPE only if the matching is reliable.

5. This is a great idea. I also don't like how long it takes to parse the 250Mb NVD CPE Dictionary XML file. I continue playing with it and it turns out that to improve reliability a extensive parsing of the file is required. It would totally make sense to move this part into a separate process and provide only a post-processed dictionary that syft client could be downloading directly and optionally a path to the official NVD XML file for offline installation would improve the UX. One thing to note though, is that the NVD CPE Dictionary XML file is updated really really often. This does not mean that the processed CPE Dictionary should be updated as most of the time this is to add a new version to an existing software, which should not impact the CPE Matching process as it's not relying on the (version/edition).

Few things that I noticed and I am working on:
* CPE Dictionary contains Title & References URL. It seems that the matching is much more effective with those two attributes rather than relying only on Vendor & Product. I started a PR for go-rpmdb to include URL from RPM Metadata in order to be able to easily do a match.
* Having some integration tests done with standard images (Ubuntu, RHEL, Debian & some really popular (maybe the project with the highest number of version in the CPE Dictionary?) Java / NodeJS / Python projects to ensure the CPE matching is working

Of course, this is will requires lot of more discussion / work / PoC & testing!

luhring:
Hi @Kissy,

This is all great to hear. It sounds like we're well-aligned!

>I started this PR because we were very interested to implement grype (and the admission controller for K8s). Unfortunately few tests on our projects (Java mostly) turned out that we had dozens of false positive / missed CPE compared to our actual implementation of OWASP dependency-check. Even a plain Oracle Linux image with Tomcat installed already reported false positive. This was a no-go for the company to migrate to grype and we will stick to OWASP Dependency checks for Java.

This is a really helpful usage insight. And it's yet another motivation to get Grype's matching better! Thanks for sharing.

Regarding the **completeness of the official CPE dictionary** — Thanks for the explanation. I think I'm still a bit confused. I wrote a script to go through a given NVD vulnerabilities JSON file from our feeds service, and for each listed vulnerable CPE, check if it's found in the CPE dictionary. The results show that there are several vulnerable CPEs not found in the CPE dictionary.

Sometimes, it's close but doesn't match. Here's one vulnerable CPE:

`cpe:2.3:a:oracle:communications_unified_inventory_management:7.2.4.2:*:*:*:*:*:*:*`

The CPE dictionary has this software listed but for other versions, e.g.:

`cpe:2.3:a:oracle:communications_unified_inventory_management:7.3:*:*:*:*:*:*:*`
`cpe:2.3:a:oracle:communications_unified_inventory_management:7.3.0:*:*:*:*:*:*:*`
`cpe:2.3:a:oracle:communications_unified_inventory_management:7.3.2:*:*:*:*:*:*:*`
`cpe:2.3:a:oracle:communications_unified_inventory_management:7.3.4:*:*:*:*:*:*:*`
`cpe:2.3:a:oracle:communications_unified_inventory_management:7.3.5:*:*:*:*:*:*:*`

Other times, there's no entry in the dictionary even close to the vulnerable CPE. Here's one such vulnerable CPE:

`cpe:2.3:a:sap:s4score:1.01:*:*:*:*:*:*:*`

I'm not finding the string `s4score` anywhere in the CPE dictionary.

Can you tell if there's something about how I'm analyzing this that's misguided? I'd love to be confident in the CPE dictionary as a source of truth for vulnerability data.

Kissy:
Hello, 

I was running the same kind of test as yours, as I also noticed some missing entries. 
NVD does not explain exactly how they are building the Dictionary, however after few analysis it seems that they are taking only the last 1 to 2 year to build the dictionary hence the missing entries for old CVE or missing old versions.

For the version I don't think it's an issue because I think it's best to not use it to match it (to allow new version to still match).
Maybe it will be best to build a CPE dictionary based on the actual CVE list instead of using the one provided by NVD.

westonsteimel:
So I have also just started taking a look at CPE matching stuff and noticed that the [CPE Match Feed](https://nvd.nist.gov/vuln/data-feeds#cpeMatch) might have all CPEs that appear in a CVE, but it will have an empty list for the `cpe_name` attribute it there isn't an entry for it in the Official CPE Dictionary.

From that I'm seeing there are currently 297,434 CPE's in the match feed, and of those 136,973 are missing an official CPE dictionary entry.  And then looking at the official CPE dictionary, there are 613,816 entries, but 469,124 of them don't yet have any CVEs that they match.

westonsteimel:
I've started some work myself on mapping from packages to the corresponding CPE components in the NVD at https://github.com/westonsteimel/package-metadata.  I've only made it through reviewing an initial set of python packages at the moment, but do plan on continuing through other package ecosystems in the future.  I basically took information from the GitHub Security Advisory feed and the pyupio db to collect the CVE's they identified for packages and then pulled all the CPE mappings associated with those CVE's and manually reviewed them.  You can match a package to multiple CPE configurations; however, where I have found multiple CPE configs referencing the same packages I have been issuing requests to the [CPE dictionary team](cpe_dictionary@nist.gov) to try and collapse to a single CPE and deprecate the others.  They are generally fairly responsive.  Anyways, just thought I'd share if it's helpful to any of the work any of you are doing.

luhring:
@westonsteimel That's really cool! And that seems like a ton of work. So if I'm understanding correctly, the flow is:

1. Use a source like GHSA to map packages to CVEs
2. Find the listed CPEs for each CVE
3. Review manually

And thus this becomes a quasi-authoritative lookup mechanism for mapping packages to CPEs that are useful for CVE matching?

luhring:
Also, just an update from the core team on this — we plan on making a significant investment in this effort, and more generally addressing false positives in Grype (the primary downstream consumer of these CPEs), starting in the next few months. It's super important to us.

westonsteimel:
Thanks @luhring!  It is definitely quite a bit of work, but I'm just chipping away at it as I have time, so it hasn't been too overwhelming thus far.  I just hope it ends up being useful to someone at some point.  Yes, your understanding of my current workflow is correct.  I have a few quickly hacked-together scripts at https://github.com/westonsteimel/nvd-data-analysis/ that I used to generate the initial set of stuff to review, and I also generate a [json file](https://raw.githubusercontent.com/westonsteimel/nvd-data-analysis/main/data/cpe/unique_prefixes.json) of all of the unique`vendor:product` combos that exist.  I usually search through that to try and uncover multiple CPE configs for the same package that weren't present in the CVE's known from GHSA, and then try to (eventually) follow up with the CPE Dictionary team to get them down to one official entry.  At this point I have a fairly long list of them that I haven't sent over yet, but I'll get to it eventually. I'm also trying to capture things like the source url as that can potentially be used to find other matches (since CVE's often link to GitHub issues, PR's etc related to the affected packages).  There is also a list that Debian started at https://salsa.debian.org/security-tracker-team/security-tracker/-/blob/master/data/CPE/list to use to try and reconcile their security tracker with the NVD to find anything that may have been missed.  Not sure if they're actually using it, but I figured I'd have a look through it after I finish up with my initial set.

I started it hoping that it might potentially help accuracy of vulnerability detection across the spectrum, and it has certainly helped address some issues in the NVD itself as well as uncovering some specific instances where incorrect packages were being associated with CVE's either in the NVD or by consumers of the NVD stream (an example was with https://snyk.io/vuln/SNYK-PYTHON-QLIB-1054635 which references the `qlib` package on pypi, but in looking at the CVE and associated github activity, I realised the actual affected package is called `pyqlib` and reached out to them, and they corrected it in https://snyk.io/vuln/SNYK-PYTHON-PYQLIB-1085990)

Anyways, looking forward to hearing more about the team's plan on reducing false positives etc in grype.  I'm pretty passionate about that at the moment and am very glad it's being looked into.

luhring:
@westonsteimel 

> I just hope it ends up being useful to someone at some point.

It seems likely that it will be... :)

> There is also a list that Debian started [...] I figured I'd have a look through it after I finish up with my initial set.

That's interesting — I'm curious how many of these exist in the wild. Let me know what you find if you get to this!

> Anyways, looking forward to hearing more about the team's plan on reducing false positives etc in grype. I'm pretty passionate about that at the moment and am very glad it's being looked into.

Absolutely, will do. It's been great seeing our interests aligning around this, and I'm looking forward to making headway on this!

luhring:
Doing some cleanup, so closing this for now. But this has been a great exploration, and we'll continue to reference these notes as we iterate on Grype's matching strategy. Also, we can always re-open this if anyone wants to continue down this path!

PR #318, opened 2021-02-01
Allow path prefixes

### LINKED ISSUE #315: Missing support for path prefixes during import to Anchore API
This means that Syft's import feature **does not work** with the standard way we deploy Enterprise via Helm, where the API paths should be prefixed with `/v1/`. For details on the chart config, see https://github.com/anchore/anchore-charts/blob/8ff7bc108290ce07a6f2d92db5aa1be11985dc23/stable/anchore-engine/templates/enterprise_ui_configmap.yaml#L20

We're finding that users are trying to specify path prefixes with the host parameter (e.g. `-H` on the CLI), such as `https://mydeployment.example.com/v1`.

At a minimum:

1. We need a way to allow the user to specify path prefixes as they tell Syft where their deployment is located (we should **not** assume that the user's API is deployed at a root path for a hostname).
1. We need to ensure users have a smooth interaction with Anchore Enterprise by default.

luhring:
Closes #315 

This PR:

- broadens what users can specify using the `--host` / `-H` configuration value (adds support for path prefixes, e.g. `-H localhost:8000/v1`)
- ensures that, for now, our base paths for API requests end with `/v1`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -159,12 +159,12 @@ func doImport(src source.Source, s source.Metadata, catalog *pkg.Catalog, d *dis
 	}
 
 	c, err := anchore.NewClient(anchore.Configuration{
-		BasePath: appConfig.Anchore.Host,
+		BaseURL:  appConfig.Anchore.Host,
 		Username: appConfig.Anchore.Username,
 		Password: appConfig.Anchore.Password,
 	})
 	if err != nil {
-		return fmt.Errorf("failed to create anchore client: %+v", err)
+		return fmt.Errorf("unable to upload results: %w", err)
 	}
 
 	importCfg := anchore.ImportConfig{
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -158,19 +158,10 @@ func doImport(src source.Source, s source.Metadata, catalog *pkg.Catalog, d *dis
 		}
 	}
 
-	var scheme string
-	var hostname = appConfig.Anchore.Host
-	urlFields := strings.Split(hostname, "://")
-	if len(urlFields) > 1 {
-		scheme = urlFields[0]
-		hostname = urlFields[1]
-	}
-
 	c, err := anchore.NewClient(anchore.Configuration{
-		Hostname: hostname,
+		BasePath: appConfig.Anchore.Host,
 		Username: appConfig.Anchore.Username,
 		Password: appConfig.Anchore.Password,
-		Scheme:   scheme,
 	})
 	if err != nil {
 		return fmt.Errorf("failed to create anchore client: %+v", err)
--- a/internal/anchore/client.go
+++ b/internal/anchore/client.go
@@ -3,6 +3,7 @@ package anchore
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/anchore/client-go/pkg/external"
 	"github.com/anchore/syft/internal"
@@ -10,11 +11,10 @@ import (
 )
 
 type Configuration struct {
-	Hostname  string
+	BasePath  string
 	Username  string
 	Password  string
 	UserAgent string
-	Scheme    string
 }
 
 type Client struct {
@@ -29,16 +29,14 @@ func NewClient(cfg Configuration) (*Client, error) {
 		cfg.UserAgent = fmt.Sprintf("%s / %s %s", internal.ApplicationName, versionInfo.Version, versionInfo.Platform)
 	}
 
-	if cfg.Scheme == "" {
-		cfg.Scheme = "https"
-	}
+	basePath := ensureURLHasScheme(cfg.BasePath) // we can rely on the built-in URL parsing for the scheme, host,
+	// port, and path prefix, as long as a scheme is present
 
 	return &Client{
 		config: cfg,
 		client: external.NewAPIClient(&external.Configuration{
-			Host:      cfg.Hostname,
+			BasePath:  basePath,
 			UserAgent: cfg.UserAgent,
-			Scheme:    cfg.Scheme,
 		}),
 	}, nil
 }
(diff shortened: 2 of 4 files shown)


internal/anchore/client_test.go
```diff
@@ -0,0 +1,109 @@
...
+		},
+		{
+			url:      "http://localhost/v1",
+			suffix:   "/v1",
+			expected: "http://localhost/v1",
+		},
+		{
+			url:      "http://localhost/v1/",
+			suffix:   "/v1",
+			expected: "http://localhost/v1//v1",
```
alfredodeza, line 91: we aren't sanitizing this? this looks... unexpected

internal/anchore/client.go
```diff
@@ -56,3 +57,27 @@ func (c *Client) newRequestContext(parentContext context.Context) context.Contex
...
+	if !hasScheme(url) {
+		return fmt.Sprintf("%s://%s", defaultScheme, url)
+	}
+
+	return url
+}
+
+func ensureURLHasSuffix(url, suffix string) string {
+	if !strings.HasSuffix(url, suffix) {
+		return url + suffix
```
alfredodeza, line 79: this doesn't look robust to support trailing slashes with leading slashes. Seems like Go can handle this better with `url.Parse` and `path.Join`

```go
package main

import (
	"fmt"
	"log"
	"net/url"
	"path"
)

func main() {
	u, err := url.Parse("http://localhost/")
	if err != nil {
		log.Fatal(err)
	}
	u.Path = path.Join(u.Path, "/v1/")
	s := u.String()
	fmt.Println(s)
}
```

Prints:

```
http://localhost/v1
```

internal/anchore/client_test.go
luhring: 1. I see what you did there.
2. We're sanitizing the slash when pipelining the actual `basePath` value. This is meant to be a test of the single "suffixing" operation itself. I'm open to adjusting the layout of functionality, though.

internal/anchore/client.go
luhring: See [other comment](https://github.com/anchore/syft/pull/318#discussion_r568189105). We _are_ sanitizing slashes, just not in this discrete operation.

alfredodeza: I think that preventing these slashes from creating a problem is something that we should try - even though this is not the exact problem this PR is addressing. Path normalization when adding parts is common enough (Python has the same facilities), and I think we can take advantage of it in this case.

luhring: >I think that preventing these slashes from creating a problem is something that we should try

I completely agree. I think what I'm trying to communicate is that it seems like we're doing this more than you're realizing. It seems like you're focusing specifically in the function `ensureURLHasSuffix`, but there's more to the URL parsing than this function. If you look at how this function is used in context, I think you'll see a bigger picture.

Another way to frame this is to ask the question: "What's some input that produces a bad or unexpected result?"

If we can't think of any cases, then we'd have to conclude that the logic actually is robust.

Having said all that, I think we could rearrange the functionality. Let's pair on this tomorrow!

cmd/root.go
```diff
@@ -159,12 +159,12 @@ func doImport(src source.Source, s source.Metadata, catalog *pkg.Catalog, d *dis
...
 
 	c, err := anchore.NewClient(anchore.Configuration{
-		BasePath: appConfig.Anchore.Host,
+		BaseURL:  appConfig.Anchore.Host,
 		Username: appConfig.Anchore.Username,
 		Password: appConfig.Anchore.Password,
 	})
 	if err != nil {
-		return fmt.Errorf("failed to create anchore client: %+v", err)
+		return fmt.Errorf("unable to upload results: %w", err)
```
alfredodeza, line 167: yesssssssss

internal/anchore/client.go
```diff
@@ -58,26 +60,56 @@ func (c *Client) newRequestContext(parentContext context.Context) context.Contex
...
-	parts := strings.Split(url, "://")
+var ErrInvalidBaseURLInput = errors.New("invalid base URL input")
 
-	return len(parts) > 1
+func prepareBaseURLForClient(baseURL string) (string, error) {
+	if err := checkBaseURLInput(baseURL); err != nil {
+		return "", err
+	}
+
+	scheme, urlWithoutScheme := splitSchemeFromURL(baseURL)
```
alfredodeza, line 70: love it

internal/anchore/client.go
```diff
@@ -58,26 +60,56 @@ func (c *Client) newRequestContext(parentContext context.Context) context.Contex
...
+func splitSchemeFromURL(url string) (scheme, urlWithoutScheme string) {
+	if hasScheme(url) {
+		urlParts := strings.SplitN(url, "://", 2)
+		scheme = urlParts[0]
+		urlWithoutScheme = urlParts[1]
+		return
 	}
 
-	return url
+	return "", url
```
alfredodeza, line 108: since the consumer of this function will always want a scheme and a url, shouldn't the fallback to `"http"` be used here instead of being done in the caller? (not a hard comment, just curious about the alternative)
```suggestion
	return "http", url
```

internal/anchore/client_test.go
```diff
@@ -36,73 +36,174 @@ func TestHasScheme(t *testing.T) {
...
+			expectedScheme:           "",
+			expectedURLWithoutScheme: "localhost",
+		},
+		{
+			input:                    "https://anchore.com/path",
+			expectedScheme:           "https",
+			expectedURLWithoutScheme: "anchore.com/path",
+		},
+		{
+			input:                    "tcp://host:1234",
```
alfredodeza, line 176: excellent, I like having other stuff aside from http cases

internal/anchore/client.go
luhring: @alfredodeza It's funny you say that, I was thinking about doing that initially. Then I reflected and thought the process was simpler when each step was broken out — like "Okay, what's the scheme? Okay, the scheme is nothing, so here's what we'll use". This way, the function that splits the scheme off is transparent, and the consumer can decide what the default is. It's important to have a default, for sure, but this approach segregates the responsibilities of using a default (like HTTP) vs. breaking the scheme off from the rest of the URL (which is its own important operation)

alfredodeza (approved):

PR #324, opened 2021-02-12
Add handling of interrupting signals to the UI

luhring:
This PR fixes the problem where "Ctrl+C-ing" out of Syft would leave your terminal in a state where the cursor was hidden.

It adds a mechanism to the UI's `for`/`select` loop to listen for particular signals of interest and respond by breaking the loop. This allows the execution of the deferred statement that re-enables the user's cursor.

Currently, there are two signals being subscribed to, but this list is easily amended if anyone has suggestions.

- SIGINT (e.g. hitting Ctrl+C after running a syft command)
- SIGTERM (e.g. terminating a running syft process via `kill -TERM <syft-command>`)

### DIFF
--- a/internal/ui/etui/ephemeral_tui.go
+++ b/internal/ui/etui/ephemeral_tui.go
@@ -23,7 +23,9 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/signal"
 	"sync"
+	"syscall"
 
 	"github.com/anchore/syft/internal/logger"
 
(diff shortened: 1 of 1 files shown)


alfredodeza (approved):

wagoodman (approved):
👍  nit addition: also adding to the logger UI loop https://github.com/anchore/syft/blob/5370daf02754b1b52e9b1c2dc5b90f67ae3b6f6d/internal/ui/logger_output.go#L15

luhring:
@wagoodman Good thinking — I'll add that

PR #329, opened 2021-02-17
Mark package relations by file ownership

### LINKED ISSUE #166: global incremental ID creates issues for consumers
**What happened**: The `syft/pkg/catalog.go` file has a global (unexported) variable that increments whenever a new package is added to the catalog. This creates a hard constraint of not being able to control the IDs and forces consumers to use the catalog one at the time.

This particular problem was evident when 2 tests where trying to create a pkg catalog from syft. When the two tests ran, one would fail. These tests were adding 3 packages each, and they both were relying on the fact that the IDs would be 0, 1, and 2. When both added the tests, the IDs for the second test would be 3, 4 and 5. 

**What you expected to happen**: Not have a global auto-incrementing variable that is unexported, and possibly allow consumers to control it.

**How to reproduce it (as minimally and precisely as possible)**: Create two functions that add to a pkg.catalog

**Environment**:
- syft version (use `syft version -v`):

### LINKED ISSUE #198: Add a `Files` field in the `Package` struct
**What would you like to be added**: Many package specifications like Ruby Gems [provide information regarding the files](https://guides.rubygems.org/specification-reference/#files) that belong to a package. This information is currently not present in the output from syft. 

It seems that this information is very common among other package formats so that this could potentially mean a new field in the Package struct vs. just adding it to the metadata.

Either case, this is not present at all and is needed for the different analyzers that provide that information in Anchore Engine. For package types like Python, this might present a problem, where the egg parser looks at `PKG-INFO` which doesn't have access to files, although `SOURCES.txt` (within the same directory) has all the files for that package.

**Why is this needed**: Parity with Anchore Engine Analyzers, specifically related to this issue https://github.com/anchore/anchore-engine/issues/629

**Additional context**:

### LINKED ISSUE #326: Label relationships between packages
An indication on each package if it is owned by another package.

Syft raises a lot of metadata about the packages it discovers, including file ownership (on `Package.Metadata.Files[]` for some package metadata types, not all). Additionally Syft shows for each package a set of locations that led to the discovery of this package (on `Package.Location[]`). It would be great if we could pair these two bits of information to convey relationships between packages (something like "package A is owned by package B"). 

**Why is this needed**:
This is a pretty common case say when an RPM package installs a set of python packages; the RPM cataloger picks up on one and the Python cataloger picks up on the other. Both are valid packages in their own right so it doesn't make sense to de-duplicate one package over another, however, downstream consumers can make use of this extra relationship information to make better decisions about (e.g. grype shouldn't do vulnerability matching against both the RPM and the Python package, but instead only the authoritative RPM package).

**Additional context**:
This is pretty similar to #32 but is subtly different in the sense that this issue is NOT intending to filter out packages based on some criteria, but instead annotate a relationship between packages for downstream consumers.

wagoodman:
This PR adds a new `Relations` attribute onto a package to track one or more `ParentsByFileOwnership`, or packages who's `.Location[].RealPath` or `.Location.[].VirtualPath` overlaps with another packages `.Metadata.Files[].Path` (or similar field). All packages now explicitly list parents, in this way consumers can easily identify the "root" packages and filter out child packages.

TODO:
- [ ] update the imports package type to include the updated `Relations` field
- [ ] test explicitly with engine (manual task)

Relates to: https://github.com/anchore/anchore-engine/issues/315 , https://github.com/anchore/anchore-engine/issues/445 , and https://github.com/anchore/anchore-engine/issues/460
Closes #326 
Closes #166
Closes #198


syft/pkg/relations.go
```diff
@@ -0,0 +1,6 @@
+package pkg
+
+type Relations struct {
+	// ParentsByFileOwnership lists all parent packages that claim ownership of this package
```
alfredodeza, line 4: Reading this it seems that _Parents_ is needed because there can be more than one parent for a package, but that doesn't seem right for the type of relationship this is trying to establish (many-to-one). This is arguable, but I tend to think about a _Parent_ as a one way relationship to a _child_ or _children_ (package or packages in this case). 

Is it possible to rethink these variables to map the same concept?

syft/pkg/dpkg_metadata.go
```diff
@@ -44,3 +46,12 @@ func (m DpkgMetadata) PackageURL(d *distro.Distro) string {
 		"")
 	return pURL.ToString()
 }
+
+func (m DpkgMetadata) ownedFiles() (result []string) {
+	for _, f := range m.Files {
+		if f.Path != "" {
```
alfredodeza, line 52: is it not desirable to prevent duplicates here? Or does it want to intentionally capture everything?

syft/pkg/npm_metadata.go
```diff
@@ -1,5 +1,7 @@
 package pkg
 
+var _ fileOwner = (*NpmPackageJSONMetadata)(nil)
```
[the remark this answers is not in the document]
wagoodman: indeed; this is a quick assertion that the type in question (`NpmPackageJSONMetadata`) positively implements the interface (`fileOwner`). Just a note, having this line here does not aid in functionality at all, but is a nice to have in case you make a change to the code and accidentally break the interface implementation (in this case `.ownedFiles()`)

The specifics of the line are:
- `var _ fileOwner = ...`: make a variable with no name (`_`, a `BlankIdentifier`) of type `fileOwner` (our interface)
- ` = (*t)(nil) ...`: the value is the result of a type conversation. Specifically interpreting a `nil` value as a pointer of our type (`*NpmPackageJSONMetadata`).

The upsides of this approach is that you get a compiler-time check that your type truly implements the interface.

Here's an article on type assertion vs type conversion: https://www.sohamkamani.com/golang/type-assertions-vs-type-conversions/

syft/pkg/relations.go
wagoodman: In a pure sense, what you are saying is true --there should really be a single parent and a possibility of multiple children. However, it is allowable for multiple packages to claim ownership of the same files, which would register here 
 as multiple parents; whether this is valid for any particular ecosystem to do doesn't necessarily matter.

I thought about making this something like:
```
type Relations struct {
   ChildrenByFileOwnership []ID
}
```
But this has the downside that consumers have to do a bit of work to determine which packages are "root" packages (not owned by another package). By tracking parent relationships a consumer only needs to check that the `ParentsByFileOwnership` is empty to ensure it is a "root" package.

syft/pkg/dpkg_metadata.go
wagoodman: I could make this a set implementation and return the resulting slice, think it's worth it?

alfredodeza: Good question - I'm not sure! But was wondering if it is a problem to have several duplicates here. It does seem that the intent is to report all owned files, not every single file reported from the package owner(s) - in which case it would make sense to prevent duplicates. 

Either a set implementation returning a slice or preventing adding a new item if it already exists in the current slice sounds OK to me

syft/cataloger/catalog.go
```diff
@@ -39,6 +39,8 @@ func newMonitor() (*progress.Manual, *progress.Manual) {
 // request.
 func Catalog(resolver source.Resolver, theDistro *distro.Distro, catalogers ...Cataloger) (*pkg.Catalog, error) {
 	catalog := pkg.NewCatalog()
+	defer catalog.Finalize()
```
luhring, line 42: I'd suggest changing this from using `defer` to just calling the method explicitly at the right time later on in the implementation. If I'm understanding correctly, it's not universally true that we want this to run before the function returns — specifically if we're returning an error instead of the catalog

syft/pkg/catalog.go
```diff
@@ -41,43 +39,133 @@ func (c *Catalog) PackageCount() int {
...
+}
+
+func (c *Catalog) Finalize() {
+	c.markPackageOwnership()
+}
+
+// markPackageOwnership find overlaps in file ownership with a file that defines another package. Specifically, a .Location.Path of
+// a package is found to be owned by another (from the owner's .Metadata.Files[]). This relationship is captured on the
+// child package.
+func (c *Catalog) markPackageOwnership() {
```
luhring, line 133: This method is a bit on the long and nested side, it'd be nice to find ways to decompose this into more focused functions

syft/pkg/catalog.go
```diff
@@ -41,43 +39,133 @@ func (c *Catalog) PackageCount() int {
...
+	var pkgParents = make(map[ID]*strset.Set)
```
luhring, line 134: I think this function would get a bit simpler and safer if the map's value type wasn't a pointer. This would avoid the need for `nil` checks further on

syft/pkg/dpkg_metadata.go
luhring: Re set implementation — I'm not sure it's worth it.

I do have a question — are we allowing `f.Path` to be an empty string upstream? It'd be nice to be able to rely on a usable value here. I could be misunderstanding how this value gets used

syft/pkg/catalog.go
```diff
@@ -41,43 +39,133 @@ func (c *Catalog) PackageCount() int {
...
-	p.id = ID(nextPackageID)
-	nextPackageID++
+	_, exists := c.byID[p.ID]
+	if exists {
+		log.Errorf("package ID already exists in the catalog : id=%+v %+v", p.ID, p)
+		return
+	}
+
+	if p.ID == "" {
+		p.ID = newID()
```
luhring, line 77: This is a bit concerning — the nondeterminism — especially since this gets exposed outside of the implementation. But the same could be said of any persistence/store-oriented object like a database table. No change needed right now — just noting a non-ideal trait. Open to further discussion

syft/pkg/apk_metadata_test.go
```diff
@@ -31,3 +34,45 @@ func TestApkMetadata_pURL(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(strings.Join(test.expected, ","), func(t *testing.T) {
+			var i interface{}
+			i = test.metadata
+			actual := i.(fileOwner).ownedFiles()
+			for _, d := range deep.Equal(test.expected, actual) {
```
luhring, line 73: Not required — but it could be nice to start trying out Testify's assert package in earnest, in hopes of finding a better way to approach assertions in our tests

test/integration/test-fixtures/image-owning-package/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM ubuntu:20.04
+# this covers rpm-python
+RUN apt update && apt install -y python-pil
```
luhring, line 3: It could be good to pin installed dependencies. We've run into a few issues where lack of pinning meant our test outcomes changed unexpectedly down the road after a completely unrelated change

syft/pkg/apk_metadata_test.go
alfredodeza: +1 on normalizing moving forward with testify

syft/pkg/dpkg_metadata.go
alfredodeza: Looking at the consumer of this, the idea is to use the owned files to get packages by path. Not de-duplicating here means extra (unnecessary) work for the consumer.

wagoodman: re: sets / duplicates: It's not clear if deduplicating signals anything different to the consumer (the given slice, deduplicated or not still indicates ownership), but it does clarify the intent of the interface ("provide a __set__ of file paths that are owned"). I think I'll update to use a set in each, but still return a slice as the return type.

re: empty path: I felt that it was important to ensure that any implementation could not signal "ownership" of "nothing". I didn't check the underlying implementations to see if it was possible, but instead added the defense here. The coincidental implementation of underlying cataloger parsers shouldn't really drive the behavior of "fileOwner" IMHO.

syft/pkg/catalog.go
wagoodman: IDs by nature should be nondeterministic, we shouldn't be depending on a particular value and claim this is a ID --so the change from an incrementing number (which is still not deterministic since package order may vary on addition depending on the cataloger implementation... if we add concurrency for instance).

As you noted, this is a impactful change though since this is being exposed (where previously it was not). You couldn't reference theses IDs across syft runs, since the IDs are random, however, we could in the future provide a stable/deterministic package "fingerprint" field derived from the  package property values.

syft/pkg/apk_metadata_test.go
wagoodman: hadn't realized we reached consensus on using testify, I'll start using it after this PR :)

syft/pkg/catalog.go
wagoodman: the "splits" I considered seemed contrived, so I instead added comments to help.

syft/pkg/catalog.go
alfredodeza: This issue I think is relevant wrt to IDs https://github.com/anchore/syft/issues/166 - definitely something that has come up before

luhring: Ah yeah, I remember that one. I don't know about "should be nondeterministic" — I think the important attribute of IDs is uniqueness. Something consumers can use to positively find a single item.

I'm okay with this code going in as is, I was more thinking of what @alfredodeza mentioned — this has come up before, and it warrants caution.

syft/pkg/apk_metadata_test.go
luhring: 🎉 nice!

syft/pkg/catalog.go
luhring: The comments are a good start, because you've taken time to think through what each section of the method is doing. It's clear that different sections of this method are doing specific things. These comments can now provide a roadmap toward what makes sense to factor out. This will help others wrap their head around what's going on in this function more readily

syft/pkg/catalog.go
wagoodman: As written in the context of this function the use of a pointer value in a map is safe. (the use of a pointer in of itself is not "unsafe").

Also, the constructor for `strset` provides a pointer to a set, also which nil check are you referring to? The closest I found was:
```
				if _, exists := pkgParents[subPackage.ID]; !exists {
					pkgParents[subPackage.ID] = strset.New()
				}
```
which would be required either way.

syft/pkg/catalog.go
wagoodman: I guess what I'm saying is that I don't think this function is that long (35-ish lines including whitespace & comments).

I see where it __could__ get broken up into other functions, however in all cases you will be passing what could stay as local state around to other functions (`pkgParents`), which means a reader needs to understand what the separate function is doing relative to the already-prepared state from another function. In these cases I tend to prefer a slightly longer function (which is still within the linter rules) with healthy comments.

If there were multiple rules for what determines "ownership" I would probably change my mind.

syft/pkg/catalog.go
luhring: >the use of a pointer in of itself is not "unsafe"

Agreed.

>which nil check are you referring to?

An imaginary one that's not in the code yet.

As I understand this code, there's no point where a `nil` value can be assigned as a value for a map key in `pkgParents`, so in that sense we're good! The danger is in maintainability and _future_ problems.

Making this a pointer means the system is more complex because there are more possible states in the system, which means it's now less difficult to introduce problems down the road. In this code, there are two accesses of a "potentially" `nil` value (to your point, not today, but there's no guarantee of that remaining true as this code is maintained):

- The call to `.Add` on line 154
- The call `.List` on line 164

If I were a developer in the future coming to work on this function, I would try to look at this code and quickly get a sense for what's going on. I would see that the map's value type was a **pointer** to a `strset.Set`. I'd assume that this was done for a reason, and I'd further assume that the intention was so that the value could possibly be `nil`, since that's the primary impact of writing this as a pointer in the first place. I wouldn't assume that a pointer was added arbitrarily. Given this, I might alter the code in such a way that a `nil` value does become possible. In those cases, the calls to the methods `Add` and `List` would result in nil pointer dereferences. Hopefully, there are tests to catch this. But it would have been a problem that could be avoided altogether.

syft/pkg/catalog.go
wagoodman: @alfredodeza good reference, adding to the PR to close it; this PR takes your suggestion to allow IDs to be an exported field on the `Package` struct and loosens assumptions about ID with regard to catalog assignment.

@luhring Agreed that IDs should be unique, that's not in question. 

I want to clarify why I think non-determinisim is good quality to have in an ID system as a rule of thumb (note, not as a strict "always-do-this" rule).

In #166 , the fact that the ID was being derived from a global value is bad, but what is also bad is that tests were depending on predictable ID values for correctness, and the value was not being controlled by the test (which is what @alfredodeza  was looking to do). If the ID was random from the start such accidental dependence on ID values would not be possible --the tests would not have passed to begin with.

Exposing the existing IDs with the current approach (incrementing number) an end user may assume that there is an implied order to packages. They may assume that the order is stable between runs --which is true, but is only a coincidence due to the current implementation. The truth is, the ID should not convey any intrinsic information --by design or accidentally-- other than the fact that your guaranteed to not have duplicates.

Anywho, I know you weren't asking for a change here @luhring  but wanted to clarify my statements on IDs.

syft/pkg/catalog.go
luhring: 👍 We can move forward with this PR. We'll chat offline and follow up in another PR as necessary.

syft/pkg/catalog.go
luhring: 👍 We can move forward with this PR. We'll chat offline and follow up in another PR as necessary.

luhring (approved):

luhring (approved):
This was a very nice improvement to see come to Syft. 🎉  This looks good to go — pending a few minor things we chatted about earlier (like finding a more distinct and descriptive value to replace `"ownership-by-files"`.)

Jotting down a few notes from our conversation for future improvements to this code:

1. Moving relationship calculation into a package for business logic instead of presentation logic.
2. Types like `map[ID]map[ID]*strset.Set` can be difficult to reason about and prone to misuse. If storing data in this way is deemed necessary, it would be nice to hide this data structure inside an object that provides a clear API to consumers that illustrates the intended way to store and retrieve information.
3. We're probably approaching a point soon where we want to re-examine the project's package layout. We touched briefly on how some locations for vars/consts/funcs/structs are starting to feel _slightly_ out of place as we try to avoid import cycles and other design smells. This dissonance might be indicating the need for a refactor.

Issue #44, opened 2020-06-03
Add distroless image identification


wagoodman:
Should be able to identify a distroless image and determine the version (populating a `distro.Distro` object from `distro.Identify()`)

Note: anchore-engine has logic for this, which this ticket should keep in parity

dakaneye:
Any particular images in mind? I did some googling but syft seems able to resolve the distro, so maybe this image isn't truly distroless:
```
[0000] DEBUG Application config:-arch-and-opensuseleap) syft gcr.io/distroless/base:latest         
configpath: /Users/samdacanay/.syft.yaml
presenteropt: table
output: table
scopeopt: 1
scope: squashed
quiet: false
log:
  structured: false
  levelopt: debug
  level: debug
  filelocation: ""
clioptions:
  configpath: ""
  verbosity: 0
checkforappupdate: false

 ✔ Loaded image         
 ✔ Parsed image         
 ✔ Cataloged image      [0 packages]
[0000]  INFO cataloging image
[0000] DEBUG image: source=DockerDaemon location=gcr.io/distroless/base:latest processingOption=1 from-lib=steroscope
[0000] DEBUG image metadata: digest=sha256:ae9876e34217211b012fd2d9f1f1bff6041e465623546a0a1b3cbcb6217c73df mediaType=application/vnd.docker.distribution.manifest.v2+json tags=[gcr.io/distroless/base:latest] from-lib=steroscope
[0000] DEBUG layer metadata: index=0 digest=sha256:d72a74c56330b347f7d18b64d2effd93edd695fde25dc301d52c37efbcf4844e mediaType=application/vnd.docker.image.rootfs.diff.tar.gzip from-lib=steroscope
[0000] DEBUG layer metadata: index=1 digest=sha256:f47163e8de57e3e3ccfe89d5dfbd9c252d9eca53dc7906b8db60eddcb876c592 mediaType=application/vnd.docker.image.rootfs.diff.tar.gzip from-lib=steroscope
[0000]  INFO identified distro: debian 9
[0000]  INFO building the catalog
[0000] DEBUG cataloger 'dpkg-cataloger' selected '0' files
[0000] DEBUG cataloger 'bundler-cataloger' selected '0' files
[0000] DEBUG cataloger 'python-cataloger' selected '0' files
[0000] DEBUG cataloger 'rpmdb-cataloger' selected '0' files
[0000] DEBUG cataloger 'java-cataloger' selected '0' files
[...]

PR #332, opened 2021-02-21
allow ability to catalog packages from /var/lib/dpkg/status.d/

westonsteimel:
Some debian-based variants (such as [Google's Distroless images](https://github.com/GoogleContainerTools/distroless))
don't write a single file to `/var/lib/dpkg/status`, but rather write
a file per package to `/var/lib/dpkg/status.d/`

related to #44 


westonsteimel:
Example output:

```bash
$ syft gcr.io/distroless/base:latest
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged image      [6 packages]
NAME        VERSION          TYPE
base-files  9.9+deb9u13      deb
libc6       2.24-11+deb9u4   deb
libssl1.1   1.1.0l-1~deb9u3  deb
netbase     5.4              deb
openssl     1.1.0l-1~deb9u3  deb
tzdata      2021a-0+deb9u1   deb
```

luhring:
Hi @westonsteimel, thanks so much for this!

It'd be great to have at least one automated test that verifies this added code does what it's intended to. We'd be happy to help you with this, if you'd like. At a high-level, I'd suggest adding an integration test in `./test/integration/...`, and we could even use `gcr.io/distroless/base` as the fixture image (although we'd want to pin to an exact image digest or tag rather than `latest`).

westonsteimel:
Hello @luhring!  Yes I did suspect having some sort of test would be desirable, I just wasn't quite able to wrap my head around how to implement one, so any help here would be greatly appreciated!  I'll try having a look at some of the existing integration tests when I have a chance this week to see if I can figure it out, but in the meantime if anyone has further suggestions or wants to have a go at it, I'd be perfectly happy to accept that.  Thanks!

luhring:
Hi @westonsteimel, thanks, that makes sense. Let me dive into this again in the next few days and I'll get back to you with a recommendation — hopefully a good one! But no promises 😄

westonsteimel:
Sure, I did have a try with the existing "find dpkg packages" testcase in https://github.com/anchore/syft/pull/332/commits/6c80f9910f9a6cf26392c6d0ec2fb21293628e55.  I just added two new packages under `/var/lib/dpkg/status.d/`, so that verifies that syft detects packages from `/var/lib/dpkg/status` as well as any under `/var/lib/dpkg/status.d/`, which is really what I intended.

wagoodman (approved):
@westonsteimel nice job, thanks for the improvement! Luckily syft is already detecting the underlying os-release information and to close out this issue the last bit was to be able to detect packages in the way you've added.

PR #339, opened 2021-03-08
Add file metadata cataloger

wagoodman:
This PR primarily adds the new file metadata cataloger to syft. There are several other changes that were made because of this:
- The root command is being deprecated in favor of the `packages` subcommand. The root command is still in place and is an alias for the `packages` subcommand, additionally showing a deprecation warning when invoked. 
- Adds a new `power-user` subcommand, which exposes the new file-metadata cataloger functionality.
- Expands on the application configuration to account for power-user functionality and to remove command-specific validations (e.g. presenter options, which should be lazily validated)
- Refactors the presenter go packages to be oriented based on the per-subcommand they represent. Additionally migrates the `syft/presenters` to `internal/presenters` since they are command-specific and not generically helpful for library consumers.
- Migrate the existing `syft/cataloger` package to `syft/pkg/cataloger` since there are now semantically more than one "kind" of cataloger, and the existing ones are tailored just for software packages.
- Updates the JSON schema to reflect the additional (optional) fields exposed by the power-user command.
- Removes the `scope` option from the image metadata section in the JSON presenters and adds application configuration to the JSON presenter `Descriptor` object (currently for the power-user subcommand only).
- Pulls in the latest stereoscope changes (`TarIndex` for content fetching)
- Removes the `Multi*`-oriented methods from the `Resolver` and renames `Resolver` to `FileResolver`, and clarifies the names of the other constituent resolver interfaces.
- Adds CLI testing
- Migrates JSON Schema testing from integration tests to CLI tests
- Adds Benchmark testing


cmd/power_user.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var powerUserOpts = struct {
+	configPath string
+}{}
+
+var powerUserCmd = &cobra.Command{
+	Use:           "power-user [SOURCE]",
+	Short:         "Run bulk operations on container images",
+	Example:       `  {{.appName}} power-user <image>`,
```
alfredodeza, line 23: This is not rendering correctly:

```
 go run main.go power-user --help
Run bulk operations on container images

Usage:
   power-user [SOURCE] [flags]

Examples:
  {{.appName}} power-user <image>
....
```

cmd/cmd.go
```diff
@@ -3,141 +3,85 @@ package cmd
...
-	rootCmd.Flags().StringP(
-		flag, "d", "",
-		"include dockerfile for upload to Anchore Enterprise",
-	)
-	if err := viper.BindPFlag("anchore.dockerfile", rootCmd.Flags().Lookup(flag)); err != nil {
-		fmt.Printf("unable to bind flag '#{flag}': #{err}")
-		os.Exit(1)
+	if activeCmd == rootCmd {
+		// note: cobra supports command deprecation, however the command name is empty and does not report to stderr
+		fmt.Fprintln(os.Stderr, color.New(color.Bold, color.Red).Sprintf("The root command is deprecated, please use the 'packages' subcommand"))
```
alfredodeza, line 69: This is causing problems when running the power-user command:

```
$ go run main.go power-user alpine:3.12
The root command is deprecated, please use the 'packages' subcommand
 ✔ Pulled image
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged image      [14 packages]

{
...
```

.github/workflows/validations.yaml
```diff
@@ -0,0 +1,368 @@
...
+        uses: actions/cache@v2
+        with:
+          path: test/results/benchmark-main.txt
+          # use base sha for PR or new commit hash for main push in benchmark result key
+          key: ${{ runner.os }}-bench-${{ (github.event.pull_request.base.sha != github.event.after) && github.event.pull_request.base.sha || github.event.after }}
+
+      - name: Run benchmark tests
+        id: benchmark
+        run: |
+          REF_NAME=${GITHUB_REF##*/} make benchmark
```
alfredodeza, line 217: it would be useful to add a small comment on why the character replacements below are needed. These are happening because there is no support for newlines when capturing output in a github action. 

Alternatively, this could redirect output to an environment variable, which would make it easier to read (no need for character replacement):

```yaml
        run: |
          echo 'SYFT_BENCHMARK<<EOF' >> $GITHUB_ENV
          REF_NAME=${GITHUB_REF##*/} make benchmark >> $GITHUB_ENV
          echo 'EOF' >> $GITHUB_ENV
```

wagoodman: The goal was to:
1) show results in realtime
2) persist results to a location where they can be cached for future runs
3) prep the output for use with the sticky PR comment (the reason for the character replacements).

For a quick test, I took out the character replacements to see it if still worked with the github action for the sticky comments... and it did! (so no need for the extra replacements +  comments as to why... and it feels simpler 😎 )

syft/pkg/cataloger/java/test-fixtures/pom/colon-delimited-with-equals.pom.properties
```diff
@@ -2,4 +2,4 @@
 #Tue Jul 07 18:59:56 GMT 2020
 groupId:org.anchore
 artifactId: example-java=app-maven
-version:   0.1.0=something
+version:   0.1.0=something
```
luhring, line 5: Nit / future: It'd be nice to find easy ways to prevent these newline removals from adding noise to the diff. (For myself, I just check a box in my editor)

syft/pkg/cataloger/python/package_cataloger.go
```diff
@@ -28,64 +29,34 @@ func (c *PackageCataloger) Name() string {
 }
 
 // Catalog is given an object to resolve file references and content, this function returns any discovered Packages after analyzing python egg and wheel installations.
-func (c *PackageCataloger) Catalog(resolver source.Resolver) ([]pkg.Package, error) {
-	entries, err := c.getPackageEntries(resolver)
-	if err != nil {
-		return nil, err
-	}
+func (c *PackageCataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, error) {
+	// nolint:prealloc
```
luhring, line 33: Can we use a globally toggled-off `prealloc` rule?

wagoodman: the original commits are from branches old enough where the prealloc rule was still around (it's not longer there anymore). I'll remove the `nolint` here

syft/pkg/cataloger/java/test-fixtures/pom/colon-delimited-with-equals.pom.properties
wagoodman: I'd vote adding precommit hooks over synchronizing IDE settings (something I've been wanting to add to the repo anyway)

luhring: Either/both is good w/ me 👍

luhring (approved):
Very nice addition! Thanks for the thoughtful review sessions ❤️

.github/workflows/validations.yaml
wagoodman: looks like I was wrong, character replacements were needed (https://github.com/anchore/syft/pull/354). I like the suggestion you made above (neat here-doc-like trick!) but this would not meet goal 1, yes?

alfredodeza: Not sure what you mean by "show results in realtime" or how this addition is different than the redirection to an environment variable

PR #310, opened 2021-01-19
Add docker image and refactor release pipeline

luhring:
This PR replaces the GoReleaser-based release process with discrete steps as `make` targets.

The main problem addressed here is that we can run release steps on either Linux or macOS, based on what's needed by the step. Before, with GoReleaser doing all steps monolithically, this wasn't possible; and specifically, we could not both **sign our macOS assets** and **build container images**. (Thus #228)

This PR also introduces the concept of a "release follow-up pipeline". Before, we were publishing "draft" releases, but some assets had already been published as final (e.g. the version check file, the homebrew formula update). This PR separates out "follow-up" steps which won't trigger until the draft release has been published.

With these changes in place, this PR also adds a container image build and push mechanism into the release follow-up pipeline.

This PR also tests more artifacts before publishing now... specifically the homebrew formula and the container image.

Other lower-level changes:
- Changes make shell from `sh` to `bash` (to allow for more elegant scripts, e.g. allows `pushd`/`popd`)
- Adds protective shell flags (`-e`, `-o pipefail`)
- Introduces the special `.SILENT:` make target to silence recipe steps more elegantly than prefixing with `@`
- Disambiguates `VERSION` and `VERSION_TAG`. This follows from GoReleaser's separation of the `VERSION` and `TAG` template variables. This distinction matters when used in context (e.g. `VERSION` is `0.12.2`, whereas `VERSION_TAG` is `v0.12.2`).

**This PR does not add documentation** for using the container image. This documentation will be introduced in a future PR, once the new container workflow is confirmed to work, and will accompany other minor docs improvements.

**Note: This PR is likely to require follow-up fixes since it is difficult to test in its entirety with non-real releases.**

PR #344, opened 2021-03-11
Add Docker image to release process

### LINKED ISSUE #228: Docker image

wagoodman:
This adds a docker image build step to goreleaser (publishing to `anchore/syft`). This work does NOT replace the additional work proposed in https://github.com/anchore/syft/pull/310 but instead only delivers the docker image portion for now. 

Note: this adds docker to the mac instance in CI during release.

Closes #228

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,95 +1,89 @@
 name: "Release"
 on:
   push:
-#    # take no actions on push to any branch...
-#    branches-ignore:
-#      - "**"
-#    # ... only act on release tags
-#    tags:
-#      - "v*"
+    # take no actions on push to any branch...
+    branches-ignore:
+      - "**"
+    # ... only act on release tags
+    tags:
+      - "v*"
 
 env:
   GO_VERSION: "1.14.x"
 
 jobs:
-#  wait-for-checks:
-#    runs-on: ubuntu-latest # This OS choice is arbitrary. None of the steps in this job are specific to either Linux or macOS.
-#    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -83,6 +83,8 @@ jobs:
     needs: [wait-for-checks]
     runs-on: macos-latest # Due to our code signing process, it's vital that we run our release steps on macOS.
     steps:
+      - uses: docker-practice/actions-setup-docker@master
```
luhring, line 86: It'd be good to use a tag or commit if available — we've been hit in the past with GH actions changing out from under us

luhring (approved):

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM alpine:latest AS build
```
alfredodeza, line 1: would be nice to use _some_ labels here that follow the [label-schema](http://label-schema.org/rc1/) . Using labels makes it friendlier to others that want to get more information about the container itself, when it was built, and from what commit.

.github/workflows/release.yaml
wagoodman: updated to v1

Dockerfile
wagoodman: added relevant labels

PR #345, opened 2021-03-18
add cataloger for rust crates from Cargo.lock files

### LINKED ISSUE #338: Cataloger for Rust crates from Cargo.lock
It would be great to have the ability to catalog rust crates from Cargo.lock files.

**Why is this needed**:
It would increase the accuracy of the BOM generated by syft

**Additional context**:
I started looking a bit into what this would entail in a branch at https://github.com/westonsteimel/syft/tree/rust-cargo-cataloger.  I wasn't really sure on naming for `syft/pkg/type.go`, I went with `rust-cargo` for the moment, but I think it should probably just be `cargo` or maybe just `rust` as I see it is just `python` for PythonPkg and cargo is that canonical way to manage packages in the rust ecosystem.  

I also made PR https://github.com/package-url/packageurl-go/pull/24 to add support for rust to the go package url module that this project uses, but it looks like @wagoodman already has a request out there for modification to it for debian packages that has been there for awhile so I doubt that would be merged anytime soon, if ever.

westonsteimel:
Fixes #338 


syft/pkg/type.go
```diff
@@ -51,6 +53,8 @@ func (t Type) PackageURLType() string {
 		return packageurl.TypeRPM
 	case GoModulePkg:
 		return packageurl.TypeGolang
+	case RustPkg:
+		return "cargo"
```
westonsteimel, line 57: I did create https://github.com/package-url/packageurl-go/pull/24 to get a type for rust added, but judging by the activity on that repo, it seems unlikely to be merged anytime soon

syft/pkg/type.go
```diff
@@ -17,6 +17,7 @@ const (
 	JavaPkg          Type = "java-archive"
 	JenkinsPluginPkg Type = "jenkins-plugin"
 	GoModulePkg      Type = "go-module"
+	RustPkg          Type = "rust-crate"
```
westonsteimel, line 20: I struggled with what to name this, so certainly up for any feedback here.  These were the ones I considered:

- rust
- cargo
- rust-cargo
- rust-crate

I ended up with `rust-crate` for the moment because that really describes the equivalent of a package format in rust, whereas `cargo` is the tool for managing crates, and that seemed to fit the best based off of the naming for the others

syft/cataloger/rust/parse_cargo_lock_test.go
```diff
@@ -0,0 +1,99 @@
...
+			Language: pkg.Rust,
+			Type:     pkg.RustPkg,
+			Licenses: nil,
+		},
+		{
+			Name:     "winapi",
+			Version:  "0.3.9",
+			Language: pkg.Rust,
+			Type:     pkg.RustPkg,
+			Licenses: nil,
```
alfredodeza, line 67: is licenses not a thing with Rust packages? if it is, it would be useful to capture that in these tests

westonsteimel: No, there isn't any information about licenses available in the Cargo lock file

alfredodeza (approved):

syft/cataloger/rust/parse_cargo_lock_test.go
luhring: This is an interesting question! I'm very new to Rust 😃 . From what I can tell, @westonsteimel is right about licenses not being stored in **cargo.lock** files. It looks like they are stored in **cargo.toml** files (akin to a `package.json` file), but that'd be a different parser than what's tackled here. We could handle that down the road.

[Cargo.toml vs Cargo.lock](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)

[The Manifest Format (cargo.toml)](https://doc.rust-lang.org/cargo/reference/manifest.html)

westonsteimel: Yes, you can also use the [cargo metadata](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html) command to output json with everything, but that would require executing a command

syft/pkg/type.go
wagoodman: yup yup, we ran into the same thing 👍

syft/pkg/type.go
wagoodman: short answer: I think `rust-crate` is a good name 👍 

longer answer:
We struggle with this too with each new package type we add --we've had a lot of internal discussions on what a "package type" is and isn't. We haven't come to any strong conclusions. There are three fields that should be considered together when naming:
1. `Package.Type`, the packaging ecosystem (in this case we're within the `cargo` packaging ecosystem) or the packaging unit (in this case `crate`)
2. `Package.Language`, a definitive language ecosystem the package belongs to (`rust`)
3. `Package.MetadataType`, hints at what the data shape is for the `Package.Metadata` field, which should simply surface raw information from the underlying evidence parsed (in this case, from the `Cargo.lock` file).

Looking at some resources regarding rust packaging (https://rust-cli.github.io/book/tutorial/packaging.html, https://doc.rust-lang.org/cargo/appendix/glossary.html#package, https://github.com/rust-lang/cargo/issues/3380) there seems to be something hinting that the semantics are really about `cargo-packages` instead of `crates`, and to add even more confusion, there is a separate crate file format (https://users.rust-lang.org/t/where-to-learn-about-crate-file-format/11496) that seems to not be talked about much. 

That being said, from a rust-community-perspective, I think that `crate` is the unit that folks are used to thinking in, and cargo is essentially describing. (also, if we're wrong we can change it later to be more accurate)

syft/pkg/metadata.go
```diff
@@ -13,4 +13,5 @@ const (
 	NpmPackageJSONMetadataType MetadataType = "NpmPackageJsonMetadata"
 	RpmdbMetadataType          MetadataType = "RpmdbMetadata"
 	PythonPackageMetadataType  MetadataType = "PythonPackageMetadata"
+	RustCrateMetadataType      MetadataType = "RustCrateMetadata"
```
wagoodman, line 16: This new `MetadataType` is added here but is never used.

syft/cataloger/rust/cargo_metadata_package.go
```diff
@@ -0,0 +1,21 @@
+package rust
+
+import "github.com/anchore/syft/syft/pkg"
+
+type CargoMetadataPackage struct {
```
wagoodman, line 5: This would be great raw information to capture onto `Package.Metadata` (and hint the type via the new `MetadataType` you added onto `Package.MetadataType`).

This also implies that we should migrate the struct to the `syft/pkg` package and rename the struct to `CargoPackageMetadata` to mirror the naming scheme for other metadata structs in the `syft/pkg` package.

syft/pkg/type.go
westonsteimel: Yes, it is a bit confusing, and I now can't seem to find the reference I had remembered seeing in the rust book about a `crate` being similar to the idea of a package from other ecosystems, but I did find https://doc.rust-lang.org/stable/book/ch01-03-hello-cargo.html, and it mentions:

> The last line, [dependencies], is the start of a section for you to list any of your project’s dependencies. In Rust, packages of code are referred to as crates. We won’t need any other crates for this project, but we will in the first project in Chapter 2, so we’ll use this dependencies section then.

This may just be another example like Steve mentions in https://github.com/rust-lang/cargo/issues/3380#issuecomment-266106939 where it isn't always used in a clear way, or it may be as intended, so when you specify dependencies you are specifying the crate you want, in which case I think we've named it appropriately?  

I also saw https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/cargo-workspaces-for-multi-package-projects.html which seems to summarize the hierarchy fairly well.

syft/cataloger/rust/cargo_metadata_package.go
westonsteimel: Okay, I'll try to look at that a bit over the weekend.  I was a bit unsure on this as I think I was mostly looking at the go modules one as an example (and the python poetry file parser for the toml bits) and it didn't have a metadata component defined under pkg.

wagoodman: no problem! Taking a look at the poetry metadata we could have done the same thing (moved it up to `pkg/` and expose it via `Package.Metadata`) but it seems like we didn't (I admit we're being somewhat inconsistent here). 

We can also merge this PR as is without capturing the extra cargo metadata onto `Package.Metadata` --we can always do that in a follow up PR if we find we really need it. Let us know which direction you want to go in, either is OK 👍

westonsteimel: So I had a try at this in another branch https://github.com/westonsteimel/syft/commit/8b620de00cfeb0d62619cf5257dc507b6ba2f94b, but I'm running into issues getting the tests to pass.  First the unit tests passed, but the integration tests did not, and I was able to figure out that I needed to increment and generate a new json schema file (1.0.3).  I did that and the integration tests now pass, but the `TestJsonDirsPresenter` and `TestJsonImgsPresenter` unit tests fail because they expect json schema file version 1.0.2 and now get 1.0.3, and I haven't yet figured out how to resolve that one.

wagoodman: awesome! I cherry picked that commit over and made a couple minor adjustments:
- We have snapshots of expected output for the json presenter tests. We use test flags to update the test snapshots: `go test ./syft/presenters/json -update` and then manually ensure the snapshots are what we expect before committing (I did this already, so no need to update)
- I updated the `MetadataType` to reflect where the information was parsed from (a cargo package section)

wagoodman (approved):
Solid work, thanks for the contribution! 🚀

syft/pkg/metadata.go
wagoodman: nvm, it's used now 👍

alfredodeza:
congrats @westonsteimel ! Excellent work

westonsteimel:
Thanks everyone!

Issue #330, opened 2021-02-17
parse known fields in packages, allowing leniency 


alfredodeza:
As shown in issue https://github.com/anchore/syft/issues/327 - there is a need to be lenient on some of the fields when parsing. This leniency is not only meant for Python packages but should also be applied elsewhere.

The way the PKG-INFO parser works is by looking at each line and making the following assumptions:

* if the line is empty it skips to the next line
* if the line starts with whitespace it captures it as a "field body conttinuation"
* if it doesn't start with whitespace then it looks if there is a `:` to capture the field and the value

The problem with this approach is that it is trying to programmatically parse the file, assuming that any line that has at least one instance of `:` is a valid field, which it isn't.

There is also no need to capture multi-line descriptions.  And lastly, there are packages out there that have fully invalid representations of the format (as seen in the linked issue). 

In the case of Python, the [fields are known](https://packaging.python.org/specifications/core-metadata/), so this proposal should be about preference of known fields that are useful for `syft` rather than programmatically discovering every single field in the package.

In the case of the PR, it doesn't really fix the underlying programmatic access of fields, but rather, warns about seemingly invalid fields (https://github.com/anchore/syft/pull/328). The spec says "Summary" is a single line. So the parser should consume only one line and ignore the rest. That behavior will prevent unnecessary hard errors.

spiffcs:
@anchore/tools I've moved this issue onto our internal board for consideration when we start looking at future enhancements to make

wagoodman:
I think we've taken this stance, but maybe not universally so. As written this is a little too broad to be actionable.

PR #348, opened 2021-03-20
Allow for multiple pom.properties delimiters

wagoodman:
This PR fixes the parsing of pom.properties to allow for both `=` and `:` delimiters.

Additionally this PR loosens the error handling on the java parsing execution path to show warnings for individual parsing errors instead of bubbling up small parsing errors back up to the main catalog function. This allows for java packages unrelated to the parser problems to still be cataloged.


luhring (approved):

alfredodeza:
This seems related to https://github.com/anchore/syft/issues/330 that explains that strictness in parsing causes several omissions. I'd propose raising the priority of that ticket so that we can enable permissive parsing for _all parsers_

Issue #335, opened 2021-03-02
Add file metadata cataloger


wagoodman:
Add a new "power-user" command configuration to capture metadata for each file in an image, including:
- File content digests, with selectable digest algorithms (md5, sha1, sha256 at least)
- File mode
- UID
- GID

Configuration should support:
- Allow for enable/disable of the cataloger as a whole
- Takes a scope option as configuration (default: bind to user config/default)
- Should be able to specify zero or more digest algorithms (default: none)
- Should be able to specify a list of globs to capture metadata information for (default: “**”)

JSON Presentation:
- Adds a new “fileMetadata” section that is a list of objects, each object should have a file path, a layer, and the associated metadata.

luhring:
One thought on the "Configuration should support" section: are all of these items necessary for the MVP of this feature? My understanding is we're intentionally not committing to how we're exposing this functionality to users yet; and instead, for now, our goal is to satisfy a need by one particular downstream consumer — Anchore Engine. If Anchore Engine doesn't need a given piece of configurability, I'd suggest we defer adding it for now, so that we don't introduce more dimensions we need to test, and more scope that takes our time in the short term.

PR #357, opened 2021-03-26
Add link destination to file metadata

wagoodman:
- Added `LinkDestination` to `source.FileMetadata`, which is shown in the json output only when there is a value
- Makes `digests` an optional field in the json output
- Adds tests for the power-user presenter
- Renames `source.FileType` strings to be capitalized (consistent with similar fields)
- Fixes a bug where there are digests with a single value which have empty `algorithm` and `value` fields.
- Fixed a bug where sorting file metadata was not stable

Related to https://github.com/anchore/syft/issues/335

### DIFF
--- a/internal/presenter/poweruser/json_presenter_test.go
+++ b/internal/presenter/poweruser/json_presenter_test.go
@@ -1,0 +1,172 @@
+package poweruser
+
+import (
+	"bytes"
+	"flag"
+	"testing"
+
+	"github.com/sergi/go-diff/diffmatchpatch"
+
+	"github.com/anchore/syft/syft/file"
+
+	"github.com/anchore/go-testutils"
+	"github.com/anchore/syft/internal/config"
+	"github.com/anchore/syft/syft/distro"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/source"
+)
+
+var updateJSONGoldenFiles = flag.Bool("update-json", false, "update the *.golden files for json presenters")
+
+func must(c pkg.CPE, e error) pkg.CPE {
+	if e != nil {
+		panic(e)
+	}
+	return c
+}
+
+func TestJSONPresenter(t *testing.T) {
+	var buffer bytes.Buffer
+
+	catalog := pkg.NewCatalog()
+
+	catalog.Add(pkg.Package{
+		ID:      "package-1-id",
+		Name:    "package-1",
+		Version: "1.0.1",
+		Locations: []source.Location{
+			{
+				RealPath: "/a/place/a",
+			},
+		},
+		Type:         pkg.PythonPkg,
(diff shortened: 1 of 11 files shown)


internal/presenter/poweruser/json_presenter_test.go
```diff
@@ -0,0 +1,179 @@
...
+	if err := NewJSONPresenter(cfg).Present(&buffer); err != nil {
+		t.Fatal(err)
+	}
+	actual := buffer.Bytes()
+
+	if *updateJSONGoldenFiles {
+		testutils.UpdateGoldenFileContents(t, actual)
+	}
+
+	var expected = testutils.GetGoldenFileContents(t)
```
luhring, line 172: nit: It'd be a bit nicer to read if we didn't mix variable definition styles (see `actual := ...` a few lines above)

syft/file/digest_cataloger.go
```diff
@@ -74,16 +74,18 @@ func (i *DigestsCataloger) catalogLocation(resolver source.FileResolver, locatio
...
-			result[idx] = Digest{
-				Algorithm: cleanAlgorithmName(i.hashes[idx].String()),
-				Value:     fmt.Sprintf("%+x", hasher.Sum(nil)),
-			}
+	// only capture digests when there is content. It is important to do this based on SIZE and not
+	// FILE TYPE. The reasoning is that it is possible for a tar to be crafted with a header-only
+	// file type but a body is still allowed.
+	for idx, hasher := range hashers {
+		result[idx] = Digest{
+			Algorithm: cleanAlgorithmName(i.hashes[idx].String()),
```
luhring, line 87: Optional but could help — It looks like in most calls to `cleanAlgorithmName`, we have a value but are calling `.String()` on that value just in time. Usually when I notice this, I realize I should at minimum consider changing the parameter type from `string` to `fmt.Stringer`, to clean up the consuming code and consolidate the extra `.String()` work in the called function. In this particular case though, it looks like we could possibly go even further and take a parameter of type `crypto.Hash` instead. This would immediately simplify 3 out of the 4 calls being made to `cleanAlgorithmName`. The 4th call has a string already, because `NewDigestsCataloger` is taking a slice of strings. But it might be advantageous to modify this too to use `crypto.Hash` to ensure usable hash algorithms are supplied. If conversion from a string is needed because of user configuration, that should probably be pushed away from the application logic toward the user boundary as much as possible.

luhring (approved):

syft/file/digest_cataloger.go
wagoodman: I agree with the overall logic of moving the string work to the caller of NewDigestsCataloger, but I think this is a bit of a trade off since the `cleanAlgorithmName` is used for a couple purposes and will now need to be exported for both purposes (parsing strings and cleaning hash names).

PR #364, opened 2021-03-29
Ignore failures on benchmark sticky comment

wagoodman:
PRs from forks will not be able to add a sticky comment to PRs, so this step should be allowed to run without indicating a failure if the post fails (for any reason, not just the single reason listed here).

### DIFF
--- a/.github/workflows/validations.yaml
+++ b/.github/workflows/validations.yaml
@@ -216,6 +216,7 @@ jobs:
 
       - name: Update PR benchmark results comment
         uses: marocchino/sticky-pull-request-comment@v2
+        continue-on-error: true
         with:
           header: benchmark
           message: |


.github/workflows/validations.yaml
```diff
@@ -216,6 +216,7 @@ jobs:
 
       - name: Update PR benchmark results comment
         uses: marocchino/sticky-pull-request-comment@v2
+        continue-on-error: true
```
alfredodeza, line 219: if there is an error this should probably get advertised somehow

wagoodman: the error output will be in the logs for troubleshooting if needed, however, we don't want to declare the benchmark validations as failed just because the posting of the comment on the PR fails.

PR #366, opened 2021-04-01
Use sorted artifacts for consistent SBoM output in table, cyclonedx, and json presenters

### LINKED ISSUE #331: Syft BOM ordering is not always consistent
**What happened**:

Running BOM generation twice on the same image results in different output each time.

We expect output to remain the same each time we run syft against an image. I think we should be sorting both on package name and version. 

Ultimately in our pipeline we'd like to run diff on the output to determine if any packages have changed in an image. Currently not possible since the same image returns different output

Note: Similar thing happens for both table and Json output

```
$ syft node:12 > bom.out
$ syft node:12 > bom2.out
$ diff bom.out bom2.out
5d4
< agent-base                        4.3.0                              npm
6a6
> agent-base                        4.3.0                              npm
10d9
< ansi-regex                        4.1.0                              npm
12a12
> ansi-regex                        4.1.0                              npm
201d200
< iferr                             0.1.5                              npm
202a202
> iferr                             0.1.5                              npm
235d234
< isarray                           0.0.1                              npm
236a236
> isarray                           0.0.1                              npm
590d589
< ms                                2.0.0                              npm
591a591
> ms                                2.0.0                              npm
633d632
< p-limit                           2.3.0                              npm
634a634
> p-limit                           2.3.0                              npm
703d702
< retry                             0.10.1                             npm
704a704
> retry                             0.10.1                             npm
707,708d706
< safe-buffer                       5.2.0                              npm
< safe-buffer                       5.1.2                              npm
709a708,709
> safe-buffer                       5.1.2                              npm
> safe-buffer                       5.2.0                              npm
740d739
< string-width                      2.1.1                              npm
741a741
> string-width                      2.1.1                              npm
810d809
< yallist                           2.1.2                              npm
811a811
> yallist                           2.1.2                              npm
```

**Environment**:
- Output of `syft version`:

syft version
```
Application:   syft
Version:       0.12.6
BuildDate:     2021-02-05T16:42:47Z
GitCommit:     8d838b18a85d2c2f0ee976cf2af491913dbcf98b
GitTreeState:  clean
Platform:      linux/amd64
GoVersion:     go1.14.14
Compiler:      gc
```

cat /etc/os-release
```
NAME="CentOS Linux"
VERSION="8"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Linux 8"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"
CENTOS_MANTISBT_PROJECT="CentOS-8"
CENTOS_MANTISBT_PROJECT_VERSION="8"
```

zhill:
Switches use of catalog.Enumerated() to catalog.Sorted() and adds the location as a component of the sort key.

Fixes #331

### DIFF
--- a/internal/presenter/packages/cyclonedx_document.go
+++ b/internal/presenter/packages/cyclonedx_document.go
@@ -34,7 +34,7 @@ func NewCycloneDxDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) Cyc
 	}
 
 	// attach components
-	for p := range catalog.Enumerate() {
+	for _, p := range catalog.Sorted() {
 		component := CycloneDxComponent{
 			Type:       "library", // TODO: this is not accurate
 			Name:       p.Name,
--- a/internal/presenter/packages/table_presenter.go
+++ b/internal/presenter/packages/table_presenter.go
@@ -25,7 +25,7 @@ func (pres *TablePresenter) Present(output io.Writer) error {
 	rows := make([][]string, 0)
 
 	columns := []string{"Name", "Version", "Type"}
-	for p := range pres.catalog.Enumerate() {
+	for _, p := range pres.catalog.Sorted() {
 		row := []string{
 			p.Name,
 			p.Version,
--- a/syft/pkg/catalog.go
+++ b/syft/pkg/catalog.go
@@ -164,6 +164,9 @@ func (c *Catalog) Sorted(types ...Type) []*Package {
 	sort.SliceStable(pkgs, func(i, j int) bool {
 		if pkgs[i].Name == pkgs[j].Name {
 			if pkgs[i].Version == pkgs[j].Version {
+				if pkgs[i].Type == pkgs[j].Type {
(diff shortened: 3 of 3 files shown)


zhill:
Note that it won't be possible to make all JSON presenter output fully matching for diffs because of the use of UUIDs for artifacts. Similarly, in the CycloneDX output there is a timestamp component so that will trigger diffs as well. This just addresses the sorting of the artifact entries themselves. I looked at the artifactRelationships array but since the IDs are not consistent across executions, I didn't worry about sorting those since the entries use the artifact UUIDS and thus would not  be comparable across invocations.

wagoodman (approved):

luhring:
Do we think it's possible to add tests to confirm this behavior?

PR #367, opened 2021-04-01
Add secrets search capability

### LINKED ISSUE #362: Add secrets cataloger
Add a cataloger which can surface discovered secrets in the `power-user` command JSON output. Should be configurable:
- search by regex, provide a set of regexes.
- configurable skip conditions (file size)
- configurable output (do/don't show secret)

Suggested altered search strategy: should **always** search against all files within the image (not the squashed representation). 

refs:
- https://github.com/anchore/anchore-engine/blob/master/anchore_engine/conf/analyzer_config.yaml#L22-L31
- https://github.com/anchore/anchore-engine/blob/master/anchore_engine/analyzers/modules/13_secret_content_search.py

wagoodman:
Adds the ability to catalog secrets from regexes (under the `power-user` subcommand).

Specific changes:
- Added new `file.SecretsCataloger` (this can one day be repurposed into a generic search cataloger, but there was no specific need a this time)
- Adds the idea of a `file.SearchResult` to show a value, which pattern matched, and the location within the file. Note: anchore-engine requires lines, however, internally a seek position is required. Both are useful within different contexts so both information sets are on the struct.
- `power-user` tasks are now executed asynchronously 
- Updates the application config elements to respond as `loaders` and `parsers` to help push setting default values and validation down to individual structs.
- Added ETUI elements for the secrets cataloger, digests cataloger, and metadata cataloger
- Updates JSON schema to 1.0.5 with additional `secrets` section
- Adds `Size` to the file metadata parser (this was accidentally left off in the addition of the file metadata cataloger and was used in one of the secrets cataloger iterations... though not the final iteration) 
- Updates the config example with the secrets config options

Closes #362


syft/file/digest_cataloger.go
```diff
@@ -22,13 +29,22 @@ func NewDigestsCataloger(hashes []crypto.Hash) (*DigestsCataloger, error) {
 
 func (i *DigestsCataloger) Catalog(resolver source.FileResolver) (map[source.Location][]Digest, error) {
 	results := make(map[source.Location][]Digest)
+	var locations []source.Location
 	for location := range resolver.AllLocations() {
+		locations = append(locations, location)
+	}
+	stage, prog := digestsCatalogingProgress(int64(len(locations)))
+	for location := range resolver.AllLocations() {
```
luhring, line 37: It looks like we're ranging through `resolver.AllLocations()` twice now. At first it wasn't obvious why, but I'm now thinking it's because we can't get a slice upfront from the `resolver` object, and then we need to access the total count of locations for the sake of reporting to the UI. Is that right?

syft/file/secrets_cataloger.go
```diff
@@ -0,0 +1,197 @@
...
+
+var DefaultSecretsPatterns = map[string]string{
+	"aws-access-key":     `(?i)aws_access_key_id["'=:\s]*?(?P<value>(A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16})`,
+	"aws-secret-key":     `(?i)aws_secret_access_key["'=:\s]*?(?P<value>[0-9a-zA-Z/+]{40})`,
+	"pem-private-key":    `-----BEGIN (\S+ )?PRIVATE KEY(\sBLOCK)?-----((?P<value>(\n.*?)+)-----END (\S+ )?PRIVATE KEY(\sBLOCK)?-----)?`,
+	"docker-config-auth": `"auths"((.*\n)*.*?"auth"\s*:\s*"(?P<value>[^"]+)")?`,
+	"generic-api-key":    `(?i)api(-|_)?key["'=:\s]*?(?P<value>[A-Z0-9]{20,60})["']?(\s|$)`,
+}
+
+func GenerateSearchPatterns(basePatterns map[string]string, additionalPatterns map[string]string, excludePatternNames []string) (map[string]*regexp.Regexp, error) {
```
luhring, line 29: It'd be great to document this function — its implementation appears to be pretty involved

syft/file/secrets_cataloger.go
```diff
@@ -0,0 +1,197 @@
...
+	var regexObjs = make(map[string]*regexp.Regexp)
+	var errs error
+
+	addFn := func(name, pattern string) {
```
luhring, line 33: It would help readability to move this function definition outside of this function, and give it a name that describes what it's doing. I see that it needs a reference to the `regexObjs` object. I could see that being a function parameter, or, we could codify `map[string]*regexp.Regexp)` as its own type and make this a pointer receiver method instead of a standalone function

syft/file/secrets_cataloger.go
```diff
@@ -0,0 +1,197 @@
...
+		// always enable multiline search option for extracting secrets with multiline values
+		obj, err := regexp.Compile(`(?m)` + pattern)
+		if err != nil {
+			errs = multierror.Append(errs, fmt.Errorf("unable to parse %q regular expression: %w", name, err))
+		}
+		regexObjs[name] = obj
+	}
+
+	matchesExclusion := func(name string) bool {
```
luhring, line 42: It'd be nice to move this function definition outside of the `GenerateSearchPatterns` function. Even though it's clearly serving just this function, it'd help the high-level details of `GenerateSearchPatterns` be more readily parsed

syft/file/secrets_search_by_line_strategy.go
```diff
@@ -0,0 +1,123 @@
...
+			if secret != nil {
+				secrets = append(secrets, *secret)
+			}
+		}
+	}
+
+	return secrets, nil
+}
+
+func extractSecretFromPosition(resolver source.FileResolver, location source.Location, name string, pattern *regexp.Regexp, lineNo, lineOffset, seekPosition int64) (*SearchResult, error) {
```
luhring, line 71: This function looks large and complex. What are your thoughts about decomposing it?

ui/event_handlers.go
```diff
@@ -301,14 +301,145 @@ func PackageCatalogerStartedHandler(ctx context.Context, fr *frame.Frame, event
...
+		spin := color.Green.Sprint(completedStatus)
+		title = tileFormat.Sprint("Cataloged secrets")
+		auxInfo := auxInfoFormat.Sprintf("[%d secrets]", prog.SecretsDiscovered.Current())
+		_, _ = io.WriteString(line, fmt.Sprintf(statusTitleTemplate+"%s", spin, title, auxInfo))
+	}()
+	return err
+}
+
+// FileMetadataCatalogerStartedHandler shows the intermittent secrets searching progress.
+// nolint:dupl
```
luhring, line 358: Just curious why this was needed — are these two implementations different enough (e.g. due to a specific type) that they can't be factored out?

luhring (changes requested):
This looks like a lot of good work. A few comments on readability and understandability, let me know if you want to talk through these

internal/config/application.go
```diff
@@ -30,34 +37,57 @@ type Application struct {
...
 
+// init loads the default configuration values into the viper instance (before the config values are read and parsed).
+func (cfg Application) loadDefaultConfig(v *viper.Viper) {
+	// set the default values for primitive fields in this struct
+	v.SetDefault("check-for-app-update", true)
+
+	// for each field in the configuration struct, see if the field implements the loader interface and invoke it if it does
+	value := reflect.ValueOf(cfg)
+	for i := 0; i < value.NumField(); i++ {
+		// check to see if the field implements the loader interface. note: the loader method receiver is NOT a pointer
```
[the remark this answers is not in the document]
wagoodman: I'll combine them, the `note` was the extra important bit

internal/config/packages.go
```diff
@@ -1,9 +1,15 @@
...
+import "github.com/spf13/viper"
+
 type Packages struct {
 	Cataloger catalogerOptions `yaml:"cataloger" json:"cataloger" mapstructure:"cataloger"`
 }
 
-func (cfg *Packages) build() error {
-	return cfg.Cataloger.build()
+func (cfg Packages) loadDefaultConfig(v *viper.Viper) {
+	v.SetDefault("package.cataloger.enabled", true)
```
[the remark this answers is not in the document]
wagoodman: these constants would only be used in one spot (this one) ... we wouldn't be able to parameterize the other place where parts of this constant are used (e.g. the struct tag listing `package` to an individual field). I'm quite open to ways that make this safer, it since they are tied to struct tags it would be easy to go to the effort to add a const that has the same problem (someone makes an update elsewhere and it still breaks accidentally).

The point of the change in this PR was to at least bring the setting of the default values into the same file where the values reside (in the struct definition)... it was getting easier to accidentally forget to update the global function where they were all set.

syft/file/digest_cataloger.go
wagoodman: that's exactly right; I considered changing the resolver to return the slice, but decided against making an interface change for a UI enhancement.

ui/event_handlers.go
wagoodman: staying consistent with existing patterns until we decide to replace some of this with glint (which case this all changes). I could refactor all of the handlers to accept the event primitives that are parsed from each event, but again, think time is better served elsewhere since glint is on the horizon here.

syft/file/secrets_cataloger.go
wagoodman: This also has the added function of accumulating multiple errors instead of spreading that logic around (the `multierror.Append`.

I'd vote leaving this here.

syft/file/secrets_cataloger.go
wagoodman: agreed, I'll move it out.

syft/file/secrets_search_by_line_strategy.go
wagoodman: I'll split out the "seeking" logic from the "finder" logic to help out with complexity. (I don't recommend we break apart the "finder" logic into multiple parts.)

internal/config/application.go
```diff
@@ -18,6 +17,14 @@ import (
 
 var ErrApplicationConfigNotFound = fmt.Errorf("application config not found")
 
+type loader interface {
```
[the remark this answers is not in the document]
wagoodman: changed to `defaultValueLoader`

syft/file/secrets_cataloger.go
wagoodman: additionally I've moved this functionality to its own file 👍

syft/file/digest_cataloger.go
luhring: Makes sense. A "nit" comment would be to range through the `locations` instead (in the second loop) since that collection is already available in scope, rather than access the collection via multiple ways in the same set of lines, but that's an optional comment wrt this PR 👍

luhring (approved):
Looks good! Thanks for all the consideration that went into this ❤️

syft/file/digest_cataloger.go
wagoodman: good call! That was an oversight, I'll adjust

PR #370, opened 2021-04-06
Add file contents cataloger

### LINKED ISSUE #369: Implement selective "retrieve files"
Add file retriever cataloger responsible for retrieving entire file contents for select files:

Configuration:
- Allow for enable/disable
- Takes a scope option as configuration (default: bind to user config/default)
- List of globs that should be persisted (default: [])
- “Max file size”, skipping files over the configured size; default = 1MB

JSON Presentation:
- Adds a new “fileContents” section that is a list of objects, each object should have a file path, a layer, and the base64 encoded file content.

wagoodman:
Enables capturing file content for select files (or glob).

Closes #369


internal/config/application.go
```diff
@@ -38,6 +38,7 @@ type Application struct {
 	Package           Packages       `yaml:"package" json:"package" mapstructure:"package"`
 	FileMetadata      FileMetadata   `yaml:"file-metadata" json:"file-metadata" mapstructure:"file-metadata"`
 	Secrets           Secrets        `yaml:"secrets" json:"secrets" mapstructure:"secrets"`
+	Contents          contents       `yaml:"contents" json:"contents" mapstructure:"contents"`
```
wagoodman, line 41: Open question @luhring @alfredodeza : I think I should rename this from `contents` to `file-contents` to be consistent with `file-metadata` and another cataloger added in another PR. Any votes / thoughts here?

luhring: I think that makes sense

wagoodman: just made the adjustment

syft/file/contents_cataloger.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"io"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/source"
+)
+
+type ContentsCataloger struct {
+	globs              []string
+	skipFilesAboveSize int64
```
luhring, line 15: Nit: Clarify unit, e.g. `skipFilesAboveSizeInBytes`

syft/file/contents_cataloger.go
```diff
@@ -0,0 +1,68 @@
...
+func (i *ContentsCataloger) catalogLocation(resolver source.FileResolver, location source.Location) (string, error) {
+	contentReader, err := resolver.FileContentsByLocation(location)
+	if err != nil {
+		return "", err
+	}
+	defer contentReader.Close()
+
+	buf := &bytes.Buffer{}
+	if _, err = io.Copy(base64.NewEncoder(base64.StdEncoding, buf), contentReader); err != nil {
+		return "", fmt.Errorf("unable to observe contents of %+v: %+v", location.RealPath, err)
```
luhring, line 64: Nit: Do we still normally prefer `%w` for errors?

syft/file/contents_cataloger_test.go
```diff
@@ -0,0 +1,89 @@
...
+func TestContentsCataloger(t *testing.T) {
+	allFiles := []string{"test-fixtures/last/path.txt", "test-fixtures/another-path.txt", "test-fixtures/a-path.txt"}
+
+	tests := []struct {
+		name       string
+		globs      []string
+		maxSize    int64
+		files      []string
+		expected   map[source.Location]string
+		catalogErr bool
```
luhring, line 19: `catalogErr` — This doesn't look used. If that's true, I'd remove it from the struct and remove code that uses it in the test body.

syft/pkg/cataloger/common/generic_cataloger_test.go
```diff
@@ -51,7 +51,7 @@ func TestGenericCataloger(t *testing.T) {
 	}
 
 	if len(actualPkgs) != len(expectedPkgs) {
-		t.Fatalf("unexpected packages len: %d", len(actualPkgs))
+		t.Fatalf("unexpected packages len: %d != %d", len(expectedPkgs), len(actualPkgs))
```
luhring, line 54: I'd suggest changing this to use an `assert` function from testify to handle this. That should also make it more clear to the consumer of the test output which value is which.

luhring (changes requested):

syft/file/contents_cataloger.go
wagoodman: yup yup, I'll adjust

syft/pkg/cataloger/common/generic_cataloger_test.go
wagoodman: Pulled in `assert.Len` to address this 👍

luhring (approved):

wagoodman:
@luhring rebased to pull in latest changes from main (including JSON schema changes). The largest change was reconciling the schemas into v1.1.0 instead of v1.0.6.

PR #374, opened 2021-04-07
Add file classification cataloger

### LINKED ISSUE #372: Implement binary package identification
Should support identification for the following binaries:
- go 
- python
- busybox

Should at least have the following configuration options:
- Allow for enable/disable
- Takes a scope option as configuration (default: bind to user config/default)

JSON Presentation:
- Adds a new “binaryDetection”, "binaryClassifications" section that is a list of objects, each object should have a file path, a layer, the binary classification (go, busybox, or python), and the version detected. Note: consider a more generic "fileClassifications" section if it makes sense.

wagoodman:
Adds binary classification via a new `file.ClassificationCataloger` to the `power-user` command. The more generic "file" classification over "binary" classification was chosen for a few reasons:
1. the semantics in anchore-engine collects file evidence for a "binary package". In syft "package" is reserved to mean something very specific and typically does not extrapolate the existence of a "package" without direct package metadata. 
2. inferring multiple file evidence into strictly a "binary" could be over presumptuous on syft's part, where toolbox tools typically surface raw information over inferring or synthesizing information (though there are exceptions here)
3. Since synthesizing a "binary package" or similar is not preferred, having a root-level "binaryClassification" json section seemed like the incorrect choice.
4. All of the same information needed could be surfaced by describing what is actually in the image (e.g. X location was classified as a "python-binary" and Y location was classified as a "go-binary") allowing for a downstream consumer (such as anchore-engine) to collect these classifications as evidence for a "binary package". This more closely follows the toolbox strategy of "surface raw information".

This PR contains the following changes / additions:
- Adds a new `file.ClassificationCataloger`
- Adds a new `file.Classifier` based on file path selection by regex, content selection by regex, and metadata capturing by named capture groups from the content regexes (very similar to anchore-engine in this regard).
- Adds new `file.DefaultClassifiers` that cover detecting the same file evidence that anchore-engine requires ("python-binary", "cpython-source", "go-binary", "go-binary-hint", "busybox-binary")
- Adds tests around the existing `internal.MatchCaptureGroups` (and renames to a more specific name: `internal.MatchNamedCaptureGroups`)

What is NOT done in this PR but we have an opportunity to do later:
- Expose the ability for users to craft their own file classifiers via configuration

Closes #372


syft/file/classfication_cataloger_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+
+			c, err := NewClassificationCataloger(DefaultClassifiers)
+			if err != nil && !test.constructorErr {
+				t.Fatalf("could not create cataloger (but should have been able to): %+v", err)
+			} else if err == nil && test.constructorErr {
+				t.Fatalf("expected constructor error but did not get one")
+			} else if test.constructorErr && err != nil {
```
luhring, line 107: Nit: I might suggest checking for this case _first_, so you can return early without needing any other logic. The remaining checks could be a separate statement altogether.

syft/file/classfication_cataloger_test.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+func TestClassifierCataloger_DefaultClassifiers_PositiveCases(t *testing.T) {
+	tests := []struct {
+		name           string
+		fixtureDir     string
+		location       string
+		expected       []Classification
+		constructorErr bool
+		catalogErr     bool
```
luhring, line 17: What is the function of `catalogErr`?

syft/file/classfication_cataloger_test.go
```diff
@@ -0,0 +1,164 @@
...
+			if err != nil {
+				t.Fatalf("could not create resolver: %+v", err)
+			}
+
+			actualResults, err := c.Catalog(resolver)
+			if err != nil && !test.catalogErr {
+				t.Fatalf("could not catalog (but should have been able to): %+v", err)
+			} else if err == nil && test.catalogErr {
+				t.Fatalf("expected catalog error but did not get one")
+			} else if test.catalogErr && err != nil {
```
luhring, line 126: Not required here, maybe useful in the future — testify has some nifty error-testing functions like `assert.Error`, `assert.NoError`, and `assert.EqualError` that could be included on the test cases directly to reduce the need for the test body to test for each permutation.

syft/file/classification_cataloger.go
```diff
@@ -0,0 +1,38 @@
...
+	return &ClassificationCataloger{
+		classifiers: classifiers,
+	}, nil
+}
+
+func (i *ClassificationCataloger) Catalog(resolver source.FileResolver) (map[source.Location][]Classification, error) {
+	results := make(map[source.Location][]Classification)
+
+	numResults := 0
+	for location := range resolver.AllLocations() {
```
luhring, line 22: Open question (for longer term) — do we think it's worth considering a slice-returning method on the resolver, if consumers are having to count out enumerated results?

luhring (commented):
Looks good — "commenting" here so we can answer some open questions. Once those are answered I'll proceed with approve/request changes

syft/file/classfication_cataloger_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestClassifierCataloger_DefaultClassifiers_PositiveCases(t *testing.T) {
+	tests := []struct {
+		name           string
+		fixtureDir     string
+		location       string
+		expected       []Classification
+		constructorErr bool
```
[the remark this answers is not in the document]
wagoodman: copy-pasta from another test, I'll remove

syft/file/classfication_cataloger_test.go
wagoodman: copy-pasta from another test, I'll remove

syft/file/classfication_cataloger_test.go
wagoodman: ahh yeah, I learned that in your last PR for templating, I can adjust here

syft/file/classification_cataloger.go
wagoodman: the thought was if there are several locations, a chan would generate them one at a time, so a caller doesn't have to hold all locations in memory --we have multiple callers here in parallel. This interface method has come up a few times in the same way in PR comments ("why a channel over a slice?")... so it sounds like there is an overwhelming desire to make this a slice

internal/regex_helpers.go
```diff
@@ -0,0 +1,39 @@
...
+	// encapsulated in an optional capture group, there will be results for each character, but the results will match
+	// on nothing. The only "true" match will be at the end ("match against").
+	allMatches := regEx.FindAllStringSubmatch(content, -1)
+	for matchIdx, match := range allMatches {
+		// fill a candidate results map with named capture group results, accepting empty values, but not groups with
+		// no names
+		results := make(map[string]string)
+		for nameIdx, name := range regEx.SubexpNames() {
+			if nameIdx <= len(match) && len(name) > 0 {
+				results[name] = match[nameIdx]
```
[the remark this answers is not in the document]
wagoodman: yup yup, since the intention is that the return is a mapping, and the mapping key is the capture group name, there can be only one value. My take: callers shouldn't be providing regexes with multiple capture groups with the same name.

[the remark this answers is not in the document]
luhring: Okay cool, that makes sense. In that case, I'd leave just a "nit" here:

>callers shouldn't be providing regexes with multiple capture groups with the same name

I'd say this explicitly in the doc comment, since the caller would have no other way to find that out until runtime.

internal/regex_helpers.go
```diff
@@ -0,0 +1,39 @@
...
+			}
+		}
+		// note: since we are looking for the first best potential match we should stop when we find the first one
+		// with non-empty results.
+		if len(results) > 0 {
+			foundNonEmptyValue := false
+			for _, value := range results {
+				if value != "" {
+					foundNonEmptyValue = true
```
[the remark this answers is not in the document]
wagoodman: simplified the best I could in 59f0ffd6f7e1e318d747d9d5564d070e437b3439 , the remaining complications appear to be more related to the stdlib regex API at this point --shout out if you see other opportunities here that are worth while.

internal/regex_helpers.go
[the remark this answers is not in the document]
wagoodman: done in 527dc39c82857f6df5717bb213febfa4f1e1c36d

syft/file/classfication_cataloger_test.go
wagoodman: righto, this will be addressed a different way (using the new `assert.NoError` and removing the bool test options)

syft/file/classfication_cataloger_test.go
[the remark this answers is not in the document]
wagoodman: done in 47f1a5c221e9fc611875fa08dec70341dd1253be

syft/file/classfication_cataloger_test.go
wagoodman: Done in 47f1a5c221e9fc611875fa08dec70341dd1253be

syft/file/classfication_cataloger_test.go
wagoodman: Done in 47f1a5c221e9fc611875fa08dec70341dd1253be

wagoodman:
@luhring  just rebased with the latest json schema changes (v1.1.0)

internal/regex_helpers.go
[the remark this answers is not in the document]
luhring: Thanks!

syft/file/classification_cataloger.go
luhring: Okay, cool. Maybe at least as another option. If other consumers are getting value specifically from the `chan` approach, that'd be good to leave. No change needed in this PR!

luhring (approved):
Looks great!

Issue #371, opened 2021-04-06
Report package DB verification metadata


wagoodman:
Report on RPM, DPKG, and APK file manifest data for each file owned by each package found in the respective database:
- digest
- digest algorithm
- mode
- user
- group
- size
- conffile listing (if available)

wagoodman:
Note for DPKG: anchore-engine today only captures the package name, digest, digest algorithm, and indication if this is a conf file per https://github.com/anchore/anchore-engine/blob/v0.8.2/anchore_engine/analyzers/utils.py#L1863

_This means that the following fields will be left out_: mode, user, group, file size

Actions needed for DPKG: 
- pull out the `conffile` listing, combining information from both the status file and the info/conffile

wagoodman:
### Update for APK
anchore-engine today only captures package name, digest, digest algorithm, mode, group, and user per https://github.com/anchore/anchore-engine/blob/v0.8.2/anchore_engine/analyzers/utils.py#L1975

_This means that the following fields will be left out of the syft implementation_: conffile, size

Note: today we have a `Checksum` field. We could make this agnostic by providing a `digest` field with `value` and `algorithm` fields. APK only uses sha1, but it could help consumers by being consistent here.

wagoodman:
### Update for RPM

anchore-engine today captures: digest, digest algorithm, mode, group, user, size, and indication if this is a conffile
See: https://github.com/anchore/anchore-engine/blob/v0.8.2/anchore_engine/analyzers/utils.py#L1756

Actions needed for RPM:
- add user
- add group
- add conffile indication
- add digest algorithm

Note: today we have a`SHA256` field. We could make this agnostic by providing a `digest` field with `value` and `algorithm` fields. This way we could help consumers by being consistent.

PR #375, opened 2021-04-09
Update package verification fields for DPKG

wagoodman:
This PR implements the suggested improvements from https://github.com/anchore/syft/issues/371#issuecomment-815932176, specifically:
- add `conffile` listing to the `files` field, combining information from both the status file and the info/conffile
- normalize the `md5` field to `digest`, with subfields `value` and `algorithm`
- updated JSON schema with `DpkgMetadata` changes (which is a partially breaking change)
- discovered that the `artifacts.metadata` `anyOf` clause was being generated incorrectly, resulting in an always-passing validation --this was fixed as well.

Partially addresses #371


syft/pkg/cataloger/deb/cataloger.go
```diff
@@ -56,93 +59,166 @@ func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, error)
...
-					p.Locations = append(p.Locations, *copyrightLocation)
-				}
+	// get file listing (package files + additional config files)
+	files, infoLocations := getAdditionalFileListing(resolver, dbLocation, p)
+loopNewFiles:
+	for _, newFile := range files {
+		for _, existingFile := range metadata.Files {
+			if existingFile.Path == newFile.Path {
+				// skip adding this file since it already exists
+				continue loopNewFiles
```
alfredodeza, line 101: This looks like it isn't doing anything... I don't see what is the condition where it would do something else when the file doesn't already exist

syft/pkg/cataloger/deb/cataloger.go
```diff
@@ -56,93 +59,166 @@ func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, error)
...
+	// persist alterations
+	p.Metadata = metadata
+
+	// persist location information from each new source of information
+	p.Locations = append(p.Locations, infoLocations...)
+}
+
+func getAdditionalFileListing(resolver source.FileResolver, dbLocation source.Location, p *pkg.Package) ([]pkg.DpkgFileRecord, []source.Location) {
+	// ensure the file list is an empty collection (not nil)
+	var files = make([]pkg.DpkgFileRecord, 0)
```
alfredodeza, line 121: I think we've gone back-and-forth on this before, but what is the benefit of creating a slice of `0` instead of (I know the linter would complain):

```suggestion
	var files []pkg.DpkgFileRecord
```

syft/pkg/cataloger/deb/cataloger.go
```diff
@@ -56,93 +59,166 @@ func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, error)
...
-	return results, nil
+
+	conffilesReader, conffilesLocation := fetchConffileContents(resolver, dbLocation, p)
+
+	if conffilesReader != nil {
+		// attach the file list
+		files = append(files, parseDpkgConffileInfo(md5Reader)...)
+
+		// keep a record of the file where this was discovered
+		if md5Location != nil {
```
alfredodeza, line 143: shouldn't this be `conffilesReader != nil`?

If it should be changed, it seems that this is missing a test case

syft/pkg/cataloger/deb/cataloger.go
```diff
@@ -56,93 +59,166 @@ func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, error)
...
 
-	reader, err := resolver.FileContentsByLocation(*md5SumLocation)
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to fetch deb md5 contents (%+v): %w", p, err)
+	return md5Reader, location
+}
+
+func fetchConffileContents(resolver source.FileResolver, dbLocation source.Location, p *pkg.Package) (io.ReadCloser, *source.Location) {
+	var reader io.ReadCloser
+	var err error
```
alfredodeza, line 180: why is this `var` needed?

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -48,20 +51,54 @@ func parseDpkgStatus(reader io.Reader) ([]pkg.Package, error) {
...
+	err = mapstructure.Decode(dpkgFields, &entry)
+	if err != nil {
+		return pkg.DpkgMetadata{}, err
+	}
+
+	name, version := extractSourceVersion(entry.Source)
+	if version != "" {
+		entry.SourceVersion = version
+		entry.Source = name
+	}
```
alfredodeza, line 74: for my clarification only, this means that if version is empty, we don't want to capture the name?

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -48,20 +51,54 @@ func parseDpkgStatus(reader io.Reader) ([]pkg.Package, error) {
...
+	// there may be an optional conffiles section that we should persist as files
+	if conffilesSection, exists := dpkgFields["Conffiles"]; exists && conffilesSection != nil {
+		if sectionStr, ok := conffilesSection.(string); ok {
+			entry.Files = parseDpkgConffileInfo(strings.NewReader(sectionStr))
+		}
+	}
+
+	if entry.Files == nil {
+		// ensure that the collection is always allocated
+		entry.Files = make([]pkg.DpkgFileRecord, 0)
```
alfredodeza, line 85: This is surprising... isn't there a way for this to be defined beforehand instead of having to check like this?

alfredodeza: Ohhh I guess not because the struct wasn't initialized. Shouldn't there be a constructor for this?

alfredodeza (changes requested):

syft/pkg/cataloger/deb/cataloger.go
wagoodman: The key to this logic is the argument to `continue`.

There are two loops here; the outer one is a named loop (`loopNewFiles`) looping through the new files and an inner loop that is checking the existing files. The last line in the outer loop (`loopNewFiles`) will append the new file. However, the `continue loopNewFiles` statement occurs when the new path matches an existing path. The optional argument to `continue` here is key, as it is noting to continue the outer loop iteration and not the inner loop iteration.

syft/pkg/cataloger/deb/cataloger.go
wagoodman: I can update the comment to be more specific on the reasoning here...

The reason for the allocation is for later presentation. That is, the metadata objects to not have redundant definitions, and the common team consensus on displaying collections in the JSON presenter is to always show empty collections (e.g. `{}`) instead of `nil`. This requires that objects with collections that will be presented have allocations in advance. The comment was meant to portray this, but only does so partially --I'll adjust.

syft/pkg/cataloger/deb/cataloger.go
wagoodman: nice catch 👍 there happens to not be an example where there are conffiles and not md5sums (which I don't think would be seen on a system, the inverse is true... where you will find md5sums but not conffiles)

syft/pkg/cataloger/deb/cataloger.go
wagoodman: we are assigning to an existing `md5Reader` variable with the line:
```golang
    md5Reader, err = resolver.FileContentsByLocation(*location)
```
since this is done conditionally, we need an existing variable in the outer scope. Since an error is also returned, we need a variable for that too in the outer scope.

syft/pkg/cataloger/deb/parse_dpkg_status.go
wagoodman: good question --I only migrated this part of the code (already existed in this file). Looks like it was added in https://github.com/anchore/syft/pull/297 ... maybe an oversight?

syft/pkg/cataloger/deb/cataloger.go
alfredodeza: oh wow - this is something I'd not seen before, thanks for explaining it

syft/pkg/cataloger/deb/parse_dpkg_status.go
wagoodman: Right, we don't have a constructor here (or in many [any?] of the underlying pkg.*Metadata structs). We typically have a `parse*()` function that is acting like the constructor for an object. I'll move this line to where `entry` is created to make this more straight forward 👍

syft/pkg/cataloger/deb/parse_dpkg_status.go
alfredodeza: I didn't catch that this was migrated over. No strong opinions here, just that this looked odd. If you think this shouldn't get a bit reworked that's fine. If in doubt though, I'd recommend opening up a ticket to look at this later on (no need to block this PR on this)

alfredodeza (approved):
This is also good to go, I think that the dpkg source+version situation needs at least an issue.

wagoodman:
@alfredodeza created https://github.com/anchore/syft/issues/377

PR #378, opened 2021-04-13
Add ability to pull images directly from a registry

wagoodman:
Adds the ability to pull images directly from a registry, no container runtime is required:
```
syft registry:reop/myimage:latast
```

Authentication and connection options are configurable under the new `registry` configuration section. A single set of credentials may be provided via environment variables, multiple credentials must be provided via a config file. See the README for the details.


cmd/packages.go
```diff
@@ -32,7 +32,7 @@ const (
   {{.appName}} {{.command}} alpine:latest -vv            show verbose debug information
 
   Supports the following image sources:
-    {{.appName}} {{.command}} yourrepo/yourimage:tag     defaults to using images from a Docker daemon
+    {{.appName}} {{.command}} yourrepo/yourimage:tag     defaults to using images from a Docker daemon. If docker is not present, the image is pulled directly from the registry.
```
luhring, line 35: nit: Would be good to be consistent on capitalizing "Docker"

internal/config/registry.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (cfg registry) loadDefaultValues(v *viper.Viper) {
+	v.SetDefault("registry.insecure-skip-tls-verify", false)
+	v.SetDefault("registry.auth", []RegistryCredentials{})
+}
+
+func (cfg *registry) parseConfigValues() error {
+	// there may be additional credentials provided by env var that should be appended to the set of credentials
+	// note: only a authority and password would be required since a token
```
alfredodeza, line 33: seems like this comment got cut

internal/config/registry.go
```diff
@@ -0,0 +1,68 @@
...
+	authority, username, password, token :=
+		os.Getenv("SYFT_REGISTRY_AUTH_AUTHORITY"),
+		os.Getenv("SYFT_REGISTRY_AUTH_USERNAME"),
+		os.Getenv("SYFT_REGISTRY_AUTH_PASSWORD"),
+		os.Getenv("SYFT_REGISTRY_AUTH_TOKEN")
+
+	if authority != "" && password != "" {
```
alfredodeza, line 40: if I'm configuring this with a token only then this condition wouldn't happen, is that an oversight? or when would a token be prepended?

syft/source/source.go
```diff
@@ -25,7 +23,7 @@ type Source struct {
 type sourceDetector func(string) (image.Source, string, error)
 
 // New produces a Source based on userInput like dir: or image:tag
-func New(userInput string) (Source, func(), error) {
+func New(userInput string, registryOptions *image.RegistryOptions) (Source, func(), error) {
```
luhring, line 26: In the future, it'd be nice to be able to abstract this, such that other consumers (such as in the changes below) don't need to worry about answering for this parameter if they don't care about registry concerns. I have no useful ideas right now though!

luhring (approved):

syft/source/source.go
wagoodman: I agree, and also came up blank here (without resorting to injecting not through the constructor or odd `init` voodoo)

internal/config/registry.go
wagoodman: excellent call, adjusting now. I'm going to see if I can get a cli-level test to cover this

internal/config/registry.go
wagoodman: yup yup, I'll adjust here

wagoodman: conclusion: comment was not needed, the code has been adjusted to allow for token

internal/config/registry.go
wagoodman: cli test added for asserting registry options in general

test/cli/packages_cmd_test.go
```diff
@@ -137,3 +137,75 @@ func TestPackagesCmdFlags(t *testing.T) {
...
+				assertInOutput(`no registry credentials configured, using the default keychain`),
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			cmd, stdout, stderr := runSyftCommand(t, test.env, test.args...)
+			for _, traitFn := range test.assertions {
+				traitFn(t, stdout, stderr, cmd.ProcessState.ExitCode())
```
luhring, line 202: nit: It might be more clear when reading line 202 if the individual slice item was named something like `assertFn` (or something that includes "assert") — just since it's a recognizable/idiomatic part of a test body

luhring (approved):

PR #384, opened 2021-04-14
Modify registry credential verification

### LINKED ISSUE #385: Registry credentials should require username and password
Today it is possible for the config to pass partial credentials to stereoscope for registry authentication. We should ensure that both username and password OR token is required before appending credentials from environment variables onto the config.

wagoodman:
Enforces that user and password OR a token must be present --no partial values allowed now. Sister PR to https://github.com/anchore/grype/pull/293

Closes #385

### DIFF
--- a/internal/config/registry_test.go
+++ b/internal/config/registry_test.go
@@ -1,0 +1,59 @@
+package config
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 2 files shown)


internal/config/registry_test.go
```diff
@@ -0,0 +1,59 @@
+package config
```
luhring, line 1: This is one of the most elegantly written/formatted test files I've ever seen

luhring (approved):

PR #387, opened 2021-04-15
Expand RPM DB fields

wagoodman:
Makes the following adjustments to the RPM DB cataloger:
- adds username
- adds groupname
- adds file flags (`c` indicates conffile)
- add digest algorithm
- replaces the SHA256 field with the shared definition for `file.Digest`

Partially addresses #371


alfredodeza (approved):
vhpy

PR #389, opened 2021-04-15
Support msrc types

### LINKED ISSUE #381: msrc package type
**What would you like to be added**: Package Type support for MSRC

**Additional context**:
Probably just the types so that they can be filled in by injected sbom. No direct support for scanning/parsing to provide MSRC package type (which is basically KBs in reality)

alfredodeza:
Closes #381

### DIFF
--- a/syft/distro/identify_test.go
+++ b/syft/distro/identify_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/anchore/syft/internal"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/anchore/syft/syft/source"
 )
@@ -88,9 +89,14 @@ func TestIdentifyDistro(t *testing.T) {
 
 	observedDistros := internal.NewStringSet()
 	definedDistros := internal.NewStringSet()
+
 	for _, distroType := range All {
 		definedDistros.Add(string(distroType))
 	}
+
+	// Somewhat cheating with Windows. There is no support for detecting/parsing a Windows OS, so it is not
+	// possible to comply with this test unless it is added manually to the "observed distros"
+	definedDistros.Remove(string(Windows))
 
 	for _, test := range tests {
 		t.Run(test.fixture, func(t *testing.T) {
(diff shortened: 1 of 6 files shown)


syft/distro/identify_test.go
```diff
@@ -88,6 +89,11 @@ func TestIdentifyDistro(t *testing.T) {
 
 	observedDistros := internal.NewStringSet()
 	definedDistros := internal.NewStringSet()
+
+	// Somewhat cheating with Windows. There is no support for detecting/parsing a Windows OS, so it is not
+	// possible to comply with this test unless it is added manually to the "observed distros"
+	observedDistros.Add(string(Windows))
```
wagoodman, line 95: agreed on not requiring testing of windows, however, I feel that removing the windows distro type from the `definedDistros` variable is a bit better than saying we observed windows during testing.

wagoodman (changes requested):
I see that there is a need to capture KB metadata, but does that need to be done within `pkg`? That is, adding a new package type and package metadata seems to imply that this is package metadata, when this is really more akin to distro metadata.

syft/distro/identify_test.go
alfredodeza: I see those as the same thing but I'm fine swapping it around

alfredodeza:
> I see that there is a need to capture KB metadata, but does that need to be done within `pkg`? That is, adding a new package type and package metadata seems to imply that this is related to a package, when this is really more akin to distro metadata.

The "packages" are really distro information. There aren't any packages here, I'm overloading the package to represent each KB installed in a Windows OS. The metadata is not (in my testing) used, but it represents better what the fields should look like, and opens up the ability to add more metadata later.

syft/distro/identify_test.go
alfredodeza: This got swapped around

wagoodman (commented):
Thanks for the offline conversation regarding this approach and why it plays nicely into the existing grype patterns. Though it would be nice to capture this under `distro` as auxiliary metadata that can get picked up by a custom grype matcher, currently the matchers are oriented around matching packages... and even though they are passed distro information, we would need to add an alternate matching pattern that could take extra time. 

I think you're right --we can get away with not doing that work for now until we find other examples where a similar matching pattern would be needed, then take the effort to refactor. 

(you can ignore my comment here https://github.com/anchore/syft/pull/389#pullrequestreview-637661558)

wagoodman (approved):

PR #392, opened 2021-04-19
Refactor pom properties handling relative to parent package

wagoodman:
Today the java packages processed are not in a guaranteed order. Additionally, java packages alter parent package metadata based on information discovered in other child package processing paths (specifically in pom properties processing). This can be problematic given the first note (order not guaranteed) and if you find a jar with multiple pom properties.

This PR refactors the pom processing path to only alter the parent if there is missing information, otherwise no change to the parent is made.

### DIFF
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -98,7 +98,7 @@ func (j *archiveParser) parse() ([]pkg.Package, error) {
 	}
 
 	// find aux packages from pom.properties
-	auxPkgs, err := j.discoverPkgsFromPomProperties(parentPkg)
+	auxPkgs, err := j.discoverPkgsFromAllPomProperties(parentPkg)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -184,69 +181,82 @@ func (j *archiveParser) discoverPkgsFromPomProperties(parentPkg *pkg.Package) ([
 			continue
 		}
 
-		if propsObj.Version != "" && propsObj.ArtifactID != "" {
+		if propsObj.Version == "" || propsObj.ArtifactID == "" {
```
luhring, line 184: ❤️ Love this inversion and the early return. Way easier to follow! ❤️

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -184,69 +181,82 @@ func (j *archiveParser) discoverPkgsFromPomProperties(parentPkg *pkg.Package) ([
...
-				Language:     pkg.Java,
-				Type:         pkg.JavaPkg,
-				MetadataType: pkg.JavaMetadataType,
-				Metadata: pkg.JavaMetadata{
-					VirtualPath:   virtualPath,
-					PomProperties: propsObj,
-					Parent:        parentPkg,
-				},
-			}
+		if parentPkg == nil {
```
luhring, line 189: ❤️ Love this inversion and the early return. Way easier to follow! ❤️

luhring (approved):
This is a huge net improvement, and helps build momentum for anything additional we want to look at down the road. Thanks!!!

PR #397, opened 2021-04-19
Add hyphen replacement logic for CPE generation

### LINKED ISSUE #361: Add hyphen replacement in CPE generator
Today the CPE generator in anchore-engine [substitutes `-` for `_` within product names](https://github.com/anchore/anchore-engine/blob/master/anchore_engine/services/policy_engine/engine/loaders.py#L1132) for a candidate CPE for java packages --Syft should make the same replacement when [generating candidate names](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/cpe.go#L109).

wagoodman:
Adds hyphen replacement for product names as described in #361 (for all package types, not just java packages).

Closes #361


syft/pkg/cataloger/cpe.go
```diff
@@ -80,6 +80,11 @@ func candidateVendors(p pkg.Package) []string {
 	switch p.Language {
 	case pkg.Python:
 		vendors = append(vendors, fmt.Sprintf("python-%s", p.Name))
+		if strings.Contains(p.Name, "-") {
```
dakaneye, line 83: Why not just always do the replacement? Seems like negligible performance cost, but maybe something I'm missing too...my thinking is it'd just simplify this block
```
vendors = append(vendors.fmt.Sprintf("python_%s", strings.ReplaceAll(p.Name, "-", "_")))
```

dakaneye: Also are there some language distributions that need a hyphen? Or is it inconsistent across languages in NVD?

syft/pkg/cataloger/cpe.go
```diff
@@ -98,8 +103,7 @@ func candidateVendors(p pkg.Package) []string {
 
 func candidateProducts(p pkg.Package) []string {
```
zhill, line 104: The python-|_ should be an option in the product name as well

zhill (commented):
Just some questions on why approaches used in one area and not another. I queried the CPE dict and found python- prefixes in both vendor (project names) and product names.

syft/pkg/cataloger/cpe.go
wagoodman: actually, looks like adjusting for [this comment](https://github.com/anchore/syft/pull/397#discussion_r616235716) will handle this case since vendors are duplicate of the product candidates.

syft/pkg/cataloger/cpe.go
wagoodman: righto, adjusted in the [latest commit](https://github.com/anchore/syft/pull/397/commits/924e9e6e6ad268cfe72dd77de250c16e7c68d0a6)

luhring (approved):

PR #399, opened 2021-04-19
Don't use pom groupId for Jenkins and Jira plugins

### LINKED ISSUE #395: Using pom groupId leading to bad CPEs for plugins, leading to false positives
This is specific to Java packages...

E.g.

1. Plugins with groupId: `com.atlassian.jira.plugins` — matches **_Jira_** itself
2. Plugins with groupId: `com.cloudbees.jenkins.plugins` — matches **_Jenkins_** itself

luhring:
Closes #395

### DIFF
--- a/syft/pkg/cataloger/cpe.go
+++ b/syft/pkg/cataloger/cpe.go
@@ -54,6 +54,9 @@ func (s candidateStore) getCandidates(t pkg.Type, key string) []string {
 	return value
 }
 
+const pomPropertiesGroupIDJenkinsPlugins = "com.cloudbees.jenkins.plugins"
+const pomPropertiesGroupIDJiraPlugins = "com.atlassian.jira.plugins"
+
 func newCPE(product, vendor, version, targetSW string) wfn.Attributes {
 	cpe := *(wfn.NewAttributesWithAny())
 	cpe.Part = "a"
@@ -100,7 +103,7 @@ func candidateTargetSoftwareAttrs(p pkg.Package) []string {
 	var targetSw []string
 	switch p.Language {
 	case pkg.Java:
-		targetSw = append(targetSw, "java", "maven")
+		targetSw = append(targetSw, candidateTargetSoftwareAttrsForJava(p)...)
 	case pkg.JavaScript:
 		targetSw = append(targetSw, "node.js", "nodejs")
 	case pkg.Ruby:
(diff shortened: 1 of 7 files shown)


syft/pkg/cataloger/cpe.go
```diff
@@ -82,37 +83,115 @@ func candidateVendors(p pkg.Package) []string {
...
-	default:
-		return products
 	}
 	return products
 }
+
+func candidateVendorsForJava(p pkg.Package) []string {
+	var vendors []string
+	if p.MetadataType == pkg.JavaMetadataType {
+		if product := vendorFromPomPropertiesGroupID(p); product != "" {
```
alfredodeza, line 113: this is not product right? I think it is vendor? 
```suggestion
		if vendor := vendorFromPomPropertiesGroupID(p); vendor != "" {
```

luhring: Yep, correct. I'm working on a revision to this branch that cleans some of this up and adds test, will push shortly

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -230,44 +225,42 @@ func (j *archiveParser) packagesFromPomProperties(propsObj *pkg.PomProperties, p
...
+
+	return nil
 }
 
 // discoverPkgsFromNestedArchives finds Java archives within Java archives, returning all listed Java packages found and
 // associating each discovered package to the given parent package.
 func (j *archiveParser) discoverPkgsFromNestedArchives(parentPkg *pkg.Package) ([]pkg.Package, error) {
 	var pkgs = make([]pkg.Package, 0)
 
-	if !j.detectNested {
```
wagoodman, line 267: nit question: why move this from here? the original code makes it more difficult to misuse this function relative to the `archiveParser` configuration.

wagoodman (approved):

syft/pkg/cataloger/java/archive_parser.go
luhring: Good question. The value of `j.detectNested` is always `true` in the application code. Because of that, I believe we could consider removing it in the future, to reduce the number of variables needing to be consulted in the logic.

In the test code, it looks like the value of `j.detectNested` is always `false`. I think this warrants some discussion. As it reads now, it looks like that parameter exists to facilitate testing, although it's never used with the same value in testing (`false`) as it is in application code (`true`). If it's configurable in the application code, we should consider testing all states in the test code. And, if it's not needed in the application code, we should consider whether or not it's needed in the test code. I was wondering if it was used in testing out of convenience, so we can reuse nested archives for tests that only need to expect a single package. I could probably benefit from learning more of the context here.

In the meantime, I wanted to surface this check so the conditionality was more obvious to the caller, and so that we avoid the call to `discoverPkgsFromNestedArchives` if there's not a need to invoke it. (I'll also admit, **_subjectively_**, this was reading better to me: "Do we need to get nested archives? Yes? Okay, then let's do it." vs. "Get me the nested archives. (Okay, but we might not actually.)" — again, subjective 😃 )

For this particular PR, I can move the check back if that's more palatable, and we can follow up with a discussion after the PR is merged.

PR #402, opened 2021-04-21
Enhance CPE generation for java GroupId and filtering

wagoodman:
This adds the following features to the CPE generation logic:
- Adds `io.jenkins.plugins`, `io.jenkins-ci.plugins`, `org.jenkins.plugins`, and `org.jenkins-ci.plugins` as known jenkins plugins GroupIDs to identify a `JenkinsPlugin` package type.
- Enhances the GroupID logic to search for value prefixes instead of exact match (this covers cases such as `io.jenkins.plugin.bluesteel`)
- Adds ability to filter CPEs from the final generated list based on a filter function (which gets the CPE in question and the package it should describe). In this way there is a sanity check for known bad generated values.

### DIFF
--- a/internal/string_helpers.go
+++ b/internal/string_helpers.go
@@ -2,6 +2,7 @@ package internal
 
 import "strings"
 
+// HasAnyOfPrefixes returns an indication if the given string has any of the given prefixes.
 func HasAnyOfPrefixes(input string, prefixes ...string) bool {
 	for _, prefix := range prefixes {
 		if strings.HasPrefix(input, prefix) {
--- a/internal/string_helpers.go
+++ b/internal/string_helpers.go
@@ -1,0 +1,13 @@
+package internal
+
+import "strings"
+
+func HasAnyOfPrefixes(input string, prefixes ...string) bool {
+	for _, prefix := range prefixes {
+		if strings.HasPrefix(input, prefix) {
+			return true
+		}
+	}
+
+	return false
+}
--- a/syft/pkg/java_metadata.go
+++ b/syft/pkg/java_metadata.go
@@ -1,6 +1,19 @@
 package pkg
 
-import "github.com/package-url/packageurl-go"
+import (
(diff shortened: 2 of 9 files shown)


syft/pkg/java_metadata.go
```diff
@@ -1,6 +1,19 @@
...
 
-import "github.com/package-url/packageurl-go"
+import (
+	"github.com/anchore/syft/internal"
+	"github.com/package-url/packageurl-go"
+)
+
+const PomPropertiesGroupIDJiraPlugins = "com.atlassian.jira.plugins"
+
+var JenkinsPluginPomPropertiesGroupIDs = []string{
```
luhring, line 10: nit: Would be nice to name these two related things with a consistent phrase pattern (I'm good with either of the existing patterns winning): `PomPropertiesGroupIDJiraPlugins` vs. `JenkinsPluginPomPropertiesGroupIDs`

internal/string_helpers.go
```diff
@@ -0,0 +1,13 @@
+package internal
+
+import "strings"
+
+func HasAnyOfPrefixes(input string, prefixes ...string) bool {
```
luhring, line 5: Good call moving this here 👍

luhring: nit: doc comment

luhring (approved):

syft/pkg/java_metadata.go
wagoodman: I tried on `PomPropertiesGroupIDsJenkinsPlugins`  and `PomPropertiesGroupIDsJenkinsPlugin` for size, it seemed odd with the order and plurality. I can swap out `PomPropertiesGroupIDJiraPlugins` for `JiraPluginPomPropertiesGroupID`.

internal/string_helpers.go
wagoodman: it was good where it was at, but then I made use of it in another package. I'll update the doc string

PR #403, opened 2021-04-22
Update parent pom persistence with regard to shaded jars

wagoodman:
It was discovered that we're potentially missing java dependencies for jars that have multiple pom properties. This seems to be the case especially for shaded jars, which may have more than one pom properties (one for each packaged dependency). This PR attempts to close the gap on correctness here, only updating the parent package if there is a positive match of the parent package identity and the discovered identity from the pom properties. 

The main change made was related to how we match a parent and child package; The artifactID and parent package name must match, they cannot be subsets of one another. This may regress other cases, however, the previous code was too presumptive (see the comments). 

This came from an investigation of looking at the `cloudbees-mm-core` image and why it was missing the `cloudbees-analytics-segment-driver` nested jar in the SBOM, which is a shaded jar.

The code was further decomposed as well (the tests were updated, but not decomposed further).

### DIFF
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -184,14 +184,17 @@ func (j *archiveParser) discoverPkgsFromAllPomProperties(parentPkg *pkg.Package)
 			continue
 		}
 
-		pkgs = append(pkgs, j.packagesFromPomProperties(*pomProperties, parentPkg)...)
+		pkgFromPom := j.newPackageFromPomProperties(*pomProperties, parentPkg)
+		if pkgFromPom != nil {
+			pkgs = append(pkgs, *pkgFromPom)
+		}
 	}
 	return pkgs, nil
 }
 
 // packagesFromPomProperties processes a single Maven POM properties for a given parent package, returning all listed Java packages found and
 // associating each discovered package to the given parent package.
-func (j *archiveParser) packagesFromPomProperties(pomProperties pkg.PomProperties, parentPkg *pkg.Package) []pkg.Package {
+func (j *archiveParser) newPackageFromPomProperties(pomProperties pkg.PomProperties, parentPkg *pkg.Package) *pkg.Package {
 	// keep the artifact name within the virtual path if this package does not match the parent package
 	vPathSuffix := ""
 	if !strings.HasPrefix(pomProperties.ArtifactID, parentPkg.Name) {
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -297,3 +267,53 @@ func (j *archiveParser) discoverPkgsFromNestedArchives(parentPkg *pkg.Package) (
 
 	return pkgs, nil
 }
+
+func packageIdentitiesMatch(p pkg.Package, parentPkg *pkg.Package) bool {
```
luhring, line 271: This function looks intricate enough to warrant some unit tests

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -297,3 +267,53 @@ func (j *archiveParser) discoverPkgsFromNestedArchives(parentPkg *pkg.Package) (
...
+	pkgKey := uniquePkgKey(&p)
+	parentKey := uniquePkgKey(parentPkg)
+
+	// the name/version pair matches...
+	matchesParentPkg := pkgKey == parentKey
```
luhring, line 276: I'd suggest returning early when possible to simplify reading. For example, if `pkgKey == parentKey` is `true`, nothing else in the function body can change the function's return value. It would be helpful to the reader to express that via an early return.

syft/pkg/cataloger/java/archive_parser.go
luhring: (And, it's awesome that this was broken out!! 👍 👍 )

luhring (changes requested):

syft/pkg/cataloger/java/archive_parser.go
wagoodman: I thought about it but decided not to for now

> The code was further decomposed as well (the tests were updated, but not decomposed further).

There are tests that exercise this function.

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -297,3 +267,53 @@ func (j *archiveParser) discoverPkgsFromNestedArchives(parentPkg *pkg.Package) (
...
+
+	// the virtual path matches...
+	matchesParentPkg = matchesParentPkg || parentPkg.Metadata.(pkg.JavaMetadata).VirtualPath == p.Metadata.(pkg.JavaMetadata).VirtualPath
+
+	metadata, ok := p.Metadata.(pkg.JavaMetadata)
```
luhring, line 281: Question: (just noticed this) It looks like `p.Metadata` is used as a `pkg.JavaMetadata` (via an assertion) on line 279, but then that type assertion is verified afterward on this line. Is there a reason to not check the type first?

wagoodman: This is left over from moving from the last function (where the metadata was a already available)... I added a safe type assertion out of habit, but it technically is not needed here. I'll adjust this to be more consistent.

syft/pkg/cataloger/java/archive_parser.go
luhring: Okay, I think I see where you're coming from. The only code path I saw that didn't appear to be tested by the consuming function's test was the `if !ok` path, which is gone now, so that's a moot concern. It looks like this function is getting tested in an indirect way via existing tests, so that's good. My preference would be that we add localized tests for this function in the future, but I'm up for deferring that beyond this PR. Thanks for the consideration.

syft/pkg/cataloger/java/archive_parser.go
luhring: My understanding of the code is that, in looking at this function in isolation, a panic is still possible, but that in the context of the rest of the current codebase, there would never be a situation where `p.Metadata` is set to a `nil` value or non-`pkg.JavaMetadata` type, so a panic isn't possible currently. And, it looks like such a panic would be caught in the consuming function's test, so that's good. 👍

syft/pkg/cataloger/java/archive_parser.go
luhring: Thanks for this change, it helps a lot!

syft/pkg/cataloger/java/archive_parser.go
wagoodman: agreed on adding localized tests in the future 👍

PR #405, opened 2021-04-22
Add for known bad CPE field combinations for jenkins package

wagoodman:
This adds a filter to eliminate guesses that would result in matching jenkins server without having a package name that minimally reflects this.

Additionally updates the jira filter to consider `wfn.Any` vendor values.


syft/pkg/cataloger/cpe.go
```diff
@@ -37,14 +37,22 @@ var productCandidatesByPkgType = candidateStore{
 }
 
 var cpeFilters = []filterFn{
-	// nolint: goconst
 	func(cpe pkg.CPE, p pkg.Package) bool {
 		// jira / atlassian should not apply to clients
-		if cpe.Vendor == "atlassian" && cpe.Product == "jira" && strings.Contains(p.Name, "client") {
-			return true
+		if cpe.Product == "jira" && strings.Contains(strings.ToLower(p.Name), "client") {
```
luhring, line 42: nit: Now that we're amassing some nontrivial, specialized rules in our list of `filterFn`, it would be nice to codify these as named functions that get referenced here in this list. But I'm making this a "nit" due to the context of our CPE implementation future

luhring (approved):

syft/pkg/cataloger/cpe.go
wagoodman: agreed with the long term approach, and held off for similar considerations 👍

PR #410, opened 2021-05-06
update readme to point to grype

alfredodeza:
Signed-off-by: Alfredo Deza <adeza@anchore.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![License: Apache-2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/anchore/syft/blob/main/LICENSE)
 [![GitHub go.mod Go version](https://img.shields.io/github/go-mod/go-version/anchore/syft.svg)](https://github.com/anchore/syft)
 
-A CLI tool and go library for generating a Software Bill of Materials (SBOM) from container images and filesystems.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,7 @@
 [![License: Apache-2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/anchore/syft/blob/main/LICENSE)
 [![GitHub go.mod Go version](https://img.shields.io/github/go-mod/go-version/anchore/syft.svg)](https://github.com/anchore/syft)
 
-A CLI tool and go library for generating a Software Bill of Materials (SBOM) from container images and filesystems.
+A CLI tool and go library for generating a Software Bill of Materials (SBOM) from container images and filesystems, to help detect vulnerabilities when used with a vulnerability matcher like [Grype](https://github.com/anchore/grype)
```
luhring, line 9: This is a good direction. I think we should be clear that Syft and its output are valuable in their own right, too. What about something like this?

>A CLI tool and Go library for generating a Software Bill of Materials (SBOM) from container images and filesystems. Great for enabling vulnerability detection using tools like [Grype](https://github.com/anchore/grype).

luhring (changes requested):

README.md
alfredodeza: Before I commit/push, how about this variation:

```
Exceptional for vulnerability detection when used with a scanner tool like [Grype](https://github.com/anchore/grype)
```

luhring: @alfredodeza I think that's good enough 👍

alfredodeza: updated

PR #427, opened 2021-06-02
Allow for RPM package epoch to be optionally provided in the version field

### LINKED ISSUE #408: Report Epoch in RPM version
Today we don't report epoch in the RPM version: https://github.com/anchore/syft/blob/v0.15.1/syft/pkg/cataloger/rpmdb/parse_rpmdb.go#L52

We should report the epoch in the version string when it is available (see the code comment above).

wagoodman:
This PR makes a few changes to how RPM versions are displayed:
- Epoch is optionally included in the `pkg.Package.Version` field, only if the epoch version is explicitly set for that package in the RPM DB (0 is not assumed)
- Release field is included in the `pkg.Package.Version` field
- for pURLs representing RPMs, the epoch is specified as a qualifier and not as part of the version

~Note: this will remain in draft until all changes to the grype matching are identified and drafted.~ (confirmed intended behavior)

Closes #408

### DIFF
--- a/syft/pkg/cataloger/package_url_test.go
+++ b/syft/pkg/cataloger/package_url_test.go
@@ -34,30 +34,6 @@ func TestPackageURL(t *testing.T) {
 			pkg: pkg.Package{
 				Name:    "name",
 				Version: "v0.1.0",
-				Type:    pkg.PythonPkg,
-			},
-			expected: "pkg:pypi/name@v0.1.0",
-		},
-		{
-			pkg: pkg.Package{
-				Name:    "name",
-				Version: "v0.1.0",
-				Type:    pkg.PythonPkg,
-			},
-			expected: "pkg:pypi/name@v0.1.0",
-		},
-		{
-			pkg: pkg.Package{
-				Name:    "name",
-				Version: "v0.1.0",
-				Type:    pkg.PythonPkg,
-			},
-			expected: "pkg:pypi/name@v0.1.0",
-		},
-		{
-			pkg: pkg.Package{
-				Name:    "name",
-				Version: "v0.1.0",
 				Type:    pkg.GemPkg,
 			},
 			expected: "pkg:gem/name@v0.1.0",
@@ -96,13 +72,31 @@ func TestPackageURL(t *testing.T) {
 				Type:    pkg.RpmPkg,
 				Metadata: pkg.RpmdbMetadata{
 					Name:    "name",
-					Version: "v0.1.0",
-					Epoch:   2,
+					Version: "0.1.0",
+					Epoch:   intRef(2),
 					Arch:    "amd64",
 					Release: "3",
 				},
 			},
-			expected: "pkg:rpm/centos/name@2:v0.1.0-3?arch=amd64",
+			expected: "pkg:rpm/centos/name@0.1.0-3?arch=amd64&epoch=2",
+		},
+		{
+			distro: &distro.Distro{
+				Type: distro.CentOS,
+			},
+			pkg: pkg.Package{
+				Name:    "bad-name",
+				Version: "bad-v0.1.0",
+				Type:    pkg.RpmPkg,
+				Metadata: pkg.RpmdbMetadata{
+					Name:    "name",
+					Version: "0.1.0",
+					Epoch:   intRef(),
+					Arch:    "amd64",
+					Release: "3",
+				},
+			},
+			expected: "pkg:rpm/centos/name@0.1.0-3?arch=amd64",
 		},
 		{
 			distro: &distro.Distro{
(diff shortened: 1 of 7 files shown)


syft/pkg/cataloger/package_url_test.go
```diff
@@ -136,3 +130,10 @@ func TestPackageURL(t *testing.T) {
 		})
 	}
 }
+
+func intRef(i ...int) *int {
```
luhring, line 134: This is clever. But I think it would be more straightforward to take only a single `int` and return a pointer to it, and in cases where the consumer just needs `nil`, to use `nil` directly, since seeing `Epoch: intRef()` is less immediately obvious than seeing `Epoch: nil`.

syft/pkg/cataloger/rpmdb/parse_rpmdb.go
```diff
@@ -74,6 +76,19 @@ func parseRpmDB(resolver source.FilePathResolver, dbLocation source.Location, re
 	return allPkgs, nil
 }
 
+// The RPM naming scheme is [name]-[version]-[release]-[arch], where version is implicitly expands to [epoch]:[version].
+// RPM version comparison depends on comparing at least the version and release fields together as a subset of the
+// naming scheme. This toElVersion function takes a RPM DB package information and converts it into a minimally comparable
+// version string, containing epoch (optional), version, and release information. Epoch is an optional field and can be
+// assumed to be 0 when not provided for comparison purposes, however, if the underlying RPM DB entry does not have
+// an epoch specified it would be slightly disingenuous to display a value of 0.
+func toElVersion(metadata pkg.RpmdbMetadata) string {
```
luhring, line 85: It's not clear to me what this function name (`toElVersion`) means... is this a domain-specific concept?

luhring (changes requested):

syft/pkg/cataloger/rpmdb/parse_rpmdb.go
wagoodman: The name derives from domain-specific concepts, yes.

RedHat has a "version naming scheme" of `name-version-release-arch` where the `version-release` is what is used to make comparisons within the RPM software --this is approach is shared with CentOS as well. You'll often see "rhel" which stands for RedHat Enterprise Linux or "el" which generically stands for Enterprise Linux and includes CentOS and similar flavors (basically means "*EL").

Seeing that this `version-release` pair is what is common to use as a package version for RHEL based distros, it made sense to name this function in a way that distinguishes it from just the name "toVersion", and "toVersionRelease" or "toVersionReleasePair" really didn't resonate / make sense. This is versioning scheme applies fairly universally within the EL ecosystem, so I ended up calling it "toElVersion" since this is the version value that can be compared within the EL ecosystem.

syft/pkg/cataloger/package_url_test.go
wagoodman: updated in https://github.com/anchore/syft/pull/427/commits/afbd8f8ea02746624e037801d288da66df1a1314

syft/pkg/cataloger/rpmdb/parse_rpmdb.go
luhring: Cool, this is helpful! Thanks!

I don't terribly mind the `VersionRelease` naming approach. Lately I've been working with Kubernetes code, and they have a struct that describes a resource's group, version, and kind (each as its own struct field). For the name of the struct, rather than invent a new term, they call it `GroupVersionKind`. I thought that was interesting... it was surprisingly straightforward.

But I see your point about the desire to harken back to the Enterprise Linux domain explicitly.

What about something like `toEnterpriseLinuxVersion`? It's longer than the acronym, but it's completely unambiguous.

A next step down in my mind would be `toELVersion`, to capitalize both letters of the acronym. This seems to be the convention in Go, to have all letters of the acronym uppercase if the first letter is uppercase. This acronym approach is less ideal though, from my perspective.

syft/pkg/cataloger/package_url_test.go
luhring: Awesome, looks good!

syft/pkg/cataloger/rpmdb/parse_rpmdb.go
alfredodeza: I don't have strong opinions here but I don't think there is an _"EL ecosystem"_. "RedHat Enterprise Linux" is one thing as a whole. The problem does bubble up when packages use/advertise the `"el"` as part of their convention, and those packages can be seen in CentOS, RHEL, and others. 
"EL" is domain specific for sure, and if I had to pick I'd prefer `toELVersion`

wagoodman: 👍  renaming to `toELVersion`

luhring (approved):

PR #428, opened 2021-06-03
Add support for processing files with prepended bytes before the zip archive

### LINKED ISSUE #413: Improve discovery of app.jar packaging (self-executing Jars)
Users of popular java frame work spring often package thier app as /app.jar.  This JAR file usually has a shell script prepended to them, ostensibly to enable them to configure their environment at runtime.  The script headers lead us to believe that it comes from the Spring framework / build tools, and we have little reason to believe that this is an uncommon way to distribute/deploy JAR files.

**Why is this needed**:
Steps to reproduce
syft packages [redacted]

Expected result
spring-core should appear in the output, as it's bundled in the /app.jar in the image

**Additional context**:

wagoodman:
Today syft unzips java jars and catalogs all contents, however, jars (which are typically zip files) with shell scripts prepended to the archive are [not supported by the stdlib](https://github.com/golang/go/issues/10464). This PR adds support for processing the files as described in the above golang issue + https://github.com/anchore/syft/issues/413#issuecomment-853293746 . 

> We could infer the archive length from the end of central directory record by reading from the end of the file, create a new io.ReaderAt via io.SectionReader that is bounded to only the archive based on the length from the ECOD, and create a new zip reader via zip.NewReader() (instead of zip.OpenReader). We'd need to make certain that the underlying reader passed to the SectionReader is closed after each use since there is no Close() operation available from a zip.Reader.

Specifically, this PR adds a new `ZipReadCloser` that is a drop-in replacement for the stdlib `zip.ReadCloser`, but adds support for the above mentioned case (ignoring prepended bytes).

This additionally updates the `TraverseFilesInZip` util function to only use the new `ZipReadCloser`, and in doing so all of the helper functions that use `TraverseFilesInZip` also no longer use `zip.ReadCloser`.

The Springboot test-fixture has been updated to generate the self-executing tar which exhibits this behavior (which should be the same kind of jar as described in #413 ).

Closes #413


internal/file/zip_file_helpers_test.go
```diff
@@ -59,73 +46,70 @@ func createZipArchive(t *testing.T, sourceDirPath, destinationArchivePath string
...
-	}
-
 	archivePrefix, err := ioutil.TempFile("", "syft-ziputil-archive-TEST-")
 	if err != nil {
-		return cleanup(cleanupFns), "", fmt.Errorf("unable to create tempfile: %+v", err)
+		t.Fatalf("unable to create tempfile: %+v", err)
 	}
-	cleanupFns = append(cleanupFns, func() error { return os.Remove(archivePrefix.Name()) })
+
+	t.Cleanup(
```
luhring, line 77: TIL about `t.Cleanup` 😄

internal/file/zip_file_manifest.go
```diff
@@ -1,23 +1,36 @@
...
+func NewZipFileManifest(archivePath string) (ZipFileManifest, error) {
+	zipReader, err := OpenZip(archivePath)
+	manifest := make(ZipFileManifest)
+	if err != nil {
+		return manifest, fmt.Errorf("unable to open zip archive (%s): %w", archivePath, err)
+	}
+	defer func() {
+		err = zipReader.Close()
+		if err != nil {
+			log.Errorf("unable to close zip archive (%s): %+v", archivePath, err)
```
luhring, line 26: Does this execution path necessarily result in the app terminating non-zero? I'm thinking back to our team's prior discussion about having a correlation between error-level log entries and app exit code. If this particular error isn't bringing down the app process, I'd suggest changing to `log.Warnf`. What are your thoughts?

internal/file/zip_read_closer.go
```diff
@@ -0,0 +1,149 @@
...
+
+// ZipReadCloser is a drop-in replacement for zip.ReadCloser (from zip.OpenReader) that additionally considers zips
+// that have bytes prefixed to the front of the archive (common with self-extracting jars).
+type ZipReadCloser struct {
+	*zip.Reader
+	io.Closer
+}
+
+// OpenZip provides a ZipReadCloser for the given filepath.
+func OpenZip(filepath string) (*ZipReadCloser, error) {
```
luhring, line 26: nit: Can `OpenZip` be unexported?

internal/file/zip_file_traversal_test.go
```diff
@@ -127,41 +125,119 @@ func TestUnzipToDir(t *testing.T) {
...
+func prepZipSourceFixtureWithPrependedBytes(tb testing.TB, value string) func(tb testing.TB) string {
+	if len(value) == 0 {
+		tb.Fatalf("no bytes given to prefix")
 	}
+	return func(t testing.TB) string {
+		archivePath := prepZipSourceFixture(t)
 
-	t.Logf("running from: %s", cwd)
+		// create a temp file
+		tmpFile, err := ioutil.TempFile("", "syft-ziputil-archive-TEST-")
```
luhring, line 170: `"syft-ziputil-archive-TEST-"` — This string might be worth extracting as a constant, partially for consistency, partially for developer ease as we add tests.

internal/file/zip_file_traversal_test.go
```diff
@@ -127,41 +125,119 @@ func TestUnzipToDir(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("unable to create tempfile: %+v", err)
+		}
+		defer tmpFile.Close()
+
+		// write junk to the temp file
```
alfredodeza, line 176: this is not writing _junk_, it is writing the _value_ argument. Or is it considered junk?

internal/file/zip_file_traversal_test.go
```diff
@@ -127,41 +125,119 @@ func TestUnzipToDir(t *testing.T) {
...
+
+			assertZipSourceFixtureContents(t, actual, expected)
+		})
 	}
+}
 
-	cwd, err := os.Getwd()
-	if err != nil {
-		t.Errorf("unable to get cwd: %+v", err)
+func prepZipSourceFixtureWithPrependedBytes(tb testing.TB, value string) func(tb testing.TB) string {
```
alfredodeza, line 162: is this really prepending _bytes_? why is it taking a `string` as an argument in that case? It seems that the name wants to be a full sentence - can it be enhanced with commented docs instead while reducing the name to something shorter?

internal/file/zip_file_traversal_test.go
```diff
@@ -127,41 +125,119 @@ func TestUnzipToDir(t *testing.T) {
...
+			t.Fatalf("unable to copy source to dest: %+v", err)
+		}
+
+		sourceFile.Close()
 
-	aFilePath := filepath.Join("some-dir", "a-file.txt")
-	bFilePath := filepath.Join("b-file.txt")
+		// remove the original archive and replace it with the temp file
+		if err := os.Remove(archivePath); err != nil {
+			t.Fatalf("unable to remove original source archive")
```
alfredodeza, line 196: can this include the path it couldn't remove?

internal/file/zip_file_traversal_test.go
```diff
@@ -127,41 +125,119 @@ func TestUnzipToDir(t *testing.T) {
...
+		}
 
-	expected := map[string]string{
-		aFilePath: "A file! nice!",
-		bFilePath: "B file...",
+		if err := os.Rename(tmpFile.Name(), archivePath); err != nil {
+			t.Fatalf("unable to move new archive to old path")
```
alfredodeza, line 200: same here, I think this would benefit from including the path information

internal/file/zip_read_closer.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// Make sure directoryOffset points to somewhere in our file.
+	if o := int64(d.directoryOffset); o < 0 || o >= size {
+		return 0, zip.ErrFormat
+	}
+	return startOfArchive, nil
+}
+
+func findSignatureInBlock(b []byte) int {
```
alfredodeza, line 137: is this func also derived from the zip package? it looks terrifying

internal/file/zip_file_manifest.go
wagoodman: this was moving the code (I didn't change any contents), but I can make this update

internal/file/zip_read_closer.go
wagoodman: it can, but we should be opening all zips with this util now (including future uses of `zip.OpenReader` should use this instead). It happens to be that no other function internally yet.

internal/file/zip_file_traversal_test.go
wagoodman: these should be roughly unique for each test or helper, I'll adjust the name.

internal/file/zip_file_traversal_test.go
wagoodman: swapped to `prependZipSourceFixtureWithString`

luhring (approved):

PR #433, opened 2021-06-09
Export the presenter interface so it may be used by external projects

dakaneye:


wagoodman (approved):
Thanks for the update 👍 

Just a heads up, we may be changing the way this works in the near future (how data makes it to presenters and surrounding topics), but we'll make certain to keep a way to obtain a presenter exported.

PR #437, opened 2021-06-13
Fix yarn.lock parsing

### LINKED ISSUE #430: erroneous npm package versions appearing in output
**What happened**:
When using `syft`, we are occasionally seeing multiple copies of NPM packages in the output, referencing versions that do not exist.

Steps to Reproduce
This assumes that one has node.js and `npm` 6.x or 7.x installed on their machine.  These steps were run on macOS 11.2.3, but I have no reason to believe that this is platform-dependent:

npm install react-native
syft packages dir:node_modules

Expected Result
array-unique 0.3.2 npm

Actual Behavior
array-unique 0.2.4 npm
array-unique 0.3.2 npm
array-unique 4.10.1 npm

Errata
This behavior is not limited to the `array-unique` package, but it is being used as an example.

There are no such versions of the `array-unique` package as 0.2.4 and 4.10.1.  Those version strings DO exist in the `yarn.lock` file, but attached to completely different packages.

**Environment**:
- Output of `syft version`:
Application:   syft
Version:       0.16.1
BuildDate:     2021-05-25T22:04:01Z
GitCommit:     8be0d988e7e793d62afd2e9e2a881428322329cb
GitTreeState:  clean
Platform:      darwin/amd64
GoVersion:     go1.16.4
Compiler:      gc
sw_vers 
ProductName:	macOS
ProductVersion:	11.2.3
BuildVersion:	20D91

luhring:
Fixes #430

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_yarn_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_yarn_lock.go
@@ -5,9 +5,8 @@ import (
 	"fmt"
 	"io"
 	"regexp"
-	"strings"
 
-	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/pkg/cataloger/common"
 )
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/javascript/parse_yarn_lock.go
```diff
@@ -5,69 +5,62 @@ import (
...
 )
 
 // integrity check
 var _ common.ParserFn = parseYarnLock
 
-var composedNameExp = regexp.MustCompile("^\"(@{1}[^@]+)")
-var simpleNameExp = regexp.MustCompile(`^[a-zA-Z\-]+@`)
-var versionExp = regexp.MustCompile(`^\W+(version)\W+`)
+var (
+	composedNameExp = regexp.MustCompile(`^"(@[^@]+)`)
```
wagoodman, line 18: nit / optional: it'd be nice to have some match examples in comments around these regex lines

wagoodman (approved):

Issue #444, opened 2021-06-23
Enhance SPDX Support


wagoodman:
Completing https://github.com/anchore/syft/issues/213 adds support for generating SPDX documents, however, there are several opportunities to expand upon full what can be expressed in an SPDX document. For instance, we have a file cataloger and are aware for select package types which files belong to a package... we could be leveraging the "FilesAnalyzed" package indication and showing the file digests for these package files. Additionally we could be adding relationships between these files and packages. For packages with transitive dependencies known, we could be building a dependency graph. There are many directions this can take, this issue is here as a placeholder to show that we could be supporting more of what can be expressed in an SPDX document.

bureado:
See #451 as it's possible that `filesAnalyzed` mandates `packageVerificationCode` as per: https://spdx.github.io/spdx-spec/3-package-information/#39-package-verification-code

WRT describing dependencies in the SPDX output, I particularly love the idea because IMHO it's existing metadata which can be potentially valuable for the SBOM consumer, but I can also see the argument for "flat SBOMs". I would say there are two levels, one would be literally taking a big string with the dependency information for each package and adding it as a comment or something semantically equivalent to a comment in the SPDX document, the next level would be converting those package names from their native names to the SPDX identifier or purl spec for proper linkage.

Another thing which I'm not sure if you're interested in considering is adding any leftover _files_ to the final SBOM document, in particular SPDX supports files as you know: https://spdx.github.io/spdx-spec/4-file-information/

This would be for anything else in the tree that hasn't been accounted for and that is relevant to the SBOM (i.e., not stuff under /home, etc., unless explicitly included), when thinking in terms of utilities I think about the output of `debsums -c` or `cruft` as potential signals that I might need to explicitly call out a file in addition to packages. One particularly interesting scenario is for `Dockerfile` with `RUN wget foo` that leave files we might want to account for, maybe it's the golang release, or terraform, or the jx client, we can already hash them and run libmagic on them but if we can also make it easy for SBOM publishers to identify which layer introduced the file maybe they can easily add a comment showing what line in the Dockerfile (or in their `packer` script or in their `debootstrap` includes...) introduced the component as some form of auxiliary attestation.

Just a few ideas. Thanks for giving SPDX enrichment additional thought!

PR #445, opened 2021-06-23
Add SPDX support

### LINKED ISSUE #213: Add support for SPDX SBOM format
**What would you like to be added**:  Please add support for SPDX

**Why is this needed**:  There is an ecosystem of open source projects that have adopted this format, and it would be great to let this community be able to work with your tools as well. 

**Additional context**:
Details on the format can be found at: https://spdx.github.io/spdx-spec/
Happy to work with anyone interested in doing the port,  I just lack the "go" programming chops to submit a PR in myself.

wagoodman:
This PR adds initial SPDX support for the tag-value and json formats via the `spdx` and `spdx-json` output options. Any discovered packages and package metadata that is compatible with the SPDX spec has been added. What has not been added yet is any relationship data as well as file data (files owned by packages and their metadata) --this will be added in a follow up (https://github.com/anchore/syft/issues/444).

Specific additions:
- Adds `internal/presenter/packages/spdx_*_presenter` for both new formats
- Adds `internal/presenter/packages/model/spdx22` structs that match the ontology needed for the JSON presenter (specific to SPDX 2.2)
- Refactors presenter tests for all presenters to use shared helpers (lots of code reuse here)
- Adds go-generated code based on the SPDX license list under `internal/spdxlicense` (there is an integration test to ensure this list always remains up to date)

Closes #213


internal/presenter/packages/spdx_helpers.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	if len(parsedLicenses) == 0 {
+		return "NOASSERTION"
+	}
+
+	return strings.Join(parsedLicenses, " AND ")
+}
+
+func noneIfEmpty(value string) string {
```
dakaneye, line 61: nit: this seems like it could be a more global string helper

(feel free to ignore this comment tho)

internal/spdxlicense/generate_license_list.go
```diff
@@ -0,0 +1,92 @@
+// +build ignore
```
dakaneye, line 1: what's the purpose of this code?

dakaneye (approved):
wow this was a LOT!

I like consolidating the presenter tests the way you did, all of this code appears to have great test coverage in general. 

I didn't go over what each field was in each struct, but the go code itself looks great, and easy to follow. 

I'm also not super familiar with the spdx stuff and wanted some additional background on the license and schema things that were added but don't want to block based on that.

internal/presenter/packages/spdx_helpers.go
wagoodman: The returned "NONE" value is pretty coupled to the SPDX spec and it's verbiage --for now I think it belongs with the SPDX helpers

internal/spdxlicense/generate_license_list.go
wagoodman: SPDX spec dictates which authoritative license names can be reported out in an SPDX SBOM. They maintain the authoritative list at https://spdx.org/licenses/licenses.json and the version of this list needs to be additionally reported in the SBOM. This `generate_license_list.go` responds to the `go generate ./...` command in this repo based on the `//go:generate go run generate_license_list.go` comment in `license.go`. 

Specifically the `// +build ignore` comment in `generate_license_list.go` is meant to keep this go source file out of the syft build. That is, when running `go build ./...` this "build tag" is used to tell the compiler to skip this file altogether when making the syft binary. The `generate_license_list.go` file is still kept specially near the file it is generating for simplicity (`license_list.go`).

So in order to build syft with the latest license list (pulled from https://spdx.org/licenses/licenses.json) one needs to run:
```
go generate ./...
go build ./...
```

But if you want to keep the existing license list and just build syft then it you do the same as you would today:
```
go build ./...
```

note: I've got an integration test that checks to see if there is a later version of the license list than what we have now --if there is then the test fails and notes in the output as to why.

wagoodman:
@dakaneye  re: what's being added schema-wise for SPDX. I pulled in the SPDX JSON schema and added a JSON schema validation test as a CLI test to ensure that, for the image tested against, syft produces valid JSON relative to the published SPDX JSON schema (v2.2).

PR #447, opened 2021-06-26
Minimal changes to use ID_LIKE

bureado:
While scanning a Debian derivative that is not mapped by `syft` I discovered that the entire `name:version:like` tuple is `nil`. The functional impact is limited as the _catalogers_ still find the DPKG files, but there's an empty `distro` object in the final JSON and the `purl` URLs aren't computed either.

This PR tries lazily to achieve that by avoiding returning `nil` early and attempting mapping again with the `ID_LIKE` value, if it was detected.

This of course means the `purl` URLs might be inaccurate and it might warrant some form of debug warning to the user, or a strict mode.

One final element to consider is that the distro _name_ (as declared in, say, `/etc/os-release`) seems to get effectively overridden by the way `syft` calls distros (the mapping result) but this is minor (it can be fixed in the resulting output very easily) and I didn't suggest any changes for that.

I can build locally, but I was unable to run unit or integration tests locally as I use `podman` and didn't want to mess the `Makefile` too much by trying to hack the calls.

Thank you for the opportunity to learn a little bit from the codebase and for all the very cool work happening with `syft` lately. While this is a minor patch, I still look forward to learning from your observations, input and continued work.

### DIFF
--- a/syft/distro/identify.go
+++ b/syft/distro/identify.go
@@ -97,8 +97,7 @@ func assemble(name, version, like string) *Distro {
 
 	// If it's an unknown distro, try mapping the ID_LIKE
 	if !ok && len(like) != 0 {
-		name = like
-		distroType, ok = IDMapping[name]
+		distroType, ok = IDMapping[like]
 	}
 
 	if ok {
--- a/syft/distro/identify.go
+++ b/syft/distro/identify.go
@@ -95,6 +95,12 @@ func assemble(name, version, like string) *Distro {
 		return nil
 	}
 
+	// If it's an unknown distro, try mapping the ID_LIKE
+	if !ok && len(like) != 0 {
+		name = like
+		distroType, ok = IDMapping[name]
+	}
+
 	if ok {
 		distro, err := NewDistro(distroType, version, like)
 		if err != nil {
--- a/syft/distro/identify.go
+++ b/syft/distro/identify.go
@@ -91,7 +91,7 @@ func assemble(name, version, like string) *Distro {
 	distroType, ok := IDMapping[name]
 
 	// Both distro and version must be present
-	if len(name) == 0 {
+	if len(name) == 0 && len(version) == 0 {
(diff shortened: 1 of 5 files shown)


bureado:
To test this, you could `sed 's/debian/foo/g' /etc/os-release && echo ID_LIKE=debian >> /etc/os-release` or similar on a standard `debian:latest` image.

I was trying to run `syft` on the Azure Cloud Shell base image referred to here: https://github.com/Azure/CloudShell/blob/master/linux/base.Dockerfile#L16

The changes you would see between builds are, for example:

```
--- a.json      2021-06-25 21:08:20.889421445 -0700
+++ b.json      2021-06-25 21:07:38.721420687 -0700
@@ -43350,9 +43350,9 @@
   }
  },
  "distro": {
-  "name": "",
-  "version": "",
-  "idLike": ""
+  "name": "debian",
+  "version": "10",
+  "idLike": "debian"
  },
```

Or:

```
--- a.json      2021-06-25 21:08:20.889421445 -0700
+++ b.json      2021-06-25 21:07:38.721420687 -0700
@@ -32,7 +32,7 @@
     "cpe:2.3:a:adduser:adduser:3.118:*:*:*:*:*:*:*",
     "cpe:2.3:a:*:adduser:3.118:*:*:*:*:*:*:*"
    ],
-   "purl": "",
+   "purl": "pkg:deb/debian/adduser@3.118?arch=all",
    "metadataType": "DpkgMetadata",
    "metadata": {
     "package": "adduser",
```

wagoodman (approved):
This is a great idea --incorporating the distro `like` field was added after the initial implementation. Using the `like` field as an additional indicator of the name is a great way to cover the case you're pointing out here.

100% agreed with your point about being able to run testing without docker. Though some container runtime would be required in some cases, we should be more amenable to other available container runtime options here.

I've added a unit test to cover the case in question 👍

Issue #416, opened 2021-05-20
Disk space not freed after syft command


rmkanda:
**What happened**:
Whenever I ran syft command to analyse a image the disk usage goes up and once the syft is completed. 
The utilized disk space is not freed.

What is the temporary directory? Can the disk space freed automatically?

**Environment**:
- Output of `syft version`: 0.15.2
- OS MacOS

luhring:
Hi @rmkanda! Thanks for reporting this. Our expectation is that all files we create **_are cleaned up_** after the run completes. (The exceptions to this are in cases where the app crashes or when the app is terminated early.) So this is definitely a problem, and we need to look more into it.

>What is the temporary directory?

We use your OS's tmp directory (seen by running `echo $TMPDIR`), and within that directory, we create a few directories as needed for the analysis job. These directory names should include the names of our projects, like `syft` and `stereoscope`.

I think I'm able to reproduce what you're describing...

### How to reproduce

I have a split screen terminal. On the left, I’m running `watch -n 0.2 'ls -l $TMPDIR | grep -i stereoscope | wc -l'` . On the right, I’m running `syft registry:anchore/anchore-engine:latest` repeatedly. And then I’m watching the count of stereoscope temp dirs go up and down.

I’m finding that this count returns to the starting point _only sometimes_.

I'll be looking more into what might be causing this.

DatGameh:
Hi @luhring, I have a few question about these temp files:

For context, I intend to use Syft for large-volume scans of about ~1000-2000 images at a time, but due to this bug, these large `stereoscope` directories of ~10GB each are created in the `tmp` folder but never deleted. This quickly consumes hundreds of GB of storage in a short period.

![image](https://github.com/anchore/syft/assets/71962500/741a641c-b703-476f-8205-ec9009f3de26)

There are two things I would like to ask regarding this:
1. Can I safely delete all these `stereoscope` and `syft` directories in `tmp` without breaking Syft?
2. Is it possible to define where Syft saves these temp files? 
    * For instance, instead of the OS's 'tmp' directory, save them in a `temp` directory in my project that gets deleted after the program finishes.

Thank you!

kzantow:
@DatGameh first of all, I should note that Syft _should_ be cleaning this stuff up and if it's not, that's a bug we need to fix (I thought it had been, when this was closed). To answer your specific questions:
1. After Syft runs, you can definitely delete those directories, they won't be used again
[...]

PR #448, opened 2021-06-27
Split UI from event handling

wagoodman:
Today the event loop is coupled with the specific UI details, which is not ideal. This PR splits out those concerns, migrating the event loop from the `ui` package to `cmd` and extracting the `ui.UI` function into an interface for setting up a UI, handling individual events from the event loop, and tearing down the UI. This allows for flow control, signal handling, and worker/event management to be shared, and the UI specifics to be implemented behind an interface --this allows for the ability to swap out the ETUI implementation for the logger implementation, but now without duplicating flow control logic.

Closes https://github.com/anchore/syft/issues/416


cmd/packages.go
```diff
@@ -186,9 +185,13 @@ func bindPackagesConfigOptions(flags *pflag.FlagSet) error {
 }
 
 func packagesExec(_ *cobra.Command, args []string) error {
-	errs := packagesExecWorker(args[0])
-	ux := ui.Select(appConfig.CliOptions.Verbosity > 0, appConfig.Quiet)
-	return ux(errs, eventSubscription)
+	return eventLoop(
+		packagesExecWorker(args[0]),
```
luhring, line 189: nit: If we're passing in a specific arg, it could be a nice hint to the reader to identify the arg via an explicit parameter for `packagesExec`, rather than receive the whole slice and grab item `0` just in time.

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/gookit/color v1.2.7
 	github.com/hashicorp/go-multierror v1.1.0
 	github.com/hashicorp/go-version v1.2.0
+	github.com/k0kubun/go-ansi v0.0.0-20180517002512-3bf9e2903213
```
luhring, line 24: Just double-checking that this was intended, since I couldn't remember if we're adding new ANSI logic here.

ui/event_handlers.go
```diff
@@ -22,8 +23,8 @@ import (
 )
 
 const maxBarWidth = 50
-const statusSet = common.SpinnerDotSet // SpinnerCircleOutlineSet
-const completedStatus = "✔"            // "●"
+const statusSet = components.SpinnerDotSet // SpinnerCircleOutlineSet
+const completedStatus = "✔"                // "●"
```
luhring, line 27: Just curious: What is `"●"` for?

luhring (approved):
Thanks! 🎉

go.mod
wagoodman: I've used this package before and decided to use it for the hide/show of the cursor (was intended)

ui/event_handlers.go
wagoodman: I can remove this --it was used during early development when I was flopping between options

PR #449, opened 2021-06-27
Improve temp dir generation + cleanup

wagoodman:
This PR pulls in the changes from https://github.com/anchore/stereoscope/pull/71 as well as imports the `file.TempDirGenerator` to follow the same pattern used in stereoscope. In the near future we can split out some of the `file` package functionality in stereoscope into another repo.

Note: tests will fail until the stereoscope PR is merged and the go.mod is updated here. Also note that the base branch is `split-ui-and-event-handling` which should be merged first and the base here changed to `main`.

Partially addresses https://github.com/anchore/syft/issues/416


wagoodman:
Given the comments on https://github.com/anchore/stereoscope/pull/71 this will need to be brought in under https://github.com/anchore/syft/pull/448 to account for cleanup interface changes.

PR #455, opened 2021-06-30
Add help message when no arguments are provided

wagoodman:
Now when no arguments are provided, the default help text is displayed:
```
$ syft packages
Generate a packaged-based Software Bill Of Materials (SBOM) from container images and filesystems

Usage:
   packages [SOURCE] [flags]

...

an image/directory argument is required
exit status 1

```

### DIFF
--- a/cmd/packages.go
+++ b/cmd/packages.go
@@ -47,7 +47,7 @@ const (
 
 var (
 	packagesPresenterOpt packages.PresenterOption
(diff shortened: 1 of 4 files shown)


luhring (approved):
Looks good! We can probably remove the extra lines of output after `Use " [command] --help"  ...`, but that can be done later. Having the help text back is the high order bit here 🎉 

image

PR #456, opened 2021-06-30
Redirect cursor hide/show to stderr

wagoodman:
https://github.com/anchore/syft/pull/448 pulled in an external lib to perform hiding/showing the cursor. However, this needs to be done relative to the right output stream (in our case, stderr), otherwise `syft | less` will continue to show the cursor.

This PR restores this functionality.

### DIFF
--- a/internal/ui/ephemeral_terminal_ui.go
+++ b/internal/ui/ephemeral_terminal_ui.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"sync"
 
@@ -11,7 +12,6 @@ import (
 	"github.com/anchore/syft/internal/logger"
 	syftEvent "github.com/anchore/syft/syft/event"
 	"github.com/anchore/syft/ui"
-	"github.com/k0kubun/go-ansi"
 	"github.com/wagoodman/go-partybus"
 	"github.com/wagoodman/jotframe/pkg/frame"
 )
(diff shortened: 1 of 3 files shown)


luhring (approved):

internal/ui/ephemeral_terminal_ui.go
```diff
@@ -126,14 +128,22 @@ func (h *ephemeralTerminalUI) flushLog() {
...
 
 func (h *ephemeralTerminalUI) Teardown(force bool) error {
 	h.closeScreen(force)
-	ansi.CursorShow()
+	showCursor(h.output)
 	return nil
 }
+
+func hideCursor(output io.Writer) {
+	fmt.Fprint(output, "\x1b[?25l")
```
dakaneye, line 144: can we get a method comment to add some context here? I have some general questions that might be implicit in the method name, but want to make sure...
* what does the `\x1b[?25l` code correspond to? 
* why does printing it to the specified output stream hide the cursor? 

These questions also apply to the showCursor method

dakaneye (commented):
I'll be the first to admit my ANSI skills are lacking,  but from what I can tell in the code changes, it looks like things were already logging to stderr, and had some questions re: the actual hide and show cursor methods (sorry if those questions are things I should know already)

The other question I have is if there's any testing required (but you've mentioned before that we haven't typically unit tested the ETUI stuff in the past)

internal/ui/ephemeral_terminal_ui.go
wagoodman: sorry @dakaneye , just as I merged your message popped up (I didn't refresh). I'll come back after the next couple of meetings and provide further context here 👍

PR #467, opened 2021-07-27
Power user dir

houdini91:
Hi guys.
My first contribution here, sorry if i am missing something.

Add support of using directory type sources on the **power-user** command.

**Following support have been addressed**
- Digest, contents and secret catalogers require `read` permissions, encountering  a permission error fails the catalogers.
    **Fix**: Log the error (debug) instead of failing, skip ahead.
- Digest and metadata cataloger will fail saying all files sent do not exists in resolver  - cataloger fails.
    **Reason**: `directoryResolver.AllLocations`, `directoryResolver.FilesByGlob` returns locations that is not connected to the directory resolver file references.
    **PR**: Return locations that are connected to directory resolver file references.
    
**Test**
- Added cli test `test/cli/power_user_cmd_test.go`, `default-dir-results-w-pkg-coverage` using the `dir:test-fixtures/image-pkg-coverage` source.
 - Added cli test `test/cli/power_user_cmd_test.go`, `defaut-secrets-dir-results-w-reveal-values` using the `dir:test-fixtures/image-secrets` source.
- Unit test - location assertion changed from structural compare to string compare on the location real path field.
 (`TestDirectoryResolverDoesNotIgnoreRelativeSystemPaths`, `TestClassifierCataloger_DefaultClassifiers_PositiveCases`)

**Open issue**
- Should directory resolver locations include the relative path in the virtual path fields and use the reference path for the real path field?

Hope minor patch helps.
Ps love your work.

### DIFF
--- a/internal/err_helper.go
+++ b/internal/err_helper.go
@@ -30,9 +30,9 @@ func IsErrPath(err error) bool {
 }
 
 func IsErrPathPermission(err error) bool {
-	path_err, ok := err.(ErrPath)
+	pathErr, ok := err.(ErrPath)
 	if ok {
-		return os.IsPermission(path_err.Err)
+		return os.IsPermission(pathErr.Err)
 	}
 	return ok
 }
--- a/internal/err_helper.go
+++ b/internal/err_helper.go
@@ -1,7 +1,9 @@
 package internal
 
 import (
+	"fmt"
 	"io"
+	"os"
 
 	"github.com/anchore/syft/internal/log"
 )
@@ -11,4 +13,26 @@ func CloseAndLogError(closer io.Closer, location string) {
 	if err := closer.Close(); err != nil {
 		log.Warnf("unable to close file for location=%q: %+v", location, err)
 	}
+}
+
+type ErrObserve struct {
+	Path string
+	Err  error
+}
+
+func (e ErrObserve) Error() string {
+	return fmt.Sprintf("unable to observe contents of %+v: %v", e.Path, e.Err)
+}
+
+func IsErrObserve(err error) bool {
+	_, ok := err.(ErrObserve)
+	return ok
+}
+
+func IsErrObservePermission(err error) bool {
+	observe_err, ok := err.(ErrObserve)
+	if ok {
+		return os.IsPermission(observe_err.Err)
+	}
+	return ok
 }
--- a/internal/err_helper.go
+++ b/internal/err_helper.go
@@ -15,24 +15,24 @@ func CloseAndLogError(closer io.Closer, location string) {
 	}
 }
 
-type ErrObserve struct {
+type ErrPath struct {
 	Path string
 	Err  error
 }
 
-func (e ErrObserve) Error() string {
+func (e ErrPath) Error() string {
 	return fmt.Sprintf("unable to observe contents of %+v: %v", e.Path, e.Err)
 }
 
-func IsErrObserve(err error) bool {
-	_, ok := err.(ErrObserve)
+func IsErrPath(err error) bool {
+	_, ok := err.(ErrPath)
 	return ok
 }
 
-func IsErrObservePermission(err error) bool {
-	observe_err, ok := err.(ErrObserve)
+func IsErrPathPermission(err error) bool {
+	path_err, ok := err.(ErrPath)
 	if ok {
-		return os.IsPermission(observe_err.Err)
+		return os.IsPermission(path_err.Err)
 	}
 	return ok
 }
--- a/syft/source/directory_resolver.go
+++ b/syft/source/directory_resolver.go
@@ -41,17 +41,10 @@ type directoryResolver struct {
 	errPaths      map[string]error
 }
 
-func newDirectoryResolver(fileTree *filetree.FileTree, root string, pathFilters ...pathFilterFn) (*directoryResolver, error) {
+func newDirectoryResolver(root string, pathFilters ...pathFilterFn) (*directoryResolver, error) {
 	cwd, err := os.Getwd()
 	if err != nil {
 		return nil, fmt.Errorf("could not create directory resolver: %w", err)
-	}
-
-	var resolverFileTree *filetree.FileTree
-	if fileTree == nil {
-		resolverFileTree = filetree.NewFileTree()
-	} else {
-		resolverFileTree = fileTree
 	}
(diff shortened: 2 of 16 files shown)


houdini91:
**Update**: Added support for `Filetree` sharing between directory resolver - this feature is needed so that catalogers running concurrently can work on the same references.

**Following issues are targeted**
* Share `Filetree` between resolver via `source` struct. 
* Thread save directory resolver indexing - Adding a lock around the indexing logic
* Share indexed file information - copied shared tree references before indexing.
* Fix file metadata GID,UID sample (Linux based only).

**Test**
* Added shared directory unit test - `TestDirectoryResolver_SharedTreeMultipleFilesByPath`

wagoodman:
@houdini91 very nice addition!

I see the performance reason for introducing the shared trees between `directoryResolver` instances, and it's a good one. However, I think there is an opportunity to simplify that approach while still keeping the same performance. Instead of creating multiple instances of the `directoryResolver` and injecting a shared file tree as a dependency, it could be simpler to memoize the `Source. FileResolver()` method such that only one `directoryResolver` instance is created and cached for future calls. Additional benefit is that the `FileTree` remains an internal concern of the `directoryResolver` and there would no longer be a need to keep a mutex on `Source`.

What do you think about this as an alternative approach to the shared filetree?

syft/source/directory_resolver.go
```diff
@@ -276,15 +306,22 @@ func (r *directoryResolver) AllLocations() <-chan Location {
 func (r *directoryResolver) FileMetadataByLocation(location Location) (FileMetadata, error) {
 	info, exists := r.infos[location.ref.ID()]
 	if !exists {
-		return FileMetadata{}, fmt.Errorf("location: %+v : %w", location, os.ErrExist)
+		return FileMetadata{}, fmt.Errorf("location: %+v : %w", location, os.ErrNotExist)
+	}
+
+	uid := -1
+	gid := -1
+	if stat, ok := info.Sys().(*syscall.Stat_t); ok {
```
wagoodman, line 314: good call 👍

internal/err_helper.go
```diff
@@ -12,3 +14,25 @@ func CloseAndLogError(closer io.Closer, location string) {
 		log.Warnf("unable to close file for location=%q: %+v", location, err)
 	}
 }
+
+type ErrObserve struct {
```
wagoodman, line 18: nit: I feel that there may be a more descriptive name for this error. Maybe `ErrPath`?

houdini91: Done

syft/source/directory_resolver.go
houdini91: @wagoodman  I totally agree, sharing a _directoryResolver_ is probably simpler than sharing a _FileTree_ pointer.
Originally thought about sharing the _directoryResolver_ pointer but i was not sure how it will effect parts i have not yet explored.

Can you please clarify a bit on the _Mutex_ comment? 
It seems to me that even if _source.FileResolver()_ creates only one directoryResolver (Stores pointer in struct) the initialization process still can create a race (Two threads initialization the first directoryResolver)
Eg:
```
// Called by catalogers child threads concurrently 
func (s Source) FileResolver(scope Scope) (FileResolver, error) {
	switch s.Metadata.Scheme {
	case DirectoryScheme:
                if s.dirResolver == nil {
                          // Not thread safe
		          resolver, err := newDirectoryResolver(s.Tree, s.Metadata.Path)
                          if err == nil {
                                 s.dirResolver = resolver
                          }
                }
                return s.dirResolver, err
...
...
```
* Maybe we can initialize the directory resolver in the source _New()_ constructor function? (Called by main thread)?
* Maybe leave the Mutex and only replace the FileTree pointer with a directoryResolver pointer?

houdini91:
@wagoodman  I totally agree, sharing a _directoryResolver_ is probably simpler than sharing a _FileTree_ pointer.
Originally thought about sharing the _directoryResolver_ pointer but i was not sure how it will effect parts i have not yet explored.

Can you please clarify a bit on the _Mutex_ comment? 
It seems to me that even if _source.FileResolver()_ creates only one directoryResolver (Stores pointer in struct) the initialization process still can create a race (Two threads initialization the first directoryResolver)
Eg:
```
// Called by catalogers child threads concurrently 
func (s Source) FileResolver(scope Scope) (FileResolver, error) {
	switch s.Metadata.Scheme {
	case DirectoryScheme:
                if s.dirResolver == nil {
                          // Not thread safe
		          resolver, err := newDirectoryResolver(s.Tree, s.Metadata.Path)
                          if err == nil {
                                 s.dirResolver = resolver
                          }
                }
                return s.dirResolver, err
...
...
```
* Maybe we can initialize the directory resolver in the source _New()_ constructor function? (Called by main thread)?
* Maybe leave the Mutex and only replace the FileTree pointer with a directoryResolver pointer?

wagoodman:
> Maybe we can initialize the directory resolver in the source New() constructor function? (Called by main thread)?

This wouldn't be ideal since any `FileResolver ` calls may be passed different scopes. I realize that for the case of the directory resolve this doesn't technically matter (since a directory resolver is like having an image with a single layer, so different scope options have no effect), but doing resolver initialization in two places where it could be focused to one is where the concern is.

> Maybe leave the Mutex and only replace the FileTree pointer with a directoryResolver pointer?

You're right, I overlooked this --the mutex for the resolver technically should already be there today. Let's leave the mutex 👍

houdini91:
@wagoodman 

> it could be simpler to memoize the `Source. FileResolver()` method such that only one `directoryResolver` instance
Updated code to reflect change. 

* One issue i had was that the _source struct_ was not passed around by value causing any pointer initialization not be reflected outside of the instance scope.
I have updated code to use a pointer of the _source struct_ instead.

wagoodman:
@houdini91 this looks good, thanks for the updates! Only one last change to get past validations: https://github.com/anchore/syft/pull/467/checks?check_run_id=3508971289 a linter failure for a variable name. Once passing we'll merge this in.

houdini91:
@wagoodman Fixed lint error,
 your welcome It was my pleasure.

luhring (approved):
Thanks for this! Approving per [review](https://github.com/anchore/syft/pull/467#issuecomment-912745156) by @wagoodman

PR #472, opened 2021-08-05
Enhance CPE generation 

### LINKED ISSUE #396: Use of asterisk (*) in CPEs leading to many false positives in vulnerability matching
It would be ideal to find ways to minimize the use of stars (a.k.a. asterisks / `*`) in CPE fields in generated CPEs to minimize the number of false positives in downstream vulnerability matching.

The ultimate goal is to reduce the count of false positives that Grype produces. The amount of FP reduction considered acceptable will be a judgment call by the team.

We'll need a feedback loop, where we can make targeted changes to Syft's CPE-generation logic and evaluate if that had the intended effect of reducing false positives in Grype's matching.

Strategies to consider trying:

1. ~Replace all `*` (any) with `-` (N/A).~
2. Do the above replacement for only the VENDOR field.

Use the available test harness tooling to get the results for any given change made.

### LINKED ISSUE #471: Enhance CPE generation to improve downstream matching in grype
After completing #396  there is room for further improvements to help with FP matches in grype.

- [x] Re-enable searching for jenkins (and other) plugins https://github.com/anchore/syft/blob/v0.19.1/syft/pkg/cataloger/cpe.go#L334 
- [x] Generate CPE candidates to attempt to match on umbrella projects from sub-component packages (e.g. match on a vendor of `cloudbees` from a package called `cloudbees-some-component`) based on hyphen/underscore splitting.
- [x] When generating vendors and products during CPE generation for java packages, both the artifactID and groupID should be used for both vendor and product values, not just artifactIDs for products and groupIDs for vendors in isolation.
- [x] Add “Extension-Name” and "Automatic-Module-Name" for groupID processing for java packages during CPE generation (a source for product and vendor information).
- [x] Add “ruby-lang” vendor value for gems.

wagoodman:
Closes #396 
Closes #471


syft/pkg/cataloger/cpe.go
```diff
@@ -323,15 +343,152 @@ func groupIDFromPomProperties(p pkg.Package) string {
...
+		// there is a strong indication that the artifact ID is really a group ID, don't use it
+		return ""
+	}
+	return artifactID
 }
 
-func shouldConsiderGroupID(groupID string) bool {
-	if groupID == "" {
-		return false
+func groupIDsFromJavaPackage(p pkg.Package) (groupIDs []string) {
```
luhring, line 354: What is the purpose of naming the function's result parameter?

luhring: (ditto this question for other occurrences below, but I'll keep to just one PR comment for all of them)

syft/pkg/cataloger/cpe.go
```diff
@@ -323,15 +343,152 @@ func groupIDFromPomProperties(p pkg.Package) string {
...
+			}
+		}
+
+		// keep track of the trailing separator for the next loop
+		lastToken = rawCandidate[len(rawCandidate)-1]
+	}
+	return results
+}
+
+func trimHyphenOrUnderscore(r rune) bool {
```
luhring, line 455: It'd be nice to have a doc comment that explains why this function exists and what it does. (I did figure this out through looking around, but it'd be nice to have the answer more immediately if possible.)

syft/pkg/cataloger/cpe_specificity.go
```diff
@@ -1,31 +1,46 @@
...
 type ByCPESpecificity []wfn.Attributes
 
-// Implementing sort.Interface
 func (c ByCPESpecificity) Len() int      { return len(c) }
 func (c ByCPESpecificity) Swap(i, j int) { c[i], c[j] = c[j], c[i] }
 func (c ByCPESpecificity) Less(i, j int) bool {
-	return countSpecifiedFields(c[i]) > countSpecifiedFields(c[j])
+	iScore := countSpecifiedFields(c[i])
+	jScore := countSpecifiedFields(c[j])
+	if iScore == jScore {
```
luhring, line 18: nit: Some vertical space surrounding the if block could help readers follow what's happening, and visualize the distinct concerns being addressed

syft/pkg/cataloger/cpe_specificity.go
```diff
@@ -1,31 +1,46 @@
...
+	checksForSpecifiedField := []func(cpe wfn.Attributes) (bool, int){
+		func(cpe wfn.Attributes) (bool, int) { return cpe.Part != "", 2 },
+		func(cpe wfn.Attributes) (bool, int) { return cpe.Vendor != "", 3 },
+		func(cpe wfn.Attributes) (bool, int) { return cpe.Product != "", 4 },
+		func(cpe wfn.Attributes) (bool, int) { return cpe.Version != "", 1 },
+		func(cpe wfn.Attributes) (bool, int) { return cpe.TargetSW != "", 1 },
 	}
 
-	count := 0
+	var weightedCount int
```
luhring, line 37: I'd suggest being explicit about the initial value here, even if there's no difference in behavior. I think it helps the reader understand that we're counting, and that we may end up returning `0` here.

syft/pkg/cataloger/cpe_specificity.go
```diff
@@ -1,31 +1,46 @@
...
 }
 
 func countSpecifiedFields(cpe wfn.Attributes) int {
-	checksForSpecifiedField := []func(cpe wfn.Attributes) bool{
-		func(cpe wfn.Attributes) bool { return cpe.Part != "" },
-		func(cpe wfn.Attributes) bool { return cpe.Product != "" },
-		func(cpe wfn.Attributes) bool { return cpe.Vendor != "" },
-		func(cpe wfn.Attributes) bool { return cpe.Version != "" },
-		func(cpe wfn.Attributes) bool { return cpe.TargetSW != "" },
+	checksForSpecifiedField := []func(cpe wfn.Attributes) (bool, int){
```
luhring, line 29: Given this change in the function's behavior, I'd suggest updating the function name to reflect the implementation (i.e. indicate that there's more happening than just a counting of fields that are specified). I'm not opinionated on the new name but can provide ideas if that's useful

syft/pkg/cataloger/cpe.go
```diff
@@ -1,16 +1,27 @@
 package cataloger
```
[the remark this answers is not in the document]
wagoodman: agreed (I resisted a lot of possible refactors on this PR that I think there is a lot of room for).

internal/string_helpers.go
```diff
@@ -13,6 +13,16 @@ func HasAnyOfPrefixes(input string, prefixes ...string) bool {
 	return false
 }
 
+func ContainsString(input string, collection []string) bool {
```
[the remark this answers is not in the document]
wagoodman: It had been slice-based in the past and it tipped over in the later commits in the PR to set-based. I need to comb back through the PR and do more slice-to-set conversions here 👍

syft/pkg/cataloger/cpe_specificity.go
wagoodman: I can make the update, but I don't think I follow why the assigning a 0 when a zero-value will do helps with readability.

syft/pkg/cataloger/cpe.go
wagoodman: this was for convenience for simpler functions, no other reason

internal/string_helpers.go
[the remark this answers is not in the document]
wagoodman: I think I'll limit this transition to only the newly changed code --I'll remove the new string helper and use a set for this. When we do [future refactors](https://github.com/anchore/syft/pull/472#discussion_r685266208) we can think about the value of moving more of the cpe generation logic to set-based objects.

syft/pkg/cataloger/cpe_specificity.go
luhring: I appreciate this deference greatly 🙏 The change is helpful to me, and you've given me something to think about better articulating in the future. 🤔

syft/pkg/cataloger/cpe.go
luhring: Interesting, okay thanks! 👍

luhring (approved):
Looks great! 👏

wagoodman:
rebased / force pushed to incorporate https://github.com/anchore/syft/pull/474

Issue #242, opened 2020-10-26
Add Pipenv support (Pipfile.lock)


wagoodman:
Add support for Pipenv by cataloging packages found in the `Pipfile.lock`.

(note: this is work being broken off from anchore/grype#29)

luhring:
@wagoodman Should we close this issue at this point? (in light of #473) Or is there more to do?

wagoodman:
good catch -- nope, no more to do 👍

PR #473, opened 2021-08-09
Added parser for Pipfile.lock to cataloger

Zilborg:
Signed-off-by: Nikita <33390074+Zilborg@users.noreply.github.com>

Issue: #242

### DIFF
--- a/syft/pkg/cataloger/python/parse_pipfile_lock.go
+++ b/syft/pkg/cataloger/python/parse_pipfile_lock.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/pkg/cataloger/common"
@@ -48,7 +49,7 @@ func parsePipfileLock(_ string, reader io.Reader) ([]pkg.Package, error) {
 			return nil, fmt.Errorf("failed to parse Pipfile.lock file: %w", err)
 		}
 		for name, pkgMeta := range lock.Default {
-			version := pkgMeta.Version[2:len(pkgMeta.Version)]
+			version := strings.TrimPrefix(pkgMeta.Version, "==")
 			packages = append(packages, pkg.Package{
 				Name:     name,
 				Version:  version,
--- a/syft/pkg/cataloger/python/parse_pipfile_lock.go
+++ b/syft/pkg/cataloger/python/parse_pipfile_lock.go
@@ -56,7 +56,6 @@ func parsePipfileLock(_ string, reader io.Reader) ([]pkg.Package, error) {
 				Type:     pkg.PythonPkg,
 			})
 		}
-
 	}
 
 	return packages, nil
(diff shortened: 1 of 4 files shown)


syft/pkg/cataloger/python/parse_pipfile_lock.go
```diff
@@ -0,0 +1,63 @@
...
+
+	for {
+		var lock PipfileLock
+		if err := dec.Decode(&lock); err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("failed to parse Pipfile.lock file: %w", err)
+		}
+		for name, pkgMeta := range lock.Default {
+			version_ := pkgMeta.Version[2:len(pkgMeta.Version)]
```
wagoodman, line 51: clarification: is the version always prepended with `==` in every instance in `pipfile.lock`? or are there cases where other values could be expected?

wagoodman:
a couple of notes: 
- if you rebase your branch the integration test should pass
- `make lint-fix` should either fix or at least highlight locally the lint failure

syft/pkg/cataloger/python/parse_pipfile_lock.go
Zilborg: > clarification: is the version always prepended with `==` in every instance in `pipfile.lock`? or are there cases where other values could be expected?

`Pipfile.lock` is automatically generated and is a mirror of the current dependencies. (Like [package-lock.json](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json) in npm)

wagoodman: agreed that the pipfile.lock is automatically generated. I wanted to understand more about the possible values that get prepended to the version, in this case `==` is pretty common by inspection. Would we ever expect any other value? Such as `>=`, `>`, `~`, etc... in place of the stripped `==`?

If not then it would be slightly more ideal to use:

```suggestion
			version_ := strings.TrimPrefix(pkgMeta.Version, "==")
```

syft/pkg/cataloger/python/parse_pipfile_lock.go
```diff
@@ -0,0 +1,62 @@
...
+			packages = append(packages, pkg.Package{
+				Name:     name,
+				Version:  version_,
```
wagoodman, line 54: Is there a specific reason for the suffixed `_` on the `version` variable name? If not can you remove it?
```suggestion
				Version:  version,
```

Zilborg:
fixed `_version` in both cases + rebased branch to actual version. Now all should be fine. I hope)

wagoodman:
thanks for the contribution, @Zilborg !

syft/pkg/cataloger/python/parse_pipfile_lock.go
```diff
@@ -48,7 +49,7 @@ func parsePipfileLock(_ string, reader io.Reader) ([]pkg.Package, error) {
 			return nil, fmt.Errorf("failed to parse Pipfile.lock file: %w", err)
 		}
 		for name, pkgMeta := range lock.Default {
-			version := pkgMeta.Version[2:len(pkgMeta.Version)]
+			version := strings.TrimPrefix(pkgMeta.Version, "==")
```
Zilborg, line 52: Cool! Thx! 👍

wagoodman (approved):

PR #482, opened 2021-08-13
Add option to enable http connection to registries 

kzantow:
Adds an option to allow connection to registries over `http`. This is not a strict requirement for https://github.com/anchore/grype/issues/334 but implemented for feature parity. Related to: https://github.com/anchore/grype/pull/380

Added just a very basic test, as this is just pushing configuration to `go-containerregistry`


luhring (approved):
Looks good!

PR #483, opened 2021-08-18
Fix go module tidy checks for release pipeline

### LINKED ISSUE #458: Pipelines allow unclean go.sum files that block our release pipeline
In neither of our primary pipelines (for PRs, and for releases) do we catch "unclean" `go.sum` files, where "unclean" is defined as having excess line items that would not be present after running `go mod tidy`.

We should catch this scenario in our Static Analysis pipeline (which runs for PRs), which is as far upstream we can detect this problem, short of changes to developer environments. We have most of the pieces in place to do this. But since we run our `check-go-mod-tidy` make target after having run `go mod tidy`, we aren't catching the problem. To correctly catch the problem in the Static Analysis pipeline, we need to:

- [x] remove the `go mod tidy` command from the bootstrap step (**Note**: this leaves `go mod download` as the last Go command in the bootstrapping process)
- [x] modify `go-mod-tidy-check.sh` to detect and error out upon finding a diff between _"A"_ and _"B"_, where: _"A"_ is the original `go.sum`, and _"B"_ is state of `go.sum` after running `go mod tidy`.

wagoodman:
Closes #458

### DIFF
--- a/.github/scripts/go-mod-tidy-check.sh
+++ b/.github/scripts/go-mod-tidy-check.sh
@@ -1,27 +1,31 @@
 #!/usr/bin/env bash
-set -u
+set -eu
 
-TMP_GOMOD=$(mktemp)
-TMP_GOSUM=$(mktemp)
+ORIGINAL_STATE_DIR=$(mktemp -d "TEMP-original-state-XXXXXXXXX")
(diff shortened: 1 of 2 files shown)


luhring (approved):
👏

PR #487, opened 2021-08-22
Add support for older CentOS versions (6 & 5) by checking additional release files for information

### LINKED ISSUE #429: Distro not detected for centos:6
1. on linux os
2. docker pull centos:6
3. syft centos:6 -o json
4. distro data is null
图片

dakaneye:
Fixes #429

### DIFF
--- a/syft/distro/identify.go
+++ b/syft/distro/identify.go
@@ -35,6 +35,16 @@ var identityFiles = []parseEntry{
 		path: "/bin/busybox",
 		fn:   parseBusyBox,
 	},
+	{
+		// check for centos:6
+		path: "/etc/system-release-cpe",
+		fn:   parseSystemReleaseCPE,
+	},
+	{
+		// last ditch effort for determining older centos version distro information
+		path: "/etc/redhat-release",
+		fn:   parseRedhatRelease,
+	},
 }
 
 // Identify parses distro-specific files to determine distro metadata like version and release.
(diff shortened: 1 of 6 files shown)


syft/distro/identify.go
```diff
@@ -145,3 +155,45 @@ func parseBusyBox(contents string) *Distro {
 	}
 	return nil
 }
+
+// example CPE: cpe:/o:centos:linux:6:GA
+var systemReleaseCpeMatcher = regexp.MustCompile(`cpe:\/o:(.*?):.*?:(.*?):.*?$`)
```
wagoodman, line 160: This is a good approach, however, syft already has CPE parsing built in that we can leverage https://github.com/anchore/syft/blob/main/syft/pkg/cpe.go#L12  --can we switch to using this?

dakaneye: @wagoodman Just tried that, but it creates an import cycle :(. Unless i'm doing something horribly wrong?

wagoodman: nope, you are not doing anything wrong --it looks like pkg depends on distro to support pURLs. This hints at a refactor that we can do later, either striving for pkg to not depend on distro (medium lift-ish) or migrating cpe out of pkg (small lift). We can defer that for now.

Can you leave a comment in the code that points out we should use the CPE parser once it is importable? (no code change requested at this point 👍 )

wagoodman (approved):
looks good, we can come back and refactor this to use the existing CPE code when there is no longer a cycle (there are multiple hints to migrate CPE related things to a separate package)

syft/distro/identify.go
dakaneye: comment added!

PR #493, opened 2021-09-03
Disable docker releases

wagoodman:
Currently cannot install docker onto the mac runners in GHA, which we need for signing. The workaround is to disable docker releases for the short-term until we have a fix in place (but we cannot block binary releases on this)

Example infinite loop:
https://github.com/anchore/syft/actions/runs/1199137004


luhring (approved):
🐳

PR #496, opened 2021-09-08
(#342) - remove strong distro check

### LINKED ISSUE #342: Remove strong distro type
Today there is a `distro.Type` that is not internally used within syft (but is within `grype`).  We should consider starting to trust raw os-release data and pass it through to the JSON presenters. We should also consider removing this `distro.Type` type from syft and move it to grype.

**Additional context**:
This is from a discussion from https://anchorecommunity.slack.com/archives/C019BUXV7R6/p1614950344012600 and https://github.com/anchore/syft/pull/341

spiffcs:
### Works to Address issue from: #342  

Syft currently implements a STRONG check against the distro type when scanning a container image. If a distro is not in the known list of types it is labeled `UnknownDistroType` and `Identify` returns `nil`.

#342 centers around moving this STRONG check over to [grype](https://github.com/anchore/grype). The thought here is that syft should be able to generate an SBOM with data as close to the input as possible. Grype should be responsible for knowing which types it can accept.

This first change is just so that custom/unknown distros can be parsed and will not return as 
```
 "distro": {
  "name": "",
  "version": "",
  "idLike": ""
 },
```

Before we can close the issue, we should discuss possible breaking changes with @wagoodman where `Type` is removed. Grype will be responsible for mapping a distros `Name => Type` for the vulnerability matching section.

#### Changelog
- Add type coercion so that if we fail to find in map of known the ID is still used in distro construction
- Update test to show that `unmatchable` is now `custom` and can generate a distro


syft/distro/distro.go
```diff
@@ -8,23 +8,28 @@ import (
 
 // Distro represents a Linux Distribution.
 type Distro struct {
-	Type       Type
+	Type       string
```
spiffcs, line 11: With `Type` being removed it feels wrong to keep this field named as `Type`.

Since we have a field called `IDLike` I would be in favor of renaming this `ID` so that we can get as close to the source naming of these fields as possible.

syft/distro/distro.go
```diff
@@ -8,23 +8,28 @@ import (
...
 	Version    *hashiVer.Version
 	RawVersion string
 	IDLike     string
 }
 
 // NewDistro creates a new Distro object populated with the given values.
-func NewDistro(t Type, ver, like string) (Distro, error) {
+func NewDistro(name, ver, like string) (Distro, error) {
```
spiffcs, line 18: Same issue here - We pass name, but it's actually NOT the name field from the file we parse. I can rename this to id and then it will match in the function and be less confusing to readers who come back here.

```
ID: id,
```

syft/distro/distro.go
```diff
@@ -33,7 +38,7 @@ func NewDistro(t Type, ver, like string) (Distro, error) {
 
 // Name provides a string repr of the distro
 func (d Distro) Name() string {
-	return string(d.Type)
+	return d.Type
```
spiffcs, line 41: Should we add a `Name` field to `Distro`. We can then account for these three fields.
```
NAME
ID
ID_LIKE
```

syft/distro/identify.go
```diff
@@ -90,35 +90,21 @@ identifyLoop:
 		}
 	}
 
-	if distro != nil && distro.Type == UnknownDistroType {
```
spiffcs, line 93: Since we're always taking the supplied field and not returning a strong type I don't know if this check is still neccesary. 

Should we be checking if distro.Type is `""`

internal/spdxlicense/generate_license_list.go
```diff
@@ -1,3 +1,4 @@
+//go:build ignore
```
luhring, line 1: Just curious — what is this for?

syft/distro/distro.go
luhring: I like this idea — the more we can reuse well-known nomenclature, the better.

We'd just need to be sure to understand the consequences downstream; that is, was there any existing logic that really did think of this as a "Type" for some reason? (This is an exported field that is used in several places, even just within Syft itself.)

syft/distro/distro.go
luhring: Agreed — we should be consistent here, _especially_ since `name` is a real, separate thing.

syft/distro/distro.go
luhring: I can't think of a reason not to!

I was just refreshing myself on the [spec for os-release](https://www.linux.org/docs/man5/os-release.html). I don't know that we need to reflect _all_ fields, but this list of fields may be worth consulting as we thinking about the future of distro identification.

spiffcs:
Also general comments - I did some research into where Grype would break if these changes were merged or not thought through completely.

Since we changed the API of the type to not have a `String` method anymore to obtain the `type` then this breaks:
https://github.com/anchore/grype/blob/77b1da95fab47df2f69725f04c15454d89f3c76b/grype/matcher/common/distro_matchers.go#L49-L53

There is also deeper coupling in grype-db here:
https://github.com/anchore/grype-db/blob/72ff1b90af677519387ba475d223c9c3c5c94c68/pkg/db/v3/namespace.go#L39-L69

It's still an open question where we want the Strong list to exist as far as whitelisting which inputs are valid, but removing them from syft does seem to shift the onus onto grype or grype-db.

syft/distro/identify.go
luhring: Good point! It looks like it's no longer that "we failed to identify a thing" — we're just always surfacing what's there, which may be nothing

syft/distro/type.go
```diff
@@ -1,64 +0,0 @@
-package distro
-
-// Type represents the different Linux distribution options
-type Type string
-
-const (
```
luhring, line 6: I think these constants might still be useful, even if they aren't of type `Type` any more — it's nice to govern the expected string values by referencing a centrally defined value, rather than relying on the value (e.g. `"opensuse-leap"` to be specified correctly at every use. What are your thoughts?

syft/distro/identify.go
luhring: Thinking more on this — I'm even wondering what use we have for a "nillable" value to be returned for all of these related functions (that is, return type of `*Distro` and conditionally returning a `nil` value). What's the difference between that and a zero-value `Distro` struct?

luhring (commented):
Just did a preliminary pass to respond to your comments and take a look around at the changes. Nice work so far!

syft/distro/distro.go
spiffcs: We can punt this to the next PR when @wagoodman is back.

internal/spdxlicense/generate_license_list.go
spiffcs: removed - my local golang configuration updated this

syft/distro/type.go
spiffcs: Constants have now been kept in favor of a more gradual change to move the functionality to Grype.

syft/distro/distro.go
```diff
@@ -19,10 +19,12 @@ func NewDistro(t Type, ver, like string) (Distro, error) {
 	if ver == "" {
 		return Distro{Type: t}, nil
 	}
+
```
luhring, line 22: I ❤️ vertical separation for readability

syft/distro/distro.go
```diff
@@ -39,7 +41,7 @@ func (d Distro) Name() string {
 // MajorVersion returns the major version value from the pseudo-semantically versioned distro version value.
 func (d Distro) MajorVersion() string {
 	if d.Version == nil {
-		return fmt.Sprint("(version unknown)")
+		return "(version unknown)"
```
luhring, line 44: Nice catch

luhring (approved):
Nice work!

PR #501, opened 2021-09-13
Revert "Disable docker releases"

wagoodman:
Reverts anchore/syft#493

Testing the docker install process on a fork shows that this issue has been resolved. There is the larger issue of this occurring again in the future --it would be ideal to solve this by removing the dependency on `docker-practice/actions-setup-docker@v1` and docker desktop for mac in general. (Migrating to builds via podman + qemu?)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -94,15 +94,11 @@ jobs:
     needs: [quality-gate]
     runs-on: macos-latest # Due to our code signing process, it's vital that we run our release steps on macOS.
     steps:
+      - uses: docker-practice/actions-setup-docker@v1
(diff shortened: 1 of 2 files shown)


kzantow (approved):
I see the comment:

> we are having an infinite loop of installing docker

Is this not an issue any more? Fixed in `docker-practice/actions-setup-docker@v1`?

wagoodman:
It looks like it has https://github.com/docker-practice/actions-setup-docker/commit/51fd5009666ea5e3382fe125abf5d73d31386b59 a few days after the problem was noticed on our end. I tried something like this with a different brew formula commit without luck.

Issue #476, opened 2021-08-10
Adapt new and existing package metadata as SPDX relationships


wagoodman:

spiffcs:
I've been doing research on this since Friday afternoon and this AM. The first questions I have concerns this repository:
https://github.com/spdx/tools-golang

The library is hosted by the spdx organization and defines roughly the same model we do in `presenter/packagers/model/spdx22`.

If we want to express SPDX relationships for 2.2 correctly I think the first discussion should be around if we move to their model or keep our presenter model.

The tough part with this proposal is that their model is not designed as a presenter (no JSON tags).

They take a pretty interesting approach as far as marshalling/unmarshalling.
 
Check out their parsing code [here](https://github.com/spdx/tools-golang/blob/main/jsonloader/parser2v2/parse_annotations.go).

@wagoodman if you have time today can we talk a bit about the complexity of juggling our JSON model against the spdx tool model and then wrapping that into a "correct" presenter?

kzantow:
One thing of note: I think CycloneDX also has a way of specifying dependencies using a `bom-ref`: https://cyclonedx.org/use-cases/#dependency-graph so we would quite possibly want this handled somehow in our own model

spiffcs:
Added - https://github.com/anchore/syft/pull/507 as a starting point to build out the initial `ROOT` --> `Package` relationships

spiffcs:
https://github.com/anchore/syft/pull/507/files has been updated to now populate `Files` and include vertices between `Packages` and `Files` in the `Relationships` field

spiffcs:
Before we dig further into relationships it's probably worth tackling some of the prioritized bugs we have surrounding SPDX. I pulled in https://github.com/anchore/syft/issues/460 to make some progress in cleaning up our license section.

spiffcs:
@luhring The next steps for making this better is starting to dive on the architecture changes we talked about on Monday.

https://github.com/anchore/syft/issues/516

We'll work on refining this issue so we have a clear path to get syfts command API in the right place.

wagoodman:
I've broken out package relationships into https://github.com/anchore/syft/issues/572 such that package catalogers raise this information before presenters/formats can leverage them (such as SPDX).

PR #507, opened 2021-09-14
Populate Files and Relationship fields for spdx-json output

spiffcs:
Work for #476 

Initial Support for `RELATIONSHIPS` in the `spdx-json` output. 

This PR adds support for the following:
- populating the `files` field of the spdxJSON output
- associating spdx `Packages` with the new File resources in the `hasFiles` field.
- populating the `Relationship` field with the vertices associating Files and Packages

File support only comes from `catalogers` who support the `FileOwner` interface.

The current list is
- apk
- dpkg
- python
- rmdb

Link to JSON schema:
https://github.com/spdx/spdx-spec/blob/a08ffa4b883b35574ae0df8f4566cd031341ba3e/schemas/spdx-schema.json#L423-L558

Snippets of the new output generated with the `spdxJson` option -  pullfrom node:latest
```
 "files": [
  {
   "SPDXID": "SPDXRef-File-adduser-deluser.conf",
   "name": "deluser.conf",
   "licenseConcluded": "",
   "fileName": "/etc/deluser.conf"
  },
  {
   "SPDXID": "SPDXRef-File-adduser-adduser",
   "name": "adduser",
   "licenseConcluded": "",
   "fileName": "/usr/sbin/adduser"
  },
  {
   "SPDXID": "SPDXRef-File-adduser-deluser",
   "name": "deluser",
   "licenseConcluded": "",
   "fileName": "/usr/sbin/deluser"
  },
  {
   "SPDXID": "SPDXRef-File-adduser-adduser.conf",
   "name": "adduser.conf",
   "licenseConcluded": "",
   "fileName": "/usr/share/adduser/adduser.conf"
  },

```

```
   "hasFiles": [
    "SPDXRef-File-autotools-dev-dh_autotools-dev_restoreconfig",
    "SPDXRef-File-autotools-dev-dh_autotools-dev_updateconfig",
    "SPDXRef-File-autotools-dev-NEWS.Debian.gz",
    "SPDXRef-File-autotools-dev-README.Debian.gz",
    "SPDXRef-File-autotools-dev-TODO",
    "SPDXRef-File-autotools-dev-changelog.Debian.gz",
    "SPDXRef-File-autotools-dev-changelog.gz",
    "SPDXRef-File-autotools-dev-copyright",
    "SPDXRef-File-autotools-dev-upstream.mail.template",
    "SPDXRef-File-autotools-dev-dh_autotools-dev_restoreconfig.1.gz",
    "SPDXRef-File-autotools-dev-dh_autotools-dev_updateconfig.1.gz",
    "SPDXRef-File-autotools-dev-config.guess",
    "SPDXRef-File-autotools-dev-config.sub",
    "SPDXRef-File-autotools-dev-autotools_dev.pm"
   ],
```

```
 "relationships": [
  {
   "spdxElementId": "SPDXRef-Package-deb-adduser-3.118",
   "relationshipType": "CONTAINS",
   "relatedSpdxElement": "SPDXRef-File-adduser-deluser.conf"
  },
  {
   "spdxElementId": "SPDXRef-Package-deb-adduser-3.118",
   "relationshipType": "CONTAINS",
   "relatedSpdxElement": "SPDXRef-File-adduser-adduser"
  },
  {
   "spdxElementId": "SPDXRef-Package-deb-adduser-3.118",
   "relationshipType": "CONTAINS",
   "relatedSpdxElement": "SPDXRef-File-adduser-deluser"
  },
  {
   "spdxElementId": "SPDXRef-Package-deb-adduser-3.118",
   "relationshipType": "CONTAINS",
   "relatedSpdxElement": "SPDXRef-File-adduser-adduser.conf"
  },
  {
   "spdxElementId": "SPDXRef-Package-deb-adduser-3.118",
   "relationshipType": "CONTAINS",
   "relatedSpdxElement": "SPDXRef-File-adduser-TODO"
  },
```


seabass-labrax:
This is a very exciting addition to Syft! Keep up the great work, @spiffcs :)

internal/presenter/packages/model/spdx22/relationship.go
```diff
@@ -2,10 +2,12 @@ package spdx22
 
 type Relationship struct {
 	// SPDX ID for SpdxElement. A related SpdxElement.
-	RelatedSpdxElement string `json:"relatedSpdxElement"`
+	SpdxElementID string `json:"spdxElementId"`
 	// Describes the type of relationship between two SPDX elements.
 	RelationshipType RelationshipType `json:"relationshipType"`
-	Comment          string           `json:"comment,omitempty"`
+	// SPDX ID for SpdxElement.  A related SpdxElement.
```
luhring, line 8: Is this comment a duplicate of https://github.com/anchore/syft/pull/507/files#diff-c37457a5802e2e325dba628b81a3270da8709a215b74f1c421be73c8a60552edR4? Which field should it be used to describe?

internal/spdxlicense/generate_license_list.go
```diff
@@ -1,3 +1,4 @@
+//go:build ignore
```
luhring, line 1: I think you mentioned this line is getting added by accident. You might want to make a local adjustment to avoid accidentally checking this in 🤔

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -39,7 +39,8 @@ func (pres *SPDXJsonPresenter) Present(output io.Writer) error {
 	return enc.Encode(&doc)
 }
 
-func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx22.Document {
+// NewSPDXJsonDocument creates and populates a new JSON document struct that follows the SPDX 2.2 spec from the given cataloging results.
+func NewSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx22.Document {
```
luhring, line 43: Just curious: Why did this get exported?

wagoodman:
> Sub-packages should not be nested inside a Package Information section, 
but should be separate and should use a Relationship to clarify.

This is good information. I think today we have some catalogers that raise up transitive dependencies and others that only raise up direct dependencies. This implies a couple of things:
1. Adding a relationship to the doc ID for all existing packages that are cataloged today may not be correct --we need to only do this for only direct dependencies.
2. This hints at catalogers should be able to optionally raise up relationship information, to both support declaring when there is a "root" package and to pave the way for transitive dependencies.

internal/presenter/packages/spdx_json_presenter.go
spiffcs: We export NewJsonDocument as it is consumed by the poweruser package. It seemed incorrect to have one Document constructor exposed while the other was private. I also think it's something that should have an official Documentation string highlighted above the function since its counterpart would also end up in the godoc if generated.

internal/spdxlicense/generate_license_list.go
spiffcs: Yep - thought I removed this configuration but it seems like it's back with a vengence. I'll look at what magic nonsense exists within my defaults that is appending this.

internal/presenter/packages/model/spdx22/relationship.go
spiffcs: Yea I'll be more explicit here and use the direct json schema comment for both of these fields

internal/spdxlicense/generate_license_list.go
spiffcs: @luhring looks like my tooling is trying to migrate us to the current spec:
https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

spiffcs:
> > Sub-packages should not be nested inside a Package Information section,
> > but should be separate and should use a Relationship to clarify.
> 
> This is good information. I think today we have some catalogers that raise up transitive dependencies and others that only raise up direct dependencies. This implies a couple of things:
> 
> 1. Adding a relationship to the doc ID for all existing packages that are cataloged today may not be correct --we need to only do this for only direct dependencies.
> 2. This hints at catalogers should be able to optionally raise up relationship information, to both support declaring when there is a "root" package and to pave the way for transitive dependencies.

Yea this is the direction we should proceed. 

I can start tonight looking at how to make those changes as part of this PR. I don't want this going in if `Adding a relationship to the doc ID for all existing packages that are cataloged ` is incorrect if only SOME are direct. I'll do some work learning more about the cataloguer so I can find out which are smarter than the others around direct/transitive dependencies.

internal/presenter/packages/model/spdx22/relationship.go
spiffcs: updated to most recent specification comments:
https://github.com/spdx/spdx-spec/blob/3d156919a06c07781bd47dc26c75efcfc9df21cd/schemas/spdx-schema.json#L691-L718

spiffcs: Better explanation of the transition
https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives

I'm going to remove it for now, but will keep an eye on the transition going forward. I don't know if it's my `LSP` that is doing this or if it's some configuration I updated in my dotfiles when go1.17 dropped which has slipped my mind.

luhring: Sounds good. And thanks for the links — TIL!

luhring: Actually, I think we've crossed threads...

spiffcs: yea I think I missed the commit when I did a `rebase -i` to fix signature stuff. Addressing now - thanks for following up.

internal/presenter/packages/spdx_json_presenter.go
luhring: Gotcha, interesting! My thoughts, in no particular order:

- 100% agree re: doc comments 🌟 
- Architectural question: Do we plan on creating SPDX documents for any non-package commands?
- My pref in general is not to export something unless it's being used internally in another package, or we expect it to be used by a consumer of our module. In this case, it looks like this code is under `internal/` so I don't think it would be consumed by another module. Is that right?

spiffcs:
After talking with @wagoodman we're going to keep some parts of this branch. We'll table the `Package` problem for AFTER we bubble up the file relationships. I'm going to proceed with adding the `File` section of the document and then populating `Relationships` with the edge file relationships we already generate [here](https://github.com/anchore/syft/blob/9ec09add676a96c6d4f61cbe11415868be38c30c/syft/pkg/ownership_by_files_relationship.go)

cc @luhring for an update on progress/direction

internal/spdxlicense/generate_license_list.go
spiffcs: Better explanation of the transition
https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives

I'm going to remove it for now, but will keep an eye on the transition going forward. I don't know if it's my `LSP` that is doing this or if it's some configuration I updated in my dotfiles when go1.17 dropped which has slipped my mind.

internal/presenter/packages/model/spdx22/file.go
```diff
@@ -20,22 +20,22 @@ type File struct {
...
+	FileTypes []string `json:"fileTypes,omitempty"`
 	// This field provides a place for the SPDX file creator to record potential legal notices found in the file.
 	// This may or may not include copyright statements.
 	NoticeText string `json:"noticeText,omitempty"`
 	// Indicates the project in which the SpdxElement originated. Tools must preserve doap:homepage and doap:name
 	// properties and the URI (if one is known) of doap:Project resources that are values of this property. All other
 	// properties of doap:Projects are not directly supported by SPDX and may be dropped when translating to or
-	// from some SPDX formats.
-	ArtifactOf []string `json:"artifactOf"`
+	// from some SPDX formats(depricated).
```
luhring, line 39: nit: spelling: `depricated`

internal/presenter/packages/model/spdx22/relationship.go
luhring: I still think I'm not following what this comment is describing, sorry!

internal/presenter/packages/spdx_helpers.go
```diff
@@ -29,6 +30,52 @@ func getSPDXExternalRefs(p *pkg.Package) (externalRefs []spdx22.ExternalRef) {
...
+	files = make([]spdx22.File, 0)
+	fileIDs = make([]string, 0)
+	relationships = make([]spdx22.Relationship, 0)
+
+	pkgFileOwner, ok := p.Metadata.(pkg.FileOwner)
+	if !ok {
+		return files, fileIDs, relationships
+	}
+
+	for _, ownedFilePath := range pkgFileOwner.OwnedFiles() {
```
spiffcs, line 43: I just realized that the OwnedFiles is a full shotgun blast and has support from to exclude certain things.
https://github.com/anchore/syft/blob/9fe1da8ee60e128a5a61f406c0a03e178f4287d4/syft/pkg/ownership_by_files_relationship.go#L63-L67

I'm open to updating the signature so we can make the function smarter and have it only return a filtered list, but that means `matchesAny` gets repeated across implementations. Maybe there is a good middle ground here.

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -48,9 +49,12 @@ func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx
 		name = srcMetadata.Path
 	}
 
+	spdxDocumentID := spdx22.ElementID("DOCUMENT").String()
```
spiffcs, line 52: We're not using this anymore in the return. It was part of old relationship code. I'll restore it.

internal/presenter/packages/model/spdx22/relationship.go
spiffcs: Here is a link to the spec from their repo:
https://github.com/spdx/spdx-spec/blob/3d156919a06c07781bd47dc26c75efcfc9df21cd/schemas/spdx-schema.json#L697-L711

`spdxElementId`: "Id to which the SPDX element is related"
`relatedSpdxElement`: "SPDX ID for SpdxElement.  A related SpdxElement."

The rest of the model seems to follow these comments or the documentation closely so I was trying to stay in that theme. I do agree it is not the best description.

internal/presenter/packages/spdx_json_presenter.go
spiffcs: lowercased - if we start doing anything in `poweruser` with the function we can export it then.

internal/presenter/packages/model/spdx22/file.go
spiffcs: updated

internal/presenter/packages/model/spdx22/relationship.go
luhring: Ah I see. Got it, thanks! 🙏

internal/presenter/packages/spdx_helpers.go
spiffcs: @wagoodman Any opinions on where we want to go here? You have a deeper knowledge of Cataloger so I'll defer to you.

Do you think we should just repeat this logic in presenter or push the onus of `matchesAny` onto the respective `OwnedFiles` implementations?

internal/presenter/packages/spdx_json_presenter.go
spiffcs: removed

internal/presenter/packages/spdx_helpers.go
wagoodman: we went into a deeper answer on this offline; the short answer is this block doesn't apply to the relationship being added.

internal/presenter/packages/spdx_helpers.go
```diff
@@ -29,6 +30,52 @@ func getSPDXExternalRefs(p *pkg.Package) (externalRefs []spdx22.ExternalRef) {
...
+		baseFileName := filepath.Base(ownedFilePath)
+		fileSpdxID := spdx22.ElementID(fmt.Sprintf("File-%s-%s", p.Name, baseFileName)).String()
```
wagoodman, line 45: I'm not certain the base name will be good enough for an ID --what if there are two files with the same names in different paths for the same package? Could we use the whole path? or a hash of the path?

internal/presenter/packages/spdx_helpers.go
```diff
@@ -29,6 +30,52 @@ func getSPDXExternalRefs(p *pkg.Package) (externalRefs []spdx22.ExternalRef) {
...
+
+		fileIDs = append(fileIDs, fileSpdxID)
+
+		files = append(files, spdx22.File{
+			FileName:  ownedFilePath,
+			FileTypes: matchFileTypes(baseFileName),
```
wagoodman, line 51: since this function isn't implemented yet we could probably delete it and the FileTypes entry here (if the spec permits it). (would the basename be enough to lookup the filetype here? I'm not certain it would)

wagoodman (approved):
We tend to add information onto to common model that gets used by the presenters when possible. This PR is adding the relationship within a specific presenter --I think this is the only way to go to keep this PR small, but that does hint at future refactor that we can iterate on.

internal/presenter/packages/spdx_helpers.go
spiffcs: That's a good point - I can hash the full path and add it as a suffix to this `Sprintf`

internal/presenter/packages/spdx_helpers.go
spiffcs: The basename is not enough in this case for file detection. I'll remove it as it's not a required type.

PR #509, opened 2021-09-20
Extend license mapping for common SPDX license names 

### LINKED ISSUE #460: Some package licenses are not being reported in the SPDX format
**What happened**:
`syft ubuntu:latest -o json` reports that the adduser package has a license of `GPL-2` , but with the SPDX format the license is listed as NOASSERTION.

The SPDX output should have reported the correct SPDX identifier (`GPL-2.0`).

spiffcs:
Fixes #460

When running `syft ubuntu:latest -o spdx-json` on the current version of syft you'll see the following license output:
Screen Shot 2021-09-20 at 11 06 53 AM

SPDX will show a `NO ASSERTION` block even when a valid `GPL` license is present.

A valid license can be confirmed by running `syft ubuntu:latest -o json > test.json`:
Screen Shot 2021-09-20 at 11 12 26 AM

With the new code changes, a license mapping has been added where the stored debian license does not match its SPDX counterpart:
Screen Shot 2021-09-20 at 11 14 54 AM


internal/presenter/packages/spdx_helpers.go
```diff
@@ -87,6 +87,13 @@ func getSPDXLicense(p *pkg.Package) string {
 	for _, l := range p.Licenses {
 		if value, exists := spdxlicense.ID(l); exists {
 			parsedLicenses = append(parsedLicenses, value)
+			continue
+		}
+
+		// we did not find an explicit spdx license
+		// check congruent debian mappings
+		if value, exists := spdxlicense.DebianID(l); exists {
```
wagoodman, line 95: in this case we'd be potentially mapping the license as if it were debian related without verifying that it is debian related.

internal/spdxlicense/license.go
```diff
@@ -4,9 +4,25 @@ import (
 	"strings"
 )
 
+var debianLicenseLabels = map[string]string{
```
wagoodman, line 7: where was this list derived from? Are we certain that `gpl-2` means exactly 2.0 and not 2.1 (or more generically the latest v2 option)?

internal/spdxlicense/license.go
```diff
@@ -4,9 +4,25 @@ import (
...
+}
+
 //go:generate go run generate_license_list.go
 
 func ID(id string) (string, bool) {
 	value, exists := licenseIDs[strings.ToLower(id)]
 	return value, exists
 }
+
+func DebianID(id string) (string, bool) {
```
wagoodman, line 25: opinion: it seems a little awkward to expose something that is specific to debian here instead of encapsulating this decision.

spiffcs: Yea I agree - I'll workshop it a bit more and see if I can wrap this into the ID method rather than have something this specific.

internal/spdxlicense/license.go
spiffcs: Good point - I'll dig a bit more into why license are formatted as `GPL-X` on the ubuntu image and what that means as far as how it fits into latest license options.

internal/presenter/packages/spdx_helpers.go
spiffcs: Yea - this stems from the other comment about not having too specific a function. If I wrap this all into the `spdxlicense.ID` function where we first check the `go:generate` list, then we can just update the naming so that our second check isn't debian specific, but more of a check for incongruent mappings as they appear in the SPDX spec.

internal/spdxlicense/license.go
spiffcs: Here are the docs for the debian [spec](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-short-name).

"For SPDX compatibility, versions with trailing dot-zeroes are considered to be equivalent to versions without (e.g., “2.0.0” is considered equal to “2.0” and “2”)."

Based on that, we can be confident that `gpl-2` means exactly 2.0.

The 'latest' is covered where a `+` is added as a suffix to the short name.

The list was derived from checking generated list at `license_list.go`.

There are potentially more to add based on the confusion surrounding `short name` equivalence in the debian space.

Cases like `2` are not covered by our generated `license_list` from [here](https://spdx.org/licenses).

internal/presenter/packages/spdx_helpers.go
spiffcs: @wagoodman cleaned up the interface so that we are just using ID now. Map has been updated to known short name exceptions so it can be expanded to include other oddities that might be found in other disto.

internal/spdxlicense/license.go
spiffcs: Removed and expanded ID functionality to check a known list of short name exceptions that can be distro agnostic.

internal/spdxlicense/license.go
```diff
@@ -4,9 +4,26 @@ import (
 	"strings"
 )
 
+// https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-short-name
```
wagoodman, line 7: thanks for adding the link; mind adding a little more verbiage around what this is and why there are these exceptions?

internal/spdxlicense/license.go
wagoodman: It looks like there have been specific cases added for GPL and LGPL, is there a way we could process the license list to get this for all other license types with versions in them?

spiffcs:
@wagoodman on further investigation while trying to get the exceptions generated by our build code I found that there are a couple of deprecated licenses [here](https://spdx.github.io/license-list-data/)

Search `Deprecated License Identifiers`

Should we update our generator to filter deprecated license Identifiers?

If we remove the deprecated licenses then we can write generator code specific to the spdx guidelines here:
https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-field

Here is an example of what is left for gpl when we remove deprecated license
```
	"gpl-1.0-only":                         "GPL-1.0-only",
	"gpl-1.0-or-later":                     "GPL-1.0-or-later",
	"gpl-2.0-only":                         "GPL-2.0-only",
	"gpl-2.0-or-later":                     "GPL-2.0-or-later",
	"gpl-3.0-only":                         "GPL-3.0-only",
	"gpl-3.0-or-later":                     "GPL-3.0-or-later",
```

We can have it so that `(1, 1.0, 1.0.0)` ----> gpl-1.0-only. `(1+, 1.0+, 1.0.0+)` then match to `GPL-1.0-or-later`

I think it would be best to filter the deprecated license from `https://spdx.org/licenses/licenses.json` and add logic to the generation code that handles identification exceptions as we find them through the parsing process (starting with the debian guidelines we found).

internal/spdxlicense/license.go
```diff
@@ -1,12 +1,38 @@
...
+// If a license generated in license_list.go is not found when looking up by ID,
+// then the ID function will use these regular expressions to help resolve cases where
+// x.0.0 and x are supplied as version numbers. For SPDX compatibility, versions with trailing
+// dot-zeroes are considered to be equivalent to versions without (e.g., “2.0.0” is considered equal to “2.0” and “2”).
+// EX: gpl-2+ ---> GPL-2.0+
+// EX: gpl-2.0.0-only ---> GPL-2.0-only
+// See the debian link for more details on the spdx license differences
+var (
+	zero   = regexp.MustCompile(`^((.*).0)(.*)$`)
+	noZero = regexp.MustCompile(`^(.*-)([1-9])(.*)`)
```
kzantow, line 18: I don't think it's a _huge_ problem for this at all, but the first regex will match things like `GPL-2.0.1`, later becoming `GPL-2.1` maybe the first regex should be `^(.*\.0)\.0(.*)$` and do `idBytes = zero.ReplaceAll([]byte(lowerID), []byte("${1}${2}"))`? and the second would match `GPL-2.0.1-stuff` and later become `GPL-2.0.1.0-stuff`, maybe use `^(.*\-\d)([^.1-9]+)$`? and later `noZero.ReplaceAll([]byte(lowerID), []byte("${1}.0${2}"))`?

spiffcs: Nice! I've update these to be more exclusive. I think the only spot where it's a little funky now is the `noZero` case.

`gpl-2` is not matched on anymore which causes the ReplaceAll to not fire just leaving us with `gpl-2` and no trailing `.0`

internal/spdxlicense/license_test.go
```diff
@@ -0,0 +1,44 @@
...
+			"cc-by-nc-3-de",
+			"CC-BY-NC-3.0-DE",
+		},
+	}
+
+	for _, test := range tests {
+		got, exists := ID(test.shortName)
+		if exists {
+			assert.Equal(t, got, test.spdx)
+			return
```
wagoodman, line 40: this return short circuits the tests (only one passing case is tested). With `t.Run()` the return would work out alright, but there is some room for testify to make this look even simpler:
```golang
		t.Run(test.shortName, func(t *testing.T) {
			got, exists := ID(test.shortName)
			assert.True(t, exists)
			assert.Equal(t, test.spdx, got)
		})
```

spiffcs: Updated - realized this when testing suggestions from @kzantow

internal/spdxlicense/license.go
```diff
@@ -1,12 +1,38 @@
...
+)
+
 //go:generate go run generate_license_list.go
 
 func ID(id string) (string, bool) {
```
wagoodman, line 23: this approach is changing the input license at runtime, but we have an opportunity to generate all possible answers in the lookup at build time (over in `generate_license_list.go`)

internal/spdxlicense/license_test.go
spiffcs: simplified with newest commit

internal/spdxlicense/license.go
spiffcs: If we move if over into generation should we be explicit in the capture groups for only debian short license:
https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-short-name

Or do we want our map to just add cases `x-1` and `x-1.0.0`

How high do we want the version numbers to go in the generation?

spiffcs: Since we are iterating over this list I think it's pretty simple to add `2` extra key given the example
```
	"gpl-1.0-only":                         "GPL-1.0-only",
	"gpl-1.0-or-later":                     "GPL-1.0-or-later",
	"gpl-2.0":                              "GPL-2.0",
	"gpl-2.0+":                             "GPL-2.0+",
	"gpl-2.0-only":                         "GPL-2.0-only",
	"gpl-2.0-or-later":                     "GPL-2.0-or-later",
	"gpl-2.0-with-autoconf-exception":      "GPL-2.0-with-autoconf-exception",
	"gpl-2.0-with-bison-exception":         "GPL-2.0-with-bison-exception",
	"gpl-2.0-with-classpath-exception":     "GPL-2.0-with-classpath-exception",
	"gpl-2.0-with-font-exception":          "GPL-2.0-with-font-exception",
	"gpl-2.0-with-gcc-exception":           "GPL-2.0-with-GCC-exception",
	"gpl-3.0":                              "GPL-3.0",
	"gpl-3.0+":                             "GPL-3.0+",
	"gpl-3.0-only":                         "GPL-3.0-only",
	"gpl-3.0-or-later":                     "GPL-3.0-or-later",
	"gpl-3.0-with-autoconf-exception":      "GPL-3.0-with-autoconf-exception",
	"gpl-3.0-with-gcc-exception":           "GPL-3.0-with-GCC-exception",
```

So `gpl-3.0-only` becomes `gpl-3.0-only`, `gpl-3-only` `gpl-3.0.0-only` and they all point to the same `GPL-3.0-only`

spiffcs: @wagoodman bumped the code to generate the map at build time with entries. 

Tests I wrote for runtime still work which says to me 👍

internal/spdxlicense/license_list.go
```diff
@@ -14,42 +14,80 @@ var licenseIDs = map[string]string{
...
 	"barr":                                 "Barr",
 	"beerware":                             "Beerware",
+	"bittorrent-1":                         "BitTorrent-1.0",
 	"bittorrent-1.0":                       "BitTorrent-1.0",
+	"bittorrent-1.0.0":                     "BitTorrent-1.0",
 	"bittorrent-1.1":                       "BitTorrent-1.1",
 	"blessing":                             "blessing",
+	"blueoak-1.0":                          "BlueOak-1.0.0",
 	"blueoak-1.0.0":                        "BlueOak-1.0.0",
+	"blueoak-1.0.0.0":                      "BlueOak-1.0.0",
```
spiffcs, line 90: Technically correct, but weird

internal/spdxlicense/license_list.go
```diff
@@ -308,67 +560,118 @@ var licenseIDs = map[string]string{
...
 	"oldap-1.2":                            "OLDAP-1.2",
 	"oldap-1.3":                            "OLDAP-1.3",
 	"oldap-1.4":                            "OLDAP-1.4",
+	"oldap-2":                              "OLDAP-2.0",
 	"oldap-2.0":                            "OLDAP-2.0",
+	"oldap-2.0.0":                          "OLDAP-2.0",
+	"oldap-2.0.0.1":                        "OLDAP-2.0.1",
 	"oldap-2.0.1":                          "OLDAP-2.0.1",
-	"oldap-2.1":                            "OLDAP-2.1",
+	"oldap-2.1":                            "OLDAP-2.0.1",
```
spiffcs, line 674: also incorrect here

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -91,3 +92,81 @@ func main() {
 		log.Fatalf("unable to generate template: %+v", err)
 	}
 }
+
+func processSPDXLicense(result LicenseList) map[string]string {
+	// first pass build map
```
wagoodman, line 97: probably worth beefing up the comment since the two-pass approach is very intentional

wagoodman (commented):
I think it's worth capturing some odd cases in tests that are commented or in the description for this PR (e.g. `w3c-20150513.0.0` vs `w3c-20150513`)

wagoodman (approved):
Nice! This was a subtly tricky one. I added a couple of comments about code comments, but otherwise looks great 👍

internal/spdxlicense/generate/generate_license_list.go
spiffcs: comment added

spiffcs:
@wagoodman comments and test cases added to aid future readers

wagoodman (approved):

spiffcs:
Thanks for the addition - I'll merge this when the checks complete.

internal/spdxlicense/license_test.go
```diff
@@ -35,6 +35,7 @@ func TestIDParse(t *testing.T) {
 			"CC-by-nc-3-de",
 			"CC-BY-NC-3.0-DE",
 		},
+                # the below few cases are NOT expected, however, seem unavoidable given the current approach
```
spiffcs, line 38: @wagoodman you might have another language on the brain. I think this should be `//`

wagoodman: I like to throw in breaking code into other folks PRs as a new hobby of mine...  couldn't get past you!

PR #511, opened 2021-09-21
lower log file permissions to 0644

spiffcs:
Removes permissive file creation and sets default to 0644


kzantow (approved):
Very good change :)

luhring (approved):
LGTM 👍

PR #512, opened 2021-09-22
Use Anchore fork of packageurl lib without replace directive

luhring:
As [discussed in the community Slack](https://anchorecommunity.slack.com/archives/C0190NF9TSM/p1632315222000900), using a `replace` directive in a module's `go.mod` means that projects like Syft will build differently themselves than when imported by consumers and then built. This could mean that consumers of Syft experience unexpected behavior, or even potentially broken builds, since they're using a state different than we're testing in Syft.

This PR follows up https://github.com/anchore/packageurl-go/pull/2 by removing the `replace` directive from Syft's `go.mod`, thereby pivoting Syft to rely more directly on Anchore's fork.

### DIFF
--- a/.bouncer.yaml
+++ b/.bouncer.yaml
@@ -6,7 +6,7 @@ permit:
   - ISC
 ignore-packages:
   # packageurl-go is released under the MIT license located in the root of the repo at /mit.LICENSE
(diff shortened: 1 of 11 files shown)


kzantow (approved):
LGTM

Issue #497, opened 2021-09-09
SPDX output is not consistently sorted


kzantow:
**What happened**:
`syft -o spdx ...` and `syft -o spdx-json ...` produce differently sorted output across different runs.

**What you expected to happen**:
The output is sorted consistently.

**How to reproduce it (as minimally and precisely as possible)**:
Run `syft -o spdx` multiple times and `diff` the results, these should not show lines that are in different order.

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: 0.21.0
- OS (e.g: `cat /etc/os-release` or similar):

spiffcs:
Just an update here - it's not the sorting that is the issue:

`diff -u /tmp/alpine.json:2 /tmp/alpine.json:latest` where both of those documents are the same run shows inconsistencies in CPE generation.

Specifically, sometimes a cpe will use `_` while in other runs it will use `_`.

After this is fixed I'll double check the sorting portion to be sure the output is consistently sorted, but the primary source of the `diff` pain comes from this weird flip of character between `-` and `_`

json-document
![Screen Shot 2021-09-30 at 11 47 39 AM](https://user-images.githubusercontent.com/32073428/135488884-ae64a338-ae99-4c39-bd66-24a245757f1c.png)

spdx
![Screen Shot 2021-09-30 at 11 36 24 AM](https://user-images.githubusercontent.com/32073428/135488914-41e9a4e8-5426-4d71-891f-8103dd492443.png)

seabass-labrax:
This is great :grinning: Reproducible builds with Syft-generated SPDX documents coming soon, in CI pipelines near you!

spiffcs:
Just adding a regression test in a separate PR to make sure we keep it that way!

spiffcs:
Opening until we close the issue with the regression test.

PR #522, opened 2021-09-30
497 stable sorted CPE array (JSON and SPDX)

spiffcs:
## Attempts to Solve #497 

To see the current issue you can:
```
syft -o json alpine:latest > /tmp/alpine.json:latest
syft -o json alpine:latest > /tmp/alpine.json:latest:2 
diff -u /tmp/alpine.json:2 /tmp/alpine.json:latest
```

This will produce a diff where the `cpes` field for some packages is sorted differently from run to run.

In the case where `countFieldLength` is equal for two sequential CPE in a given list of CPE, then our sort function would return a non deterministic order for cases as seen below:
```
cpe:2.3:a:alpine-keys:alpine_keys:2.3-r1:*:*:*:*:*:*:*
cpe:2.3:a:alpine-keys:alpine-keys:2.3-r1:*:*:*:*:*:*:*
cpe:2.3:a:alpine_keys:alpine_keys:2.3-r1:*:*:*:*:*:*:*
cpe:2.3:a:alpine_keys:alpine-keys:2.3-r1:*:*:*:*:*:*:*
cpe:2.3:a:alpine:alpine_keys:2.3-r1:*:*:*:*:*:*:*
cpe:2.3:a:alpine:alpine-keys:2.3-r1:*:*:*:*:*:*:*
```

```
// If both Less(i, j) and Less(j, i) are false,
// then the elements at index i and j are considered equal.
// Sort may place equal elements in any order in the final result,
// while Stable preserves the original input order of equal elements.
```

If the below change is too specific or fragile I can explore using the `Stable` interface to see if that at least keeps us consistent across runs.

The array order coming into the sort is always consistent, but where we get the `diff` between runs is when ` Less(i, j) and Less(j, i) are false,`.

This PR assigns a priority based on the character `-` for the `Vendor` and `Product` fields.
Given the above example the ideal sorted output would be:
```
xxx-xxx-xxx
xxx-xxx_xxx
xxx_xxx-xxx
xxx_xxx_xxx
xx-xx
xx_xx
```

Currently I'm trying to find images which break the assumptions found in this code.

### DIFF
--- a/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
+++ b/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
@@ -18,16 +18,18 @@ func (c BySpecificity) Less(i, j int) bool {
 	iScore := weightedCountForSpecifiedFields(c[i])
 	jScore := weightedCountForSpecifiedFields(c[j])
 
-	if iScore == jScore {
-		if countFieldLength(c[i]) == countFieldLength(c[j]) {
-			// we want this to be < than since we want
-			// - to come before _ in vendor:product
-			return c[i].BindToFmtString() < c[j].BindToFmtString()
-		}
+	// check weighted sort first
+	if iScore != jScore {
+		return iScore > jScore
+	}
 
+	// sort longer fields to top
+	if countFieldLength(c[i]) != countFieldLength(c[j]) {
 		return countFieldLength(c[i]) > countFieldLength(c[j])
 	}
-	return iScore > jScore
+
+	// if score and length are equal then text sort
+	return c[i].BindToFmtString() < c[j].BindToFmtString()
 }
 
 func countFieldLength(cpe wfn.Attributes) int {
--- a/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
+++ b/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
@@ -30,6 +30,11 @@ func (c BySpecificity) Less(i, j int) bool {
 	return iScore > jScore
 }
 
+// if countFieldLength is equal for adjacent members
+// Less has the option of randomly ordering the items
+// dashIndex is used to stabilize the sort so that Vendor
+// and Product combinations are given a stable hierarchy
+// if their countFieldLength sums are equal.
 func dashIndex(cpe wfn.Attributes) int {
 	count := 0
 	dash := "-"
--- a/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
+++ b/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
@@ -2,6 +2,7 @@ package cpe
 
 import (
 	"sort"
+	"strings"
 
 	"github.com/facebookincubator/nvdtools/wfn"
 )
@@ -19,9 +20,28 @@ func (c BySpecificity) Less(i, j int) bool {
 	jScore := weightedCountForSpecifiedFields(c[j])
 
 	if iScore == jScore {
+		if countFieldLength(c[i]) == countFieldLength(c[j]) {
+			// lower index count prefered
+			return dashIndex(c[i]) > dashIndex(c[j])
+		}
+
 		return countFieldLength(c[i]) > countFieldLength(c[j])
 	}
 	return iScore > jScore
+}
+
+func dashIndex(cpe wfn.Attributes) int {
+	count := 0
+	dash := "-"
+	if strings.Contains(cpe.Vendor, dash) {
+		count += 2
+	}
+
+	if strings.Contains(cpe.Product, dash) {
+		count += 1
+	}
+
+	return count
 }
 
 func countFieldLength(cpe wfn.Attributes) int {
--- a/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
+++ b/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
@@ -34,11 +34,11 @@ func dashIndex(cpe wfn.Attributes) int {
 	count := 0
 	dash := "-"
 	if strings.Contains(cpe.Vendor, dash) {
-		count = count + 2
+		count += 2
 	}
 
 	if strings.Contains(cpe.Product, dash) {
-		count = count + 1
+		count++
 	}
 
 	return count
--- a/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
+++ b/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
@@ -21,7 +21,7 @@ func (c BySpecificity) Less(i, j int) bool {
 
 	if iScore == jScore {
 		if countFieldLength(c[i]) == countFieldLength(c[j]) {
-			// lower index count prefered
+			// lower index count preferred
 			return dashIndex(c[i]) > dashIndex(c[j])
 		}
 
@@ -34,11 +34,11 @@ func dashIndex(cpe wfn.Attributes) int {
 	count := 0
 	dash := "-"
 	if strings.Contains(cpe.Vendor, dash) {
-		count += 2
+		count = count + 2
 	}
 
 	if strings.Contains(cpe.Product, dash) {
-		count += 1
+		count = count + 1
 	}
 
 	return count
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/common/cpe/sort_by_specificity.go
```diff
@@ -19,11 +20,30 @@ func (c BySpecificity) Less(i, j int) bool {
...
+		}
+
 		return countFieldLength(c[i]) > countFieldLength(c[j])
 	}
 	return iScore > jScore
 }
 
+func dashIndex(cpe wfn.Attributes) int {
+	count := 0
+	dash := "-"
```
luhring, line 35: nit: This could be a `const` since we never need to assign a new value to it.

luhring (commented):
Question: If the objective of this PR is to ensure that SPDX output is consistently sorted on the whole, is there a test we can introduce to make that assertion, beyond CPEs?

syft/pkg/cataloger/common/cpe/sort_by_specificity.go
spiffcs: I need to rename the PR. The bug was `sorted spdx output`, but that was not actually the issue. The issue was that the diffs found when running multiple of the same image were located in the `cpes` array.

spiffcs: Updated to: stable sorted CPE array (JSON and SPDX)

spiffcs:
> Question: If the objective of this PR is to ensure that SPDX output is consistently sorted on the whole, is there a test we can introduce to make that assertion, beyond CPEs?

Yea let me also see if I can add an integration level test that asserts there are no regressions on this front as far as producing a consistent document across multiple runs.

syft/pkg/cataloger/common/cpe/sort_by_specificity.go
```diff
@@ -19,11 +20,30 @@ func (c BySpecificity) Less(i, j int) bool {
...
+			// lower index count preferred
+			return dashIndex(c[i]) > dashIndex(c[j])
+		}
+
 		return countFieldLength(c[i]) > countFieldLength(c[j])
 	}
 	return iScore > jScore
 }
 
+func dashIndex(cpe wfn.Attributes) int {
```
wagoodman, line 33: A comment here as to why this is here would go a long way --It isn't obvious that this is here purely to stabilize the sort

spiffcs: comment added here. I'm also open to talk about this being a pretty fragile assumption. It was a patch that was very target at the alpine example I listed in the PR. I'm still running some other images to see where/how this might break and if I can find a larger and more general solution.

wagoodman: agreed --it's always worth a little extra time up front to attempt and avoid solving the same problem twice!

kzantow: I'm just mostly confused why are we not just text sorting? We could sort based on each field split based on `:` or somesuch

spiffcs: yea there was some conversation on using 
```
return c[i].BindToFmtString() > c[j].BindToFmtString()
```

I'll check that rather than sorting based on random `-`

spiffcs:
@kzantow I added text sort as the base case if our `iScore` or `fieldLength` are both equal. I also prioritized it so it goes
```
weightedScore > (prioritize higher weight to front)
length > (prioritize longer length to front)
text < (reverse here so that - shows up in front of _)
```

spiffcs:
With the new stable sort here is a screenshot that shoes `spdx-json` is now stable for `alpine:latest`
Screen Shot 2021-09-30 at 5 21 40 PM

spiffcs:
@anchore/tools There are two places I found as candidates for regression test additions for this PR.

This first would be in the `snapshot` testing we have under the presenter package. Here we can add a more complex combination of CPE to the catalog so that every time the presenter generates we can compare to the old snapshot and say 👍 the array did not change.

https://github.com/anchore/syft/blob/6480f061734f6f7b1f3545f225851a8ae0e37c3d/internal/presenter/packages/json_presenter_test.go#L29-L37
https://github.com/anchore/syft/blob/6480f061734f6f7b1f3545f225851a8ae0e37c3d/internal/presenter/packages/utils_test.go#L106-L108

The other is we write a new integration test under `test/integration` and generate two sbom using `alpine:latest`, filter those sbom for unique values (Timestamp, ID, etc) and then compare.

I'm happy to add either, just wanted your thoughts going into Friday on if one was good enough or if we want both levels checked.

luhring:
> There are two places I found as candidates for regression test additions for this PR. [...]

Great ideas. I'm liking the second route a bit more at the moment, but I could be convinced of either direction. My initial thinking is if we can avoid the fragility that sometimes comes with snapshot testing, these tests would be more reliable and thus more trusted by the team.

luhring:
Re: the PR's purpose...

>I need to rename the PR. The bug was sorted spdx output, but that was not actually the issue. The issue was that the diffs found when running multiple of the same image were located in the cpes array.

My thinking is this: Right now, merging this PR will cause #497 to be closed. 497 is expressing that the SPDX output's sorting should be stable in general (which I agree with). So if we want to consider 497 done, I think we should have test(s) that ensure 497 is fully addressed. If we don't want to tackle all of 497 in this particular PR, I'm cool with that, too! In that case, 497 would still be unfinished (IMHO) upon merging this PR.

Curious for your latest thinking!

spiffcs:
> . So if we want to consider 497 done, I think we should have test(s) that ensure 497 is fully addressed. If we don't want to tackle all of 497 in this particular PR, I'm cool with that, too! In that case, 497 would still be unfinished (IMHO) upon merging this PR.
> 
> Curious for your latest thinking!

If we can get an integration test added that shows two runs of `-o spdx-json` consistently sorted then I think we can close #497. I've run this fix against a couple images and have not seen diffs that show unstable sorts baring the CPE issue.

wagoodman:
re: snapshot testing: I agree with @luhring  on this one --snapshot testing is great for change detection and seeing what changed, but is not great at communicating why something changed. Leaning more on unit tests for depth in cases and specific business logic verifications is the way to go here. 

If we have a concern about how things are wired up, maybe an integration or CLI level test would be useful here --generate an SBOM for the same input a few times and ensure the output is stable as a whole (like you suggested @spiffcs ). This isn't required for this PR, but could also be helpful to make certain we aren't regressing in SBOM reproducibility.

spiffcs:
> If we have a concern about how things are wired up, maybe an integration or CLI level test would be useful here --generate an SBOM for the same input a few times and ensure the output is stable as a whole (like you suggested @spiffcs ). Though it doesn't need to be limited to spdx-json, but testing all SBOM outputs (`for each output format ... for 1-5... run syft` then compare all results within each format). This isn't required for this PR, but could also be helpful to make certain we aren't regressing in SBOM reproducibility.

I think we if add this in this PR we can call #497 closed since we addressed the root of the problem and add regression. I'll write it up today.

syft/pkg/cataloger/common/cpe/sort_by_specificity_test.go
```diff
@@ -80,6 +80,25 @@ func TestCPESpecificity(t *testing.T) {
...
 		},
+		{
+			name: "sort by mix of field length, specificity, dash",
+			input: []pkg.CPE{
+				mustCPE("cpe:2.3:a:alpine-keys:alpine_keys:2.3-r1:*:*:*:*:*:*:*"),
+				mustCPE("cpe:2.3:a:alpine-keys:alpine-keys:2.3-r1:*:*:*:*:*:*:*"),
+				mustCPE("cpe:2.3:a:alpine_keys:alpine_keys:2.3-r1:*:*:*:*:*:*:*"),
+				mustCPE("cpe:2.3:a:alpine_keys:alpine-keys:2.3-r1:*:*:*:*:*:*:*"),
+				mustCPE("cpe:2.3:a:alpine:alpine_keys:2.3-r1:*:*:*:*:*:*:*"),
+				mustCPE("cpe:2.3:a:alpine:alpine-keys:2.3-r1:*:*:*:*:*:*:*"),
```
kzantow, line 91: maybe make this obviously more out of order?

kzantow (approved):
To me, this looks good -- super simple and I think it should work for solving the original issue 👍

PR #526, opened 2021-10-04
Make the build run on Fedora

joshbressers:
Fedora uses podman instead of docker for running containers. There are a few differences with podman that caused the build to fail.

### DIFF
--- a/test/integration/test-fixtures/image-large-apk-data/Dockerfile
+++ b/test/integration/test-fixtures/image-large-apk-data/Dockerfile
@@ -1,5 +1,2 @@
 FROM alpine@sha256:d9a7354e3845ea8466bb00b22224d9116b183e594527fb5b6c3d30bc01a20378
-RUN apk add --no-cache \
-            tzdata=2021a-r0 \
-            vim=8.2.2320-r0 \
-            alpine-sdk=1.0-r0
+RUN apk add --no-cache tzdata vim alpine-sdk
--- a/syft/pkg/cataloger/java/archive_parser_test.go
+++ b/syft/pkg/cataloger/java/archive_parser_test.go
@@ -90,6 +90,7 @@ func TestParseJar(t *testing.T) {
 			fixture: "test-fixtures/java-builds/packages/example-jenkins-plugin.hpi",
 			ignoreExtras: []string{
 				"Plugin-Version", // has dynamic date
+				"Built-By", // podman returns the real UID
 				"Build-Jdk",      // can't guarantee the JDK used at build time
 			},
 			expected: map[string]pkg.Package{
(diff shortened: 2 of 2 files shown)


test/integration/test-fixtures/image-large-apk-data/Dockerfile
```diff
@@ -1,5 +1,2 @@
 FROM alpine@sha256:d9a7354e3845ea8466bb00b22224d9116b183e594527fb5b6c3d30bc01a20378
-RUN apk add --no-cache \
-            tzdata=2021a-r0 \
-            vim=8.2.2320-r0 \
-            alpine-sdk=1.0-r0
+RUN apk add --no-cache tzdata vim alpine-sdk
```
luhring, line 2: We pin to dependency versions intentionally for reproducibility and to avoid surprise test failures.

What was the exact issue caused by those pinned versions? Perhaps there are newer versions that can be pinned to solve things?

joshbressers:
Here is the error I'm seeing both on Debian and Fedora. I'll update the PR with pinned versions
```
=== RUN   TestRegression212ApkBufferSize
Sending build context to Docker daemon  2.048kB
Step 1/2 : FROM alpine@sha256:d9a7354e3845ea8466bb00b22224d9116b183e594527fb5b6c3d30bc01a20378
 ---> 7731472c3f2a
Step 2/2 : RUN apk add --no-cache             tzdata=2021a-r0             vim=8.2.2320-r0             alpine-sdk=1.0-r0
 ---> Running in cba139f29b43
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/community/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  tzdata-2021c-r0:
    breaks: world[tzdata=2021a-r0]
  vim-8.2.3437-r0:
    breaks: world[vim=8.2.2320-r0]
The command '/bin/sh -c apk add --no-cache             tzdata=2021a-r0             vim=8.2.2320-r0             alpine-sdk=1.0-r0' returned a non-zero code: 2
    image_fixtures.go:156: could not build fixture image: exit status 2
```

joshbressers:
@luhring the new patch only updates the pinned versions instead of removing the versions

luhring:
@joshbressers Thanks for the update on versions!

It looks like just one commit (https://github.com/anchore/syft/pull/526/commits/6e9689491d2fad0c0733e9b5d88c886f4658244c) didn't get the "sign-off" appended to the commit, and so the DCO check is failing. You can either remove this commit via a rebase or retroactively sign it off.

Also, we require that all commits are signed now (e.g. w/ GPG), see here for details and let me know if I can help with this: https://github.com/anchore/syft/blob/main/CONTRIBUTING.md#sign-your-commits

joshbressers:
Nice on the signed commits!

I'll just delete this PR and open a new one to hide my shame and put all the bits in proper order :)

Thanks

Issue #495, opened 2021-09-08
Missing/incorrect SPDX fields: DocumentName, DocumentNamespace


kzantow:
**What happened**:
The SPDX spec [requires the DocumentName field](https://spdx.github.io/spdx-spec/document-creation-information/#64-document-name-field), which is not present.

Related: the namespace is the _same_ for all documents (e.g. `https://anchore.com/syft/image/`), which currently [goes against the spec indicating it should be uniquely identifiable](https://spdx.github.io/spdx-spec/document-creation-information/#65-spdx-document-namespace-field). Presumably the DocumentName would also be used as part of the DocumentNamespace.

NOTE: this is important in order to properly cross-reference SPDX files using the [relationship](https://spdx.github.io/spdx-spec/relationships-between-SPDX-elements/) field.

ALSO NOTE: this happens when using a _directory_ scan but this information is included if scanning a container image, however I do not believe the DocumentNamespace is correct, it does not have a UUID.

**What you expected to happen**:
A unique DocumentNamespace is generated and somehow a DocumentName is included or allowed to be specified from the command line.

**How to reproduce it (as minimally and precisely as possible)**:
`syft packages dir:. -o spdx`

**Anything else we need to know?**:
Pertinent directory scan output snippet:
```
SPDXVersion: SPDX-2.2
DataLicense: CC0-1.0
SPDXID: SPDXRef-DOCUMENT
DocumentNamespace: https://anchore.com/syft/image/
LicenseListVersion: 3.14
Creator: Organization: Anchore, Inc
Creator: Tool: syft-0.21.0
Created: 2021-09-08T01:45:39Z

##### Package: chownr
...
```

Image scan output snippet:
```
SPDXVersion: SPDX-2.2
DataLicense: CC0-1.0
SPDXID: SPDXRef-DOCUMENT
DocumentName: alpine:latest
DocumentNamespace: https://anchore.com/syft/image/alpine:latest
LicenseListVersion: 3.14
Creator: Organization: Anchore, Inc
Creator: Tool: syft-0.21.0
Created: 2021-09-08T16:40:41Z

##### Package: alpine-baselayout
...
```

**Environment**:
- Output of `syft version`: 0.21.0
- OS (e.g: `cat /etc/os-release` or similar):

spiffcs:
[...]

PR #528, opened 2021-10-04
(#495) Update `documentNamespace` uniqueness for spdx-json output

spiffcs:
Fixes: #495 

In the case where an input directory is `.` or `./` the name is stripped and a `uuidv4` becomes the namespace identifier.

When a URL is an input here is an example of the format output:
```
"documentNamespace": "https:/anchore.com/syft/image/gcr.io/distroless/nodejs-debian10-6d671418-314c-480b-a620-c86d985d5441",
```

Since `.` is unreserved I believe we can get away with keeping the baseURL for image inputs in the path:
https://datatracker.ietf.org/doc/html/rfc3986#section-2.3

### DIFF
--- a/internal/presenter/packages/spdx_json_presenter.go
+++ b/internal/presenter/packages/spdx_json_presenter.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"strings"
 	"time"
 
 	"github.com/anchore/syft/internal"
@@ -16,7 +15,7 @@ import (
 	"github.com/google/uuid"
 )
 
-const anchoreNamespace = "https://anchore.com/syft/image"
+const anchoreNamespace = "https://anchore.com/syft"
 
 // SPDXJsonPresenter is a SPDX presentation object for the syft results (see https://github.com/spdx/spdx-spec)
 type SPDXJsonPresenter struct {
@@ -51,10 +50,10 @@ func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx
 	switch srcMetadata.Scheme {
 	case source.ImageScheme:
 		name = srcMetadata.ImageMetadata.UserInput
-		identifier = strings.Trim(fmt.Sprintf("%s-%s", name, uniqueID.String()), "-")
+		identifier = fmt.Sprintf("image/%s", uniqueID.String())
 	case source.DirectoryScheme:
 		name = srcMetadata.Path
-		identifier = uniqueID.String()
+		identifier = fmt.Sprintf("dir/%s", uniqueID.String())
 	}
 
 	namespace := fmt.Sprintf("%s/%s", anchoreNamespace, identifier)
--- a/internal/presenter/packages/spdx_json_presenter.go
+++ b/internal/presenter/packages/spdx_json_presenter.go
@@ -58,7 +58,7 @@ func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx
 		if name != "." {
 			identifier = path.Join("dir", fmt.Sprintf("%s-%s", name, uniqueID.String()))
 		} else {
-			identifier = path.Join("dir", fmt.Sprintf("%s", uniqueID.String()))
+			identifier = path.Join("dir", uniqueID.String())
 		}
 	}
 
--- a/internal/presenter/packages/spdx_json_presenter.go
+++ b/internal/presenter/packages/spdx_json_presenter.go
@@ -52,10 +52,16 @@ func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx
 	switch srcMetadata.Scheme {
 	case source.ImageScheme:
 		name = cleanSPDXName(srcMetadata.ImageMetadata.UserInput)
-		identifier = path.Join("image", fmt.Sprintf("%s-%s", name, uniqueID.String()))
+		if name != "" {
+			identifier = path.Join("image", fmt.Sprintf("%s-%s", name, uniqueID.String()))
+		}
+		identifier = path.Join("image", fmt.Sprintf("%s", uniqueID.String()))
 	case source.DirectoryScheme:
 		name = cleanSPDXName(srcMetadata.Path)
-		identifier = path.Join("dir", fmt.Sprintf("%s-%s", name, uniqueID.String()))
+		if name != "" {
+			identifier = path.Join("dir", fmt.Sprintf("%s-%s", name, uniqueID.String()))
+		}
+		identifier = path.Join("dir", fmt.Sprintf("%s", uniqueID.String()))
 	}
 
 	namespace := path.Join(anchoreNamespace, identifier)
@@ -130,6 +136,10 @@ func cleanSPDXName(name string) string {
 
 	// remove : for url construction
 	name = strings.Replace(name, ":", "-", -1)
+
+	// trim bad characters
+	name = strings.Trim(name, "/")
+	name = strings.Trim(name, ".")
 
 	// clean relative pathing
 	return path.Clean(name)
--- a/internal/presenter/packages/spdx_json_presenter.go
+++ b/internal/presenter/packages/spdx_json_presenter.go
@@ -52,16 +52,10 @@ func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx
 	switch srcMetadata.Scheme {
 	case source.ImageScheme:
 		name = cleanSPDXName(srcMetadata.ImageMetadata.UserInput)
-		if name != "" {
-			identifier = path.Join("image", fmt.Sprintf("%s-%s", name, uniqueID.String()))
-		}
-		identifier = path.Join("image", fmt.Sprintf("%s", uniqueID.String()))
+		identifier = path.Join("image", fmt.Sprintf("%s-%s", name, uniqueID.String()))
 	case source.DirectoryScheme:
 		name = cleanSPDXName(srcMetadata.Path)
-		if name != "" {
-			identifier = path.Join("dir", fmt.Sprintf("%s-%s", name, uniqueID.String()))
-		}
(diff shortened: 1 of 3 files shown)


spiffcs:
I think there is a bug when scanning `directory`

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -41,12 +43,16 @@ func (pres *SPDXJsonPresenter) Present(output io.Writer) error {
...
+	uID := uuid.Must(uuid.NewRandom())
+
+	var name, namespace string
 	switch srcMetadata.Scheme {
 	case source.ImageScheme:
 		name = srcMetadata.ImageMetadata.UserInput
+		namespace = strings.Trim(fmt.Sprintf("%s-%s", name, uID.String()), "-")
 	case source.DirectoryScheme:
 		name = srcMetadata.Path
+		namespace = uID.String()
```
luhring, line 55: Why do we construct this value using two different approaches here?

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -67,7 +73,7 @@ func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx
 			LicenseListVersion: spdxlicense.Version,
 		},
 		DataLicense:       "CC0-1.0",
-		DocumentNamespace: fmt.Sprintf("https://anchore.com/syft/image/%s", srcMetadata.ImageMetadata.UserInput),
+		DocumentNamespace: fmt.Sprintf("https://anchore.com/syft/image/%s", namespace),
```
luhring, line 76: Now that the variable itself is called `namespace`, I think this would read better if we set the `DocumentNamespace` field to the value of `namespace` directly, so readers don't wonder about the two different possible meanings of "namespace".

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -41,12 +43,16 @@ func (pres *SPDXJsonPresenter) Present(output io.Writer) error {
 
 // newSPDXJsonDocument creates and populates a new JSON document struct that follows the SPDX 2.2 spec from the given cataloging results.
 func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx22.Document {
-	var name string
+	uID := uuid.Must(uuid.NewRandom())
```
luhring, line 46: It's not clear to me what the meaning of the variable name `uID` is. What does the `u` stand for in this context? Would something like `id` or `uuid` be more recognizable (while also being semantically appropriate) here?

spiffcs: yea uuid is good, but then it shadow imports over the `uuid` library so I opted to shorten it to `uID`

I can be more verbose and just call it `uniqueID` if you want.

luhring: Okay, I'm cool with verbose if it's clear! 👍  Also cool with shadowing if you don't need the package anywhere below in this scope

internal/presenter/packages/spdx_json_presenter.go
spiffcs: Good question - I opted to keep the `ImageMetadata.UserInput` as the namespace prefix for images since that's what exists on main. 

For the directory, I could not think of a great metadata identifier besides the digested contents path. The path presents a few challenges when appending it to our `https://anchore.com/syft/image/%s` format.

A user could specify an absolute path `/foo/bar` which would result in `...image//foo/bar-<uuid>`.
A user could also specify a relative path `./foo/bar` which would result in `...image/./foo/bar-<uuid>`

Rather than parse/clean all possible `fs` inputs I decided to opt for simplicity and generate a uuid since the `documentNamespace` field on main for a directory input is currently not set.

internal/presenter/packages/spdx_json_presenter.go
spiffcs: Makes sense - I can lift the `fmt.Sprintf` up so that we keep the naming similar.

spiffcs:
@luhring updated the verbosity and organization based on pr feedback. 

Let me know what you think of my comment RE: different value construction. Happy to make any edits that we settle on if we need more metadata in the namespace field for directory inputs.

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -12,8 +13,11 @@ import (
 	"github.com/anchore/syft/internal/version"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/source"
+	"github.com/google/uuid"
 )
 
+var anchoreNamespace = "https://anchore.com/syft/image"
```
luhring, line 19: nit: If this value never needs to change, we could use `const` here

internal/presenter/packages/spdx_json_presenter.go
spiffcs: That's a good callout. 

This behavior also exists currently on `main` so I'm happy to use this branch to squash this bug. I was able to reproduce it on my machine with alpine and saw this generated with the latest version of syft:

`"documentNamespace": "https://anchore.com/syft/image//Users/hal/development/images/alpine",`

We could run the resulting `namespace` string through a cleaner function that strips all bad prefixes.

We could also adopt a common method of just appending a `uuid` since the sbom have other fields that provide context as to what was scanned.

luhring: These are good ideas. I don't have a strong suggestion here. Here's what I do think...

1. We should probably use the same logic for building this string, regardless of an image or directory path. This lets us factor out this string generation as a function, if we want. Also, we should probably remove "image" from `"https://anchore.com/syft/image"`, since this value is getting used in both paths (this is true both _already_ and potentially _going forward_, too).

2. We could go the route of cleaning the user input value, but I could also see that potentially getting hairy. It'd be great if there was a convention we could lean on that exists already — one that has accounted for producing safe "job identifying" strings. I'm not sure if this exists today. So perhaps for now, UUIDs alone are enough (instead of incorporating user input)?

spiffcs: Yea - I think UUIDs alone are good enough until we start seeing more of a standard being adopted across the sbom community. I'll also update the path so we have `syft/dir` vs `syft/image` depending on the input contents

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -41,14 +44,19 @@ func (pres *SPDXJsonPresenter) Present(output io.Writer) error {
...
 // newSPDXJsonDocument creates and populates a new JSON document struct that follows the SPDX 2.2 spec from the given cataloging results.
 func newSPDXJsonDocument(catalog *pkg.Catalog, srcMetadata source.Metadata) spdx22.Document {
-	var name string
+	uniqueID := uuid.Must(uuid.NewRandom())
+
+	var name, identifier string
 	switch srcMetadata.Scheme {
 	case source.ImageScheme:
 		name = srcMetadata.ImageMetadata.UserInput
+		identifier = fmt.Sprintf("image/%s", uniqueID.String())
```
kzantow, line 53: Note: I think this identifier is meant to include the name, e.g. it should be something like: `http://anchore.com/syft/image/alpine-latest-SOMEUUID`. See: [DocumentNamespace spec](https://spdx.github.io/spdx-spec/document-creation-information/#652-intent), should have: `http://[CreatorWebsite]/[pathToSpdx]/[DocumentName]-[UUID]` this was why I rolled it into the same issue as missing DocumentName.

kzantow (changes requested):
Left some comments in the issue about this, too: https://github.com/anchore/syft/issues/495#issuecomment-933971588

internal/presenter/packages/spdx_json_presenter.go
spiffcs: Nice! Good to see it so explicitly formated in their docs. This format is pretty easy for images, but one of the review comments I got was that we want it to be similar across directory and image inputs. 

@luhring what are your thoughts given the spec information? Is it ok if we have some different formatting options for image vs directory?

What should the document name be when we scan a directory or OCI directory source?

luhring (approved):
🙌

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -113,3 +129,14 @@ func newSPDXJsonElements(catalog *pkg.Catalog) ([]spdx22.Package, []spdx22.File,
 
 	return packages, files, relationships
 }
+
+func cleanSPDXName(name string) string {
+	// remove # according to specification
+	name = strings.Replace(name, "#", "-", -1)
+
+	// remove : for url construction
+	name = strings.Replace(name, ":", "-", -1)
```
kzantow, line 138: is this needed?

internal/presenter/packages/spdx_json_presenter.go
```diff
@@ -41,14 +46,25 @@ func (pres *SPDXJsonPresenter) Present(output io.Writer) error {
...
-		name = srcMetadata.ImageMetadata.UserInput
+		name = cleanSPDXName(srcMetadata.ImageMetadata.UserInput)
+		input = "image"
 	case source.DirectoryScheme:
-		name = srcMetadata.Path
+		name = cleanSPDXName(srcMetadata.Path)
+		input = "dir"
+	}
+
+	if name != "." {
```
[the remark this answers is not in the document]
spiffcs: I think it's ok to have this solution while we sort out what `DocumentName` should be in this case.

internal/presenter/packages/spdx_json_presenter.go
spiffcs: yea ":" is not in that non-reserved list so an input like `alpine:latest` would need to be stripped

internal/presenter/packages/spdx_json_presenter.go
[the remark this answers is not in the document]
kzantow: Your call! I don't consider it a blocker. Rest is 👍

kzantow (approved):
Huge step forward for mankind!

Issue #478, opened 2021-08-10
Catalog Go modules used in Go binaries


luhring:
### Value

In order to better understand what software is present in a given context, Syft could inspect compiled Go binaries in order to determine the dependencies (called **[Go modules](https://golang.org/ref/mod)**) that have been included in the compiled binary.

This is especially valuable for scans of **container images**, where the only artifact present is the Go binary, and there are no manifest-like files (such as `go.mod` or `go.sum`) remaining in the image.

### Hypothesis

A complete list of a binary's included Go modules is included in the binary itself, and can be inspected by the natively available Go tooling. This can be done on the CLI using `go version -m <path/to/Go/binary>`, as well as with the Go standard library [functionality](https://github.com/golang/go/blob/go1.16.5/src/cmd/go/internal/version/version.go#L169) used by this CLI command.

### Implementation considerations

1. In order to prevent Syft from needing to inspect every file present in the scan target to test if the file is a Go binary, Syft could potentially search through only a prefiltered set of files — specifically, files that are binaries. If https://github.com/anchore/syft/issues/336 is implemented first, Syft (when scanning _images_) could shorten its execution time by querying Stereoscope for files that are binaries.
2. It's important to remember that the version of Go that Syft uses **might differ** from the version of Go that a discovered binary file uses. Care should be taken to ensure that Syft's Go binary analysis can be done with as many versions of Go as possible.

houdini91:
+1 
**Implementation considerations**
1) `file/lib-magic` tool/library do something similar (it does not seem to be strait forward as searching magic numbers) - [ https://github.com/file/file](url), although sadly binding to `lib-magic` this will force a dynamic link (i think), maybe someone implemented a pure go magic library.

2) I taken a look at `go version` source code (elf) it looks for the modules information out of a pointer in the .go-buildinfo section, i am guessing this will not change much between go versions.
Any way the actual binaries  `version` can be extracted as well to be compared to supported go versions.

wagoodman:
re: magic number processing, we've been careful to keep syft as a standalone static binary so pulling in a binding to a C lib would not be ideal. On #336 there is a [prototype branch in syft and stereoscope](https://github.com/anchore/syft/issues/336#issuecomment-875890241) for adding this support.

re: go version source, agreed. I think the best way forward is to pull in that code (since it isn't exported) and keep track on major go versions if that section definition changes.

fg-j:
[...]

PR #529, opened 2021-10-04
Add query by MIME type to source.FileResolver

wagoodman:
This PR adds the ability to fetch file locations from an image by MIME type from the `source.FileResolver`. This is useful in scenarios where you want to process only a certain kind of file (e.g. tar.gz or tgz) but can't rely exclusively on file extensions.

Partially implements anchore/syft#478


syft/source/directory_resolver.go
```diff
@@ -207,6 +216,7 @@ func (r directoryResolver) FilesByPath(userPaths ...string) ([]Location, error)
 			log.Warnf("unable to get file by path=%q : %+v", userPath, err)
 			continue
 		}
+		// TODO: why not use stored metadata?
```
wagoodman, line 219: note: I left this as a question for future us, not necessarily one to solve on this PR

spiffcs (approved):

PR #530, opened 2021-10-05
Add option to output SBOM report to a file

wagoodman:
This PR adds a new `--file` option that writes the report bytes out to the given file path. Note: since the same functionality will be ported to grype it was decided that `-f` will not be used since that option is already taken in grype.

### DIFF
--- a/cmd/event_loop.go
+++ b/cmd/event_loop.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/anchore/syft/internal/log"
@@ -14,11 +15,13 @@ import (
 // signal interrupts. Is responsible for handling each event relative to a given UI an to coordinate eventing until
 // an eventual graceful exit.
 // nolint:gocognit,funlen
-func eventLoop(workerErrs <-chan error, signals <-chan os.Signal, subscription *partybus.Subscription, ux ui.UI, cleanupFn func()) error {
+func eventLoop(workerErrs <-chan error, signals <-chan os.Signal, subscription *partybus.Subscription, cleanupFn func(), uxs ...ui.UI) error {
 	defer cleanupFn()
 	events := subscription.Events()
 	var err error
-	if ux, err = setupUI(subscription.Unsubscribe, ux); err != nil {
+	var ux ui.UI
+
+	if ux, err = setupUI(subscription.Unsubscribe, uxs...); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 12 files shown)


cmd/event_loop.go
```diff
@@ -81,7 +81,7 @@ func eventLoop(workerErrs <-chan error, signals <-chan os.Signal, subscription *
 func setupUI(unsubscribe func() error, ux ui.UI) (ui.UI, error) {
 	if err := ux.Setup(unsubscribe); err != nil {
 		// replace the existing UI with a (simpler) logger UI
-		ux = ui.NewLoggerUI()
+		ux = ui.NewLoggerUI(os.Stdout)
```
luhring, line 84: Want to make sure I follow: Why `os.Stdout` here?

internal/ui/select.go
```diff
@@ -11,7 +12,7 @@ import (
 
 // Select is responsible for determining the specific UI function given select user option, the current platform
 // config values, and environment status (such as a TTY being present).
-func Select(verbose, quiet bool) UI {
+func Select(verbose, quiet bool, reportWriter io.Writer) UI {
```
luhring, line 15: It'd be nice to explain the correct usage of this new parameter (even very briefly) in the function's doc comment

internal/config/application.go
```diff
@@ -29,6 +29,7 @@ type parser interface {
 type Application struct {
 	ConfigPath         string             `yaml:",omitempty" json:"configPath"`                                                         // the location where the application config was read from (either from -c or discovered while loading)
 	Output             string             `yaml:"output" json:"output" mapstructure:"output"`                                           // -o, the Presenter hint string to use for report formatting
+	File               string             `yaml:"file" json:"file" mapstructure:"file"`                                                 // -f, the file to write report output to
```
luhring, line 32: We should probably remove `-f` from this line's comment for now

cmd/root.go
```diff
@@ -40,6 +40,17 @@ func init() {
 
 	rootCmd.PersistentFlags().CountVarP(&persistentOpts.Verbosity, "verbose", "v", "increase verbosity (-v = info, -vv = debug)")
 
+	rootCmd.PersistentFlags().StringP(
+		"file", "f", "",
```
luhring, line 44: Should we remove the `"f"` for now?

cmd/report_writer.go
```diff
@@ -0,0 +1,29 @@
...
+	case 0:
+		return os.Stdout, nop, nil
+	default:
+		reportFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+		if err != nil {
+			return nil, nop, fmt.Errorf("unable to create report file: %w", err)
+		}
+		return reportFile, func() error {
+			if !appConfig.Quiet {
+				_, _ = fmt.Fprintf(os.Stderr, "Report written to %q\n", path)
```
luhring, line 24: Curious: Can we not use logging here?

cmd/root.go
wagoodman: this was my bad --I removed it but was restoring work from the reflog and accidentally brought it back. Removing again...

cmd/event_loop.go
luhring: Disregard the above — in favor of this comment -> https://github.com/anchore/syft/pull/530#discussion_r722450745

wagoodman: I've done a tiny refactor so that the caller can provide fallback UIs to use in case the preferred one does not work

spiffcs (approved):

luhring (approved):

PR #534, opened 2021-10-05
478 identify go binaries

spiffcs:


syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	sort.SliceStable(pkgsSlice, func(i, j int) bool {
+		return pkgsSlice[i].Name < pkgsSlice[j].Name
+	})
+
+	return pkgsSlice, nil
+}
+
+// The build info blob left by the linker is identified by
```
wagoodman, line 57: regarding the files we copied in from the internal go code, a couple of asks:
- can we keep them in their own (original) files? In this way if we need to incorporate upstream updates it makes it easier
- at the top of each file copied into our repo, can you add a note that mentions something to the affect of "this code was copied from <GH link to file>"?

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -0,0 +1,125 @@
...
+	pkgsSlice := make([]pkg.Package, 0)
+
+	// Identify if bin was compiled by go
+	x, err := openExe(reader)
+	if err != nil {
+		return pkgsSlice, err
+	}
+
+	_, mod := findVers(x)
+	fields := strings.Fields(mod)
```
wagoodman, line 25: Splitting off the parsing of the bytes from the build section into a separate function that can be unit tested would be ideal here (both for reading of this function and for an extra opportunity to test this interface)

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// slice off root pacakge info
+	var separator int
+	for x, field := range fields {
+		if field == packageIdentifier {
```
wagoodman, line 30: open question: do we want to only preserve dependencies? Or should we keep a package for the root module (without a version since it isn't available)?

syft/pkg/cataloger/golang/utils.go
```diff
@@ -0,0 +1,198 @@
...
+	"debug/elf"
+	"debug/macho"
+	"debug/pe"
+	"fmt"
+	"io"
+)
+
+// An exe is a generic interface to an OS executable (ELF, Mach-O, PE, XCOFF).
+// TODO: add back close method since we have a closer
+type exe interface {
```
wagoodman, line 14: `Close()` was removed from here, however, with the `io.ReadCloser` being passed in it doesn't look like omitting this function is necessary.

syft/pkg/cataloger/golang/parse_go_bin.go
wagoodman, line 25: TIL about `strings.Fields` --nice!

syft/pkg/cataloger/golang/parse_go_bin.go
spiffcs: Let's stick with preserving dependencies for the time being. I don't think a document would be accurate if the software sourced itself as a dependency

internal/ui/select.go
```diff
@@ -5,7 +5,7 @@ import (
 	"os"
 	"runtime"
 
-	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
```
spiffcs, line 8: I don't know how this came through, but my linter was yelling about it so I updated it here

syft/pkg/cataloger/golang/parse_go_bin.go
spiffcs: I am adding a `Location` to each package so that there is additional metadata on the path where this was found. Not always analogous to the root package, but let me know if you have a good spot in mind to place the root package.

syft/pkg/cataloger/golang/bin_cataloger.go
```diff
@@ -0,0 +1,58 @@
...
+// Catalog is given an object to resolve file references and content, this function returns any discovered Packages after analyzing rpm db installation.
+func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, error) {
+	fileMatches, err := resolver.FilesByMIMEType(mimeTypes...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find bin by mime types: %w", err)
+	}
+
+	var pkgs []pkg.Package
+	for _, location := range fileMatches {
+		r, err := resolver.FileContentsByLocation(location)
```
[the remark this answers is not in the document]
spiffcs: This is being closed now

syft/pkg/cataloger/golang/exe.go
```diff
@@ -0,0 +1,269 @@
+// This code was copied from the Go std library.
```
spiffcs, line 1: Add link to Github for this file

spiffcs: No lint copied files

syft/pkg/cataloger/golang/parse_go_bin_test.go
```diff
@@ -0,0 +1,57 @@
+package golang
```
spiffcs, line 1: Check other go mod directives for logic on parsing

spiffcs:
> This implementation is different than our internal one (we generate an anchore_hints.json file at binary build time so we use the compiler to output the result and massage it), but one issue we recently hit is that our initial implementation (now fixed) included the v in the go version strings (e.g. v1.1.1) but the CPE strings in the NVD seem to never contain the v. I see that this is including the v in the output. For the generated CPE string is the v being stripped?

Good callout @cburgess and thanks for the context. I ran locally just to get a sample and it looks like it's being stripped. I'll do some due diligence on the code to find where this is happening.

This screenshot is from the JSON output running on this branch.
Screen Shot 2021-10-06 at 5 17 23 PM

spiffcs:
I'll update the failing integration tests. Looks like there was a regression covering if a cataloger was removed.

cburgess:
The other thing our internal implementation does is extract the version of go used to build the binary and add that to the hints file. I believe that’s in a different location in the binary but I’m thinking it would be nice to provide that as well, possibly controlled by a flag. There have been cases in the past of a bug/vulnerability in the compiler being inherited by the binaries it produces.

PR #536, opened 2021-10-06
Update linter + fix whitespace

wagoodman:
Follow up from https://github.com/anchore/grype/pull/442#discussion_r722612680 (I realize for grype). While I was at it I updated our linter.

### DIFF
--- a/cmd/report_writer.go
+++ b/cmd/report_writer.go
@@ -11,18 +11,22 @@ import (
 
 func reportWriter() (io.Writer, func() error, error) {
 	nop := func() error { return nil }
-
 	path := strings.TrimSpace(appConfig.File)
+
 	switch len(path) {
 	case 0:
 		return os.Stdout, nop, nil
+
 	default:
 		reportFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+
 		if err != nil {
 			return nil, nop, fmt.Errorf("unable to create report file: %w", err)
 		}
+
 		return reportFile, func() error {
 			log.Infof("report written to file=%q", path)
+
 			return reportFile.Close()
 		}, nil
 	}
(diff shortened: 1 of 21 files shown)


cmd/report_writer.go
```diff
@@ -11,18 +11,22 @@ import (
...
 	case 0:
 		return os.Stdout, nop, nil
+
 	default:
 		reportFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+
 		if err != nil {
 			return nil, nop, fmt.Errorf("unable to create report file: %w", err)
 		}
+
```
luhring, line 26: ❤️  — very easy to follow, love it!

syft/pkg/cataloger/deb/cataloger.go
```diff
@@ -36,9 +36,9 @@ func (c *Cataloger) Name() string {
 }
 
 // Catalog is given an object to resolve file references and content, this function returns any discovered Packages after analyzing dpkg support files.
-// nolint:funlen
+
```
luhring, line 39: This blank line could have very easily resulted from a [comment I made](https://github.com/anchore/grype/pull/442#discussion_r722612680) 😄  So I wanted to chime in — blank lines between a declaration and its doc comment is not something that's important to me. Also, I think it breaks the association of the documentation to the declared thing. I haven't tested this on pkg.go.dev, but I have tested it in GoLand, and I find that the blank line causes the "hover over" documentation to no longer include the doc comment text.

luhring:
(Looks good ✅  — I didn't approve only because I couldn't remember if "auto merge" was enabled, and I wanted us to chat through https://github.com/anchore/syft/pull/536#discussion_r723178891 before merging 😃  )

spiffcs (approved):

PR #537, opened 2021-10-06
Allow Syft to build with podman instead of Docker

joshbressers:
This includes one small change to allow syft to build with podman as the container runtime instead of docker. 

The problem involved checking jar files built in a container. The jar archive parser test resulted in the Built-By flag being set to 1000 when podman was used (that's my UID, I assume the UID will match whoever runs the build, I did not try other UIDs). When native docker is used the Built-By flag is "?". This patch removes checking this specific flag as it is not a setting that affects the jar file in a meaningful way.

With this change the build now completes successfully


joshbressers:
I should add, this is the cleaned up version of https://github.com/anchore/syft/pull/526

PR #543, opened 2021-10-11
Add new format abstraction

wagoodman:
**This is not ready for final review yet for a few reasons**:
- implementing multiple formats should be done with multiple PRs
- there is an unnecessary breaking change with json encoding (remove scope) that can be deferred
- needs a good rebase

---

Partially addresses https://github.com/anchore/grype/issues/395

Today we have the `presenter` abstraction to write out internal SBOM results to a particular format (e.g. JSON, table, SPDX, CycloneDX). All model details about the format itself is contained under the presenter objects themselves. This abstraction was never meant to encapsulate handling intake (parsing) of these formats. As we add more formats it would be ideal to be able to parse these formats for use in other tools (such as syft). For this reason this PR adds a new `format` abstraction for specifying encoding/decoding details for a particular SBOM format (and the presenter can use format encoders to deal with presentation concerns).

The [first commit](https://github.com/anchore/syft/commit/9ded1c4c22c3c527e1b57e390ec53cb5ccc0a8b0) of this PR adds the new format abstractions:
- `type Encoder ...`: a function signature for encoding a given set of SBOM objects into bytes that are written to a writer.
- `type Decoder ...`: a function signature for decoding an SBOM from a reader and returning SBOM objects.
- `type Validator ...`: a function signature for observing the bytes of an SBOM document via a reader and returns any errors if the given document is not of a specific format.
- `type Format ...`: ties together the above bits of functionality into a single object tied to a specific `format.Option` (with helper functions)

This new pattern allows for:
- specifying not only encoding, but also decoding of a format
- separates the concerns of presenting to a writer (behavior oriented) from specifying the SBOM format document shape (data oriented). 
- allows for "one way" and "two way" format conversions. That is, you can have a format that has both encoding and decoding supported (such as spdx-json), or a format that only allows for encoding (such as table [not implemented yet]).
- enables encode operations without needing to use the presenter abstraction (which is meant specifically for deferred execution of encoding bound to a certain SBOM object)

This PR has moved a lot of code and large test fixtures. To better orient what is really happening I'd suggest reviewing a commit at a time in order... but these are the most important ones:
- Take a look at the new `format` abstraction first:  https://github.com/anchore/syft/commit/9ded1c4c22c3c527e1b57e390ec53cb5ccc0a8b0
- The next commits go over specific implementations that return the format struct: https://github.com/anchore/syft/commit/c1346ad62c72e2d20637830c36bd753c09ac2eb2, https://github.com/anchore/syft/commit/f6cc0c86289f22b7ea73437327967c0c0143c068
- We expose this new encode/decode behavior in the main syft lib API: https://github.com/anchore/syft/commit/12b44af471c77ae8b3d0ae537a29dec08fa1952c

The remaining commits involve adding tests, moving existing test-fixtures, removing some presenter implementations, and making common SPDX helpers available to other packages.... essentially it's what "shakes out" after getting the abstraction and initial implementations added.

Why move both JSON and SPDX-JSON? I attempted to move just SPDX-JSON to implement https://github.com/anchore/grype/issues/395, however, in doing so there was a lot of overlap with the JSON decoding implementation in grype. Instead of having this live in two places it was easier conceptually to go ahead an move the JSON decoding into the new approach.

**Regarding details of the SPDX-JSON fields**: new fields were added to the SPDX-JSON structs such that encoding-decoding loops are NOT lossy. That is, Before this PR if you were to encoding a SPDX-JSON document it would be lossy relative to the internal syft document... with the adjustments in this PR this is no longer the case. Why do this? In order to support keeping the `package.Metadata` fields that are sensitive to vulnerability scanning in grype. However, selecting ONLY the fields that grype needs is brittle from syft's point of view ("why were _these_ fields persisted and _those_ not?!?"). The initial implementation is lossless for this reason (and the added reason that selecting a different shape would require custom `pkg.*Metadata` struct definitions for each customization, which seems difficult to maintain!).

### DIFF
--- a/internal/anchore/import_package_sbom.go
+++ b/internal/anchore/import_package_sbom.go
@@ -25,7 +25,7 @@ type packageSBOMImportAPI interface {
 	ImportImagePackages(context.Context, string, external.ImagePackageManifest) (external.ImageImportContentResponse, *http.Response, error)
 }
 
-func packageSbomModel(s source.Metadata, catalog *pkg.Catalog, d *distro.Distro, scope source.Scope) (*external.ImagePackageManifest, error) {
+func packageSbomModel(s source.Metadata, catalog *pkg.Catalog, d *distro.Distro, _ source.Scope) (*external.ImagePackageManifest, error) {
 	var buf bytes.Buffer
 	pres := formats.ByOption(format.JSONOption).Presenter(catalog, &s, d)
 	err := pres.Present(&buf)
--- a/internal/anchore/import_package_sbom_test.go
+++ b/internal/anchore/import_package_sbom_test.go
@@ -1,7 +1,6 @@
 package anchore
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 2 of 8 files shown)


wagoodman:
Closing to keep the context here --a new PR will introduce the new pattern and one ported format

Issue #540, opened 2021-10-07
Capture additional go package data


wagoodman:
Derived from conversations on https://github.com/anchore/syft/pull/534#issuecomment-937915373

Today any go packages cataloged do not have any additional data captured on `pkg.Metadata`, such as the h1 digest (found in the go.sum), which version of go the module is intended for (found in the go.mod), the go version used to compile the dependency (applicable for the go binary cataloger only), go module replacements (for both the go.mod cataloger and the binary cataloger), etc.

Given the set of distinct data that we could capture, and the fact that they are somewhat different depending on if it was found in a go.mod or a binary (and in the future we may be getting this information from source directly) --all of this may hint at there being multiple `GoMetadata*` structs that are tailored for each case, not just one.

PR #546, opened 2021-10-12
540 add additional go package metadata

spiffcs:
Fixes(#540)

### Add GolangBinMetadataType.

The two fields included:
- GoCompiledVersion
- H1Digest

#### GoCompiledVersion
Takes the version used to compile the parent binary and adds it as as metadata to the dependencies

#### H1Digest
Takes the checksum for each dependency and adds it as metadata.

I could not think of any good additions for the metadata type for the `mod` parser. Here is the open discussion surrounding the "truthiness" of go.mod from the cloud native slack:
https://cloud-native.slack.com/archives/CGEQHPYF4/p1633986885169500

Options for Go Mod Parser Metadata:
- h1 digest requires us to parse and match go.sum (we can do this as part of this PR if we want to add this metadata)
- Go module replacement information

### Regarding Go module replacement information
Is this correct for the cataloger? https://github.com/anchore/syft/blob/10368f6fe8889f641c34703c054424f35f114ad5/syft/pkg/cataloger/golang/parse_go_mod.go#L37-L44

`m.New.Path` would be a completely new path and NOT overwrite a Path found in `file.Require`

Do we want to add metadata information that describes the `Old` Path and Version
```
// A Version (for clients, a module.Version) is defined by a module path and version pair.
// These are stored in their plain (unescaped) form.
type Version struct {
	// Path is a module path, like "golang.org/x/text" or "rsc.io/quote/v2".
	Path string

	// Version is usually a semantic version in canonical form.
	// There are three exceptions to this general rule.
	// First, the top-level target of a build has no specific version
	// and uses Version = "".
	// Second, during MVS calculations the version "none" is used
	// to represent the decision to take no version of a given module.
	// Third, filesystem paths found in "replace" directives are
	// represented by a path with an empty version.
	Version string `json:",omitempty"`
}
```

### DIFF
--- a/syft/pkg/golang_mod_metadata.go
+++ b/syft/pkg/golang_mod_metadata.go
@@ -1,3 +1,0 @@
-package pkg
-
-type GolangModMetadata struct{}
--- a/syft/pkg/metadata.go
+++ b/syft/pkg/metadata.go
@@ -16,7 +16,6 @@ const (
 	RustCargoPackageMetadataType MetadataType = "RustCargoPackageMetadata"
 	KbPackageMetadataType        MetadataType = "KbPackageMetadata"
 	GolangBinMetadataType        MetadataType = "GolangBinMetadata"
-	GolangModMetadataType        MetadataType = "GolangModMetadata"
 )
 
 var AllMetadataTypes = []MetadataType{
@@ -30,5 +29,4 @@ var AllMetadataTypes = []MetadataType{
 	RustCargoPackageMetadataType,
 	KbPackageMetadataType,
 	GolangBinMetadataType,
-	GolangModMetadataType,
 }


kzantow (approved):
You are the winner! More code deleted than added!

deitch:
Did this actually go in? I ran `syft -o spdx-json` on a directory with go, I see versions but not h1 hashes, e.g.:

```json
  {
   "SPDXID": "SPDXRef-1761ac713ff9ae54",
   "name": "github.com/BurntSushi/toml",
   "licenseConcluded": "NONE",
   "externalRefs": [
    {
     "referenceLocator": "cpe:2.3:a:BurntSushi:toml:v0.3.1:*:*:*:*:*:*:*",
     "referenceType": "cpe23Type"
    },
    {
     "referenceCategory": "PACKAGE_MANAGER",
     "referenceLocator": "pkg:golang/github.com/BurntSushi/toml@v0.3.1",
     "referenceType": "purl"
    }
   ],
   "filesAnalyzed": false,
   "licenseDeclared": "NONE",
   "sourceInfo": "acquired package info from go module information: vendor/github.com/containerd/containerd/go.mod",
   "versionInfo": "v0.3.1"
  },
```

Same for other output formats.

kzantow:
Hi @deitch -- it looks like this change would only affect the Syft JSON output and CycloneDX, not SPDX.

I ran the latest Syft release against itself and see entries in CycloneDX like: 
```json
        {
          "name": "syft:metadata:goCompiledVersion",
          "value": "go1.18.6"
        },
        {
          "name": "syft:metadata:h1Digest",
          "value": "h1:a2VclLzOGrwOHDiV8EfBGhvjHvP46CtW5j6POvhYGGo="
        },
```

And Syft format:
```json
      "metadata": {
        "goCompiledVersion": "go1.18.6",
        "architecture": "amd64",
        "h1Digest": "h1:ij3fYGe8zBF4Vu+g0oT7mB06r8sqGWKuJu1yXeR4by8=",
        "mainModule": "github.com/anchore/syft"
      }
```

deitch:
Ah, that does, indeed, explain it.

> The problem with SPDX is... where to put that data? There do not really seem to be appropriate fields

Could we use `externalRefs`? I am not sure if there is an official list of reference categories; I could not find in [the appropriate section of the spdx spec](https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field), which leads me to think it may be somewhat extensive?

Alternatively, is the [purl for go spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang) extensible enough to support it?

deitch:
Ah, I see the externalRefs in [Annex F](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/). In [Other](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f5-other)? 🤷‍♂️

kzantow:
@deitch -- yes Annex F has that information about externalRefs. I'd hesitate to use `other` for reasons previously mentioned: it is unclear what that information is, and reading SPDX format, we couldn't assume `other` is this hash value.

I've created an issue to track this so we don't keep commenting on a closed PR -- please bring the discussion there: https://github.com/anchore/syft/issues/1261

PR #548, opened 2021-10-13
Add windows support

spiffcs:
Syft now compiles on windows and can be executed. 

Picture of compile and help command running:
demo

Picture of compile and base command running:
image


spiffcs:
Syft runs after removing the offending `filepath.Clean` not sure if this is the entire fix and how this might break other cases:

demo-fix

Replaced locally removing this function:
https://github.com/anchore/stereoscope/blob/53801100857890df6c7e3ec6e80d47a7f45334d4/pkg/file/path.go#L19-L26

PR #551, opened 2021-10-14
update to point at new script for new release

spiffcs:
Small patch to use new upstream script which adds `all` as a valid case for downloading goreleaser.

We should eventually migrate to one of the official install instructions here:
https://goreleaser.com/install/

To test:

`make bootstrap` <-- this should now return status code 0

Important changes:
https://gist.github.com/spiffcs/479255e52143bb68bb0b2974080bb0a4#file-goreleaser-sh-L232-L253

Add `all` to allow the script to consume the new release binary.


spiffcs:
I can also add the script to the repository. I wanted to keep the Makefile pretty consistent since other tools were sourcing from external scripts

Issue #555, opened 2021-10-16
Encapsulate all artifacts in syft JSON output


wagoodman:
Today we output a json structure similar to the following:
```
{
   artifacts: [
      # list of packages
   ],
   relationships: [
      # list of package relationships
   ]
   distro: {...},
   ...
}
```

As we move forward and want to capture more kinds of artifacts we could consider moving to something closer to this:

```
{
   artifacts: {
      packages: [
         # list of packages
      ],
      files: [
         # list of files
      ],
      distro: {...},
      # more artifact types...
   },
   relationships: [
      # list of relationships for ANY artifact
   ],
   ...
}
```

In this way we can agnostically express all artifacts kinds without filling the root-level object up with new elements.

Issue #554, opened 2021-10-16
Promote cataloging task pattern


wagoodman:
Today the package command is a single task that is responsible for invoking all of the package catalogers and returning a presenter capable of encoding an SBOM in a particular format.

At the same time, the power-user command can enable and disable different catalogers and format the SBOM document more dynamically. Internally the power-user command is made of a set of tasks that is dynamic based on the configuration. Additionally, all tasks write to a single result document instance, encapsulating all possible things that can be cataloged. 

There are elements of the internal power-user command approach that would be ideal to promote forward as patterns we want to see on the package command ([or any command that will do cataloging in the future](https://github.com/anchore/syft/issues/516)).

Goals:
1. Have a singular analysis result document to encapsulate all things discovered. Possibly a new `sbom` package with a struct called `Document`. (tracked in https://github.com/anchore/syft/issues/555)
2. Move towards a task-based approach for separating work and attaching the result to the analysis document

These goals focus on how the work of cataloging is completed and persisting in a programmatic way that is useful to others.

spiffcs:
Dependent on review of #606. When that is in then we can start building tasks around the package command that talk to this new document.

spiffcs:
Grabbing this while waiting for review on our new media type.

Issue #516, opened 2021-09-27
Syft command restructure


spiffcs:
**What would you like to be added**:

Currently syft's root and packages commands produce the same `package` specific output:
`syft packages node:latest > /dev/null `
`syft node:latest > /dev/null`
 
![Screen Shot 2021-09-27 at 11 06 10 AM](https://user-images.githubusercontent.com/32073428/134935109-1e361bcd-8ac7-4b9f-b31b-2a55c775009a.png)

syft also has a `power-user` command which produces a more verbose output concerning `packages`, `secrets`, `file metadata` `file-digests`:
`syft power-use node:latest > /dev/null`

![Screen Shot 2021-09-27 at 11 09 39 AM](https://user-images.githubusercontent.com/32073428/134935722-b6632eed-3000-41cb-8f24-30680524757b.png)

**I believe the `syft [noun]` pattern is not the space where we want to be focusing development moving forward.**

Given that syft is concerned with sbom generation, I propose we look to move our command structure towards `syft [verb]`, starting with `syft create` or `syft describe` in order to generate an sbom.

The goal of this command API change follows improving a few key points.

1) Our current path of coupling `presenter/noun` logic and structures will lead to a good bit of unmanageable code sprawl. We can already see here that the `poweruser` config struct has a host of VERY specific and useful information that only it has access to. Rather than reimplement/confuse these structures for each `noun` we should look to shift them lower in the program so presenters interpret all possible SBOM entities the same way.
https://github.com/anchore/syft/blob/e5caba043d9686383853689c5a9e886030d9207d/internal/presenter/poweruser/json_document_config.go#L11-L21

2) As we move into wanting to describe more SBOM entities, code paths like `presenter/secrets` `presenter/files` `presenter/relationships` just do not scale into the user behavior or usefulness we want for the tool.  `syft create` should already do analysis for the different entities with sane defaults and a config input that can increase/decrease the entities analyzed for the command.

3) Refocus presenter to be `presenter/spdx` `presenter/json` `presenter/cyclonedx`. Formats as described previously should not be children of nouns like `packages`. The current program architecture does just this `syft/internal/presenter/packages/spdx.go,json_pacakge.go,table,etc`. Refactoring to `syft [verb] -o [presenter]` allows us much more space to design common logic/structures when dealing with different SBOM entities.

**Why is this needed**:
Remove formal sbom entity semantics from command API to refocus program on its main directive of generating sbom.

**Additional context**:
TODO

[...]

PR #605, opened 2021-10-29
Add SPDX tag-value format object

### LINKED ISSUE #590: Port remaining package presenters to format objects
https://github.com/anchore/syft/pull/550 introduced a new approach to express encoding and decoding formats. We should port the existing presenters to format objects with encoders (represented by this issue) and in the future add decoding and detection capabilities as needed.

Formats that need to be ported:
- [x] Syft JSON #550
- [x] CycloneDX #589 
- [x] Text #604 
- [x] Table #586 
- [x] SPDX JSON #584 
- [x] SPDX Tag-Value #605

wagoodman:
Closes #590  

This PR ports the last of the package presenter objects to a format object (with encoding only, no decoder or validator). Since this is the last object there is a little more clean up work that was additionally done:
- Swaps the `presenter` utils to the `formats` utils in the `cmd` package.
- Removes the `presenter` package altogether, swapping for [a single global definition](https://github.com/wagoodman/go-presenter) 

Future work:
- Remove the power-user presenter by combining it with the packages format object (related to #555 , #516, and #554)


spiffcs (approved):

spiffcs:
So happy this is here!

Issue #556, opened 2021-10-16
Prefer artifact relationships over package relationships


wagoodman:
**Context**
Today we capture only package-to-package relationships. It isn't possible to express things such as a file-to-package or file-to-file relationship.

**The goals of this issue are**:
1. Promote relationships as a first class ~artifact~ objects raised by catalogers.
2. Enable describing relationships for things other than packages (such as files)

To goal 1: catalogers today only raise `[]pkg.Package`, which means that any relationships you are adding must be added after the cataloging step, after connectivity data may be missing (this is the case for most ecosystems). If the cataloger were to additionally return `[]Relationship` it would allow the catalogers to be more expressive.

To goal 2: Today's [`pkg.Relationship`](https://github.com/anchore/syft/blob/main/syft/pkg/relationship.go) should probably be promoted to a root-level package. This additionally implies that the existing `pkg.ID` will need to be replaced with something more agnostic... with additional requirements that anything that wants to express a relationship with something else have an ID that works globally. 

This work is most likely coupled to how IDs are expressed in packages; that is, should they remain as UUIDs? or something more stable like a package fingerprint? See https://github.com/anchore/syft/issues/363 for more details.

PR #606, opened 2021-10-29
Introduce a single SBOM document

wagoodman:
This PR replaces `poweruser.JSONDocumentConfig` with `sbom.SBOM`. Specifically this datastructure is being promoted as a first-class object to be used internally for tasking, presentation, and processing concerns. This is proposed in https://github.com/anchore/syft/issues/555#issuecomment-952071565 but closely affects #554 and #556. 

The main proposal of the PR is the following datastructure: https://github.com/anchore/syft/blob/a26e9e4eb726b70e59662a7196a4de22fe7d01b7/syft/sbom/sbom.go#L10-L23

All other changes attempt to promote usage of this data structure over passing parts of an SBOM around, for example... replacing:
```golang
func Encode(catalog *pkg.Catalog, metadata *source.Metadata, dist *distro.Distro, scope source.Scope, option format.Option) ([]byte, error) { ... }
```

with:
```golang
func Encode(s sbom.SBOM, option format.Option) ([]byte, error) { ... }
```

Notes:
- Relationships has not been included but will later be added as a field under `sbom.SBOM` in other PRs implementing #556
- `source.Scope` was removed from any presenter output as described in https://github.com/anchore/syft/issues/555#issuecomment-952071565


internal/anchore/import_package_sbom.go
```diff
@@ -24,10 +26,19 @@ type packageSBOMImportAPI interface {
 	ImportImagePackages(context.Context, string, external.ImagePackageManifest) (external.ImageImportContentResponse, *http.Response, error)
 }
 
-func packageSbomModel(s source.Metadata, catalog *pkg.Catalog, d *distro.Distro, scope source.Scope) (*external.ImagePackageManifest, error) {
+func packageSbomModel(srcMetadata source.Metadata, catalog *pkg.Catalog, d *distro.Distro, _ source.Scope) (*external.ImagePackageManifest, error) {
```
spiffcs, line 29: Do we need to keep the `_` value for `source.Source` here or can we update the signature at this point to exclude that value?

internal/formats/spdx22json/test-fixtures/snapshot/TestSPDXJSONImagePresenter.golden
```diff
@@ -3,44 +3,15 @@
...
  "creationInfo": {
-  "created": "2021-10-22T19:25:38.341582Z",
+  "created": "2021-10-29T16:26:09.001799Z",
   "creators": [
    "Organization: Anchore, Inc",
    "Tool: syft-[not provided]"
   ],
   "licenseListVersion": "3.14"
  },
- "syftSourceData": {
```
spiffcs, line 13: Moving to the new model it looks like we lose this `syftSourceData` information?

Is this something we no longer want to include in the output?

internal/formats/spdx22json/to_format_model.go
```diff
@@ -19,9 +19,9 @@ import (
 )
 
 // toFormatModel creates and populates a new JSON document struct that follows the SPDX 2.2 spec from the given cataloging results.
-func toFormatModel(catalog *pkg.Catalog, srcMetadata *source.Metadata, _ *distro.Distro, _ source.Scope) model.Document {
-	name := documentName(srcMetadata)
-	packages, files, relationships := extractFromCatalog(catalog)
+func toFormatModel(s sbom.SBOM) model.Document {
```
spiffcs, line 22: Love the simplification of all the signatures!

internal/formats/syftjson/test-fixtures/snapshot/TestImagePresenter.golden
```diff
@@ -88,8 +88,7 @@
...
    "config": "eyJhcmNoaXRlY3R1cmUiOiJhbWQ2NCIsImNvbmZpZyI6eyJFbnYiOlsiUEFUSD0vdXNyL2xvY2FsL3NiaW46L3Vzci9sb2NhbC9iaW46L3Vzci9zYmluOi91c3IvYmluOi9zYmluOi9iaW4iXSwiV29ya2luZ0RpciI6Ii8iLCJPbkJ1aWxkIjpudWxsfSwiY3JlYXRlZCI6IjIwMjEtMTAtMDRUMTE6NDA6MDAuNjM4Mzk0NVoiLCJoaXN0b3J5IjpbeyJjcmVhdGVkIjoiMjAyMS0xMC0wNFQxMTo0MDowMC41OTA3MzE2WiIsImNyZWF0ZWRfYnkiOiJBREQgZmlsZS0xLnR4dCAvc29tZWZpbGUtMS50eHQgIyBidWlsZGtpdCIsImNvbW1lbnQiOiJidWlsZGtpdC5kb2NrZXJmaWxlLnYwIn0seyJjcmVhdGVkIjoiMjAyMS0xMC0wNFQxMTo0MDowMC42MzgzOTQ1WiIsImNyZWF0ZWRfYnkiOiJBREQgZmlsZS0yLnR4dCAvc29tZWZpbGUtMi50eHQgIyBidWlsZGtpdCIsImNvbW1lbnQiOiJidWlsZGtpdC5kb2NrZXJmaWxlLnYwIn1dLCJvcyI6ImxpbnV4Iiwicm9vdGZzIjp7InR5cGUiOiJsYXllcnMiLCJkaWZmX2lkcyI6WyJzaGEyNTY6ZmI2YmVlY2I3NWIzOWY0YmI4MTNkYmYxNzdlNTAxZWRkNWRkYjNlNjliYjQ1Y2VkZWI3OGM2NzZlZTFiN2E1OSIsInNoYTI1NjozMTliNTg4Y2U2NDI1M2E4N2I1MzNjOGVkMDFjZjAwMjVlMGVhYzk4ZTdiNTE2ZTEyNTMyOTU3ZTEyNDRmZGVjIl19fQ==",
-   "repoDigests": [],
-   "scope": "Squashed"
```
spiffcs, line 92: Looks like we also lose the scope data point in the presenter

spiffcs (commented):
First pass. 

I like the new `SBOM` struct and how it just encapsulates current information. Is there a future where the underlying `Artifacts` struct is changed as we build into developing relationships between everything?

Also small comments on data loss across the new outputs of the golden snapshots. I wasn't sure if we wanted to start excluding the information, or if it got lost as a consequence of moving to the new encode architecture.

internal/formats/spdx22json/test-fixtures/snapshot/TestSPDXJSONImagePresenter.golden
wagoodman: I noticed this while preparing the PR --this field was never meant to be included to begin with. I added it accidentally in https://github.com/anchore/syft/pull/584 while removing features from https://github.com/anchore/syft/pull/578 . This is probably better served in its own PR (or at least in the PR description here!)

internal/formats/syftjson/test-fixtures/snapshot/TestImagePresenter.golden
wagoodman: indeed, this is one of the notes from the PR description (the second bullet at the bottom)

internal/anchore/import_package_sbom.go
wagoodman: 💯 I'll remove this reference and clean up the call stack here

wagoodman:
@spiffcs 

> Is there a future where the underlying Artifacts struct is changed as we build into developing relationships between everything?

Indeed! I have some of that drafted out in #607, specifically: https://github.com/anchore/syft/blob/c3a65f5fa2a0a511d4183133ce3d8c457e5a5a50/syft/sbom/sbom.go#L13 

This is where the relationships object gets promoted to in the end (the `sbom.SBOM` struct).

cmd/packages.go
```diff
@@ -260,21 +261,29 @@ func packagesExecWorker(userInput string) <-chan error {
...
 			Type:  event.PresenterReady,
-			Value: f.Presenter(catalog, &src.Metadata, d, appConfig.Package.Cataloger.ScopeOpt),
+			Value: f.Presenter(sbomResult),
 		})
 	}()
 	return errs
 }
 
-func runPackageSbomUpload(src *source.Source, s source.Metadata, catalog *pkg.Catalog, d *distro.Distro, scope source.Scope) error {
+func runPackageSbomUpload(src *source.Source, s source.Metadata, catalog *pkg.Catalog, d *distro.Distro) error {
```
luhring, line 286: **Question:** Would we expect to see this function's parameter list start using the unified SBOM type as well?

internal/formats/spdx22json/to_format_model.go
luhring: This! ⬆️

luhring (approved):
Very nice work! 👏 

I had just one question from curiosity

cmd/packages.go
spiffcs: Would we still need `*source.Source` passed in? so `(src *source.Source, sbom.SBOM)`

I don't see anywhere else in the code where those two types are paired up which gives me some pause.

If we assemble the SBOM type and then just have to deconstruct its parts for the ImportConfig then in my mind its a bit clearer to have them separated in the signature so there is one less indirection.

```
anchore.ImportConfig{
    Catalog: sbom.Artifacts.PackageCatalog
}
```

vs

```
	importCfg := anchore.ImportConfig{
		Catalog:                 catalog,
	}
```

spiffcs: @wagoodman this was the only outstanding question left here

wagoodman: I hesitated with updating the import path with the same pattern --but in hind sight I should have. I'll update in a follow up PR.

spiffcs (approved):

wagoodman:
@spiffcs thanks a ton for solving what ended up being a caching issue I was having --I was starting to go a little crazy there 🙌

cmd/packages.go
wagoodman: Follow up PR https://github.com/anchore/syft/pull/621

Issue #558, opened 2021-10-16
Top-level API should be more composable


wagoodman:
**What would you like to be added**:
More reusable primitives when syft is used as a library. This would be able to do at least the following tasks:

-  get source objects from descriptions + configuration
-  catalog packages and persist work into an `sbom.Document` (as proposed in https://github.com/anchore/syft/issues/554)
-  catalog other artifacts (e.g. files, secrets, etc)
-  encode sbom (initial work from https://github.com/anchore/syft/pull/550 need to be adjusted when the sbom document is incorporated)
-  decode sbom (initial work from https://github.com/anchore/syft/pull/550 need to be adjusted when the sbom document is incorporated)

This should probably be done after (that is, this issue is most likely related to...):
- Task pattern for catalogers + encapsulating sbom document: https://github.com/anchore/syft/issues/554
- Utilizing the format pattern introduced in https://github.com/anchore/syft/pull/550 (possibly provide a shim for presenter-only options today so they are represented as formats)
- Using a more agnostic source object (not coupled to syft) https://github.com/anchore/syft/issues/557

**Why is this needed**:
The existing top-level API functions are for a full parse-catalog processing unit, where as separating parsing and cataloging into separate functions is useful if you want to reuse the same parsed image (which takes a while) with different catalogers calls (which relatively speaking doesn't take that long).

This would additionally improve interoperability with `anchorectl`.

**Tasks**
- [x] #1751
- [x] #1383
- [x] #1764
- [x] Provide API proposal (source API refactor PR and SBOM builder refactor PR has this)
- [x] Provide examples for using this API https://github.com/anchore/syft/pull/2517

wagoodman:
From refinement:
- Do we want to encapsulate the stereoscope primitives or not? Should not have to do `stereoscope.*` to interact with syft, even if the objects from stereoscope is what we end up allowing to be passed around the API.
- It would be ideal to not expose behavior (e.g. tasks) and instead expose declarations (e.g. cataloger configurations?)

wagoodman:
Also, I think we should distinguish scoped capabilities of syft vs the pure definitions needed to interop with primitives that syft works with.

An example: Syft can encode and decode various formats (spdx, cyclonedx, etc). Additionally, syft has the definition of what a `Format` is... something that can encode an SBOM into bytes, decode an SBOM from bytes, and validate the underlying bytes to see if they represent the format the object represents.

These are probably not strongly separated enough in today's API. 

[...]

PR #607, opened 2021-10-31
First class relationships

wagoodman:
This PR primarily introduces the concept of an `artifact.Relationship` (instead of a `pkg.Relationship`) and adds a new element to the `sbom.SBOM` struct (as a root level element, a sibling of the `Artifact` field).

As a result the following additions/changes were also made to support introducing `artifact.Relationship`:
1. Create the idea of an `artifact.ID` as a type
2. Introduce a new `Identifiable` interface for objects that can return a stable `artifact.ID` (and thereby be able to be part of a `artifact.Relationship`)
3. Hoist up the package fingerprinting to be a general purpose utility in the `artifact` package
4. Hoist up the existing `ownership-by-file-overlap` relationship to now be discovered/created just after the cataloging phase (instead of within the `syftjson` format encoder as it is today)
5. Ignore select metadata struct fields and made subtle changes to CPE generation behavior to further stabilize the package ID (based on an object hash)
6. Migrate the encode-decode cycle test as an integration test (utilizing all of the known package types) to prove that a package ID remains stable upon multiple encode-decode cycles
7. ~split `source.Location` into `source.Coordinates`. This new object has the minimal information needed to get to a path within a source. This new `source.Coodinates` implements `artifact.Identifiable` such that files can form relationships with packages (setting up for Port the SPDX JSON files relationships in a separate PR)~  This work has been broken out into a separate PR

Additional considerations:
- I attempted to remove the `pkg.Catalog` altogether, however, elected to leave it for now to keep this PR smaller and increment towards a final state in a future TBD PR. This keeps the same semantics as the catalog does today in the sense that when a package is added to the catalog it is assumed that it will not be mutated. This is further enforced now by never exposing references to packages that are contained within the catalog.

Partially addresses #556 ; remaining work:
- Port the SPDX JSON files relationships into the new artifact relationships (instead of doing do in the encoder)


syft/artifact/relationship.go
```diff
@@ -1,20 +1,19 @@
-package pkg
+package artifact
 
 const (
 	// OwnershipByFileOverlapRelationship indicates that the parent package owns the child package made evident by the set of provided files
 	OwnershipByFileOverlapRelationship RelationshipType = "ownership-by-file-overlap"
 )
 
+type Identifiable interface {
+	Identity() ID
```
luhring, line 9: nit / curious: Could we call the method `ID()`? This to me would read as a conventional Go "getter", where the method name matches the type of what's returned. The way I think of it, we're getting the "ID" (the identification), not the "identity", which would be more broad. What are your thoughts?

wagoodman: I originally did have this as ID, one of the soft issues with this is that Package has a field called ID that needs to be exported. Up for naming suggestions here.

luhring: Interesting. Is `Package.ID` a field that might be going away soon? (I'm vaguely remembering this from a conversation last week...)

wagoodman: Possibility, but not finalized. We could make that change on this PR. I'll review tomorrow and get a more concrete answer.

luhring: Okay cool! Just exploring. I'll think on a method name, too.

spiffcs: `Identifier`? Always a good excuse to follow the `er` convention and an `Identifier` implementing `Identify` feels correct?

cmd/packages.go
```diff
@@ -248,7 +250,7 @@ func packagesExecWorker(userInput string) <-chan error {
 		}
 		defer cleanup()
 
-		catalog, d, err := syft.CatalogPackages(src, appConfig.Package.Cataloger.ScopeOpt)
+		catalog, relationships, d, err := syft.CatalogPackages(src, appConfig.Package.Cataloger.ScopeOpt)
```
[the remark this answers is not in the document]
wagoodman: I resisted on making this update here. I think this is better fit under https://github.com/anchore/syft/issues/554 
 and/or https://github.com/anchore/syft/issues/558 where we consider the units of work, what they produce, or what they modify. That is, since `syft.CatalogPackages` only generates a part of an SBOM today it wasn't clear if this should be modified to generate an entire SBOM (it seemed awkward from the power user command objects since then there would be multiple SBOM objects to deal with, which seem semantically wrong)

internal/formats/spdx22json/to_format_model.go
```diff
@@ -39,37 +39,32 @@ func toFormatModel(catalog *pkg.Catalog, srcMetadata *source.Metadata, _ *distro
...
+func documentName(srcMetadata source.Metadata) string {
+	switch srcMetadata.Scheme {
+	case source.ImageScheme:
+		return cleanSPDXName(srcMetadata.ImageMetadata.UserInput)
+	case source.DirectoryScheme:
+		return cleanSPDXName(srcMetadata.Path)
 	}
+
 	// TODO: is this alright?
 	return uuid.Must(uuid.NewRandom()).String()
```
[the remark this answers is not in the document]
wagoodman: agreed --we should make an issue for this to track separately 👍

internal/formats/syftjson/to_format_model.go
```diff
@@ -85,29 +89,26 @@ func toPackageModel(p *pkg.Package) model.Package {
...
 	result := make([]model.Relationship, len(relationships))
 	for i, r := range relationships {
 		result[i] = model.Relationship{
-			Parent:   string(r.Parent),
-			Child:    string(r.Child),
+			Parent:   string(r.From.Identity()),
+			Child:    string(r.To.Identity()),
 			Type:     string(r.Type),
-			Metadata: r.Metadata,
+			Metadata: r.Data,
```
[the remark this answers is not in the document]
wagoodman: Indeed we should. Right now I'm only at the point where I'm making non-breaking changes (so not changing the JSON schema). We could get this work in and done without making any user-surface area changes (and there are a lot of changes to get in before schema changes, so it seemed like a good "stopping point").

I'll leave this out for now, but will continually re-evaluate. If there are other breaking changes that need to be done that are similar then adding this in earlier may make sense.

internal/anchore/import_package_sbom.go
```diff
@@ -8,26 +8,33 @@ import (
...
 type packageSBOMImportAPI interface {
 	ImportImagePackages(context.Context, string, external.ImagePackageManifest) (external.ImageImportContentResponse, *http.Response, error)
 }
 
-func packageSbomModel(s source.Metadata, catalog *pkg.Catalog, d *distro.Distro, scope source.Scope) (*external.ImagePackageManifest, error) {
+func packageSbomModel(srcMetadata source.Metadata, catalog *pkg.Catalog, d *distro.Distro, _ source.Scope) (*external.ImagePackageManifest, error) {
 	var buf bytes.Buffer
 
-	err := syftjson.Format().Presenter(catalog, &s, d, scope).Present(&buf)
+	// TODO: once the top-level API is refactored and SBOMs are the unit of work, then this function will be passed an SBOM and there would be no more need to create an SBOM object here.
```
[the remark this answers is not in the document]
wagoodman: yeah --the "draftyness" of this PR kinda deserved an ephemeral comment in a few places :)

wagoodman:
I'm going to rebase since #606 was merged and has made this look a little crazy.

wagoodman:
Since the package fingerprinting and SBOM data structure PRs have landed I think I'm ready to start tackling replacing ID assignment with the fingerprinting method. This implies considering possible removal of the `pkg.Catalog` --I'll see what shakes out.

syft/artifact/relationship.go
wagoodman: I think I'm going to try and go the direction of an `ID()` method as a unravel the package ID assignments in the `pkg.Catalog` and start relying on the new fingerprint method.

luhring:
>One complexity that has been introduced that I want to note before I get too far: since the package ID is a function of the bytes on the pkg.Package struct [...], upon translating from one format to another we need to keep track of the string IDs from a static document and look up the object that they represent.

Question: Are we treating the static package ID values on any formats (that Syft can decode) as "okay to discard" during decoding? Specifically, I'm asking about more than using the IDs for things like discovering relationships, I'm asking about preserving the literal ID values themselves.

This is a really interesting perspective, and it's making me question an early opinion I had of not retaining static ID values on native Syft packages (in the business layer). As we think about using Syft as a library, I'm wondering if consumers of the library might have a reasonable expectation that they can find the original ID values (say, from there SPDX document that they Syft-decoded), within the business objects that Syft exposes in the library, for some task they're trying to accomplish. (Just wondering... I don't have an exact scenario in mind. Maybe some sort of policy logic.)

wagoodman:
> I'm wondering if consumers of the library might have a reasonable expectation that they can find the original ID values (say, from there SPDX document that they Syft-decoded)

With the last few commits in to prove out the PoC, I've been wondering the same thing.

As a `<side note>`... as of right now (at the time of this writing) there is no ID problem across multiple formats from a user perspective. That is, ID is not a field that is exposed in the existing CycloneDX and SPDX implementations (SPDX [uses name/type](https://github.com/anchore/syft/blob/bb0f35bac4e2e610e61be55b27b3810099794479/internal/formats/spdx22tagvalue/to_format_model.go#L95) which should change since this can create problems and we do not use the [bom-ref field](https://cyclonedx.org/docs/1.2/json/#metadata_component_bom-ref) in the [definition of a component]( https://github.com/anchore/syft/blob/main/internal/formats/cyclonedx12xml/model/component.go#L6-L20), but this will change as we need component refs in the near future).
`</side note>`

We've changed ID to be behavior of a package and not a static (random) data value on a package. One nicety of a stabilizing ID based on contents is that the ID matches between multiple runs of analyzing the same source. The downside is that if the package data is lossy from converting from format-to-format then the ID will be different between documents.

I think we may have reached two mutually exclusive use cases here: one use case wanting ID stability based on "the definition of a package" and another use case wanting ID stability across different descriptions of the same basic package.

We cannot use the same field to handle both use cases. Either we will need to declare that one of these use cases is unsupported or we will need to split up responsibilities between two different fields. (anyone see a third option?)

syft/artifact/relationship.go
wagoodman: Identity has been renamed to ID 👍

luhring:
That `<side note>` is a very good one to realize. 😄 

>I think we may have reached two mutually exclusive use cases here: one use case wanting ID stability based on "the definition of a package" and another use case wanting ID stability across different descriptions of the same basic package.
>
>We cannot use the same field to handle both use cases. Either we will need to declare that one of these use cases is unsupported or we will need to split up responsibilities between two different fields. (anyone see a third option?)

Well said. My two cents: in general, I still favor the `ID()` approach where it's a stable identifier yielded from a deterministic function that considers the package's (or other artifact's) distinct identity. Maybe we start with this, and when (if?) the need arises for capturing other identifiers, we solve that in a designated part of the data structure (e.g. `originalID` or something). WDYT?

wagoodman:
I landed on the same side of the fence as you @luhring  --I think this is good enough for now, the new behavior is better than before, and when we have a specific use case that is raising a problem we can solve it then.

internal/formats/spdx22json/to_format_model.go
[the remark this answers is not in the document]
spiffcs: https://github.com/anchore/syft/issues/622

spiffcs:
> Maybe we start with this, and when (if?) the need arises for capturing other identifiers, we solve that in a designated part of the data structure (e.g. `originalID` or something). WDYT?

I don't think bifurcating to other designated spots for an `originalID` is what we want here. I think in future PR it will be important to identify the base structs/fields that each format share.

> one use case wanting ID stability based on "the definition of a package" and another use case wanting ID stability across different descriptions of the same basic package.

If the definition of a package (before enoding to a format) is the same at the bottom of syft, then I believe there is a scenario where we fulfill both cases here so long as the generated ID happens BEFORE any format/presentation specific structs have a possibility of being hashed.

cmd/power_user.go
```diff
@@ -109,28 +110,61 @@ func powerUserExecWorker(userInput string) <-chan error {
...
+		errs <- err
+		return
+	}
+
+	for _, relationship := range relationships {
+		c <- relationship
+	}
+}
+
+func mergeResults(cs ...<-chan artifact.Relationship) <-chan artifact.Relationship {
```
spiffcs, line 151: Nit: When dealing with more complex types like `<-chan artifact.Relationship` I find it easier to reason about the function signature if the return value is also declared `(results <-chan artifact.Relationship)`.

syft/artifact/id.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// ID represents a unique value for each package added to a package catalog.
+type ID string
+
+type Identifiable interface {
+	ID() ID
+}
+
+func DeriveID(obj interface{}) (ID, error) {
```
spiffcs: Some of the larger comments on this PR talk about formalizing the ID across formats. I think a more strict signature might help with this. 

If we have this as `interface` what are all possible types within this project that could be passed here?
Can we limit it to some common field shared among all types?

syft/pkg/catalog.go
```diff
@@ -38,21 +41,26 @@ func (c *Catalog) PackageCount() int {
 }
 
 // Package returns the package with the given ID.
-func (c *Catalog) Package(id ID) *Package {
+func (c *Catalog) Package(id artifact.ID) *Package {
```
spiffcs, line 44: Do we return a copy now because we want to force the immutability of the catalog?

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -21,7 +23,7 @@ var _ common.ParserFn = parseApkDB
 
 // parseApkDb parses individual packages from a given Alpine DB file. For more information on specific fields
 // see https://wiki.alpinelinux.org/wiki/Apk_spec .
-func parseApkDB(_ string, reader io.Reader) ([]pkg.Package, error) {
+func parseApkDB(_ string, reader io.Reader) ([]pkg.Package, []artifact.Relationship, error) {
```
spiffcs, line 26: Are we adding relationships as a placeholder for now? On parse what relationships are we expecting to be able to generate?

syft/pkg/cataloger/common/cpe/generate.go
```diff
@@ -83,7 +83,7 @@ func candidateVendors(p pkg.Package) []string {
 	// allow * as a candidate. Note: do NOT allow Java packages to have * vendors.
 	switch p.Language {
 	case pkg.Ruby, pkg.JavaScript:
-		vendors.addValue("*")
+		vendors.addValue(wfn.Any)
```
spiffcs, line 86: Nice find

wagoodman:
re: https://github.com/anchore/syft/pull/607#issuecomment-965328777 , I agree, the suggestion of keeping `originalID` probably isn't ideal --but we haven't made any decisions about that. What we're settling on is that we can keep the ID system as it is today (based on the object hash) and solve this later if we see a specific problem with that approach in the future. 

> If the definition of a package (before enoding to a format) is the same at the bottom of syft, then I believe there is a scenario where we fulfill both cases here so long as the generated ID happens BEFORE any format/presentation specific structs have a possibility of being hashed.

I'd have to think on this; it appears that what you're saying is true within the scope of generating a single document, but not necessarily true when converting from one format to another (something that isn't supported by syft yet anyway, so isn't a concern today, but could be in the future).

spiffcs (commented):
Comments - I'll dig in again later

syft/artifact/id.go
wagoodman: I though about doing something along those lines, where there would be a `IDFromLocation(...)` and `IDFromPackage(...)` in the `artifact` package. The first problem is that this would introduce a cyclic dependency between package and artifact --ideally it should be the artifact package that is at the "root". I then realized that all of the ID implementations would be the same anyway (pass an object into `Hash`). For this reason I went the direction of making a single utility function (also, I renamed this in a later commit to `IDFromHash(...)`).

The largest con against this decision I see is that it would be possible to get an `artifact.ID` for something that doesn't represent an artifact. However, since artifact.ID is a string, there is nothing stopping anyone from creating artifact.IDs --that is, there is no enforcement through encapsulation for this.

It could be that we move this function to a separate package... maybe `internal/hash`? Then at least there wouldn't be any problems with cyclic dependencies.

syft/pkg/catalog.go
wagoodman: exactly 👍 I just updated the PR description a minute ago to reflect these thoughts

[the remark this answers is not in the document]
wagoodman: seems like a dup comment

syft/pkg/cataloger/apkdb/parse_apk_db.go
wagoodman: Yeah, a necessary placeholder for now 👍 

Since the generic cataloger has been enhanced to [allow for parser functions to raise up artifact relationships](https://github.com/anchore/syft/blob/aba0de6e651d17a6e291e144cb92bdfd04693c69/syft/pkg/cataloger/common/parser.go#L11), and the APK cataloger [uses the generic cataloger](https://github.com/anchore/syft/blob/aba0de6e651d17a6e291e144cb92bdfd04693c69/syft/pkg/cataloger/apkdb/cataloger.go#L17), that meant that this function signature needed to be updated as well.

wagoodman: There may not be any relationships that the APK DB parser will raise up, but other parser functions may have the opportunity to raise up relationships (such as `gemfile.lock` and `rpmdb` parsers). I haven't double checked the underlying data available for APK yet, but APK might have package dependency data for us to leverage.

syft/pkg/cataloger/common/cpe/generate.go
wagoodman: Thanks... that one took a little while to find and is one of the main reasons why I beefed up the encode-decode cycle tests for all of this.

syft/pkg/cataloger/apkdb/parse_apk_db_test.go
```diff
@@ -775,7 +775,7 @@ func TestMultiplePackages(t *testing.T) {
 				}
 			}()
 
-			pkgs, err := parseApkDB(file.Name(), file)
+			pkgs, _, err := parseApkDB(file.Name(), file)
```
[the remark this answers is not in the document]
wagoodman: I went back and forth on this. There are a lot of places that ignore relationship data in test results now that [parser functions raise up artifact relationships](https://github.com/anchore/syft/blob/aba0de6e651d17a6e291e144cb92bdfd04693c69/syft/pkg/cataloger/common/parser.go#L11). I made certain to wire up anywhere that could raise up relationship data in the future in the cataloging execution path even if the cataloger in question does not surface relationships. However, I did not update the tests since this PR doesn't actually add any new relationships.

My take: any PR that is enhancing one or more catalogers to discover relationships should come with tests that prove the new functionality works. The lack of tests on a PR I think would be more of a signal than anything else.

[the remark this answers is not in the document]
wagoodman: disregard... TODO statements here we come...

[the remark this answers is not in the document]
wagoodman: 👍  https://github.com/anchore/syft/pull/607/commits/628c2e44a5ce039a1a12b8f9748e8a3b1ca8d251

wagoodman:
From an offline conversation: it looks like `ID()` on `source.Location` could be problematic when working with fields that either have a `ref` or do not. This PR opted out of this problem by ignoring the `ref` field (and `VirutalPath`), however, this won't help in situations where we index by `source.Location` but only have the `ReadPath` and `FilesystemID` (when we port over the SPDX package-to-file relationship). I'll keep looking into this.

wagoodman:
To account for the problem mentioned in https://github.com/anchore/syft/pull/607#issuecomment-965704866 I've split out `source.Location` into a new `source.Coordinates` which holds the minimal information needed for determining "where" a file is at relative to a source. (I'll update the PR description)

spiffcs (approved):

PR #608, opened 2021-11-01
Allow for cataloging a single file

### LINKED ISSUE #541: Allow cataloging from a single file
This idea came from @kzantow 

The ability to point syft at a single file and generate an SBOM from the contents. For instance, pointing to a single `gemspec` without the requirement of being installed within a `specifications` directory and raising up all of the package information we can glean from the gem-spec cataloger would be ideal.

```
syft ./my.gemspec
# catalogs the gemspec, leaving all directory contents alone
```

**Why is this needed**:
In instances where you do not want to manipulate the filesystem and isolate a single file in an empty directory for scanning. For instance, scanning a single jar in a directory of jars without moving the target jar to a new directory.

wagoodman:
This PR allows for cataloging a single file by adding a `file` scheme (in addition to the existing `dir` and image-oriented schemes):

```bash
❯ cat requirements.txt
flask==4.0.0

❯ syft requirements.txt
 ✔ Indexed requirements.txt 
 ✔ Cataloged packages      [1 packages]

NAME   VERSION  TYPE   
flask  4.0.0    python 
```

Example usage:
```bash
# explicit file scheme
$ syft file:./path/to/file

# implicit file scheme
$ syft ./path/to/file
```

Closes #541

### DIFF
--- a/syft/source/directory_resolver.go
+++ b/syft/source/directory_resolver.go
@@ -66,50 +66,74 @@ func newDirectoryResolver(root string, pathFilters ...pathFilterFn) (*directoryR
 
 func (r *directoryResolver) indexTree(root string, stager *progress.Stage) ([]string, error) {
 	log.Infof("indexing filesystem path=%q", root)
+
+	var roots []string
 	var err error
+
 	root, err = filepath.Abs(root)
 	if err != nil {
 		return nil, err
 	}
-	var roots []string
+
+	// we want to be able to index single files with the directory resolver. However, we should also allow for attempting
+	// to index paths that do not exist (that is, a root that does not exist is not an error case that should stop indexing).
+	// For this reason we look for an opportunity to discover if the given root is a file, and if so add a single root,
+	// but continue forth with index regardless if the given root path exists or not.
+	fi, err := os.Stat(root)
+	if err != nil && fi != nil && !fi.IsDir() {
+		newRoot, err := r.addPathToIndex(root, fi)
+		if err = r.handleFileAccessErr(root, err); err != nil {
+			return nil, fmt.Errorf("unable to index path: %w", err)
+		}
+
+		if newRoot != "" {
+			roots = append(roots, newRoot)
+		}
+		return roots, nil
+	}
 
 	return roots, filepath.Walk(root,
 		func(path string, info os.FileInfo, err error) error {
 			stager.Current = path
 
-			// ignore any path which a filter function returns true
-			for _, filterFn := range r.pathFilterFns {
-				if filterFn(path) {
-					return nil
-				}
-			}
-
-			if err = r.handleFileAccessErr(path, err); err != nil {
-				return err
-			}
-
-			// link cycles could cause a revisit --we should not allow this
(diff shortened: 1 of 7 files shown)


syft/source/directory_resolver.go
```diff
@@ -66,50 +66,74 @@ func newDirectoryResolver(root string, pathFilters ...pathFilterFn) (*directoryR
...
+	if err = r.handleFileAccessErr(path, err); err != nil {
+		return "", err
+	}
 
-			newRoot, err := r.addPathToIndex(path, info)
-			if err = r.handleFileAccessErr(path, err); err != nil {
-				return fmt.Errorf("unable to index path: %w", err)
-			}
+	// link cycles could cause a revisit --we should not allow this
+	if r.fileTree.HasPath(file.Path(path)) {
```
spiffcs, line 121: Are we sure we want to use file.Path here after running into the windows errors from the previous PR?

syft/source/directory_resolver.go
```diff
@@ -66,50 +66,74 @@ func newDirectoryResolver(root string, pathFilters ...pathFilterFn) (*directoryR
...
+		return "", nil
+	}
 
-			if newRoot != "" {
-				roots = append(roots, newRoot)
-			}
+	if info == nil {
```
spiffcs, line 125: Nit: I express why, but to me info != nil feels like the better path where then our default execution would return the error

syft/source/scheme.go
```diff
@@ -19,41 +19,49 @@ const (
...
 
-	// we should attempt to let stereoscope determine what the source is first --but, just because the source is a valid directory
-	// doesn't mean we yet know if it is an OCI layout directory (to be treated as an image) or if it is a generic filesystem directory.
+	// try the most specific sources first and move out towards more generic sources.
+
+	// first: let's try the image detector, which has more scheme parsing internal to stereoscope
 	source, imageSpec, err := imageDetector(userInput)
-	if err != nil {
-		return UnknownScheme, image.UnknownSource, "", fmt.Errorf("unable to detect the scheme from %q: %w", userInput, err)
+	if err == nil && source != image.UnknownSource {
```
spiffcs, line 47: This seems awkward - Note for myself to come back and read the full expanded flow.

spiffcs (commented):
First pass - I'll look again more tonight

syft/source/directory_resolver.go
wagoodman: this is gonna be a hard habit for me to break :)

syft/source/directory_resolver.go
wagoodman: it's probably since that would mirror the same operations done when checking err, but in this case it's still the unhappy path that is in the guard condition.

spiffcs (approved):

spiffcs:
@wagoodman Approved

syft/source/scheme.go
wagoodman: (I'm assuming that there were not questions/comments here)

PR #609, opened 2021-11-02
PHP parser composer.lock

Zilborg:
Signed-off-by: Nikita <33390074+Zilborg@users.noreply.github.com>

Hello, guys. 

Added parser for `composer.lock`.
Documentation: https://getcomposer.org/doc/04-schema.md

Screenshot 2021-11-02 at 15 37 06

### DIFF
--- a/syft/pkg/cataloger/php/parse_composer_lock.go
+++ b/syft/pkg/cataloger/php/parse_composer_lock.go
@@ -41,7 +41,7 @@ func parseComposerLock(_ string, reader io.Reader) ([]pkg.Package, error) {
 				Name:     name,
 				Version:  version,
 				Language: pkg.PHP,
-				Type:     pkg.PhpPkg,
+				Type:     pkg.PhpComposerPkg,
 			})
 		}
 	}
--- a/syft/pkg/cataloger/php/parse_composer_lock.go
+++ b/syft/pkg/cataloger/php/parse_composer_lock.go
@@ -1,0 +1,50 @@
+package php
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/common"
+)
+
+type ComposerLock struct {
+	Packages   []Dependency `json:"packages"`
+	PackageDev []Dependency `json:"packages-dev"`
+}
+
+type Dependency struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+// integrity check
+var _ common.ParserFn = parseComposerLock
+
+// parseComposerLock is a parser function for Composer.lock contents, returning "Default" php packages discovered.
+func parseComposerLock(_ string, reader io.Reader) ([]pkg.Package, error) {
+	packages := make([]pkg.Package, 0)
+	dec := json.NewDecoder(reader)
+
+	for {
+		var lock ComposerLock
+		if err := dec.Decode(&lock); err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("failed to parse composer.lock file: %w", err)
+		}
+		for _, pkgMeta := range lock.Packages {
+			version := pkgMeta.Version
+			name := pkgMeta.Name
+			packages = append(packages, pkg.Package{
+				Name:     name,
+				Version:  version,
+				Language: pkg.PHP,
+				Type:     pkg.PhpPkg,
+			})
+		}
+	}
+
+	return packages, nil
+}
--- a/syft/pkg/type.go
+++ b/syft/pkg/type.go
@@ -14,6 +14,7 @@ const (
 	RpmPkg           Type = "rpm"
 	NpmPkg           Type = "npm"
 	PythonPkg        Type = "python"
+	PhpPkg           Type = "php"
 	JavaPkg          Type = "java-archive"
 	JenkinsPluginPkg Type = "jenkins-plugin"
 	GoModulePkg      Type = "go-module"
@@ -29,6 +30,7 @@ var AllPkgs = []Type{
 	RpmPkg,
 	NpmPkg,
 	PythonPkg,
+	PhpPkg,
 	JavaPkg,
 	JenkinsPluginPkg,
 	GoModulePkg,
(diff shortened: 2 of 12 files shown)


spiffcs:
Thanks for the contribution @Zilborg! I'll take a look today when I have some time.

It does look like some tests are still failing at the latest commit. If you have a second today there might be some follow-up code that needs to be adjusted to rectify these. 

https://github.com/anchore/syft/runs/4079998089?check_suite_focus=true

Zilborg:
> Thanks for the contribution @Zilborg! I'll take a look today when I have some time.
> 
> It does look like some tests are still failing at the latest commit. If you have a second today there might be some follow-up code that needs to be adjusted to rectify these.
> 
> https://github.com/anchore/syft/runs/4079998089?check_suite_focus=true

The trouble in test cases for new packages type (`php`). All checks for parser was passed.

syft/pkg/type.go
```diff
@@ -14,6 +14,7 @@ const (
 	RpmPkg           Type = "rpm"
 	NpmPkg           Type = "npm"
 	PythonPkg        Type = "python"
+	PhpPkg           Type = "php"
```
wagoodman, line 17: we try to keep package names as close to the package manager in question (when possible). 

```suggestion
	PhpComposerPkg           Type = "php-composer"
```

syft/pkg/cataloger/php/parse_composer_lock.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/common"
+)
+
+type ComposerLock struct {
+	Packages   []Dependency `json:"packages"`
+	PackageDev []Dependency `json:"packages-dev"`
+}
+
+type Dependency struct {
```
wagoodman, line 17: in the future we can extract more information from the underlying JSON structure and add a new `pkg.PhpComposerMetadata` type and struct to capture additional data. 

For this PR getting composer support is a good enough step to get in on it's own 👍 .

wagoodman:
Today we have the [image cataloger set](https://github.com/anchore/syft/pull/609/files#diff-ebae68bfe16b465ffb6f9bdad5117c92ed36a6cebcad98fa2f5a4d55cfa0445cL33) that is used on container images to detect installations of packages and we have the [directory cataloger set](https://github.com/anchore/syft/pull/609/files#diff-ebae68bfe16b465ffb6f9bdad5117c92ed36a6cebcad98fa2f5a4d55cfa0445cL47) for scanning within repos and ad-hoc directories. 

The only real difference between these sets is that the image set requires that catalogers detect _installed packages_ while the directory catalogers do not have this requirement. I see the new cataloger was added to the directory cataloger set (which seems right), do you know of any other conventions that would allow us to add the same cataloger for a tweaked regex for images (installed composer packages)?

For instance, for ruby gems we use the `gemfile.lock` for directory scans and NOT image scans, however, we can discover installed gems if we look for `gemspec` files within a `specifications` directory (https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/ruby/catalogers.go#L22). 

Is there any similar convention with composer? (I'm not that familiar with the PHP ecosystem)

No worries if the answer is no! I was just checking to see if there was an opportunity here for detecting installed PHP composer packages... and if not we can get this in as-is and make a needed exception for the failing integration test.

Zilborg:
Ok, on Monday (8.11) I will see the built containers with composer packages, also, will ask your question our developer.

Zilborg:
Hello, guys. I've reversed built containers:
```
# 1. Detect installed versions
root@2be1b12ad733:/# ll /etc/php/
7.4/ 8.0/ 8.1/

# All names of installed extended packages (without versions!)  for php via system package manager (such as ‘apt install php-mysql’) 
root@2be1b12ad733:/# ll /etc/php/7.4/mods-available/
total 184
drwxr-xr-x 1 root root 4096 Nov  9 11:38 ./
drwxr-xr-x 1 root root 4096 Sep 10 20:45 ../
-rw-r--r-- 1 root root   54 Feb 20  2021 amqp.ini
-rw-r--r-- 1 root root   72 Oct 22 15:33 bcmath.ini
-rw-r--r-- 1 root root   74 Aug 26 18:51 calendar.ini
-rw-r--r-- 1 root root   71 Aug 26 18:51 ctype.ini
…

# 2. Version of this libs can get only(IMHO) via system package manager

# 3. Composer download all packages in ‘./vendor’ folder
root@e0939bb18d36:/var/www/vendor# ll
total 272
drwxr-xr-x 67 www-data www-data 4096 Nov  3 20:05 ./
drwxr-xr-x  1 www-data www-data 4096 Nov  3 20:05 ../
drwxr-xr-x  3 www-data www-data 4096 Nov  3 20:05 adoy/
drwxr-xr-x  3 www-data www-data 4096 Nov  3 20:05 alcaeus/
drwxr-xr-x  3 www-data www-data 4096 Nov  3 20:05 andres-montanez/
…

# 4. Each of folder above is ‘username/package-name’ and contains composer.json, but without these versions.
root@e0939bb18d36:/var/www/vendor/adoy/fastcgi-client# grep -rnw './' -e 'version'
./fcgiget.php:21: * @version     1.0
./src/Adoy/FastCGI/Client.php:34: * @version     1.0.0
./src/Adoy/FastCGI/Client.php:280:        return chr(self::VERSION_1)         /* version */
./src/Adoy/FastCGI/Client.php:368:        $ret['version']       = ord($data[0]);
root@e0939bb18d36:/var/www/vendor/adoy/fastcgi-client# cat composer.json
{
  "name": "adoy/fastcgi-client",
  "description": "Lightweight, single file FastCGI client for PHP.",
  "keywords": [
    "fastcgi",
    "fcgi"
  ],
  "license": "MIT",
  "authors": [
    {
      "name": "Pierrick Charron",
      "email": "pierrick@adoy.net"
    }
  ],
  "autoload": {
    "psr-0": {
      "Adoy\\FastCGI\\": "src"
    }
  }
}
```
So package version of composer packages I’ve found only in lock file.

wagoodman (approved):
@Zilborg Thanks for the confirming... we can alway try to add image support in a future PR. I've updated the integration test on your branch and rebased due to a merge conflict.

This looks great as is --thanks for the contribution!

Issue #612, opened 2021-11-02
Media type for Syft SBoM JSON format


sclevine:
**What would you like to be added**:

Declaration of an official media type for Syft's JSON SBoM format.

**Why is this needed**:

Integration with the [Cloud Native Buildpacks](https://buildpacks.io) project, which allows complete SBoM to be generated automatically (e.g., using Syft) during the application build process.

**Additional context**:

See: https://github.com/buildpacks/lifecycle/issues/755

The CNCF Buildpacks project has an API that allows SBoM files with CycloneDX and SPDX media types to be generated by buildpacks and automatically attached to container images. This allows vulnerability scanning tools that consume SBoMs (like Grype) to match software components to vulnerabilities with a strong guarantee that the SBoMs are complete (due to the contractual nature of the buildpack API). Parts of this model were assessed by the CNCF Security TAG, with notes and details [here](https://github.com/cncf/tag-security/tree/main/assessments/projects/buildpacks).

I would like users of Cloud Native Buildpacks to be able to scan buildpack-generated SBoMs with Grype, but Grype currently only supports Syft's JSON format. I'm proposing that Cloud Native Buildpacks add Syft's JSON format as a possible SBoM format, but this requires a defined a media type for Syft's JSON format. I might recommend something like: `application/vnd.syft+json`.

spiffcs:
Hey @sclevine! Is there another link to that Buildpacks project? The above link sends me to a domain that seems to be for sale.

samj1912:
@spiffcs https://buildpacks.io/ should work :)

sclevine:
Apologies, URI fixed :)

luhring:
Hey @sclevine — yes, absolutely we need to do this. We were just chatting about this today. There are other needs for Syft "types" coming up, such as in attestations.

I saw the issue you linked. It sounds like one suggestion is `application/vnd.syft+json`, is that right?

cc: @wagoodman

sclevine:
Great to hear 😄 

> I saw the issue you linked. It sounds like one suggestion is `application/vnd.syft+json`, is that right?

That's my suggestion, similar to CycloneDX's `application/vnd.cyclonedx+json`:
1. A producer-specific media type (`vnd.`)
2. Intended for use with a specific application category (`application/`)
3. In JSON format (`+json`)
 
See other examples here: https://www.iana.org/assignments/media-types/media-types.xhtml#application

Although not mandatory, it would be good practice to register whatever you choose with IANA:
[...]

PR #618, opened 2021-11-05
document current draft for iana submission

spiffcs:
WIP: Ideally we can discuss this in the afternoon and have a submission by the end of the day

### DIFF
--- a/schema/json/vnd.syft+json
+++ b/schema/json/vnd.syft+json
@@ -42,11 +42,9 @@ Interoperability considerations: This media type has the same
    interoperability considerations of application/json as described
    in [RFC8259].
 
-Published specification: The specification can be found on the
-   Syft github under the schema directory https://github.com/anchore/syft/blob/main/schema/json
-
-   The current JSON schema version, 1.1.0, can be found at
-   https://github.com/anchore/syft/blob/main/schema/json/schema-1.1.0.json
+Published specification: The specification can be found on the main
+   Syft github repository under the schema directory
+   https://github.com/anchore/syft/blob/main/schema/json
 
 Applications that use this media: This media type is used to specify
    a software bill of materials.
@@ -75,8 +73,8 @@ General Comments:
 
 Person to contact for further information:
 
-   1. Name: Christopher Phillips
-   2. Email: christopher.phillips@anchore.com
+   1. Name: Dan Luhring
+   2. Email: dan.luhring@anhore.com
 
 Intended usage: Common
 
@@ -84,4 +82,4 @@ Intended usage: Common
    specification. It is intended to be exchanged between different
    parties of the software supply chain.
 
-Author/Change controller: Christopher Phillips, on behalf of Anchore
+Author/Change controller: Dan Luhring, on behalf of Anchore
--- a/schema/json/vnd.syft+json
+++ b/schema/json/vnd.syft+json
@@ -43,7 +43,7 @@ Interoperability considerations: This media type has the same
    in [RFC8259].
 
 Published specification: The specification can be found on the main
-   Syft github repository under the schema directory
+   Syft Github repository under the schema directory
    https://github.com/anchore/syft/blob/main/schema/json
 
 Applications that use this media: This media type is used to specify
@@ -65,7 +65,7 @@ Additional information:
 
    1. Deprecated alias names for this type: N/A
    2. Magic number(s): N/A
-   3. File extension(s): .json
+   3. File extension(s): syft.json
    4. Macintosh file type code: N/A
    5. Object Identifiers: N/A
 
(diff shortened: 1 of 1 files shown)


schema/json/vnd.syft+json
```diff
@@ -0,0 +1,85 @@
...
+   transmission of the SBOM to 3rd parties. These additional
+   requirements are considered out of scope for the specification.
+   And will typically be enforced by contract or copyright terms.
+
+Interoperability considerations: This media type has the same
+   interoperability considerations of application/json as described
+   in [RFC8259].
+
+Published specification: The specification can be found on the main
+   Syft github repository under the schema directory
```
luhring, line 46: nit: `github` -> `GitHub`

schema/json/vnd.syft+json
```diff
@@ -0,0 +1,85 @@
...
+
+Fragment identifier considerations: N/A
+
+Restrictions on usage: N/A
+
+Additional information:
+
+   1. Deprecated alias names for this type: N/A
+   2. Magic number(s): N/A
+   3. File extension(s): .json
```
luhring, line 68: I like [this suggestion](https://github.com/anchore/syft/issues/612#issuecomment-962264996) of using the extension `.syft.json`.

schema/json/vnd.syft+json
```diff
@@ -0,0 +1,85 @@
...
+   4. Macintosh file type code: N/A
+   5. Object Identifiers: N/A
+
+General Comments:
+
+Person to contact for further information:
+
+   1. Name: Dan Luhring
+   2. Email: dan.luhring@anhore.com
```
luhring, line 77: `anhore` -> `anchore`

schema/json/vnd.syft+json
```diff
@@ -0,0 +1,85 @@
...
+Encoding considerations: binary
+
+   This media type has all of the same encoding considerations of
+   application/json as described in [RFC8259].
+
+Security considerations: This media type has all of the same security
+   considerations of application/json as described in [RFC8259].
+
+   Additional care should be taken when processing additionalProperties,
+   digest, files, fileClassifications <>..... <----------------------------TODO Let's get this complete
```
luhring, line 27: Question: Is it acceptable to convey the meaning of this paragraph without referring to specific field names/types from the schema? In my mind, the versioning of the schema governs what's currently accessible, and I'm thinking it'd be nice not to have to remember to update this central media type definition as the schema evolves, potentially removing/changing these referenced names each time.

What do others think?

schema/json/vnd.syft+json
```diff
@@ -0,0 +1,85 @@
...
+   transmission of the SBOM to 3rd parties. These additional
+   requirements are considered out of scope for the specification.
+   And will typically be enforced by contract or copyright terms.
+
+Interoperability considerations: This media type has the same
+   interoperability considerations of application/json as described
+   in [RFC8259].
+
+Published specification: The specification can be found on the main
+   Syft Github repository under the schema directory
```
luhring, line 46: nit: `Github` -> `GitHub`

schema/json/vnd.syft+json
```diff
@@ -0,0 +1,85 @@
...
+
+Fragment identifier considerations: N/A
+
+Restrictions on usage: N/A
+
+Additional information:
+
+   1. Deprecated alias names for this type: N/A
+   2. Magic number(s): N/A
+   3. File extension(s): syft.json
```
luhring, line 68: nit: `syft.json` -> Does this need to be prefixed with a dot (i.e. `.syft.json`)? I noticed it was `.json` previously (with a leading dot)

schema/json/vnd.syft+json
spiffcs: Resolved to be less constrained to our schema. We just call out the JSON considerations from RFC8259

luhring (approved):
This looks great, @spiffcs! 👏

schema/json/vnd.syft+json
```diff
@@ -0,0 +1,70 @@
...
+Media subtype name: vnd.syft+json
+
+Required parameters: N/A
+
+Optional parameters:
+version
+
+The version parameter refers to the Syft specification version in use.
+
+version = 1*DIGIT "." 1*DIGIT "." 1*DIGIT
```
wagoodman, line 13: Question: does `1*DIGIT` only allow for values like `5` and not `50`?

spiffcs: Good Question - here is the relevant text.
https://www.rfc-editor.org/rfc/rfc2045.html#section-4

I asked the reviewer if we could use `x.x.x` and they said it was fine extending to M/m/p here.

MIME-VERSION in their specification does only allow for a single digit

```
version := "MIME-Version" ":" 1*DIGIT "." 1*DIGIT
```

I can follow up and see what edits we might need for the submission since this is unclear.

spiffcs: I'll make a follow-up PR when I get more clarity from the governing body. As of right now, they accepted the proposal as is. Edits are welcome should this turn out to be more strict than we need for the version constraints.

PR #623, opened 2021-11-10
Introduce minimal source coordinates

wagoodman:
Today we have `source.Location` to represent any and all location information. This includes the real path (path with no links), virtual path (path that may have links), filesystem ID (image layer ID), and `file.Reference` resolved from stereoscope. All of this information is useful in different circumstances. 

However, there are critical use cases for having an object that has the minimal amount of information to be able to find a location within a source. This minimal information is real path and filesystem ID. The use case in question is being able to craft an index used within `sbom.SBOM.Artifacts` without necessarily having virtual path or a `file.Reference`. If you've indexed a location with a virtual path but want to access that index with only real path information then it would not be possible without creating additional indexes.

This PR introduces a new `source.Coordinates` object that contains this minimal information and changes all maps within `sbom.SBOM.Aritfacts` to use `source.Coordinates` as keys (instead of `source.Location`).

This PR helps to unblock anchore/grype#395 and #476 while further helping along #556.

~Note: this PR base branch is `first-class-relationships` which is under review under PR #607 . This PR will remain in draft until #607 is merged.~

### DIFF
--- a/syft/source/coordinates.go
+++ b/syft/source/coordinates.go
@@ -1,0 +1,34 @@
+package source
+
+import (
+	"fmt"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/artifact"
+)
+
+// Coordinates contains the minimal information needed to describe how to find a file within any possible source object (e.g. image and directory sources)
+type Coordinates struct {
+	RealPath     string `json:"path"`              // The path where all path ancestors have no hardlinks / symlinks
+	FileSystemID string `json:"layerID,omitempty"` // An ID representing the filesystem. For container images, this is a layer digest. For directories or a root filesystem, this is blank.
+}
+
+func (c Coordinates) ID() artifact.ID {
+	f, err := artifact.IDFromHash(c)
+	if err != nil {
+		// TODO: what to do in this case?
+		log.Warnf("unable to get fingerprint of location coordinate=%+v: %+v", c, err)
+		return ""
+	}
+
+	return f
+}
+
+func (c Coordinates) String() string {
+	str := fmt.Sprintf("RealPath=%q", c.RealPath)
+
+	if c.FileSystemID != "" {
+		str += fmt.Sprintf(" Layer=%q", c.FileSystemID)
+	}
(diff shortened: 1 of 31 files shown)


spiffcs (approved):

syft/source/coordinates.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/artifact"
+)
+
+// Coordinates contains the minimal information needed to describe how to find a file within any possible source object (e.g. image and directory sources)
+type Coordinates struct {
+	RealPath     string `json:"path"`              // The path where all path ancestors have no hardlinks / symlinks
+	FileSystemID string `json:"layerID,omitempty"` // An ID representing the filesystem. For container images this is a layer digest, directories or root filesystem this is blank.
```
luhring, line 13: nit (trying to get this a bit more clear)
```suggestion
	FileSystemID string `json:"layerID,omitempty"` // An ID representing the filesystem. For container images, this is a layer digest. For directories or a root filesystem, this is blank.
```

syft/source/location.go
```diff
@@ -3,28 +3,44 @@ package source
...
+func NewLocation(realPath string) Location {
 	return Location{
-		RealPath: path,
+		Coordinates: Coordinates{
+			RealPath: realPath,
+		},
+	}
+}
+
+func NewVirtualLocation(realPath, virtualPath string) Location {
```
luhring, line 32: nit: Can we provide a doc comment for this?

syft/source/location.go
```diff
@@ -3,28 +3,44 @@ package source
...
+	return Location{
+		Coordinates: Coordinates{
+			RealPath: realPath,
+		},
+		VirtualPath: virtualPath,
+	}
+}
+
+func NewLocationFromCoordinates(coordinates Coordinates) Location {
```
luhring, line 41: nit: Can we provide a doc comment for this?

luhring (approved):
Looks cool!

Issue #287, opened 2020-12-08
Analysis of "docker" image fails


Overcolor:
**What happened**:
I tried to scan *docker:20.10.0-rc2-git*, but Syft wanted to pull *20.10.0-rc2-git*:
`[0002] ERROR failed to catalog: could not fetch image '20.10.0-rc2-git': pull failed: Error response from daemon: pull access denied for 20.10.0-rc2-git, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

**What you expected to happen**:
Analysis of the *docker:20.10.0-rc2-git* image.

**How to reproduce it (as minimally and precisely as possible)**:
```
docker pull docker:20.10.0-rc2-git
syft docker:20.10.0-rc2-git
```
**Anything else we need to know?**:
The same error occurs with Grype.

**Environment**:
- Output of `syft version`:
Application:   syft
Version:       0.9.2
BuildDate:     2020-12-03T21:40:41Z
GitCommit:     f87c59b4eb5d5b1396dcc8ad06a07a0706c4b7c5
GitTreeState:  clean
Platform:      linux/amd64
GoVersion:     go1.14.12
Compiler:      gc
- OS (e.g: `cat /etc/os-release` or similar):
Ubuntu 20.04.1

pvnovarese:
fails the same way with any tag in the `docker` repo

slightly different failure if you try `docker` with no tag:

```
pvn@gyarados ~# syft docker
 ✔ Loaded image
[0009] ERROR failed to catalog input: could not fetch image 'docker': tarball must contain only a single image to be used with tarball.Image
```

luhring:
Hi there, thanks for submitting this! You're absolutely right, we need to make an adjustment here.

For completeness, here's what's going on. Syft (and grype) allow users to specify "schemes" in their input string, as a means of directing the tool with an explicit strategy for parsing their input. You can see our schemes by running `syft help`, for example:

```
...
You can also explicitly specify the scheme to use:
    syft docker:yourrepo/yourimage:tag          explicitly use the Docker daemon
    syft docker-archive:path/to/yourimage.tar   use a tarball from disk for archives created from "docker save"
    syft oci-archive:path/to/yourimage.tar      use a tarball from disk for OCI archives (from Podman or otherwise)
...
```

Of course, "docker" is not just a syft scheme, it's the name of a valid container image, which a user might want to inspect.

[...]

PR #626, opened 2021-11-12
287 - Add retry for image fetch using user's input

spiffcs:
Add a second call to `GetImageFromSource` if a prefix was aggressively parsed causing an image name to be shortened into an invalid image.

Docker image cases now covered:
- docker:20.10.0-rc2-git
- docker:docker:20.10.0-rc2-git

#287 (Can't link the issue for some reason)

Had to break things into smaller components because of the linter.


luhring (approved):
Love the functional decomposition! 💯

PR #628, opened 2021-11-15
upgrade docker version fix high vuln

fengshunli:


wagoodman:
Thanks @fengshunli , though, we just merged https://github.com/anchore/syft/pull/633 which bumps docker to v20.10.10 (following up from your stereoscope PR 🙌 )

PR #634, opened 2021-11-17
Migrate SPDX-JSON relationships to SBOM model

### LINKED ISSUE #556: Prefer artifact relationships over package relationships

wagoodman:
This is a follow up to #607 and #623 
Closes #556 

Changes made:
- Removes the creation of relationships in the `spdx22json` format encoder and moves file-ownership relationship creation back to the main `Catalog` function.
- Adds a new `sbom.AllCoordinates` that returns a slice of all coordinates referenced throughout all SBOM artifacts and relationships. This is needed to guarantee inclusion of all referenced files as nodes in the SBOM.
- The power-user JSON document data structure has been replaced with the `syftjson` format model and the `poweruser.Presenter` has been deleted (as well as the remainder of `internal/presenters` 🎉 🌮 )
- Combines file classifications, metadata, and contents fields for a location into a single entry. For this reason the `fileMetadata`, `fileClassifications`, and `fileContents` power-user JSON fields have been removed and replaced with a single `files` section. This requires bumping the JSON schema version from 1 to 2 .
- Adds the application configuration to the output of the `packages` output (for `syftjson` format). This now matches the same behavior as power-user.

Notes: 
- `schema/json/schema-2.0.0.json` is not intended for manual review (~1000 lines).
- this PR will remain in draft until #623 is merged.

Questions:
- a future PR could implement `syft1json` and `syft2json` formats for backwards compatibility, though this is not planned at this time. **Should we do this before bumping to schema version 2?**
- What should we do in cases where a package manager claims ownership of a file that **does not exist** (relative to the source.FileResolver)? Should we log and continue (what is implemented now)? Or make note of this in the SBOM?


wagoodman:
Merging #623 has caused a lot of merge conflicts and rebasing that is needed (note the base branch changed, so the diff is way off now). I'll adjust shortly.

wagoodman:
ok, rebased and cleaned up! 🤞

internal/formats/syftjson/model/document.go
```diff
@@ -4,10 +4,12 @@ package model
 type Document struct {
 	Artifacts             []Package      `json:"artifacts"` // Artifacts is the list of packages discovered and placed into the catalog
 	ArtifactRelationships []Relationship `json:"artifactRelationships"`
-	Source                Source         `json:"source"`     // Source represents the original object that was cataloged
-	Distro                Distro         `json:"distro"`     // Distro represents the Linux distribution that was detected from the source
-	Descriptor            Descriptor     `json:"descriptor"` // Descriptor is a block containing self-describing information about syft
-	Schema                Schema         `json:"schema"`     // Schema is a block reserved for defining the version for the shape of this JSON document and where to find the schema document to validate the shape
+	Files                 []File         `json:"files,omitempty"`   // note: must have omitempty
```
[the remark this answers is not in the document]
spiffcs: Let's see if we can sync and nail down how/why `Files` should be a sibling field to `Artifacts` -  I think I understand, but it would be good to codify it somewhere so we can be very explicit on why they are different.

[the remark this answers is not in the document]
spiffcs: Are all `File` `Artifact`, but not all `Artifact` `File`?

internal/formats/syftjson/to_format_model.go
```diff
@@ -26,6 +30,8 @@ func ToFormatModel(s sbom.SBOM, applicationConfig interface{}) model.Document {
 	return model.Document{
 		Artifacts:             toPackageModels(s.Artifacts.PackageCatalog),
 		ArtifactRelationships: toRelationshipModel(s.Relationships),
+		Files:                 toFile(s),
```
spiffcs, line 33: More to my point about nailing the distinction. A little weird that `Files` takes the entire sbom, but `Secrets` below it has a subsection from `Artifacts` that we lift to the top level of this struct.

internal/formats/table/encoder.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/olekukonko/tablewriter"
 )
 
-func encoder(output io.Writer, s sbom.SBOM) error {
+func encoder(output io.Writer, s sbom.SBOM, _ interface{}) error {
```
spiffcs, line 14: Noted comment about `function` type vs making this an `interface` from before

spiffcs (commented):
First pass - Stopped at `mimetype_helper` and will dig in more after lunch - 

To answer questions on the PR:
1) I think we should implement backwards compatibility before bumping the schema version. It's easy to lose that thread and let changes get ahead of the opportunity for us to do it now.

2) If there is metadata from a package manager that suggests and file exists, but we have read the entire tree and concluded it doesn't exist because some future build action had removed it then including it in the document might cause a kind of SBOM false positive. On the other hand, if we support annotations I think we can include that message saying - `package manager says x should exist, but we concluded it has been removed`

I think the incorrect path is including it with no context. Either do what is implemented now, or explore annotation land.

internal/formats/syftjson/to_format_model.go
```diff
@@ -40,6 +46,85 @@ func ToFormatModel(s sbom.SBOM, applicationConfig interface{}) model.Document {
 	}
 }
 
+func toSecrets(data map[source.Coordinates][]file.SearchResult) []model.Secrets {
```
spiffcs, line 49: `map[source.Coordinates][]file.SearchResult`

Is there any value in codifying these as concreate types beyond the assembly of their basic types?

Found as part of Sbom here:
https://github.com/anchore/syft/blob/4f0099583ab5b2030188de0b15367011b020957b/syft/sbom/sbom.go#L17-L25

`map[source.Coordinates][]file.SearchResult => someIndex`?

Only asking to see if additional methods could be added  which makes working with these types easier

syft/pkg/cataloger/catalog.go
```diff
@@ -67,6 +69,14 @@ func Catalog(resolver source.FileResolver, theDistro *distro.Distro, catalogers
 			// generate PURL
 			p.PURL = generatePackageURL(p, theDistro)
 
+			// create file-to-package relationships for files owned by the package
+			owningRelationships, err := packageFileOwnershipRelationships(p, resolver)
```
spiffcs, line 73: I know catalog is being given the responsibility of building these relationships so are we keeping it around going forward? Seems like an ok spot to land for relationship generation

syft/pkg/cataloger/golang/binary_cataloger.go
```diff
@@ -16,15 +16,6 @@ import (
 
 const catalogerName = "go-module-binary-cataloger"
 
-// current mime types to search by to discover go binaries
```
spiffcs, line 19: Nice glad these are going to their own thing. Syft having mime-type destructed to its own thing is a great level up

spiffcs (commented):
More comments

To Review:
- to_format_model.go (each implementation)
- new mimetype_helper
- sbom.go coordinate additions
- CoordinateSet and its use

internal/formats/spdx22json/to_format_model.go
```diff
@@ -113,16 +122,145 @@ func extractFromCatalog(catalog *pkg.Catalog) ([]model.Package, []model.File, []
...
+	}
+	return checksums
+}
+
+func toFileTypes(metadata *source.FileMetadata) (ty []string) {
+	if metadata == nil {
+		return nil
+	}
+
+	mimeTypePrefix := strings.Split(metadata.MIMEType, "/")[0]
```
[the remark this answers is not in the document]
wagoodman: indeed --it will either match up with a prefix, or not. Either way there will be at least one value in the slice and we're taking the first value.

internal/formats/spdx22json/to_format_model.go
```diff
@@ -113,16 +122,145 @@ func extractFromCatalog(catalog *pkg.Catalog) ([]model.Package, []model.File, []
...
+			Comment:            comment,
+		})
+	}
+	return result
+}
+
+func lookupRelationship(ty artifact.RelationshipType) (bool, model.RelationshipType, string) {
+	switch ty {
+	case artifact.ContainsRelationship:
+		return true, model.ContainsRelationship, ""
```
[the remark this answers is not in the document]
wagoodman: `Other` was highly suggested by the SPDX spec to have a comment, however, it doesn't seem like adding comments that copy-paste from the SPDX docs will help here (programmatically it does, but from a user/consumer sense it doesn't). Maybe in the future we could provide more context in the comments field, but for now should probably leave it out.

internal/formats/cyclonedx12xml/encoder.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/anchore/syft/syft/sbom"
 )
 
-func encoder(output io.Writer, s sbom.SBOM) error {
+func encoder(output io.Writer, s sbom.SBOM, _ interface{}) error {
```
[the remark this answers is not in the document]
wagoodman: This is a great suggestion and points out a problem that we have already started to get rid of in the code base... I'll take a closer look at options here 👍

internal/formats/syftjson/model/document.go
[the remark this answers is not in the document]
wagoodman: Right now the "artifacts" section is just packages, but in the future it may contain other fields that are sibling fields of "artifacts" today (see https://github.com/anchore/syft/issues/555). Until we tackle #555  it would be consistent to leave `Files` as a sibling of `Artifacts`.

internal/formats/syftjson/to_format_model.go
wagoodman: > A little weird that Files takes the entire sbom...

you're right, though it's still necessary. The short answer from our offline conversation is that the `Artifacts` struct is not guaranteed to have all of the nodes in the graph that can be created from `sbom.SBOM`, some of the edges in `Relationships` describe nodes that a cataloger may not have directly observed (e.g. a package manager claiming that a package owns a file that does not exist on the filesystem). 

At a minimum to describe all files one needs `sbom.Artifacts` and `sbom.Relationships` to get a full picture of all files that will be in the SBOM. Since this is a helper function of the `sbom` package it seemed right from a consumer point of view to only need to pass an SBOM to get the answer to the question "what files are in my SBOM" (as opposed to needing to pass elements of an SBOM).

internal/formats/syftjson/decoder_test.go
```diff
@@ -15,7 +15,7 @@ func TestEncodeDecodeCycle(t *testing.T) {
 	originalSBOM := testutils.ImageInput(t, testImage)
 
 	var buf bytes.Buffer
-	assert.NoError(t, encoder(&buf, originalSBOM))
+	assert.NoError(t, encoder(&buf, originalSBOM, map[string]string{"config": "value"}))
```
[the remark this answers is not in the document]
wagoodman: yeah, I'll play with this while working on https://github.com/anchore/syft/pull/634#discussion_r754564473 ... at the very least I'll provide a comment

internal/formats/syftjson/to_format_model.go
wagoodman: From our offline conversation: I'll try out making types for each of these "map" fields but probably won't add any helper functions for those types.

syft/pkg/cataloger/catalog.go
wagoodman: Cataloging has the ability to raise up relationships when it makes sense (shifting that responsibility "left" towards the catalogers when it has the "most knowledge" about the relationship in question, say, tracking a graph of transitive dependency for a package... only the cataloger has enough information to capture these relationships). Though there will be other times when crafting relationships are better served on the "right" after all cataloging is completed (say when elements from different catalogers may be related).

This will probably shake out over time and we can adjust where this lives as we see better places for it. (for now I'll leave it where it is --though shout out if you see a better spot for it)

internal/formats/spdx22json/to_format_model.go
```diff
@@ -76,19 +90,12 @@ func documentNamespace(name string, srcMetadata source.Metadata) string {
 	return path.Join(anchoreNamespace, identifier)
 }
 
-func extractFromCatalog(catalog *pkg.Catalog) ([]model.Package, []model.File, []model.Relationship) {
+func toPackages(catalog *pkg.Catalog, relationships []artifact.Relationship) []model.Package {
```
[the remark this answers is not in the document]
wagoodman: Passing around a slice of relationships isn't a large problem --even if it's a lot of them, it's s pointer to an underlying array, so I feel pretty OK with this. When it comes to code complexity, I also feel like we're OK here... specifically `fileIDSForPackage` is filtering out the set of relationships to only those that this package takes a part of and with the appropriate relationship type.

I do feel like there may be room for improvement as to where the "relationship helper function" go. This one is ad-hoc in a format-related package. That can be dealt with in the future after this PR.

internal/formats/cyclonedx12xml/encoder.go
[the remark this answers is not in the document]
wagoodman: TL;DR: I ended up adding an `sbom.Descriptor` that is added to `sbom.SBOM` that contains the application configuration, the tool name, and tool version.

I needed to refresh myself on the input going into the design for format encoding and decoding. The first approach taken (but not used) was to introduce `Encoder`, `Decoder`, and `Validator` interfaces (I re-mocked it up here https://github.com/anchore/syft/commit/5fb0235cfb71fd968af948a3d21b63ed9d972fd1). This approach looked great, however, the drawback was that you couldn't leverage the pro's of this approach. Specifically, you could push format-specific fields (such as application config) down below the interface and let the concrete implementation deal with these difference for each instance. However, that requires that this data is passed at object construction or between construction and the call to `Encode()`, neither of which is ideal since there would be a need to have a list of all formatters to reference... needing to have the data at this time would be difficult.

Function types that are assigned to the Format struct are the next best thing, since it enforces that state must be passed in at encode-time, and not earlier.

I then started to look the other way: since injection of this state via behavior seemed wrong (due to the above notes) I started looking for ways to encapsulate the app config. I had initially thrown out encoding it in `sbom.SBOM` and instead started looking into making a new envelope (something like `format.Context` that would contain `sbom.SBOM` and the application config), however, the more I poked at this the more sane it seemed to put the application config into the SBOM itself.

In syft json, CycloneDX, and SPDX, there are sections that describe the tool that created the SBOM. Now that there are encode and decode workflows we must persist this information in the model or else we will loose it in decoding operations. 

Consider an encode-decode-encode workflow with an old version of syft to a new version of syft... we encode the original SBOM with the old version of syft followed by a decode and encode with the new version of syft. This would be how you convert from an older schema version to a new schema version. Without persisting the original tool name, version, and configuration, we would have no record on how the information was created to begin with. It should only be the `schema` section that changes (to indicate the new format) but not the tool information. 

For these reasons I went the route of adding a new `sbom.Descriptor` section that persists name, version, and optional configuration for the tool that created the SBOM.

internal/formats/syftjson/to_format_model.go
wagoodman: I've added the new types didn't make any behavioral changes --the new types at the very least abstract what is in the type so if it changes later (say to a struct) then we don't need to update a bunch of function signatures.

wagoodman: Hmm, well. When I updated the `file` package to use the new `sbom.File*` types, I got this:
```
```

wagoodman:
@spiffcs I've addressed all of your comments --ready for re-review! 🎉 🌮

spiffcs (approved):

PR #635, opened 2021-11-18
add cyclone-json output format

### LINKED ISSUE #535: Swap cycloneDX lib for an official lib
Today we are using an in-house set of struct definitions for CycloneDX support, however, an official lib now exists: https://github.com/CycloneDX/cyclonedx-go . We can remove our in house definitions and use the official lib.

### LINKED ISSUE #631: Add an output type for CycloneDX JSON format

Presently you can `syft packages -o cyclonedx dir:.` and you'll get a CycloneDX report in XML. There are tools to convert it to JSON, but it's an extra step.

It would be helpful there was a `cyclonedx-json` (or something like that) output format that would directly output CycloneDX JSON.

**Why is this needed**:

It would make using `syft` more convenient for use cases that require CycloneDX JSON output. No additional tools or libraries would be required to convert from XML to JSON.

**Additional context**:

The [Buildpacks specification](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md) has been standardized around using JSON output. Many of the buildpacks we are developing are generating SBoMs using `syft`, which leaves us needing to convert from CycloneDX XML to JSON. Having an option for direct output would make our integrations easier.

Thanks!

jonasagx:
closes #631 
closes #535

topics for thought:
* this pr adds cyclone 1.3, should we keep 1.2?
* code was copied, I just repeated the structure for json & xml, please let me know if you hate it
* since moving to cycloneDX 1.3 I upgraded the local schema docs 


spiffcs:
@jonasagx I'll take a gander at this after our demo meetings

jonasagx:
@spiffcs I will redo some of it to use cycloneDX's go library

spiffcs:
> @spiffcs I will redo some of it to use cycloneDX's go library

Alright awesome - just tag me when that's in and I'll then 👀

syft/format/option.go
```diff
@@ -33,6 +35,8 @@ func ParseOption(userStr string) Option {
 		return TableOption
 	case string(CycloneDxOption), "cyclone", "cyclone-dx":
```
wagoodman, line 36: now that we have a JSON flavor of CycloneDX we should probably rename `CycloneDxOption` to `CycloneDxXMLOption` and provide additional parser variants like `cyclone-dx-xml` and `cyclone-xml`

wagoodman (approved):
Looks great!! Though the diff says otherwise, this is actually removing code to _add_ a lot of functionality, which is always awesome 💯 ❗  (the copies of the schemas and test fixtures just makes it look like we're adding more).

wagoodman (commented):
Just found this one... can you update `schema/cyclonedx/.gitignore` to include `bom.json` so we don't accidentally commit it?

internal/formats/common/cyclonedxhelpers/format.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/anchore/syft/syft/sbom"
+	"github.com/anchore/syft/syft/source"
+	"github.com/google/uuid"
+)
+
+func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
+	cdxBOM := cyclonedx.NewBOM()
+	versionInfo := version.FromBuild()
+
+	cdxBOM.SerialNumber = uuid.New().URN()
```
spiffcs, line 19: I think a small comment about the specification requiring the URN format might be nice to show this was not an arbitrary selection for the SerialNumber

internal/formats/common/cyclonedxhelpers/format.go
```diff
@@ -0,0 +1,90 @@
...
+	cdxBOM.Metadata = toBomDescriptor(internal.ApplicationName, versionInfo.Version, s.Source)
+
+	packages := s.Artifacts.PackageCatalog.Sorted()
```
spiffcs, line 22: Do we want this function to guarantee the sort, or should a user bring their own sorted packages already? This could prevent duplicate sort calls and not punish a consumer for having sorted previously in an upstream process.

internal/formats/common/cyclonedxhelpers/format.go
```diff
@@ -0,0 +1,90 @@
...
+	components := make([]cyclonedx.Component, len(packages))
+	for i, p := range packages {
+		components[i] = toComponent(p)
+	}
+	cdxBOM.Components = &components
```
spiffcs, line 27: I'm not sure if it's equivlant, but if line 23 does
```
cdxBOM.Components = make([]cyclonedx.Component, len(packages))
```

We can cut out the middle man and just append directly to the struct?
```
cdxBOM.Components = appened(cdxBOM.Components, toComponent(p))
```

This is where I don't have a preference really, but am curious about what people think from a general programming sense. Especially if append vs index assignment is faster when using Go.

internal/formats/common/cyclonedxhelpers/format.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return nil
+}
+
+func toLicenses(ls []string) *cyclonedx.Licenses {
+	if len(ls) == 0 {
+		return nil
+	}
+
+	lc := make(cyclonedx.Licenses, len(ls))
```
spiffcs, line 80: Same comment here, index assignment vs appened. When do we use one vs the other? I know we use both in this codebase so clarity on the differences and choice would help when doing PR review.

internal/formats/common/testutils/utils.go
```diff
@@ -87,6 +89,8 @@ func AssertPresenterAgainstGoldenSnapshot(t *testing.T, pres presenter.Presenter
 	if !bytes.Equal(expected, actual) {
 		dmp := diffmatchpatch.New()
 		diffs := dmp.DiffMain(string(expected), string(actual), true)
+		t.Logf("len: %d\nexpected: %s", len(expected), expected)
```
spiffcs, line 92: I see we're still using `diffmatchpath` here. 

I really love https://pkg.go.dev/github.com/google/go-cmp/cmp for these kinds of large diffs. I think the output is superior to us having to log our own `len` and do the `bytes.Equal` comparison.

The diffTesting example is pretty expressive:
https://pkg.go.dev/github.com/google/go-cmp/cmp#example-Diff-Testing

Don't have to change if you like `dmp`. I'm just throwing this library into the ring for consideration.

internal/formats/cyclonedx13json/encoder_test.go
```diff
@@ -0,0 +1,39 @@
...
+	testutils.AssertPresenterAgainstGoldenImageSnapshot(t,
+		Format().Presenter(testutils.ImageInput(t, testImage)),
+		testImage,
+		*updateCycloneDx,
+		cycloneDxRedactor,
+	)
+}
+
+func cycloneDxRedactor(s []byte) []byte {
+	serialPattern := regexp.MustCompile(`"serialNumber": "[a-zA-Z0-9\-:]+"`)
```
spiffcs, line 32: Is this for the serial number? Their specification has the official regex as if you want the whole things REDACTED
```
^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$
```

spiffcs: https://cyclonedx.org/docs/1.3/json/#serialNumber

internal/formats/cyclonedx13xml/test-fixtures/snapshot/TestCycloneDxDirectoryPresenter.golden
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<bom xmlns="http://cyclonedx.org/schema/bom/1.2" version="1" serialNumber="urn:uuid:5404937f-72d6-44a2-8e9b-954305ecb4f6">
+<bom xmlns="http://cyclonedx.org/schema/bom/1.3" serialNumber="371e31c0-d99c-4816-986e-5485246b47e3" version="1">
```
spiffcs, line 2: Is this new serial number correct for 1.3?
https://cyclonedx.org/docs/1.3/json/#serialNumber

internal/formats/cyclonedx13xml/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<bom xmlns="http://cyclonedx.org/schema/bom/1.2" version="1" serialNumber="urn:uuid:e34bad2e-cd27-483c-86dc-f4e26d6103b0">
+<bom xmlns="http://cyclonedx.org/schema/bom/1.3" serialNumber="34b83585-396b-47d5-8e7c-c78eda1fc83d" version="1">
```
spiffcs, line 2: Same question RE serial number as above

spiffcs (commented):
General comments:
- Nice work here all these changes are awesome
- Why is there an spdx.xsd file under `schema/cyclonedx/`?
- +1 on Alex's comments regarding option naming and .gitignore
-

wagoodman:
@spiffcs  The short answer for...

> Why is there an spdx.xsd file under schema/cyclonedx/?

...is because the CycloneDX schema references the SPDX schema directly: https://github.com/anchore/syft/pull/635/files#diff-b31077573587a0c621de4b49d380b3a97e609a39299e35b6d688b50d64871aa3L27 . During schema validation it is useful to have them both locally persisted.

internal/formats/common/cyclonedxhelpers/format.go
wagoodman: We do want to guarantee sorting here. Why? Overall we want to have the output documents be as stable as possible; the same bytes being encoded twice should result in the same serialized bytes. Since the package catalog is a map under the hood the order of packages is not guaranteed to be stable, so sorting is necessary in the format-model domain.

internal/formats/common/cyclonedxhelpers/format.go
wagoodman: In general I'm a +0 in the category of performance between these options... for readability I'm a +1 for your suggestion, @spiffcs in general. However, in this case I don't think that'll work; the field on the struct isn't a slice, it's a pointer to a slice:
```
type BOM struct {
	...
	Components         *[]Component         `json:"components,omitempty" 
        ...
}
```

internal/formats/common/cyclonedxhelpers/format.go
wagoodman: `<reflection on the meta question about reviews>`

I think both are OK and valid in different circumstances (my take is that this is in the "up to the PR author" thing, but not out of bounds for review comment... lots of space for "depending on the context"). Technically via assignment has fewer allocations, but again, I'm a +0 on that concern. 

More 2 🪙 ... from a review point of view, we are trying to prefer code in that is:
1. harder to abuse / misuse 
2. resistant to future problems / brittleness 
3. easy to read / grok / understand

These are all mostly subjective, but are more clear when comparing implementation alternatives.

Take for instance incorporating @spiffcs suggestion:

```
func toLicenses(ls []string) *cyclonedx.Licenses {
	var lc cyclonedx.Licenses
	for _, licenseName := range ls {
		lc = append(lc, cyclonedx.LicenseChoice{
			License: &cyclonedx.License{
				Name: licenseName,
			},
		})
	}

	return &lc
}
```

The upside about this is that the implementation is shorter without increasing the "comprehensional density" (warning: made up phrase)... that is, I don't need to "unpack" extra hidden behavior from this alternative implementation. But it expresses the same behavior in a smaller space with fewer concerns (e.g. no need for a guard clause, better default nil behavior, one less variable to worry about).

That all being said, I personally consider comments about these kinds of things to be in the `nit` category. What I specifically mean by a "nit" comment is a "take it or leave it, optional comment" aimed solely at potential improvements to the code but do not block the approval process --it's up to the PR author to moderate the comment (incorporate or not).

`</reflection>`

internal/formats/common/testutils/utils.go
wagoodman: I like both cmd and dmp. The upside with dmp is that you get the entire output formatted with a colorized diff. The upside with cmp is that you get a summary of the changes in a typical patch format. 

The pro to dmp is that you get the entire document which gives you full context as to where the problem might be. That is also the downside... you get the full document, which could be rather large! The pro to cmp is that it gives you a summary, but that is also it's downside... sometimes the summary doesn't provide enough context to figure the exact problem.

(I've been pretty torn on this one)

spiffcs:
After reading @wagoodman comments I'm good with this PR. The only outstanding question I still have is the `urn` format found in the specification vs the one found during our snapshot generation.

internal/formats/cyclonedx13xml/test-fixtures/snapshot/TestCycloneDxDirectoryPresenter.golden
jonasagx: I was not right, the integration test also complained

internal/formats/cyclonedx13xml/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
jonasagx: Fixed

Issue #246, opened 2020-10-28
Full cataloging of archive contents


luhring:
**What would you like to be added**: The ability for Syft to inspect the contents of archive files (tar, zip, jar/war/etc.) using all catalogers, such that any inner contents can be surfaced to the user.

**Why is this needed**: There's a general case to be made, as well as a more specific one.

The general case is that inspecting archive contents recursively means that Syft is diving deeper into the given image's or directory's content. Since Syft's prime directive is to "uncover what's in an image [or directory]", the ability to "see into" the file tree encapsulated by archive files allows Syft to more completely fulfill its mission, which means users get a more complete understanding of what they're building/shipping/deploying/etc. This is true today while Syft focuses on packages, and it becomes even more true as Syft is extended to discover secrets, specific files, etc.

The specific case is that the nature of Java packages is that the on-disk format can (and does) contain files of arbitrary types. This means that a JAR file can easily contain more "packages" than just Java applications and libraries. For example, a web application can have JavaScript dependencies, even if the web application is packaged as a JAR file. Ideally, a Java app that uses a JS package that has a reported vulnerability will, itself, have a reported vulnerability. But this isn't always the case, and the richer the information Syft produces about a given package, the more possibilities open up for identification and remediation of problems.

wagoodman:
This could leverage the new location-by-MIME-type options added in https://github.com/anchore/syft/pull/529 . 

One challenge is that this cataloger should try to not overlap with other catalogers (such as python wheels and java jars) or risk raising up duplicate packages.

luhring:
There are some interesting considerations around implementation, I think.

One approach would be that this feature _does not_ introduce a new cataloger, and instead provides our existing catalogers what they need to catalog within archives. The ideal situation (IMHO) is that archives can be traversed transparently, as if the file tree just continued into them. This way, we can use our existing cataloger pattern and catalogers to find NPM packages, Python packages, anything really. This would mean that we'd never risk getting duplicate packages.

We have a "resolver" abstraction today that let's catalogers ask questions about files (e.g. via glob patterns), while the cataloger is tasked with surfacing a specific kind of package. It could be that this feature takes place "below" the resolver abstraction as opposed to "above" it.

There are other approaches that would work, too! Just dropping this here for consideration.

wagoodman:
>  It could be that this feature takes place "below" the resolver abstraction as opposed to "above" it.

❤️

edonadei:
[...]

PR #637, opened 2021-11-18
Catalog archive contents for single-file input

wagoodman:
This PR adds support for cataloging the contents of a given archive:
```
# unarchive automatically for input from the "file" scheme
❯ syft file:./hike-gem.tar.gz
...
NAME                               VERSION                         TYPE
hike                               1.2.3                           gem

# unarchive automatically when the "file" scheme is detected
❯ syft ./hike-gem.zip
...
NAME                               VERSION                         TYPE
hike                               1.2.3                           gem

# preserves existing image archive processing
❯ syft ~/oci-images/ubi8.tar
...
NAME                                    VERSION                          TYPE
PyGObject                               3.28.3                           python
PySocks                                 1.6.8                            python
acl                                     2.2.53-1.el8                     rpm
audit-libs                              3.0-0.17.20191104git1c2f876.el8  rpm
basesystem                              11-5.el8                         rpm
```

Supports zip, tar, and rar archive formats and the following compression formats:
- brotli (br)
- bzip2 (bz2)
- flate (zip)
- gzip (gz)
- lz4
- snappy (sz)
- xz
- zstandard (zstd)

Note: this does NOT unarchive contents recursively. That is, if the given archive contains an archive, then the nested archive will be treated as a file and NOT unarchived. Recursive unarchiving should/could be handled under https://github.com/anchore/syft/issues/246.

### DIFF
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -101,10 +101,7 @@ func generateFileSource(fs afero.Fs, location string) (*Source, func(), error) {
 		return &Source{}, func() {}, fmt.Errorf("given path is not a directory (path=%q): %w", location, err)
 	}
 
-	s, cleanupFn, err := NewFromFile(location)
-	if err != nil {
-		return &Source{}, func() {}, fmt.Errorf("could not populate source from path=%q: %w", location, err)
-	}
+	s, cleanupFn := NewFromFile(location)
 
 	return &s, cleanupFn, nil
 }
@@ -122,9 +119,9 @@ func NewFromDirectory(path string) (Source, error) {
 }
 
 // NewFromFile creates a new source object tailored to catalog a file.
-func NewFromFile(path string) (Source, func(), error) {
+func NewFromFile(path string) (Source, func()) {
 	var analysisPath = path
-	var cleanupFn func()
+	var cleanupFn = func() {}
 
 	// if the given file is an archive (as indicated by the file extension and not MIME type) then unarchive it and
 	// use the contents as the source. Note: this does NOT recursively unarchive contents, only the given path is
@@ -148,7 +145,7 @@ func NewFromFile(path string) (Source, func(), error) {
 			Path:   path,
 		},
 		path: analysisPath,
-	}, cleanupFn, nil
+	}, cleanupFn
 }
 
 // NewFromImage creates a new source object tailored to catalog a given container image, relative to the
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -198,7 +198,7 @@ func unarchiveToTmp(path string, unarchiver archiver.Unarchiver) (string, func()
 
 	cleanupFn := func() {
 		if err := os.RemoveAll(tempDir); err != nil {
-			log.Errorf("unable to cleanup archive tempdir: %+v", err)
+			log.Warnf("unable to cleanup archive tempdir: %+v", err)
 		}
 	}
 
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -7,11 +7,14 @@ package source
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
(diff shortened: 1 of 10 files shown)


syft/source/source.go
```diff
@@ -7,21 +7,25 @@ package source
...
 // Source is an object that captures the data source to be cataloged, configuration, and a specific resolver used
 // in cataloging (based on the data source and configuration)
 type Source struct {
 	Image             *image.Image // the image object to be cataloged (image only)
-	DirectoryResolver *directoryResolver
 	Metadata          Metadata
-	Mutex             *sync.Mutex
+	directoryResolver *directoryResolver
+	path              string
+	mutex             *sync.Mutex
```
luhring, line 28: Yay for reduced scope! 🎉

syft/source/source.go
```diff
@@ -168,3 +192,18 @@ func (s *Source) FileResolver(scope Scope) (FileResolver, error) {
...
+
+func unarchiveToTmp(path string, unarchiver archiver.Unarchiver) (string, func(), error) {
+	tempDir, err := ioutil.TempDir("", "syft-archive-contents-")
+	if err != nil {
+		return "", func() {}, fmt.Errorf("unable to create tempdir for archive processing: %w", err)
+	}
+
+	cleanupFn := func() {
+		if err := os.RemoveAll(tempDir); err != nil {
+			log.Errorf("unable to cleanup archive tempdir: %+v", err)
```
luhring, line 204: Should this be a "warn" level error? It looks like we're not bubbling this error up, so the app could still exit 0 here. Is that right?

syft/source/source.go
```diff
@@ -97,34 +101,54 @@ func generateFileSource(fs afero.Fs, location string) (*Source, func(), error) {
...
 			Path:   path,
 		},
+		path: path,
 	}, nil
 }
 
-// NewFromDirectory creates a new source object tailored to catalog a given filesystem directory recursively.
-func NewFromFile(path string) (Source, error) {
+// NewFromFile creates a new source object tailored to catalog a file.
+func NewFromFile(path string) (Source, func(), error) {
```
luhring, line 125: This function never returns an error value. Is `error` part of the signature to meet expectations of an interface/type?

syft/source/source.go
```diff
@@ -97,34 +101,54 @@ func generateFileSource(fs afero.Fs, location string) (*Source, func(), error) {
...
+	var analysisPath = path
+	var cleanupFn func()
+
+	// if the given file is an archive (as indicated by the file extension and not MIME type) then unarchive it and
+	// use the contents as the source. Note: this does NOT recursively unarchive contents, only the given path is
+	// unarchived.
+	envelopedUnarchiver, err := archiver.ByExtension(path)
+	if unarchiver, ok := envelopedUnarchiver.(archiver.Unarchiver); err == nil && ok {
+		unarchivedPath, tmpCleanup, err := unarchiveToTmp(path, unarchiver)
```
luhring, line 134: I'm wondering if the cleanup function value should be captured as `cleanupFn` (already declared), rather than introduce a new variable here. `NewFromFile` returns a cleanup function, it could be useful to return the cleanup task definition that comes from `unarchiveToTmp` in the event of an error being returned from `unarchiveToTmp`. What are your thoughts here?

wagoodman: We only want to replace `cleanupFn` with another function once we are certain that `unarchiveToTmp` has produced a function (there is no error returned).

wagoodman: Thanks for the offline sync --I've updated this section to always ensure that a cleanup function will be provided.

luhring: Looks good, thanks!

luhring (approved):
Nice enhancement! 🙏

PR #639, opened 2021-11-19
Update concurrency primitives for relationships

spiffcs:
Discussion on a separate PR was made around how we populate relationships.

This pr removes a level of concurrency and returns a slice. If we're good with this we can apply it to the `packages` command.

Previous discussion: https://github.com/anchore/syft/pull/636#discussion_r752440947

### DIFF
--- a/cmd/power_user.go
+++ b/cmd/power_user.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"sync"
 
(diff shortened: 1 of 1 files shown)


spiffcs:
As @luhring commented previously this is a net loss of code and a pretty slick simplification regarding the removal of a level of unneeded concurrency and go routine tracking.

luhring (approved):
Nice!

PR #643, opened 2021-11-23
Adds composer installed.json parser

### LINKED ISSUE #642: Support for PHP/composer installed.json files (i.e. syft php image support)
First, thanks all for the great tool and all the hard work. Really fantastic stuff.

Presently, there is no so-called "image support" (as discussed in [here](https://github.com/anchore/syft/pull/609#issuecomment-958140573)) for composer. 

What I would like to see added is support for the `./vendor/composer/installed.json` file, which is a statement of what is _actually_ installed (rather than, with `composer.lock` what _should_ be installed - see this [github discussion](https://github.com/composer/composer/discussions/9686)).

**Why is this needed**:
This is needed because presently we're not scanning images for php-packages.

bomoko:
closes #642 

This is, as far as possible, a copy-pasta from the earlier PHP composer.lock parser. I'm still new to the code base, so I've tried to work out what needs to be changed/added, etc.

Thanks for taking the time to look at this.

signed off by Blaize M. Kaye: blaize.kaye@amazee.com


spiffcs:
Thanks for the PR @bomoko! I'll take a look when I have some free cycles.

wagoodman:
Nice work @bomoko --great follow up to https://github.com/anchore/syft/pull/609. I made a couple of tweaks, but otherwise looks great 💯 :
- Unexported the PHP types used for unmarshaling
- Small rename of the PHP cataloger names

wagoodman (approved):

wagoodman:
One last change is needed before we can merge this, @bomoko;  we require that all commits are signed before merging into the baseline (see [CONTRIBUTING.md](https://github.com/anchore/syft/blob/main/CONTRIBUTING.md#sign-your-commits)) . Shout out if you have any questions about this!

bomoko:
Thanks so much @wagoodman - I _think_ I understand how to sign the previous commits if I rebase and amend with `-S`, does this seem right?
Problem with this is that I believe I end up signing your commits (whether I rebase on the branch directly, or if I cherry-pick into another branch). I'm a little unsure about how best to deal with that (or whether it's even a problem).

Is it a problem that I end up signing your commits? Or do the commits simply need to _be signed_?
Alternatively, I could make a new commit out of yours and sign that.
Any advice you could give here would be great.

wagoodman:
@bomoko indeed rebasing with `-S` will do the trick 👍 (it's OK if you end up signing my commits)

bomoko:
Awesome, thanks @wagoodman - since signing the commit changes the hashes, I think I'm going to need to push force the branch to get the history right. Any objections? (I can also open a new PR)

PR #644, opened 2021-11-24
Fix directory resolver path resolution

wagoodman:
When a single file archive is provided syft had been returning paths relative to the temporary directory created for the archive instead of relative to the root of the archive itself:
```
❯ SYFT_FILE_METADATA_CATALOGER_ENABLED=true syft file:/tmp/syft.tar -o json | jq '.files'
...
  {
    "id": "ee5a4e3ef9a56d72",
    "location": {
      "path": "/var/folders/8x/gw98pp6535s4r8drc374tb1r0000gn/T/syft-archive-contents-770869922/syft/source/source.go"
    },
    "metadata": {
      "mode": 644,
      "type": "RegularFile",
      "userID": 501,
      "groupID": 20,
      "mimeType": "text/plain"
    },
    "digests": [
      {
        "algorithm": "sha256",
        "value": "f101eac52429e3fc2da43ffbb80bdacba6b8b80b9703470b2ff9d7218e9bc9ad"
      }
    ]
  },

...
```

Note the path is `/var/folders/8x/gw98pp6535s4r8drc374tb1r0000gn/T/syft-archive-contents-770869922/syft/source/source.go` instead of `syft/source/source.go`.

This PR fixes this behavior:
```
❯ SYFT_FILE_METADATA_CATALOGER_ENABLED=true syft file:/tmp/syft.tar -o json | jq '.files'
...
  {
    "id": "c45caba6a786688c",
    "location": {
      "path": "syft/source/source.go"
    },
    "metadata": {
      "mode": 644,
      "type": "RegularFile",
      "userID": 501,
      "groupID": 20,
      "mimeType": "text/plain"
    },
    "digests": [
      {
        "algorithm": "sha256",
        "value": "f101eac52429e3fc2da43ffbb80bdacba6b8b80b9703470b2ff9d7218e9bc9ad"
      }
    ]
  },
...
```

### DIFF
--- a/syft/source/directory_resolver.go
+++ b/syft/source/directory_resolver.go
@@ -33,10 +33,11 @@ type pathFilterFn func(string) bool
 
 // directoryResolver implements path and content access for the directory data source.
 type directoryResolver struct {
-	path     string
-	cwd      string
-	fileTree *filetree.FileTree
-	metadata map[file.ID]FileMetadata
+	path              string
+	cwdRelativeToRoot string
+	cwd               string
+	fileTree          *filetree.FileTree
+	metadata          map[file.ID]FileMetadata
 	// TODO: wire up to report these paths in the json report
 	pathFilterFns  []pathFilterFn
 	refsByMIMEType map[string][]file.Reference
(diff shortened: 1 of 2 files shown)


syft/source/directory_resolver.go
```diff
@@ -212,7 +226,9 @@ func (r directoryResolver) requestPath(userPath string) (string, error) {
 func (r directoryResolver) responsePath(path string) string {
 	// always return references relative to the request path (not absolute path)
 	if filepath.IsAbs(path) {
-		return strings.TrimPrefix(path, r.cwd+string(filepath.Separator))
+		// we need to account for the cwd relative to the running process and the given root for the directory resolver
+		prefix := filepath.Clean(filepath.Join(r.cwd, r.cwdRelativeToRoot))
```
spiffcs, line 230: If someone runs against a linux box on windows I think we run into the issue where this will not do the paths correctly right?

wagoodman: I don't think I understand your specific point (yet). The directory resolver is only used against the host's filesystem and not container images. This means that we shouldn't need to worry about cross-platform concerns such as using posix paths for a linux container image being scanned on a windows host. In all cases the scan will be against (and thus relative to) the hosts filesystem paradigm.

spiffcs: That's my fault - I thought that this could double as doing a directory resolution within a container as well.

spiffcs (approved):

PR #649, opened 2021-12-01
Fix SPDX namespace value

### LINKED ISSUE #622: SPDX2.2 JSON format should not use UUID random
**What happened**:
When generating an SPDX 2.2 format SBOM, the `documentName` function should not have a `uuid` as the base return case.

https://github.com/anchore/syft/blob/bb0f35bac4e2e610e61be55b27b3810099794479/internal/formats/spdx22json/to_format_model.go#L49-L59

I expect the base case name to return a string that is in line with the specification found [here](https://spdx.github.io/spdx-spec/document-creation-information/#64-document-name-field)

wagoodman:
The current SPDX namespace values being produced by syft look like:
```
# note the "https:/" instead of "https://"
https:/anchore.com/syft/...
```

This PR fixes this behavior, moves name and namespace related helpers to the `spdxhelpers` package, and adds assertion around testing to ensure that `source.Scheme`s are fully tested (while writing tests for this it was noticed that cases for `FileScheme` were missing).

Additionally closes #622

### DIFF
--- a/internal/formats/common/spdxhelpers/document_name_test.go
+++ b/internal/formats/common/spdxhelpers/document_name_test.go
@@ -1,0 +1,70 @@
+package spdxhelpers
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/anchore/syft/syft/source"
+	"github.com/scylladb/go-set/strset"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_DocumentName(t *testing.T) {
+	allSchemes := strset.New()
+	for _, s := range source.AllSchemes {
+		allSchemes.Add(string(s))
+	}
+	testedSchemes := strset.New()
+
+	tests := []struct {
+		name        string
+		inputName   string
+		srcMetadata source.Metadata
+		expected    string
+	}{
+		{
+			name:      "image",
+			inputName: "my-name",
+			srcMetadata: source.Metadata{
+				Scheme: source.ImageScheme,
+				ImageMetadata: source.ImageMetadata{
+					UserInput:      "image-repo/name:tag",
+					ID:             "id",
+					ManifestDigest: "digest",
+				},
+			},
+			expected: "image-repo/name-tag",
+		},
+		{
+			name:      "directory",
+			inputName: "my-name",
+			srcMetadata: source.Metadata{
+				Scheme: source.DirectoryScheme,
+				Path:   "some/path/to/place",
+			},
+			expected: "some/path/to/place",
+		},
+		{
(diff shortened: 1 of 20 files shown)


internal/formats/common/spdxhelpers/document_namespace_test.go
```diff
@@ -0,0 +1,70 @@
...
+				Scheme: source.FileScheme,
+				Path:   "some/path/to/place",
+			},
+			expected: "https://anchore.com/syft/file/my-name-",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			actual := DocumentNamespace(test.inputName, test.srcMetadata)
+			assert.True(t, strings.HasPrefix(actual, test.expected), fmt.Sprintf("actual namespace %q", actual))
```
luhring, line 61: Are we asserting via `strings.HasPrefix` because of the UUIDs being appended to these strings?

internal/formats/common/spdxhelpers/document_name_test.go
```diff
@@ -0,0 +1,73 @@
...
+			require.NoError(t, err)
+			assert.True(t, strings.HasPrefix(actual, test.expected), fmt.Sprintf("actual name %q", actual))
+
+			// track each scheme tested (passed or not)
+			testedSchemes.Add(string(test.srcMetadata.Scheme))
+		})
+	}
+
+	// assert all possible schemes were under test
+	assert.ElementsMatch(t, allSchemes.List(), testedSchemes.List(), "not all source.Schemes are under test")
```
luhring, line 72: Nice idea 👍

luhring (commented):
Looks good! Just had a question for you

internal/formats/common/spdxhelpers/document_namespace_test.go
wagoodman: Indeed 👍

internal/formats/common/spdxhelpers/document_name_test.go
wagoodman: Thanks! I try to keep collections of things under test in a way where there is an automatic failure when new items are added to the collection. This one snuck by me.

internal/formats/common/spdxhelpers/document_namespace_test.go
luhring: (non-blocking) It _may_ help to have a comment about that, it took me a couple minutes to realize that, but that could have just been me

luhring (approved):

PR #651, opened 2021-12-01
Fix content access for directory resolver

wagoodman:
The directory resolver `FileContentsByLocation` method currently does not rely on `file.Reference` to fetch content, which may lead to empty responses when infact there is content available.


spiffcs:
Paired on this together approved

spiffcs (approved):

PR #657, opened 2021-12-08
Add cataloging of macho multi-architecture binaries

wagoodman:
Adds the ability to catalog [Macho multiarchitecture binaries](https://en.wikipedia.org/wiki/Fat_binary) that are built by the go compiler.

Additionally this PR:
- raises up the architecture for a given package within a binary in the package metadata section
- adds the go binary metadata struct to the JSON schema

This relates to https://github.com/goreleaser/goreleaser/pull/2648#discussion_r765129074

### DIFF
--- a/syft/pkg/cataloger/golang/exe.go
+++ b/syft/pkg/cataloger/golang/exe.go
@@ -25,6 +25,7 @@ type exe interface {
 	// ReadData reads and returns up to size byte starting at virtual address addr.
 	ReadData(addr, size uint64) ([]byte, error)
 
+	// ArchName returns a string that represents the CPU architecture of the executable.
 	ArchName() string
 
 	// DataStart returns the writable data segment start address.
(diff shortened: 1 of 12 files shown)


syft/pkg/cataloger/golang/exe.go
```diff
@@ -57,23 +60,35 @@ func openExe(file io.ReadCloser) (exe, error) {
...
 	if bytes.HasPrefix(data, []byte("\xFE\xED\xFA")) || bytes.HasPrefix(data[1:], []byte("\xFA\xED\xFE")) {
 		e, err := macho.NewFile(f)
 		if err != nil {
 			return nil, err
 		}
-		return &machoExe{file, e}, nil
+		return []exe{&machoExe{file, e}}, nil
+	}
+
+	if bytes.HasPrefix(data, []byte("\xCA\xFE\xBA\xBE")) || bytes.HasPrefix(data[1:], []byte("\xCA\xFE\xBA\xBF")) {
```
luhring, line 82: It'd be good to add a sense of meaning to this `if` line in order to motivate these `bytes.HasPrefix` calls for the reader. If these conditions "mean" something, that'd be great as a factored out function, named for what it's doing. WDYT?

syft/pkg/cataloger/golang/exe.go
```diff
@@ -24,6 +25,8 @@ type exe interface {
 	// ReadData reads and returns up to size byte starting at virtual address addr.
 	ReadData(addr, size uint64) ([]byte, error)
 
+	ArchName() string
```
luhring, line 28: nit: It'd be nice to have a doc comment here

luhring (commented):
I tried it out and it works like a charm! See comments on code specifics, curious for your thoughts 😃

spiffcs (commented):
LGTM - 👍 to @luhring comments, but nothing else to add

syft/pkg/cataloger/golang/exe.go
wagoodman: I can do that for the new code added, however, for code that was copied out of the stdlib I have a preference to leave as is (otherwise when we incorporate new stdlib code we need to re-re-factor it, which isn't fun).

wagoodman:
Looks like in the future we might be able to get this from exported stdlib code (not released yet though, so not today) https://github.com/golang/go/commit/b37a5391f9e452aa779205add12bd89f44e3fcf0

luhring (approved):

PR #663, opened 2021-12-10
Recover from panics from the stdlib when cataloging malformed binaries

wagoodman:
Addresses anchore/grype#526

When processing malformed binaries the stdlib does not take into consideration several edge cases, resulting in a panic instead of returning an error. When this happens grype (or syft) will exit with an error. 

This PR attempts to recover from any panics from this specific execution path until these cases can be dealt with within the stdlib. This allows for syft/grype to continue cataloging paths that didn't have problems and still produce a report instead of exiting early.

### DIFF
--- a/syft/pkg/cataloger/golang/parse_go_bin_test.go
+++ b/syft/pkg/cataloger/golang/parse_go_bin_test.go
@@ -1,6 +1,7 @@
 package golang
 
 import (
+	"io"
 	"testing"
 
 	"github.com/anchore/syft/syft/pkg"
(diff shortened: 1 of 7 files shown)


syft/pkg/cataloger/golang/parse_go_bin_test.go
```diff
@@ -144,17 +145,38 @@ func TestBuildGoPkgInfo(t *testing.T) {
...
+		wantErr  assert.ErrorAssertionFunc
+	}{
+		{
+			name: "recovers from panic",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			pkgs, err := parseGoBin(source.NewLocation("some/path"), nil, freakOut)
+			assert.Error(t, err)
```
spiffcs, line 178: Interesting, so we still want this inability to parse the binary to bomb out the whole process? I thought we could just log a warning and still give people an SBOM here that didn't include the unhandled binary

test/cli/test-fixtures/image-bad-binaries/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM debian:sid
+ADD sources.list /etc/apt/sources.list.d/sources.list
+RUN apt update -y && apt install -y dpkg-dev
+# this as a "macho-invalid" directory which is useful for testing
```
spiffcs, line 4: 🥳

syft/pkg/cataloger/golang/parse_go_bin_test.go
spiffcs: Edit: NVM I see that `parseGoBin` does not report its error up the chain.

spiffcs (approved):

syft/pkg/cataloger/golang/parse_go_bin_test.go
wagoodman: the caller is responsible for doing this, yes. This is only the parser function, but the cataloger makes the decision to continue with warning on parser error conditions.

wagoodman:
As a side note, though this panic has been solved, I'm still populating a directory with all of the package sources to scan and see if there are any more surprises

PR #667, opened 2021-12-13
ignore target link files based on path

### LINKED ISSUE #615: Syft indexing should ignore non-regular files
_(I'm extracting this from a conversation from https://anchorecommunity.slack.com/archives/C019BUXV7R6/p1635509628025900?thread_ts=1635430847.021800&cid=C019BUXV7R6)_

Syft is hanging when attempting to index paths that are non-regular files (such as FIFO):
```
  File: '/run/dmeventd-client'
  Size: 0         	Blocks: 0          IO Block: 4096   fifo
Device: 14h/20d	Inode: 12419       Links: 1
Access: (0600/prw-------)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2021-04-06 08:33:27.368000000 +0000
Modify: 2021-04-06 08:33:27.368000000 +0000
Change: 2021-04-06 08:33:27.368000000 +0000
 Birth: -
```

Indexing should skip theses files instead of hard coding ignore paths such as `/run` since these can occur anywhere in the filesystem.

### LINKED ISSUE #665: Syft hangs when encounting symlink (in /run/udev/) to a character device (in /dev/)
**What happened**:

Given a desire to scan a host
When it reaches a symlink that point to a character special device that would block on read
It hangs

Given a desire to scan a host
When it reaches a symlink that point to a character special device (OR a socket) [that could block on read]
It should skip it as unreadable and continue to completion.

On a machine with a sound driver (I couldn't think of other characters that would be likely to block on read for a CI/CD environment).

```
syft packages dir:/run/
```

Hangs when Indexing file:/run/udev/static_node-tags/uaccess/snd\x2fseq

/run/udev/static_node-tags/uaccess/snd\x2fseq is a symlink to /dev/snd/seq, which is a character device

strace shows syft doing the following

image

It would appear that syft is correctly detecting the symlink, but then not doing the check (that it would normally do if I run syft against file:/dev/snd/seq) in detecting it as  character special device.

**Environment**:
- Output of `syft version`: Version 0.32.0

Downloaded today from GitHub release using install.sh script from this repository.

image

Red Hat Enterprise Linux release 8.4 (Ootpa)   --- running as a graphical workstation with sound capability.

jonasagx:
Makes several changes:
- The directory resolver will now return paths relative to the root of the resolver, not CWD
- The directory resolver prunes file types from the index which are not regular files, directories, or symlinks
- Adds and modifies several tests around the directory resolver

Closes #615 ; Closes #665


rspilker (changes requested):
The error message "Could not find files by blog" should be "by Glob"

syft/source/directory_resolver.go
```diff
@@ -202,6 +209,7 @@ func (r directoryResolver) addPathToIndex(p string, info os.FileInfo) (string, e
 		r.refsByMIMEType[metadata.MIMEType] = append(r.refsByMIMEType[metadata.MIMEType], *ref)
 	}
 
+	log.Infof("indexed filesystem path=%q", p)
```
wagoodman, line 212: this log line is pretty verbose for the execution path it's in --was it meant to be used for troubleshooting then removed?

syft/source/directory_resolver.go
```diff
@@ -171,6 +171,13 @@ func (r directoryResolver) addPathToIndex(p string, info os.FileInfo) (string, e
 		if err != nil {
 			return "", fmt.Errorf("unable to readlink for path=%q: %w", p, err)
 		}
+
+		for _, filterFn := range r.pathFilterFns {
```
wagoodman, line 175: Filtering should occur within `indexPath`. If there was a need to put filtering logic here it may mean that this is a usage of `addPathToIndex` where there should be an `indexPath`

wagoodman: I misread this.

wagoodman (commented):
Looks like there is an edge case in irregular files.

Overall this is a step in the right direction but I think there are more things we can do when it comes to hinting on file types and not paths.

luhring (approved):
Nice work!! 👏 👏 👏

PR #669, opened 2021-12-13
Update README

davideicardi:
Update information about available output format.

According to #635 (cc @jonasagx) I have added `cyclonedx-json` as a possible output and for what I can see the version is 1.3, not 1.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ syft packages <image> -o <format>
 Where the `formats` available are:
 - `json`: Use this to get as much information out of Syft as possible!
 - `text`: A row-oriented, human-and-machine-friendly output.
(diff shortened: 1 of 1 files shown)


spiffcs (approved):

spiffcs:
@davideicardi I see you have signed the commit. Thanks for doing that:

Contributing to syft also requires using GPG or S/MIME, you can sign tags and commits locally. These tags or commits are marked as verified on GitHub so other people can be confident that the changes come from a trusted source.

See [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) on how to configure this and get your commits verified.

davideicardi:
@spiffcs Done

luhring:
Thanks @davideicardi!

PR #685, opened 2021-12-14
Add some additional property fileds to cycloneDX output.

pbalogh-sa:
Signed-off-by: Peter Balogh <p.balogh.sa@gmail.com>

Related https://github.com/anchore/grype/issues/481

This PR adds some additional properties to the CycloneDX output. After converting the generated CyclonDX format to Syft JSON format with these properties, the Grype using the converted json can find the vulnerabilities.

### DIFF
--- a/internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
+++ b/internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
@@ -16,7 +16,17 @@
       "type": "container",
       "name": "user-image-input",
       "version": "sha256:2731251dc34951c0e50fcc643b4c5f74922dad1a5d98f302b504cf46cd5d9368"
-    }
+    },
+    "properties": [
+      {
+        "name": "distroName",
+        "value": "debian"
+      },
+      {
+        "name": "distroVersion",
+        "value": "1.2.3"
+      }
+    ]
   },
   "components": [
     {
@@ -30,13 +40,37 @@
           }
         }
       ],
-      "purl": "a-purl-1"
+      "purl": "a-purl-1",
+      "properties": [
+        {
+          "name": "path",
+          "value": "/somefile-1.txt"
+        },
+        {
+          "name": "layerID",
+          "value": "sha256:d8fa3ce64b8228215c105c9a307ef852b27bf6075cec407ad54509f0a6121c1a"
+        }
+      ]
     },
     {
       "type": "library",
       "name": "package-2",
       "version": "2.0.1",
-      "purl": "a-purl-2"
+      "purl": "a-purl-2",
+      "properties": [
+        {
+          "name": "source",
+          "value": ""
+        },
+        {
+          "name": "path",
+          "value": "/somefile-2.txt"
+        },
+        {
+          "name": "layerID",
+          "value": "sha256:a2f67dceefd26123b36e015c60bb458d624a72f752801e9999d5b604764afecd"
+        }
+      ]
     }
   ]
 }
--- a/internal/formats/cyclonedx13xml/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
+++ b/internal/formats/cyclonedx13xml/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
@@ -13,6 +13,10 @@
       <name>user-image-input</name>
       <version>sha256:2731251dc34951c0e50fcc643b4c5f74922dad1a5d98f302b504cf46cd5d9368</version>
     </component>
+    <properties>
+      <property name="distroName">debian</property>
+      <property name="distroVersion">1.2.3</property>
+    </properties>
   </metadata>
   <components>
     <component type="library">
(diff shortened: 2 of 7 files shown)


wagoodman (commented):
@pbalogh-sa stellar PR --thanks for opening! The use of properties here helps a TON in the sense of unlocking the use of using CycloneDX formatted documents as input in grype (e.g. the source/origin properties you added).

I think the only thing that looks out of place is the `distro` information which has been attached to the `BomDescriptor` section. This section is meant to describe the tool and context that created the SBOM. The `distro` itself is from the result of cataloging and not truly meant for this BOM Descriptor section. 

I'm not certain that there is a good spot for this information in a top-level section in CycloneDX (if there is, shout out! I could have easily missed something here). But, it may be that the easiest path forward is to encode the distro information directly on packages which this information matters for. For instance, capturing distro name and version for `rpm`, `deb` and `apk` as properties on the package (but not capture this information for non-distro packages, such as python, java, etc).

Let me know what you think about this approach @pbalogh-sa  --happy to chat it through more.

pbalogh-sa:
@wagoodman  I wanted to convert cyclonedx format to syft json, and as I see in the syft json the distro information is in top-level, and not in package level. So This was the reason why I put it into the bomDescriptor section. The distro is only crucial in the case when an image was the input. Putting the distro information at the package level makes sense. For example, if we analyze a source directory and after it, we analyze the docker image we built using that source code we will be able to merge two BOM into one because all information will be in the package level.

wagoodman:
Using the BOM Descriptor is very convenient, but technically not the reason for the section, which is why I can't recommend heading further down that path (maybe keep it in our back pocket as a workaround incase other solutions don't work as well as we hope?).

I think the only two downsides (so far) to adding distro information to each package is:
1. the SBOM gets larger due to redundant information --this is more of an annoyance than a problem I suspect.
2. what if there are zero OS packages found? An odd case, but with the new `--exclude` option it's possible. In this case we wouldn't have any distro information in the SBOM. This is bad, but possibly not a show stopper for this path (still notable).

> ...we will be able to merge two BOM into one because all information will be in the package level

That's true! There is a benefit there I didn't think of. 

When we get to format conversion, I think this will complicate the case where we are trying to get information transformed back into the syft-native model, but I think we can solve that one later (so shouldn't be a restriction here).

@pbalogh-sa are you OK with the per-package approach?

internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
```diff
@@ -30,13 +40,37 @@
...
-      "purl": "a-purl-1"
+      "purl": "a-purl-1",
+      "properties": [
+        {
+          "name": "path",
+          "value": "/somefile-1.txt"
+        },
+        {
+          "name": "layerID",
+          "value": "sha256:d8fa3ce64b8228215c105c9a307ef852b27bf6075cec407ad54509f0a6121c1a"
```
spiffcs, line 51: Looks like this value should be being redacted per the encoder test changes.

Should value here be `redacted`?

internal/formats/cyclonedx13xml/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
```diff
@@ -24,11 +28,20 @@
         </license>
       </licenses>
       <purl>a-purl-1</purl>
+      <properties>
+        <property name="path">/somefile-1.txt</property>
+        <property name="layerID">sha256:d8fa3ce64b8228215c105c9a307ef852b27bf6075cec407ad54509f0a6121c1a</property>
```
spiffcs, line 33: Same comment re: `redacted` from the above discussion

spiffcs (commented):
@pbalogh-sa! Thanks for the excellent PR. @wagoodmans comments get to the core of the specification questions I might have here.

I added some small comments regarding the tests to just be sure the correct values are being redacted.

Thanks again!

internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
pbalogh-sa: Testing it locally I got this layerID but in the CI I got different one. I think the test image that was built is different in my case.

pbalogh-sa:
> are you OK with the per-package approach? or want to keep looking for alternate solutions? (there could be another option in the spec for us)

@wagoodman Currently distro is in the level of `syft.SBOM.Artifacts` struct. So it would be hard change now I think.

internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxImagePresenter.golden
pbalogh-sa: I think it's enough to test if the `layerID` is proper `sha256`, because we get it from the `location.FileSystemID` and just add this info to cycloneDX output.

wagoodman:
@pbalogh-sa We recently got in a feature that may help here: https://github.com/anchore/syft/pull/754 

Now we're encoding distro information onto each package within the pURL. In the near future the pURL will be used downstream in grype for vulnerability matching (enabling OS-level matching as we do today). 

Does this change fit your needs?

pbalogh-sa:
@wagoodman Thanks, along with https://github.com/anchore/syft/pull/710 can fit my needs. I will check it.

pbalogh-sa:
In the case of Java artifactID and groupID are missing.

wagoodman:
@pbalogh-sa With #758 merged, I think the remaining items on this PR are:

- remove the `distro` object from being encoded into the `BomDescriptor`
- rebase to pull in the latest changes

Let me know if there is anything I can do to help (or if I got the above list wrong)!

pbalogh-sa:
implemented by: #710 and #758

PR #687, opened 2021-12-14
Handle extra empty lines in Java manifest parsing

luhring:
For https://github.com/anchore/grype/issues/538


wagoodman (approved):
Nice! This addresses the cases that the issue covers, I'll follow up with a PR that covers a few more

PR #688, opened 2021-12-14
Prefer warning over erroring out when parsing java manifests

wagoodman:
When originally developing the java manifest parsing it was heavy in returning errors over returning partial results. This PR adjusts the behavior of the parser to be "best-effort" and return results when possible (even if partial).


luhring (approved):
Awesome

Issue #682, opened 2021-12-14
Add lpkg as java package format


ybroeker:
**What would you like to be added**:
Support for lpgk packages as Java archives, similar to jar and war.

**Why is this needed**:
Liferay bundles applications as lpkg (zip files containing jars/wars and some properties). syft currently ignores these, so many used libraries are not discovered

**Additional context**:

PR #694, opened 2021-12-15
Add lpkg as java package format

jonasagx:
also cleaned up linter warning

Closes: #682

### DIFF
--- a/syft/pkg/apk_metadata_test.go
+++ b/syft/pkg/apk_metadata_test.go
@@ -105,8 +105,7 @@ func TestApkMetadata_FileOwner(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(strings.Join(test.expected, ","), func(t *testing.T) {
-			var i interface{}
-			i = test.metadata
+			var i interface{} = test.metadata
 			actual := i.(FileOwner).OwnedFiles()
 			for _, d := range deep.Equal(test.expected, actual) {
 				t.Errorf("diff: %+v", d)
--- a/syft/pkg/apk_metadata_test.go
+++ b/syft/pkg/apk_metadata_test.go
@@ -105,8 +105,7 @@ func TestApkMetadata_FileOwner(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(strings.Join(test.expected, ","), func(t *testing.T) {
-			var i interface{} = test.metadata
-			actual := i.(FileOwner).OwnedFiles()
+			actual := test.metadata.OwnedFiles()
 			for _, d := range deep.Equal(test.expected, actual) {
 				t.Errorf("diff: %+v", d)
 			}
--- a/syft/pkg/catalog.go
+++ b/syft/pkg/catalog.go
@@ -4,12 +4,10 @@ import (
 	"sort"
 	"sync"
 
+	"github.com/anchore/syft/internal"
+	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/jinzhu/copier"
-
-	"github.com/anchore/syft/internal"
-
-	"github.com/anchore/syft/internal/log"
 )
 
 // Catalog represents a collection of Packages.
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -23,6 +23,10 @@ var archiveFormatGlobs = []string{
 	"**/*.ear",
 	"**/*.jpi",
 	"**/*.hpi",
+	"**/*.lpkg", // Zip-compressed package used to deploy applications
+	// (aka plugins) to Liferay users that can be installed on their
+	// websites, contains .JAR files and a .PROPERTIES file, the latter
+	//  has information about the application and installation requirements.
 }
 
 type archiveParser struct {
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -6,8 +6,8 @@ import (
 	"path"
 	"strings"
 
-	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/internal/file"
+	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/pkg"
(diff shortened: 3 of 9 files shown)


syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -23,6 +23,10 @@ var archiveFormatGlobs = []string{
 	"**/*.ear",
 	"**/*.jpi",
 	"**/*.hpi",
+	"**/*.lpkg", // Zip-compressed package used to deploy applications
+	// (aka plugins) to Liferay users that can be installed on their
+	// websites, contains .JAR files and a .PROPERTIES file, the latter
+	// has information about the application and installation requirements.
```
luhring, line 29: What do you think it would look like to add a test for this?

luhring (commented):
Thanks for doing this! One question for you to chat about re: testing

syft/pkg/cataloger/java/archive_parser.go
jonasagx: I added a file test. Their docs and tools are pretty outdated and I can't find sample projects, even on GitHub.

syft/pkg/apk_metadata_test.go
```diff
@@ -105,8 +105,7 @@ func TestApkMetadata_FileOwner(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(strings.Join(test.expected, ","), func(t *testing.T) {
-			var i interface{}
-			i = test.metadata
+			var i interface{} = test.metadata
 			actual := i.(FileOwner).OwnedFiles()
```
kzantow, line 109: curious if there is a reason to not just do:
```suggestion
			actual := test.metadata.(FileOwner).OwnedFiles()
```

syft/pkg/catalog.go
```diff
@@ -4,12 +4,10 @@ import (
...
 	"sync"
 
-	"github.com/anchore/syft/syft/artifact"
-	"github.com/jinzhu/copier"
-
 	"github.com/anchore/syft/internal"
-
 	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/jinzhu/copier"
```
kzantow, line 10: nit: prefer not to include import-only changes, I think?

kzantow (commented):
It looks like all of these changes are `gofmt` or `goimports` except for `archive_filename.go`, `archive_filename_test.go`, and `archive_parser.go`, unless they are causing failures with the PR, maybe just include the relevant changes to make a review a bit more concise? I don't see any `.lpkg` files included, though, which makes me think there is limited testing for this - is it sufficient?

jonasagx:
@kzantow while exploring the project my editor found places that were not properly formatted, I will try not to pollute PR too much going forward.

On testing: we don't add jar/war/hpi files within the repo, that is why there are automated building steps for most formats but war. War files are a super set of jar plus Servlet, JSP, HTML, JavaScript, but we only care about their jar content. Lpkg files are jars plus a property file. I wrote lpkg tests based off war validation that seems sufficient.

syft/pkg/apk_metadata_test.go
jonasagx: it can actually be simpler: `actual := test.metadata.OwnedFiles()`

syft/pkg/cataloger/java/archive_parser_test.go
```diff
@@ -283,9 +283,7 @@ func TestParseJar(t *testing.T) {
 				// ignore select fields (only works for the main section)
 				for _, field := range test.ignoreExtras {
 					if metadata.Manifest != nil && metadata.Manifest.Main != nil {
-						if _, ok := metadata.Manifest.Main[field]; ok {
-							delete(metadata.Manifest.Main, field)
-						}
+						delete(metadata.Manifest.Main, field)
```
spiffcs, line 286: I like that this simplifies the test! Just for my own sake, why do we no longer need this check for `ok`?

syft/pkg/dpkg_metadata_test.go
```diff
@@ -86,9 +86,7 @@ func TestDpkgMetadata_FileOwner(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(strings.Join(test.expected, ","), func(t *testing.T) {
-			var i interface{}
-			i = test.metadata
-			actual := i.(FileOwner).OwnedFiles()
+			actual := test.metadata.OwnedFiles()
```
spiffcs, line 89: Nice! 

What change went in that made so that metadata no longer hast to be cast to `FileOwner` I don't see it in this code diff so did we update something in another repo?

kzantow (approved):
This looks okay;  I would definitely like to have some sort of tests beyond the one included but I understand the difficulty adding these tests due to Liferay's paid development model. The only other issue I have is that there are really only 3 files with changes but a _lot_ of what appear to be unrelated things in the PR -- in particular `goimports` formatting and:
```
-			var i interface{}
-			i = test.metadata
-			actual := i.(FileOwner).OwnedFiles()
+			actual := test.metadata.OwnedFiles()
```
and
```
-						if _, ok := metadata.Manifest.Main[field]; ok {
-							delete(metadata.Manifest.Main, field)
-						}
+						delete(metadata.Manifest.Main, field)
```
I would probably just remove those changes and have the pertinent stuff in here, but they all look like improvements so I wouldn't hold up merging this on account of them!

syft/pkg/cataloger/java/archive_parser_test.go
jonasagx: Go is smart: 
> The delete built-in function deletes the element with the specified key (m[key]) from the map. If m is nil or there is no such element, delete is a no-op.

https://pkg.go.dev/builtin#delete

syft/pkg/dpkg_metadata_test.go
jonasagx: Maybe this https://github.com/jonasagx/syft/commit/93d00dc3409ded9a31aa8b52b0512d96bc252f6b

However, if `test.metadata` implements [FileOwner](https://github.com/jonasagx/syft/blob/main/syft/pkg/file_owner.go#L7) there is no need to cast it. [Deep equal](https://github.com/jonasagx/syft/blob/main/syft/pkg/dpkg_metadata_test.go#L90) accepts `interface{}` so any type can be compared.

PS: <3 hast

Issue #221, opened 2020-10-14
Exclude path option


wagoodman:
**What would you like to be added**:
Ability to exclude paths from being scanned (mirror of issue https://github.com/anchore/grype/issues/183)

**Why is this needed**:
You may already have a container running made from the image you want to catalog, however, there may be volume mounts or container additions that you'd like to explicitly ignore.

dpippenger:
It's interesting this issue has seen so little attention. Providing a --exclude flag for directories would open up syft for generating an SBOM for the container host itself. Doing this would create a whole new functionality for syft/grype and Anchore engine for the cost of one cli argument.

wagoodman:
I want to extend the original CLI ergonomic suggestions put forth by @yudong from https://github.com/anchore/syft/issues/221#issuecomment-936436399 :
```bash
# exclude single path
syft my/image:latest --exclude /some/path

# exclude multiple paths
syft my/image:latest -e /some/dir -e /some/file -e /something/else

# exclude by glob
syft my/image:latest -e /some/place/*.txt
```

And additionally being able to express this via the application configuration:
```yaml
exclude:
  - /some/path
  - /some/other/place
  - /a/**/glob
  - *.txt
```

This should work the same regardless if we are scanning an image or a directory.

This new behavior suggests the following changes:
- Adding the ability to bail early when indexing individual paths within layer tars (based on path/glob exclusion rules): https://github.com/anchore/stereoscope/blob/003132a67c108b437dafa50bfbc3fa4e89944cbe/pkg/image/layer.go#L145-L148
- Ability to prune entire branches from `filepath.Walk` based on path/glob exclusion rules within the syft directory resolver: https://github.com/anchore/syft/blob/main/syft/source/directory_resolver.go#L98-L107

A few notes that could influence the implementation:
- syft and stereoscope has normalized glob resolution (in string matching and otherwise) with [doublestar](https://github.com/bmatcuk/doublestar) and not with the globbing implementation from the standard lib
- there is already [path filtering on the directory resolver](https://github.com/anchore/syft/blob/main/syft/source/directory_resolver.go#L112-L116) but must visit all nodes discovered by walking, which may not be ideal. Migrating that mechanism to a new mechanism that affects the walking trajectory itself would be ideal.

cameronkerrnz:
[...]

PR #695, opened 2021-12-15
Add --exclude flag

kzantow:
This adds the `--exclude` flag to exclude paths from directory, archive, and image scans. This addresses #221

TODO:
- [x] add tests
- [x] add documentation
- [x] add more tests
- [x] add even more tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,15 +98,17 @@ registry:yourrepo/yourimage:tag        pull image directly from a registry (no c
 Syft can exclude files and paths from being scanned within a source by using glob expressions
 with one or more `--exclude` parameters:
 ```
-syft <source> --exclude ./out/**/*.json --exclude /etc
+syft <source> --exclude './out/**/*.json' --exclude /etc
 ```
 **Note:** in the case of _image scanning_, since the entire filesystem is scanned it is
-possible to use absolute paths, e.g. `/etc` or `/usr/**/*.txt` whereas _directory scans_
+possible to use absolute paths like `/etc` or `/usr/**/*.txt` whereas _directory scans_
 exclude files _relative to the specified directory_. For example: scanning `/usr/foo` with
-`--exclude ./package.json` would exclude `/usr/foo/package.json` and `--exclude **/package.json`
+`--exclude ./package.json` would exclude `/usr/foo/package.json` and `--exclude '**/package.json'`
 would exclude all `package.json` files under `/usr/foo`. For _directory scans_,
 it is required to begin path expressions with `./`, `*/`, or `**/`, all of which
-will be resolved _relative to the specified scan directory_.
+will be resolved _relative to the specified scan directory_. Keep in mind, your shell
+may attempt to expand wildcards, so put those parameters in single quotes, like:
+`'**/*.json'`.
 
 ### Output formats
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<p align="center">
+<p style="text-align: center">
     <img src="https://user-images.githubusercontent.com/5199289/136844524-1527b09f-c5cb-4aa9-be54-5aa92a6086c1.png" width="271" alt="Cute pink owl syft logo">
 </p>
 
@@ -91,6 +91,12 @@ oci-dir:path/to/yourimage              read directly from a path on disk for OCI
 dir:path/to/yourproject                read directly from a path on disk (any directory)
 file:path/to/yourproject/file          read directly from a path on disk (any single file)
 registry:yourrepo/yourimage:tag        pull image directly from a registry (no container runtime required)
+```
+
+Syft can exclude files and paths from being scanned within a source by using
+one or more `--exclude` parameters:
+```
+syft packages path/to/dir --exclude **/*.png --exclude **/generated/**
 ```
 
 ### Output formats
@@ -218,6 +224,12 @@ file: ""
 # enable/disable checking for application updates on startup
 # same as SYFT_CHECK_FOR_APP_UPDATE env var
 check-for-app-update: true
+
+# a list of globs to exclude. same as --exclude ; for example:
+# exclude:
+#   - '/etc/**'
+#   - '/tmp/**/*.png'
+exclude:
 
 # cataloging packages is exposed through the packages and power-user subcommands
 package:
--- a/README.md
+++ b/README.md
@@ -93,13 +93,20 @@ file:path/to/yourproject/file          read directly from a path on disk (any si
 registry:yourrepo/yourimage:tag        pull image directly from a registry (no container runtime required)
 ```
 
-#### Excluding file paths
+### Excluding file paths
 
 Syft can exclude files and paths from being scanned within a source by using glob expressions
 with one or more `--exclude` parameters:
 ```
-syft packages path/to/dir --exclude **/*.png --exclude **/generated/**
+syft <source> --exclude ./out/**/*.json --exclude /etc
 ```
+**Note:** in the case of _image scanning_, since the entire filesystem is scanned it is
+possible to use absolute paths, e.g. `/etc` or `/usr/**/*.txt` whereas _directory scans_
+exclude files _relative to the specified directory_. For example: scanning `/usr/foo` with
+`--exclude ./package.json` would exclude `/usr/foo/package.json` and `--exclude **/package.json`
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-<p align="center">
+<p style="text-align: center">
```
luhring, line 1: I may be misremembering, but I think I remember someone making this update a while back and finding that's not supported — can you confirm?

README.md
```diff
@@ -93,6 +93,12 @@ file:path/to/yourproject/file          read directly from a path on disk (any si
 registry:yourrepo/yourimage:tag        pull image directly from a registry (no container runtime required)
 ```
 
+Syft can exclude files and paths from being scanned within a source by using
+one or more `--exclude` parameters:
+```
+syft packages path/to/dir --exclude **/*.png --exclude **/generated/**
+```
+
```
luhring, line 101: It'd be nice to have a section header for this content. A lot of folks have been asking about this feature, so being able to give them a direct link to this section would be really valuable!

cmd/packages.go
```diff
@@ -132,6 +132,11 @@ func setPackageFlags(flags *pflag.FlagSet) {
 		"include dockerfile for upload to Anchore Enterprise",
 	)
 
+	flags.StringArrayP(
+		"exclude", "", nil,
+		"exclude paths using a glob expression",
```
luhring, line 137: nit: I don't want to make this too verbose, but I think "exclude" can be ambiguous with our tool, and it might be worthwhile to specify that this is excluding paths **from being scanned** (i.e. Syft will never touch them), as opposed to other kinds of exclusion Syft could've been doing, like filtering output.

syft/source/excluding_file_resolver.go
```diff
@@ -0,0 +1,71 @@
...
+	return r.delegate.FileContentsByLocation(location)
+}
+
+func (r *excludingResolver) FileMetadataByLocation(location Location) (FileMetadata, error) {
+	return r.delegate.FileMetadataByLocation(location)
+}
+
+func (r *excludingResolver) HasPath(path string) bool {
+	return r.delegate.HasPath(path)
+}
```
luhring, line 32: Just curious, should filter logic come into play with these methods, too? Like if the caller is getting file content, or asking if the path exists, and the path should be filtered, should we just say "no"?

syft/source/excluding_file_resolver_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			resolver := &mockResolver{
+				locations: test.locations,
+			}
+			excludingResolver := NewExcludingResolver(resolver, test.excludeFn)
+
+			fc, _ := excludingResolver.FilesByPath()
+			assert.Equal(t, test.expectedLength, len(fc))
```
luhring, line 62: (totally optional: there's an `assert.Len` function that makes this slightly cleaner)

luhring (commented):
This is so exciting! 🎉  I left a first pass of code comments. Nothing major. I'm taking this for a test drive now, and may want to sync up with you about it

luhring:
I have a thought on ergonomics, and I'm not sure where to take this, but wanted to express and discuss it. I'm attempting to take my developer hat off, and put my user hat on. 😬 

I find myself reaching for the same kind of simple expressions I'd use in a `.gitignore`, like this:

```
$ dir:. --exclude 'unwanted-folder'
```

But that doesn't work... and if I put my developer hat back on, I get that `.gitignore` doesn't match the behavior of the glob library we're using, where `**` is needed. And a separate possible issue is that I didn't preface my directory name with `./`.

But given that, it'd be really really (really) nice to have the ability to express a path relative to the dir I'm scanning, like this:

```
$ dir:. --exclude './unwanted-folder/**'
```

In my testing, this doesn't work. I glean from the examples that it's  **not intended to work**. So no flaw there. But I would like to open the discussion for making this work. Thoughts? 😄

README.md
kzantow: you can just edit in the browser to see. the `align` tag is not actually supposed to be part of html5. latest chrome looks like: 
image

kzantow: (I only changed it because intellij complained "hey this is an invalid attribute" so i looked it up

kzantow:
Yeah @luhring the note in the description explains why. I also do not like this at all -- I would much prefer all the paths be "relatively absolute" to the root of the directory being scanned. I wasn't quite sure how to accomplish this without doing string manipulation on every path test, but I have an idea I will try out.

ASIDE: working on this made me think that we should actually be using "relatively absolute" paths in the directory resolver, and if you need file contents or other info, it knows the prefix to use etc..

README.md
kzantow: this is not really part of the PR, so I'll just revert it anyway

syft/source/excluding_file_resolver.go
kzantow: it isn't _exactly_ necessary based on how this is used, but yes, it probably should also apply here

README.md
luhring: @kzantow Oh yeah — I totally think `style=...` is better, I just thought there was a specific incompatibility w/ GitHub or something. I could be totally crazy :D

luhring (approved):
LGTM!

syft/source/directory_resolver.go
```diff
@@ -95,7 +92,7 @@ func (r *directoryResolver) indexTree(root string, stager *progress.Stage) ([]st
 	fi, err := os.Stat(root)
 	if err != nil && fi != nil && !fi.IsDir() {
 		// note: we want to index the path regardless of an error stat-ing the path
-		newRoot := r.indexPath(root, fi, nil)
+		newRoot, _ := r.indexPath(root, fi, nil)
```
spiffcs, line 95: Do we still wan to totally drop the error here for the r.indexPath? Since we're surfacing the error maybe want something where it could be used in an `if err != nil { warn/debug }` rather than `fatal`

syft/source/source.go
```diff
@@ -26,28 +29,38 @@ type Source struct {
...
 		return &Source{}, func() {}, fmt.Errorf("unable to parse input=%q: %w", userInput, err)
 	}
 
+	source := &Source{}
+	fn := func() {}
+
 	switch parsedScheme {
 	case FileScheme:
-		return generateFileSource(fs, location)
+		source, fn, err = generateFileSource(fs, location)
```
spiffcs, line 50: Would it be ok to call these cleanup instead of fn? I think fn might be too generic in this case. During the review I was wondering what new functionality was added but then had to trace that it was the old cleanup function

syft/source/source.go
```diff
@@ -26,28 +29,38 @@ type Source struct {
...
+		source, fn, err = generateDirectorySource(fs, location)
 	case ImageScheme:
-		return generateImageSource(location, userInput, imageSource, registryOptions)
+		source, fn, err = generateImageSource(location, userInput, imageSource, registryOptions)
+	default:
+		err = fmt.Errorf("unable to process input for scanning: '%s'", userInput)
+	}
+
+	if err == nil {
+		source.Exclusions = exclusions
```
spiffcs, line 60: Does it make sense to make exclusions compacted into where we generate the source that way we don't have to modify it after it has been returned from the specific constructors?

This would also allow us to keep the same branching logic from before.

Not a hard ask or anything just a thought since I think this might have to be refactored when taking into account windows paths. Each generate function will have to do some specific mutation to the exclusions to make it match with the source being generated in some instances.

syft/source/directory_resolver.go
kzantow: The only "error" being returned is `fs.SkipDir`, which is a special thing to return during the walk that prevents traversal of subdirectories. This was ignoring anyone  errors already, effectively.

syft/source/source.go
kzantow: The problem is I need this data when `FileResolver(...)` is called, so it doesn't really do a lot of good to pass it down and the only caller is in the `cmd` package, which doesn't seem like the right place to set this (and makes it harder to test)

spiffcs (approved):

spiffcs:
This can be merged - @kzantow and then I'll rebase my windows changes on top of it and see what I can do about making --exclude compatible on windows

PR #698, opened 2021-12-15
Performance improvements around package ID

### LINKED ISSUE #696: Unable to catalog image gitlab/gitlab-ce:latest
**What happened**:
When cataloging docker images using a script `syft` got stuck at one image and continued to consume quite a lot of CPU.
The image that caused trouble is `gitlab/gitlab-ce:latest`

I would expect that `syft` would catalog this image as good as any other image.

Download the image and catalog it:
```
docker pull gitlab/gitlab-ce:latest 
syft -vv -o json --file gitlab_gitlab-ce:latest.sbom.json gitlab/gitlab-ce:latest
...
[0000] DEBUG no new syft update available
[0000] DEBUG image: source=DockerDaemon location=gitlab/gitlab-ce:latest from-lib=stereoscope
[0083] DEBUG image metadata: digest=sha256:a2bf5ef04c22b5530d9c57aa5f20b55601c85fa2393d2d81e120d235f2a39ce4 mediaType=application/vnd.docker.distribution.manifest.v2+json tags=[gitlab/gitlab-ce:latest] from-lib=stereoscope
...
[0150] DEBUG cataloging with "apkdb-cataloger"
[0153] DEBUG discovered 0 packages
[0153] DEBUG cataloging with "go-module-binary-cataloger"
[0164] DEBUG discovered 1490 packages

<stuck here>
```
Scanning the same image with `grype` goes well.

**Environment**:
- Output of `syft version`:
Application:   syft
Version:       0.32.2
BuildDate:     2021-12-14T17:38:52Z
GitCommit:     727b84ce0da67a35899e966eb4408b3e6553ed63
GitTreeState:  clean
Platform:      linux/amd64
GoVersion:     go1.16.11
Compiler:      gc

NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

wagoodman:
`Package.ID()` uses `hashistructure.Hash()` to use the contents of a package as the definition of the ID of the package. This is problematic as ID() has been more widely used in the codebase performance-wise.

This PR makes the following changes to the baseline:
- Adds an unexported `id` field onto `pkg.Package` which is read by `Package.ID()`. The `id` field is set only when a new `Package.SetID()` is called. It is up to the creator of the package to determine which this should be called.
- The common parser functions now return `[]*Package` instead of `[]Package`. This is due to the fact that parser functions can not be the finalizer of a package (the `common.GenericCataloger` is) however, the same object reference is required for any `artifact.Relationship` objects created, which their `From` and `To` fields (that implement `artifact.Identifiable`). This is consistent with the logic that both parser functions and catalogers may change attributes on a `Package` and the `Package` is only finalized when returned from the `Cataloger`.
- `source.Location` no longer implements `hashstructure.Hash`. `Location.ID()` is conveyed through the embedded `source.Coordinate` which implements `artifact.Identifiable`. 
- `pURL` and `CPE`s are no longer considered when determining a package ID; all of these values are synthesized from other package information. Note: this could change in the future when considering populating these values from other sources, but for now this seems correct and is convenient for the `cataloger.Catalog` function that sets these values for all objects. 

Time to run `syft gitlab/gitlab-ce:latest -o json` for versions of syft:
- v0.30.1 (before performance regression): ~49 seconds
- v0.32.2 (after performance regression): ~ 5 hours
- this PR (after remediation): ~60 seconds

Closes #696

### DIFF
--- a/internal/formats/syftjson/encoder.go
+++ b/internal/formats/syftjson/encoder.go
@@ -8,7 +8,7 @@ import (
 )
 
 func encoder(output io.Writer, s sbom.SBOM) error {
-	doc := ToFormatModel(s)
+	doc := toFormatModel(s)
 
 	enc := json.NewEncoder(output)
 	// prevent > and < from being escaped in the payload
--- a/syft/artifact/id.go
+++ b/syft/artifact/id.go
@@ -13,7 +13,7 @@ type Identifiable interface {
 	ID() ID
 }
 
-func IDFromHash(obj interface{}) (ID, error) {
+func IDByHash(obj interface{}) (ID, error) {
 	f, err := hashstructure.Hash(obj, hashstructure.FormatV2, &hashstructure.HashOptions{
 		ZeroNil:      true,
 		SlicesAsSets: true,
--- a/syft/pkg/catalog.go
+++ b/syft/pkg/catalog.go
@@ -77,10 +77,10 @@ func (c *Catalog) Add(p Package) {
 	defer c.lock.Unlock()
 
 	id := p.ID()
-
 	if id == "" {
 		log.Warnf("found package with empty ID while adding to the catalog: %+v", p)
 		p.SetID()
+		id = p.ID()
 	}
 
 	// store by package ID
@@ -147,7 +147,7 @@ func (c *Catalog) Sorted(types ...Type) (pkgs []Package) {
 	sort.SliceStable(pkgs, func(i, j int) bool {
 		if pkgs[i].Name == pkgs[j].Name {
 			if pkgs[i].Version == pkgs[j].Version {
-				if pkgs[i].Type == pkgs[j].Type {
+				if pkgs[i].Type == pkgs[j].Type && len(pkgs[i].Locations) > 0 && len(pkgs[j].Locations) > 0 {
 					return pkgs[i].Locations[0].String() < pkgs[j].Locations[0].String()
 				}
 				return pkgs[i].Type < pkgs[j].Type
--- a/syft/pkg/catalog.go
+++ b/syft/pkg/catalog.go
@@ -51,6 +51,7 @@ func (c *Catalog) Package(id artifact.ID) *Package {
 		log.Warnf("unable to copy package id=%q name=%q: %+v", id, v.Name, err)
 		return nil
 	}
+	p.id = v.id
 	return &p
 }
 
@@ -75,8 +76,12 @@ func (c *Catalog) Add(p Package) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	// note: since we are capturing the ID, we cannot modify the package being added from this point forward
 	id := p.ID()
+
+	if id == "" {
+		log.Warnf("found package with empty ID while adding to the catalog: %+v", p)
+		p.SetID()
+	}
 
 	// store by package ID
 	c.byID[id] = p
--- a/syft/pkg/cataloger/deb/cataloger.go
+++ b/syft/pkg/cataloger/deb/cataloger.go
@@ -68,6 +68,8 @@ func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, []arti
 
 			// fetch additional data from the copyright file to derive the license information
 			addLicenses(resolver, dbLocation, p)
+
+			p.SetID()
 		}
 
 		allPackages = append(allPackages, pkgs...)
--- a/syft/pkg/cataloger/javascript/parse_package_json.go
+++ b/syft/pkg/cataloger/javascript/parse_package_json.go
@@ -50,7 +50,6 @@ type Repository struct {
 // ---> name: "Isaac Z. Schlueter" email: "i@izs.me" url: "http://blog.izs.me"
 var authorPattern = regexp.MustCompile(`^\s*(?P<name>[^<(]*)(\s+<(?P<email>.*)>)?(\s\((?P<url>.*)\))?\s*$`)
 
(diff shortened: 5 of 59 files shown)


internal/formats/syftjson/encoder.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func encoder(output io.Writer, s sbom.SBOM) error {
-	doc := ToFormatModel(s)
+	doc := toFormatModel(s)
```
spiffcs, line 11: Nice privatization here

syft/artifact/id.go
```diff
@@ -13,7 +13,7 @@ type Identifiable interface {
 	ID() ID
 }
 
-func IDFromHash(obj interface{}) (ID, error) {
+func IDByHash(obj interface{}) (ID, error) {
```
spiffcs, line 16: Nice I also thought this was a little awkward when we were debugging this AM.

syft/pkg/catalog.go
```diff
@@ -51,6 +51,7 @@ func (c *Catalog) Package(id artifact.ID) *Package {
 		log.Warnf("unable to copy package id=%q name=%q: %+v", id, v.Name, err)
 		return nil
 	}
+	p.id = v.id
```
spiffcs, line 54: I like propagating this up from the catalog

syft/pkg/cataloger/deb/cataloger.go
```diff
@@ -68,6 +68,8 @@ func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, []arti
 
 			// fetch additional data from the copyright file to derive the license information
 			addLicenses(resolver, dbLocation, p)
+
+			p.SetID()
```
spiffcs, line 72: Maybe a comment for context that THIS is where the package become immutable?

I think it's going to be hard to enforce this by convention going forward so I'm trying to think of a better way than a giant warning comment.

syft/pkg/cataloger/javascript/parse_package_json.go
```diff
@@ -135,7 +134,7 @@ func licenseFromJSON(b []byte) (string, error) {
 	return "", errors.New("unable to unmarshal license field as either string or object")
 }
 
-func licensesFromJSON(p PackageJSON) ([]string, error) {
```
spiffcs, line 138: Why the change from function to method here? Just preference?

syft/pkg/cataloger/javascript/parse_package_lock_test.go
```diff
@@ -18,26 +20,11 @@ func assertPkgsEqual(t *testing.T, actual []pkg.Package, expected map[string]pkg
...
-
-		if a.Type != expectedPkg.Type {
-			t.Errorf("%s : bad package type: %+v, expected: %+v", a.Name, a.Type, expectedPkg.Type)
-		}
-
-		if len(a.Licenses) < len(expectedPkg.Licenses) {
-			t.Errorf("%s : bad package licenses count: '%+v'", a.Name, a.Licenses)
-		}
-
+		assert.True(t, ok)
```
spiffcs, line 23: Whoa nice simplification

syft/pkg/package.go
```diff
@@ -14,31 +14,35 @@ import (
 // Package represents an application or library that has been bundled into a distributable format.
 // TODO: if we ignore FoundBy for ID generation should we merge the field to show it was found in two places?
 type Package struct {
+	id           artifact.ID       `hash:"ignore"`
```
spiffcs, line 17: huzza

spiffcs (approved):
Approved with minor comments - Really awesome update

syft/pkg/cataloger/deb/cataloger.go
luhring: Agreed. I think the need for this `SetID` method is the only thing in this PR that seems like it needs continued thought and follow-up — in a **_future_** PR. I can see this getting difficult to enforce and reason about for future developers, and it'd be great to find ways to reduce the burden on catalogers and achieve more immutability. But I think the overall design here is definitely an improvement. 💯

luhring (approved):
Great performance improvement! 👏

syft/artifact/id.go
wagoodman: yeah, I realized I almost got the name right, so was a good time to tweak it

syft/pkg/cataloger/javascript/parse_package_json.go
wagoodman: initially I was going the route of custom per-cataloger package constructors, but realized it wasn't going to work given some extra information I found. I left the per-package creation updates as is though (this was one of them).

syft/pkg/cataloger/javascript/parse_package_lock_test.go
wagoodman: thanks --when fixing some tests I got to do some upgrades :)

wagoodman:
>  I think the need for this SetID method is the only thing in this PR that seems like it needs continued thought and follow-up

Agreed. I should have mentioned this in the PR description but forgot to:My opinion on `SetID()` is **_I very much dislike it_** --however it does solve the problem in a flexible fashion so hit my "good enough" threshold. All other alternatives had a higher burden on developers regarding brittleness / cognitive load.

PR #699, opened 2021-12-16
683 Add windows translation for posix paths

spiffcs:
## Add windows support for `dir:<directory>` scheme
image

### Summary
Syft consumes `FileTree` [^1] from stereoscope to build an index of the scanned input. It uses this `FileTree` along with `os.Adapter`[^2] to integrate with the [doublestar](https://github.com/bmatcuk/doublestar) library for path pattern matching and globing support.

It it's current form the `FileTree`  [^1] is indexed using the true `\\` windows path separators as keys to discover the virtual file nodes. The screen shot below shows some debugger output where `nodes` is a map and `node.ID` is the Windows specific path.

image.

According to [doublestar](https://github.com/bmatcuk/doublestar) documentation:

> This function assumes that your pattern uses / as the path separator even if that's not correct for your OS (like Windows). If you aren't sure if that's the case, you can use filepath.ToSlash() on your pattern before calling Glob().

This assumption is made because doublestar v4 is compatible with the new `fs.FS`.

> Note that paths are slash-separated on all systems, even Windows. Paths containing other characters such as backslash and colon are accepted as valid, but those characters must never be interpreted by an FS implementation as path element separators. [^3]

Given that the new `Glob` method expects an `fs.FS` interface where paths are slash-separated this PR's methods had to be introduced at the directory resolver layer in order to translate windows paths to the posix path the doublestar glob matcher can interact with. 
image

[^1]: https://github.com/anchore/stereoscope/blob/5a5e323a5c89082fce2624f2c72d168bfac04614/pkg/filetree/filetree.go#L21-L36
[^2]: https://github.com/anchore/stereoscope/blob/5a5e323a5c89082fce2624f2c72d168bfac04614/pkg/filetree/glob.go#L127-L131
[^3]: https://pkg.go.dev/io/fs#ValidPath

### TODO
- [ ] Unit tests POSIX -> Windows
- [ ] Unit tests Windows -> POSIX
- [x] Edge case identification
- [ ] Explore other FS integrations and reduce code complexity


spiffcs:
Doing this between two different computers and rebasing a couple of times I think the history is in a poor state for review. I might just cherry-pick the stuff I need into another branch and delete this one to clear it all up

spiffcs:
Closing this in favor of a new PR that will be opened shortly

PR #704, opened 2021-12-17
Adjust CPE generation for log4j

kzantow:
Some false positives for the log4j vulnerability are showing up as noted in https://github.com/anchore/grype/issues/552 This will add filtering for them.

NOTE: there is some request to leave the current behavior as-is: https://github.com/anchore/grype/issues/552#issuecomment-996502034

### DIFF
--- a/syft/pkg/cataloger/common/cpe/filter_test.go
+++ b/syft/pkg/cataloger/common/cpe/filter_test.go
@@ -165,3 +165,101 @@ func Test_disallowJiraClientServerMismatch(t *testing.T) {
 		})
 	}
 }
+
+func Test_disallowLog4JCVE202144228FalsePositives(t *testing.T) {
+	tests := []struct {
+		name     string
+		cpe      pkg.CPE
+		pkg      pkg.Package
+		expected bool
+	}{
+		// filter out
+		{
+			name: "filter out log4j-api java packages",
+			cpe:  mustCPE("cpe:2.3:a:apache:log4j:3.2:*:*:*:*:*:*:*"),
+			pkg: pkg.Package{
+				Name: "log4j-api",
+				Type: pkg.JavaPkg,
+			},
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/common/cpe/filter_test.go
```diff
@@ -165,3 +165,110 @@ func Test_disallowJiraClientServerMismatch(t *testing.T) {
...
+				Name: "log4j-api",
+				Type: pkg.JavaPkg,
+			},
+			expected: true,
+		},
+		{
+			name: "filter out log4j-slf4j-impl java packages",
+			cpe:  mustCPE("cpe:2.3:a:apache:log4j:3.2:*:*:*:*:*:*:*"),
+			pkg: pkg.Package{
+				Name: "log4j-slf4j-impl",
```
wagoodman, line 190: is the reason why `log4j-slf4j-impl` was added was because the `log4j-slf4j` package as a whole is an exception?

kzantow: yes. I'm pretty sure nothing related to `log4j-slf4j` is vulnerable. `slf4j` == Simple Logging Facade for Java. It's effectively an API that you program logging agnostically and plug in various back-ends. The `log4j-slf4j-impl` is just an adapter that still requires a vulnerable `log4j-core` library to be present

kzantow: UPDATE: yes, if you look at the `log4j-slf4j-impl` [dependency tree](https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/dependencies.html#Dependency_Tree), you will see it depends on: `org.apache.logging.log4j:log4j-core:jar:2.17.0 (runtime)` (and earlier vulnerable versions, where applicable)

PR #705, opened 2021-12-19
Fix unhelpful error message for oci-archive scheme

### LINKED ISSUE #701: Misleading error message when oci-archive scheme used on non-OCI archives
**What happened**:
I get "no such file or directory" for an existing oci:archive

The oci archive is picked.

```
$ ls -la image.tar 
-rw-r--r-- 1 docker docker 1400545280 Dec 17 12:30 /tmp/test/image.tar
$ docker run -v $(pwd):/x anchore/syft packages oci-archive:/x/image.tar
1 error occurred:
	* failed to determine image source: could not fetch image '/x/image.tar': unable to open OCI tarball: open oci-archive:/x/image.tar: no such file or directory
```
Same with local installation and absolute or relative paths.

**Environment**:
- Output of `syft version`:
0.33.0

Linux, Docker

luhring:
This fixes #701, ~although this is likely to revert the fix applied for #287 in the PR #626~ (**EDIT**: This is no longer true, we can preserve the retry functionality used in #626 with some minor tweaks 🎉 ).

Due to the complexity of the call graph for determining and processing an image source, it's probably best if user input is processed only one way if a scheme is specified, to avoid even more potential execution paths. What was happening that led to the misleading error message in #701 was that a second attempt of interpreting user input caused the error message from the first attempt to be swallowed (an error message which would've helped the user figure out the issue more quickly).

~This PR is intended to be dependent on https://github.com/anchore/stereoscope/pull/92, so I'll start by opening this in **draft**, with a `replace` directive left in `go.mod`.~ **EDIT**: With the stereoscope PR merged, this PR has been updated to use the latest commit of stereoscope.

### DIFF
--- a/cmd/packages.go
+++ b/cmd/packages.go
@@ -264,7 +264,7 @@ func packagesExecWorker(userInput string) <-chan error {
 
 		src, cleanup, err := source.New(userInput, appConfig.Registry.ToOptions(), appConfig.Exclusions...)
 		if err != nil {
-			errs <- fmt.Errorf("analysis failed: %w", err)
+			errs <- fmt.Errorf("failed to construct source from user input %q: %w", userInput, err)
 			return
 		}
 		if cleanup != nil {
--- a/cmd/packages.go
+++ b/cmd/packages.go
@@ -264,7 +264,7 @@ func packagesExecWorker(userInput string) <-chan error {
 
 		src, cleanup, err := source.New(userInput, appConfig.Registry.ToOptions(), appConfig.Exclusions...)
 		if err != nil {
-			errs <- fmt.Errorf("failed to determine image source: %w", err)
+			errs <- fmt.Errorf("analysis failed: %w", err)
 			return
 		}
 		if cleanup != nil {
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -51,7 +51,7 @@ func New(userInput string, registryOptions *image.RegistryOptions, exclusions ..
 	case DirectoryScheme:
 		source, cleanupFn, err = generateDirectorySource(fs, location)
 	case ImageScheme:
-		source, cleanupFn, err = generateImageSource(location, userInput, imageSource, registryOptions)
+		source, cleanupFn, err = generateImageSource(location, imageSource, registryOptions)
 	default:
 		err = fmt.Errorf("unable to process input for scanning: '%s'", userInput)
 	}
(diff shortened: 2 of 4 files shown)


cmd/packages.go
```diff
@@ -255,7 +255,7 @@ func packagesExecWorker(userInput string) <-chan error {
 
 		src, cleanup, err := source.New(userInput, appConfig.Registry.ToOptions())
 		if err != nil {
-			errs <- fmt.Errorf("failed to determine image source: %w", err)
+			errs <- fmt.Errorf("analysis failed: %w", err)
```
spiffcs, line 258: I think the error message still reads ok ok when it says `failed to determine image source` - It helps determine if the error came on sourcing the image VS processing it's packages `analysis failed` is a little vague and could belong to any part of the program

syft/source/source.go
```diff
@@ -44,19 +44,17 @@ func New(userInput string, registryOptions *image.RegistryOptions) (*Source, fun
...
 
-func generateImageSource(location, userInput string, imageSource image.Source, registryOptions *image.RegistryOptions) (*Source, func(), error) {
+func generateImageSource(location string, imageSource image.Source, registryOptions *image.RegistryOptions) (*Source,
+	func(), error) {
 	img, err := stereoscope.GetImageFromSource(location, imageSource, registryOptions)
 	if err != nil {
-		log.Debugf("error parsing location: %s after detecting scheme; pulling image: %s", location, userInput)
-		// we may have been to aggressive reading the source hint
-		// try the input as supplied by the user if our initial parse failed
-		img, err = stereoscope.GetImageFromSource(userInput, imageSource, registryOptions)
```
spiffcs, line 59: I don't think I'm in favor of removing this call to fix the error message since it is catching a lot of cases that might fail initially for the user. 

Could we elevate the `debug` to be an `info` message to give the user context in this scenario that we could not parse the provided location and are trying one more variation?

spiffcs:
Also I don't remember doing #287 with the mentality that it was a half-baked fix that needed more thought. If we felt that way initially then we probably should not have merged it or mentioned it on the PR.

cmd/packages.go
luhring: I think you're right, we can make this message a bit more specific. "Failed to determine image source" is wrong, though, since this code is no longer used just to determine an _image_ source.

syft/source/source.go
luhring: I think we should sync on this. I have a few concerns about not respecting user-specified schemes as explicit intent from the user.

luhring:
> Also I don't remember doing #287 with the mentality that it was a half-baked fix that needed more thought. If we felt that way initially then we probably should not have merged it or mentioned the hesitancy about it on the PR.

@spiffcs To be clear, I don't think it was a half-baked fix! I think it was a solution that cured the immediate pain in #287 — but I do remember noting the need for follow-up regarding schemes. And you left a comment here about this: https://github.com/anchore/syft/issues/287#issuecomment-967713766.

Seeing this bug (#701) arise here has helped me to iterate on my thinking about how strictly we handle user input when they use schemes.

cmd/packages.go
spiffcs: `Failed to parse image source: %s, %w` Where `%v` is the `userInput`

spiffcs (approved):
👍 LGTM

luhring:
I'll make one more adjustment that we discussed, such that images named `docker:...` and `registry:...` get the same pull retry behavior that other images benefit from — where we first check for the Docker daemon, and fallback to a direct registry pull. This determination has been centralized an exposed in a newly exported function via https://github.com/anchore/stereoscope/pull/94. Once that's merged, I'll update this PR to use it.

syft/source/source.go
```diff
@@ -114,6 +114,7 @@ func getImageWithRetryStrategy(userInput, location string, imageSource image.Sou
 		err,
 	)
 
+	imageSource = image.DetermineImagePullSource(userInput)
```
wagoodman, line 117: nit: a healthy comment as to why the image source is redetermined would go a long way

wagoodman (approved):
added a small nit on comments --looks great. Thanks for taking the extra time to think about behavioral consistency for this feature 💯

spiffcs:
LGTM

PR #681, opened 2021-12-14
679 pip vcs url

spiffcs:
## Add initial support for pip VCS URL specification

Issue linked, see [here](https://packaging.python.org/en/latest/specifications/direct-url/#example-direct-url-json) for more information.

Screenshot from a repository with direct VCS installation of httpie run in virtual env.

TODO:
- Discuss struct flattening and camel vs underscore for decoding/encoding
- Fix/Add unit tests

Screen Shot 2021-12-14 at 1 16 16 AM

PR #708, opened 2021-12-20
Add tests for direct-url information and add it to the output purl

samj1912:

Builds on top of #681 to add tests and PURL support


syft/pkg/cataloger/package_url_test.go
```diff
@@ -24,9 +24,30 @@ func TestPackageURL(t *testing.T) {
 		},
 		{
 			pkg: pkg.Package{
-				Name:    "name",
-				Version: "v0.1.0",
+				Name:    "bad-name",
+				Version: "bad-v0.1.0",
 				Type:    pkg.PythonPkg,
+				Metadata: pkg.PythonPackageMetadata{
```
spiffcs, line 30: Thanks for adding these cases! Since we're adding new cases to this `TestPackageURL` test it might be good to take the opportunity to add a `name` field to the test struct `lines 13-15`. This field should just be used as a label to state which case the test struct is covering so when we come back to look at this file we have context for the different tests under `TestPackageURL`

syft/pkg/python_package_metadata.go
```diff
@@ -21,17 +23,45 @@ type PythonFileRecord struct {
...
+	AuthorEmail          string                     `json:"authorEmail" mapstruct:"Authoremail"`
+	Platform             string                     `json:"platform" mapstruct:"Platform"`
+	Files                []PythonFileRecord         `json:"files,omitempty"`
+	SitePackagesRootPath string                     `json:"sitePackagesRootPath"`
+	TopLevelPackages     []string                   `json:"topLevelPackages,omitempty"`
+	DirectURLOrigin      *PythonDirectURLOriginInfo `json:"directUrlOrigin,omitempty"`
+}
+
+type DirectURLOrigin struct {
+	URL         string      `json:"url"`
```
spiffcs, line 47: It's interesting that these are tagged as underscore.

I know the specification you linked has them tagged this way so it would not unmarshal correctly if we did it with camel case. 

I think I defer to @wagoodman on this for how we want to keep the Metadata shape consistent

I think we have two options here:
1 - create unmarshal structs that then `1-1` translate over to our `Metadata` struct
2 - custom unmarshal to try both cases?

I know the structs here are correct after reviewing the spec, it's just a matter of getting the tagging and presentation correct now.

spiffcs (commented):
Small comments on presentation and test context but thanks so much for the follow-up here. I've been stuck on another issue since I filed the first PR so this helps a ton.

syft/pkg/python_package_metadata.go
samj1912: IIUC these structs are not directly used anywhere except for unmarshalling the input `direct_url.json` files and we flatten them out in metadata when we use `PythonDirectURLOriginInfo` struct which has fields as camelCase.

syft/pkg/cataloger/package_url_test.go
samj1912: Added - also backfilled it for some other tests. :)

syft/pkg/python_package_metadata.go
spiffcs: That is absolutely my bad - I thought DirectURLOrigin was being put onto metadata not `DirectURLOriginInfo`

spiffcs (approved):

PR #710, opened 2021-12-21
Improve CycloneDX format output

samj1912:
This commit dramatically improves the information we add in the
cyclonedx output -

It adds -

* CPEs
* Authors
* Publishers
* External References (Website, Distribution, VCS)
* Description


samj1912:
cc: @VinodAnandan, @coderpatros from CycloneDX team

VinodAnandan:
Thank you @samj1912 !!
Cc + @nscuro, @stevespringett

samj1912:
I am facing an issue with CPE binding strings that are generated by https://github.com/facebookincubator/nvdtools

It looks like it has 11 parts - but the cyclonedx schema https://cyclonedx.org/docs/1.3/xml/#type_cpe and the upstream CPE schema (https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd) both require it to have 10 parts

This is causing the nvdtools library to add an additional `:*` which is not matched by the cyclonedx and cpe schema and causing integration test errors - 


The ways I see around this - 

1. Serialize the CPE as an URI instead of a formatted string (the facebook nvdtools library handles this correctly and this is correctly recognized by the cyclonedx schema)
2. Fix the nvdtools binding function in syft (It seems to add an `Other` part to the string, which is probably what is causing the issue) See https://regexr.com/6c0et
3. Use a different cpe parsing library that follows the spec closely

samj1912:
Created issue #712 - will fix it in a different PR

samj1912:
Note on the commit that adds metadata props - 


```
This commit adds arbitrary properties from the syft metadata to
CycloneDX. This builds on top of the previous work on enriching
the cyclonedx output. Creating a separate PR for this work as this
might be a point of contention whether we want to add these properties
or not or the way the properties are currently extracted using struct
tags.

NOTE: Diff to look at samj1912/syft-1@enrich-cdx...enrich-cdx-props

This allows us to use struct tags to add cyclonedx property values. It works in the follow way -

If it detects an explicit cyclonedx struct tag and it will use the value of that tag as the name for the property.

It stringifies the value of the struct field to get the prop value since cyclonedx props are strings only.

If the cyclonedx struct tag is not added or if the stringified value is empty, the prop is not added.
```

internal/formats/common/cyclonedxhelpers/hashes.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+func Hashes(p pkg.Package) *[]cyclonedx.Hash {
+	hashes := []cyclonedx.Hash{}
+	if hasMetadata(p) {
+		if metadata, ok := p.Metadata.(pkg.CargoPackageMetadata); ok && metadata.Checksum != "" {
```
wagoodman, line 11: Today we surface checksum digest data from package metadata (in some cases). Additionally we have a file cataloger that observes the actual digests of the files from the source directly (container image or filesystem). Though it is convenient to report out the package metadata we should always report out the real filesystem hashes instead. We can cross correlate the filenames/paths from the package metadata and use that to access any digest information that the `sbom.SBOM` might have (the file digest cataloger is not enabled by default, so there may be no results).

internal/formats/common/cyclonedxhelpers/licenses.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/anchore/syft/internal/spdxlicense"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+func Licenses(p pkg.Package) *cyclonedx.Licenses {
+	lc := cyclonedx.Licenses{}
+	for _, licenseName := range p.Licenses {
+		if value, exists := spdxlicense.ID(licenseName); exists {
```
spiffcs, line 12: Interesting! I wasn't able to find a shared list between `spdx` and `cyclonedx` is this for convenience or did you see in their specification somewhere that they are doing licenses the same way?

internal/formats/common/cyclonedxhelpers/properties.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+func Properties(p pkg.Package) *[]cyclonedx.Property {
```
spiffcs, line 11: Is it correct to deconstruct each package and place its separated `Field -> Value` as a list of properties here?

It does move us towards making sure all information is encoded in the format. The definition of `properties` worries me a little though:

```
Provides the ability to document properties in a name-value store. This provides flexibility to include data not officially supported in the standard without having to use additional namespaces or create extensions. Unlike key-value stores, properties support duplicate names, each potentially having different values.
```

If we put data here that IS supported by the specification is the duplicate data too much noise for the final data shape?

internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxDirectoryPresenter.golden
```diff
@@ -1,10 +1,10 @@
 {
   "bomFormat": "CycloneDX",
   "specVersion": "1.3",
-  "serialNumber": "urn:uuid:a81dc685-cf22-48e0-bda5-65ea1a8bca5b",
+  "serialNumber": "urn:uuid:3305b0e9-bc4a-46f1-87fe-e9707cca5211",
```
spiffcs, line 4: Is this going to be updated each time now? Should we include this and timestamp as redacted fields? I don't think this is something from your changes so maybe I can file a separate PR for this.

spiffcs (commented):
Comments added thanks again @samj1912 for the great PR!

I didn't cover the cpe changes since I added comments in the dependent PR which we can merge before moving on this one.

internal/formats/common/cyclonedxhelpers/licenses.go
stevespringett: CycloneDX support SPDX license IDs and expressions along with unresolved license names. See: https://cyclonedx.org/use-cases/#license-compliance

spiffcs: Thanks for the link @stevespringett!

internal/formats/common/cyclonedxhelpers/hashes.go
wagoodman: I don't think we should enable or disable specific catalogers conditionally based on which format is specified. I think that would start to get difficult to reason about. 

We should think about what catalogers we want in the default cataloger set in the future though. With the subcommand as `packages` enabling the digests cataloger doesn't make sense, but it could in the near future with other subcommands or deprecating the `packages` subcommand.

internal/formats/common/cyclonedxhelpers/properties.go
wagoodman: Fair question. My take: I think it may be easier to lean on the side of larger documents to keep the implementation simpler. If we run into a problem in the future where there are a lot of duplicate fields we could start adding the complexity of removing them.

internal/formats/common/cyclonedxhelpers/properties.go
```diff
@@ -0,0 +1,64 @@
...
+	props := []cyclonedx.Property{}
+	props = append(props, getCycloneDXProperties(p)...)
+	if len(p.Locations) > 0 {
+		for _, l := range p.Locations {
+			props = append(props, getCycloneDXProperties(l.Coordinates)...)
+		}
+	}
+	if hasMetadata(p) {
+		props = append(props, getCycloneDXProperties(p.Metadata)...)
```
wagoodman, line 20: We should additionally capture "pkg.MetadataType" as a property to help in future decoding efforts (how to take a bag of fields and generate a syft package when decoding a syft-generated CycloneDX document back into native syft objects)

internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxDirectoryPresenter.golden
wagoodman: I think the uuid is already redacted in testing https://github.com/anchore/syft/blob/main/internal/formats/cyclonedx13json/encoder_test.go#L32

internal/formats/common/cyclonedxhelpers/properties.go
samj1912: Currently I have only annotated fields that do not directly map to a cyclonedx field with the `cyclonedx` struct tag to allow us to avoid duplication. The properties logic only includes fields with the `cyclonedx` struct tag.

internal/formats/cyclonedx13json/test-fixtures/snapshot/TestCycloneDxDirectoryPresenter.golden
samj1912: I had to update the golden snapshots due to the addition of new fields. As a consequence the uuids were updated when I ran the regeneration commands.

wagoodman: @spiffcs idea for a follow-up PR: we could move the redactor calls to before the "update snapshot" call, therefor we wouldn't need to see UUIDs in the diff, we'd only see "redacted" which would never change.

spiffcs:
Now that the CPE changes have been merged it looks like this one just has some conflicts @samj1912

spiffcs:
Thanks so much for the huge lift on these two massive PR

samj1912:
@wagoodman @spiffcs - made several improvements to the properties output. We now store metadatatype and also pkg.Type. We also store the distro information in the top level sbom metadata properties. I believe this should result in all the information that grype may need to do vuln. matching at this point.

samj1912:
Note - just found out the PR at https://github.com/anchore/syft/pull/685 and the discussion therein. I have updated my PR to include the distro information as a component with type `operating-system` (as noted in the cyclonedx spec). This should alleviate the concerns that were noted at https://github.com/anchore/syft/pull/685 around top level bom metadata use. We should also be able to link appropriate packages to the operating system component through the use of relationships, avoiding duplication of data.

internal/formats/common/cyclonedxhelpers/hashes.go
samj1912: We could definitely use cyclonedx file components to store the file level metadata (if it exists) and correlate that to the package component using bom refs in the future. I am happy to address it in a subsequent PR. This PR is getting a bit too big as is. We could only do it if the file level data/catalogers are enabled.

internal/formats/common/cyclonedxhelpers/external_references.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+func ExternalReferences(p pkg.Package) *[]cyclonedx.ExternalReference {
+	refs := []cyclonedx.ExternalReference{}
+	if hasMetadata(p) {
+		switch metadata := p.Metadata.(type) {
+		case pkg.ApkMetadata:
```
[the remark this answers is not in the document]
samj1912: I have added the package's metadata type as a property for the entire component. That should allow us to determine how these and other component attributes were generated.

samj1912:
@wagoodman / @spiffcs if you could please take a look again that would be great :)

spiffcs:
Yep! Sorry for the delay here @samj1912. I've been looking into signing sbom at the moment. Next time I get to review PRS this is at the top of the stack.

internal/formats/common/cyclonedxhelpers/hashes.go
wagoodman: agreed on not adding scope to this PR, however, I think the only remaining blocking comment (for me) is the use of hash information from package metadata. We should only be including hash information based on direct observation of the file contents and not trusting of the packaging metadata. 

We should remove this function from this PR for the meantime until we can include hashes sourced from the file itself (which is another feature/PR altogether).

wagoodman (commented):
@samj1912 the PR is just about there! My only blocking comment is this one: https://github.com/anchore/syft/pull/710/files#r784914277 .

spiffcs (commented):
Viewed the rest of the file and LGTM. The only blocking comment here is what Alex had mentioned regarding including hashes from metadata files. I also agree with him that we should only be including hashes that we have explicitly read rather than pulling them from the metadata entries.

samj1912:
@wagoodman / @spiffcs should be updated now :)

wagoodman (approved):
@samj1912 stellar contribution... thanks for your work on this!

samj1912:
Thanks @wagoodman! Is this good to merge?

PR #714, opened 2021-12-21
Fix CPE encode/decode when it contains special chars

### LINKED ISSUE #426: Failed to parse CPE - unbind formatted string
**What happened**:
Grype errored and failed to scan an image due to failing to parse CPEs generated based on package names.

Do not error. Generate CPEs based on package names that don't fail to parse in the CPE parser.

Using an image with weird Java package names like:
```
"Jetty :: Http Utility"
"Jetty :: Utilities"
```
Produce sbom.json with syft: `syft -o json image:tag > syft-sbom.json`
Scan with grype using the sbom: `grype -o json sbom:syft-sbom.json > grype-vulns.json`

This will fail with an error similar to:
```
[0004] ERROR failed to catalog: failed to parse CPE="cpe:2.3:a:Jetty :: Http Utility:Jetty :: Http Utility:11.0.2.v20210329:*:*:*:*:java:*:*": unbind formatted string: could not unbind attribute at pos 17
```

**Environment**:
- Output of `grype version`:
```
grype version
Application:          grype
Version:              0.11.0
BuildDate:            2021-04-22T21:51:14Z
GitCommit:            0a46e54aae837135f73d2ff18a84061189ee6f28
GitTreeState:         clean
Platform:             linux/amd64
GoVersion:            go1.16.3
Compiler:             gc
Supported DB Schema:  2
```

### LINKED ISSUE #712: Syft can generate invalid CPEs for debs and rpms with epochs
**What happened**:

Currently debs and rpms may have a `:` in their version to indicate epochs. Syft currently passes these directly to the CPE generator, generating CPEs with an extra colon which violate the CPE spec

Syft should generate valid CPEs

Eg -

```
$ syft centos:7 -o json | grep cpe | grep vim-minimal
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [159 packages]
    "cpe:2.3:a:vim-minimal:vim-minimal:2:7.4.629-7.el7:*:*:*:*:*:*:*",
    "cpe:2.3:a:vim-minimal:vim_minimal:2:7.4.629-7.el7:*:*:*:*:*:*:*",
    "cpe:2.3:a:vim_minimal:vim-minimal:2:7.4.629-7.el7:*:*:*:*:*:*:*",
    "cpe:2.3:a:centos:vim-minimal:2:7.4.629-7.el7:*:*:*:*:*:*:*",
    "cpe:2.3:a:vim:vim-minimal:2:7.4.629-7.el7:*:*:*:*:*:*:*",

```

**Environment**:
- Output of `syft version`:
```
syft version
Application:   syft
Version:       0.33.0
BuildDate:     2021-12-16T14:16:33Z
GitCommit:     a27907659d6c43f5bef2f76a9e381f2bf697a64f
GitTreeState:  clean
Platform:      darwin/amd64
GoVersion:     go1.16.12
Compiler:      gc
```

samj1912:
Fixes #712 
Fixes #426 

Currently syft seems to generate invalid CPEs which do not
conform with the official CPE spec. This is because the underlying
nvdtools library is not a completely spec compliant implementation
and has some interesting bugs/issues.

The following are the list of issues I have encountered with nvdtools:

1. It parses strings which are not CPEs incorrectly as valid CPEs. This
messes up our filter function which is supposed to filter out any
incorrect CPEs we generate. In order to fix this, I have introduced
a new regex in the NewCPE function which follows the upstream spec and
filters out any incorrect CPEs.

2. Introduce wfn.WFNize for any cpe attributes we infer from packages.
This ensures that we are escaping and quoting any special characters
before putting them into CPEs. Note that nvdtools has yet another bug
in the WFNize function, specifically the "addSlashesAt" part of the
function which stops the loop as soon as it encounters ":" a valid
character for a WFN attribute after quoting, but the way nvdtools
handles it causes it to truncate strings that container ":". As a result
strings like "prefix:1.2" which would have been quoted as "prefix\:1.2"
end up becoming "prefix" instead causing loss of information and
incorrect CPEs being generated. As a result in such cases, we remove out
strings containing ":" in any part entirely for now. This is similar
to the way we were handling CPE filtering in the past with http urls as
vendor strings

3. Add special handling for version which contain ":" due to epochs in
debian and rpm. In this case, we strip out the parts before ":" i.e.
the epoch and only output the actual function. This ensures we are not
discarding valid version strings due to pt #.2.

In the future we should look at moving to a more spec compliant cpe
parsing library to avoid such shenanigans.

Edit on Point 2 after the second commit - 

Remove WFNize for input strings
WFNize seems to not be part of the standard as per
pkg.go.dev/github.com/facebookincubator/nvdtools@v0.1.4/wfn#WFNize
and seems to have bugs/issues with encode/decode cycles, so I am
just removing it at this point and relying on the CPE regex to filter
out invalid CPEs for now.

### DIFF
--- a/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
+++ b/syft/pkg/cataloger/common/cpe/sort_by_specificity.go
@@ -29,6 +29,9 @@ func (c BySpecificity) Less(i, j int) bool {
 	}
 
 	// if score and length are equal then text sort
+	// note that we are not using CPEString from the syft pkg
+	// as we are not encoding/decoding this CPE string so we don't
+	// need the proper quoted version of the CPE.
 	return c[i].BindToFmtString() < c[j].BindToFmtString()
 }
 
--- a/syft/pkg/cpe.go
+++ b/syft/pkg/cpe.go
@@ -64,7 +64,24 @@ func normalizeCpeField(field string) string {
 	if field == "*" {
 		return wfn.Any
 	}
-	return strings.ReplaceAll(wfn.StripSlashes(field), `\/`, "/")
+	return stripSlashes(field)
+}
+
+// stripSlashes is a reverse of the sanitize function below.
+// It correctly removes slashes that are followed by allowed puncts.
+// This is to allow for a correct round trip parsing of cpes with quoted characters.
+func stripSlashes(s string) string {
+	const allowedPunct = "-!\"#$%&'()+,./:;<=>@[]^`{|}!~"
+	buf := make([]byte, 0, len(s))
+	for i, c := range s {
+		c := byte(c)
+		if c == '\\' && i+1 < len(s) && strings.IndexByte(allowedPunct, s[i+1]) != -1 {
+			continue
+		} else {
+			buf = append(buf, c)
+		}
+	}
+	return string(buf)
 }
 
 func CPEString(c CPE) string {
--- a/syft/pkg/cpe.go
+++ b/syft/pkg/cpe.go
@@ -13,17 +13,17 @@ type CPE = wfn.Attributes
 // This regex string is taken from
 // https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd which has the official cpe spec
 // This first part matches CPE urls and the second part matches binding strings
-const cpeRegexString = ((`([c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\._\-~%]*){0,6})`) +
+const cpeRegexString = ((`^([c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\._\-~%]*){0,6})`) +
 	// Or match the CPE binding string
 	// Note that we had to replace '`' with '\x60' to escape the backticks
-	`|(cpe:2\.3:[aho\*\-](:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\*\-]))(:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){4})`)
+	`|(cpe:2\.3:[aho\*\-](:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\*\-]))(:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){4})$`)
 
 var cpeRegex = regexp.MustCompile(cpeRegexString)
 
 func NewCPE(cpeStr string) (CPE, error) {
 	// We should filter out all CPEs that do not match the official CPE regex
 	// The facebook nvdtools parser can sometimes incorrectly parse invalid CPE strings
-	if !cpeRegex.Match([]byte(cpeStr)) {
+	if !cpeRegex.MatchString(cpeStr) {
 		return CPE{}, fmt.Errorf("failed to parse CPE=%q as it doesn't match the regex=%s", cpeStr, cpeRegexString)
 	}
 	value, err := wfn.Parse(cpeStr)
--- a/syft/pkg/cpe.go
+++ b/syft/pkg/cpe.go
@@ -66,3 +66,47 @@ func normalizeCpeField(field string) string {
 	}
 	return strings.ReplaceAll(wfn.StripSlashes(field), `\/`, "/")
 }
+
+func CPEString(c CPE) string {
+	output := CPE{}
+	output.Vendor = sanitize(c.Vendor)
+	output.Product = sanitize(c.Product)
+	output.Language = sanitize(c.Language)
+	output.Version = sanitize(c.Version)
+	output.TargetSW = sanitize(c.TargetSW)
+	output.Part = sanitize(c.Part)
+	output.Edition = sanitize(c.Edition)
+	output.Other = sanitize(c.Other)
+	output.SWEdition = sanitize(c.SWEdition)
+	output.TargetHW = sanitize(c.TargetHW)
+	output.Update = sanitize(c.Update)
+	return output.BindToFmtString()
+}
+
+// sanitize is a modified version of WFNize function from nvdtools
+// that quotes all the allowed punctation chars with a slash and replaces
+// spaces with underscores. It differs from the upstream implmentation as
+// it does not use the buggy nvdtools implementation, specifically the "addSlashesAt" part of the
+// function which stops the loop as soon as it encounters ":" a valid
+// character for a WFN attribute after quoting, but the way nvdtools
+// handles it causes it to truncate strings that container ":". As a result
+// strings like "prefix:1.2" which would have been quoted as "prefix\:1.2"
+// end up becoming "prefix" instead causing loss of information and
+// incorrect CPEs being generated.
+func sanitize(s string) string {
+	const allowedPunct = "-!\"#$%&'()+,./:;<=>@[]^`{|}!~"
+	// replace spaces with underscores
+	in := strings.ReplaceAll(s, " ", "_")
+	// we allocate 2x the input to the buffer assuming
+	// we have to quote every character in the input
+	buf := make([]byte, 0, 2*len(in))
+	for _, c := range in {
+		c := byte(c)
+		if strings.IndexByte(allowedPunct, c) != -1 {
+			buf = append(buf, '\\', c)
+		} else {
+			buf = append(buf, c)
+		}
+	}
+	return string(buf)
+}
--- a/syft/pkg/cpe.go
+++ b/syft/pkg/cpe.go
@@ -72,16 +72,15 @@ func normalizeCpeField(field string) string {
 // This is to allow for a correct round trip parsing of cpes with quoted characters.
 func stripSlashes(s string) string {
 	const allowedPunct = "-!\"#$%&'()+,./:;<=>@[]^`{|}!~"
-	buf := make([]byte, 0, len(s))
+	sb := strings.Builder{}
 	for i, c := range s {
-		c := byte(c)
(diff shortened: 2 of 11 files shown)


wagoodman:
@samj1912  thanks for jumping on this!

I have a few reflections on the PR specifically:

_(regarding epoch)_ I'm not certain it is safe to drop epochs from versions, though I see exactly why it was done. If a tool downstream of syft extracts the version from the CPE and uses that as a basis of comparison with other versions from the same product, without epoch it is not clear if the comparison is valid or not (since only versions within the same epoch are directly comparable, otherwise the epoch values themselves should be compared).

_(regarding WFNize)_ You're right that wfn.WFNize doesn't quite do the trick, and the encode/decode path remaining stable is an important item for us. Unless I'm missing something I think the lib is missing something akin to "wfn.UnWFNize" for escaped values... something that isn't easily bolted on in syft (but possible I think).

_(regarding nvdtools replacement?)_ You made an observation in the community slack that I wanted to highlight:

>  I think it might be best to use a diff. go cpe library given all the bugs that I pointed out

I think this is definitely an option that should be on the table. If we dive down this path I think we need to consider how that will impact grype matching by CPE (which is I think one of the reasons this lib was chosen to begin with).

samj1912:
https://github.com/knqyf263/go-cpe seems to be the only other cpe parsing lib that I can find. From a cursory look, it seems to handle quoting and parsing better than the current library but it seems mostly unmaintained at this point.

wagoodman:
I've found a few others, but the one you found is at the top my list of ones to try 👍 . The cases that a CPE library needs to cover (at least) are:

1. valid decode of a CPE formatted string
2. be able to encode a CPE object into a formatted string
3. be able to compare two CPEs (objects or strings)

(there is a chance I missed a case here, but these are the big ones)

The third case comes up in Grype, specifically here:  https://github.com/anchore/grype/blob/4f964c4ee26ad01a80b8bcffb6bf23c0afb71d09/grype/vulnerability/store_adapter.go#L103

This function wraps the nvdtools `MatchWithoutVersion` method: https://github.com/facebookincubator/nvdtools/blob/952d0696a19961ffb4bc18c415df62ea4d00d2fa/wfn/matcher.go#L39-L50

If the new CPE lib doesn't have the exact comparison method we're interested, that may be OK. As long as we can craft it with the components that the lib provides (ideally easily 😄  ). 

Happy to help on this one --odds are this is a pretty big undertaking (could be wrong though!). Feel free to reach out to me on slack!

samj1912:
From slack https://anchorecommunity.slack.com/archives/C4PJFNEEM/p1640131075279100?thread_ts=1640100243.273100&cid=C4PJFNEEM - 

> @Alex Goodman I think I managed to solve the issue - lmk if my reasoning makes sense - 
> 
> instead of sanitizing at cpe generation time, I now do so at decode, this ensures that an encode<>decode cycle produces the same cpe reproducibly and we do not rely on cpe generation to sanitize the output. The wfnize function should now correctly be unwfnize by the string slashes that is called when NewCPE is used and similarly all format decoders now use the new CPEString function which correctly escapes all unwanted chars.

samj1912:
I also noticed that my regex check that was introduced in `NewCPE` was causing a lot of unquoted CPEs to be filtered out before my last change, which is why the encode/decode cycle test was passing - since for all problematic CPEs - it will just filtering them out instead of properly encoding/decoding them. My latest changes should ensure that all CPEs are preserved as before while still being valid now that they are quoted properly.

samj1912:
@wagoodman this should be ready for review now - I had to update the decode function (normalize) to match the encode function (sanitize) since StripSlashes was also not doing what it is supposed to.

syft/pkg/cpe.go
```diff
@@ -2,14 +2,30 @@ package pkg
...
 
 type CPE = wfn.Attributes
 
+// This regex string is taken from
+// https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd which has the official cpe spec
+// This first part matches CPE urls and the second part matches binding strings
+const cpeRegexString = ((`([c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\._\-~%]*){0,6})`) +
+	// Or match the CPE binding string
+	// Note that we had to replace '`' with '\x60' to escape the backticks
+	`|(cpe:2\.3:[aho\*\-](:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\*\-]))(:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){4})`)
```
wagoodman, line 19: this regex is pretty important from the point of view that it restricts which values are allowed to be expressed as CPE objects. It would be ideal to have a good body of tests around `NewCPE` that exercises both positive and negative cases.

samj1912: @wagoodman - added some tests case - also created a data dump using the NVD CPE db and created some random test cases from it to verify our parser.

syft/pkg/cataloger/common/cpe/sort_by_specificity.go
```diff
@@ -29,6 +29,9 @@ func (c BySpecificity) Less(i, j int) bool {
 	}
 
 	// if score and length are equal then text sort
+	// note that we are not using CPEString from the syft pkg
```
spiffcs, line 32: Nice thanks for updating this comment

syft/pkg/cpe.go
```diff
@@ -48,5 +64,66 @@ func normalizeCpeField(field string) string {
...
-	return strings.ReplaceAll(wfn.StripSlashes(field), `\/`, "/")
+	return stripSlashes(field)
+}
+
+// stripSlashes is a reverse of the sanitize function below.
+// It correctly removes slashes that are followed by allowed puncts.
+// This is to allow for a correct round trip parsing of cpes with quoted characters.
+func stripSlashes(s string) string {
+	const allowedPunct = "-!\"#$%&'()+,./:;<=>@[]^`{|}!~"
+	buf := make([]byte, 0, len(s))
```
spiffcs, line 75: Nit `strings.Builder` vs `buffer` here seems like a small consideration since builder gives us `copyCheck` for free.

syft/pkg/cpe.go
```diff
@@ -48,5 +64,66 @@ func normalizeCpeField(field string) string {
...
+// strings like "prefix:1.2" which would have been quoted as "prefix\:1.2"
+// end up becoming "prefix" instead causing loss of information and
+// incorrect CPEs being generated.
+func sanitize(s string) string {
+	const allowedPunct = "-!\"#$%&'()+,./:;<=>@[]^`{|}!~"
+	// replace spaces with underscores
+	in := strings.ReplaceAll(s, " ", "_")
+	// we allocate 2x the input to the buffer assuming
+	// we have to quote every character in the input
+	buf := make([]byte, 0, 2*len(in))
```
spiffcs, line 119: Also nit: not sure if the 2x allocation vs just using `strings.Builder` here makes the assembly faster or not.

syft/pkg/cpe_test.go
```diff
@@ -80,3 +83,98 @@ func Test_normalizeCpeField(t *testing.T) {
...
+	testCases := []struct {
+		CPEString string `json:"cpe-string"`
+		CPEUrl    string `json:"cpe-url"`
+		WFN       CPE    `json:"wfn"`
+	}{}
+	out, err := ioutil.ReadFile("test-fixtures/cpe-data.json")
+	if err != nil {
+		t.Fatal("Unable to read test-fixtures/cpe-data.json: ", err)
+	}
+	json.Unmarshal(out, &testCases)
```
spiffcs, line 97: Nice use of test generation from json file

spiffcs (commented):
Small nits for discussion and thanks so much for the PR @samj1912!

spiffcs:
Core functionality of the PR looks like a great update to help with the issues in the `nvdtools` library. Test cases look good and round trip has been implemented correctly (sanitize <-> original). Small nits on the string building (buffer vs in-library builder). Excellent addition 🥳 .

syft/pkg/cpe_test.go
```diff
@@ -80,3 +83,98 @@ func Test_normalizeCpeField(t *testing.T) {
...
 	}
 }
+
+func Test_CPEParser(t *testing.T) {
+	testCases := []struct {
+		CPEString string `json:"cpe-string"`
+		CPEUrl    string `json:"cpe-url"`
+		WFN       CPE    `json:"wfn"`
+	}{}
+	out, err := ioutil.ReadFile("test-fixtures/cpe-data.json")
```
wagoodman, line 93: this test fixture file has a lot of cases! Is there a way to reduce this to a minimum (or at least smaller) set of cases that still reflect all of the features tested?

wagoodman (commented):
Really great work @samj1912 ! I only have https://github.com/anchore/syft/pull/714#discussion_r777706490 as a comment (as a side note, I don't ever remember leaving a comment on a review that says "this may be too many tests" ... it's a nice problem to have!). If you think that the set of tests can't easily be reduced, we can get this in as is and maybe work on reducing it later (having the tests is better than not).

syft/pkg/cpe.go
```diff
@@ -2,14 +2,30 @@ package pkg
...
 
 type CPE = wfn.Attributes
 
+// This regex string is taken from
+// https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd which has the official cpe spec
+// This first part matches CPE urls and the second part matches binding strings
+const cpeRegexString = ((`^([c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\._\-~%]*){0,6})`) +
+	// Or match the CPE binding string
+	// Note that we had to replace '`' with '\x60' to escape the backticks
+	`|(cpe:2\.3:[aho\*\-](:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\*\-]))(:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&'\(\)\+,/:;<=>@\[\]\^\x60\{\|}~]))+(\?*|\*?))|[\*\-])){4})$`)
```
wagoodman, line 19: this is quite a scary regex! Is this check 100% necessary? or can it be left off during CPE construction? (I presume the very-large test fixture of CPEs covers some of these regex paths in testing)

samj1912: @wagoodman I think it is necessary since we use the `NewCPE` to validate and filter out invalid CPEs. This ensures that syft never puts/parses an invalid CPE in an sbom. I have seen issues where syft outputs CPEs which are not accepted/matched by https://nvd.nist.gov/products/cpe/search

wagoodman: Though this is a large test fixture you're right in the sense that this is making things better. We can always cull back in a future PR or remove if we replace the CPE library we're using in the future.

wagoodman (approved):
Solid work, thanks for the contribution! 💯

PR #721, opened 2021-12-22
Remove unstable CLI test (bad binaries)

wagoodman:
I haven't been able to reproduce locally yet, but in CI this has proven to be unstable.

For the same commit:
   - pass: https://github.com/anchore/syft/runs/4611344142?check_suite_focus=true
   - fail: https://github.com/anchore/syft/runs/4611343586?check_suite_focus=true

For the meantime this test will be commented out, but should be added back in as soon as possible.


spiffcs (approved):
Approved removing flakey test for the interim

PR #727, opened 2021-12-31
support .par for java ecosystems

westonsteimel:
Support .par format for #728.  This succesfully detects log4j in the sample .par file at https://github.com/palantir/log4j-sniffer/blob/develop/examples/inside_a_par/wrapped_in_a_par.par.

### DIFF
--- a/syft/pkg/cataloger/java/archive_filename.go
+++ b/syft/pkg/cataloger/java/archive_filename.go
@@ -113,7 +113,7 @@ func (a archiveFilename) extension() string {
 
 func (a archiveFilename) pkgType() pkg.Type {
 	switch strings.ToLower(a.extension()) {
-	case "jar", "war", "ear", "lpkg":
+	case "jar", "war", "ear", "lpkg", "par":
 		return pkg.JavaPkg
 	case "jpi", "hpi":
 		return pkg.JenkinsPluginPkg
--- a/syft/pkg/cataloger/java/archive_filename_test.go
+++ b/syft/pkg/cataloger/java/archive_filename_test.go
@@ -51,6 +51,13 @@ func TestExtractInfoFromJavaArchiveFilename(t *testing.T) {
 			ty:        pkg.JavaPkg,
 		},
 		{
+			filename:  "pkg-extra-field-maven-4.3.2-rc1.par",
+			version:   "4.3.2-rc1",
+			extension: "par",
+			name:      "pkg-extra-field-maven",
+			ty:        pkg.JavaPkg,
+		},
+		{
 			filename:  "/some/path/pkg-extra-field-maven-4.3.2-rc1.jpi",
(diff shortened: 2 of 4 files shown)


syft/pkg/cataloger/java/archive_filename.go
```diff
@@ -113,7 +113,7 @@ func (a archiveFilename) extension() string {
 
 func (a archiveFilename) pkgType() pkg.Type {
 	switch strings.ToLower(a.extension()) {
-	case "jar", "war", "ear", "lpkg":
+	case "jar", "war", "ear", "lpkg", "par":
```
spiffcs, line 116: I'm not to familiar with the `parchive` format, but I did some initial research and found [this](https://github.com/Parchive/par2cmdline) as a starting point. 

Should we be matching on `par2` as well?
Is it correct to always associate `.par` extensions with `pkg.JavaPkg`?

I don't know the answers to these just yet, but wanted to surface these questions as part of this PR discussion.

spiffcs (commented):
Thanks @westonsteimel solid improvement to our coverage of different archive formats! I just had two questions I added inline regarding the association with `pkg.JavaPkg` and the matching pattern we want to finalize on.

wagoodman:
I don't know a lot about par files, are there any possible differences that we should expect in the archive format or what goes into a `par` file relative to a `jar`? I'm asking relative to our existing java build test fixtures: https://github.com/anchore/syft/tree/main/syft/pkg/cataloger/java/test-fixtures/java-builds ... I was pondering if we should add a test fixture for par explicitly or not. We haven't done this for all java formats if the only difference from syft's perspective is a file extension difference (e.g. JPI vs HPI... I think we only cover JPI).

syft/pkg/cataloger/java/archive_filename.go
westonsteimel: I'm definitely not an expert on this either, I was mostly going based off of the examples from other scanners, but I expect this isn't parchive but rather the Sun Portal Server format at https://docs.oracle.com/cd/E19528-01/819-5030/gbjsz/index.html, which is just another format containing .jar files

westonsteimel: Also, this isn't entirely related, but I did have a thought that it might be useful in the future to extend the catalogers to allow matching on custom patterns via config

westonsteimel: And searching a bit more it seems like .par might be used frequently with SAP applications

spiffcs: Ahh that makes much more sense. Thanks for more context here. I'll see if I can assemble one as well and try and scan it with the changes. You're right that finding real examples of them has proven difficult.

wagoodman (approved):
nice job 💯

PR #732, opened 2022-01-03
Add support for multiple output files in different formats

kzantow:
This PR adds support for multiple `-o` / `--output` options (and similarly in yaml). It continues to support the yaml syntax with a single option of `output: 'json'`. Additionally, each output option can specify a file by appending `=<file>`. For example, to output a Syft JSON _and_ SPDX JSON, one could use: `syft -o json=my.syft.json -o spdx-json=my.spdx.json`

TODO:

- [x] refactor all references to presenter
- [x] add tests


README.md
```diff
@@ -238,8 +252,8 @@ check-for-app-update: true
 
 # a list of globs to exclude from scanning. same as --exclude ; for example:
 # exclude:
-#   - '/etc/**'
-#   - './out/**/*.json'
+#   - "/etc/**"
```
kzantow, line 255: just making this consistent with the other examples using `"` instead of `'`

cmd/packages.go
```diff
@@ -6,11 +6,12 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/anchore/syft/internal/formats"
```
kzantow, line 9: why the heck did `goimports` do this?

cmd/packages.go
```diff
@@ -62,13 +63,11 @@ var (
 		Args:          validateInputArgs,
 		SilenceUsage:  true,
 		SilenceErrors: true,
-		PreRunE: func(cmd *cobra.Command, args []string) error {
-			// set the presenter
-			presenterOption := format.ParseOption(appConfig.Output)
-			if presenterOption == format.UnknownFormatOption {
-				return fmt.Errorf("bad --output value '%s'", appConfig.Output)
+		PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+			writers, err = formats.MakeWriters(formats.ParseOptions(appConfig.Output, format.TableOption, appConfig.File))
```
kzantow, line 67: doing this here so parameter validation happens early, is this right?

cmd/packages.go
```diff
@@ -101,14 +100,14 @@ func setPackageFlags(flags *pflag.FlagSet) {
...
-		fmt.Sprintf("report output formatter, options=%v", format.AllOptions),
+	flags.StringArrayP(
+		"output", "o", []string{string(format.TableOption)},
+		fmt.Sprintf("report output format, options=%v", format.AllOptions),
 	)
 
 	flags.StringP(
 		"file", "", "",
-		"file to write the report output to (default is STDOUT)",
+		"file to write the default report output to (default is STDOUT)",
```
kzantow, line 110: should this be deprecated?

internal/formats/report_writer.go
```diff
@@ -0,0 +1,144 @@
+package formats
```
kzantow, line 1: I wonder if this would be better in `syft/format`? And/or possibly abstracted with an interface so Grype could also use this code to support multiple outputs in a similar way?

internal/ui/ephemeral_terminal_ui.go
```diff
@@ -36,22 +36,20 @@ import (
...
-	frame        *frame.Frame
-	logBuffer    *bytes.Buffer
-	uiOutput     *os.File
-	reportOutput io.Writer
+	unsubscribe func() error
+	handler     *ui.Handler
+	waitGroup   *sync.WaitGroup
+	frame       *frame.Frame
+	logBuffer   *bytes.Buffer
+	uiOutput    *os.File
```
kzantow, line 44: is there a reason this is `os.File` instead of `io.Writer`?

internal/ui/logger_ui.go
```diff
@@ -1,23 +1,18 @@
...
 	"github.com/wagoodman/go-partybus"
 )
 
 type loggerUI struct {
-	unsubscribe  func() error
-	reportOutput io.Writer
+	unsubscribe func() error
 }
 
 // NewLoggerUI writes all events to the common application logger and writes the final report to the given writer.
```
kzantow, line 13: I didn't quite follow how the events are written to the "common application logger" here?

syft/event/parsers/parsers.go
```diff
@@ -109,17 +110,17 @@ func ParseFileIndexingStarted(e partybus.Event) (string, progress.StagedProgress
 	return path, prog, nil
 }
 
-func ParsePresenterReady(e partybus.Event) (presenter.Presenter, error) {
+func ParsePresenterReady(e partybus.Event) (*formats.SBOMWriter, error) {
```
kzantow, line 113: I also am a bit confused about go pointers vs structs. I will do a little reading, but could use a bit of explanation when to use one vs the other...

internal/output/writer.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/anchore/syft/syft/format"
+	"github.com/anchore/syft/syft/sbom"
+	"github.com/hashicorp/go-multierror"
+)
+
+type StreamWriter struct {
```
wagoodman, line 13: `StreamWriter` appears to be an internal concern for this package since `MakeWriter` returns `sbom.Writer` so should be unexported

internal/ui/ephemeral_terminal_ui.go
wagoodman: hmm, I don't recall a specific reason why it's not an `io.Writer`

internal/output/writer.go
kzantow: It is.. except as I've structured the tests, I need to reference it in another package

internal/ui/logger_ui.go
wagoodman: yeah, the phrasing is a bit weird. We could probably even rename this to "NopUI" since it doesn't respond to any events and the caller is expected to have configured the logger to write out to stderr (and be the UI).

cmd/output_writer_test.go
```diff
@@ -0,0 +1,129 @@
...
+			writer, err := makeWriter(test.outputs, file)
+
+			if test.err {
+				assert.Error(t, err)
+				return
+			} else {
+				assert.NoError(t, err)
+			}
+
+			mw := writer.(*output.MultiWriter)
```
wagoodman, line 110: this test is focused around testing `makeWriter` and diving into the concerns of functions in the `output` package. I think it would be easier/cleaner to test the logic of `parseOptions` directly, and not `cmd.makeWriter` at all. Then there could be specific tests in `output` for `output.MakeWriter` to test correct functionality (if we think that's needed).

cmd/packages.go
wagoodman: Is `-o` intended to replace `--file`?

cmd/power_user.go
```diff
@@ -140,8 +144,8 @@ func powerUserExecWorker(userInput string) <-chan error {
 		s.Relationships = append(s.Relationships, mergeRelationships(relationships...)...)
 
 		bus.Publish(partybus.Event{
-			Type:  event.PresenterReady,
-			Value: syftjson.Format().Presenter(s),
+			Type:  event.SBOMReady,
```
wagoodman, line 147: nit: pairing a function with something called "SBOMReady" makes it seem like there is an SBOM object that a subscriber would get, but instead gets a callback. Something like "Exit" doesn't imply receiving anything but doesn't exclude the possibility of a "finalizer" function.

cmd/output_writer_test.go
kzantow: yeah, definitely would clean things up. will get this done right away, not sure what I was thinking. probably programmer laziness

cmd/power_user.go
kzantow: fair point. also, you mentioned the bus isn't exactly supposed to be an event system, so 👍

wagoodman (commented):
Looks great! I only had minor code comments --can you also add a CLI test to test the wiring of multiple `-o` options and assert the formats were written correctly? I think we have some helpers that could get repurposed for this.

cmd/output_writer_test.go
wagoodman: just finished the review: you can ignore my comment on the need to add a unit test for `output.MakeWriter` if we're adding a CLI test anyway 👍

cmd/packages.go
kzantow: I don't really know, but the functionality is redundant. it would seem outputting a table to a file is fairly useless, so generally a user would have to do `-o spdx-json --file <output-file>`, for example. But this is already covered by the `-o spdx-json=<output-file>`...

wagoodman: I see your point on redundancy. 

My take: I think it makes sense to leave the `--file` flag as it's not harming anything by leaving it. We probably don't need to add a deprecation notice if we aren't removing it (we could always decide to remove it in the future).

cmd/output_writer_test.go
```diff
@@ -0,0 +1,78 @@
...
+			_, err := makeWriter(test.outputs, file)
+
+			if test.err {
+				assert.Error(t, err)
+				return
+			} else {
+				assert.NoError(t, err)
+			}
+
+			for _, expected := range test.expected {
```
wagoodman, line 67: nit: could use `t.Cleanup()` to remove files created while testing (dunno if there is a more helpful function in testify for this)

wagoodman: nvm, just saw `t.TempDir()`, I always forget about that option!

wagoodman (approved):
Looks great 👍 nice work!

seabass-labrax:
Great addition; thanks! :smiley:

PR #735, opened 2022-01-06
683 windows filepath

spiffcs:
## Support Windows Directory Resolver
- Add function that converts windows to posix functionality
- Add function that converts posix to windows
- Add build tags to remove windows developer environment errors
- redact carriage return specific windows issues

### Summary
When scanning on windows using the `dir` directive the current directory resolver implementation cannot lookup file nodes that match glob patterns in the `filetree.` By converting Windows filepaths on the way in so that their locations match the `posix` standard, we allow for syft to match on directory resolver matching criteria.

Paths are then translated back into windows paths on the way out so the presentation layer shows users the correct on disk details for the given SBOM items.

#### Considertations
- Exclude functionality needs to be updated in a separate PR
- Location data regarding `Realpath` and `VirtualPath` is sometimes incorrect. I'm currently trying to track down where this bug exists. Should `VirtualPath` ever just be the `Posix` representation as found in the file tree or should we be converting these back?

#### Screenshot demo showing functionality on Windows 11:
image

### DIFF
--- a/syft/source/directory_resolver.go
+++ b/syft/source/directory_resolver.go
@@ -22,6 +22,8 @@ import (
 	"github.com/wagoodman/go-progress"
 )
 
+const WindowsOS = "windows"
+
 var unixSystemRuntimePrefixes = []string{
 	"/proc",
 	"/dev",
@@ -145,7 +147,7 @@ func (r *directoryResolver) indexPath(path string, info os.FileInfo, err error) 
 	}
 
 	// here we check to see if we need to normalize paths to posix on the way in coming from windows
-	if runtime.GOOS == "windows" {
+	if runtime.GOOS == WindowsOS {
 		path = windowsToPosix(path)
 	}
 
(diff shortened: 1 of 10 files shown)


syft/source/directory_resolver.go
```diff
@@ -367,7 +384,13 @@ func (r directoryResolver) FileContentsByLocation(location Location) (io.ReadClo
...
 		return nil, fmt.Errorf("file content is inaccessible path=%q", location.ref.RealPath)
 	}
-	return file.NewLazyReadCloser(string(location.ref.RealPath)), nil
+	// RealPath is posix so for windows directory resolver we need to translate
+	// to its true on disk path.
+	filePath := string(location.ref.RealPath)
+	if runtime.GOOS == WindowsOS {
+		filePath = posixToWindows(filePath)
+	}
+	return file.NewLazyReadCloser(filePath), nil
```
wagoodman, line 393: nice find

wagoodman (approved):
Really nice work --I don't think we need to add window-specific tests for this work quite yet. We can take that on in the future.

PR #738, opened 2022-01-07
Improve SPDX decoding functionality

kzantow:
Adds the ability to ingest SPDX JSON and Tag-Value formats and convert them to internal Syft data model.

This will be the basis for: https://github.com/anchore/grype/issues/395

Note: there is a branch with [a quick hack of SBOM input for Syft](https://github.com/anchore/syft/compare/add-spdx-decoder-take-2...kzantow-anchore:sbom-input?expand=1)


internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+			IDLike:     []string{name},
+			Version:    version,
+			VersionID:  version,
+		}
+	}
+
+	return nil
+}
+
+func findLinuxReleaseByPURL(doc *spdx.Document2_2) *linux.Release {
```
wagoodman, line 95: This function looks like it could be simplified some, either by using guard clauses to reduce indention:
```golang
func findLinuxReleaseByPURL(doc *spdx.Document2_2) *linux.Release {
	for _, p := range doc.Packages {
		purlValue := extractPURL(p)
		if purlValue == "" {
			continue
		}
		
		purl, err := packageurl.FromString(purlValue)
		if err != nil {
			log.Warnf("unable to parse purl: %s", purlValue)
			continue
		} 
		for _, qualifier := range purl.Qualifiers {
			if qualifier.Key != "distro" {
				continue
			}

			parts := strings.Split(qualifier.Value, "-")
			if len(parts) != 1 {
				continue
			}

			name := parts[0]
			version := parts[1]
			return &linux.Release{
				PrettyName: name,
				Name:       name,
				ID:         name,
				IDLike:     []string{name},
				Version:    version,
				VersionID:  version,
			}	
		}
	}


	return nil
}
```

...and/or function decomposition.

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+	}
+	return out
+}
+
+func collectSyftPackages(spdxIDMap map[string]interface{}, doc *spdx.Document2_2) {
+	for _, p := range doc.Packages {
+		syftPkg := toSyftPackage(p)
+		spdxIDMap[string(p.PackageSPDXIdentifier)] = syftPkg
+		for _, f := range p.Files {
```
wagoodman, line 222: it looks like we're decoding `package.Files` into the syft `pkg.Package.Locations` however the inverse is not true (we are not encoding `pkg.Package.Locations` into `package.Files` when making SPDX documents). As a high level comment we should be in parity in the encoding and decoding paths for how we interpret/use fields.

This comment may apply to other fields as well.

internal/formats/spdx22json/decoder_test.go
```diff
@@ -0,0 +1,105 @@
...
+			path:          "alpine-3.10.syft.spdx.json",
+			packages:      []string{"busybox", "libssl1.1", "ssl_client"},
+			relationships: []string{"busybox", "busybox", "libssl1.1", "libssl1.1"},
+		},
+		{
+			path:          "alpine-3.10.vendor.spdx.json",
+			packages:      []string{"alpine", "busybox", "ssl_client"},
+			relationships: []string{},
+		},
+		{
```
wagoodman, line 29: it looks like a few of these test fixtures assert we find no packages, but also decode correctly. Was it intended to find packages in these tests (outside of the `fail: true` cases)?

syft/pkg/url.go
```diff
@@ -11,13 +11,13 @@ import (
 )
 
 const (
-	purlArchQualifier   = "arch"
-	purlDistroQualifier = "distro"
-	purlEpochQualifier  = "epoch"
-	purlVCSURLQualifier = "vcs_url"
+	ArchQualifier   = "arch"
```
wagoodman, line 14: note on naming: renaming these removes the context that it is specifically for pURLs and not for just syft packages.

kzantow: I thought about this, but the fact these are all labeled as "Qualifier" seemed to be sufficient? Maybe we should name these either `PURLQualifier<Name>` or `<Name>PURLQualifier`? They do need to be exported.

internal/formats/spdx22json/decoder_test.go
kzantow: It was not intended to find packages but rather ensure that the SPDX parsed properly. Maybe this should actually end up as an error if we don't find any packages due to the data present?

internal/formats/common/spdxhelpers/to_syft_model.go
kzantow: I'm not sure that is entirely accurate. We should try our best to process any incoming SPDX document rather than just support those created by Syft.

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+	return &sbom.SBOM{
+		Artifacts: sbom.Artifacts{
+			PackageCatalog:    catalog,
+			LinuxDistribution: findSyftLinuxRelease(doc),
+		},
+		Relationships: toSyftRelationships(spdxIDMap, doc),
+	}, nil
+}
+
+func findLinuxReleaseByDocument(doc *spdx.Document2_2) *linux.Release {
```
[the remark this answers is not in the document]
kzantow: This pattern is not widely adopted and also is generally the image scanned rather than a distro. I'd be fine taking this out although it _does_ work in some very limited cases. Not enough cases to make it especially useful, though.

internal/formats/spdx22json/decoder_test.go
wagoodman: > It was not intended to find packages but rather ensure that the SPDX parsed properly.

gotcha, that makes sense 👍 

> Maybe this should actually end up as an error if we don't find any packages due to the data present?

no, I think what you have here already is better; we shouldn't require the existence of packages from SBOMs

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+	if r != nil {
+		return r
+	}
+
+	return nil
+}
+
+func findSpdxRelationshipByType(doc *spdx.Document2_2, typ string) *spdx.Relationship2_2 {
+	for _, r := range doc.Relationships {
+		if typ == r.Relationship {
```
wagoodman, line 143: this utility function looks like it will only return the first instance of a matching relationship type, however, there may be several instances of a type

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+		spdxIDMap[string(f.FileSPDXIdentifier)] = toSyftLocation(f)
+	}
+}
+
+func toSyftRelationships(spdxIDMap map[string]interface{}, doc *spdx.Document2_2) []artifact.Relationship {
+	var out []artifact.Relationship
+	for _, r := range doc.Relationships {
+		// FIXME what to do with r.RefA.DocumentRefID and  r.RefA.SpecialID
+		if r.RefA.DocumentRefID != "" && requireAndTrimPrefix(r.RefA.DocumentRefID, "DocumentRef-") != string(doc.CreationInfo.SPDXIdentifier) {
+			log.Debugf("relationship to external document: %+v", r)
```
wagoodman, line 181: nit: 
```suggestion
			log.Debugf("ignoring relationship to external document: %+v", r)
```

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+		toPackage, toPackageOk := b.(*pkg.Package)
+		toLocation, toLocationOk := b.(*source.Location)
+		if !fromOk || !(toPackageOk || toLocationOk) {
+			log.Debugf("unable to find valid relationship mapping from SPDX 2.2 JSON, ignoring: (from: %+v) (to: %+v)", a, b)
+			continue
+		}
+		var to artifact.Identifiable
+		var typ artifact.RelationshipType
+		switch r.Relationship {
+		case "CONTAINS":
```
wagoodman, line 196: nit: we could move the relationship constants from https://github.com/anchore/syft/blob/main/internal/formats/spdx22json/model/relationship.go to the SPDX helpers and use them instead of raw strings

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+			if toLocationOk {
+				from.Locations = append(from.Locations, *toLocation)
```
wagoodman, line 198: same comment from https://github.com/anchore/syft/pull/738#discussion_r796005173 , a contains relationship to a file doesn't necessarily translate to a syft `pkg.Package.Location`

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+		case "OVERWRITE":
+			typ = artifact.OwnershipByFileOverlapRelationship
+			to = toPackage
+		}
+		if typ != "" && to != nil {
+			out = append(out, artifact.Relationship{
+				From: from,
+				To:   to,
+				Type: typ,
+				Data: nil, // FIXME should there be anything for this Data?
```
wagoodman, line 211: answer: there is nothing needed there today for the `Data` field

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+		var to artifact.Identifiable
+		var typ artifact.RelationshipType
+		switch r.Relationship {
+		case "CONTAINS":
+			if toLocationOk {
+				from.Locations = append(from.Locations, *toLocation)
+				typ = artifact.ContainsRelationship
+				to = toLocation
+			}
+		case "OVERWRITE":
```
wagoodman, line 202: it's not clear that overwrite means the same thing as "OwnershipByFileOverlapRelationship".

internal/formats/syftjson/to_syft_model.go
```diff
@@ -42,6 +46,59 @@ func toSyftLinuxRelease(d model.LinuxRelease) *linux.Release {
...
+		return nil
+	}
+	to, ok := idMap[relationship.Child].(artifact.Identifiable)
+	if !ok {
+		log.Warnf("relationship mapping to key %s is not a valid artifact.Identifiable type: %+v", relationship.Child, idMap[relationship.Child])
+		return nil
+	}
+	typ := artifact.RelationshipType(relationship.Type)
+
+	switch typ {
```
wagoodman, line 86: if this is intended to encapsulate all syft-expressible relationships it would be less brittle to have a helper that checks against all possible relationships from a shared collection (something like `artifacts.AllRelationships()` that returns a slice, or similar). This way as we add relationships this section naturally grows.

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+		MetadataType: metadataType,
+		Metadata:     metadata,
+	}
+
+	sP.SetID()
+
+	return &sP
+}
+
+func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interface{}) {
```
wagoodman, line 296: Adding unit tests for this function would be ideal

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,373 @@
...
+func (p *pkgInfo) qualifierValue(name string) string {
+	for _, q := range p.purl.Qualifiers {
+		if q.Key == name {
+			return q.Value
+		}
+	}
+	return ""
+}
+
+func extractPkgInfo(p *spdx.Package2_2) pkgInfo {
```
wagoodman, line 263: should this PR absorb the functionality from https://github.com/anchore/grype/pull/607 regarding extracting source information from pURLs as well? (this would need to change to account for this)

kzantow: Just so I understand -- this is what we discussed basically incorporating a [port of this commit](https://github.com/anchore/grype/pull/607/commits/7b6058049eed63907dc11b7dbdc00019bcf18e31)?

internal/formats/common/spdxhelpers/to_syft_model.go
kzantow: Ended up deleting this after addressing some other feedback 🎉

internal/formats/syftjson/to_syft_model.go
kzantow: This has been removed during some refactoring

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,343 @@
...
+				Algorithm: string(digest.Algorithm),
+				Value:     digest.Value,
+			})
+		}
+	}
+	return digests
+}
+
+func toFileMetadata(f *spdx.File2_2) (meta source.FileMetadata) {
+	if len(f.FileType) > 0 {
```
jonasagx, line 104: NIT: with `range` if the slice is empty nothing happens, so this `if` check is redundant

internal/formats/spdx22json/decoder_test.go
```diff
@@ -0,0 +1,105 @@
...
+						p, ok := rel.From.(*pkg.Package)
+						if ok && p.Name == pkgName {
+							continue relationships
+						}
+					}
+					assert.NoError(t, fmt.Errorf("Unable to find relationship: %s", pkgName))
+				}
+			}
+
+			fmt.Printf("%+v \n\n\n\n", sbom)
```
jonasagx, line 102: NIT: I assume this print can be removed

jonasagx, line 102:
```suggestion

```

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,343 @@
...
+	var out []artifact.Relationship
+	for _, r := range doc.Relationships {
+		// FIXME what to do with r.RefA.DocumentRefID and  r.RefA.SpecialID
+		if r.RefA.DocumentRefID != "" && requireAndTrimPrefix(r.RefA.DocumentRefID, "DocumentRef-") != string(doc.CreationInfo.SPDXIdentifier) {
+			log.Debugf("ignoring relationship to external document: %+v", r)
+			continue
+		}
+		a := spdxIDMap[string(r.RefA.ElementRefID)]
+		b := spdxIDMap[string(r.RefB.ElementRefID)]
+		from, fromOk := a.(*pkg.Package)
```
jonasagx, line 137: Food for thought: will go 1.18, with generics, helps us simplify conversions like that?

kzantow: I don't think so. the problem is this is runtime and generics as far as I understand them are only compile time

internal/formats/common/spdxhelpers/to_syft_model.go
kzantow: I *think* this has been done already, please do check and tell me if I've missed something!

wagoodman (approved):
I think we're at the finish line 🏁  Awesome work!

It would be a good idea to hold off merging this until we finalize the grype branch that uses this work and verify that everything integrates as expected 👍

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,339 @@
...
+	"github.com/anchore/syft/syft/source"
+)
+
+func ToSyftModel(doc *spdx.Document2_2) (*sbom.SBOM, error) {
+	spdxIDMap := make(map[string]interface{})
+
+	s := &sbom.SBOM{
+		Artifacts: sbom.Artifacts{
+			PackageCatalog:    pkg.NewCatalog(),
+			FileMetadata:      map[source.Coordinates]source.FileMetadata{},
```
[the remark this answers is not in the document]
kzantow: ok, will remember this and follow on a fix later :)

internal/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	return digests
+}
+
+func toFileMetadata(f *spdx.File2_2) (meta source.FileMetadata) {
+	// FIXME Syft is currently lossy due to the SPDX 2.2.1 spec not supporting arbitrary mimetypes
+	for _, typ := range f.FileType {
+		switch FileType(typ) {
+		case ImageFileType:
+			meta.MIMEType = "image/"
```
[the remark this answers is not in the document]
kzantow: not really, but we don't have any additional info, this was the lossiness I was referring to

PR #743, opened 2022-01-11
Align SPDX export more with SPDX 2.2 specification

kzantow:
This aligns the SPDX output more correctly with the SPDX 2.2 spec.

These changes are required for Syft's generated SPDX to import properly using the [spdx/tools-golang jsonloader](https://github.com/spdx/tools-golang/blob/main/jsonloader/jsonloader.go#L15)

Links to some pertinent resources:
* https://spdx.github.io/spdx-spec/package-information
* https://github.com/spdx/spdx-spec/blob/master/schemas/spdx-schema.json

### DIFF
--- a/internal/formats/common/spdxhelpers/origintor.go
+++ b/internal/formats/common/spdxhelpers/origintor.go
@@ -22,7 +22,7 @@ func Originator(p pkg.Package) string {
 			if author == "" {
 				author = metadata.AuthorEmail
 			} else if metadata.AuthorEmail != "" {
-				author = fmt.Sprintf("%s <%s>", author, metadata.AuthorEmail)
+				author = fmt.Sprintf("%s (%s)", author, metadata.AuthorEmail)
 			}
 		case pkg.GemMetadata:
 			if len(metadata.Authors) > 0 {
--- a/internal/formats/common/spdxhelpers/origintor.go
+++ b/internal/formats/common/spdxhelpers/origintor.go
@@ -1,36 +1,39 @@
 package spdxhelpers
 
 import (
-	"fmt"
-
 	"github.com/anchore/syft/syft/pkg"
 )
 
+// Originator needs to conform to the SPDX spec here:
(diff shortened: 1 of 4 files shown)


internal/formats/common/spdxhelpers/origintor.go
```diff
@@ -6,31 +6,35 @@ import (
...
-			return author
 		case pkg.GemMetadata:
 			if len(metadata.Authors) > 0 {
-				return metadata.Authors[0]
+				author = metadata.Authors[0]
 			}
-			return ""
 		case pkg.RpmdbMetadata:
-			return metadata.Vendor
+			return "Organization: " + metadata.Vendor
```
kzantow, line 32: Should this _also_ be `Person: ` ?

internal/formats/spdx22json/model/element.go
```diff
@@ -3,7 +3,7 @@ package model
 type Element struct {
 	SPDXID string `json:"SPDXID"`
 	// Identify name of this SpdxElement.
-	Name string `json:"name"`
+	Name string `json:"name,omitempty"`
```
kzantow, line 6: Not sure if this will have other adverse effects since it is used by things such as [spdx22json.Package](https://github.com/anchore/syft/blob/main/internal/formats/spdx22json/model/package.go#L4) [.Item.Element](https://github.com/anchore/syft/blob/dfefd2ea4e9d44187b4f861bc202970247dd34c8/internal/formats/spdx22json/model/item.go#L4) but not allowed in [spdx22json.File](https://github.com/anchore/syft/blob/dfefd2ea4e9d44187b4f861bc202970247dd34c8/internal/formats/spdx22json/model/file.go#L20-L30), something like this is needed.

internal/formats/common/spdxhelpers/origintor.go
wagoodman: I think we won't be able to tell 100% of the time (org is a good guess)

wagoodman (approved):

PR #752, opened 2022-01-18
Update Syft formats for SyftJson

Toure:
This change will introduce omitempty struct tag to PackageCustomData.
This struct tag will cause null and empty values to be dropped on serialization
for consumers downstream.

Signed-off-by: Toure Dunnon <toure.dunnon@anchore.com>


spiffcs (dismissed):

spiffcs:
@Toure is there more context for this change or a reference for why we need to have these values omitted?

In what cases are you seeing these show up as null on the JSON output?

Vijay-P:
> 

```
        {
            "id": "2666e846-97b0-4f6a-8701-f0bd19ac5e83",
            "name": "Authlib",
            "version": "0.15.2",
            "type": "python",
            "foundBy": "python-index-cataloger",
            "locations": [
                {
                    "path": "/Users/vijay/Documents/enterprise/requirements-test.txt"
                }
            ],
            "licenses": [],
            "language": "python",
            "cpes": [
                "cpe:2.3:a:python_Authlib:python-Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python_Authlib:python_Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python-Authlib:python-Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python-Authlib:python_Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python-Authlib:Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:Authlib:python-Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:Authlib:python_Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python_Authlib:Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python:python_Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python:python-Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:Authlib:Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python:Authlib:0.15.2:*:*:*:*:python:*:*",
                "cpe:2.3:a:python-Authlib:python-Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python_Authlib:python_Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python-Authlib:python_Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python_Authlib:python-Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python_Authlib:Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:Authlib:python-Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python-Authlib:Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:Authlib:python_Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python:python_Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python:python-Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:Authlib:Authlib:0.15.2:*:*:*:*:*:*:*",
                "cpe:2.3:a:python:Authlib:0.15.2:*:*:*:*:*:*:*"
            ],
            "purl": "pkg:pypi/Authlib@0.15.2",
            "metadataType": "",
            "metadata": null
        },
```

Toure:
> @Toure is there more context for this change or a reference for why we need to have these values omitted?
> 
> In what cases are you seeing these show up as null on the JSON output?

When scanning Anchore Enterprise the output SBoM will contain a null value for metadata and metadataType attribute, as an example, Python_Authlib returns a null value for metadata which causes document imports to fail, this change would allow us to omit those fields.

spiffcs:
Nice thanks for the example and comments! I think it's fine to drop these fields specifically since they would always be a nulll type and not something like `""` or `0`.

I will defer approval in this case to @wagoodman because maybe he has context or a good reason why we might want to update the `UnMarshalJSON` function to account for `"metadataType": "",` and set a value rather than drop the field completely from the output.

spiffcs (commented):
I'll sync with the tools team about the above comments. I might just be being too careful here about dropping fields from the output if empty.

spiffcs:
The CLI tests are still failing - so those will need a quick update to reflect the new desired output.

wagoodman:
@Toure I can help with the json schema update (where CLI tests are failing)

wagoodman (approved):

PR #754, opened 2022-01-18
Improve package URL support

wagoodman:
This PR does primarily enhances up support for OS-related packages to include distro version information in any pURL generated:
```diff
- pkg:alpine/name@1.0.1?arch=amd64
+ pkg:alpine/name@1.0.1?arch=amd64&distro=alpine-3.4.6
```
The `ID` and `VersionID` are encoded into the pURL as a `distro` get-param; see [the pURL spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#deb) for more info. 

Additionally this PR:
- migrates the existing package URL functionality to the `pkg` package (since this semantically represents packages)
- beefs up testing around pURL generation to ensure that we are exercising all package types
- during this work it was found that the NPM metadata filename within `pkg` was inconsistent --the filename was updated

Related to https://github.com/anchore/grype/issues/395

Note: this PR cannot be merged until https://github.com/anchore/syft/pull/753 is merged and this PR is rebased.


kzantow (approved):
If you think this is ready, 👍

syft/pkg/dpkg_metadata_test.go
```diff
@@ -12,39 +12,52 @@ import (
...
+				VersionID: "16.04",
 			},
 			metadata: DpkgMetadata{
 				Package:      "p",
 				Source:       "s",
 				Version:      "v",
 				Architecture: "a",
 			},
-			expected: "pkg:deb/ubuntu/p@v?arch=a",
+			expected: "pkg:deb/ubuntu/p@v?arch=a&distro=ubuntu-16.04",
```
kzantow, line 45: One question, though: should these be more like `pkg:deb/ubuntu/p@v?arch=a&distro=16.04` without the `ubuntu-`? (After looking again at: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#deb )

wagoodman:
(force pushed to rebase from main now that https://github.com/anchore/syft/pull/753 is merged)

syft/pkg/dpkg_metadata_test.go
wagoodman: Honestly, I'm not certain. The spec only has one "good" example with this here: https://github.com/package-url/purl-spec/blob/1b1e9b2afa6de1c855225fb08cb46878ad653925/PURL-SPECIFICATION.rst#some-purl-examples
```
pkg:deb/debian/curl@7.50.3-1?arch=i386&distro=jessie
pkg:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
```

The debian distro code-name is a bad idea for us, so I didn't approach it that way... so the fedora example was used as a basis. That being said, there is essentially 0 verbiage on correct usage of `distro` even though it is a suggested qualifier (that we need). 

References:
- https://github.com/package-url/purl-spec/tree/a21a9ecbab171fe06d1ac88843bc3721a42dbed0#known-qualifiers-keyvalue-pairs
- https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm

Since there is little direction from the spec I'm open to other options here.

wagoodman: From an offline conversation: we'll leave this as is for now since it seems more semantically correct to the term "distro" but are open to changing in the future.

wagoodman:
looks like the codeql workflow hasn't kicked in after a long while (hours)

Screen Shot 2022-01-19 at 5 11 14 PM

I think this is due to the fact that the base branch was originally not `main` but was automatically switched when the reference PR merged.

wagoodman:
I'm going to close and reopen this PR to hopefully trigger the pull_request events for this PR.

wagoodman:
That's more like it 💯 

Screen Shot 2022-01-19 at 5 24 32 PM

PR #755, opened 2022-01-19
update sign sed command to include windows zip

spiffcs:
- remove replacements so install script is correct
- update checksum sign so windows zip is included


kzantow (approved):
LGTM 🤷

PR #758, opened 2022-01-20
[CycloneDX] Add artifactID and groupID to the cycloneDX properties (support lower level struct as properties)

pbalogh-sa:
Signed-off-by: Peter Balogh <p.balogh.sa@gmail.com>

related: #685 , #710 
In the case of java add artifactID and groupID pomProperties to the cycloneDX properties.
- support adding lower level struct to the properties.
- add `groupID` as a `group` to the CDX component in the case of Java

### DIFF
--- a/internal/formats/common/cyclonedxhelpers/properties.go
+++ b/internal/formats/common/cyclonedxhelpers/properties.go
@@ -34,8 +34,8 @@ func getCycloneDXProperties(m interface{}) *[]cyclonedx.Property {
 	structType := structValue.Type()
 	for i := 0; i < structValue.NumField(); i++ {
 		if name, value := getCycloneDXPropertyName(structType.Field(i)), getCycloneDXPropertyValue(structValue.Field(i)); name != "" && value != "" {
-			// n the case of the value is a struct and has cyclonedx tag with name "-"
-			//
+			// In the case of the value is a struct and has cyclonedx tag with name "-"
+			// call the getCycloneDXProperties recursively.
 			if name == "-" && reflect.ValueOf(value).Kind() == reflect.Struct {
 				props = append(props, *getCycloneDXProperties(value)...)
 			} else {
--- a/internal/formats/common/cyclonedxhelpers/properties.go
+++ b/internal/formats/common/cyclonedxhelpers/properties.go
@@ -28,17 +28,22 @@ func Properties(p pkg.Package) *[]cyclonedx.Property {
 func getCycloneDXProperties(m interface{}) *[]cyclonedx.Property {
 	props := []cyclonedx.Property{}
 	structValue := reflect.ValueOf(m)
-	// we can only handle top level structs as interfaces for now
 	if structValue.Kind() != reflect.Struct {
 		return &props
 	}
 	structType := structValue.Type()
 	for i := 0; i < structValue.NumField(); i++ {
 		if name, value := getCycloneDXPropertyName(structType.Field(i)), getCycloneDXPropertyValue(structValue.Field(i)); name != "" && value != "" {
-			props = append(props, cyclonedx.Property{
-				Name:  name,
-				Value: value,
-			})
+			// n the case of the value is a struct and has cyclonedx tag with name "-"
+			//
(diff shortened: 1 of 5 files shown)


internal/formats/common/cyclonedxhelpers/properties.go
```diff
@@ -63,6 +68,8 @@ func getCycloneDXPropertyValue(field reflect.Value) string {
 		return ""
 	case reflect.Ptr:
 		return getCycloneDXPropertyValue(reflect.Indirect(field))
+	case reflect.Struct:
```
samj1912, line 71: We should move this up to line 65 instead. Having the `field.CanInterface()` check is important to avoid potential panics during reflection.

syft/pkg/java_metadata.go
```diff
@@ -32,8 +32,8 @@ type JavaMetadata struct {
 type PomProperties struct {
 	Path       string            `mapstructure:"path" json:"path"`
 	Name       string            `mapstructure:"name" json:"name"`
-	GroupID    string            `mapstructure:"groupId" json:"groupId"`
-	ArtifactID string            `mapstructure:"artifactId" json:"artifactId"`
+	GroupID    string            `mapstructure:"groupId" json:"groupId" cyclonedx:"groupID"`
```
samj1912, line 35: I believe we should explicitly handle groupId/artifactId the way it is specified in the CDX spec - i.e. using the structured `group` and `name` field of a component. https://cyclonedx.org/docs/1.4/json/#components_items_name rather than a property.

We can also take a look at the reference CycloneDX maven implementation for reference on how it handles these fields - https://github.com/CycloneDX/cyclonedx-maven-plugin/blob/684b4d59012396e8548faa4d8460bc49c2af79f9/src/main/java/org/cyclonedx/maven/BaseCycloneDxMojo.java#L436-L439

cc: @stevespringett who will know more.

internal/formats/common/cyclonedxhelpers/properties.go
```diff
@@ -28,17 +28,22 @@ func Properties(p pkg.Package) *[]cyclonedx.Property {
...
-				Value: value,
-			})
+			// In the case of the value is a struct and has cyclonedx tag with name "-"
+			// call the getCycloneDXProperties recursively.
+			if name == "-" && reflect.ValueOf(value).Kind() == reflect.Struct {
+				props = append(props, *getCycloneDXProperties(value)...)
+			} else {
+				props = append(props, cyclonedx.Property{
+					Name:  name,
+					Value: fmt.Sprintf("%s", value),
```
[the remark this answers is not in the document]
pbalogh-sa: I added a check for the type of value.

internal/formats/common/cyclonedxhelpers/properties.go
pbalogh-sa: I added `canInterface` check for filed.

syft/pkg/java_metadata.go
pbalogh-sa: I think the lower level support of structs can be feasible for the future.
At the moment the `group` is missing and putting it into the CDX properties don't cause any issues.
For example, there is a type and language in the properties that can be easily extracted from the PURL.

pbalogh-sa: @samj1912  I added `group` filed with `groupID` in the case of java. I think the `grouID` and `artifactID` still can be a part of the properties. WDYT?

stevespringett: The `group` field in CycloneDX is intended to be a drop-in for Maven groupId and NPM scope names. The CycloneDX implementations for Maven, Gradle, Webpack, and NodeJS all use the group field.

stevespringett: See also https://cyclonedx.org/docs/1.4/json/#components_items_group

pbalogh-sa: @stevespringett Thanks for your explanation. So filling the group filed with groupID is OK.

wagoodman (approved):
Excellent work 🏅  --this will help immensely with the upcoming anchore/grype#481 work! 🙌

Issue #510, opened 2021-09-20
Ability to sign or attest the generated SBOM


luhring:
### Background

Syft can already generate SBOMs in a variety formats (SPDX, CycloneDX, lossless Syft JSON, etc.). These SBOMs can be saved and used in a number of downstream tasks, such as vulnerability scanning w/ [Grype](https://github.com/anchore/grype).

We recently looked at a **supply chain security workflow** that leverages Cosign to sign the SBOM from Syft before it's used downstream. This workflow was published as a blog post: https://anchore.com/blog/drop-an-sbom-how-to-secure-your-software-supply-chain-using-open-source-tools/

Ideally, there would be **no opportunity for the SBOM to be tampered with** before it's signed.

### The new feature

Syft could have an option to sign (or even _attest_) the SBOM immediately as it's produced, using an identity provided by the user, to avoid any chance of tampering with the SBOM.

### Implementation

This could be implemented by integrating [Cosign](https://github.com/SigStore/cosign) into Syft. By doing so, this could allow Syft users to pass Cosign-related keys and/or configuration to Syft.

A simple case would be to use Cosign's `sign-blob` functionality to sign the SBOM:

```
syft example:latest -o spdx --sign ./cosign.key
```

Syft could even leverage Cosign's [Keyless Signatures](https://github.com/sigstore/cosign/blob/main/KEYLESS.md) workflow (creating keys via OIDC auth and publishing the signature to Rekor), to avoid the need for users to specify any keys:

```
syft example:latest -o spdx --sign-keyless
```

Alternatively (or in addition), Syft could create an [in-toto attestation](https://github.com/in-toto/attestation/blob/main/spec/README.md) for the SBOM (as the ["predicate"](https://github.com/in-toto/attestation/blob/main/spec/README.md#predicate)) with respect to the image (as the "subject"), akin to Cosign's `attest` command:

```
syft example:latest -o spdx --attest ./cosign.key
```

### Open questions

1. What CLI syntax makes the most sense to the user for these operations?
2. Should Syft just be able to **_sign_** the SBOM? **_Attest_** the SBOM? Both?
3. Should Syft always push the signatures or attestations to the registry?
4. Should Syft output the signature/attestation somewhere local, and if so, where to? (the SBOM itself is already being sent to stdout)
5. Should Syft refuse to sign/attest images _not_ in a registry? (The input to the signing is a registry manifest.)
[...]

PR #759, opened 2022-01-20
Add attestation workflow to README

spiffcs:
While #510 is still in flight and waiting on changes in the upstream sigstore codebase I added the current attestation workflow for how to attach a syft generated SBOM to a signed image.


README.md
```diff
@@ -414,3 +414,36 @@ anchore:
   dockerfile: ""
 
 ```
+
+### Attaching an SBOM to an image as an attestation
```
luhring, line 418: One small consideration: since "attach" and "attest" have two similar but distinct meanings within the cosign ecosystem, we may want to avoid mixing those terms when describing a single functionality, to help avoid confusion for users.

README.md
```diff
@@ -414,3 +414,36 @@ anchore:
...
+```
+
+You should see this output along with the attached SBOM.
+```bash
+Verification for test/image:latest --
+The following checks were performed on each of these signatures:
+  - The cosign claims were validated
+  - The signatures were verified against the specified public key
+  - Any certificates were verified against the Fulcio roots.
+```
```
luhring, line 447: nit: We might not want to render this as "bash", since it's just English text (albeit outputted in the shell)

luhring (commented):
Great idea and great write-up! I added a couple of nit thoughts for your consideration, but this looks nice.

jonasagx (approved):

PR #765, opened 2022-01-23
Refactor install.sh

wagoodman:
This PR attempts to cleanup the install.sh script, which was originally generated by [godownloader](https://github.com/goreleaser/godownloader). Godownloader uses the project's `.goreleaser.yaml` to autogenerate the install.sh to ensure that the releases always stay in step with the install.sh features (as long as you continually run godownloader). We have not been re-running godownloader since it is now a deprecated project and there is no guarantee that the generated install.sh will function with the release.

This, coupled with multiple recent install.sh related issues that have been filed (in both syft and grype) has led to this PR, which makes the following changes:
- refactor of the install.sh to be more friendly to continual maintenance (remove dependence on mutating global variables, breakdown functions into smaller more focused responsibilities, etc)
- reworks the logic of the install.sh script to depend on the published checksums file for each release. This file is searched for the appropriate asset to install. Note: dmg and darwin zip files are not included in the checksums file in a production release
- adds unit tests for key bash functions, including serving a current release-like snapshot directory and testing the download and install functions to ensure that all assets can be installed by the install.sh functions
- adds unit test to test that install.sh can download assets from an actual release. This is due to the fact that the signing process changes the checksums file and zip/dmg assets for mac, so testing only against snapshots is not good enough.
- adds acceptance tests that use the install.sh script with more modern `bash` and less modern `sh` environments to prove portability on linux systems
- adds acceptance test to ensure that install.sh script works on a mac to install a DMG

The above changes are useful for the following primary point: _as we create snapshot builds (which is essentially the output of the `.goreleaser.yaml` specific configuration) we are testing the install.sh against a pseudo-release, so we can see format/os/arch related failures before cutting releases._

There are some caveats:
1.  snapshot != release: the DMG is not created by the build process, but instead, the signing process, which is not performed on snapshot builds. For this reason all-things related to installing DMGs is not included in unit testing or the search-by-checksum file approach. Instead a single acceptance test is used to ensure that installing from an existing release with the current install.sh works alright. (maybe in the future DMG creation can be brought under the standard asset release process?)
2. The acceptance-level install.sh usage is against the latest release. This leaves us open to possible problems down the line when/if we introduce breaking changes to how we install assets. Ideally we would test against a mocked current release like the unit tests do [+ mocking the github api calls] ... however, that doesn't seem necessary at this time.

Another way to look at these caveats (after an offline sync during review):
- We are partially testing a T=0: "most" of the upcoming release via snapshots, there is a small uncovered part of snapshot-oriented testing that is not covered (see caveat 1).
- We are fully testing a T=-1: "all" of the last release is tested via downloading and installing that release.
- When we release at T=1: there is still a risk of failure due to caveat 1 even though testing before the release passed, leaving  you in an unreleasable state (remember, now you need to pass tests for against your latest "broken" release, so you're stuck). The workaround is to comment out offending tests in the short term.

Ideally:
- snapshot = release: this requires refactoring the release pipeline to not create new assets during the signing process and to be able to complete the signing process locally during snapshot builds.
- acceptance-level testing should only test against the current snapshot (T=0), not the latest release (T=-1).

Alternative approach considered: I prototyped a version of the install.sh that uses the github API to discover the available assets, however, this was more complicated than using purely the checksums file and we would still use the checksums file anyway for verification of downloads.

Related to anchore/project-toolbox#113


westonsteimel:
@wagoodman, I just gave the install.sh script a try on my M1 and got the following:

```sh
[info] checking github for the current release tag
[info] using release tag='v0.36.0' version='0.36.0' os='darwin' arch='arm64'
[error] could not find release asset for os='darwin' arch='arm64' format='zip'
[error] failed to install syft
```

wagoodman:
great catch @westonsteimel , I see why that wouldn't work --the zip checksum is removed from the checksums file on release (but not snapshot builds), a behavior we hope to correct soon when bumping goreleaser and changing the release process. 

I'll make some updates to account for this 👍

wagoodman:
@westonsteimel mind trying it out now on your M1?

westonsteimel:
@wagoodman, it works now!

test/install/2_download_asset_test.sh
```diff
@@ -0,0 +1,43 @@
+. test_harness.sh
```
spiffcs, line 1: Is this the name we want for this file? Should it be `download_asset_release_test`?

test/install/Makefile
```diff
@@ -0,0 +1,103 @@
...
+BUSYBOX_IMAGE=busybox:1.35
+
+ENVS=./environments
+DOCKER_RUN=docker run --rm -t -w /project/test/install -v $(shell pwd)/../../:/project
+UNIT=make unit-local
+
+# acceptance testing is running the current install.sh against the latest release. Note: this could be a problem down
+# the line if there are breaking changes made that don't align with the latest release (but will be OK with the next
+# release)
+acceptance=sh -c '../../install.sh -b /usr/local/bin  && syft version'
```
spiffcs, line 15: Would it be ok to uppercase this variable so that we make it distinct from our make target on line 31?

test/install/Makefile
```diff
@@ -0,0 +1,103 @@
...
+		$(acceptance)
+	@echo "\n*** test note: you should see syft spit out a 'x509: certificate signed by unknown authority' error --this is expected ***"
+
+busybox-1.35:
+	$(call title,busybox-1.35 - build environment)
+	docker pull $(BUSYBOX_IMAGE)
+
+## For CI ########################################################
+
+.PHONY: cache.fingerprint
```
spiffcs, line 100: I think we can remove all the `PHONY` directives here since these targets will not be dependencies on outside files/folders

test/install/test_harness.sh
```diff
@@ -0,0 +1,163 @@
...
+    exit 2
+  fi
+}
+
+log_test_case() {
+  echo "  running $@"
+}
+
+run_test_case_with_snapshot_release() {
+  log_test_case ${@:1}
```
spiffcs, line 77: In this case why does line 77 not execute, but line 83 will execute?

spiffcs (approved):
Approved with comments

install.sh
```diff
@@ -342,51 +282,452 @@ hash_sha256_verify() {
...
+)
+
+# extract_json_value [json] [key]
+#
+# outputs value of the key from the given json string
+#
+extract_json_value() (
+  json="$1"
+  key="$2"
+  key_value=$(echo "${json}" | grep  -o "\"$key\":[^,]*[,}]" | tr -d '",}')
```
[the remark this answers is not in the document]
wagoodman: only in multiline cases --yes this is a limitation. In condensed/non-multiline cases (which the GitHub API does) this is not a problem (we match on `}`). We can revisit this in the future if it becomes a problem.

test/install/Makefile
wagoodman: I realized after the review why this is here: to ignore the fact that the file already exists. When invoked we always want to run this target, regardless of the file already existing. The fingerprint should always be evaluated with the latest source. A real makefile would add all dependent files as targets themselves, but I'll skip that :)

test/install/test_harness.sh
wagoodman: in bash (not posix I think) when you pass items as parameters they are not evaluated, unless explicitly done so (e.g. `$($1)`). When used alone as not parameters they are implicitly evaluated --a more explicit approach would be to use `eval ${@:1}` I think.

PR #768, opened 2022-01-24
Add dependencies to cyclonedx

spiffcs:
Updated CycloneDX relationships based off of #706


hectorj2f:
Lgtm

internal/formats/common/cyclonedxhelpers/format.go
```diff
@@ -97,18 +105,63 @@ func toBomDescriptor(name, version string, srcMetadata source.Metadata) *cyclone
 	}
 }
 
+func lookupRelationship(ty artifact.RelationshipType) bool {
```
wagoodman, line 108: nit: the name of this function doesn't really convey what its purpose is, which is to filter down relationship types to those that are meant for package dependencies

internal/formats/common/cyclonedxhelpers/format.go
```diff
@@ -97,18 +105,63 @@ func toBomDescriptor(name, version string, srcMetadata source.Metadata) *cyclone
...
+	switch ty {
+	case artifact.OwnershipByFileOverlapRelationship:
```
wagoodman, line 110: I'm not 100% sure that this relationship belongs here, but I could be wrong

spiffcs: I can remove this one for now.

spiffcs:
@wagoodman updated based on the PR feedback

internal/formats/common/cyclonedxhelpers/format.go
wagoodman: That's better; what about something like `isExpressiblePackageRelationship`? Beyond naming, I think a helpful comment to help future devs figure which qualities of a relationship indicate that they should be included here.

internal/formats/common/cyclonedxhelpers/format.go
```diff
@@ -97,18 +105,61 @@ func toBomDescriptor(name, version string, srcMetadata source.Metadata) *cyclone
...
+	}
+	return false
+}
+
+func toDependencies(relationships []artifact.Relationship) []cyclonedx.Dependency {
+	result := make([]cyclonedx.Dependency, 0)
+	for _, r := range relationships {
+		exists := relationshipTypeExists(r.Type)
+		if !exists {
+			log.Warnf("unable to convert relationship from CycloneDX 1.3 JSON, dropping: %+v", r)
```
wagoodman, line 127: nit: this warning indicates a specific CycloneDX version, but this helper could be used for any CycloneDX version

spiffcs:
looks like my changes to the tests did not get factored in

wagoodman:
> looks like my changes to the tests did not get factored in

also other changes didn't survive the rebase:
```
...
func lookupRelationship(ty artifact.RelationshipType) bool {
...
```

wagoodman (approved):
survived the rebase!

hectorj2f:
Thanks guys 👏🏻

PR #777, opened 2022-01-26
Extract language and package type from pURLs on SBOM decode

wagoodman:
This PR paves the way to use pURLs as a way to indicate package language and package type, two attributes that are not easily encoded into in-spec properties for CycloneDX and SPDX. This will be useful when decoding SBOM documents that did not necessarily originate from Syft, but do have pURL information available.

Possible future SPDX decode function:
```golang
func toSyftPackage(p model.Package) pkg.Package {
	purl := spdxhelpers.ExtractPURL(p.ExternalRefs)
	sP := pkg.Package{
		Type:     pkg.TypeFromPURL(purl), //  <---- new function 
		Name:     p.Name,
		Version:  p.VersionInfo,
		Licenses: spdxhelpers.ParseLicense(p.LicenseDeclared),
		CPEs:     spdxhelpers.ExtractCPEs(p.ExternalRefs),
		PURL:     purl,
		Language: pkg.LanguageFromPURL(purl),  //  <---- new function 
	}

	sP.SetID()

	return sP
}

```

Additionally adds pURL support for cargo and npm package types.

Related to anchore/grype#395

### DIFF
--- a/syft/pkg/npm_package_json_metadata.go
+++ b/syft/pkg/npm_package_json_metadata.go
@@ -1,11 +1,32 @@
 package pkg
 
+import (
+	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/syft/linux"
+)
+
+var _ urlIdentifier = (*NpmPackageJSONMetadata)(nil)
+
 // NpmPackageJSONMetadata holds extra information that is used in pkg.Package
 type NpmPackageJSONMetadata struct {
+	Name        string   `mapstructure:"name" json:"name"`
+	Version     string   `mapstructure:"version" json:"version"`
 	Files       []string `mapstructure:"files" json:"files,omitempty"`
 	Author      string   `mapstructure:"author" json:"author"`
 	Licenses    []string `mapstructure:"licenses" json:"licenses"`
 	Homepage    string   `mapstructure:"homepage" json:"homepage"`
 	Description string   `mapstructure:"description" json:"description"`
 	URL         string   `mapstructure:"url" json:"url"`
+}
+
+// PackageURL returns the PURL for the specific NPM package (see https://github.com/package-url/purl-spec)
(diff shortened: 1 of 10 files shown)


syft/pkg/npm_package_json_metadata.go
```diff
@@ -1,11 +1,32 @@
...
+// PackageURL returns the PURL for the specific NPM package (see https://github.com/package-url/purl-spec)
+func (p NpmPackageJSONMetadata) PackageURL(_ *linux.Release) string {
+	return packageurl.NewPackageURL(
+		packageurl.TypeNPM,
+		"",
+		p.Name,
+		p.Version,
+		nil,
+		"",
+	).ToString()
```
jonasagx, line 31: NIT: go's https://pkg.go.dev/fmt#Stringer uses `String() string` as signature.

jonasagx: So for example 
```golang
purl := packageurl.NewPackageURL(
 		packageurl.TypeNPM,
 		"",
 		p.Name,
 		p.Version,
 		nil,
 		"",
 	)

fmt.Printf("purl: %s", purl) // Printf will look for purl.String()
```

jonasagx: nvm there is an alternative: https://github.com/package-url/packageurl-go/blob/master/packageurl.go#L189

jonasagx (approved):

kzantow (approved):
👍

syft/pkg/npm_package_json_metadata.go
wagoodman: it seems like `p.ToString()` and `p.String()` are really the same call on this object (not certain why they didn't just provide `String` and leave off `ToString`)

PR #785, opened 2022-01-31
510 - SBOM attestation stdout

spiffcs:
### Attestation Stdout

This PR adds the initial version of the `syft attest` command. A user can use `attest` to generate a summary of discovered packages formatted as the predicate to an image attestation.

To run this command you'll need to generate a key via `cosign generate-key-pair`. You can either pass it directly to the command or have syft read from the default location `./cosign.key`

```
go run main.go attest --key cosign.key --output json anchore/syft:latest 
```

Notes:
- compile size increases from ~22mb --> ~60mb due to new library imports
- CI time has increased for static analysis
- keyless workflow will be added after; see [here](https://github.com/anchore/syft/issues/510) for outline of what is to come
- snapshot build times have increased with the introduction of new cosign libraries

TODO:

- [x] update user input keysign options to be more friendly
- [x] update to allow password inputs for keys
- [x] inject command context to leverage correct cosign function signatures
- [x] unit/integration coverage
- [x] command coverage 
- [x] rebase
- [x] BUG: keyreference panic found for certain keyoptions
- [x] Question: Integration with SBOM writer/encoding pattern or keep write to stdout separate flow
- [x] Question: static-analysis has increased to 4m beyond 2m timeout


luhring:
Very excited about this!

I've started playing with the functionality on this branch (at e51334d), and here are some notes I have so far:

1. I'm seeing the text "an error occurred..." get outputted to STDERR in all the ways I've run it so far.

```console
$ go run . attest alpine:latest --key ~/cosign.key
an error occurred: no provider found for that key reference, will try to load key from disk...
...
```

2. Unless my cosign private key password is set in `COSIGN_PASSWORD`, I'm getting prompted to enter the key's password **two times**, even when the first entry is correct.

```console
$ go run . attest alpine:latest --key ./cosign.key
an error occurred: no provider found for that key reference, will try to load key from disk...
Enter password for private key:
Enter password for private key again:
```

Also, I'm finding that if I don't provide the correct entry on the second attempt too, I get the message `reading key: passwords do not match`.

3. The attestation bundle is getting sent to STDERR, and it should be STDOUT.

```console
$ go run . attest alpine:latest --key ./cosign.key 1> std.out 2> std.err
$ head -c 200 std.*
==> std.err <==
an error occurred: no provider found for that key reference, will try to load key from disk...
{"payloadType":"application/syft.in-toto+json","payload":"eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdGF0ZW1
==> std.out <==
$ wc -c std.*
   80468 std.err
       0 std.out
   80468 total
```

4. The attestation bundle's `payloadType` field value is `application/syft.in-toto+json`, but it should be `application/vnd.in-toto+json`.

Just dumping what I have so far — I'll keep playing!

spiffcs:
@luhring Just to go through the points you made:
1) This comes from the cosign library when trying to read the key from disk. Do you want me to see if we can eliminate this error from the output?

2) I set the password to be `confirm` https://github.com/anchore/syft/blob/2df53eab7d957731b1d86b0e9b274a4b08363372/cmd/attest.go#L89-L110

Do you want just one pass for the input?

3) This has been updated to stdout

4) I will update the payloaType field

luhring:
@spiffcs Thanks!

1. If we can eliminate the message, I think that'd be ideal. My reaction to it is that it's confusing, and as a user I'm not sure what I'm supposed to do with that information. If it's difficult to suppress, this is good feedback for the Cosign community.

2. Hmm, maybe I'm just missing something... what is the motivation for asking the user for their private key passphrase twice?

Makefile
```diff
@@ -111,6 +111,7 @@ bootstrap-tools: $(TEMPDIR)
 	curl -sSfL https://raw.githubusercontent.com/anchore/chronicle/main/install.sh | sh -s -- -b $(TEMPDIR)/ v0.3.0
 	.github/scripts/goreleaser-install.sh -d -b $(TEMPDIR)/ v1.4.1
 	GOBIN="$(shell realpath $(TEMPDIR))" go install github.com/neilpa/yajsv@v1.4.0
+	GOBIN="$(shell realpath $(TEMPDIR))" go install github.com/sigstore/cosign/cmd/cosign@latest
```
wagoodman, line 114: cosign should be pinned to a specific version (not `latest`)

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+		return fmt.Errorf("could not support attestation for %s; please try one of these scheme: %v", userInput, validAttestScheme)
+	default:
+		return nil
+	}
+}
+
+func attestExec(ctx context.Context, _ *cobra.Command, args []string) error {
+	// can only be an image for attestation or OCI DIR
+	userInput := args[0]
+	err := validateScheme(userInput)
```
wagoodman, line 145: this validation only partially covers all of the source type cases. `generateSBOM` can still create SBOMs for sources that are files or dirs:
```
syft attest .
```

We can export `detectScheme` so that something before `generateSBOM` is called to help with this (https://github.com/anchore/syft/blob/5debdcbc9d2cffda6d15f11ae2437a163bdf0e78/syft/source/source.go#L41)

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+	if err != nil {
+		return err
+	}
+
+	ko := sign.KeyOpts{
+		KeyRef:   keyPath,
+		PassFunc: passFunc,
```
wagoodman, line 152: I think we can modify this to not require a password for passwordless keys (don't have the details this second / happy to pair on it)

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+	rootCmd.AddCommand(attestCmd)
+}
+
+func setAttestFlags(flags *pflag.FlagSet) {
+	// Key options
+	flags.StringVarP(&keyPath, "key", "", "cosign.key",
+		"private key to use to sign attestation",
+	)
+
+	flags.StringArrayVarP(&attestationOutput,
```
wagoodman, line 263: question for later: how difficult would it be to allow for multiple `-o` options being accepted like with the root command.

cmd/packages.go
```diff
@@ -236,54 +235,66 @@ func isVerbose() (result bool) {
...
+	return s
+}
+
 func packagesExecWorker(userInput string, writer sbom.Writer) <-chan error {
 	errs := make(chan error)
 	go func() {
 		defer close(errs)
-
-		tasks, err := tasks()
+		s, src, err := generateSBOM(userInput, errs)
```
wagoodman, line 283: should have a `nil` check on `s` to ensure that an SBOM is provided or else error out (possible panic here)

test/cli/attest_cmd_test.go
```diff
@@ -0,0 +1,50 @@
+package cli
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestAttestCmdFlags(t *testing.T) {
```
wagoodman, line 8: a great addition to this test would be to call out to `cosign` and verify the output with the [flow you added in the readme](https://github.com/anchore/syft#adding-an-sbom-to-an-image-as-an-attestation)

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+			errs <- err
+			return
+		}
+	}()
+	return errs
+}
+
+func assertPredicateType(output format.Option) string {
+	switch output {
+	case format.SPDXTagValueOption, format.SPDXJSONOption:
```
wagoodman, line 208: as we add more spdx and cyclonedx formats and versions we should probably use helper functions to select from a given format to protect against brittleness (something like `formats.IsSPDX(output)` which would return a bool).

What do you think?

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+		return in_toto.PredicateSPDX
```
wagoodman, line 209: question: does the predicate additionally encode the format of the payload? (e.g. tag-value vs json) Or is that hinted at somewhere else in the data structure?

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+	default:
+		return in_toto.StatementInTotoV01
+	}
+}
+
+func generateAttestation(predicate []byte, src *source.Source, sv *sign.SignerVerifier, output format.Option) error {
+	predicateType := assertPredicateType(output)
+
+	h, _ := v1.NewHash(src.Image.Metadata.ManifestDigest)
```
wagoodman, line 218: probably shouldn't ignore this error

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+
+	wrapped := dsse.WrapSigner(sv, "application/vnd.in-toto+json")
```
wagoodman, line 220: nit: could be a good constant (then the name would reflect what this represents)

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	payload, err := json.Marshal(sh)
+	if err != nil {
+		return err
+	}
+
+	signedPayload, err := wrapped.SignMessage(bytes.NewReader(payload), signatureoptions.WithContext(context.Background()))
+	if err != nil {
+		return errors.Wrap(err, "signing")
```
wagoodman, line 238: nit: 
```suggestion
		return errors.Wrap(err, "unable to sign SBOM")
```

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+
+func init() {
+	setAttestFlags(attestCmd.Flags())
+	rootCmd.AddCommand(attestCmd)
+}
+
+func setAttestFlags(flags *pflag.FlagSet) {
+	// Key options
+	flags.StringVarP(&keyPath, "key", "", "cosign.key",
+		"private key to use to sign attestation",
```
wagoodman, line 260:
```suggestion
		"path to the private key file to use for attestation",
```

cmd/attest.go
```diff
@@ -0,0 +1,267 @@
...
+	)
+
+	flags.StringArrayVarP(&attestationOutput,
+		"output", "o", []string{string(format.JSONOption)},
+		fmt.Sprintf("attestation output format, options=%v", format.AllOptions),
```
wagoodman, line 265: nit: the phrasing makes it seem like this controls the attestation format instead of whats within the attestation, maybe something like
```suggestion
		fmt.Sprintf("SBOM output format, options=%v", format.AllOptions),
```

cmd/packages.go
```diff
@@ -236,54 +235,66 @@ func isVerbose() (result bool) {
...
+			Configuration: appConfig,
+		},
+	}
+
+	s = buildRelationships(s, src, tasks, errs)
+
+	return &s, src, nil
+}
+
+func buildRelationships(s sbom.SBOM, src *source.Source, tasks []task, errs chan error) sbom.SBOM {
```
wagoodman, line 266: instead of passing a copy of `sbom.SBOM` and returning `sbom.SBOM` it looks like this is a good candidate for passing a reference to the sbom and returning nothing

wagoodman:
nit: probably makes sense to add `cosign.key` and `cosign.pub` to the `.gitignore`

wagoodman:
re: binary size. Increasing by 3x does seem a little excessive. Tagging onto our earlier conversation about upx... _I think either a light-take on upx or doing nothing is the right call here_.

We don't really need expensive options such as `-9` or `--brute` to get the filesize to a "reasonable" size. This tilts the time-space tradeoff back towards the save-some-time side so that our snapshot builds don't take that long.

```
$ ls -alh ./syft
-rwxr-xr-x  1 -  -    66M Feb 14 17:20 ./syft
                                                                                                                                 
$ time upx -1 ./syft
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2020
UPX 3.96        Markus Oberhumer, Laszlo Molnar & John Reiser   Jan 23rd 2020

        File size         Ratio      Format      Name
   --------------------   ------   -----------   -----------
  69166256 ->  23650320   34.19%   macho/amd64   syft                          

Packed 1 file.
upx -1 ./syft  2.15s user 0.09s system 98% cpu 2.282 total
                                                                                                                                     
$ ls -alh ./syft    
-rwxr-xr-x  1 -  -    23M Feb 14 17:20 ./syft

```

For +~3 seconds processing per-binary, that size is just about as good as today:
```
$ syft version                     
Application:     syft
Version:         0.37.10
...
                                                                                                                           
$ ls -alh $(realpath $(which syft))
-r-xr-xr-x  1 -  -    22M Feb  7 19:05 /usr/local/Cellar/syft/0.37.10/bin/syft

```

That's a pretty good time-space trade off and I think will hold us over for now.

UPX works by adding a just-in-time decompression of part of the binary, which means a small time cost is taken for every invocation of syft. How much time?

```
# today
$ time SYFT_CHECK_FOR_APP_UPDATE=false syft version
Application:     syft
Version:         0.37.10
...
SYFT_CHECK_FOR_APP_UPDATE=false syft version  0.02s user 0.01s system 118% cpu 0.025 total


# with upx
$ time SYFT_CHECK_FOR_APP_UPDATE=false ./syft version
Application:     syft
Version:         0.37.10-SNAPSHOT-ba41a59
...
SYFT_CHECK_FOR_APP_UPDATE=false ./syft version  0.28s user 0.04s system 103% cpu 0.303 total

```

This seems to add about ~0.3 seconds to the syft startup time (on my box). 

...**Conclusion: `upx -1` seems like the right call here if we're convinced that 3x larger binary is not desirable... otherwise the right call is to do nothing.**

Here's a commit that prototypes out the changes --feel free to pull it onto this branch: https://github.com/anchore/syft/commit/ba41a591bf332fe9763650b68bc4dcdeae5e48f9

cc: @anchore/tools

cmd/attest.go
spiffcs: I'll modify this for the next PR when we tackle the passwordless! I've pinned this comment for the new issue.

cmd/attest.go
spiffcs: I don't think it would be too difficult. I can file this as a separate issue so we are sure to cover multiple output

cmd/packages.go
spiffcs: Checking now for nill sbom no panic

cmd/attest.go
spiffcs: Error checked

cmd/attest.go
spiffcs: Updated this to be better error descriptor

cmd/attest.go
spiffcs: updated phrasing

cmd/packages.go
spiffcs: updated to pass by pointer with no return

spiffcs:
Finishing the test updates described.

cmd/attest.go
spiffcs: Since we're only doing SPDXJSON, CycloneDXJSON, and SYFTJSON for this predicate type we'll modify this and keep this function around after talking about it over zoom.

README.md
```diff
@@ -68,6 +69,7 @@ nix-shell -p syft
 
 ## Getting started
 
+#### SBOM
```
spiffcs, line 72: Remove titles no segregation here.

README.md
```diff
@@ -85,6 +87,17 @@ To include software from all image layers in the SBOM, regardless of its presenc
 syft packages <image> --scope all-layers
 ```
 
+#### Attestation
```
spiffcs, line 90: Update possibly to SBOM Attestation

wagoodman (approved):
Great work --this is a huge feature 🎉 Can't wait for the keyless work next 🤓

Issue #733, opened 2022-01-04
unable to catalog dpkg package=/var/lib/dpkg/status


foolm00n:
**What happened**:
1 error occurred:
	* unable to catalog dpkg package=/var/lib/dpkg/status: bad installed-size value="2400MB": strconv.Atoi: parsing "2400MB": invalid syntax

**What you expected to happen**:
Didn't expect the error. The size of /var/lib/dpkg/status is actually 2.4MB not "2400MB"
It runs without errors on another machine with very similar configuration.

**How to reproduce it (as minimally and precisely as possible)**:
As root, run syft dir:/ -v -o json --file sbom.json

**Anything else we need to know?**:
Without verbose mode, syft gets stuck and hangs at "Cataloging packages [packages: 0]" stage

**Environment**:
- Output of `syft version`:
```
Application:   syft
Version:       0.34.0
BuildDate:     2021-12-22T21:15:39Z
GitCommit:     7f8cb0bd8068709982ef23e931e1ceaa2dfac955
GitTreeState:  clean
Platform:      linux/amd64
GoVersion:     go1.16.12
Compiler:      gc
```

- OS (e.g: `cat /etc/os-release` or similar):
```
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

jonasagx:
Thanks for posting this bug @foolm00n
There seem to be two problems:

1. dpkg meta data
It looks like your `/var/lib/dpkg/status` has a package where the installed-size field is not just an integer, so Syft should accept file sizes with scale suffixes, e.g. 1MB. I can't reproduce it on my debian boxes since `installed-size` probably comes from a dpkg application meta-data. Would you mind sharing which package has the `installed-size` with `MB` or any other value that is not simply an integer?

2. hanging problem
We are investigating cases where it happens. Also keep in mind that scanning all your file system might take awhile.

foolm00n:
[...]

PR #786, opened 2022-02-01
ignore minor parsing error when reading dpkg status files

jonasagx:
Fixes https://github.com/anchore/syft/issues/733

Question: should we add a smarter parser to guess approximate installed-size
value?
A: yes, this PR uses go-humanize to parse storage metadata better

This PR also drops dpkg cataloger from returning parsing errors, they should not stop Syft. Warning with context were added.


syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -6,12 +6,12 @@ import (
...
 	"io"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/anchore/syft/internal"
-
+	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/pkg"
+	"github.com/jonasagx/datasize"
```
wagoodman, line 14: looks like this should be `github.com/c2h5oh/datasize`

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -164,11 +164,12 @@ func handleNewKeyValue(line string) (string, interface{}, error) {
 		// further processing of values based on the key that was discovered
 		switch key {
 		case "InstalledSize":
-			numVal, err := strconv.Atoi(val)
+			var s datasize.ByteSize
+			err := s.UnmarshalText([]byte(val))
 			if err != nil {
-				return "", nil, fmt.Errorf("bad installed-size value=%q: %w", val, err)
+				log.Warnf("bad installed-size value=%q: %w", val, err)
```
wagoodman, line 170: should probably return early here after warning (`return key, nil, nil`)

syft/pkg/cataloger/deb/parse_dpkg_status.go
wagoodman: we use `github.com/dustin/go-humanize` for some of these cases, would that work instead of pulling in a new lib?

jonasagx: it did work, thanks

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -154,7 +150,7 @@ func extractSourceVersion(source string) (string, string) {
 }
 
 // handleNewKeyValue parse a new key-value pair from the given unprocessed line
-func handleNewKeyValue(line string) (string, interface{}, error) {
+func handleNewKeyValue(line string) (key string, val interface{}, err error) {
```
kzantow, line 153: `key` is redefined below, this might be confusing?

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -59,9 +58,6 @@ func parseDpkgStatusEntry(reader *bufio.Reader) (pkg.DpkgMetadata, error) {
 	var retErr error
 	dpkgFields, err := extractAllFields(reader)
 	if err != nil {
-		if !errors.Is(err, errEndOfPackages) {
-			return pkg.DpkgMetadata{}, err
-		}
```
kzantow, line 64: looking below at `extractAllFields`, it appears to also only return fields,nil or nil,err; so continuing processing here also would be wrong, no?

syft/pkg/cataloger/deb/parse_dpkg_status_test.go
```diff
@@ -242,3 +259,153 @@ func TestSourceVersionExtract(t *testing.T) {
...
+			wantErr: assert.NoError,
+		},
+		{
+			name: "duplicated key",
+			input: bufio.NewReader(strings.NewReader(`
+Package: apt
+Package: apt-get
+
+`)),
+			wantErr: func(t assert.TestingT, err error, i ...interface{}) bool {
```
kzantow, line 282: Might be slightly easier to understand by just having an error message (like `expectedError: "duplicate key discovered: Package"`) instead of a callback here, then if `expectedError != ""` in the actual test below, do this?

syft/pkg/cataloger/deb/parse_dpkg_status_test.go
```diff
@@ -242,3 +259,153 @@ func TestSourceVersionExtract(t *testing.T) {
...
+				expected := errors.New("duplicate key discovered: Package")
+				return errors.Is(err, expected)
+			},
+		},
+		{
+			name: "no match for continuation",
+			input: bufio.NewReader(strings.NewReader(`  Package: apt
```
kzantow, line 289: Maybe just include the string here and then do `bufio.NewReader(strings.NewReader(` in the test below?

kzantow (commented):
A couple suggestions I think might help maintainability and unclear about a couple error handling path changes. The main bit of the `go-humanize` addition looks great, though!

syft/pkg/cataloger/deb/cataloger.go
```diff
@@ -53,7 +53,7 @@ func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, []arti
 		pkgs, err := parseDpkgStatus(dbContents)
 		internal.CloseAndLogError(dbContents, dbLocation.VirtualPath)
 		if err != nil {
-			return nil, nil, fmt.Errorf("unable to catalog dpkg package=%+v: %w", dbLocation.RealPath, err)
+			log.Warnf("dpkg cataloger: unable to catalog package=%+v: %w", dbLocation.RealPath, err)
```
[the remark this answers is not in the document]
jonasagx: `continue` in this situation would be redundant, because if `pkgs` is empty then iterating over it will yield nothing. But I this redundancy would facilitate reading. I will add it

syft/pkg/cataloger/deb/parse_dpkg_status.go
jonasagx: 👍🏼 I removed the `var` declaration from the function body

wagoodman (commented):
This comment: https://github.com/anchore/syft/pull/786#discussion_r799693381 was resolved but it doesn't look like there's a corresponding change or additional conversation here.

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -164,11 +161,11 @@ func handleNewKeyValue(line string) (string, interface{}, error) {
 		// further processing of values based on the key that was discovered
 		switch key {
 		case "InstalledSize":
-			numVal, err := strconv.Atoi(val)
+			s, err := humanize.ParseBytes(val)
 			if err != nil {
-				return "", nil, fmt.Errorf("bad installed-size value=%q: %w", val, err)
+				log.Warnf("extracting dpkg fields: bad installed-size value=%q: %w", val, err)
```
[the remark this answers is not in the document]
jonasagx: @wagoodman I moved the warning log one caller up https://github.com/jonasagx/syft/blob/733-fix-dpkg-parser/syft/pkg/cataloger/deb/cataloger.go#L56 where we can expose more error messages, and not break the deb cataloging

syft/pkg/cataloger/deb/parse_dpkg_status_test.go
jonasagx: I refactored the test so we still use tests but in a simpler manner, hopefully more readable this time. 
I would prefer to stick to error var, they are a simple but rich API that supports error encapsulation, and checking -- which we can't do as easily with pure strings.

syft/pkg/cataloger/deb/cataloger_test.go
```diff
@@ -12,7 +12,6 @@ import (
 )
 
 func TestDpkgCataloger(t *testing.T) {
-
```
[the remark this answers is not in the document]
jonasagx: nah, just don't like empty lines right after function signatures

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -133,7 +133,8 @@ func extractAllFields(reader *bufio.Reader) (map[string]interface{}, error) {
 			var val interface{}
 			key, val, err = handleNewKeyValue(line)
 			if err != nil {
-				return nil, err
+				log.Warnf("parsing key from line: %s err: %v", line, err)
```
wagoodman, line 136: Same comment from earlier, as an error with chaining there is enough context to glean where the message is coming from. As a log warning there isn't enough context here to tell it's a dpkg status failure when appearing in the terminal.

syft/pkg/cataloger/deb/parse_dpkg_status_test.go
```diff
@@ -260,9 +260,9 @@ func TestSourceVersionExtract(t *testing.T) {
 	}
 }
 
-func errIs(expected error) assert.ErrorAssertionFunc {
+func assertEqualErr(expected error) assert.ErrorAssertionFunc {
 	return func(t assert.TestingT, err error, i ...interface{}) bool {
-		return errors.Is(err, expected)
+		return assert.Contains(t, err.Error(), expected.Error())
```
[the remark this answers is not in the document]
jonasagx: <del>I noticed that `errors.Is` wasn't enough to compare if an error includes another one, and `assert.Equal` as too struct, so I went with a poor-man's error inclusion assert, where it is all just struct comparison.</del>

Actually I should just use `error.As` 🤦🏼 
Thanks for pointing this out.

syft/pkg/cataloger/deb/parse_dpkg_status.go
jonasagx: The log line should have more context info now

syft/pkg/cataloger/deb/parse_dpkg_status_test.go
```diff
@@ -20,11 +25,13 @@ func compareEntries(t *testing.T, left, right pkg.DpkgMetadata) {
...
-		name     string
-		expected pkg.DpkgMetadata
+		name        string
+		expected    pkg.DpkgMetadata
+		fixturePath string
 	}{
 		{
-			name: "Test Single Package",
+			name:        "Test Single Package",
+			fixturePath: filepath.Join("test-fixtures", "status", "single"),
```
kzantow, line 34: FWIW all OSes can handle `test-fixtures/status/single`, including Windows

syft/pkg/cataloger/deb/parse_dpkg_status_test.go
```diff
@@ -242,3 +259,135 @@ func TestSourceVersionExtract(t *testing.T) {
...
+			name:    "fail parsing installed-size",
+			line:    "Installed-Size: 1bla",
+			wantKey: "",
+			wantErr: assertAs(fmt.Errorf("unhandled size name: %s", "bla")),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotKey, gotVal, err := handleNewKeyValue(tt.line)
+			tt.wantErr(t, err, fmt.Sprintf("handleNewKeyValue(%v)", tt.line))
```
kzantow, line 387: I would probably have `wantErr string` and then something like:
```
if tt.wantErr != "" {
  assertAs(fmt.Error(tt.wantErr))(t, err)
} else {
  assert.NoError(t, err)
}
```
... so each test case is made simpler

kzantow (approved):
LGTM; just a couple suggestions that I wouldn't hold this up for

syft/pkg/cataloger/deb/parse_dpkg_status.go
```diff
@@ -154,21 +155,21 @@ func extractSourceVersion(source string) (string, string) {
...
+		key = strings.TrimSpace(line[0:i])
 		// mapstruct cant handle "-"
 		key = strings.ReplaceAll(key, "-", "")
 		val := strings.TrimSpace(line[i+1:])
 
 		// further processing of values based on the key that was discovered
 		switch key {
 		case "InstalledSize":
-			numVal, err := strconv.Atoi(val)
+			s, err := humanize.ParseBytes(val)
```
spiffcs, line 168: What's the purpose of `humanize`?  
Are the `line` variables being parsed consistently formatted under `InstalledSize`?

Why are we returning an interface here? Do we want to coerce the value so this function always returns a consistent type?

spiffcs (commented):
small comment about how we are parsing the size variables

spiffcs (approved):

syft/pkg/cataloger/deb/parse_dpkg_status_test.go
jonasagx: Yes, I like the test running step be as simple as possible, where each test case has all the information for validation, so each case is just configuration.

PR #789, opened 2022-02-04
Normalize snapshot and release artifacts

wagoodman:
This PR refactors the release pipeline in the following ways:
1. Brings the snapshot artifacts in line with the release artifacts. In this way what we test against is what we will expect to see during the release process (see the `snapshot != release` caveat in https://github.com/anchore/syft/pull/765 ).
2. The signing path can be tested locally with self-signed certs using a new `make snapshot-with-signing` target.
3. Bumps the goreleaser version, which will prevent the deprecation warning for our brew installs.

Limitations:
1. Notarization cannot easily be tested locally
2. How the signing material is prepared is not the same locally vs in the release pipeline, thus, cannot be tested locally


.github/scripts/apple-signing/run.sh
```diff
@@ -0,0 +1,65 @@
...
+      ;;
+  esac
+
+  . "$SCRIPT_DIR"/sign.sh
+
+  # load up all signing material into a keychain (note: this should set the MAC_SIGNING_IDENTITY env var)
+  setup_signing
+
+  # sign all of the binaries in the archive and recreate the input archive with the signed binaries
+  sign_archive "$archive_abs_path" "$MAC_SIGNING_IDENTITY"
```
wagoodman, line 44: nit: change to `sign_binaries_in_archive`

.github/scripts/apple-signing/prep-signing-dev.sh
```diff
@@ -0,0 +1,163 @@
...
+            -out "$CERT_FILE" \
+            -extfile "$EXT_FILE" \
+            -passin "pass:$KEY_PASSWORD" \
+            -extensions $EXT_SECTION
+
+  commentary "verify the certificate: we should see our extensions"
+  openssl x509 -text -noout -in $CERT_FILE | grep -A1 'X509v3' || exit_with_error "could not find x509 extensions in certificate"
+
+  title "export cert and private key to .p12 file"
+  openssl pkcs12 \
```
wagoodman, line 111: nit: add a note that this was helpful to make, but a p12 may not actually be needed.

.github/scripts/apple-signing/prep-signing-dev.sh
```diff
@@ -0,0 +1,163 @@
...
+    # note: add-trusted-cert requires user interaction
+    commentary "adding the developer certificate as a trusted certificate... (requires user interaction)"
+    security add-trusted-cert -d -r trustRoot -k "$KEYCHAIN_PATH" "$CERT_FILE"
+  else
+    set -e
+    commentary "...dev cert has already been imported onto the dev keychain"
+  fi
+
+  # remove any generated cert material since the keychain now has all of this material loaded
+  rm -rf "${DIR}"
```
wagoodman, line 148: consider providing a clean up script to delete the keychain such that it is untrusted. Maybe have a env var that can persist this keychain for ease during development, but default it would always remove the keychain.

luhring (approved):
Huge improvement! We paired on this in a sync-up, and it looks merge-able pending the comments already noted 💯

.github/scripts/apple-signing/prep-signing-dev.sh
wagoodman: added `.github/scripts/apple-signing/cleanup.sh` which is invoked in the makefile in the `snapshot-with-signing` target.

PR #796, opened 2022-02-05
Stabilize CI behavior

wagoodman:
There are multiple instance of:
1. OOM during snapshot builds (https://github.com/anchore/syft/actions/runs/1799784721/attempts/2)
2. install.sh tests failing periodically due to http server not being ready (https://github.com/anchore/syft/runs/5078759447?check_suite_focus=true)

This PR ~reduces the build parallelism~ fixes the signing bash script logging to not hang in CI and increases the wait time after setting up the http server for install.sh testing (which ideally would be event driven and not a sleep --happy to take suggestions here).


wagoodman:
it looks like with these changes I'm still seeing hung runs in CI, which means that the parallelism changes didn't fix the underlying problem

Screen Shot 2022-02-05 at 12 10 11 PM

jonasagx (approved):

PR #797, opened 2022-02-06
Fix intermittent slow CLI tests

wagoodman:
CLI tests may run considerably slower (several minutes) on the registry auth tests when there is a non-registry service running on port 5000 (the same port that the registry tests reference). Turning off the offending service solves the problem, however, 5000 tends to be a commonly used port. Swapping to a port that is (probably) not in use speeds up the tests from minutes to seconds (~2 seconds). The test now references the extremely useful [QOTD port](https://en.wikipedia.org/wiki/QOTD) in the hopes to glean occasional wisdom.

### DIFF
--- a/test/cli/packages_cmd_test.go
+++ b/test/cli/packages_cmd_test.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


test/cli/packages_cmd_test.go
```diff
@@ -209,55 +209,55 @@ func TestRegistryAuth(t *testing.T) {
 	}{
 		{
 			name: "fallback to keychain",
-			args: []string{"packages", "-vv", "registry:localhost:5000/something:latest"},
+			args: []string{"packages", "-vv", "registry:localhost:17/something:latest"},
```
luhring, line 212: nit: Since we have a handful of test cases all using the same value in setups and assertions, it might be a good idea to extract out the common parts as constants (scoped to the `TestRegistryAuth` func), such as `localhost:17` and/or `localhost:17/something:latest`.

luhring (approved):
>The test now references the extremely useful [QOTD port](https://en.wikipedia.org/wiki/QOTD) in the hopes to glean occasional wisdom.

Lol 😄

PR #805, opened 2022-02-07
use SYFT_LOG_FILE env var

jonasagx:
[README](https://github.com/anchore/syft/blob/main/README.md#L391) mentions SYFT_LOG_FILE but is not functional. This PR reads SYFT_LOG_FILE during log initialization and sets log level to debug.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -13,7 +13,6 @@ import (
 	"github.com/anchore/syft/internal/version"
 	"github.com/anchore/syft/syft"
 	"github.com/gookit/color"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"github.com/wagoodman/go-partybus"
@@ -89,11 +88,6 @@ func initAppConfig() {
 }
 
 func initLogging() {
-	if logFile, ok := os.LookupEnv("SYFT_LOG_FILE"); ok && logFile != "" {
-		appConfig.Log.FileLocation = logFile
-		appConfig.Log.LevelOpt = logrus.DebugLevel
-	}
-
 	cfg := logger.LogrusConfig{
 		EnableConsole: (appConfig.Log.FileLocation == "" || appConfig.CliOptions.Verbosity > 0) && !appConfig.Quiet,
 		EnableFile:    appConfig.Log.FileLocation != "",
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -13,6 +13,7 @@ import (
 	"github.com/anchore/syft/internal/version"
 	"github.com/anchore/syft/syft"
 	"github.com/gookit/color"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"github.com/wagoodman/go-partybus"
@@ -90,6 +91,7 @@ func initAppConfig() {
 func initLogging() {
 	if logFile, ok := os.LookupEnv("SYFT_LOG_FILE"); ok && logFile != "" {
 		appConfig.Log.FileLocation = logFile
+		appConfig.Log.LevelOpt = logrus.DebugLevel
 	}
 
 	cfg := logger.LogrusConfig{
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -88,6 +88,10 @@ func initAppConfig() {
 }
 
 func initLogging() {
+	if logFile, ok := os.LookupEnv("SYFT_LOG_FILE"); ok && logFile != "" {
+		appConfig.Log.FileLocation = logFile
+	}
+
 	cfg := logger.LogrusConfig{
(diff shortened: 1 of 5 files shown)


luhring:
Adding "bug" label to this PR since we don't have a corresponding issue, and we want to keep track of this as a bug for release notes and future reference. 😃 

(I've reproduced this bug on my machine.)

cmd/cmd.go
```diff
@@ -88,6 +89,11 @@ func initAppConfig() {
 }
 
 func initLogging() {
+	if logFile, ok := os.LookupEnv("SYFT_LOG_FILE"); ok && logFile != "" {
+		appConfig.Log.FileLocation = logFile
+		appConfig.Log.LevelOpt = logrus.DebugLevel
+	}
+
```
luhring, line 96: Hmm, in theory, we shouldn't need to do this...

When you get a chance, it's worth a study of how application config works in Syft (and Grype)! It's pretty cool. We use [Viper](https://github.com/spf13/viper) with [Cobra](https://github.com/spf13/cobra) to read in config values from multiple sources in priority order (command line flags, environment variables, and config files).

We should be able to count on the [`Application.Log.FileLocation`](https://github.com/anchore/syft/blob/main/internal/config/logging.go#L13) config value being set correctly as we initialize Cobra and Viper, and it looks like that's not happening for some reason. 🤔  That'd be a good place to begin the hunt for the root cause of this bug.

kzantow: Hah, I was just thinking this. I would note that sometimes "magic" like this makes it harder to track down where problems are because there's nothing in the codebase to search for.

wagoodman: most of this magic is from leveraging viper and cobra, which is a good thing (stand on the shoulder of giants)! This does mean getting acquainted with cobra and viper and how they interact. Cobra gives us CLI parsing, Viper gives us config parsing, and together they provide a 12-factor-like experience when it comes to these domains (Env vars override config values, CLI switches override config and env vars).

The definitions for these elements come in three parts:
1. define the CLI switch [`-s`](https://github.com/anchore/syft/blob/v0.37.10/cmd/packages.go#L92-L94)
2. define the config element `package.cataloger.scope` [from here](https://github.com/anchore/syft/blob/main/internal/config/pkg.go#L9) to [ultimately here](https://github.com/anchore/syft/blob/main/internal/config/cataloger_options.go#L13)
3. bind `-s` and `package.cataloger.scope` together [here](https://github.com/anchore/syft/blob/v0.37.10/cmd/packages.go#L146-L148)

With this we can specify the scope from the CLI:
```
$ syft -s <scope> myimage:latest
```

... from an environment variable (based off of the config path)
```
$ SYFT_PACKAGE_CATALOGER_SCOPE=scope syft myimage:latest 
```

or in a config:
```yaml
# .syft.yaml
package:
   cataloger:
       scope: <scope>
```
```
$ syft -c .syft.yaml
```

With a couple of exceptions, all input options must be defined in the application config. This is not true of the CLI options: not all possible input is expressible via the CLI. 

The next part that matters (step 4, really) is where does the default value come from? This can be different depending on how everything is wired up... is there a CLI switch for the config value? If yes, then look for the cobra configured default on the CLI flag object, viper config binding uses that value as the default for the application config. If no, then the default must come from somewhere for binding to work. This is where the problem for the log file configurable is occurring. There is some subtle behavior in how cobra and viper work with the config binding feature that I think is causing the problem. Specifically, there is no default value for `SYFT_LOG_FILE`/`log.file` so the binding never takes effect.

That is, when looking up the config flag for `package.cataloger.scope` if there is no default value then I think the binding doesn't work as expected.

We specify config-only defaults (options without CLI switches) within the `loadDefaultValues` method on the config struct: https://github.com/anchore/syft/blob/main/internal/config/logging.go#L17

I think setting the default for `log.file` is missing here.

kzantow: Maybe just tests for everything we expect as input?

wagoodman: tests are always good, and we do have tests that cover a small portion of config bindings, but these kinds of tests are brittle to the expansion of options over time. Without mechanisms that cause test failures when detecting uncovered cases it's easy to introduce these kinds of bad behaviors over time.

My comment was trying to bring forth ideas to protect against this class of bug --this is not the first time that we've had this problem, which is why we put in CLI-level tests for this (which doesn't scale well).

One thought I have is to write a helper for the binding operation, which can be used to perform the binding but also double check that the intended effect was had or else return an error. Another thought is to somehow provide zero-valued defaults for all values in our config struct (maybe with reflection?). There may be a mechanism that already exists in viper that protects against this that I'm not aware of.

jonasagx: Thank you for the info. Setting "log.file" in https://github.com/anchore/syft/blob/main/internal/config/logging.go#L17 fixed the issue -- but now a log file is always created, even when `SYFT_LOG_FILE` is _not_ set. I am checking how to make it dependent on the env var.

I agree with Keith, it deserves an test (probably an integration one)

wagoodman (approved):

PR #811, opened 2022-02-10
Add CycloneDX decoder

kzantow:
This PR adds support for decoding CycloneDX XML and JSON formats 🎉 

This is a prerequisite for https://github.com/anchore/grype/issues/481


syft/pkg/cataloger/rust/cargo_metadata.go
```diff
@@ -1,11 +1,19 @@
...
+	"reflect"
+
+	"github.com/anchore/syft/syft/pkg"
+)
 
 type CargoMetadata struct {
 	Packages []pkg.CargoPackageMetadata `toml:"package"`
 }
 
+func init() {
```
kzantow, line 13: is this `init()` function the right place to do something like this?

internal/formats/common/cyclonedxhelpers/cpe.go
```diff
@@ -1,12 +1,30 @@
...
 	// Since the CPEs in a package are sorted by specificity
 	// we can extract the first CPE as the one to output in cyclonedx
 	if len(p.CPEs) > 0 {
 		return pkg.CPEString(p.CPEs[0])
 	}
 	return ""
 }
+
+func decodeCPEs(c *cyclonedx.Component) []pkg.CPE {
+	// FIXME we not encoding all the CPEs (see above), so here we just use the single provided one
```
kzantow, line 19: This is something we've decided to do as part of this PR, this is a known deficiency that needs to be addressed at some point soon. It probably can have CPEs generated automatically or something similar, like is in the Grype PR here: https://github.com/anchore/grype/pull/606

internal/formats/common/cyclonedxhelpers/component.go
```diff
@@ -1,27 +1,164 @@
...
+	}
+
+	return p
+}
+
+func decodeLocations(c *cyclonedx.Component) (out []source.Location) {
+	if c.Properties != nil {
+		props := *c.Properties
+		for i := 0; i < len(props)-1; i++ {
+			if props[i].Name == "path" && props[i+1].Name == "layerID" {
```
kzantow, line 66: This is a pretty janky thing: we have multiple props with the same name -- each location struct is output with path=.. layerID=..; we should figure out if there is a better way to accomplish this

samj1912: I have been thinking about this from when I implemented this. What if we added the index in the property name, so path.0, layerId.0 and so on for any array like elements? Or maybe we can even utilize the syntax from gron https://github.com/tomnomnom/gron#usage ?

internal/formats/common/cyclonedxhelpers/cpe.go
samj1912: @kzantow this might help -> https://github.com/anchore/syft/pull/819

syft/pkg/cataloger/rust/cargo_metadata.go
kzantow: Spoke with @luhring and it seems like we should probably move `cargo_metadata.go` up to `syft/pkg`; I'll make this change.

wagoodman: why so? This cargo metadata is meant for parsing a cargo metadata file and it's `pkg.CargoPackageMetadata` that resides on the `pkg.Package.Metadata` itself

kzantow: @wagoodman this would be consistent with all the other package metadata types being defined in [syft/pkg](https://github.com/anchore/syft/tree/main/syft/pkg). But I'm happy to hear other options; maybe move the `pkg.CargoPackageMetadata` definition here?

wagoodman: I see the confusion: we're in agreement about the fact that we shouldn't be using `rust.CargoMetadata{}` here. What I was confused at is that there is already cargo metadata up in `pkg` (https://github.com/anchore/syft/blob/main/syft/pkg/cargo_package_metadata.go#L10), it looks like this was left behind mistakenly. We should indeed switch to using the metadata that is up in pkg (as was originally intended).

wagoodman (commented):
The original intent of the formats package was to have the following structure:
```
├── common
│   ├── spdxhelpers/     # common spdx functions that to not reference model packages (only uses strings, ints, etc. ... primitives) 
│   └── ...
└── spdx22json
    ├── decoder.go
    ├── encoder.go
    ├── format.go         # wire up all functions to return a format object
    ├── model/
    │   └──    ...         # the "format model" contains only struct definitions, no behavior
    ├── to_format_model.go    # behavior to translate from syft sbom.SBOM to the format model  
    ├── to_syft_model.go      # behavior to translate from the format model to syft sbom.SBOM 
    └── validator.go
```

This pattern has quickly eroded with the introduction of 3rd party libs that are shared across multiple formats. This isn't a bad thing since we're getting a lot of functionality and models from these libs.

From the SPDX and CycloneDX work it seems like we're moving towards (using spdx randomly as an example):
```
├── common
│   └── spdxhelpers/     # do all of the model translation work for all spdx formats
│         ├── (all of the helpers)
│         ├── to_syft_model.go
│         └── to_format_model.go 
└── spdx22json
    ├── decoder.go    
    ├── encoder.go     
    ├── format.go         # wire up all functions to return a format object
    └── validator.go
```

I realize this PR isn't the first to start this, but I still wanted to point out the direction change. 

This PR is drifting in a slightly different direction (again, spdx choice here is arbitrary):
```
├── common
│   └── spdxhelpers/     # do all of the model translation work for all spdx formats
│         ├── (all of the helpers)
│         ├── decoder.go    
│         ├── encoder.go     
│         ├── to_syft_model.go
│         ├── to_format_model.go 
│         └── validator.go
└── spdx22json
    └── format.go         # wire up all functions to return a format object
```

Here the spdx helpers are also making format choices by putting the encoder/decoders into the helpers package.

My ask: we should be consistent across formats when possible regarding these kinds of conventions. I think we should keep the encoder/decoders/validators in the respective format package and not in the helper packages.

internal/formats/common/cyclonedxhelpers/decoder_test.go
```diff
@@ -0,0 +1,258 @@
...
+			sbom, err := toSyftModel(&test.input)
+			assert.NoError(t, err)
+
+		test:
+			for _, e := range test.expected {
+				if e.os != "" {
+					assert.Equal(t, e.os, sbom.Artifacts.LinuxDistribution.ID)
+					assert.Equal(t, e.ver, sbom.Artifacts.LinuxDistribution.VersionID)
+				}
+				if e.pkg != "" {
```
wagoodman, line 214: refactoring this to make it more readable would be helpful (e.g. flipping the conditions and using guard clauses instead of deep nesting)

kzantow: Can't use guards for most of these, because they are not terminal conditions; swapped the one that I could, though.

syft/pkg/cataloger/rust/cargo_metadata.go
kzantow: I think the right thing to do here is move the `cargo_metadata.go`, I've done this -- please let me know if I am misunderstanding

wagoodman:
From an earlier conversation about https://github.com/anchore/syft/pull/811#pullrequestreview-884746258 , let chat about this further in a followup PR (not blocking for this one).

wagoodman (approved):
🎉

PR #816, opened 2022-02-12
Add option to limit to a single filesystem (#674)

### LINKED ISSUE #674: An option to limit to a single filesystem (like -xdev)
An option to limit to a single filesystem would be great. To borrow from find's manpage:

    -xdev  Don't descend directories on other filesystems.

**Why is this needed**:
Scanning a directory like / ends up in /proc/, /dev/ and other places.
Sometimes that leads to a hang, like in #665.

ajvpot:
Resolves #674 (eventually).

POSIX standard defines `-xdev` operand in `find` as:
> The primary shall always evaluate as true; it shall cause find not to continue descending past directories that have a different device ID (st_dev, see the [stat()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/stat.html) function defined in the System Interfaces volume of POSIX.1-2017).

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html#tag_20_47_05

Work in progress, adds an option that causes the syft DirectoryResolver to not walk paths on devices with a different st_dev than the root.

TODO:

- [ ] CLI Opts
- [ ] Support allowing filesystems by type(?)
- [ ] Support only allowing not nodev filesystems from `/proc/filesystems`(???)
- [ ] Windows(?)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -101,7 +101,7 @@ require (
 	github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
-	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2 // indirect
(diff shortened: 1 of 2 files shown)


ajvpot:
I'm cleaning this up, will reopen when I have something more concrete.

Issue #737, opened 2022-01-06
Catalog discovered SBOMs


wagoodman:
**What would you like to be added**:
If a container image (or directory) being scanned contains an SBOM it would be interesting to consider taking additional actions based on detecting that it's an SBOM. These actions might be:
- merge the discovered SBOM with the current SBOM being created
- additionally output the discovered SBOM and create a relationship between the two
- summarize the SBOM in a subsection of the SBOM being created
- another option?

This would give syft a more intelligent avenue for SBOM data discovered other than reporting it as "another file" discovered within the scanned source.

**Why is this needed**:
This could play into https://github.com/anchore/syft/issues/31 , but that is not clear yet.

**Additional context**:
If a "merge" of the discovered SBOM and the current SBOM is selected, it is important to be transparent about which packages were not directly detected by a package cataloger and which were (and specifically, which SBOM file each package came from). Additionally, if the merge path is detected this has certain implications about the ID-system for packages (there may be overlap in IDs used, which can get complicated fast).

bureado:
`Built-Using` comes to mind as a readily available attribute that indicates a software component of interest that is unlikely to be present at the time of scanning a binary artifact and that might be in the user's best interest to see in the resulting SBOM, see https://github.com/anchore/syft/issues/572#issuecomment-1002187843. That's for "system" components. Clearly this has an even wider applicability for application components, e.g., a build system that outputted a binary and an SBOM for said binary and placed both in the image that `syft` will scan; in such a scenario it might be safer to assume that the user wants the resulting SBOM to collate the results. As discussed in #31 and other places, the big question is what are the input formats of those hints, because in general the build systems and tooling that output such SBOMs aren't doing so in SPDX or Cyclone (although that's rapidly changing and in general we should expect that to be the norm soon) I'm talking particularly about https://reproducible-builds.org/docs/recording/ and for example the output of tools such as TestifySec's witness or basically any CI/build system-attached recorders that might output supply chain-relevant information.

samj1912:
Dependent upon implementation of decoders for all the SBOM formats the syft supports. 

CycloneDX -> #811 
SPDX -> #738

tofay:
> a build system that outputted a binary and an SBOM for said binary and placed both in the image that syft will scan; in such a scenario it might be safer to assume that the user wants the resulting SBOM to collate the results

[...]

Issue #723, opened 2021-12-23
Support lossless conversion to any SBOM format


wagoodman:
_This idea came from @samj1912 on a recent community meeting as well as a previous prototype._

Today we've started adding the ability to encode and decode from any SBOM format. However, not all fields are translatable from every format to every other format. The syft-json format contains all possible information that we can express and converting to SPDX or CycloneDX is usually lossy. This means when we start to implement the `syft convert` command in the future, there will be paths that would lead to information loss which are not apparent to the user.

We could keep extra information in all SBOM formats encoded out-of-band of the spec in select fields. For instance, in cyclone DX we could preserve most unsupported package fields within the `properties` of the package object. For SPDX we could encode the original unsupported fields into a comments field with bas64 encoding. 

Something similar to this was done in the past as a prototype with SPDX json: https://github.com/anchore/syft/pull/578 We ended up not going that directing and keep as in-spec as possible.

samj1912:
Relevant to this is the cyclonedx's SPDX taxonomy > https://github.com/CycloneDX/cyclonedx-property-taxonomy/pull/7

Also related to #563

spiffcs:
https://github.com/anchore/syft/pull/964 was merged recently getting us a bit closer on this issue, but I can't confirm if it's lossless at this point.

Issue #563, opened 2021-10-16
Support format SBOM conversion


wagoodman:
Syft can output multiple SBOM formats, however, once you have that format you can't convert to another format. This could be most useful if you have a syftjson formatted document and you want to produce SPDX (which should be lossless). In these cases something like this would be nice:

```
syft my-image:latest -o json > original.json
syft convert original.json --to spdx > original.spdx
syft convert original.json --to cyclonedx > original-cyclonedx.xml
```

It's not clear what to do in cases where there is potentially lossy behavior (e.g. convert from cyclonedx to syftjson):
- Warn the user that this is a lossy conversion and continue
- Stop the conversation (possibly bypass with `--force` or similar option)
- Do nothing --this is an end user concern (I feel that this is not a good option)
- Possibly guarantee non-lossy behavior by including all additional (not in specification) fields into all document formats (there is a heavy pro, and several cons)

One question that comes to mind: do we want to restrict these conversions to only documents that syft created to begin with? Or be able to generically convert between formats for a document that was generated from another (non-syft) tool? (does thing bring on more complexity? or not?... if so, how much?)

hectorj2f:
I'd start converting Syft to any of those, as detailed in the example. Another option could be to support multiple output formats instead of a single output format as today. wdyt ?

luhring:
>Another option could be to support multiple output formats instead of a single output format as today. wdyt ?

This definitely makes sense — and I'd say this is a separate feature, which we happen to have captured in #325

samj1912:
Relevant to this is the cyclonedx's SPDX taxonomy > https://github.com/CycloneDX/cyclonedx-property-taxonomy/pull/7

wagoodman:
from refinement:
- This ties into the github dependency API work in progress, pending  https://github.com/anchore/syft/pull/836
- Should we be warning on a per-field aspect? if so, how would we do this?
- It may be that the cyclonedx and spdx encoders/decoders have gotten good enough where we don't need to warn on anything.
- We can go an encode-decode-encode pass to detect if we should output a warning (potential loss in data). This isn't perfect as data ordering may have caused this.
- Do we want to capture intent of the `convert` functionality? That is, this is a utility to convert SBOMs that are similar to syft output, not cover all features in all SBOM specs. Suggestion: we can add verbiage in the readme and help text around this being for "packages", thus does not consider VEX or similar components (so that information would be lost).
- Maybe we cover some of the caveats in a blog post to show the most useful use cases this covers.
[...]

PR #819, opened 2022-02-13
Add support for recording multiple CPEs in cyclonedx

### LINKED ISSUE #818: Add support for multiple CPEs in CycloneDX

Currently syft only records the first CPE in its outputs for CycloneDX. While looking at the CycloneDX spec and the SPDX spec, I remembered that similar to SPDX, cyclonedx supports adding external references of type "advisory" in a URI format. CPE, coincidentally can be stored either in a string format or a URI format. Syft already supports parsing URL formatted CPEs, but it doesn't currently export CPEs in URI format. If we combine these two properties, we should be able to save all the CPE information in the CycloneDX format as well while still storing the primary or what we assume is the canonical CPE in the CycloneDX CPE field which only allows for 1 CPE. 

**Why is this needed**:

While there can only be 1 canonical CPE for any package, it is impossible for scanners to accurately determine this. While syft puts the most specific CPE in the CycloneDX CPE field, there are times where CPEs have alternative vendors etc. that we may want to capture. For these use cases, we should allow syft to store all its generated CPEs in the output cyclonedx bom.

samj1912:
Fixes #818 

For reasoning and logic behind the CPE serialization decision please take a look at the issue. We still store our most specific CPE in the cyclonedx CPE field but for all the other CPEs, we store them as external references.

### DIFF
--- a/syft/pkg/cpe.go
+++ b/syft/pkg/cpe.go
@@ -99,6 +99,22 @@ func CPEString(c CPE) string {
 	return output.BindToFmtString()
 }
 
+func CPEURI(c CPE) string {
+	output := CPE{}
+	output.Vendor = sanitize(c.Vendor)
+	output.Product = sanitize(c.Product)
+	output.Language = sanitize(c.Language)
+	output.Version = sanitize(c.Version)
+	output.TargetSW = sanitize(c.TargetSW)
+	output.Part = sanitize(c.Part)
+	output.Edition = sanitize(c.Edition)
+	output.Other = sanitize(c.Other)
+	output.SWEdition = sanitize(c.SWEdition)
+	output.TargetHW = sanitize(c.TargetHW)
+	output.Update = sanitize(c.Update)
+	return output.BindToURI()
+}
+
 // sanitize is a modified version of WFNize function from nvdtools
 // that quotes all the allowed punctation chars with a slash and replaces
 // spaces with underscores. It differs from the upstream implmentation as
--- a/syft/pkg/cpe_test.go
+++ b/syft/pkg/cpe_test.go
@@ -50,6 +50,10 @@ func TestNewCPE(t *testing.T) {
 				t.Errorf("mismatched entries:\n\texpected:%+v\n\t  actual:%+v\n", CPEString(test.expected), CPEString(actual))
 			}
 
+			if CPEURI(test.expected) != test.input {
+				t.Errorf("mismatched entries:\n\texpected:%+v\n\t  actual:%+v\n", CPEURI(test.expected), test.input)
+			}
+
 		})
 	}
 }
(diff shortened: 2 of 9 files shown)


syft/pkg/cpe_test.go
```diff
@@ -105,6 +109,9 @@ func Test_CPEParser(t *testing.T) {
 			assert.Equal(t, c1, test.WFN)
 			assert.Equal(t, c2, test.WFN)
 			assert.Equal(t, CPEString(test.WFN), test.CPEString)
+			assert.Equal(t, CPEURI(c2), test.CPEUrl)
+			assert.Equal(t, CPEURI(c1), test.CPEUrl)
+			assert.Equal(t, must(NewCPE(CPEURI(c1))), test.WFN)
```
samj1912, line 114: This ensures that we are also making our URL generation also go through the same rigorous CPE encoding/decoding test that our CPE strings go through.

syft/pkg/cpe.go
```diff
@@ -99,6 +99,22 @@ func CPEString(c CPE) string {
...
+	output.Vendor = sanitize(c.Vendor)
+	output.Product = sanitize(c.Product)
+	output.Language = sanitize(c.Language)
+	output.Version = sanitize(c.Version)
+	output.TargetSW = sanitize(c.TargetSW)
+	output.Part = sanitize(c.Part)
+	output.Edition = sanitize(c.Edition)
+	output.Other = sanitize(c.Other)
+	output.SWEdition = sanitize(c.SWEdition)
+	output.TargetHW = sanitize(c.TargetHW)
```
samj1912, line 113: This mimics our custom CPEString logic above to handle special characters in the CPE.

samj1912:
cc: @coderpatros & @stevespringett to validate if this seems like a reasonable usage of the cyclonedx spec. to get around accurate CPE determination limitation that scanning tools face. Detailed reasoning at https://github.com/anchore/syft/issues/818

stevespringett:
I think using properties is more appropriate. External references in CycloneDX are typically URIs to other resources. This varies a lot from SPDX which applies the external reference name to mean potential alternative identifiers.

We have a formal property taxonomy as well. https://github.com/CycloneDX/cyclonedx-property-taxonomy

We could register `anchore` as a top-level namespace if desired?

samj1912:
Thanks for the feedback, I have opened a PR to register a `syft` namespace since we are also using cyclonedx properties for storing other fields. The PR should first be reviewed by syft maintainers before any further action is taken -> https://github.com/CycloneDX/cyclonedx-property-taxonomy/pull/8


cc: @luhring, @wagoodman, @spiffcs, @kzantow please take a look. My reasoning for choosing `syft` over `anchore` was in case there was ever some movement in the future on https://github.com/ossf/tac/issues/70. The choice of a namespace like syft ensures that the namespace is easily identifiable and closely associated to syft while still being impervious to such changes.

samj1912:
Separately it looks like there is an spdx taxonomy in work as well which contains the CPE field at https://github.com/CycloneDX/cyclonedx-property-taxonomy/pull/7 which might also be suitable for storing this and other SPDX fields. This might also be useful with reference to #563, #723 and #737

kzantow:
@samj1912 since https://github.com/anchore/syft/pull/849 has been merged, which outputs `syft:` namespace'd CPEs, I'm going to close this one; please reopen if you disagree!

PR #820, opened 2022-02-14
Update cyclonedx to v1.4

### LINKED ISSUE #744: Add support for CycloneDX 1.4
CycloneDX 1.4 has been released. Syft should be updated to reflect the new spec.
https://cyclonedx.org/news/cyclonedx-v1.4-released/

samj1912:

Fixes #744


kzantow (commented):
One issue we have not solved yet - especially for decoding - is how to accommodate multiple versions of a spec, I think we should be mindful to keep the version information as part of the package name. However, I might suggest something to possibly avoid the sprawl of version/format/etc. if we have some base thing like `cyclonedx/v1_3` and within that `format_json.go`, etc. this also lets us keep all the `v1_3` stuff together, then copy it to `cyclonedx/v1_4` and update whatever is needed, etc. and at some point deprecate support for older versions. WDYT?

samj1912:
Noting this thread from slack ->


> [Keith Zantow](https://app.slack.com/team/U029YCTBC1K)  [1 hour ago](https://anchorecommunity.slack.com/archives/C4PJFNEEM/p1644849581941089?thread_ts=1644829204.379759&cid=C4PJFNEEM)
> Question: does this drop support for 1.3? It looks like 1.4 is a superset of 1.3, so all should be well in this case, but we need to have a strategy to support multiple versions -- at least for the decoding side of things.
> [sam](https://app.slack.com/team/U02PUPE1FGT)  [23 minutes ago](https://anchorecommunity.slack.com/archives/C4PJFNEEM/p1644851440434959?thread_ts=1644829204.379759&cid=C4PJFNEEM)
> 1.4 is a superset of 1.3 - so a 1.3 document can be decoded into a 1.4 document but not vice versa
> [sam](https://app.slack.com/team/U02PUPE1FGT)  [23 minutes ago](https://anchorecommunity.slack.com/archives/C4PJFNEEM/p1644851454686999?thread_ts=1644829204.379759&cid=C4PJFNEEM)
> the only real change is the bomVersion field
> [Keith Zantow](https://app.slack.com/team/U029YCTBC1K)  [20 minutes ago](https://anchorecommunity.slack.com/archives/C4PJFNEEM/p1644851643087439?thread_ts=1644829204.379759&cid=C4PJFNEEM)
> great, I figured as much for this one

wagoodman:
I still think we should hold off on merging this until we get the CycloneDX decoding PR in and then upgrade to 1.4

kzantow:
@samj1912 could you regenerate the golden files and other things based on the recent decoder merge? I think we can move forward on this after that.

samj1912:
@kzantow sorry for the long delay, had to play catch up with the upstream changes a couple of times which caused some rebase conflicts. Should be good to go now!

kzantow (approved):
LGTM; thanks for getting this done, @samj1912 !

PR #824, opened 2022-02-14
SBOM input

kzantow:
Quick hack to import SBOMs into Syft using the `sbom:<file>` syntax.

**At a minimum, this needs tests to get merged.**

### DIFF
--- a/cmd/packages.go
+++ b/cmd/packages.go
@@ -3,8 +3,10 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


wagoodman:
Is the reason for allowing an `sbom` scheme to be able to support format conversions (tracked https://github.com/anchore/syft/issues/563)? Or is there other use cases for this?

kzantow:
@wagoodman yeah, it is basically to support #563 ; I just wrote this to understand missing data while working on SPDX import, and thought it would be worthwhile to keep around at least as a reference

kzantow:
Closing this for now; will be covered by a bit of a different solution

PR #827, opened 2022-02-15
Use go 1.18 buildinfo to catalog binaries

### LINKED ISSUE #718: Capture additional go 1.18 based binary information
Binaries build with go 1.18+ have additional build information available in the binary `buildinfo` sections:
```
$ go version -m /tmp/syft
/tmp/syft: go1.18beta1
	path	github.com/anchore/syft
	mod	github.com/anchore/syft	(devel)
	dep	github.com/CycloneDX/cyclonedx-go	v0.4.0	h1:Wz4QZ9B4RXGWIWTypVLEOVJgOdFfy5mcS5PGNzUkZxU=
	dep	github.com/adrg/xdg	v0.2.1	h1:VSVdnH7cQ7V+B33qSJHTCRlNgra1607Q8PzEmnvb2Ic=
	dep	github.com/anchore/client-go	v0.0.0-20210222170800-9c70f9b80bcf	h1:DYssiUV1pBmKqzKsm4mqXx8artqC0Q8HgZsVI3lMsAg=
	dep	github.com/anchore/go-presenter	v0.0.0-20211102174526-0dbf20f6c7fa	h1:mDLUAkgXsV5Z8D0EEj8eS6FBekolV/A+Xxbs9054bPw=
	dep	github.com/anchore/go-rpmdb	v0.0.0-20210914181456-a9c52348da63	h1:C9W/LAydEz/qdUhx1MdjO9l8NEcFKYknkxDVyo9LAoM=
	dep	github.com/anchore/go-version	v1.2.2-0.20200701162849-18adb9c92b9b	h1:e1bmaoJfZVsCYMrIZBpFxwV26CbsuoEh5muXD5I1Ods=
	dep	github.com/anchore/packageurl-go	v0.0.0-20210922164639-b3fa992ebd29	h1:K9LfnxwhqvihqU0+MF325FNy7fsKV9EGaUxdfR4gnWk=
	dep	github.com/anchore/stereoscope	v0.0.0-20211203160213-5a5e323a5c89	h1:pcMFN7wjIQIxdZm8NA0R3JiRRFn5Eyx9mhagHOVS4Bw=
	dep	github.com/andybalholm/brotli	v1.0.4	h1:V7DdXeJtZscaqfNuAdSRuRFzuiKlHSC/Zh3zl9qY3JY=
	dep	github.com/antihax/optional	v1.0.0	h1:xK2lYat7ZLaVVcIuj82J8kIro4V6kDe0AUDFboUCwcg=
	dep	github.com/bmatcuk/doublestar/v2	v2.0.4	h1:6I6oUiT/sU27eE2OFcWqBhL1SwjyvQuOssxT4a1yidI=
	dep	github.com/bmatcuk/doublestar/v4	v4.0.2	h1:X0krlUVAVmtr2cRoTqR8aDMrDqnB36ht8wpWTiQ3jsA=
	dep	github.com/containerd/containerd	v1.5.8	h1:NmkCC1/QxyZFBny8JogwLpOy2f+VEbO/f6bV2Mqtwuw=
	dep	github.com/containerd/stargz-snapshotter/estargz	v0.10.0	h1:glqzafvxBBAMo+x2w2sdDjUDZeTqqLJmqZPY05qehCU=
	dep	github.com/docker/cli	v20.10.10+incompatible	h1:kcbwdgWbrBOH8QwQzaJmyriHwF7XIl4HT1qh0HTRys4=
	dep	github.com/docker/distribution	v2.7.1+incompatible	h1:a5mlkVzth6W5A4fOsS3D2EO5BUmsJpcB+cRlLU7cSug=
	dep	github.com/docker/docker	v20.10.11+incompatible	h1:OqzI/g/W54LczvhnccGqniFoQghHx3pklbLuhfXpqGo=
	dep	github.com/docker/docker-credential-helpers	v0.6.4	h1:axCks+yV+2MR3/kZhAmy07yC56WZ2Pwu/fKWtKuZB0o=
	dep	github.com/docker/go-connections	v0.4.0	h1:El9xVISelRB7BuFusrZozjnkIM5YnzCViNKohAFqRJQ=
	dep	github.com/docker/go-units	v0.4.0	h1:3uh0PgVws3nIA0Q+MwDC8yjEPf9zjRfZZWXZYDct3Tw=
	dep	github.com/dsnet/compress	v0.0.2-0.20210315054119-f66993602bf5	h1:iFaUwBSo5Svw6L7HYpRu/0lE3e0BaElwnNO1qkNQxBY=
	dep	github.com/dustin/go-humanize	v1.0.0	h1:VSnTsYCnlFHaM2/igO1h6X3HA71jcobQuxemgkq4zYo=
	dep	github.com/facebookincubator/nvdtools	v0.1.4	h1:x1Ucw9+bSkMd8DJJN4jNQ1Lk4PSFlJarGOxp9D6WUMo=
	dep	github.com/fsnotify/fsnotify	v1.4.9	h1:hsms1Qyu0jgnwNXIxa+/V/PDsU6CfLf6CNO8H7IWoS4=
	dep	github.com/gabriel-vasile/mimetype	v1.3.0	h1:4YOHITFLyYwF+iqG0ybSLGArRItynpfwdlWRmJnd75E=
	dep	github.com/go-restruct/restruct	v1.2.0-alpha	h1:2Lp474S/9660+SJjpVxoKuWX09JsXHSrdV7Nv3/gkvc=
	dep	github.com/gogo/protobuf	v1.3.2	h1:Ov1cvc58UF3b5XjBnZv7+opcTcQFZebYjWzi34vdm4Q=
	dep	github.com/golang/protobuf	v1.5.2	h1:ROPKBNFfQgOUMifHyP+KYbvpjbdoFNs+aK7DXlji0Tw=
	dep	github.com/golang/snappy	v0.0.3	h1:fHPg5GQYlCeLIPB9BZqMVR5nR9A+IM5zcgeTdjMYmLA=
	dep	github.com/google/go-containerregistry	v0.7.0	h1:u0onUUOcyoCDHEiJoyR1R1gx5er1+r06V5DBhUU5ndk=
	dep	github.com/google/uuid	v1.2.0	h1:qJYtXnJRWmpe7m/3XlyhrsLrEURqHRM2kxzoxXqyUDs=
	dep	github.com/gookit/color	v1.2.7	h1:4qePMNWZhrmbfYJDix+J4V2l0iVW+6jQGjicELlN14E=
	dep	github.com/hashicorp/errwrap	v1.0.0	h1:hLrqtEDnRye3+sgx6z4qVLNuviH3MR5aQ0ykNJa/UYA=
	dep	github.com/hashicorp/go-multierror	v1.1.0	h1:B9UzwGQJehnUY1yNrnwREHc3fGbC2xefo8g4TbElacI=
	dep	github.com/hashicorp/go-version	v1.2.0	h1:3vNe/fWF5CBgRIguda1meWhsZHy3m8gCJ5wx+dIzX/E=
	dep	github.com/hashicorp/hcl	v1.0.0	h1:0Anlzjpi4vEasTeNFn2mLJgTSwt0+6sfsiTG8qcWGx4=
	dep	github.com/jinzhu/copier	v0.3.2	h1:QdBOCbaouLDYaIPFfi1bKv5F5tPpeTwXe4sD0jqtz5w=
	dep	github.com/klauspost/compress	v1.13.6	h1:P76CopJELS0TiO2mebmnzgWaajssP/EszplttgQxcgc=
	dep	github.com/klauspost/pgzip	v1.2.5	h1:qnWYvvKqedOF2ulHpMG72XQol4ILEJ8k2wwRl/Km8oE=
	dep	github.com/magiconair/properties	v1.8.5	h1:b6kJs+EmPFMYGkow9GiUyCyOvIwYetYJ3fSaWak/Gls=
	dep	github.com/mattn/go-colorable	v0.0.9	h1:UVL0vNpWh04HeJXV0KLcaT7r06gOH2l4OW6ddYRUIY4=
	dep	github.com/mattn/go-isatty	v0.0.6	h1:SrwhHcpV4nWrMGdNcC2kXpMfcBVYGDuTArqyhocJgvA=
	dep	github.com/mattn/go-runewidth	v0.0.7	h1:Ei8KR0497xHyKJPAv59M1dkC+rOZCMBJ+t3fZ+twI54=
	dep	github.com/mgutz/ansi	v0.0.0-20200706080929-d51e80ef957d	h1:5PJl274Y63IEHC+7izoQE9x6ikvDFZS2mDVS3drnohI=
	dep	github.com/mholt/archiver/v3	v3.5.1	h1:rDjOBX9JSF5BvoJGvjqK479aL70qh9DIpZCl+k7Clwo=
	dep	github.com/mitchellh/go-homedir	v1.1.0	h1:lukF9ziXFxDFPkA1vsr5zpc1XuPDn/wFntq5mG+4E0Y=
	dep	github.com/mitchellh/hashstructure/v2	v2.0.2	h1:vGKWl0YJqUNxE8d+h8f6NJLcCJrgbhC4NcD46KavDd4=
	dep	github.com/mitchellh/mapstructure	v1.4.1	h1:CpVNEelQCZBooIPDn+AR3NpivK/TIKU8bDxdASFVQag=
	dep	github.com/nwaples/rardecode	v1.1.0	h1:vSxaY8vQhOcVr4mm5e8XllHWTiM4JF507A0Katqw7MQ=
	dep	github.com/olekukonko/tablewriter	v0.0.4	h1:vHD/YYe1Wolo78koG299f7V/VAS08c6IpCLn+Ejf/w8=
	dep	github.com/opencontainers/go-digest	v1.0.0	h1:apOUWs51W5PlhuyGyz9FCeeBIOUDA/6nW8Oi/yOhh5U=
	dep	github.com/opencontainers/image-spec	v1.0.2	h1:9yCKha/T5XdGtO0q9Q9a6T5NUCsTn/DrBg0D7ufOcFM=
	dep	github.com/pelletier/go-toml	v1.9.3	h1:zeC5b1GviRUyKYd6OJPvBU/mcVDVoL1OhT17FCt5dSQ=
	dep	github.com/pierrec/lz4/v4	v4.1.2	h1:qvY3YFXRQE/XB8MlLzJH7mSzBs74eA2gg52YTk6jUPM=
	dep	github.com/pkg/errors	v0.9.1	h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
	dep	github.com/pkg/profile	v1.5.0	h1:042Buzk+NhDI+DeSAA62RwJL8VAuZUMQZUjCsRz1Mug=
	dep	github.com/scylladb/go-set	v1.0.2	h1:SkvlMCKhP0wyyct6j+0IHJkBkSZL+TDzZ4E7f7BCcRE=
	dep	github.com/sirupsen/logrus	v1.8.1	h1:dJKuHgqk1NNQlqoA6BTlM1Wf9DOH3NBjQyu0h9+AZZE=
	dep	github.com/spdx/tools-golang	v0.1.0	h1:iDMNEPqQk6CdiDj6eWDIDw85j0wQ3IR3pH9p0X05TSQ=
	dep	github.com/spf13/afero	v1.6.0	h1:xoax2sJ2DT8S8xA2paPFjDCScCNeWsg75VG0DLRreiY=
	dep	github.com/spf13/cast	v1.3.1	h1:nFm6S0SMdyzrzcmThSipiEubIDy8WEXKNZ0UOgiRpng=
	dep	github.com/spf13/cobra	v1.2.1	h1:+KmjbUw1hriSNMF55oPrkZcb27aECyrj8V2ytv7kWDw=
	dep	github.com/spf13/jwalterweatherman	v1.1.0	h1:ue6voC5bR5F8YxI5S67j9i582FU4Qvo2bmqnqMYADFk=
	dep	github.com/spf13/pflag	v1.0.5	h1:iy+VFUOCP1a+8yFto/drg2CJ5u0yRoB7fZw3DKv/JXA=
	dep	github.com/spf13/viper	v1.8.1	h1:Kq1fyeebqsBfbjZj4EL7gj2IO0mMaiyjYUWcUsl2O44=
	dep	github.com/subosito/gotenv	v1.2.0	h1:Slr1R9HxAlEKefgq5jn9U+DnETlIUa6HfgEzj0g5d7s=
	dep	github.com/ulikunitz/xz	v0.5.9	h1:RsKRIA2MO8x56wkkcd3LbtcE/uMszhb6DpRf+3uwa3I=
	dep	github.com/vbatts/tar-split	v0.11.2	h1:Via6XqJr0hceW4wff3QRzD5gAk/tatMw/4ZA7cTlIME=
	dep	github.com/vifraa/gopom	v0.1.0	h1:v897eVxf6lflkEXzPmKbo4YhX2oS/LGjz7cqjWnSmCU=
	dep	github.com/wagoodman/go-partybus	v0.0.0-20210627031916-db1f5573bbc5	h1:phTLPgMRDYTizrBSKsNSOa2zthoC2KsJsaY/8sg3rD8=
	dep	github.com/wagoodman/go-progress	v0.0.0-20200731105512-1020f39e6240	h1:r6BlIP7CVZtMlxUQhT40h1IE1TzEgKVqwmsVGuscvdk=
	dep	github.com/wagoodman/jotframe	v0.0.0-20211129225309-56b0d0a4aebb	h1:Yz6VVOcLuWLAHYlJzTw7JKnWxdV/WXpug2X0quEzRnY=
	dep	github.com/x-cray/logrus-prefixed-formatter	v0.5.2	h1:00txxvfBM9muc0jiLIEAkAcIMJzfthRT6usrui8uGmg=
	dep	github.com/xi2/xz	v0.0.0-20171230120015-48954b6210f8	h1:nIPpBwaJSVYIxUFsDv3M8ofmx9yWTog9BfvIu0q41lo=
	dep	golang.org/x/crypto	v0.0.0-20210322153248-0c34fe9e7dc2	h1:It14KIkyBFYkHkwZ7k45minvA9aorojkyjGk9KJ5B/w=
	dep	golang.org/x/mod	v0.4.2	h1:Gz96sIWK3OalVv/I/qNygP42zyoKp3xptRVCWRFEBvo=
	dep	golang.org/x/net	v0.0.0-20211111160137-58aab5ef257a	h1:c83jeVQW0KGKNaKBRfelNYNHaev+qawl9yaA825s8XE=
	dep	golang.org/x/oauth2	v0.0.0-20211104180415-d3ed0bb246c8	h1:RerP+noqYHUQ8CMRcPlC2nvTa4dcBIjegkuWdcUDuqg=
	dep	golang.org/x/sync	v0.0.0-20210220032951-036812b2e83c	h1:5KslGYwFpkhGh+Q16bwMP3cOontH8FOep7tGV86Y7SQ=
	dep	golang.org/x/sys	v0.0.0-20211110154304-99a53858aa08	h1:WecRHqgE09JBkh/584XIE6PMz5KKE/vER4izNUi30AQ=
	dep	golang.org/x/term	v0.0.0-20210615171337-6886f2dfbf5b	h1:9zKuko04nR4gjZ4+DNjHqRlAJqbJETHwiNKDqTfOjfE=
	dep	golang.org/x/text	v0.3.6	h1:aRYxNxv6iGQlyVaZmk6ZgYEDa+Jg18DxebPSrd6bg1M=
	dep	golang.org/x/xerrors	v0.0.0-20200804184101-5ec99f83aff1	h1:go1bK/D/BFZV2I8cIQd1NKEZ+0owSTG1fDTci4IqFcE=
	dep	google.golang.org/genproto	v0.0.0-20211111162719-482062a4217b	h1:qvEQEwKjZRAg6rjY/jqfJ7T8/w/D7jTIFJGcaSka96k=
	dep	google.golang.org/grpc	v1.42.0	h1:XT2/MFpuPFsEX2fWh3YQtHkZ+WYZFQRfaUgLZYj/p6A=
	dep	google.golang.org/protobuf	v1.27.1	h1:SnqbnDw1V7RiZcXPx5MEeqPv2s79L9i7BJUlG/+RurQ=
	dep	gopkg.in/ini.v1	v1.62.0	h1:duBzk771uxoUuOlyRLkHsygud9+5lrlGjdFBb4mSKDU=
	dep	gopkg.in/yaml.v2	v2.4.0	h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
	build	-compiler=gc
	build	CGO_ENABLED=1
	build	CGO_CFLAGS=
	build	CGO_CPPFLAGS=
	build	CGO_CXXFLAGS=
	build	CGO_LDFLAGS=
	build	GOARCH=amd64
	build	GOOS=darwin
	build	GOAMD64=v1
	build	vcs=git
	build	vcs.revision=b7979dbc7d62c195a662c022c0b8c4ecbc069958
	build	vcs.time=2021-12-21T15:13:15Z
	build	vcs.modified=true
```

The new notable section is the lines prefixed with `build`. The existing `pkg.GoBinaryMetadata` can be expanded to include this new build information. 

**Why is this needed**:
This further pins down information that the go compiler knows about the state of the filesystem and inputs provided to the compiler, which are critical perspectives and should be incorporated into our SBOM output.

**Additional context**:
go 1.18 has much of the needed functionality exposed for interrogating other binaries (and not just for the running process) within the `debug/buildinfo` package. The downside is that these packages do not support multiarchitecture binary support for macho (darwin) binaries. We should consider trying to use the new `buildinfo` package to get this new functionality but also consider other approaches (such as taking in stdlib code internally or taking a hybrid approach where we delete our existing buildinfo code except for the multiarchitecture binary case). We should not regress (remove) the multiarchitecture binary support (which is needed for goreleaser integration support).

See https://go.dev/blog/go1.18beta1 and https://tip.golang.org/doc/go1.18#debug/buildinfo

Also, further investigation is needed to see if there is ever a version provided for the module itself (e.g. `mod	github.com/anchore/syft	(devel)` ... does this always say "devel"? or could there be other values here?). If top-level version information can be extracted that is very valuable and should also be considered in scope (if possible).

jonasagx:
**_Note: this cannot be merged until go 1.18 is released._**

## additional module to hold build info
```go
NAME               VERSION                             TYPE
golang.org/x/sys   v0.0.0-20220223155357-96fed51e1446  go-module
golang.org/x/term  v0.0.0-20210927222741-03fcf44c2211  go-module
universal.bin/v0                                       go-module # this one is the main module of this example
```

## golangci-lint doesn't work with go1.18
This PR added a golangci-lint workflow running on go1.17. Running other checkers on go1.17 won't work, e.g. bouncer fails to recognize `runtime/debug` (introduced in 1.18).

## added support to xcoff files
Go provides support to xcoff binaries via an internal library (IDK why), to support this file I copied part of this library into syft. Keep in mind the distinction between binary format and architecture.

## fixed bug where `foundBy` was empty
known bug, present in older versions of syft

## missing
- [x] use vcs.revision as main module version, when present
- [x] don't use [mod.Main.Version](https://github.com/jonasagx/syft/blob/golang-cataloger-refactor/syft/pkg/cataloger/golang/parse_go_bin.go#L170 
) to move data around 
- [ ] make golangci-lint required for merging
- [x] Update json schema
- [x] review on `getGOARCHFromBin` tests

Closes #718 


wagoodman:
@jonasagx we should try and capture the main module and only persist the build options on that module. The version could either be blank or be the vcs.revision if it exists (not ideal).

jonasagx:
@wagoodman yes, I changed it so only the main mod gets build configs. On version, what about just use "(devel)" from the main module?

syft/pkg/cataloger/golang/version.go
```diff
@@ -113,114 +31,71 @@ func isExe(file string, info fs.FileInfo) bool {
...
-	if err != nil || len(hdr) < 2*ptrSize {
-		return ""
-	}
-	dataAddr := readPtr(hdr)
-	dataLen := readPtr(hdr[ptrSize:])
-	data, err := x.ReadData(dataAddr, dataLen)
-	if err != nil || uint64(len(data)) < dataLen {
-		return ""
+	// adding macho multi-architecture support (both for 64bit and 32 bit)... this case is not in the stdlib yet
+	if bytes.HasPrefix(data, []byte("\xCA\xFE\xBA\xBE")) || bytes.HasPrefix(data, []byte("\xCA\xFE\xBA\xBF")) {
```
wagoodman, line 86: this seems to be mixing levels of concerns here; You could probably use https://github.com/anchore/go-macholibre/blob/main/universal_binary.go#L139 to determine if this should be accessed via macholibre

syft/pkg/cataloger/golang/version.go
```diff
@@ -1,109 +1,27 @@
...
+	"debug/buildinfo"
+	"errors"
+	"io"
+	"io/fs"
+	"os"
+	"runtime"
+	"runtime/debug"
+	"strings"
+
+	macho "github.com/anchore/go-macholibre"
```
wagoodman, line 21: there is also a `macho` package in the stdlib; why alias this import? (without the alias items would be accessed via `macholibre`)

go.mod
```diff
@@ -56,61 +55,56 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
-require github.com/sigstore/cosign v1.5.2-0.20220302132634-73235ac02e4a
+require (
+	github.com/in-toto/in-toto-golang v0.3.4-0.20211211042327-af1f9fb822bf
```
wagoodman, line 59: any reason for moving in-toto-golang to the next require section?

jonasagx: Weird,  I can't reproduce it with mod tidy, but it is fixed now

wagoodman:
re: devel, this value isn't really useful since it would be the same value for any main module. That's why I feel that the version value should either be blank or attempt to use the vcs.version.

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,49 +34,46 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
 }
 
-func parseGoBin(location source.Location, reader io.ReadCloser, opener exeOpener) (pkgs []pkg.Package, err error) {
-	var exes []exe
-	// it has been found that there are stdlib paths within openExe that can panic. We want to prevent this behavior
-	// bubbling up and halting execution. For this reason we try to recover from any panic and return an error.
-	defer func() {
-		if r := recover(); r != nil {
-			err = fmt.Errorf("recovered from panic while parse go binary at %q: %+v", location.RealPath, r)
+func getGOARCH(settings []debug.BuildSetting) string {
```
wagoodman, line 37: this had been based on the architecture from the binary --it's ok to now source this from the build settings, however, do we know if this value is guaranteed to be set for all modules? (honest question, I haven't checked the go source code on this one)

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,49 +34,46 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
+		m[s.Key] = s.Value
+	}
+	return m
+}
 
-	// filter mod dependencies: [dep, name, version, sha]
-	for scanner.Scan() {
-		fields := strings.Fields(scanner.Text())
+func buildGoPkgInfo(location source.Location, mod *debug.BuildInfo) []pkg.Package {
+	pkgs := make([]pkg.Package, 0)
```
wagoodman, line 56: nit: allocating the slice isn't necessary here

syft/pkg/cataloger/golang/parse_go_bin_test.go
```diff
@@ -11,24 +11,63 @@ import (
...
 	tests := []struct {
 		name     string
-		mod      string
+		mod      *debug.BuildInfo
 		expected []pkg.Package
 	}{
 		{
 			name:     "buildGoPkgInfo parses a blank mod string and returns no packages",
-			mod:      "",
+			mod:      &debug.BuildInfo{},
```
wagoodman, line 51: this isn't checking the case where `nil` is passed

syft/pkg/cataloger/golang/version.go
```diff
@@ -113,114 +30,70 @@ func isExe(file string, info fs.FileInfo) bool {
...
-	if err != nil || len(hdr) < 2*ptrSize {
-		return ""
-	}
-	dataAddr := readPtr(hdr)
-	dataLen := readPtr(hdr[ptrSize:])
-	data, err := x.ReadData(dataAddr, dataLen)
-	if err != nil || uint64(len(data)) < dataLen {
-		return ""
+	if macho.IsUniversalMachoBinary(f) {
+		var readers []io.ReaderAt
```
wagoodman, line 85: super nit: moving this declaration towards where its used helps with reading, in this case just before the `range ers`

syft/pkg/cataloger/golang/version.go
```diff
@@ -113,114 +30,70 @@ func isExe(file string, info fs.FileInfo) bool {
...
+		ers, err := macho.ExtractReaders(f)
```
wagoodman, line 86: nit: `ers` vs `err` for variable names seems tricky, maybe consider a rename of `ers`?

syft/pkg/cataloger/golang/parse_go_bin.go
jonasagx: https://github.com/golang/go/blob/bf97c99b62fe7d6652cc8c807dbc91998d488a01/src/cmd/go/internal/load/pkg.go#L2336 sets the architecture, it happens inside `setBuildInfo()` which states:

```
// setBuildInfo gathers build information, formats it as a string to be
// embedded in the binary, then sets p.Internal.BuildInfo to that string.
// setBuildInfo should only be called on a main package with no errors.
//
// This information can be retrieved using debug.ReadBuildInfo.
```
https://github.com/golang/go/blob/bf97c99b62fe7d6652cc8c807dbc91998d488a01/src/cmd/go/internal/load/pkg.go#L2211L2218

jonasagx: I am looking into compatibility with older version of go binaries that won't have the key/pair

wagoodman: > I am looking into compatibility with older version of go binaries that won't have the key/pair

I think this hints at still sourcing this from the binary instead of the build settings.

go.sum
```diff
@@ -1415,6 +1389,8 @@ github.com/jmespath/go-jmespath/internal/testify v1.5.1 h1:shLQSRRSCCPj3f2gpwzGw
 github.com/jmespath/go-jmespath/internal/testify v1.5.1/go.mod h1:L3OGu8Wl2/fWfCI6z80xFu9LTZmf1ZRjMHUOPmWr69U=
 github.com/jmoiron/sqlx v1.2.0/go.mod h1:1FEQNm3xlJgrMD+FBdI9+xvCksHtbpVBBw5dYhBSsks=
 github.com/joho/godotenv v1.3.0/go.mod h1:7hK45KPybAkOC6peb+G5yklZfMxEjkZhHbwpqxOKXbg=
+github.com/jonasagx/stereoscope v0.0.0-20220314180508-62daa1e1d05e h1:LqeODh2PUdYkA/KdtYtDBGF/vfeVZab0Qsywld1jsWA=
+github.com/jonasagx/stereoscope v0.0.0-20220314180508-62daa1e1d05e/go.mod h1:OVVJ4y/L26pspeeHNvNa7GeQfoLPWxJe13iODl9x5l4=
```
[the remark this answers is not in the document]
jonasagx: It is from testing another PR, it will be updated once I merge stereoscope. Alex and I are going over that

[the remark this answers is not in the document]
jonasagx: updated

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +63,112 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
+		}
+		arch = fmt.Sprintf("%d", f.Machine)
+	case bytes.HasPrefix(ident, []byte("\xFE\xED\xFA")) || bytes.HasPrefix(ident[1:], []byte("\xFA\xED\xFE")):
+		f, err := macho.NewFile(r)
+		if err != nil {
+			return "", errUnrecognizedFormat
+		}
+		arch = f.Cpu.String()
+	case bytes.HasPrefix(ident, []byte{0x01, 0xDF}) || bytes.HasPrefix(ident, []byte{0x01, 0xF7}):
+		arch = "xcoff"
```
wagoodman, line 129: @wagoodman need to double check this value

wagoodman, line 129: @wagoodman need to double check this value

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +63,112 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
+			log.Warnf("golang cataloger: bin parsing: getting arch from binary: %v", err)
+			continue
 		}
-	}()
 
-	// Identify if bin was compiled by go
-	exes, err = opener(reader)
-	if err != nil {
-		return pkgs, err
+		builds[i].Settings = append(builds[i].Settings, debug.BuildSetting{Key: GOARCH, Value: a})
```
wagoodman, line 86: we shouldn't be modifying the build settings; the architecture should be raised up separately

wagoodman, line 86: we shouldn't be modifying the build settings; the architecture should be raised up separately

luhring:
Nice! Just took this for a spin, and here are my comments... @jonasagx @wagoodman 

I ran this, just to test the latest code (at ed876d0) on the Syft binary that was in my path (built with go1.17.7):

```console
$ go run . file:$(which syft) -o json
```

1. I see `"foundBy": "",` for all of the packages. Shouldn't we show the Go cataloger value here?

2. For the non-main modules, I see `goBuildSettings` is set to `null`, but we do have a value for `goCompiledVersion` always set. IIUC, we should either set BOTH of these values for all modules, or, set both of these values ONLY for the main module, and not for other modules. To me, they both seem central to the binary build. What's the thinking here?

3. Is there a precedent for setting values to `null` in the JSON output? I couldn't remember if I had seen this before, but it looked relatively unfamiliar, so just asking for a sanity check...

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +63,112 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
-		if fields[0] == replaceIdentifier {
-			// replace the previous entry in the package slice
-			pkgsSlice[len(pkgsSlice)-1] = newGoBinaryPackage(name, version, h1Digest, goVersion, arch, location)
-		}
+	var empty debug.Module
+	if mod.Main == empty {
+		return pkgs
 	}
-	return pkgsSlice
+	mod.Main.Version = ""
```
wagoodman, line 168: we shouldn't modify this value

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,27 @@
...
+       - main
+  pull_request:
+
+env:
+  GO_VERSION: "1.17.x"
+  GO_STABLE_VERSION: true
+
+jobs:
+  golangci:
+    name: "GolangCI parkour around go 1.18 🤦"
```
wagoodman, line 15: we shouldn't need this new workflow, but if we did we would need this check to be aligned with the release quality gate

wagoodman, line 15: we shouldn't need this new workflow, but if we did we would need this check to be aligned with the release quality gate

.github/workflows/validations.yaml
```diff
@@ -7,7 +7,8 @@ on:
   pull_request:
 
 env:
-  GO_VERSION: "1.17.x"
+  GO_VERSION: "1.18.x"
```
wagoodman, line 10: we could run all of static-analysis under 1.17, which would minimize changes

wagoodman, line 10: we could run all of static-analysis under 1.17, which would minimize changes

.golangci.yaml
```diff
@@ -42,7 +42,11 @@ linters:
     - unused
     - varcheck
     - whitespace
-
+run:
+  timeout: 4m
+  # skip-files:
```
wagoodman, line 47: do we need this? it seems like dead configuration

wagoodman, line 47: do we need this? it seems like dead configuration

Makefile
```diff
@@ -3,7 +3,7 @@ TEMPDIR = ./.tmp
 RESULTSDIR = test/results
 COVER_REPORT = $(RESULTSDIR)/unit-coverage-details.txt
 COVER_TOTAL = $(RESULTSDIR)/unit-coverage-summary.txt
-LINTCMD = $(TEMPDIR)/golangci-lint run --tests=false --timeout=4m --config .golangci.yaml
+# LINTCMD = $(TEMPDIR)/golangci-lint run --tests=false --timeout=4m --config .golangci.yaml
```
wagoodman, line 6: if we run the static analysis job under 1.17 we wouldn't need to remove this

wagoodman, line 6: if we run the static analysis job under 1.17 we wouldn't need to remove this

Makefile
```diff
@@ -132,7 +131,9 @@ lint: ## Run gofmt + golangci lint checks
 	@test -z "$(shell gofmt -l -s .)"
 
 	# run all golangci-lint rules
-	$(LINTCMD)
+# 	NOTE(jonasagx): As of Mar-4th, 2022: golangci-lint runs on its own worklow, with go 1.17,
+# 	because incompatibilities with go 1.18
+#	$(LINTCMD)
```
wagoodman, line 136: if we run the static analysis job under 1.17 we wouldn't need to remove this

wagoodman, line 136: if we run the static analysis job under 1.17 we wouldn't need to remove this

Makefile
```diff
@@ -142,7 +143,7 @@ lint: ## Run gofmt + golangci lint checks
 lint-fix: ## Auto-format all source code + run golangci lint fixers
 	$(call title,Running lint fixers)
 	gofmt -w -s .
-	$(LINTCMD) --fix
+# 	$(LINTCMD) --fix
```
wagoodman, line 146: if we run the static analysis job under 1.17 we wouldn't need to remove this

wagoodman, line 146: if we run the static analysis job under 1.17 we wouldn't need to remove this

go.mod
```diff
@@ -327,3 +291,7 @@ require (
 	gopkg.in/ini.v1 v1.66.2 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+require github.com/docker/docker v20.10.12+incompatible
```
wagoodman, line 295: nit: since these were explicitly pinned we should have comments for these

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +63,112 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
 	}
-	return pkgs, err
+
+	return ""
 }
 
-func buildGoPkgInfo(location source.Location, mod, goVersion, arch string) []pkg.Package {
-	pkgsSlice := make([]pkg.Package, 0)
-	scanner := bufio.NewScanner(strings.NewReader(mod))
+func getGOARCHFromBin(r io.ReaderAt) (string, error) {
```
wagoodman, line 100: this function is a rewrite from a set of functions that were extracted from the standard lib in a previous version of go. Now that this has been rewritten we should try to add more tests for this functionality.

test/integration/regression_go_bin_scanner_arch_test.go
```diff
@@ -9,9 +9,9 @@ import (
 
 func TestRegressionGoArchDiscovery(t *testing.T) {
 	const (
-		expectedELFPkg   = 3
-		expectedWINPkg   = 3
-		expectedMACOSPkg = 3
+		expectedELFPkg   = 4
```
wagoodman, line 12: what changed that caused these numbers to be bumped?

syft/pkg/cataloger/golang/version.go
```diff
@@ -1,226 +1,99 @@
...
+		log.Warnf("golang cataloger: opening binary: %v", err)
+		return nil
 	}
-	defer x.Close()
 
-	vers, mod := findVers(x)
-	if vers == "" {
-		if mustPrint {
-			fmt.Fprintf(os.Stderr, "%s: go version not found\n", file)
+	builds := make([]*debug.BuildInfo, 0)
```
wagoodman, line 55: nit: could be `var builds []*debug.BuildInfo`

syft/pkg/cataloger/golang/version.go
```diff
@@ -1,226 +1,99 @@
...
+	for _, r := range readers {
+		bi, err := buildinfo.Read(r)
+		if err != nil {
+			if pathErr := (*os.PathError)(nil); !errors.As(err, &pathErr) {
```
wagoodman, line 59: I don't think I understand the assignment to `pathErr` before the error check. Can you walk me through what this line does?

syft/pkg/cataloger/golang/version.go
```diff
@@ -1,226 +1,99 @@
...
-// findVers finds and returns the Go version and module version information
-// in the executable x.
-func findVers(x exe) (vers, mod string) {
-	// Read the first 64kB of text to find the build info blob.
-	text := x.DataStart()
-	data, err := x.ReadData(text, 64*1024)
+
+// openExe opens file and returns it as io.ReaderAt.
+func getReaders(f unionReader) ([]io.ReaderAt, error) {
+	data := make([]byte, 16)
```
wagoodman, line 74: the 16-byte slice + readfull + seek seems like dead code -- is this doing anything?

test/integration/regression_go_bin_scanner_arch_test.go
jonasagx: we now add the main package

syft/pkg/cataloger/golang/version.go
```diff
@@ -1,226 +1,99 @@
...
-		info = i
-	}
-
-	if !isExe(file, info) {
-		if mustPrint {
-			fmt.Fprintf(os.Stderr, "%s: not executable file\n", file)
-		}
-		return
+func scanFile(reader unionReader, filename string, mode os.FileMode) []*debug.BuildInfo {
+	if !isExe(filename, mode) {
```
wagoodman, line 42: I think querying the cataloger already accounts for this kind of filtering (only requesting executables, though it does it by asking the resolve by MIME type instead of checking file extensions and mode values). So it shouldn't be needed. Was there a specific reason why an exe check was added (or kept around from previous code)?

syft/pkg/cataloger/golang/version.go
```diff
@@ -1,226 +1,99 @@
 // This code was copied from the Go std library.
```
wagoodman, line 1: this statement is no longer true -- before this PR it was true. It seems like this file has gone through a rewrite entirely. Also, the name `version.go` came from the standard lib as well; is there a better name for this now?

jonasagx: removed, thx

syft/pkg/cataloger/golang/version.go
jonasagx: you're right, removing it

.github/workflows/validations.yaml
jonasagx: done

.golangci.yaml
jonasagx: removed

Makefile
jonasagx: done

syft/pkg/cataloger/golang/version.go
jonasagx: It check whether the error is wraps a path error, looking back at it I can't justify it being there. Removing

syft/pkg/cataloger/golang/version.go
jonasagx: This func came from golang's file scan function, removing it

.golangci.yaml
jonasagx: removed

.github/workflows/golangci-lint.yml
jonasagx: we do because `bouncer` fails on on go1.17 trying to read `runtime/debug`, and it doesn't support exclusion of internal packages. The simplest is to isolate golangci on go1.17

syft/pkg/cataloger/golang/parse_go_bin.go
jonasagx: added test, and support to scoff files

test/integration/regression_go_bin_scanner_arch_test.go
wagoodman: excellent 👍

luhring:
@jonasagx This is looking great. I have two "new new" requests for change, now that I've been trying to use this new output data a bit more...

1. It'd be great to specify what the "main module" is for any given Go package, as an additional field on the Go metadata struct, whose type is `string` (to store values such as `"github.com/anchore/syft"`). ([Here's a helpful reference for how Go defines **"main module"**.](https://go.dev/ref/mod#glos-main-module)) This way, if there's some information we're only capturing on the Go package **_for the main module_**, a user can be loooking at a _non-main module_ (i.e. a dependency of that module) and find the right main module, which lets them find those additional bits of data.

2. I think we should set the `h1Digest` field on the metadata struct to `omitempty`. I think this would help the output be just a bit more clear — if we have an `h1Digest` value we can report on, we show it, and if not, we don't mention it. So with this, we'd never expect to see an empty string value for the field.

And to clarify, neither of these two requests are on the changes you've made in this PR — just additional polish I think we should achieve as we further our support of the Go ecosystem via this PR!

syft/pkg/cataloger/golang/test-fixtures/archs/Makefile
```diff
@@ -0,0 +1,8 @@
+GO_1_17 = /Users/jonas/.gvm/gos/go1.17.8/bin/go
```
wagoodman, line 1: this is a hardcoded value to your user. If the problem is needing to build with specific versions of go, we should probably containerize these steps.

syft/pkg/cataloger/golang/test-fixtures/archs/Makefile
```diff
@@ -0,0 +1,8 @@
...
+
+build:
+	GOARCH=arm64 GOOS=darwin $(GO_1_17) build -o hello-mach-o-arm64 main.go
```
wagoodman, line 4: ideally we don't want to capture these in the git source tree, instead, keeping these in gitignored cache directories that CI can restore on an as needed basis. If the cache is not in place then the makefile during test setup could be invoked to populate the cache.

syft/pkg/cataloger/golang/test-fixtures/archs/main.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+func main() {
+	println("свобода!")
```
wagoodman, line 4: ❤️

.golangci.yaml
```diff
@@ -43,7 +43,7 @@ linters:
     - varcheck
     - whitespace
 
-# do not enable...
+      # do not enable...
```
wagoodman, line 46: nit: remove added whitespace

wagoodman (commented):
For other reviewers, here's the diff for the json schema updates:
```diff
# $ diff schema/json/schema-3.1.0.json schema/json/schema-3.1.1.json
432a433,440
>         "goBuildSettings": {
>           "patternProperties": {
>             ".*": {
>               "type": "string"
>             }
>           },
>           "type": "object"
>         },

```

syft/pkg/cataloger/golang/internal/xcoff/file.go
```diff
@@ -0,0 +1,688 @@
+// Copyright 2018 The Go Authors. All rights reserved.
```
wagoodman, line 1: nit: mind adding a comment at the top of source files copied in from the stdlib that they came from there? I think maybe even adding a new readme under `golang/internal` that explains context here would be great

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -1,26 +1,53 @@
...
-const (
-	packageIdentifier = "dep"
-	replaceIdentifier = "=>"
+const GOARCH = "GOARCH"
+
+var (
+	// errUnrecognizedFormat is returned when a given executable file doesn't
+	// appear to be in a known format, or it breaks the rules of that format,
+	// or when there are I/O errors reading the file.
+	errUnrecognizedFormat = errors.New("unrecognized file format")
```
wagoodman, line 27: this isn't a very helpful error message; we should convey more context here (what was attempted to be parsed? what was expected? is there another error that we should be wrapping here?)

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -1,26 +1,53 @@
...
 )
 
-type exeOpener func(file io.ReadCloser) ([]exe, error)
+func makeGoMainPackage(mod *debug.BuildInfo, arch string, location source.Location) pkg.Package {
+	gbs := getBuildSettings(mod.Settings)
+	main := newGoBinaryPackage(&mod.Main, mod.GoVersion, arch, location, gbs)
+	main.Version = ""
+
+	if v, ok := gbs["vcs.revision"]; ok {
+		main.Version = v
```
wagoodman, line 36: _not a comment to change anything on this PR, just a thought_: we could toy with changing the version to be in the format of `v0.0.0-<date>-<commit>` if we could derive the date correctly from the build info (which I'm not certain is possible). This would make it consistent to go module version formatting.

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +67,121 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
-	// bubbling up and halting execution. For this reason we try to recover from any panic and return an error.
-	defer func() {
-		if r := recover(); r != nil {
-			err = fmt.Errorf("recovered from panic while parse go binary at %q: %+v", location.RealPath, r)
+// getArchs finds a binary architecture by two ways:
+// 1) reading build info from binaries compiled by go1.18+
+// 2) reading file headers from binaries compiled by < go1.18
+func getArchs(readers []io.ReaderAt, builds []*debug.BuildInfo) []string {
+	if len(readers) != len(builds) {
+		log.Errorf("golang cataloger: bin parsing: number of builds and readers doesn't match")
```
wagoodman, line 75: should be a `log.Warnf`

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +67,121 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
+		return nil
+	}
+
+	archs := make([]string, len(builds))
+	for i, build := range builds {
+		archs[i] = getGOARCH(build.Settings)
+	}
+
+	// if architecture was found via build settings return
+	if archs[0] != "" {
```
wagoodman, line 85: this accesses the index without checking the length, which could be 0 (thus panic)

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +67,121 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
+		return archs
+	}
+
+	for i, r := range readers {
+		a, err := getGOARCHFromBin(r)
+		if err != nil || a == "" {
+			log.Warnf("golang cataloger: bin parsing: getting arch from binary: %v", err)
```
wagoodman, line 92: we're logging the error even though the `if` case may show that as `nil`. We should have two different log cases for this instead.

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -39,57 +67,121 @@ func newGoBinaryPackage(name, version, h1Digest, goVersion, architecture string,
...
+	}
 
-		if fields[0] == packageIdentifier {
-			pkgsSlice = append(pkgsSlice, newGoBinaryPackage(name, version, h1Digest, goVersion, arch, location))
-		}
-		if fields[0] == replaceIdentifier {
-			// replace the previous entry in the package slice
-			pkgsSlice[len(pkgsSlice)-1] = newGoBinaryPackage(name, version, h1Digest, goVersion, arch, location)
-		}
+	var empty debug.Module
```
wagoodman, line 178: nit: this case isn't obvious when it would happen, a comment would be nice here to convey context to other developers

syft/pkg/cataloger/golang/scan_bin.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		builds = append(builds, bi)
+	}
+
+	archs := getArchs(readers, builds)
+
+	return builds, archs
+}
+
+// openExe opens file and returns it as io.ReaderAt.
```
wagoodman, line 47:
```suggestion
// getReaders extracts one or more io.ReaderAt objects representing binaries that can be processed (multiple binaries in the case for multi-architecture binaries).
```

syft/pkg/cataloger/golang/scan_bin.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		log.Warnf("golang cataloger: opening binary: %v", err)
+		return nil, nil
+	}
+
+	var builds []*debug.BuildInfo
+	for _, r := range readers {
+		bi, err := buildinfo.Read(r)
+		if err != nil {
+			log.Warnf("golang cataloger: scanning file %s: %v\n", filename, err)
```
wagoodman, line 36:
```suggestion
			log.Warnf("golang cataloger: scanning file %s: %v", filename, err)
```

syft/pkg/cataloger/golang/test-fixtures/archs/README.md
```diff
@@ -0,0 +1,4 @@
+The assets in this folder come from two places:
+
+* Golang's internal test data
```
wagoodman, line 3: this isn't true within the `archs` directory, but is true within the `test-fixtures` directory, right?

wagoodman (approved):
I only had `nit`-level comments at this point (and one potential panic comment). 

Nice work!! Just in time for the go 1.18 release too 🚀 😎

syft/pkg/cataloger/golang/internal/xcoff/file.go
jonasagx: done

syft/pkg/cataloger/golang/parse_go_bin.go
jonasagx: right, I am now wrapping the returned error with more context and only using this error when no other info is available.

syft/pkg/cataloger/golang/parse_go_bin.go
jonasagx: fixed

syft/pkg/cataloger/golang/parse_go_bin.go
jonasagx: fixed

PR #832, opened 2022-02-17
Npm purl namespace decode error on dir target

houdini91:

Seems that when running on a directory target the NPM purls does not split the namespace correctly.
For example a npm package `@aws-sdk/credential-provider-ini` version 3.41.0
 
**Dir target:**
PURL = pkg:npm/@aws-sdk%2Fcredential-provider-ini@3.41.0
**Image target**
PURL = pkg:npm/@aws-sdk/credential-provider-ini@3.41.0

Ps: i also think the `@` included in namespace or name be encoded (%40) if i am not mistaken.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/anchore/go-rpmdb v0.0.0-20210914181456-a9c52348da63
 	github.com/anchore/go-testutils v0.0.0-20200925183923-d5f45b0d3c04
 	github.com/anchore/go-version v1.2.2-0.20200701162849-18adb9c92b9b
-	github.com/anchore/packageurl-go v0.0.0-20210922164639-b3fa992ebd29
+	github.com/anchore/packageurl-go v0.1.1-0.20220314153042-1bcd40e5206b
 	github.com/anchore/stereoscope v0.0.0-20220307154759-8a5a70c227d3
 	github.com/antihax/optional v1.0.0
 	github.com/bmatcuk/doublestar/v4 v4.0.2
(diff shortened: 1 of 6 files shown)


houdini91:
@wagoodman @spiffcs 
Hi guys.
I am guessing build fails because `main` is broken?

kzantow (approved):
Awesome; thanks for this!

houdini91:
@kzantow 
Hi i read up on packageurl spec.
 [https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst](url)
 It seems the name should be extracted after the last `/` . 
 I think better code would be 
 
```
	lastSepIndex := strings.LastIndex(p.Name, "/")
	if lastSepIndex != -1 {
		namespace = p.Name[:lastSepIndex]
		name = p.Name[lastSepIndex+1:]
	}
```

PS: the reason i used `fields := strings.SplitN(p.Name, "/", 2)` in the current code snippet from `/home/mikey/scribe/syft/syft/pkg/npm_package_json_metadata.go` but i think it should be changed as well.
Any thoughts?

houdini91:
Hi @kzantow 
I read a bit on `https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst`
It seems the name is found after the last `\` in the purl.
Should the npm name be parsed in the same manner?
```

		lastSepIndex := strings.LastIndex(p.Name, "/")
		if lastSepIndex != -1 {
			namespace = p.Name[:lastSepIndex]
			name = p.Name[lastSepIndex+1:]
		}
```
Instead of
```
		fields := strings.SplitN(p.Name, "/", 2)
		if len(fields) > 1 {
			namespace = fields[0]
			name = fields[1]
		}
```
Any thoughts?

kzantow:
> `PURL-SPECIFICATION.rst` It seems the name is found after the last `\` in the purl

By this you mean `/`, I assume?

> Should the npm name be parsed in the same manner?

Yeah, agree with this. Even if there any cases more path segments -- something like `@org/something/something`, I'd _think_ the `@org/something` should be treated as the namespace.

kzantow:
After reading the PURL spec a bit more, I think we will always want to try splitting the name to namespace/name as you suggested. I think this is perhaps not the right spot to fix the underlying issue, however -- we should not be encoding these differently and need to figure out why this is happening

houdini91:
Ok i get that, what ever you think is best.
This PR was pushed as a simple patch, your may be right there is a underlining problem that needs fixing else where. (you may close this PR if were going a different route). 

Here our my thoughts
This PR tackles how `syft` splits the NPM name read from the package.json.
I am not sure  why  the  directory and image target use different libraries to  parse the name. but they both should do the same thing  at the very least.
(i am not sure if all package managers names need splitting in the same manner but it is a possibility). 

https://github.com/anchore/packageurl-go/pull/4(url) Will mitigate some of the encoding issues but i am not sure what parsing issues may arise when providing problematic  `name` and `namespace`  to  `NewPackageURL` in explicit manner.
For example name including a `/`.

kzantow:
@houdini91 I'm circling back to this -- is it okay to push a couple updates here?

houdini91:
Of course no problem you may commit . 
Do you want me to try and rebase the branch?

PR #834, opened 2022-02-17
Fix panic in reqs txt parsing

### LINKED ISSUE #831: scan crash with panic: runtime error: index out of range [1] with length 1 when parsing invalid formatted requirements.txt file
**What happened**:
When trying to parse invalid formatted requirements.tst file, syft crashed with the following stack trace
panic: runtime error: index out of range [1] with length 1

goroutine 83 [running]:
github.com/anchore/syft/syft/pkg/cataloger/python.parseRequirementsTxt({0xc000289ade, 0x1050368}, {0x7f3d99e3b3d8, 0xc0002ce9f0})
	/home/ec2-user/syft/syft/pkg/cataloger/python/parse_requirements.go:48 +0x50c
github.com/anchore/syft/syft/pkg/cataloger/common.(*GenericCataloger).Catalog(0xc0003b42e0, {0x106c7a8, 0xc0007a96c0})
	/home/ec2-user/syft/syft/pkg/cataloger/common/generic_cataloger.go:51 +0x23e
github.com/anchore/syft/syft/pkg/cataloger.Catalog({0x106c7a8, 0xc0007a96c0}, 0x0, {0xc00017ea80, 0xc, 0x0})
	/home/ec2-user/syft/syft/pkg/cataloger/catalog.go:55 +0x3bf
github.com/anchore/syft/syft.CatalogPackages(0xc0001367e0, {{0x0, 0x0, {0xeec274, 0x0}}})
	/home/ec2-user/syft/syft/lib.go:67 +0x405
github.com/anchore/syft/cmd.generateCatalogPackagesTask.func1(0xc000152160, 0x0)
	/home/ec2-user/syft/cmd/tasks.go:48 +0x68
github.com/anchore/syft/cmd.runTask(0x0, 0x0, 0x0, 0xc0000ba180, 0xc000797570)
	/home/ec2-user/syft/cmd/tasks.go:222 +0x85
created by github.com/anchore/syft/cmd.packagesExecWorker.func1
	/home/ec2-user/syft/cmd/packages.go:273 +0x4c7

My expectation is that invalid formatted requirements.txt files will be skipped, and that scan will continue

Scan a directory with invalid formatted requirements.txt. This example of content of requirements.txt cause a crash

keyring >= 4.1.1            # Minimum version 4.1.1
coverage != 3.5             # Version Exclusion. Anything except version 3.5
Mopidy-Dirble ~= 1.1        # Compatible release. Same as >= 1.1, == 1.*
Django == 2.3.4
SomeProject ==5.4 ; python_version < '3.8'
numpyNew; sys_platform == 'win32'
numpy >= 3.4.1; sys_platform == 'win32'

**Environment**:
Note: I've built it manually in order to add debug for identifying the file that cause the crash
- Output of `syft version`:
Application:     syft
Version:         0.38.0-SNAPSHOT-52d2e62
BuildDate:       2022-02-15T22:13:18Z
GitCommit:       52d2e62cdcc08967ead746820346766e12698488
GitDescription:  v0.38.0-dirty
Platform:        linux/amd64
GoVersion:       go1.17.7
Compiler:        gc

NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"

luhring:
Fixes #831

This PR fixes an "index out of range" panic that was occurring while parsing a `requirements.txt` file.

### Diving into the problem

The root cause of the panic was that the code was splitting a line on `==` and trying to access `parts[1]` without confirming that `parts` contained a second item. This was able to happen because we were removing comments from the line **in between** the check for two parts (from a split on `==`) and splitting the original line on `==` another time to find the package's name and version.

This is the line that caused the panic:

```
Mopidy-Dirble ~= 1.1 # Compatible release. Same as >= 1.1, == 1.*
```

Value of `parts` — at the initial length check:
```js
[
  "Mopidy-Dirble ~= 1.1 # Compatible release. Same as >= 1.1, ",
  "1.*",
]
```

Value of `parts` — at the access of `parts[1]`:
```js
[
  "Mopidy-Dirble ~= 1.1 ",
]
```

### Additional changes

1. There were additional cases for `requirements.txt` lines surfaced in #831 that were not being explicitly tested, so I added those cases into the test fixture. (e.g. lines with environment markers)

2. I adjusted the Python parser tests to use [go-cmp](https://github.com/google/go-cmp). I noticed we've started using this library more in our tests, and I can see why... It's much easier to interpret the diff that's reported for a failing test than with other diffing implementations I've seen, and it also means we don't need to write our own diff implementations as often (go-cmp seems to handle a large set of cases). I had originally started off this PR by writing a failing test case, but it was difficult to figure out what the actual problem was when a test failed. Once I put in go-cmp, it become significantly easier. To leave the code in a more consistent state, I adjusted the other two Python parsers that had been using `assertPackagesEqual` to instead use go-cmp.

3. The switch to go-cmp surfaced an issue where our parsing of `pipfile.lock` was nondeterministic, due to _ranging over a map_ that was decoded from JSON. I noticed that go-cmp was diffing the set of packages in an order-sensitive manner, and so `TestParsePipFileLock` would sometimes pass and sometimes fail, depending on the order of the packages returned by `parsePipfileLock`. I added sorting to `parsePipfileLock` to have the result slice have a stable sort order.

4. I refactored `parseRequirementsTxt` in an attempt to make the logical conditions of interest more explicit in the code.

### DIFF
--- a/syft/pkg/cataloger/python/parse_requirements.go
+++ b/syft/pkg/cataloger/python/parse_requirements.go
@@ -22,12 +22,11 @@ func parseRequirementsTxt(_ string, reader io.Reader) ([]*pkg.Package, []artifac
 	scanner := bufio.NewScanner(reader)
 	for scanner.Scan() {
 		line := scanner.Text()
-
-		line = strings.TrimRight(line, "\n")
+		line = trimRequirementsTxtLine(line)
 
 		switch {
-		case strings.HasPrefix(line, "#"):
-			// commented out line, skip
+		case len(line) == 0:
+			// nothing to parse on this line
 			continue
 		case strings.HasPrefix(line, "-e"):
 			// editable packages aren't parsed (yet)
@@ -62,16 +61,33 @@ func parseRequirementsTxt(_ string, reader io.Reader) ([]*pkg.Package, []artifac
 	return packages, nil, nil
 }
 
+// trimRequirementsTxtLine removes content from the given requirements.txt line
+// that should not be considered for parsing.
+func trimRequirementsTxtLine(line string) string {
+	line = strings.TrimSpace(line)
+	line = removeTrailingComment(line)
+	line = removeEnvironmentMarkers(line)
+
+	return line
+}
+
 // removeTrailingComment takes a requirements.txt line and strips off comment strings.
 func removeTrailingComment(line string) string {
-	parts := strings.Split(line, "#")
-	switch len(parts) {
-	case 1:
+	parts := strings.SplitN(line, "#", 2)
+	if len(parts) < 2 {
 		// there aren't any comments
-		return line
-	default:
-		// any number of "#" means we only want the first part, assuming this
-		// isn't prefixed with "#" (up to the caller)
-		return parts[0]
 	}
+
+	return parts[0]
+}
+
+// removeEnvironmentMarkers removes any instances of environment markers (delimited by ';') from the line.
+// For more information, see https://www.python.org/dev/peps/pep-0508/#environment-markers.
+func removeEnvironmentMarkers(line string) string {
+	parts := strings.SplitN(line, ";", 2)
+	if len(parts) < 2 {
+		// there aren't any comments
+	}
+
+	return parts[0]
 }
--- a/syft/pkg/cataloger/python/parse_requirements.go
+++ b/syft/pkg/cataloger/python/parse_requirements.go
@@ -24,34 +24,32 @@ func parseRequirementsTxt(_ string, reader io.Reader) ([]*pkg.Package, []artifac
 		line := scanner.Text()
 		line = trimRequirementsTxtLine(line)
 
-		switch {
-		case len(line) == 0:
+		if line == "" {
 			// nothing to parse on this line
 			continue
-		case strings.HasPrefix(line, "-e"):
+		}
+
+		if strings.HasPrefix(line, "-e") {
 			// editable packages aren't parsed (yet)
 			continue
-		case len(strings.Split(line, "==")) < 2:
-			// a package without a version, or a range (unpinned) which
-			// does not tell us exactly what will be installed
-			// XXX only needed if we want to log this, otherwise the next case catches it
-			continue
-		case len(strings.Split(line, "==")) == 2:
-			// remove comments if present
-			uncommented := removeTrailingComment(line)
-			// parse a new requirement
-			parts := strings.Split(uncommented, "==")
-			name := strings.TrimSpace(parts[0])
-			version := strings.TrimSpace(parts[1])
-			packages = append(packages, &pkg.Package{
-				Name:     name,
-				Version:  version,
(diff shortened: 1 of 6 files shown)


syft/pkg/cataloger/python/parse_requirements.go
```diff
@@ -22,37 +22,34 @@ func parseRequirementsTxt(_ string, reader io.Reader) ([]*pkg.Package, []artifac
...
+		if !strings.Contains(line, "==") {
+			// a package without a version, or a range (unpinned) which does not tell us
+			// exactly what will be installed.
 			continue
 		}
+
+		// parse a new requirement
+		parts := strings.Split(line, "==")
+		name := strings.TrimSpace(parts[0])
+		version := strings.TrimSpace(parts[1])
```
kzantow, line 46: would there be a case that the line is `something something ==` and still panics? the old check was `len(strings.Split(line, "==")) == 2`

kzantow (approved):
a couple small things noted, otherwise LGTM

syft/pkg/cataloger/python/parse_requirements.go
```diff
@@ -62,16 +59,37 @@ func parseRequirementsTxt(_ string, reader io.Reader) ([]*pkg.Package, []artifac
...
+
+	return parts[0]
+}
+
+// removeEnvironmentMarkers removes any instances of environment markers (delimited by ';') from the line.
+// For more information, see https://www.python.org/dev/peps/pep-0508/#environment-markers.
+func removeEnvironmentMarkers(line string) string {
+	parts := strings.SplitN(line, ";", 2)
+	if len(parts) < 2 {
+		// there aren't any comments
```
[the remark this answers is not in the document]
luhring: Good catch!! 😅  I'll update that

syft/pkg/cataloger/python/parse_requirements.go
luhring: If I'm understanding correctly, the two implementations are logically equivalent. If you split the string `"foo =="` on `==`, you'll get a slice with two elements: `["foo ", ""]`.

So I think the answer to "can it still panic" is "no"... but let me know if I'm missing something!

syft/pkg/cataloger/python/parse_requirements.go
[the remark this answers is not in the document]
luhring: Fixed in 446226f

syft/pkg/cataloger/python/parse_requirements.go
kzantow: Right, then LGTM 👍

PR #836, opened 2022-02-17
Export GitHub format

kzantow:
This adds some experimental support for a new format.


samj1912:
This looks very interesting. Is there a related issue that I can read up on to find more about the background/motivation behind the PR?

samj1912:
Related to https://github.blog/2022-02-22-github-advisory-database-now-open-to-community-contributions/?

kzantow:
@samj1912 only tangentially related to that, at this point.

internal/formats/github/format.go
```diff
@@ -0,0 +1,30 @@
...
+
+const ID sbom.FormatID = "github-1-json"
+
+func Format() sbom.Format {
+	return sbom.NewFormat(
+		ID,
+		func(writer io.Writer, sbom sbom.SBOM) error {
+			bom := toGithubModel(&sbom)
+
+			// bytes, err := json.Marshal(bom)
```
luhring, line 18: This looks like dead code, can this be removed?

internal/formats/github/format.go
```diff
@@ -0,0 +1,30 @@
+package github
+
+import (
+	"encoding/json"
+	"io"
+
+	"github.com/anchore/syft/syft/sbom"
+)
+
+const ID sbom.FormatID = "github-1-json"
```
luhring, line 10: Does "1" represent the schema version here? Curious if this aligns with a named schema version they've provided

internal/formats/github/github_dependency_api.go
```diff
@@ -0,0 +1,78 @@
...
+const (
+	DependencyRelationshipDirect   DependencyRelationship = "direct"
+	DependencyRelationshipIndirect DependencyRelationship = "indirect"
+)
+
+// DependencyScope A notation of whether the dependency is required for the primary
+// build artifact (runtime), or is only used for development.
+// Future versions of this specification may allow for more granular
+// scopes, like `runtimeserver`, `runtimeshipped`,
+// `developmenttest`, `developmentbenchmark`.
```
luhring, line 56: Is this a system we came up with, or is this coming from GitHub?

internal/formats/github/encoder.go
```diff
@@ -0,0 +1,154 @@
...
+		path := toPath(s.Source, p, false)
+		manifest, ok := manifests[path]
+		if !ok {
+			manifest = &Manifest{
+				Name: path,
+				File: FileInfo{
+					SourceLocation: path,
+				},
+				Metadata: Metadata{
+					"syft:path": toPath(s.Source, p, true),
```
luhring, line 93: The conditionality of `toPath` with the `full` parameter is making this function a bit more difficult for me to understand... I was wondering about what to do about it, but then a question occurred to me — might it makes sense to include the filesystem ID as a separate metadata field here, rather than included it somehow in a single string with the path?

internal/formats/github/format.go
kzantow: It was intended to... but they are actually at version 0 🤦

internal/formats/github/github_dependency_api.go
kzantow: This came directly from their typescript definitions

internal/formats/github/encoder.go
kzantow: agree, originally I was trying to use this as the `file/source_location` but there are currently some issues with that and I don't think that will be viable in the near term

internal/formats/github/encoder.go
```diff
@@ -0,0 +1,158 @@
...
+func toPath(s source.Metadata, p pkg.Package) string {
+	if len(p.Locations) > 0 {
+		location := &p.Locations[0]
+		switch s.Scheme {
+		case source.ImageScheme:
+			in := strings.ReplaceAll(s.ImageMetadata.UserInput, ":/", "//")
+			path := strings.TrimPrefix(location.RealPath, "/")
+			return fmt.Sprintf("%s:/%s", in, path)
+		case source.FileScheme:
+			return fmt.Sprintf("%s:/%s", s.Path, strings.TrimPrefix(location.RealPath, "/"))
```
luhring, line 75: I was just playing with this, and when I scan individual files, I get this kind of output:

```json
"name": "/Users/dan/homebrew/bin/syft:/Users/dan/homebrew/Cellar/syft/0.41.1/bin/syft",
```

What is the motivation for this string format? I'm wondering if it's easier to read to just include the real path or something, WDYT?

kzantow: The motivation for this was like:
```
/some/tar/archive.tar.gz:/some/path/within/it
```
... but maybe that is really not what is happening here, I can definitely change this... but I think this is actually illustrating one annoyance with our different resolvers: image resolvers have a path relative to the image... we should have paths relative to the directory scanned here instead of full paths

luhring: I like this scheme for archives (like TAR) FWIW!

For other things, I agree that it's a bit awkward.

luhring (approved):

PR #850, opened 2022-02-25
Cpe generation update

spiffcs:
Update CPE generation to return `nil` instead of `empty`.

When Grype takes an sbom generated by syft, an `empty` CPE will throw an error causing matches to not be returned.

This change makes it so that empty cpe can no longer be possible in the syft SBOM output.


syft/pkg/cataloger/common/cpe/generate.go
```diff
@@ -8,18 +8,23 @@ import (
...
+func newCPE(product, vendor, version, targetSW string) *wfn.Attributes {
 	cpe := *(wfn.NewAttributesWithAny())
 	cpe.Part = "a"
 	cpe.Product = product
 	cpe.Vendor = vendor
 	cpe.Version = version
 	cpe.TargetSW = targetSW
-	return cpe
+	if pkg.ValidateCPEString(pkg.CPEString(cpe)) != nil {
+		log.Warnf("could not generate CPE for %+v", cpe)
```
kzantow, line 24: Should this be a warning? Probably `Debugf` since we'll be using this function in multiple places from multiple SBOM inputs

kzantow (approved):
LGTM although I'm worried there are cases that using a warning will result in a lot of log noise

syft/pkg/cataloger/common/cpe/generate.go
spiffcs: Here is a screenshot what the output looks like with the warning. I lean slightly towards keeping this as a warning since I think this is a category of alerting the user that something went wrong vs a procedural step like `cataloging with "ruby-gemspec-cataloger"`

Screen Shot 2022-02-25 at 2 19 02 PM

syft/pkg/cataloger/common/cpe/generate.go
```diff
@@ -8,18 +8,23 @@ import (
 	"strings"
 
 	"github.com/anchore/syft/internal"
+	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/facebookincubator/nvdtools/wfn"
 )
 
-func newCPE(product, vendor, version, targetSW string) wfn.Attributes {
+func newCPE(product, vendor, version, targetSW string) *wfn.Attributes {
```
kzantow, line 16: A second look -- I'm a little confused here, shouldn't we be able to generate a valid CPE all the time given these arguments? Even if we need to encode them a certain way?

spiffcs: Vendor, product, and other fields can contain special characters:
```
"david rodríguez",
"andré medeiros",
```

So the following CPE that we currently allow through are actually invalid according the to specification:
```
     19     "cpe:2.3:a:david-rodríguez:bundler:2.1.4:*:*:*:*:*:*:*",
     20     "cpe:2.3:a:david_rodríguez:bundler:2.1.4:*:*:*:*:*:*:*",
     21     "cpe:2.3:a:andré-medeiros:bundler:2.1.4:*:*:*:*:*:*:*",
     22     "cpe:2.3:a:andré_medeiros:bundler:2.1.4:*:*:*:*:*:*:*",
```

Adding `ValidateCPEString` allows us to short circuit any invalid strings we might generate.

syft/pkg/cataloger/common/cpe/generate.go
kzantow: My concern is more... let's say we import SPDX or CycloneDX that has thousands of components and thousands of CPEs, we'll see this error a bunch of the tool used to generate said document isn't doing a good job, even though we might be able to still provide some useful things like Grype vuln matching with generated CPEs

syft/pkg/cataloger/common/cpe/generate.go
spiffcs: When grype takes an sbom as an input it will try to decode those CPE. This will throw an error and make it so that vulnerabilities are not returned for the given package it failed to parse CPE for.

syft/pkg/cataloger/common/cpe/generate.go
kzantow: ... which leans to my second comment, can't we just encode these somehow? https://github.com/anchore/syft/pull/850#discussion_r815033319

syft/pkg/cataloger/common/cpe/generate.go
kzantow: That's not entirely true. If we can still generate CPEs based on the package information, we could still return accurate matches in Grype.

spiffcs: This does not shortcircuit all CPE generation. We are still generating package information CPE.

See how in the test case we still expect the other noninvalid utf8 character cpes to be generated by a package.
```
				"cpe:2.3:a:*:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:bundler:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:ruby-lang:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:ruby:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:ruby_lang:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:jessica-lynn-suttles:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:jessica_lynn_suttles:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:stephanie-morillo:bundler:2.1.4:*:*:*:*:*:*:*",
				"cpe:2.3:a:stephanie_morillo:bundler:2.1.4:*:*:*:*:*:*:*",
```

kzantow: So based on CPE spec:
> Value strings assigned to attributes of WFNs SHALL be non-empty contiguous strings of bytes encoded
using printable Unicode Transformation Format-8 (UTF-8) characters with hexadecimal values between
x00 and x7F [RFC3629].

This tells me, we _cannot_ encode this and these characters are _invald for CPEs_. So maybe just check the inputs for any characters outside the range?

spiffcs: > That's not entirely true. If we can still generate CPEs based on the package information, we could still return accurate matches in Grype.<

This is true and it's what is causing https://github.com/anchore/grype/pull/645 to fail on the integration tests.

In previous syft version the invalid CPE were not being generated. In the current version (not including this PR), invalid CPE are being included in the `syft.json` output. 

When grype gets to this part of the code an error is returned rather than a list of `Match`
https://github.com/anchore/grype/blob/3a22a56d114354984d06c9cc1c94cf53ccc7b919/grype/search/criteria.go#L26-L31

This causes vulnerabilities to not be included in the final report when the SBOM input is fed in.

Vulnerabilities are still included when scanning an image since grype never has to decode information the way it does  when an SBOM is passed.

spiffcs: > 

That is what the function `ValidateCPEString` does.

spiffcs: That function is also what's used during grype when `ByPackageCPE` so we can be sure our output is sanitized in the same way we are checking input when grype is checking a `syftjson` SBOM

kzantow: So `ValidateCPEString` validates a _string_ against a regex, I get it. This isn't bad, it's just a lot of intermediate steps that are unnecessary when one can just validate the characters for each of the inputs. Anyway, as noted I don't think that should hold up this PR. 👍

spiffcs: It's a good idea to check all the inputs, but I'm hesitant on adding more checks into this process. We have already have `sanitize` and `ValidateCPEString` and I do not totally understand the difference between them other than this comment. 

```
// sanitize is a modified version of WFNize function from nvdtools
// that quotes all the allowed punctation chars with a slash and replaces
// spaces with underscores. It differs from the upstream implmentation as
// it does not use the buggy nvdtools implementation, specifically the "addSlashesAt" part of the
// function which stops the loop as soon as it encounters ":" a valid
// character for a WFN attribute after quoting, but the way nvdtools
// handles it causes it to truncate strings that container ":". As a result
// strings like "prefix:1.2" which would have been quoted as "prefix\:1.2"
// end up becoming "prefix" instead causing loss of information and
// incorrect CPEs being generated.
```

Based on that comment I'm not sure sanitize is the right spot to add more character checks. 

I'm up for changing the validation process if `ValidateCPEString` is a lot of intermediate steps that are unnecessary @wagoodman what do you think?

wagoodman: >  it's just a lot of intermediate steps that are unnecessary when one can just validate the characters for each of the inputs

agreed on this being a lot of steps, and there is room for improvement, but I think it's the right call for the scope of this PR. This PR is only meant to address the bad behavior being seen with empty CPEs (thus dropping invalid characters and keeping generated CPEs within spec... which is consistent with todays behavior). We won't be able to fix the shortcomings of the facebook CPE tooling in this PR

Using the existing `ValidateCPEString ` check gives us assurances that the eventual encoded string will be valid relative to the CPE spec for formatted strings without having to write more CPE helpers relative to the spec... (I'd rather use that time to swap out the CPE lib entirely since we know it's non-compliant).

syft/pkg/cataloger/common/cpe/generate.go
wagoodman: I could go either way on this (leave it or take it out). Taking it out is alright since its warning about synthetic data problems and not from data we found in the source scanned.

wagoodman (approved):

PR #858, opened 2022-03-02
Reduce PR check failures

wagoodman:
- Adds polling check for install tests to ensure the snapshot server is up
- Elevates CodeQL to only run on main (not on PRs)
- De-duplicates checks as best as possible; CI checks only run on PRs and pushes to main. Pushes to feature branches will no longer trigger CI.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -7,10 +7,10 @@ name: "CodeQL Security Scan"
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


wagoodman:
I have a parallel PR going that shows the behavior from a fork: https://github.com/anchore/syft/pull/859 (not intended to be merged)

kzantow (approved):
BASH magic there :)

PR #864, opened 2022-03-03
Migrate format definitions to sbom package

wagoodman:
This PR exposes more functionality at the top-level API with regards to formats, moving us towards the direction mentioned in https://github.com/anchore/syft/issues/558#issuecomment-1058162562 . 

The changes made:
- removes the `syft/format` package entirely... moving many of the definitions to `syft/sbom`. Why? Many of the definitions here don't have semantics that stand on their own (for instance, format of what?). The extra context of `sbom` helps with understanding what the format is for.
- migrates format functionality (behavior) and definitions relating to capabilities specific to syft to the `syft` package (what was`format.Option`, `format.Identify`, etc.). It is important to distinguish syft capabilities separate from the underlying definition of what a format is. That is, the definition of a format (an object that has encoding, decoding, validation relative to a bag of bytes and an sbom object) lives separately from the user-options to select formats that happen to be implemented by syft (e.g. `format.CycloneDxXMLOption`)
- Un-exports the `format` object (that now resides in the `sbom` package) and introduces a `Format` interface. The constructor for the unexported `format` object remains as an easy to to assemble common functionality across format implementations. The interface is useful in the sense that the abstraction can be used with more flexibility in the top-level api (someone can bring their own format without having to fit it into the same functionality that the `format` struct provides).
- Formats are decoupled from `format.Options` and instead can take one or more names that represents the format. This is part of the effort of decoupling syft capabilities and format abstractions --we shouldn't have the abstraction return a `format.Option` where syft controls the full set of valid formats (somewhat defeating the purpose of abstracting it). (caveat: though the option is a string type, it would force folks to do weird stuff like `format.Options(myFormatName)`, which is not useful)
- Migrates the implementations for `sbom.Writer`s out to `sbom` (instead of in the cmd pkg) as there are now folks that want to use this externally.

Summary of Top level `syft` API changes:
```golang
// migrated from today's "format" package
func FormatByOption(option FormatOption) sbom.Format 

// added as a helper since 
func FormatByName(name string) sbom.Format

// migrated from today's "format" package
func IdentifyFormat(by []byte) sbom.Format

func Encode(s sbom.SBOM, f sbom.Format) ([]byte, error)

func Decode(reader io.Reader) (*sbom.SBOM, sbom.Format, error)
```


internal/formats/cyclonedx13xml/format.go
```diff
@@ -3,14 +3,14 @@ package cyclonedx13xml
...
 
-func Format() format.Format {
-	return format.NewFormat(
-		format.CycloneDxXMLOption,
+func Format(names ...string) sbom.Format {
+	return sbom.NewFormat(
 		encoder,
 		cyclonedxhelpers.GetDecoder(cyclonedx.BOMFileFormatXML),
 		cyclonedxhelpers.GetValidator(cyclonedx.BOMFileFormatXML),
+		append(names, "cyclonedx-xml", "cyclone-xml", "cyclonedx")...,
```
kzantow, line 14: I'm confused by this -- so if I specify "cyclonedx", I might get both XML and JSON formats?

syft/encode_decode.go
```diff
@@ -28,19 +20,17 @@ func Encode(s sbom.SBOM, option format.Option) ([]byte, error) {
 }
 
 // Decode takes a reader for an SBOM and generates all internal SBOM elements.
-func Decode(reader io.Reader) (*sbom.SBOM, format.Option, error) {
+func Decode(reader io.Reader) (*sbom.SBOM, sbom.Format, error) {
```
kzantow, line 23: 👍

syft/formats.go
```diff
@@ -0,0 +1,77 @@
...
+				return f
+			}
+		}
+	}
+	return nil
+}
+
+func IdentifyFormat(by []byte) sbom.Format {
+	for _, f := range formats {
+		if err := f.Validate(bytes.NewReader(by)); err != nil {
```
kzantow, line 71: Nothing that will hold up this PR or even is relevant for the change here; just noting I think we we should revisit using Validate to identify a format -- when an error occurs, in what might be an otherwise valid file the user never will get notified what is wrong, this has come up just recently when someone tried to use SPDX as input but the file was invalid...

syft/formats.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+func FormatByOption(option FormatOption) sbom.Format {
+	return FormatByName(string(option))
+}
+
+func FormatByName(name string) sbom.Format {
+	for _, f := range formats {
+		for _, formatName := range f.Names() {
```
kzantow, line 60: So here... if I specify name = "spdx", I will get back the SPDXTagValueFormatOption only because it's first in the list; maybe we should just remove the ambiguity? I see now,  as it's written everything will continue to work correctly, though.

test/integration/encode_decode_cycle_test.go
```diff
@@ -2,14 +2,14 @@ package integration
 
 import (
 	"bytes"
+	"github.com/stretchr/testify/require"
```
kzantow, line 5: :sadpanda:

kzantow (approved):
As-is I think this works; a few comments -- the main one being the ambiguity between the "spdx" and "cyclonedx" names being used for two formats. It appears this will work because of the ordering of the format array, but it's not especially obvious this is the case if someone decides to add more cyclonedx formats, for example, and orders them alphabetically...

wagoodman:
I received some offline feedback I want to try and incorporate. The summary is attempting to resolve Formats by IDs instead of a collection of names would be ideal.

internal/formats/cyclonedx13xml/format.go
wagoodman: almost, but this is about to change... a commit incoming pushes aliases for formats into the `syft` package and the IDs for formats will reside with the format object itself, making this handling more obvious.

syft/formats.go
wagoodman: A validator is meant to drive this identify function, and not necessarily validate the entire document. Take for instance the syft json validator:
```golang
func validator(reader io.Reader) error {
	type Document struct {
		Schema model.Schema `json:"schema"`
	}

	dec := json.NewDecoder(reader)

	var doc Document
	err := dec.Decode(&doc)
	if err != nil {
		return fmt.Errorf("unable to decode: %w", err)
	}

	// note: we accept all schema versions
	if strings.Contains(doc.Schema.URL, "anchore/syft") {
		return nil
	}
	return fmt.Errorf("could not extract syft schema")
}

```
If this is invalid json, we bail. If there is no schema block, we bail. If the schema URL doesn't reference anchore/syft, we bail. No other fields are referenced... and this is intentional.

Per https://github.com/anchore/syft/blob/main/syft/format/validator.go#L4-L12 a `Validator` is intentionally scoped as small as possible (for now).

> I think we we should revisit using Validate to identify a format 

I agree. I think maybe `Validate` should be changed to a name that more clearly indicates it's being used to identify the format. Maybe something like `IdentifiesAs([]bytes) bool` would be better (in which case this is an easy change to the format interface). We can later add full-format validators and use `Validate([]byte) error` for that.

Agreed on maybe a follow up PR.

syft/formats.go
wagoodman: yeah, but again, this will change with the latest commit.

test/integration/encode_decode_cycle_test.go
wagoodman: :sadderpanda:

wagoodman:
A change is still needed --with the format IDs being in `internal` they are not useful for external folks. I'll move these to the `syft` package.

wagoodman:
The downside with having the IDs drive the help output is that we can't prefer showing aliases to the user. Here's an excerpt from the help output for `-o`:
```
  -o, --output stringArray    report output format, options=[syft-3-json cyclonedx-1-xml cyclonedx-1-json spdx-2-tag-value spdx-2-json syft-table syft-text] (default [syft-table])

```

wagoodman:
Here's an example help output with the alias changes:
```
  -o, --output stringArray         report output format, options=[syft-json cyclonedx-xml cyclonedx-json spdx-tag-value spdx-json table text] (default [table])

```

PR #866, opened 2022-03-04
Add platform selection

wagoodman:
Adds `--platform` option to `packages`, `attest`, and root commands, allowing someone to specify values similar to `docker pull`, for example:

- os only override: `windows`
- architecture override: `arm64`
- combination: `linux/arm`
- architecture variants: `arm/v8`
- all options: `windows/arm/v6`

Additionally if you attempt to specify `--platform` on non-image sources you will receive an error.

cc: @errordeveloper


syft/source/source.go
```diff
@@ -147,10 +153,19 @@ func getImageWithRetryStrategy(in Input, registryOptions *image.RegistryOptions)
 		opts = append(opts, stereoscope.WithRegistryOptions(*registryOptions))
 	}
 
+	if in.Platform != "" {
```
spiffcs, line 156: Nice defaults

spiffcs (approved):

PR #868, opened 2022-03-07
Update yarn.lock parser to support latest (berry v3) format

cipher-ardvark:
Update the regex for the javascript yarn.lock parser to support the latest
version of Yarn (at the time v3) which has changed the format substantially.
Added more unit test for the regex based parsers to ensure future changes
do not break any existing formats.


wagoodman (approved):
thanks for upgrading our yarn support 🎉 ! (I only made a couple of minor changes to the test additions)

cipher-ardvark:
Thanks for the code cleanup, I did my best to take your changes and sign my old commits. Let me know if there is anything else you would like changed.

PR #876, opened 2022-03-09
Add artifacthub owner

wagoodman:
Config is documented here: https://github.com/artifacthub/hub/blob/v1.6.0/docs/metadata/artifacthub-repo.yml

This PR is meant only to establish ownership, a future PR will add the repo ID and more owners.


luhring (approved):
👍  (noted as a temporary solution)

PR #877, opened 2022-03-09
Add SchemaVersion to version command output

spiffcs:
See attachment for new output and sample command updates

Screen Shot 2022-03-09 at 10 28 28 AM

Signed-off-by: Christopher Phillips <christopher.phillips@anchore.com

### DIFF
--- a/internal/version/build.go
+++ b/internal/version/build.go
@@ -22,14 +22,14 @@ var platform = fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
 
 // Version defines the application version details (generally from build information)
 type Version struct {
-	Version        string `json:"version"`        // application semantic version
-	SchemaVersion  string `json:"schemaVersion"`  // application semantic JSON schema version
-	GitCommit      string `json:"gitCommit"`      // git SHA at build-time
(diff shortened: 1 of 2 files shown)


internal/version/build.go
```diff
@@ -21,6 +23,7 @@ var platform = fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
 // Version defines the application version details (generally from build information)
 type Version struct {
 	Version        string `json:"version"`        // application semantic version
+	SchemaVersion  string `json:"schemaVersion"`  // application semantic JSON schema version
```
wagoodman, line 26: nit: maybe `jsonSchemaVersion` in case we end up adding schemas for other formats in the future?

wagoodman (approved):
Thanks for the addition here --added a note on a rename, but shout out if you think that doesn't work and we can keep chatting 👍

spiffcs:
> Thanks for the addition here --added a note on a rename, but shout out if you think that doesn't work and we can keep chatting 👍

Yep! Let me get the formatting fixed with that new rename and I'll push the changes here.

spiffcs:
Screen Shot 2022-03-09 at 11 00 21 AM

I know it looks weird in the editor, but the spacing like that comes out correctly on the terminal

PR #880, opened 2022-03-09
RPM Epoch should be optional in the json schema

wagoodman:
Today the jsonschema denotes that epoch is not optional --this is not correct. An epoch of 0 does not necessarily mean the same thing as no epoch, which is why the RPM Metadata carefully tracks `*int` for this field. This PR adjusts the json schema to match reality, allowing nullable values for the `epoch` field within the `RpmdbMetadata` definition. This has some questions on historical data access so the `revision` was bumped over the `addition` field.

Additionally, since last this schema was generated NPM package JSON now additionally captures name and version, however, since additional values are allowed this is not a breaking change.

Diff from the old schema to the new schema provided here for review convenience:

```diff
# diff schema/json/schema-3.0.1.json schema/json/schema-3.1.0.json
546a547,548
>         "name",
>         "version",
553a556,561
>         "name": {
>           "type": "string"
>         },
>         "version": {
>           "type": "string"
>         },
1099c1107,1114
<           "type": "integer"
---
>           "oneOf": [
>             {
>               "type": "integer"
>             },
>             {
>               "type": "null"
>             }
>           ]

```

The same diff with a little more context:
```
# diff -C 10 schema/json/schema-3.0.1.json schema/json/schema-3.1.0.json
*** schema/json/schema-3.0.1.json       Wed Feb  9 13:55:34 2022
--- schema/json/schema-3.1.0.json       Wed Mar  9 14:00:23 2022
***************
*** 537,563 ****
--- 537,571 ----
          },
          "cpeName": {
            "type": "string"
          }
        },
        "additionalProperties": true,
        "type": "object"
      },
      "NpmPackageJSONMetadata": {
        "required": [
+         "name",
+         "version",
          "author",
          "licenses",
          "homepage",
          "description",
          "url"
        ],
        "properties": {
+         "name": {
+           "type": "string"
+         },
+         "version": {
+           "type": "string"
+         },
          "files": {
            "items": {
              "type": "string"
            },
            "type": "array"
          },
          "author": {
            "type": "string"
          },
          "licenses": {
***************
*** 1089,1109 ****
          "files"
        ],
        "properties": {
          "name": {
            "type": "string"
          },
          "version": {
            "type": "string"
          },
          "epoch": {
!           "type": "integer"
          },
          "architecture": {
            "type": "string"
          },
          "release": {
            "type": "string"
          },
          "sourceRpm": {
            "type": "string"
          },
--- 1097,1124 ----
          "files"
        ],
        "properties": {
          "name": {
            "type": "string"
          },
          "version": {
            "type": "string"
          },
          "epoch": {
!           "oneOf": [
!             {
!               "type": "integer"
!             },
!             {
!               "type": "null"
!             }
!           ]
          },
          "architecture": {
            "type": "string"
          },
          "release": {
            "type": "string"
          },
          "sourceRpm": {
            "type": "string"
          },

```

cc: @rbrady @Toure

### DIFF
--- a/internal/constants.go
+++ b/internal/constants.go
@@ -6,5 +6,5 @@ const (
 
 	// JSONSchemaVersion is the current schema version output by the JSON encoder
 	// This is roughly following the "SchemaVer" guidelines for versioning the JSON schema. Please see schema/json/README.md for details on how to increment.
-	JSONSchemaVersion = "3.0.1"
+	JSONSchemaVersion = "3.1.0"
 )
--- a/internal/formats/syftjson/test-fixtures/snapshot/TestDirectoryEncoder.golden
+++ b/internal/formats/syftjson/test-fixtures/snapshot/TestDirectoryEncoder.golden
@@ -88,7 +88,7 @@
   }
  },
  "schema": {
-  "version": "3.0.1",
-  "url": "https://raw.githubusercontent.com/anchore/syft/main/schema/json/schema-3.0.1.json"
+  "version": "3.1.0",
+  "url": "https://raw.githubusercontent.com/anchore/syft/main/schema/json/schema-3.1.0.json"
  }
 }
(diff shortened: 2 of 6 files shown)


kzantow (approved):
LGTM although pretty difficult to review the entire schema file :)

PR #882, opened 2022-03-10
use filepath.Base() instead of path.Base() for temp files

Crysinyan:
`path.Base()` returns the entire input path if operating system is Windows.
this is resolved by using `filepath.Base()` instead.

using `path.Base()` produces multiple instances of "Path could not be found", due to the fact that the temporary directory:
`archivePath := filepath.Join(tempDir, "archive-"+name)`
is faulty, `name` contains the entire path of the input archive instead of the base name.
path

switching to using `filepath.Base()` resolves this issue by return the correct base name for the package.
filepath

An issue with symlinks still remains however.

It also seems like `path.Base()` and other functions from `path` are bein used in multiple places.
I unfortunately don't know enough about Go to know if these also should be changed in favor of `filepath`.

### DIFF
--- a/syft/pkg/cataloger/java/save_archive_to_tmp.go
+++ b/syft/pkg/cataloger/java/save_archive_to_tmp.go
@@ -5,14 +5,13 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 
 	"github.com/anchore/syft/internal/log"
 )
 
 func saveArchiveToTmp(archiveVirtualPath string, reader io.Reader) (string, string, func(), error) {
-	name := path.Base(archiveVirtualPath)
+	name := filepath.Base(archiveVirtualPath)
 	tempDir, err := ioutil.TempDir("", "syft-archive-contents-")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Crysinyan:
I would like to add that I don't usually create PRs or use Github that frequently.
if there is any problems with how this PR was made or if it should have been an issue instead I would welcome the input :)

kzantow:
@Crysinyan first thing is could you sign-off and sign your commit(s)? https://github.com/anchore/syft/blob/main/CONTRIBUTING.md

spiffcs:
There is some nuance here in how `path.Join` works vs `filepath` that I'm trying to remember for OSX vs Windows and our particular use case. 

Basic Gist:

Running `syft` on a windows machine but scanning a linux distro based container. We opted to use `path` so that when stereoscope and syft was building the virtual representation of the filesystem the paths would not be broken cross-platform.

I'm not totally sure how this might affect certain cases updating the `Base` function but will take some time to investigate and understand this PR since this seems to be giving us coverage that we're missing when trying to build an SBOM.

wagoodman (approved):
Thanks for the bug fix! (I've signed off on the commit on your behalf since the code delta is small and stdlib-only)

PR #883, opened 2022-03-10
Include vendored modules in Go Module package list

### LINKED ISSUE #871: `syft packages` fails to catalog golang binary's modules for binary built with vendored modules
**What happened**:
I wanted to first vendor my Golang app's dependencies using `go mod vendor`, then compile my binary, then catalog the packages in the binary using `syft packages` against the build output directory.

I expected the `artifacts` array in the `syft` output to contain the module built into the binary, as it does for non-vendored apps. Instead, the `artifacts` array was empty.

Clone [this repro sample app](https://github.com/fg-j/explorations/blob/34c7640a05be9f213ff976bb6cb42de121609f48/golang-syft-repro) and follow the instructions in the README. 

The reproduction steps are a stripped down version of what happens in the [Paketo](https://paketo.io/) Golang buildpack. I'm a project maintainer and hoping to use `syft` to generate an SBOM of Go modules for built binaries.

**Environment**:
- Output of `syft version`:
-  syft version
Application:     syft
Version:         0.40.0
BuildDate:       2022-03-02T16:44:56Z
GitCommit:       1e75cb0418df15f8bc8193b1b8a3b3a8d81cb3b0
GitDescription:  v0.40.0
Platform:        darwin/amd64
GoVersion:       go1.17.7
Compiler:        gc
  MacOS 11.16.1

fg-j:
Resolves #871

### DIFF
--- a/syft/pkg/cataloger/golang/parse_go_bin.go
+++ b/syft/pkg/cataloger/golang/parse_go_bin.go
@@ -65,23 +65,27 @@ func parseGoBin(location source.Location, reader io.ReadCloser, opener exeOpener
 func buildGoPkgInfo(location source.Location, mod, goVersion, arch string) []pkg.Package {
 	pkgsSlice := make([]pkg.Package, 0)
 	scanner := bufio.NewScanner(strings.NewReader(mod))
-
-	// filter mod dependencies: [dep, name, version, sha]
 	for scanner.Scan() {
 		fields := strings.Fields(scanner.Text())
 
-		// must have dep, name, version, sha
-		if len(fields) < 4 {
+		if len(fields) < 3 {
 			continue
 		}
 
-		if fields[0] == packageIdentifier || fields[0] == replaceIdentifier {
-			name := fields[1]
-			version := fields[2]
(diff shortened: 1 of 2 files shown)


spiffcs (approved):
Thanks for the quick fix! Looks good to me. Looking for a second from @wagoodman

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -65,23 +65,27 @@ func parseGoBin(location source.Location, reader io.ReadCloser, opener exeOpener
...
 
-		if fields[0] == packageIdentifier || fields[0] == replaceIdentifier {
-			name := fields[1]
-			version := fields[2]
-			h1Digest := fields[3]
+		name := fields[1]
+		version := fields[2]
+		h1Digest := ""
+		if len(fields) >= 4 {
+			h1Digest = fields[3]
```
wagoodman, line 79: nit: capturing the information about this case being for possible vendoring would be great --to help future devs in this file

wagoodman (approved):
nice addition 💯

PR #885, opened 2022-03-11
Brew install gon failed

kzantow:
Release is failing due to invalid brew formula. Instead, directly install `gon`.


wagoodman (approved):
we could probably try this out in a one-off CI test to see that `go install` bin is indeed in path (without having to cut a release)

spiffcs (approved):

PR #895, opened 2022-03-17
Fix panic when CycloneDX BOM missing metadata.component

kzantow:
There is a case where a community member is merging Syft-generated CycloneDX SBOMs using `cyclonedx-cli`, which results in a missing `metadata.component`, causing a panic.


luhring (approved):
LGTM!

PR #897, opened 2022-03-17
Correct go.mod to enforce go 1.18

06kellyjac:
Since syft now depends on debug/buildinfo go 1.18 is required to build syft and as such go.mod needs updating

> Go 1.18 Release Notes
> New debug/buildinfo package

<https://tip.golang.org/doc/go1.18>

Signed-off-by: 06kellyjac <jack@control-plane.io>


06kellyjac:
If you're hesitant to upgrade to 1.18 then you'll want to revert #827 for the time being because syft just wont build on 1.17

```
λ go version
go version go1.17.7 linux/amd64
λ go build .
syft/pkg/cataloger/golang/scan_bin.go:5:2: package debug/buildinfo is not in GOROOT (/nix/store/a2m476dmim68g082fcia3wrbrzhhfgmv-go-1.17.7/share/go/src/debug/buildinfo)
```

06kellyjac:
> It doesn't indicate which version should be used for building, which determines the version of the standard library used, the Go runtime, etc.

the go1.17 standard library doesn't have `debug/buildinfo`

In the exact section you link it says

> **Syntax**
> `go minimum-go-version`
> minimum-go-version
> > The minimum version of Go required to compile packages in this module.

The minimum version of Go required to compile the packages in this module is the minimum version you can use for building, and that is 1.18

luhring:
Hi @06kellyjac — you're absolutely right. Sorry about that.

What I should've said is that it's valid to build with a later version of Go than what's stated in `go.mod`, and that's what we're currently doing in Syft.

When we moved to 1.18, we did consider the `go.mod` line in question, and I think the original idea was to use this `go.mod` line as a tool-based constraint to prevent usage of 1.18's **language features** like generics while we learn more about their idiomatic use in the Go ecosystem. So one path forward here to make this more obvious would be to _add a comment to `go.mod`_ that explains why this project says Go 1.17 when it needs Go 1.18, which is otherwise weird and appears to be a mistake.

But there might be a better solution here, too. Since, like you said, this helps Go understand what compiler version is required, bumping this to Go 1.18 would also help Syft's library consumers to know about a potential breakage sooner rather than later, since Syft now could not be used as a module in projects using versions earlier than Go 1.18.

The core team would like to think on this for just a day or so and figure out what path makes the most sense. We want make sure we're not creating unnecessary problems for people.

Thank you so much for raising this and prompting this discussion!

luhring:
> Id be interested to see what options you all come up with

Hi @06kellyjac, thanks for bearing with us. 😃  We've decided we like the idea of adjusting `go.mod` to say "1.18". It honors the guidance put forth in the Go module docs, and it helps to alert Syft's library consumers earlier about potential breakage. We'll keep an eye out on PRs to understand how fast new language features are making it into the code base and handle any concerns with those language feature introductions as they arise.

Thanks so much for the PR — I'm going to approve and merge as-is! 🙏

luhring (approved):
Thanks again!

PR #898, opened 2022-03-17
Fix CycloneDX license decoding

kzantow:
Working through more CycloneDX decoding issues, another nil pointer dereference was identified.


spiffcs (approved):

spiffcs:
Small nit on the PR title - probably should be specific here and say `license` panic for speicificty

PR #900, opened 2022-03-18
Correct ID handling during syftjson decoding

kzantow:
During Syft JSON decoding, an ID is not being set by calling `pkg.SetID()`, however this may differ from the ID in the file. To fix this, we use the same semantics to call `SetID` but also track aliases where these IDs have changed in order to properly decode relationships.

This corrects the issue where users importing Syft JSON see warnings such as:

```
WARN found package with empty ID while adding to the catalog: Pkg(name="Django" version="3.2.12" type="python" id="") from-lib=syft
WARN found package with empty ID while adding to the catalog: Pkg(name="HikariCP" version="2.5.1" type="java-archive" id="") from-lib=syft
...
```

### DIFF
--- a/internal/formats/syftjson/model/package.go
+++ b/internal/formats/syftjson/model/package.go
@@ -129,6 +129,7 @@ func (p *Package) UnmarshalJSON(b []byte) error {
 		if err := json.Unmarshal(unpacker.Metadata, &payload); err != nil {
 			return err
 		}
+		p.Metadata = payload
 	default:
 		log.Warnf("unknown package metadata type=%q for packageID=%q", p.MetadataType, p.ID)
 	}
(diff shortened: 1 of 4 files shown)


spiffcs:
I was doing some debugging when reviewing this PR and noticed that the input package ID does not correspond with the package ID grype regenerates.

Input:
Screen Shot 2022-03-18 at 3 35 32 PM

ID Found in Grype for the same package:
Screen Shot 2022-03-18 at 3 35 23 PM

Should grype be resetting this or should it just be using the ID from the input?

spiffcs (approved):

spiffcs:
PR Review done over zoom! Thanks for helping catch the bug on the golang metadata decoder which caused the different ID and having it all accounted for!

PR #901, opened 2022-03-18
add case to decode BinMetadata

spiffcs:
Add case where `pkg.GolangBinMetadataType` can be accounted for during `syftjson` model package unmarshal. See screenshot before after where a warning was firing because this case was not accounted for:

After shows the fix:

Before:
Screen Shot 2022-03-18 at 1 52 40 PM

After:
Screen Shot 2022-03-18 at 1 51 25 PM

TODO:
- [ ]  Add unit coverage

### DIFF
--- a/internal/formats/syftjson/model/package.go
+++ b/internal/formats/syftjson/model/package.go
@@ -124,6 +124,11 @@ func (p *Package) UnmarshalJSON(b []byte) error {
 			return err
 		}
 		p.Metadata = payload
+	case pkg.GolangBinMetadataType:
+		var payload pkg.GolangBinMetadata
+		if err := json.Unmarshal(unpacker.Metadata, &payload); err != nil {
+			return err
+		}
 	default:
 		log.Warnf("unknown package metadata type=%q for packageID=%q", p.MetadataType, p.ID)
 	}
--- a/internal/formats/syftjson/model/package_test.go
+++ b/internal/formats/syftjson/model/package_test.go
@@ -1,0 +1,52 @@
+package model
+
+import "testing"
+
+func TestUnmarshalPackage(t *testing.T) {
+	tests := []struct {
+		name        string
+		p           *Package
+		packageData []byte
+	}{
+		{
+			name:        "Package.UnmarshalJSON will unmarshal blank json",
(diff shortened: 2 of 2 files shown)


internal/formats/syftjson/model/package.go
```diff
@@ -124,6 +124,11 @@ func (p *Package) UnmarshalJSON(b []byte) error {
 			return err
 		}
 		p.Metadata = payload
+	case pkg.GolangBinMetadataType:
```
kzantow, line 127: There's one test that flexes each of the types, or at least most of them: https://github.com/anchore/syft/blob/main/test/integration/encode_decode_cycle_test.go I wonder if we need to add a go binary as one of the options; I think it would be mostly as simple as adding a small go binary here?: https://github.com/anchore/syft/tree/main/test/integration/test-fixtures/image-pkg-coverage/pkgs

kzantow (approved):
👍

internal/formats/syftjson/model/package.go
jonasagx: For flexible reproducibility (where anyone can recreate the binaries quickly, or adapt them if necessary), we can make the binaries on the fly, like we do here https://github.com/anchore/syft/tree/main/syft/pkg/cataloger/golang/test-fixtures/archs

spiffcs: Yea - if we add one here I think we do it as a package `main` file and then compile the program pre-test. We then use the temp file as input and clean it up after. This way we don't commit binaries into the repo since we have ambitions on getting the OSS scorecard numbers up. I can do that as a separate PR to this one as soon as I have extra bandwidth.

kzantow: Yeah, agree not a fan of committing binaries and I know there was talk about moving to Git LFS, I think the unit test is good enough for now 👍

wagoodman: We could curl a go binary in https://github.com/anchore/syft/blob/main/test/integration/test-fixtures/image-pkg-coverage/Dockerfile so we don't have to commit one. Maybe a multistage image, where we use an older syft image and pull the binary?

Issue #518, opened 2021-09-28
Report known unknowns directly in the output SBOM


wagoodman:
It would be ideal if the output SBOM of syft included a description of what is in scope and out of scope as clearly as possible. This includes (but not limited to):
- which catalogers were used during the SBOM generation process
- what was the configuration of each cataloger (should be covered in the app config section already)
- what paths were included in the search vs excluded from the search (as input)
- what paths were attempted to be parsed/searched but could not be for various reasons (IO error, permissions error, parse error)

Why? A good SBOM describes the positive cases of "what" was actually found (X packages, Y files, etc). A great SBOM should describe negative conditions. This doesn't apply to the cases themselves, such as "list all of the packages you didn't find" (which makes no sense), but instead the runtime, configuration, and environmental factors that lead to "places not searched". This helps SBOM consumers better reproduce results and understand the boundaries of what was "in scope" vs "out of scope" during the analysis.

PR #904, opened 2022-03-21
Less verbose logging in Golang Cataloger

jonasagx:
## Before
image

## After
image

### DIFF
--- a/syft/pkg/cataloger/golang/parse_go_bin.go
+++ b/syft/pkg/cataloger/golang/parse_go_bin.go
@@ -25,6 +25,10 @@ var (
 	// appear to be in a known format, or it breaks the rules of that format,
 	// or when there are I/O errors reading the file.
 	errUnrecognizedFormat = errors.New("unrecognized file format")
+
+	// errNotGoExe is returned when a given executable file is valid but does
+	// not contain Go build information.
+	errNotGoExe = errors.New("not a Go executable")
 )
 
 func makeGoMainPackage(mod *debug.BuildInfo, arch string, location source.Location) pkg.Package {
--- a/syft/pkg/cataloger/golang/parse_go_bin.go
+++ b/syft/pkg/cataloger/golang/parse_go_bin.go
@@ -25,10 +25,6 @@ var (
 	// appear to be in a known format, or it breaks the rules of that format,
 	// or when there are I/O errors reading the file.
 	errUnrecognizedFormat = errors.New("unrecognized file format")
-
-	// errNotGoExe is returned when a given executable file is valid but does
-	// not contain Go build information.
-	errNotGoExe = errors.New("not a Go executable")
 )
 
 func makeGoMainPackage(mod *debug.BuildInfo, arch string, location source.Location) pkg.Package {
--- a/syft/pkg/cataloger/golang/scan_bin.go
+++ b/syft/pkg/cataloger/golang/scan_bin.go
@@ -3,6 +3,7 @@ package golang
 
 import (
 	"debug/buildinfo"
+	"errors"
 	"io"
 	"runtime/debug"
 
@@ -32,8 +33,13 @@ func scanFile(reader unionReader, filename string) ([]*debug.BuildInfo, []string
 	var builds []*debug.BuildInfo
 	for _, r := range readers {
 		bi, err := buildinfo.Read(r)
-		if err != nil {
+		if errors.As(err, &errNotGoExe) {
 			log.Debugf("golang cataloger: scanning file %s: %v", filename, err)
+			return nil, nil
+		}
+
+		if err != nil {
+			log.Warnf("golang cataloger: scanning file %s: %v", filename, err)
 			return nil, nil
 		}
 		builds = append(builds, bi)
--- a/syft/pkg/cataloger/golang/scan_bin.go
+++ b/syft/pkg/cataloger/golang/scan_bin.go
@@ -3,7 +3,6 @@ package golang
 
 import (
 	"debug/buildinfo"
-	"errors"
 	"io"
 	"runtime/debug"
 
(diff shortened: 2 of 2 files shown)


syft/pkg/cataloger/golang/scan_bin.go
```diff
@@ -33,7 +33,7 @@ func scanFile(reader unionReader, filename string) ([]*debug.BuildInfo, []string
 	for _, r := range readers {
 		bi, err := buildinfo.Read(r)
 		if err != nil {
-			log.Warnf("golang cataloger: scanning file %s: %v", filename, err)
+			log.Debugf("golang cataloger: scanning file %s: %v", filename, err)
```
luhring, line 36: So my opinion is that this **_should_** be a `WARN` log entry. If Syft knows for sure that it missed an opportunity to scan and surface a piece of the software, that's noteworthy (the same idea that's culminated in https://github.com/anchore/syft/issues/518). For example, it might be that the scanned Go binaries has no known vulnerabilities... except in the section that Syft missed. It's important to have a way to signal to the user the difference between "Everything is fine ✅ " and "Everything _that we scanned_ is fine, but we weren't able to scan everything. ⚠️ "

luhring (commented):
Hey @jonasagx — thanks for the attention to the log noisiness. I'm not sure about this particular change, and I've added a comment. Curious for your thoughts!

syft/pkg/cataloger/golang/scan_bin.go
wagoodman: > If Syft knows for sure that it missed an opportunity to scan and surface a piece of the software, that's noteworthy (the same idea that's culminated in https://github.com/anchore/syft/issues/518). For example, it might be that the scanned Go binaries has no known vulnerabilities... except in the section that Syft missed. It's important to have a way to signal to the user the difference between "Everything is fine ✅ " and "Everything that we scanned is fine, but we weren't able to scan everything. ⚠️ "

100% agree.

However, in this case this is a little bit grey. The golang cataloger selects candidate files based on MIME type (get all executables). There is no distinction of the binary being analyzed being a go-compiled binary (we don't know if it should be). Defaulting to warning here seems wrong since it seems to imply that we were expecting that all inputs are go-compiled binaries (which is not true). 

That being said, I also agree that we don't want to not report up skipping a file that was intended to be analyzed by the golang cataloger but was not. We should show warnings for these cases.

Given the above context, we should probably default to having this being a `debug` (though it does really seem like it should be warning at a first glance).

jonasagx: I like the idea of debug `not a go executable` and warn on any other error. My design north here is: how would things look like once we add more binary catalogers? In this future scenario log messages of the type `not a X executable` are more noise than information, but we are not there yet.

syft/pkg/cataloger/golang/parse_go_bin.go
```diff
@@ -25,6 +25,10 @@ var (
 	// appear to be in a known format, or it breaks the rules of that format,
 	// or when there are I/O errors reading the file.
 	errUnrecognizedFormat = errors.New("unrecognized file format")
+
+	// errNotGoExe is returned when a given executable file is valid but does
+	// not contain Go build information.
+	errNotGoExe = errors.New("not a Go executable")
```
wagoodman, line 31: I don't think this will work since this is a different error instance than the actual error being raised. Does the stdlib expose this error to use directly?

kzantow:
Based on an earlier discussion, it sounds like this might be better suited as data exported by Syft instead of any sort of log warnings at all?

syft/pkg/cataloger/golang/scan_bin.go
```diff
@@ -26,20 +25,23 @@ func scanFile(reader unionReader, filename string) ([]*debug.BuildInfo, []string
...
 		bi, err := buildinfo.Read(r)
-		if errors.As(err, &errNotGoExe) {
-			log.Debugf("golang cataloger: scanning file %s: %v", filename, err)
-			return nil, nil
-		}
 
-		if err != nil {
-			log.Warnf("golang cataloger: scanning file %s: %v", filename, err)
+		// note: the stdlib does not export the error we need to check for
+		if err.Error() == "not a Go executable" {
```
kzantow, line 37: Is this matching some text from the go lib?

wagoodman: indeed, slightly more context here: https://github.com/anchore/syft/pull/904#discussion_r832209830

kzantow (approved):
👍

syft/pkg/cataloger/golang/scan_bin.go
```diff
@@ -25,17 +25,27 @@ func scanFile(reader unionReader, filename string) ([]*debug.BuildInfo, []string
...
+
+		// note: the stdlib does not export the error we need to check for
 		if err != nil {
-			log.Warnf("golang cataloger: scanning file %s: %v", filename, err)
+			if err.Error() == "not a Go executable" {
+				// since the cataloger can only select executables and not distinguish if they are a go-compiled
+				// binary, we should not show warnings/logs in this case.
+				return nil, nil
+			}
+			// in this case we could not read the or parse the file, but not explicitly because it is not a
```
luhring, line 43: >but not explicitly because it is not a go-compiled binary (though it still might be)

nit: this phrasing is just a little difficult to parse, I think because of having "but not ... not ... though ...". If we can think of a more direct way to say this, I think that'd help (still thinking 🤔 )

luhring (approved):

spiffcs (approved):

syft/pkg/cataloger/golang/scan_bin.go
luhring: Maybe something like "We don't know if this is a Go binary, but we weren't able to read it, so we should alert the user."?

(I'm still making sure I understand the intent of this comment, so let me know if I'm off-base!)

wagoodman (approved):

PR #906, opened 2022-03-21
Improve docker config support

wagoodman:
Today when fetching images from the docker daemon, we rely on the default behavior from the API client and some extra behavior implemented in stereoscope. With https://github.com/anchore/stereoscope/pull/114 now stereoscope attempts to use the docker config as a starting point for creating the docker API client used for interacting with the docker daemon. This means that now syft can respond to the current docker context configured, among other functionalities that the docker config provides. This PR bumps stereoscope to pull in this enhancement.


luhring (approved):
🎉

wagoodman:
Pending https://github.com/anchore/stereoscope/pull/115

PR #910, opened 2022-03-23
835 - Keyless Support for SBOM Attestations

spiffcs:
## Summary:

Update the `syft attest` command to use the cosign keyless workflow. 

Users should be able to run `syft attest -o syft-json alpine:latest` and generate an `In-Toto` [Attestation](https://github.com/in-toto/attestation) serialized as a `DSSE` envelope defined [here](https://github.com/secure-systems-lab/dsse/blob/master/envelope.md#dsse-envelope), which is signed by `github`, `google`, or microsoft OIDC providers. They should only be able to UPLOAD attestation for images they own.

Screen Shot 2022-05-03 at 12 40 02 PM

Users in this case will not have to bring their own key. Instead, ephemeral keys and certificates will be generated and signed automatically by the `fulcio` root CA. These signatures are then stored in the `rekor` transparency log which provides an attestation as to when the signature was created. The new feature supports Oauth flows through the browser. Device flows have not been tested and will be updated in follow up PR.

To test the new functionality on this branch you can run:
`go run cmd/syft/main.go attest -o syft-json <IMAGE YOU HAVE WRITE ACCESS TO>`
`COSIGN_EXPERIMENTAL=1 cosign verify-attestation <IMAGE YOU HAVE WRITE ACCESS TO>`
### Documentation
- Update README
- Update ETUI to show progress for attestation flows


cmd/syft/cli/options/attest.go
```diff
@@ -6,23 +6,80 @@ import (
...
 		"path to the private key file to use for attestation")
 
-	return bindAttestationConfigOptions(cmd.PersistentFlags(), v)
+	cmd.Flags().StringVarP(&o.Cert, "cert", "", "",
+		"path to the x.509 certificate in PEM format to include in the OCI Signature")
+
+	cmd.Flags().BoolVarP(&o.NoUpload, "no-upload", "", false,
+		"do not upload the generated attestation")
+
+	cmd.Flags().BoolVarP(&o.Force, "force", "", false,
```
wagoodman, line 47: `force` seems a little misaligned for what the functionality is here (which is auto confirm). Maybe adding only a short hand flag for `-y`?

cmd/syft/cli/options/fulcio.go
```diff
@@ -0,0 +1,48 @@
...
+		"skip verifying fulcio published to the SCT (this should only be used for testing).")
+	return bindFulcioConfigOptions(cmd.Flags(), v)
+}
+
+func bindFulcioConfigOptions(flags *pflag.FlagSet, v *viper.Viper) error {
+	if err := v.BindPFlag("attest.fulcio_url", flags.Lookup("fulcio-url")); err != nil {
+		return err
+	}
+
+	if err := v.BindPFlag("attest.fulcio_identity_token", flags.Lookup("identity-token")); err != nil {
```
wagoodman, line 39: consistency comment: should be `attest.fulcio-identity-token ` (same for other config variables)

cmd/syft/cli/options/oidc.go
```diff
@@ -0,0 +1,48 @@
...
+// AddFlags implements Interface
+func (o *OIDCOptions) AddFlags(cmd *cobra.Command, v *viper.Viper) error {
+	cmd.Flags().StringVar(&o.Issuer, "oidc-issuer", DefaultOIDCIssuerURL,
+		"OIDC provider to be used to issue ID token")
+
+	cmd.Flags().StringVar(&o.ClientID, "oidc-client-id", "sigstore",
+		"OIDC client ID for application")
+
+	cmd.Flags().StringVar(&o.RedirectURL, "oidc-redirect-url", "",
+		"OIDC redirect URL (Optional). The default oidc-redirect-url is 'http://localhost:0/auth/callback'.")
```
wagoodman, line 29: default in the description string doesn't match what the flag default is

wagoodman (commented):
when running I noticed that the default format (table) isn't compatible with this feature, can we conditionally change the format to syft-json in this case?

cmd/syft/cli/attest/attest.go
```diff
@@ -183,23 +209,72 @@ func generateAttestation(predicate []byte, src *source.Source, sv *sign.SignerVe
...
+	}
+
+	// Attach the attestation to the entity.
+	newSE, err := mutate.AttachAttestationToEntity(se, sig)
+	if err != nil {
+		return err
+	}
+
+	// Publish the attestations associated with this entity
+	err = ociremote.WriteAttestations(digest.Repository, newSE)
```
wagoodman, line 255: is there a way to get observability into the upload and publish that to the bus for the ETUI to pick up on?

cmd/syft/cli/attest/attest.go
```diff
@@ -183,23 +209,72 @@ func generateAttestation(predicate []byte, src *source.Source, sv *sign.SignerVe
...
 	}
 
+	if app.Attest.NoUpload {
+		bus.Publish(partybus.Event{
+			Type: event.Exit,
+			Value: func() error {
+				_, err := os.Stdout.Write(signedPayload)
+				return err
+			},
+		})
```
wagoodman, line 219: missing an early return to not upload

cmd/syft/cli/attest/attest.go
```diff
@@ -183,23 +209,72 @@ func generateAttestation(predicate []byte, src *source.Source, sv *sign.SignerVe
...
 			_, err := os.Stdout.Write(signedPayload)
 			return err
 		},
 	})
-
 	return nil
 }
 
+/*
 func findValidDigest(digests []string) string {
```
wagoodman, line 271: nit: dead code -- is it worth deleting?

cmd/syft/cli/attest/attest.go
```diff
@@ -183,23 +209,72 @@ func generateAttestation(predicate []byte, src *source.Source, sv *sign.SignerVe
...
+				_, err := os.Stdout.Write(signedPayload)
+				return err
+			},
+		})
+	}
+
+	return uploadAttestation(app, signedPayload, digest, sv)
+}
+
+func uploadAttestation(app *config.Application, signedPayload []byte, digest name.Digest, sv *sign.SignerVerifier) error {
```
wagoodman, line 225: a great bonus here would be to add comments either in a block before hand or strewn throughout this function as to what is going on, noting the interaction with external entities and what gets stored where (what the transparency log is and what gets uploaded, what gets uploaded to the OCI registry and the important annotations, etc).

cmd/syft/cli/attest/attest.go
```diff
@@ -214,3 +289,31 @@ func formatPredicateType(format sbom.Format) string {
...
+	rekorClient, err := rekor.NewClient(rekorURL)
+	if err != nil {
+		return nil, err
+	}
+	entry, err := upload(rekorClient, rekorBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Infof("tlog entry created with index: %v", *entry.LogIndex)
```
wagoodman, line 317: nit:
```suggestion
	log.Infof("transparency log entry created with index: %v", *entry.LogIndex)
```

wagoodman, line 317: should do a sanity nil check on `entry.LogIndex` before deferencing

wagoodman, line 317: nit: should this be debug? or if it's worthy information, should this information be elevated to the ETUI with upload progress (even if infinite spinner)?

cmd/syft/cli/attest/attest.go
```diff
@@ -183,23 +209,72 @@ func generateAttestation(predicate []byte, src *source.Source, sv *sign.SignerVe
...
+	// Publish the attestations associated with this entity
+	err = ociremote.WriteAttestations(digest.Repository, newSE)
+	if err != nil {
+		return err
+	}
+
 	bus.Publish(partybus.Event{
 		Type: event.Exit,
 		Value: func() error {
 			_, err := os.Stdout.Write(signedPayload)
```
wagoodman, line 263: if we're uploading should we be outputting the signed payload to the screen? or suppress this?

cmd/syft/cli/options/fulcio.go
```diff
@@ -0,0 +1,48 @@
+package options
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+const DefaultFulcioURL = "https://fulcio.sigstore.dev"
```
wagoodman, line 9: nit: this can be unexported

cmd/syft/cli/options/fulcio.go
```diff
@@ -0,0 +1,48 @@
...
+func (o *FulcioOptions) AddFlags(cmd *cobra.Command, v *viper.Viper) error {
+	// TODO: change this back to api.SigstorePublicServerURL after the v1 migration is complete.
+	cmd.Flags().StringVar(&o.URL, "fulcio-url", DefaultFulcioURL,
+		"address of sigstore PKI server")
+
+	cmd.Flags().StringVar(&o.IdentityToken, "identity-token", "",
+		"identity token to use for certificate from fulcio")
+
+	cmd.Flags().BoolVar(&o.InsecureSkipFulcioVerify, "insecure-skip-verify", false,
+		"skip verifying fulcio published to the SCT (this should only be used for testing).")
```
wagoodman, line 30: What's the SCT? The docs should probably reflect the full name.

cmd/syft/cli/attest/attest.go
```diff
@@ -59,20 +71,23 @@ func Run(ctx context.Context, app *config.Application, args []string) error {
...
+		}
 
-	ko := sign.KeyOpts{
-		KeyRef:   app.Attest.Key,
-		PassFunc: passFunc,
+		ko.PassFunc = passFunc
 	}
 
-	sv, err := sign.SignerFromKeyOpts(ctx, "", "", ko)
+	sv, err := sign.SignerFromKeyOpts(ctx, "", "", *ko)
```
wagoodman, line 90: should nil check `ko` first

spiffcs: It's small so I updated it to pass by value instead of pointer here.

cmd/syft/cli/attest/attest.go
spiffcs: That's a good question. I don't think we have access to things within `WriteAttestation`, but let me see what we can pull off of newSE to publish. Right now we only publish the payload. This is the field that gets revealed when calling `COSIGN_EXPERIMENTAL=1 cosign verify-attestation` so I think it's the most important bit to report on during publish to the ETUI.

cmd/syft/cli/attest/attest.go
spiffcs: deleted

cmd/syft/cli/attest/attest.go
spiffcs: I think we should still be publishing the payload to the screen on upload so the user is able to inspect what was uploaded, but I can also see how suppressing is gives a cleaner execution and we don't pollute the screen since payloads are large and noticeably indecipherable on first glance

cmd/syft/cli/attest/attest.go
spiffcs: cc @luhring on this one. I'm on the fence about including this in the default output. I'll demote it to debug for now, but if we come up with a good reason in this thread to why it should stick around I can elevate it into the ETUI

cmd/syft/cli/options/fulcio.go
spiffcs: Also I used SCT incorrectly here so double good call out.

SCT is “Signed Certificate Timestamp”. We need this flag so that we don't have to setup a root CA for fulcio during testing.  Fulcio returns a “Signed Certificate Timestamp”.

See here if you want the deep dive:
https://blog.chainguard.dev/a-fulcio-deep-dive/

cmd/syft/cli/options/oidc.go
spiffcs: Oooo thanks for the find here

cmd/syft/cli/attest/attest.go
spiffcs: cc @luhring for input

cmd/syft/cli/options/attest.go
spiffcs: I totally agree with the comment. The reasoning here is that I'm trying to keep our `attest` command as close to the cosign one as possible for compatibility. They have this option for the reasons stated in the description. I can still change it if you want just giving context for why it is the way it is.

cmd/syft/cli/options/attest.go
```diff
@@ -6,23 +6,80 @@ import (
...
+		"skip warnings and confirmations")
+
+	cmd.Flags().BoolVarP(&o.Recursive, "recursive", "", false,
+		"if a multi-arch image is specified, additionally sign each discrete image")
+
+	cmd.Flags().BoolVarP(&o.Replace, "replace", "", false,
+		"")
```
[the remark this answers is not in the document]
spiffcs: removed this flag as it was not being used currently

cmd/syft/cli/attest/attest.go
spiffcs: Talked offline about these comments - this is less about showing more information and instead adding interaction to the ETUI that displays progress and steps surrounding the attestation uploads

cmd/syft/cli/attest/attest.go
luhring: Does cosign show this? I don't remember for sure but I thought it **_does_** — since it's a convenient way to find the Rekor entry that was just created (i.e. after you run this, you can use a Rekor client to just to a lookup by index)

cmd/syft/cli/attest/attest.go
luhring: Yeah no strong opinion, but I think I'd lean toward not outputting it for now for the "clean feel" reason

cmd/syft/cli/attest/attest.go
spiffcs: comments added @wagoodman

cmd/syft/cli/attest/attest.go
spiffcs: removed output on upload

wagoodman (approved):

jauderho:
Right now I use

```
          cosign sign docker.io/${{ env.REPOSITORY }}@${{ steps.push-step.outputs.digest }}
          syft docker.io/${{ env.REPOSITORY }}@${{ steps.push-step.outputs.digest }} -o json > docker_sbom.json
          cosign attest --predicate docker_sbom.json docker.io/${{ env.REPOSITORY }}@${{ steps.push-step.outputs.digest }}
          cosign verify-attestation docker.io/${{ env.REPOSITORY }}@${{ steps.push-step.outputs.digest }}
```

Just to confirm that with this change, I'll be to do the following instead?

```
          cosign sign docker.io/${{ env.REPOSITORY }}@${{ steps.push-step.outputs.digest }}
          syft attest -o syft-json docker.io/${{ env.REPOSITORY }}@${{ steps.push-step.outputs.digest }} 
          cosign verify-attestation docker.io/${{ env.REPOSITORY }}@${{ steps.push-step.outputs.digest }}
```

COSIGN_EXPERIMENTAL=1 is set elsewhere.

developer-guy:
This PR will also fix #990, by default, it'll attach attestation to an image requested to generate an attestation.
cc: @dentrax @luhring

PR #914, opened 2022-03-23
Ensure that all cyclonedx components have bom-refs

samj1912:

BOM-Refs are important for VEX to refer to components uniquely. This allows us to create an independent vex document and refer to components within the SBOM through a BOM Link (see https://cyclonedx.org/capabilities/bomlink/)

The current logic for setting the bom ref is - 

- PURL + syft id as a qualifier if PURL exists 
- syft artifact id otherwise.

cc: @stevespringett @coderpatros  

This is needed for independent VEX documents in grype.


internal/formats/common/cyclonedxhelpers/component.go
```diff
@@ -25,6 +25,11 @@ func encodeComponent(p pkg.Package) cyclonedx.Component {
 		properties = &props
 	}
 
+	bomRef := p.PURL
```
kzantow, line 28: Maybe `p.ID()` would be a better thing to use here? You should not have to call `p.SetID()`, as it is called if for some reason a package is added to the catalog without having an ID.

samj1912: The CycloneDX docs prefer PURL as the bom ref https://cyclonedx.org/use-cases/#dependency-graph

kzantow: Here's my concern: PURLs are not unique within the Syft data model. For example, if we scan `package-lock.json` and _also_ scan `node_modules` that has been populated, we will find 2 packages with the same PURL but they will not be unique in Syft because of the location of the evidence, whereas the `p.ID()` will be unique.

samj1912: I wouldn't prefer using syft ids as bom refs since having readable bom refs also helps with the VEX readability. Would an alternative like trying to dedupe and merge components with same ids so that we preserve the location/source they were found in, in an array but have a common top level component?

kzantow: I, personally, like this quite a bit: if we find evidence in multiple locations pointing to the same package, this should be reflected in `locations` rather than multiple packages.

kzantow:
@samj1912 what if we always include the `ID`, something like `bom-ref = fmt.Sprintf("%s-%s", p.PURL, p.ID())`? This way we can have a more readable bom-ref _and_ guarantee uniqueness?

internal/formats/cyclonedxjson/test-fixtures/snapshot/TestCycloneDxDirectoryEncoder.golden
```diff
@@ -20,6 +20,7 @@
   },
   "components": [
     {
+      "bom-ref": "b85dbb4e6ece5082",
```
kzantow, line 23: An odd thing: none of these golden examples actually include a PURL-based bom-ref; maybe something is wrong about the logic or none of them have populated PURLs?

samj1912: None of them have populated purls. if you run make validate-schema, you can see some actual examples.

internal/formats/common/cyclonedxhelpers/component.go
```diff
@@ -25,7 +26,15 @@ func encodeComponent(p pkg.Package) cyclonedx.Component {
...
 		properties = &props
 	}
-
+	bomRef := string(p.ID())
+	// try and parse the PURL if possible and append syft id to it, to make
+	// the purl unique in the BOM.
+	// TODO: In the future we may want to dedupe by PURL and combine components with
+	// the same PURL while preserving their unique metadata.
+	if parsedPURL, err := packageurl.FromString(p.PURL); err == nil {
+		parsedPURL.Qualifiers = append(parsedPURL.Qualifiers, packageurl.Qualifier{Key: "syft-id", Value: string(p.ID())})
```
wagoodman, line 35: nice touch making this a pURL qualifier

wagoodman (approved):
Nice job @samj1912 ! I only made a couple of changes:
1. Updated the common format test fixtures to have an example of a valid pURL
2. Added a few tests for the bom-ref creation logic

samj1912:
Thanks!

PR #919, opened 2022-03-24
Add dart support

ericlarssen-wf:


internal/formats/common/spdxhelpers/source_info.go
```diff
@@ -29,6 +29,8 @@ func SourceInfo(p pkg.Package) string {
 		answer = "acquired package info from rust cargo manifest"
 	case pkg.PhpComposerPkg:
 		answer = "acquired package info from PHP composer manifest"
+	case pkg.PubPkg:
```
kzantow, line 32: nit: maybe this should actually be `DartPubPkg` instead of just `PubPkg`?

syft/pkg/cataloger/cataloger.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/pkg/cataloger/apkdb"
+	"github.com/anchore/syft/syft/pkg/cataloger/dartlang"
```
kzantow, line 12: Question: is `dartlang` or `dart` more idiomatic? (I do think I prefer your choice of `dartlang` here)

syft/pkg/cataloger/dartlang/parse_pubspec_lock.go
```diff
@@ -0,0 +1,98 @@
...
+		VcsURL:    p.getVcsURL(),
+	}
+
+	return &pkg.Package{
+		Name:         name,
+		Version:      p.Version,
+		Language:     pkg.Dart,
+		Type:         pkg.PubPkg,
+		MetadataType: pkg.PubMetadataType,
+		Metadata:     *metadata,
```
kzantow, line 71: You could void a temporary pointer of the metadata, just doing:
`Metadata: pkg.PubMetadata{ ... `

syft/pkg/cataloger/dartlang/parse_pubspec_lock.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+func (p *pubspecLockPackage) getVcsURL() string {
+	if p.Source == "git" {
+		if p.Description.Path == "." {
+			return fmt.Sprintf("%s%%40%s", p.Description.URL, p.Description.ResolvedRef)
```
kzantow, line 78: Is there a reason this, I think `@` is encoded as `%40`?

syft/pkg/cataloger/dartlang/parse_pubspec_lock_test.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+func assertPackagesEqual(t *testing.T, actual []*pkg.Package, expected map[string]*pkg.Package) {
+	t.Helper()
+	if len(actual) != len(expected) {
+		for _, a := range actual {
+			t.Log("   ", a)
+		}
+		t.Fatalf("unexpected package count: %d!=%d", len(actual), len(expected))
+	}
```
kzantow, line 19: Maybe just `assert.Len(t, actual, len(expected))` ?

syft/pkg/type.go
```diff
@@ -20,6 +20,7 @@ const (
 	GoModulePkg      Type = "go-module"
 	RustPkg          Type = "rust-crate"
 	KbPkg            Type = "msrc-kb"
+	PubPkg           Type = "pub"
```
kzantow, line 23: Ahh, so this is possibly what is going to be included in the packageurl spec: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#other-candidate-types-to-define

syft/pkg/cataloger/cataloger.go
ericlarssen-wf: I was following the paradigm that the golang module set.

syft/pkg/type.go
ericlarssen-wf: Yes I have a PR https://github.com/package-url/purl-spec/pull/156

syft/pkg/cataloger/cataloger.go
wagoodman: `dart` is probably a better pick. One reason going with just `go` for the go modules was that `go` is a reserved word.

internal/formats/common/spdxhelpers/source_info.go
wagoodman: I'll second this -- `DartPubPkg` better grounds the ecosystem which this belongs without being repetitive.

syft/pkg/dart_pub_metadata.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/syft/linux"
+)
+
+type DartPubMetadata struct {
+	Name        string   `mapstructure:"name" json:"name"`
+	Version     string   `mapstructure:"version" json:"version"`
+	Description string   `mapstructure:"description" json:"description,omitempty"`
+	Authors     []string `mapstructure:"authors" json:"authors,omitempty"`
```
wagoodman, line 12: nit: we try not to `omitempty` collections, and instead enforce an empty collection

syft/pkg/type.go
```diff
@@ -20,6 +20,7 @@ const (
 	GoModulePkg      Type = "go-module"
 	RustPkg          Type = "rust-crate"
 	KbPkg            Type = "msrc-kb"
+	DartPubPkg       Type = "pub"
```
wagoodman, line 23: `dart-pub` is probably a more accurate name here

syft/pkg/cataloger/cataloger.go
```diff
@@ -43,6 +44,7 @@ func ImageCatalogers(cfg Config) []Cataloger {
 		java.NewJavaCataloger(cfg.Java()),
 		apkdb.NewApkdbCataloger(),
 		golang.NewGoModuleBinaryCataloger(),
+		dartlang.NewPubspecLockCataloger(),
```
wagoodman, line 47: question: is a pubspec lock file a solid indication of the package being positively installed? or is the presence of this file not necessarily indicative of it being installed (say if it's in a source repo)?

(We reserve the set of image catalogers by default to only be the set of catalogers that raise up installed packages)

syft/pkg/cataloger/dartlang/cataloger.go
```diff
@@ -0,0 +1,16 @@
+package dartlang
+
+import (
+	"github.com/anchore/syft/syft/pkg/cataloger/common"
+)
+
+type Cataloger struct{}
```
wagoodman, line 7: this struct definition can be deleted (appears to be unused)

wagoodman (commented):
Thanks @ericlarssen-wf  for adding dart support --this is great!  I left a few consistency comments, a question about pub lock files, and also noted the need to bump the JSON schema.

syft/pkg/dart_pub_metadata.go
ericlarssen-wf: I think this is actually extraneous and can be removed altogether.

syft/pkg/cataloger/cataloger.go
ericlarssen-wf: I would say it's akin to a go.sum or pip.lock. My current use case is to inspect artifacts generated by my companies internal build system and hope to use syft to identify that a pubspec.lock file was generated. We happen to use Docker to build said artifacts I suppose scanning a container is not required for our usecase.

ericlarssen-wf: I will remove it for now, we can explore it in the future if we need to.

internal/constants.go
```diff
@@ -6,5 +6,5 @@ const (
 
 	// JSONSchemaVersion is the current schema version output by the JSON encoder
 	// This is roughly following the "SchemaVer" guidelines for versioning the JSON schema. Please see schema/json/README.md for details on how to increment.
-	JSONSchemaVersion = "3.2.1"
+	JSONSchemaVersion = "3.2.3"
```
wagoodman, line 9: I looked back at my comment and realized my feedback has a typo (my bad!). I meant `3.2.2` (the next addition value).

wagoodman: The best way to correct this is to delete the new schema file, change the constant, and regenerate.

ericlarssen-wf:
I was not able to reproduce this issue locally, is there some kind of cache I need to clear?

wagoodman (approved):
Solid addition -- thanks for adding a new language + packaging ecosystem!

PR #926, opened 2022-03-29
Updates parsing of `yarn.lock` to use `resolved` URLs that are pulled from yarn and npm registries

### LINKED ISSUE #925: Parsing yarn.lock fails to identify the currect package and version combinations
**What happened**:
Extracting packages from [this `yarn.lock` file](https://github.com/AMoo-Miki/syft-and-yarn-workspaces/blob/main/yarn.lock) failed to identify the correct and complete packages.

The output contained:
```
 ✔ Indexed ../syft-and-yarn-workspaces
 ✔ Cataloged packages      [3 packages]
async                     0.9.2    npm
resize-observer-polyfill  1.5.2    npm
should-type               1.3.0    npm
```

I expected to see both versions of `async` and the correct `resize-observer-polyfil` package:
```
 ✔ Indexed ../syft-and-yarn-workspaces
 ✔ Cataloged packages      [4 packages]
NAME                             VERSION  TYPE
@4lolo/resize-observer-polyfill  1.5.2    npm
async                            0.9.2    npm
async                            3.2.3    npm
should-type                      1.3.0    npm
```

```
mkdir /tmp/yarn-workspace && cd /tmp/yarn-workspace
git clone https://github.com/AMoo-Miki/syft-and-yarn-workspaces.git .
syft .
```

There are 2 problems in the above example but I want to propose a common solution to solve them both.

***Why do I have 2 `async` versions?***
Working in a Yarn workspace, I can have different versions of the same package across my dependencies. `yarn.lock` would list both of them.

***Why am I overriding the package?***

One method of mitigating vulnerable nested dependencies or working around problems with abandoned nested dependencies is to *replace* them with a forked and fixed release. My sample code is extracted from a much larger code that has the abandoned `resize-observer-polyfill` deeply nested in its dependencies, which I was forced to replace with `@4lolo/resize-observer-polyfill`, a forked and fixed release.

***What am I proposing?***
Depend solely on "resolved" URLs for those entries that are pulled from `registry.yarnpkg.com`; as in `^\s+resolved\s+"https://registry.yarnpkg.com/(.+?)/-/(?:.+?)-(\d+\..+?)\.tgz`.

**Environment**:
- Output of `syft version`: 0.42.4
   * Ubuntu 20.04.2 LTS
   * macOS 12.2.1

AMoo-Miki:
Signed-off-by: Miki <amoo_miki@yahoo.com>

The changes include:
* making the parser only look for resolved URLs in `yarn.lock` files
* creating a fixture of yarn workspace with skeleton dependencies
* updating integration tests in `TestYarnPackageLockDirectory` validate the results against the fixture
* remove old yarn artifacts

#### Resolves
#925


wagoodman (approved):
Clever fix! 🙌 

I think the only caveat I see is that using the `resolved` URL will only match in cases where there is a match with `yarnpkg.com` or `npmjs.org` which means that self-hosts/alternative hosting of packages won't be able to leverage the URL.

wagoodman:
Question about conflicts, from the existing test fixture:
```
c0n-fab_u.laTION@^7.0.0:
  version "7.7.7"
  resolved "https://registry.yarnpkg.com/something-i-made-up/-/c0n-fab_u.laTION-7.7.7.tgz#b9c2bf5805f1e64aadeed6df3a2bfafb5a73f5a0"
  integrity sha512-p32cOF5q0Zqs9uBiONKYLm6BClCoBCM5O9JfeUSlnQLBTxYdTK+pW+nXflm8UkKd2UYlEbYz5qEi0JuZR9ckSw==
```

What is the right value here for the package name? `c0n-fab_u.laTION` or `something-i-made-up` ? Should we require that the outer package name (in this case `c0n-fab_u.laTION`) be a substring of the name extracted from the resolved URL? And if it's not a substring, default to the outer package name?

AMoo-Miki:
> I think the only caveat I see is that using the `resolved` URL will only match in cases where there is a match with `yarnpkg.com` or `npmjs.org` which means that self-hosts/alternative hosting of packages won't be able to leverage the URL.

Correct. For them, it will simply fallback to the name inferred from the identifier. If we find other registry with standardized naming conventions, the regex can be updated.

> What is the right value here for the package name?

While I haven't seen a spec doc for the naming convention used, all the resolved URLs I have seen from NPM and Yarn registries have `@scope/name/-/name` in them. While I am sure there are millions of package-versions on these registries, I have tested the pattern against over 4000 of them. From what i gather, the first `@scope/name` is what the package confesses to be and the second name (everything after `/-/`) is to avoid conflicts when downloading the archives prior to an install.

Of course , short of a spec doc from the registries, my 4000 cases of validation are not proof enough that an NPM/Yarn URL with a different structure doesn't exist. However, I think it is a strong indication that we shouldn't expect to see them.

We have 3 options here:
1. Continue to rely on `@scope/name/-/whatever` and handle problems when someone raises an issue.
2. Change pattern to `(@scope)/(name)/-/($2)` which will fallback to the inferred name.
3. Change pattern to `(@scope)/(name)/-/(another-name) and simply throw a warning about the package name not matching what we expected but continue to use `@scope/name` with the hopes that user will raise an issue asking about it. 

While I think option 1 is perfect because I don't care for "whatever", I think option 2 will not detract from it but option 3 will encourage feedback. 

PS, I will look into the unit test failure.

spiffcs (approved):

spiffcs:
Approved - I'm updating the unit tests on this now

PR #930, opened 2022-03-30
Deduplicate packages across multiple container image layers

### LINKED ISSUE #32: SBOM from all-layers scope showing duplicate packages
The set of analyzers will surface packages based on the small set of rules that each analyzer is coded to enforce. This may surface multiple packages from the same underlying source (e.g. python egg-info analyzer picks up a package that was also picked up by dpkg).

Note: this behavior should be optional via configuration and CLI options, defaulting to not deduping packages.

wagoodman:
This PR adds the ability to deduplicate packages that have the same information found on the same path within multiple layers. 

Additionally, `[]source.Location` has been replaced with `source.LocationSet` on the definition of a package. Why? For two reasons:
1. The ordering of otherwise the same locations for two package should not cause a distinction to be made when deriving the package ID. In this case, the package IDs should be the same (assuming all other fields on the package also match).
2. Since `pkg.Package.Locations` factors into the package ID value, it is convenient to change the response of `pkg.Package.Locations.Hash()` to control when the ID value should change.

Specific changes:
- Add `source.LocationSet`, which represents a unique set of location objects. `Hash()` on this object is derived from the st of real paths from all locations (ignores virtual paths and filesystems). This allows for IDs for locations on packages to match when they have the same locations.
- Augment `source.CoordinateSet.Hash()` to respond based off of the ordered slice that the set if made up of. Note: this change is in response to a bug found when adding tests, not to support `pkg.Package.Locations.Hash()` functionality.
- Add the java metadata virtual path to the set of encoded cyclonedx properties on a package so that encode-decode-encode cycles are stable.
- ~Includes `pURL` within the bytes that are input in the package ID hash. This is because `pkg.Package.merge()` cannot merge string fields, one would need to be dropped. Since the decision to make `pURL` had not considered the relatively new `Format.Decode()` paths from other SBOM tools, the assumption that "all pURL fields are derived from package fields" is not necessarily true.~ If a pURL does not match between the two packages then a) `pkg.Package.merge()` will return an error, and b) `pkg.Catalog.Add()` will log a warning. Since this is synthetic data (added after cataloging) there will be a encode-decode cycle mismatch for ID values, since the first ID derivation will not have a pURL associated yet (and subsequent ID derivations after an encode-decode cycle will).
- Add `pkg.Package.merge()` operation, which merges CPE fields from another package onto the current package (only if the IDs for both packages matches).
- Add `pkg.Catalog.PackagesByName()` helper, which was very convenient in testing and seems generally useful.
- Index values in `pkg.Catalog` are also now deduplicated with a new local `pkg.artifactIDSet` that is only meant for the package catalog.

Closes #32

### DIFF
--- a/syft/pkg/catalog.go
+++ b/syft/pkg/catalog.go
@@ -108,8 +108,9 @@ func (c *Catalog) Add(p Package) {
 		// there is already a package with this fingerprint merge the existing record with the new one
 		if err := existing.merge(p); err != nil {
 			log.Warnf("failed to merge packages: %+v", err)
+		} else {
+			c.addPathsToIndex(p)
 		}
-		c.addPathsToIndex(p)
 		return
 	}
 
--- a/syft/pkg/package.go
+++ b/syft/pkg/package.go
@@ -52,7 +52,7 @@ func (p *Package) merge(other Package) error {
 		return fmt.Errorf("cannot merge packages with different IDs: %q vs %q", p.id, other.id)
 	}
 	if p.PURL != other.PURL {
-		return fmt.Errorf("cannot merge packages with different pURLs: %q=%q vs %q=%q", p.id, p.PURL, other.id, other.PURL)
+		log.Warnf("merging packages have with different pURLs: %q=%q vs %q=%q", p.id, p.PURL, other.id, other.PURL)
 	}
 
 	p.Locations.Add(other.Locations.ToSlice()...)
--- a/syft/pkg/package_test.go
+++ b/syft/pkg/package_test.go
@@ -409,65 +409,6 @@ func TestPackage_Merge(t *testing.T) {
 				},
 			},
 		},
-		{
-			name: "error when there are different pURLs",
-			subject: Package{
-				Name:    "pi",
-				Version: "3.14",
-				FoundBy: "Archimedes",
-				Locations: source.NewLocationSet(
-					originalLocation,
-				),
-				Licenses: []string{
-					"cc0-1.0",
-					"MIT",
-				},
-				Language: "math",
-				Type:     PythonPkg,
-				CPEs: []CPE{
-					must(NewCPE(`cpe:2.3:a:Archimedes:pi:3.14:*:*:*:*:math:*:*`)),
-				},
-				PURL:         "pkg:pypi/pi@3.14",
-				MetadataType: PythonPackageMetadataType,
-				Metadata: PythonPackageMetadata{
-					Name:                 "pi",
-					Version:              "3.14",
-					License:              "cc0-1.0",
-					Author:               "Archimedes",
-					AuthorEmail:          "Archimedes@circles.io",
-					Platform:             "universe",
-					SitePackagesRootPath: "Pi",
-				},
-			},
-			other: Package{
-				Name:    "pi",
-				Version: "3.14",
-				FoundBy: "Archimedes",
-				Locations: source.NewLocationSet(
-					originalLocation,
-				),
-				Licenses: []string{
-					"cc0-1.0",
-					"MIT",
-				},
-				Language: "math",
-				Type:     PythonPkg,
-				CPEs: []CPE{
-					must(NewCPE(`cpe:2.3:a:Archimedes:pi:3.14:*:*:*:*:math:*:*`)),
-				},
-				PURL:         "pkg:pypi/pi-DIFFERENCE@3.14", // difference
-				MetadataType: PythonPackageMetadataType,
-				Metadata: PythonPackageMetadata{
-					Name:                 "pi",
-					Version:              "3.14",
-					License:              "cc0-1.0",
-					Author:               "Archimedes",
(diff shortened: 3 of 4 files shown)


syft/pkg/catalog.go
```diff
@@ -81,15 +88,26 @@ func (c *Catalog) Add(p Package) {
...
 	}
 
+	if existing, exists := c.byID[id]; exists {
+		// there is already a package with this fingerprint merge the existing record with the new one
+		if err := existing.merge(p); err != nil {
+			log.Warnf("failed to merge packages: %+v", err)
+		}
+		return
+	}
+
```
spiffcs, line 98: Do we need to update any of the maps on the catalog after the package has been merged or is `Add` the first time we see `p`?

syft/pkg/cpe_by_specificity.go
```diff
@@ -1,20 +1,20 @@
-package cpe
+package pkg
```
spiffcs, line 1: Was this moved because of the `merge` function that was added for packages?

syft/pkg/package.go
```diff
@@ -14,18 +14,18 @@ import (
...
-	Licenses     []string          // licenses discovered with the package metadata
-	Language     Language          `cyclonedx:"language"`     // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
-	Type         Type              `cyclonedx:"type"`         // the package type (e.g. Npm, Yarn, Python, Rpm, Deb, etc)
-	CPEs         []CPE             `hash:"ignore"`            // all possible Common Platform Enumerators (note: this is NOT included in the definition of the ID since all fields on a CPE are derived from other fields)
-	PURL         string            `hash:"ignore"`            // the Package URL (see https://github.com/package-url/purl-spec) (note: this is NOT included in the definition of the ID since all fields on a pURL are derived from other fields)
-	MetadataType MetadataType      `cyclonedx:"metadataType"` // the shape of the additional data in the "metadata" field
-	Metadata     interface{}       // additional data found while parsing the package source
+	id           artifact.ID        `hash:"ignore"`
```
spiffcs, line 17: Weird that the linter added the white space before each tag here

syft/pkg/package.go
```diff
@@ -46,3 +46,21 @@ func (p Package) ID() artifact.ID {
 func (p Package) String() string {
 	return fmt.Sprintf("Pkg(name=%q version=%q type=%q id=%q)", p.Name, p.Version, p.Type, p.id)
 }
+
+func (p *Package) merge(other Package) error {
+	if p.id != other.id {
```
spiffcs, line 51: Just to be sure - these are the only two conditions we would not want to merge on?

There is no use in comparing CPE either here even if they are ignored by the hash?

syft/source/coordinate_set.go
```diff
@@ -0,0 +1,86 @@
...
+	sort.SliceStable(coordinates, func(i, j int) bool {
+		if coordinates[i].RealPath == coordinates[j].RealPath {
+			return coordinates[i].FileSystemID < coordinates[j].FileSystemID
+		}
+		return coordinates[i].RealPath < coordinates[j].RealPath
+	})
+	return coordinates
+}
+
+func (s CoordinateSet) Hash() (uint64, error) {
```
spiffcs, line 81: Do we need this because when generating the ID `CoordinateSet` is considered?

Are there any times where we are updating/merging a coordinate set AFTER an ID has been generated?

spiffcs (commented):
Just small clarifying comments from my part. The rest LGTM. I did notice a test was failing while doing the review so I think that needs to be fixed before merge 👍

syft/pkg/catalog.go
wagoodman: let me think on that. I think you might be right about needing to update indexes for paths.

syft/pkg/package.go
wagoodman: > these are the only two conditions we would not want to merge on
 correct. The ID check covers all fields that would change an ID (which is most of them), which leaves CPEs and pURLs. CPEs are a slice, so they can be merged, but the pURL field is not a collection, thus cannot be merged.

syft/source/coordinate_set.go
wagoodman: Since LocationSet hash is based on the paths of the coordinate set, the hash of the coordinate set is unrelated to packages. This is to fix when adding files to the SBOM (and relationships to files) such that there is a consistent ID for coordinate sets (I was seeing variation when adding explicit tests for this, so provided an explicit hash function based off of a stable slice)

spiffcs (approved):

syft/pkg/cpe_by_specificity.go
wagoodman: I updated the catalog to include path indexes as well -- good call . Commit 414acc6cecdcb5a6b2dd744f484c418178d1f6c8

spiffcs:
@wagoodman I went through all the files again and nothing stood out after our call. New changes also LGTM

PR #941, opened 2022-04-06
Add digest property to parent and nested java package metadata

spiffcs:
### Summary
Adding digests to the discovered java packages will give downstream SBOM consumers more options for querying artifact registries regarding specific details surrounding SBOM content.

A followup PR in [grype](https://github.com/anchore/grype) will be made after this to allow it to optionally consume this information for matching against an upstream Maven repository for better vulnerability match fidelity 

#### Addtions
- Add digest property to JavaMetadata
- Populate external formats (cyclonedx, cyclonedx-json, spdx, spdx-json, syft-json) with new digest values

#### Todo

- [x] Update CyconeDX and SPDX output with new digest properties and `filesAnalyzed` switch
- [x] Refactor and export digest cataloger functionality to be shared across code


internal/file/digest.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/anchore/syft/internal/log"
+)
+
+const DefaultDigestAlgorithm = "sha1"
+
+// Digest takes a filepath and returns a sha1 digest for the given contents
+func Digest(filepath string) (digest string, err error) {
+	file, err := os.Open(filepath)
+	if err != nil {
+		return digest, xerrors.Errorf("unable to open file: %s - %w", filepath, err)
```
luhring, line 20: Just curious, why `xerrors.Errorf` instead of `fmt.Errorf`? (ditto for a few lines down from here, too)

internal/file/zip_file_traversal.go
```diff
@@ -91,6 +92,12 @@ func ExtractFromZipToUniqueTempFile(archivePath, dir string, paths ...string) (m
 		// provides a ReadCloser. It is up to the caller to handle closing the file explicitly.
 		defer tempFile.Close()
 
+		// grab and assign digest for the visited zip file
+		digest, err := Digest(file.FileInfo().Name())
+		if err != nil {
+			log.Warnf("failed to parse digest for file (%s): %+v", file.Name, err)
```
luhring, line 98: Re: this message... Are we "parsing" the digest? IIUC we're just calculating it

internal/formats/common/cyclonedxhelpers/external_references.go
```diff
@@ -4,12 +4,16 @@ import (
...
 	"strings"
 
+	"github.com/anchore/syft/internal/file"
+	syftFile "github.com/anchore/syft/syft/file"
+
 	"github.com/CycloneDX/cyclonedx-go"
 	"github.com/anchore/syft/syft/pkg"
 )
 
+//nolint:funlen
```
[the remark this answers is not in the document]
kzantow: looks like it needs `gocognit`, too

internal/formats/common/cyclonedxhelpers/external_references.go
```diff
@@ -4,12 +4,16 @@ import (
...
 func encodeExternalReferences(p pkg.Package) *[]cyclonedx.ExternalReference {
-	refs := []cyclonedx.ExternalReference{}
+	var refs []cyclonedx.ExternalReference
```
kzantow, line 16: Will this possibly result in some `nil` pointer issue later?

internal/file/digest.go
```diff
@@ -0,0 +1,35 @@
...
+	"encoding/hex"
+	"io"
+	"os"
+
+	"golang.org/x/xerrors"
+
+	"github.com/anchore/syft/internal/log"
+)
+
+const DefaultDigestAlgorithm = "sha1"
```
luhring, line 14: I think I'm not quite following why this const exists. I was looking at the usages through the code, and I'm wondering why we would refer to this value as opposed to the `Digest.Algorithm` that's attached to the digest value. Wouldn't that be the source of truth for which algorithm was used for the digest?

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -273,7 +288,7 @@ func discoverPkgsFromOpener(virtualPath, pathWithinArchive string, archiveOpener
 	return nestedPkgs, nestedRelationships, nil
 }
 
-func pomPropertiesByParentPath(archivePath string, extractPaths []string, virtualPath string) (map[string]pkg.PomProperties, error) {
+func pomPropertiesByParentPath(archivePath, virtualPath string, extractPaths []string) (map[string]pkg.PomProperties, error) {
```
kzantow, line 291: I think this is a good change, but it makes a bit of unnecessary noise in the PR, would it be better to do this refactoring separately?

kzantow (commented):
Generally, this looks okay, but I'm a little concerned about the `java/archive_parser_test.go` that these hashes are _different_ locally vs CI, what is the reason for this?

internal/file/digest.go
spiffcs: I think this was a mistake on my part. I thought by including this I would get more functionality from the `%w` verb, but on further inspection, I think everything I want for error wrapping is already in the std library.

internal/formats/common/cyclonedxhelpers/external_references.go
spiffcs: You can still append to and assert the length of a nil slice. Goland suggested moving from a literal to a nil slice declaration in this case.

Here is a good StackOverflow on how nil slice declarations behave:
https://stackoverflow.com/questions/44305170/nil-slices-vs-non-nil-slices-vs-empty-slices-in-go-language

internal/file/digest.go
spiffcs: This is definitely a holdover from the fact that this was the first file written vs what came out at the end. Looking again at the whole PR it probably makes sense to not play favorites here on the algorithms.

We currently have `Digest` singular on the javametadata. Looking at how this gets translated into some other specs Digests becomes plural in most cases. 

I'm going to update `Javametadata` so that `Digest` --> `Digests` so we can eliminate the incongruency and not have this default.

internal/file/zip_file_traversal.go
spiffcs: calculating, I'll update the message here

syft/pkg/cataloger/java/archive_parser_test.go
```diff
@@ -283,6 +283,11 @@ func TestParseJar(t *testing.T) {
 				metadata := a.Metadata.(pkg.JavaMetadata)
 				metadata.Parent = nil
 
+				// redact Digest which is computed differently between CI and local
+				if metadata.Digest != nil {
+					metadata.Digest = nil
+				}
```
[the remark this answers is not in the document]
spiffcs: I talked with @wagoodman about this earlier and it has something to do with the packaging on local vs in CI. I do need to double back to look at exactly why it's different, but you can see in this run where we started to notice the differences:

https://github.com/anchore/syft/runs/5858447034?check_suite_focus=true

syft/pkg/java_metadata.go
```diff
@@ -24,6 +25,7 @@ type JavaMetadata struct {
 	Manifest      *JavaManifest  `mapstructure:"Manifest" json:"manifest,omitempty"`
 	PomProperties *PomProperties `mapstructure:"PomProperties" json:"pomProperties,omitempty" cyclonedx:"-"`
 	PomProject    *PomProject    `mapstructure:"PomProject" json:"pomProject,omitempty"`
+	Digest        *file.Digest   `hash:"ignore" json:"digest,omitempty"`
```
wagoodman, line 28: nit: (opinion) a slightly more accurate name for this field is `archiveDigest`, so that it indicates what it is a digest of. what do you think?

internal/file/opener.go
```diff
@@ -7,7 +7,8 @@ import (
 
 // Opener is an object that stores a path to later be opened as a file.
 type Opener struct {
-	path string
+	path   string
+	digest string
```
wagoodman, line 11: this seems convenient but out of place for what an Opener is meant to do

internal/file/zip_file_traversal.go
```diff
@@ -91,6 +92,12 @@ func ExtractFromZipToUniqueTempFile(archivePath, dir string, paths ...string) (m
 		// provides a ReadCloser. It is up to the caller to handle closing the file explicitly.
 		defer tempFile.Close()
 
+		// grab and assign digest for the visited zip file
+		digest, err := Digest(file.FileInfo().Name())
```
wagoodman, line 96: This seems like the wrong spot for this since this is a generic archiver helper, and not necessarily scoped to just java. Maybe moving the digestion to https://github.com/anchore/syft/blob/v0.43.2/syft/pkg/cataloger/java/archive_parser.go#L72-L80 makes more sense?

internal/file/opener.go
spiffcs: 👍  Moved into the archive parser

internal/file/zip_file_traversal.go
spiffcs: moved into the archive parser

syft/file/digest.go
```diff
@@ -1,6 +1,62 @@
...
+	"encoding/hex"
+	"fmt"
+	"hash"
+	"io"
+	"os"
+
+	"github.com/anchore/syft/internal/log"
+)
+
+type HashAlgorithm string
```
wagoodman, line 17: we should be able to use the crypto.Hash object to derive the name --there is already a helper function in the digest cataloger for this

syft/file/digest.go
```diff
@@ -1,6 +1,62 @@
...
+	HashAlgoSHA1   HashAlgorithm = "SHA-1"
+	HashAlgoSHA256 HashAlgorithm = "SHA-256"
+)
+
 type Digest struct {
 	Algorithm string `json:"algorithm"`
 	Value     string `json:"value"`
 }
+
+func CalculateDigest(filepath string, algorithm HashAlgorithm) (digest *Digest, err error) {
```
wagoodman, line 30: this logic already exists in the digests cataloger and is capable of handing multiple digests at once cheaply. I think we could extract that out and use it here. 

https://github.com/anchore/syft/blob/v0.43.2/syft/file/digest_cataloger.go#L80-L118

syft/file/digest.go
```diff
@@ -1,6 +1,62 @@
...
+
+	defer func() {
+		err := file.Close()
+		if err != nil {
+			log.Warnf("unable to close source file=%q from zip=%q: %+v", file.Name, filepath, err)
+		}
+	}()
+
+	return &Digest{
+		Algorithm: string(algorithm),
```
wagoodman, line 59: I think we should try to stick to the same way the algorithm names are being cleaned in the digests cataloger (https://github.com/anchore/syft/blob/v0.43.2/syft/file/digest_cataloger.go#L80-L118)... lowercase, no dashes, etc

internal/file/zip_file_traversal.go
```diff
@@ -71,6 +71,7 @@ func TraverseFilesInZip(archivePath string, visitor func(*zip.File) error, paths
 }
 
 // ExtractFromZipToUniqueTempFile extracts select paths for the given archive to a temporary directory, returning file openers for each file extracted.
+// It also preserves the digest of the discovered zip file on the opener so that data can be used later for metadata construction
```
kzantow, line 74: I think this entire file should be reverted?

syft/file/digest.go
```diff
@@ -1,6 +1,62 @@
...
+	}
+
+	if _, err := io.Copy(h, file); err != nil {
+		return digest, fmt.Errorf("unable to calculate %s for %s: %w", algorithm, filepath, err)
+	}
+
+	defer func() {
+		err := file.Close()
+		if err != nil {
+			log.Warnf("unable to close source file=%q from zip=%q: %+v", file.Name, filepath, err)
```
wagoodman, line 54: this helper function isn't necessarily for zips (given where the helper func location is), so the warning shouldn't mention anything about zips.

Also `file.Name` and `filepath` is redundant info here.

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -157,15 +160,22 @@ func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
 		return nil, nil
 	}
 
+	// grab and assign digest for the entire archive
+	digest, err := syftFile.CalculateDigest(j.archivePath, syftFile.HashAlgoSHA1)
```
wagoodman, line 164: we could pass in the crypto.Hash here instead of making a new algorithm type

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -157,15 +160,22 @@ func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
...
+	if err != nil {
+		log.Warnf("failed to parse digest for file (%s): %+v", j.archivePath, err)
```
wagoodman, line 166: nit:
```suggestion
		log.Warnf("failed to create digest for file=%q: %+v", j.archivePath, err)
```

internal/formats/spdx22json/model/annotation.go
```diff
@@ -1,7 +1,5 @@
 package model
 
-import "time"
```
kzantow, line 3: Why are there changes to this file?

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -157,15 +160,22 @@ func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
...
 		Version:      selectVersion(manifest, j.fileInfo),
 		Language:     pkg.Java,
 		Type:         j.fileInfo.pkgType(),
 		MetadataType: pkg.JavaMetadataType,
 		Metadata: pkg.JavaMetadata{
-			VirtualPath: j.virtualPath,
-			Manifest:    manifest,
+			VirtualPath:    j.virtualPath,
+			Manifest:       manifest,
+			ArchiveDigests: []syftFile.Digest{*digest},
```
wagoodman, line 178: there isn't a nil check, so there could be a panic with `*digest`. It could be that a new `file.NewDigestOfFile(path, crypto.Hash...) ([]Digest, error)` could naturally return what you're looking for without a nil check.

internal/formats/common/cyclonedxhelpers/external_references.go
```diff
@@ -79,6 +95,16 @@ func decodeExternalReferences(c *cyclonedx.Component, metadata interface{}) {
 		meta.Homepage = refURL(c, cyclonedx.ERTypeWebsite)
 	case *pkg.GemMetadata:
 		meta.Homepage = refURL(c, cyclonedx.ERTypeWebsite)
+	case *pkg.JavaMetadata:
+		var digests []syftFile.Digest
+		ref := findExternalRef(c, cyclonedx.ERTypeBuildMeta)
+		for _, hash := range *ref.Hashes {
+			digests = append(digests, syftFile.Digest{
+				Algorithm: fmt.Sprintf("%v", hash.Algorithm),
```
wagoodman, line 103: I could be missing something, but could this be:
```suggestion
				Algorithm: hash.Algorithm,
```

internal/file/zip_file_traversal.go
wagoodman, line 74: dead comment
```suggestion
```

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -135,6 +137,7 @@ func (j *archiveParser) parse() ([]*pkg.Package, []artifact.Relationship, error)
 // discoverMainPackage parses the root Java manifest used as the parent package to all discovered nested packages.
 func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
 	// search and parse java manifest files
+	// TODO: do we want to prefer or check for pom files over manifest here?
```
kzantow, line 140: I'd proffer that we want to check both -- pom files may have a parent so the file itself won't necessarily have all the right info in it, but it _is_ evidence. Also, this doesn't seem pertinent to the change in this PR?

internal/formats/spdx22json/model/annotation.go
```diff
@@ -12,7 +10,7 @@ const (
 type Annotation struct {
 	// Identify when the comment was made. This is to be specified according to the combined date and time in the
 	// UTC format, as specified in the ISO 8601 standard.
-	AnnotationDate time.Time `json:"annotationDate"`
```
wagoodman, line 15: honest question: is this a fix? or related to the digests work?

syft/pkg/cataloger/java/archive_parser_test.go
```diff
@@ -283,6 +284,12 @@ func TestParseJar(t *testing.T) {
 				metadata := a.Metadata.(pkg.JavaMetadata)
 				metadata.Parent = nil
 
+				// redact Digest which is computed differently between CI and local
+				if len(metadata.ArchiveDigests) > 0 {
+					var digests []file.Digest
+					metadata.ArchiveDigests = digests
```
kzantow, line 290: why not just `metadata.ArchiveDigests = nil` or something similar?

syft/pkg/java_metadata.go
```diff
@@ -20,12 +21,13 @@ var jenkinsPluginPomPropertiesGroupIDs = []string{
...
-	Manifest      *JavaManifest  `mapstructure:"Manifest" json:"manifest,omitempty"`
-	PomProperties *PomProperties `mapstructure:"PomProperties" json:"pomProperties,omitempty" cyclonedx:"-"`
-	PomProject    *PomProject    `mapstructure:"PomProject" json:"pomProject,omitempty"`
-	PURL          string         `hash:"ignore" json:"-"` // pURLs and CPEs are ignored for package IDs
-	Parent        *Package       `hash:"ignore" json:"-"` // note: the parent cannot be included in the minimal definition of uniqueness since this field is not reproducible in an encode-decode cycle (is lossy).
+	VirtualPath    string         `json:"virtualPath" cyclonedx:"virtualPath"` // we need to include the virtual path in cyclonedx documents to prevent deduplication of jars within jars
+	Manifest       *JavaManifest  `mapstructure:"Manifest" json:"manifest,omitempty"`
+	PomProperties  *PomProperties `mapstructure:"PomProperties" json:"pomProperties,omitempty" cyclonedx:"-"`
+	PomProject     *PomProject    `mapstructure:"PomProject" json:"pomProject,omitempty"`
+	ArchiveDigests []file.Digest  `hash:"ignore" json:"digest,omitempty"`
```
kzantow, line 28: Are there multiple digests? Is this not just a single `.jar` scan?

spiffcs:
Thanks for the feedback everyone! I'll take another stab and incorporate your comments. I committed some of the suggested changes and need to start looking at what's failing.

syft/pkg/java_metadata.go
spiffcs: This was updated because you can generate multiple digests that use different algorithms. Having this be a list helps match up against the cyclonedx and spdx specifications which also include lists of digest in their output.

internal/formats/spdx22json/model/annotation.go
spiffcs: This was not related. This snuck in because I had initially included digests as an annotation. I can revert it, but it's worth mentioning that `time.Time` does not by default print the ISO 8601 standard. To get that standard we would call something like `time.Now().Format(time.RFC3339)` which would return a string.

internal/formats/spdx22json/model/annotation.go
spiffcs: See this comment here: https://github.com/anchore/syft/pull/941#discussion_r845580950

syft/file/digest.go
spiffcs: I'll just refactor this function out and extract the logic from the cataloger that already exists so we don't have to deal with this warning anymore.

syft/pkg/cataloger/java/archive_parser.go
spiffcs: Yea good call here - I'll just refactor based on the suggestion above to expose what's in the digest cataloger already. I can return by value here as well only if an error is not returned so we don't have to assert on the non error return value

syft/pkg/cataloger/java/archive_parser.go
spiffcs: This is pertinent to the change in this PR since pom files are where we could potentially get data that matches with the maven upstream records. The reason we are adding the digests is so that grype can eventually do a lookup. I noticed we didn't have a preference for the main package of manifest vs pom so wanted to ask the question.

spiffcs:
@wagoodman and @kzantow this has now been updated to address the last round of comments

kzantow (changes requested):
I had another pass at this, I think I identified one `nil pointer` panic, where there's `for _, hash := range *ref.Hashes {`, I'm pretty sure `ref` could be `nil` if not found.

One other thing is the CycloneDX decoder is handling the hashes, it looks like but there isn't anything for SPDX decoding. You're probably not seeing any failing tests because they haven't been added to the encode-decode-encode cycle test. They should be but I suspect that would involve some work outside the scope of this PR. So know we either need a follow-on task to do this (which I would be okay with, personally, to add the spdx formats to encode-decode-encode, which would involve adding the decoding support for the digests) or add decoding support here.

internal/formats/common/cyclonedxhelpers/external_references.go
```diff
@@ -79,6 +95,16 @@ func decodeExternalReferences(c *cyclonedx.Component, metadata interface{}) {
 		meta.Homepage = refURL(c, cyclonedx.ERTypeWebsite)
 	case *pkg.GemMetadata:
 		meta.Homepage = refURL(c, cyclonedx.ERTypeWebsite)
+	case *pkg.JavaMetadata:
+		var digests []syftFile.Digest
+		ref := findExternalRef(c, cyclonedx.ERTypeBuildMeta)
+		for _, hash := range *ref.Hashes {
```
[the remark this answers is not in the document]
spiffcs: `ERTypeBuildMeta` is a const from the cyclonedx package

```
ERTypeBuildMeta     ExternalReferenceType = "build-meta"
```

internal/formats/spdx22tagvalue/to_format_model.go
```diff
@@ -103,6 +103,24 @@ func toFormatPackages(catalog *pkg.Catalog) map[spdx.ElementID]*spdx.Package2_2
 		// the Comments on License field (section 3.16) is preferred.
 		license := spdxhelpers.License(p)
 
+		filesAnalyzed := false
+		checksums := make(map[spdx.ChecksumAlgorithm]spdx.Checksum)
+
+		// If the pkg type is Java we have attempted to generated a digest
+		// FilesAnalyzed should be true in this case
+		if p.MetadataType == pkg.JavaMetadataType {
+			javaMetadata := p.Metadata.(pkg.JavaMetadata)
```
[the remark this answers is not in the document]
spiffcs: I think having the guard of if `p.MetadataType == pkg.JavaMetadataType {` gives us the freedom here to do the conversion as it is

syft/file/digest_cataloger.go
```diff
@@ -77,30 +77,39 @@ func (i *DigestsCataloger) catalogLocation(resolver source.FileResolver, locatio
 	}
 	defer internal.CloseAndLogError(contentReader, location.VirtualPath)
 
+	digests, err := DigestsFromFile(contentReader, i.hashes)
+	if err != nil {
+		return nil, internal.ErrPath{Context: "digests-cataloger", Path: location.RealPath, Err: err}
+	}
+
+	return digests, err
```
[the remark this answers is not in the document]
spiffcs: It's common in Golang to return named values at the end of functions as well.

https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/bufio/bufio.go;l=755-798

I don't think one is more clear than the other in this case, but I'll change it

internal/formats/spdx22json/to_format_model.go
```diff
@@ -55,14 +55,32 @@ func toPackages(catalog *pkg.Catalog, relationships []artifact.Relationship) []m
 	for _, p := range catalog.Sorted() {
 		license := spdxhelpers.License(p)
 		packageSpdxID := model.ElementID(p.ID()).String()
-
+		filesAnalyzed := false
+		checksums := make([]model.Checksum, 0)
```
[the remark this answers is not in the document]
spiffcs: Updated

internal/formats/spdx22json/to_format_model.go
```diff
@@ -55,14 +55,32 @@ func toPackages(catalog *pkg.Catalog, relationships []artifact.Relationship) []m
...
+
+		// we generate digest for some Java packages
+		// see page 33 of the spdx specification for 2.2
+		// https://spdx.dev/wp-content/uploads/sites/41/2020/08/SPDX-specification-2-2.pdf
```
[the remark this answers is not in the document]
spiffcs: updated

internal/formats/common/cyclonedxhelpers/external_references.go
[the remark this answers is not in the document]
kzantow: Right but `*ref` could panic, dereferencing a nil pointer.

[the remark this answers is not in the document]
spiffcs: Updated to guard against any nil value returned from `findExternalRef`

kzantow (approved):

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -34,6 +37,11 @@ var archiveFormatGlobs = []string{
 	// project that we can build in CI feel free to include it
 }
 
+// JavaArchiveHashes are all the current hash algorithms used to calculate archive digests
+var JavaArchiveHashes = []crypto.Hash{
```
wagoodman, line 41: nit: could unexport this

wagoodman (approved):
nice work 🙌

spiffcs:
@kzantow @wagoodman I added decoding for spdx22 support can you take a look

PR #946, opened 2022-04-11
Update spdx22json to only take uppercase checksum algorithm

spiffcs:
When doing validation against SPDXJSON2.2 an error would be thrown saying `sha1` was not a valid algorithm.

This change updates the encoding for `spdx22json` so that algorithm is now uppercased according to the specification.

https://spdx.dev/wp-content/uploads/sites/41/2020/08/SPDX-specification-2-2.pdf
See page 35 Package checksum 3.10.4


kzantow (approved):

kzantow:
Good catch!

Issue #373, opened 2021-04-07
Support cataloging NuGet packages


wagoodman:
It would be useful to catalog NuGet packages. We should consider deriving this information from one or more sources:
- `packages.lock.json` which has pinned dependencies listed
-  `.nuspec` which contains package metadata

Before diving in we should consider if this should be a cataloger for images, directory, or a hybrid of both.

westonsteimel:
For the deployed artifacts of a .NET Core app I think you typically end up with something like `<app_name>.deps.json`.  This is an example output for `TestApp.deps.json`

```json
{
  "runtimeTarget": {
    "name": ".NETCoreApp,Version=v5.0",
    "signature": ""
  },
  "compilationOptions": {},
  "targets": {
    ".NETCoreApp,Version=v5.0": {
      "TestApp/1.0.0": {
        "dependencies": {
          "AWSSDK.S3": "3.7.0.10",
          "Newtonsoft.Json": "13.0.1"
        },
        "runtime": {
          "TestApp.dll": {}
        }
      },
      "AWSSDK.Core/3.7.0.10": {
        "runtime": {
          "lib/netcoreapp3.1/AWSSDK.Core.dll": {
            "assemblyVersion": "3.3.0.0",
            "fileVersion": "3.7.0.10"
          }
        }
      },
      "AWSSDK.S3/3.7.0.10": {
        "dependencies": {
          "AWSSDK.Core": "3.7.0.10"
        },
        "runtime": {
          "lib/netcoreapp3.1/AWSSDK.S3.dll": {
            "assemblyVersion": "3.3.0.0",
            "fileVersion": "3.7.0.10"
          }
        }
      },
      "Newtonsoft.Json/13.0.1": {
        "runtime": {
          "lib/netstandard2.0/Newtonsoft.Json.dll": {
            "assemblyVersion": "13.0.0.0",
            "fileVersion": "13.0.1.25517"
          }
        }
      }
    }
  },
  "libraries": {
    "TestApp/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "AWSSDK.Core/3.7.0.10": {
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-XIg3tsHLQwN1k/H2M/dUZa2hc+cUDoDJBZpppxX5XJ1vhvI/MhQR9PTBDM8rNenxTNmfLflruU51vamvzrBXeQ==",
      "path": "awssdk.core/3.7.0.10",
[...]

PR #951, opened 2022-04-13
feat: add initial dotnet-support

### LINKED ISSUE #726: .NET Core-Support
Hi Syft-Team,
are there any plans to support generating SBOMs for .NET Core projects?

**Why is this needed**:
.NET Core is a widely used ecosystem.

**Additional context**:

ckotzbauer:
Signed-off-by: Christian Kotzbauer <git@ckotzbauer.de>

This PR adds initial support for .NET (Core). It focuses on the parsing of the `*.deps.json` files in the `bin/` folders of a compiled project. This does not add support for parsing NuGet-related files which are usually only available at source-level.

I tested the changes against a compiled .NET project and against a ASP.NET Core image with the compiled project inside.

This changes are done alongside the PR adding dart-support. Please let me know, if anything is missing, wrong or should be changed.

Fixes #726
Ref #373

### DIFF
--- a/syft/pkg/cataloger/cataloger.go
+++ b/syft/pkg/cataloger/cataloger.go
@@ -11,6 +11,7 @@ import (
 	"github.com/anchore/syft/syft/pkg/cataloger/apkdb"
 	"github.com/anchore/syft/syft/pkg/cataloger/dart"
 	"github.com/anchore/syft/syft/pkg/cataloger/deb"
+	"github.com/anchore/syft/syft/pkg/cataloger/dotnet"
 	"github.com/anchore/syft/syft/pkg/cataloger/golang"
 	"github.com/anchore/syft/syft/pkg/cataloger/java"
 	"github.com/anchore/syft/syft/pkg/cataloger/javascript"
@@ -44,6 +45,7 @@ func ImageCatalogers(cfg Config) []Cataloger {
 		java.NewJavaCataloger(cfg.Java()),
 		apkdb.NewApkdbCataloger(),
 		golang.NewGoModuleBinaryCataloger(),
+		dotnet.NewDotnetDepsCataloger(),
 	}
 }
 
@@ -63,6 +65,7 @@ func DirectoryCatalogers(cfg Config) []Cataloger {
 		golang.NewGoModFileCataloger(),
 		rust.NewCargoLockCataloger(),
 		dart.NewPubspecLockCataloger(),
+		dotnet.NewDotnetDepsCataloger(),
 	}
 }
 
@@ -83,5 +86,6 @@ func AllCatalogers(cfg Config) []Cataloger {
 		golang.NewGoModFileCataloger(),
 		rust.NewCargoLockCataloger(),
 		dart.NewPubspecLockCataloger(),
+		dotnet.NewDotnetDepsCataloger(),
 	}
 }
--- a/syft/pkg/cataloger/dotnet/cataloger.go
+++ b/syft/pkg/cataloger/dotnet/cataloger.go
@@ -1,0 +1,14 @@
+package dotnet
+
+import (
+	"github.com/anchore/syft/syft/pkg/cataloger/common"
+)
+
+// NewDotnetDepsCataloger returns a new Dotnet cataloger object base on deps json files.
+func NewDotnetDepsCataloger() *common.GenericCataloger {
+	globParsers := map[string]common.ParserFn{
+		"**/*deps.json": parseDotnetDeps,
+	}
+
+	return common.NewGenericCataloger(nil, globParsers, "dotnet-deps-cataloger")
+}
--- a/syft/pkg/dotnet_deps_metadata.go
+++ b/syft/pkg/dotnet_deps_metadata.go
@@ -1,0 +1,24 @@
+package pkg
+
+import (
+	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/syft/linux"
+)
+
+type DotnetDepsMetadata struct {
+	Name    string `mapstructure:"name" json:"name"`
+	Version string `mapstructure:"version" json:"version"`
+}
+
+func (m DotnetDepsMetadata) PackageURL(_ *linux.Release) string {
+	var qualifiers packageurl.Qualifiers
+
+	return packageurl.NewPackageURL(
+		"dotnet", // TODO: packageurl.TypeDotnet,
+		"",
+		m.Name,
+		m.Version,
+		qualifiers,
+		"",
+	).ToString()
+}
--- a/syft/pkg/language.go
+++ b/syft/pkg/language.go
@@ -20,6 +20,7 @@ const (
 	Go              Language = "go"
 	Rust            Language = "rust"
 	Dart            Language = "dart"
+	Dotnet          Language = "dotnet"
 )
 
 // AllLanguages is a set of all programming languages detected by syft.
@@ -32,6 +33,7 @@ var AllLanguages = []Language{
 	Go,
 	Rust,
 	Dart,
+	Dotnet,
 }
 
 // String returns the string representation of the language.
@@ -66,6 +68,9 @@ func LanguageByName(name string) Language {
 		return Rust
 	case packageurl.TypePub, string(Dart):
 		return Dart
+	case string(Dotnet): // TODO:
+		//case packageurl.TypeDotnet, string(Dotnet):
+		return Dotnet
 	default:
 		return UnknownLanguage
 	}
--- a/syft/pkg/type.go
+++ b/syft/pkg/type.go
@@ -21,6 +21,7 @@ const (
 	RustPkg          Type = "rust-crate"
 	KbPkg            Type = "msrc-kb"
 	DartPubPkg       Type = "dart-pub"
+	DotnetPkg        Type = "dotnet"
 )
 
 // AllPkgs represents all supported package types
@@ -38,6 +39,7 @@ var AllPkgs = []Type{
 	RustPkg,
 	KbPkg,
 	DartPubPkg,
+	DotnetPkg,
 }
 
 // PackageURLType returns the PURL package type for the current package.
@@ -65,6 +67,9 @@ func (t Type) PackageURLType() string {
 		return "cargo"
 	case DartPubPkg:
 		return packageurl.TypePub
+	case DotnetPkg:
+		// TODO: return packageurl.TypeDotnet
+		return "dotnet"
 	default:
 		// TODO: should this be a "generic" purl type instead?
 		return ""
@@ -104,6 +109,9 @@ func TypeByName(name string) Type {
 		return RustPkg
 	case packageurl.TypePub:
 		return DartPubPkg
+	//case packageurl.TypeDotnet:
+	case "dotnet": // TODO:
+		return DotnetPkg
 	default:
 		return UnknownPkg
 	}
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ A CLI tool and Go library for generating a Software Bill of Materials (SBOM) fro
 - Alpine (apk)
 - Dart (pubs)
 - Debian (dpkg)
+- Dotnet (deps.json)
 - Go (go.mod, Go binaries)
 - Java (jar, ear, war, par, sar)
 - JavaScript (npm, yarn)
--- a/internal/constants.go
+++ b/internal/constants.go
@@ -6,5 +6,5 @@ const (
 
 	// JSONSchemaVersion is the current schema version output by the JSON encoder
 	// This is roughly following the "SchemaVer" guidelines for versioning the JSON schema. Please see schema/json/README.md for details on how to increment.
-	JSONSchemaVersion = "3.2.2"
+	JSONSchemaVersion = "3.2.3"
 )
--- a/internal/formats/common/spdxhelpers/source_info.go
+++ b/internal/formats/common/spdxhelpers/source_info.go
@@ -17,6 +17,8 @@ func SourceInfo(p pkg.Package) string {
 		answer = "acquired package info from pubspec manifest"
 	case pkg.DebPkg:
 		answer = "acquired package info from DPKG DB"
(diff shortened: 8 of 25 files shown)


syft/pkg/dotnet_deps_metadata.go
```diff
@@ -0,0 +1,24 @@
...
+type DotnetDepsMetadata struct {
+	Name    string `mapstructure:"name" json:"name"`
+	Version string `mapstructure:"version" json:"version"`
+}
+
+func (m DotnetDepsMetadata) PackageURL(_ *linux.Release) string {
+	var qualifiers packageurl.Qualifiers
+
+	return packageurl.NewPackageURL(
+		"dotnet", // TODO: packageurl.TypeDotnet,
```
ckotzbauer, line 17: Missing TypeDotnet

syft/pkg/language.go
```diff
@@ -66,6 +68,9 @@ func LanguageByName(name string) Language {
 		return Rust
 	case packageurl.TypePub, string(Dart):
 		return Dart
+	case string(Dotnet): // TODO:
+		//case packageurl.TypeDotnet, string(Dotnet):
```
ckotzbauer, line 72: Missing TypeDotnet

syft/pkg/type.go
```diff
@@ -65,6 +67,9 @@ func (t Type) PackageURLType() string {
 		return "cargo"
 	case DartPubPkg:
 		return packageurl.TypePub
+	case DotnetPkg:
+		// TODO: return packageurl.TypeDotnet
+		return "dotnet"
```
ckotzbauer, line 72: Missing TypeDotnet

syft/pkg/type.go
```diff
@@ -104,6 +109,9 @@ func TypeByName(name string) Type {
 		return RustPkg
 	case packageurl.TypePub:
 		return DartPubPkg
+	//case packageurl.TypeDotnet:
+	case "dotnet": // TODO:
+		return DotnetPkg
```
ckotzbauer, line 114: Missing TypeDotnet

ckotzbauer (commented):
I will open a PR against the `package-url/packageurl-go` repo when the namings are approved here.

syft/pkg/cataloger/dotnet/cataloger.go
```diff
@@ -0,0 +1,14 @@
+package dotnet
+
+import (
+	"github.com/anchore/syft/syft/pkg/cataloger/common"
+)
+
+// NewDotnetDepsCataloger returns a new Dotnet cataloger object base on deps json files.
+func NewDotnetDepsCataloger() *common.GenericCataloger {
+	globParsers := map[string]common.ParserFn{
+		"**/*deps.json": parseDotnetDeps,
```
wagoodman, line 10: Should there be an extra dot here?
```suggestion
		"**/*.deps.json": parseDotnetDeps,
```

syft/pkg/dotnet_deps_metadata.go
```diff
@@ -0,0 +1,24 @@
+package pkg
+
+import (
+	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/syft/linux"
+)
+
+type DotnetDepsMetadata struct {
```
wagoodman, line 8: This is probably a good opportunity to raise up the additional raw information for each of the packages in the `*.deps.json`, such as:
```
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-kHBB+QmosVaG6DpngXQ8OlLVVNMzltNITfsRr68Z90qO7dSqJ2EHNd8dtBU1u3AQQLqqFHOY0lfmbpexeH6Pew==",
      "path": "awssdk.core/3.7.10.6",
      "hashPath": "awssdk.core.3.7.10.6.nupkg.sha512"
```

Straight from https://github.com/dotnet/cli/blob/v2.0.0/Documentation/specs/runtime-configuration-file.md

> The type property of a dependency object defines what kind of entity satisfied the dependency. Possible values include project and package (further comments on dependency types below).
...Open Question: type is also present in the libraries section. We don't really need it in both. It's in both now because the lock file does that and we want the formats to be similar. Should we remove it?

> # libraries Section (.deps.json)
This section contains a union of all the dependencies found in the various targets, and contains common metadata for them. Specifically, it contains the type, as well as a boolean indicating if the library can be serviced (serviceable, only for package-typed libraries) and a SHA-512 hash of the package file (sha512, only for package-typed libraries.

From a high level it seems like `path`, `sha512`, `hashPath` could be valuable to capture. I'm not certain of the value of `serviceable` or `type`, but wouldn't be against including these fields in `DotnetDepsMetadata`.

syft/pkg/language.go
```diff
@@ -20,6 +20,7 @@ const (
 	Go              Language = "go"
 	Rust            Language = "rust"
 	Dart            Language = "dart"
+	Dotnet          Language = "dotnet"
```
wagoodman, line 23: Open question: do we prefer `.net` or `dotnet` for string representations of these values?

syft/pkg/dotnet_deps_metadata.go
ckotzbauer: This sounds good. I can include them. Not sure about `serviceable` and `type` too. Type would always be `package` (there's an if-statement) and `serviceable` will then always be true...

syft/pkg/cataloger/dotnet/cataloger.go
ckotzbauer: Yes, that's missing

syft/pkg/language.go
ckotzbauer: hm, I think `.net` can be used for `.NET Framework` and `.NET Core`. `dotnet` always just means `.NET Core`, so I would prefer `dotnet`.

syft/pkg/dotnet_deps_metadata.go
ckotzbauer: Added `path`, `sha512` and `hashPath`

syft/pkg/language.go
wagoodman: sounds great -- thanks for the context 👍

wagoodman:
re: packageurl-go updates, [we use a fork](https://github.com/anchore/packageurl-go) of [package-url/packageurl-go](https://github.com/package-url/packageurl-go) for this reason -- so we can get the changes needed for syft in asap without waiting for upstream. We submit PRs to upstream as well, though the merge time can take a while.

Did you want to also create a PR in our fork of packageurl-go for the `TypeDotnet` update? (if not we can leave it as is too and get the cataloger in)

syft/pkg/cataloger/cataloger.go
```diff
@@ -44,6 +45,7 @@ func ImageCatalogers(cfg Config) []Cataloger {
 		java.NewJavaCataloger(cfg.Java()),
 		apkdb.NewApkdbCataloger(),
 		golang.NewGoModuleBinaryCataloger(),
+		dotnet.NewDotnetDepsCataloger(),
```
wagoodman, line 48: Question: does the `*.deps.json` file positively indicate installed software? or is it mostly indicative of dependencies that you would find in a git repo?

This may seem like an odd question, but let me provide a little context. The image catalogers only raise up packages which have a high confidence of being installed. So for example, a ruby `*.gemspec` file alone may be an indication of an installed package or could be found in the root of a git repo and there is no software "installed". The way we distinct this is by ensuring that the gemspec is within a "specifications" directory, which is where gemspecs are found when they have been installed by the ruby toolchain (`**/specifications/**/*.gemspec`).

I'm trying to understand if the `.deps.json` falls into one or both categories. Does it indicate declared dependencies (e.g. in a source repo)? Does it indicate installed software? Or both? (and if both, can we distinct the two cases clearly enough, say as we do with gemspecs?)

Thanks in advance!

wagoodman (commented):
_(not in scope for this PR, just capturing a comment for sometime in the future after this PR is merged)_ 

We could capture [`*.runtimeconfig.json`](https://github.com/dotnet/cli/blob/v2.0.0/Documentation/specs/runtime-configuration-file.md#appnameruntimeconfigjson) in the package metadata if it's available (would require significant cataloger changes).

wagoodman:
I think the [linter issues](https://github.com/anchore/syft/runs/6197926925?check_suite_focus=true) can be fixed with `make lint-fix`

syft/pkg/cataloger/cataloger.go
ckotzbauer: The *.deps.json file is generated from the compiler, so you won't find that file in a git-repo as it would be gitignored. Does this answer the question?

ckotzbauer:
> I think the [linter issues](https://github.com/anchore/syft/runs/6197926925?check_suite_focus=true) can be fixed with `make lint-fix`

fixed

Is the CLI-Test failure related to these changes?

> Did you want to also create a PR in our fork of packageurl-go for the TypeDotnet update?

Yes, I will do a PR against the fork.

syft/pkg/cataloger/cataloger.go
wagoodman: I think it does yeah --it's distinct from nuget (a package manager which would be installations) and distinct from a declared project manifest (which would be outside install cases) but still seems strongly associated with something that has been compiled and run.

Thanks!

wagoodman:
@ckotzbauer I merged the pURL PR, so you should be good to go there.

re: CLI tests, the failure seems related to:
```
    packages_cmd_test.go:236: expected package count of 20, but found 32
...

    packages_cmd_test.go:240: STDERR:
         [0000]  INFO could not identify distro
        [0000]  INFO cataloging image
        
    packages_cmd_test.go:241: COMMAND: /home/runner/work/syft/syft/snapshot/linux-build_linux_amd64/syft packages -o json -s squashed docker-archive:test-fixtures/cache/stereoscope-fixture-image-pkg-coverage-d5baa8a0e48b30716c65fda2fecff73c8a85bdc023917b7ccae9ae0903d6d94c.tar
```

Since the `coverageImage` now includes `dotnet` packages, the total number of packages expected has grown:
- https://github.com/ckotzbauer/syft/blob/feature/dotnet/test/cli/packages_cmd_test.go#L99.

The easiest way to iterate locally on the CLI tests is:
- `make clean-snapshot snapshot` to create a build
- `make cli` to run tests against the build

I went ahead an fixed the test in question in 009dd0afadfc9ac9f46125b8148b7db4f8fa5e28, so no action needed 🎉

wagoodman:
Open question: though I've merged https://github.com/anchore/packageurl-go/pull/8 I'm currently trying to make certain I understand the right usage of package URL for a .NET package relative to what a *.deps.json describes. As I understand it a .deps.json has nothing to do with nuget. However, in terms of pURL that might not matter. That is, the information contained within deps.json simply describe .NET packages, which canonically is nuget. 

In short: should we be using the [pURL type `nuget`](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#nuget) instead of the proposed `dotnet` pURL type? 

In thinking/reading through this more I think the answer is "yes" we should use "nuget", but wanted to raise this up for comment.

ckotzbauer:
Great, thanks for fixing the CLI test.

>  I merged the pURL PR, so you should be good to go there.

I will update this PR.

> should we be using the [pURL type nuget](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#nuget) instead of the proposed dotnet pURL type?

No, I don't think so. Nuget is the most popular package-manager for dotnet, but there are alternatives (e.g. https://fsprojects.github.io/Paket/index.html). I always used Nuget before, but I think the result in deps.json would be the same with another tool, as there's no reference to "nuget" in this file.

wagoodman:
I think that all makes sense. I'll note that since the pURL type is non-standard that there is risk of needing to change this in the future, but I think for now we can operate under the assumption for now that a pURL that doesn't automatically couple .NET projects with `nuget` is the way to go in this instance.

wagoodman:
Useful JSON schema diff for reviewers
```diff
# diff schema/json/schema-3.2.2.json  schema/json/schema-3.2.3.json
279a280,295
>     "DotnetDepsMetadata": {
>       "required": [
>         "name",
>         "version"
>       ],
>       "properties": {
>         "name": {
>           "type": "string"
>         },
>         "version": {
>           "type": "string"
>         }
>       },
>       "additionalProperties": true,
>       "type": "object"
>     },
520a537,542
>         },
>         "digest": {
>           "items": {
>             "$ref": "#/definitions/Digest"
>           },
>           "type": "array"
690a713,715
>               "$ref": "#/definitions/DotnetDepsMetadata"
>             },
>             {
```

wagoodman (approved):
cc: @anchore/tools 
Looks like there are minor tweaks in grype testing needed -- I can confirm there is no apparent unexpected matching behavior with the stock matcher. We'll still need to add a .NET-specific matcher and DB namespace finder to link up GHSA matches in the future.

Overall, great work @ckotzbauer ! Thanks for taking the time on this ❤️

ckotzbauer:
Oh, just because you posted the schema diff: I extended the Metadata as part of the review, but I think I forgot to regenerate the schema.

wagoodman:
Latest diff:
```diff
# diff schema/json/schema-3.2.2.json  schema/json/schema-3.2.3.json
279a280,307
>     "DotnetDepsMetadata": {
>       "required": [
>         "name",
>         "version",
>         "path",
>         "sha512",
>         "hashPath"
>       ],
>       "properties": {
>         "name": {
>           "type": "string"
>         },
>         "version": {
>           "type": "string"
>         },
>         "path": {
>           "type": "string"
>         },
>         "sha512": {
>           "type": "string"
>         },
>         "hashPath": {
>           "type": "string"
>         }
>       },
>       "additionalProperties": true,
>       "type": "object"
>     },
520a549,554
>         },
>         "digest": {
>           "items": {
>             "$ref": "#/definitions/Digest"
>           },
>           "type": "array"
690a725,727
>               "$ref": "#/definitions/DotnetDepsMetadata"
>             },
>             {
```

PR #954, opened 2022-04-13
Add workflow for automatic PR for new stereoscope updates

kzantow:
This PR adds a nightly check for updates to stereoscope and automatically creates a PR if there has been something merged to main that has not been updated in Syft.

### DIFF
--- a/.github/workflows/update-stereoscope-release.yml
+++ b/.github/workflows/update-stereoscope-release.yml
@@ -1,0 +1,50 @@
+name: PR for latest Stereoscope release
+on:
+  schedule:
+    - cron: "0 8 * * *" # 3 AM EST
+
+  workflow_dispatch:
+
+env:
+  GO_VERSION: "1.18.x"
+  GO_STABLE_VERSION: true
+
+jobs:
+  upgrade-stereoscope:
+    runs-on: ubuntu-latest
+    if: github.repository == 'anchore/syft' # only run for main repo
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-stereoscope-release.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          stable: ${{ env.GO_STABLE_VERSION }}
+
+      - run: |
+          LATEST_VERSION=$(git ls-remote https://github.com/anchore/stereoscope main | head -n1 | awk '{print $1;}')
```
spiffcs, line 25: Are we get the current version that syft is using from the go.mod file?

If we find that this version is the same as `LATEST_VERSION` are we then able to short circuit the PR?

My thinking here is that if we can make it only every night where the versions change we have a chance to reduce the PR noise that could get introduced to our notifications feed.

spiffcs (dismissed):
Nice addition! Just one comment on possibly acting only if the versions have changed.

spiffcs (approved):

.github/workflows/update-stereoscope-release.yml
kzantow: As noted in the Grype PR, the auto-PR won't be created if there are no changes 👍

Issue #949, opened 2022-04-13
Invalid SPDXID (contains an underscore)


vargenau:
**What happened**:

The SPDXID for one package is invalid as it contains an underscore.

```
SPDXID: SPDXRef-Package-apk-ssl_client-241ed339d4367bc4
```

**What you expected to happen**:

The underscore "_" should be replaced by a dash "-".

**How to reproduce it (as minimally and precisely as possible)**:

```
syft docker:fsfe/reuse:latest -o spdx-tag-value
```

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`:

```
Application:        syft
Version:            0.44.0
JsonSchemaVersion:  3.2.2
BuildDate:          2022-04-12T18:59:04Z
GitCommit:          b46d044d7e3afc2ee09fd5ff26635f783628126e
GitDescription:     v0.44.0
Platform:           linux/amd64
GoVersion:          go1.18
Compiler:           gc
```

- OS (e.g: `cat /etc/os-release` or similar):
```
PRETTY_NAME="Ubuntu 21.10"
NAME="Ubuntu"
VERSION_ID="21.10"
VERSION="21.10 (Impish Indri)"
VERSION_CODENAME=impish
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=impish
```

PR #955, opened 2022-04-13
Ensure SPDXIDs are valid

kzantow:
SPDXIDs were being generated with in valid characters. According to the spec, these must be the format:

> "SPDXRef-"[idstring] where [idstring] is a unique string containing letters, numbers, ., and/or -

However, other characters were being included like underscores and forward slashes, this replaces them all with a dash. This should be safe to do without conflicts because our package IDs include the unique hash at the end.

This fixes: #949 and #952

### DIFF
--- a/internal/formats/common/spdxhelpers/spdxid.go
+++ b/internal/formats/common/spdxhelpers/spdxid.go
@@ -1,0 +1,13 @@
+package spdxhelpers
+
+import (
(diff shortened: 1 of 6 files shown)


spiffcs (approved):

spiffcs:
Nice! Thanks for following up on two of these issues. Great change.

Issue #953, opened 2022-04-13
Decoding of sparse CycloneDX does not set language


jonmcewen:
**What happened**:

When using grype to check a CycloneDX SBOM not produced by syft, Java vulnerabilities were not detected.

**What you expected to happen**:

Vulnerabilities should be found by language when there is no CPE and no syft metadata

**How to reproduce it (as minimally and precisely as possible)**:

Using a CycloneDX SBOM with minimal component info and known CVEs such as:

```JSON
{
  "name" : "log4j-core",
  "version" : "2.13.3",
  "purl" : "pkg:maven/org.apache.logging.log4j/log4j-core@2.13.3?type=jar",
  "type" : "library",
  "bom-ref" : "pkg:maven/org.apache.logging.log4j/log4j-core@2.13.3?type=jar"
},
```

Running grype will not find any CVEs.   

**Anything else we need to know?**:

PR coming shortly...

**Environment**:
- Output of `syft version`: v0.42.4
- OS (e.g: `cat /etc/os-release` or similar): *any*

PR #957, opened 2022-04-13
fix: #953 Derive language from pURL - https://github.com/anchore/syft…

jonmcewen:
…/issues/953

Signed-off-by: Jon McEwen <jon_mcewen@hotmail.com>

### DIFF
--- a/syft/pkg/cataloger/catalog.go
+++ b/syft/pkg/cataloger/catalog.go
@@ -69,6 +69,9 @@ func Catalog(resolver source.FileResolver, release *linux.Release, catalogers ..
 
 			// generate PURL (note: this is excluded from package ID, so is safe to mutate)
 			p.PURL = pkg.URL(p, release)
+			if p.Language == "" {
+				p.Language = pkg.LanguageFromPURL(p.PURL)
+			}
 
 			// create file-to-package relationships for files owned by the package
 			owningRelationships, err := packageFileOwnershipRelationships(p, resolver)
--- a/syft/pkg/cataloger/catalog.go
+++ b/syft/pkg/cataloger/catalog.go
@@ -69,6 +69,7 @@ func Catalog(resolver source.FileResolver, release *linux.Release, catalogers ..
 
 			// generate PURL (note: this is excluded from package ID, so is safe to mutate)
 			p.PURL = pkg.URL(p, release)
+
 			if p.Language == "" {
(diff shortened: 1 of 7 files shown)


spiffcs:
Thanks for the contribution @jonmcewen. Approved and running the checks now

spiffcs:
@jonmcewen I'm taking a look at the integration testing here to see what updates have to go in for this PR to be ok to merge

jonmcewen:
Thanks @spiffcs . Let me know if I need to do more. I had tests failing on master before the change so assumed that was something to do with my laptop setup

spiffcs:
Just a quick update on this PR. The reason the tests are failing is because we test an `encode/decode/encode` cycle of the formats. The original encoding of the cyclonedx format will NOT include certain properties that get included after a decode.

Example:

Original encode will not do the PURL language check so properties like below
```
      	        {
         	          "name": "syft:package:language",
         	          "value": "UnknownLanguage"
         	    },
```

The reason they these properties are included on the second pass is that we decode the original. This decode does the new language check, which then encodes the new property which shows we could not find a language.

I'm going to ask our team today on if we want to include ("UnknownLanguage") properties in the original encode OR if we want to redact it and not include it when doing this language from CPE check.

spiffcs (approved):
Great addition to syft's ability to generate more complete information from outside SBOM.

Looking for a second from @anchore/tools

syft/pkg/cataloger/catalog.go
```diff
@@ -69,6 +69,9 @@ func Catalog(resolver source.FileResolver, release *linux.Release, catalogers ..
 
 			// generate PURL (note: this is excluded from package ID, so is safe to mutate)
 			p.PURL = pkg.URL(p, release)
+			if p.Language == "" {
```
wagoodman, line 72: nit: an extra comment explaining why this is here would be good since there is some high-level reasoning for it that isn't apparent.

spiffcs:
Looks like my changes made the integration tests FAIL. I need to update the assertions so it sees it's ok now to have`UnknownLanguage`

spiffcs:
@jonmcewen I'll make sure to get this over the line today. We really appreciate the contribution and the thought you put into this PR. I'm just trying to get some of the encode/decode patterns for different formats reconciled with this change.

jonmcewen:
@spiffcs many thanks for getting this in and making it a pleasant contribution experience :-)

PR #961, opened 2022-04-15
Update GitHub format package_url and correlator

kzantow:
The GitHub dependency snapshot format has a couple updates encompassed here: changing the `purl` field to `package_url`, `detector.name` is required (but we are already providing that as `syft`), and `job.name` has been deprecated in favor of `job.correlator`. The latter will be filled out by the `sbom-action`, as it requires action run information.


spiffcs (approved):

jonasagx (approved):
I understood this doesn't add breaking changes, 👍

Issue #962, opened 2022-04-18
Broken shell completion - Bash


DanHam:
**What happened**:

I get an error when I attempt to source the generated Bash shell completions:

```bash
$ source <(syft completion bash)
complete: usage: complete [-abcdefgjksuv] [-pr] [-DEI] [-o option] [-A action] [-G globpat] [-W wordlist] [-F function] [-C command] [-X filterpat] [-P prefix] [-S suffix] [name ...]
```

**How to reproduce it (as minimally and precisely as possible)**:

```bash
$ source <(syft completion bash)
```

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: `syft 0.44.1`

This also occurs with `0.44.0` and `0.43.2`

- OS (e.g: `cat /etc/os-release` or similar):
```bash
$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="35 (Cinnamon)"
ID=fedora
VERSION_ID=35
VERSION_CODENAME=""
PLATFORM_ID="platform:f35"
PRETTY_NAME="Fedora Linux 35 (Cinnamon)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:35"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f35/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=35
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=35
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Cinnamon"
VARIANT_ID=cinnamon
```

spiffcs:
@DanHam we recently merged a PR that addresses this bug. If you run from main (source <(go run cmd/syft/main.go completion bash) there should be no more error.
![Screen Shot 2022-04-27 at 5 47 45 PM](https://user-images.githubusercontent.com/32073428/165636747-15f4c008-a99e-471e-ab0c-2459b045a346.png)

DanHam:
@spiffcs 

[...]

PR #965, opened 2022-04-19
refactor command package to remove globals and add dependency injection

spiffcs:
## Summary
This PR attempts to update the command package so that Syft moves away from global package variables and instead is architected around command constructors with functional options. 

## Changes
- package `cmd` separated into `cli` with `commands.go` organizing cli construction. Individual commands have constructors along with new packages named as command name. 
EX:`import github.com/anchore/syft/cmd/syft/cli/attest` includes `attest.Run` as the execution path.
- package `options` built to describe options that can be decorated onto built commands
- `eventloop` given its own package
- remove cli options and elevate to top of `config.Application` struct
- update application configuration to use dependency injection of viper and config object
- 

TODO:
- [x] Update tests in new packages to cover constructors
- [x] Review removed code and make commits that restore parity with main
- [x] Review changes and add summary of changes to config package
- [x] Review completion changes and fix generation for https://github.com/anchore/syft/issues/962
Screen Shot 2022-04-22 at 11 11 31 AM


.bouncer.yaml
```diff
@@ -5,6 +5,7 @@ permit:
   - MPL.*
   - ISC
 ignore-packages:
+  - .
```
spiffcs, line 8: This is needed since our main.go package has moved from `. =>cmd/syft/main.go`

Makefile
```diff
@@ -147,7 +147,7 @@ lint-fix: ## Auto-format all source code + run golangci lint fixers
 
 .PHONY: check-licenses
 check-licenses: ## Ensure transitive dependencies are compliant with the current license policy
-	$(TEMPDIR)/bouncer check
+	$(TEMPDIR)/bouncer check ./cmd/syft
```
spiffcs, line 150: @wagoodman if I provide the path to the main package does the bouncer check things at the top level like `internal` or  `syft`?

cmd/syft/cli/attest/attest.go
```diff
@@ -0,0 +1,214 @@
...
+		PassFunc: passFunc,
+	}
+
+	sv, err := sign.SignerFromKeyOpts(ctx, "", "", ko)
+	if err != nil {
+		return err
+	}
+	defer sv.Close()
+
+	eventBus := partybus.NewBus()
```
spiffcs, line 79: Is just in time initialization ok for these? Previously they were global vars but not used widely across the package. There was mention in a previous meeting about allowing a user to configure their own bus? Maybe not for this PR, but just want to get clarity on the mechanics of moving this from `global => local per command`

cmd/syft/cli/attest/attest.go
```diff
@@ -0,0 +1,214 @@
...
+var allowedAttestFormats = []sbom.FormatID{
+	syftjson.ID,
+	spdx22json.ID,
+	cyclonedxjson.ID,
+}
+
+var intotoJSONDsseType = `application/vnd.in-toto+json`
+
+func Run(ctx context.Context, app *config.Application, args []string) error {
+	// We cannot generate an attestation for more than one output
```
spiffcs, line 45: The order of execution/organization of this function will be changing as a consequence of the keyless attestation PR. I admit it's a little unclear at the moment and the functions `Run` and `execWorker` don't have very clear responsibilityes.

cmd/syft/cli/options/attest.go
```diff
@@ -0,0 +1,28 @@
...
+var _ Interface = (*AttestOptions)(nil)
+
+func (o *AttestOptions) AddFlags(cmd *cobra.Command, v *viper.Viper) error {
+	cmd.PersistentFlags().StringVarP(&o.Key, "key", "", "cosign.key",
+		"path to the private key file to use for attestation")
+
+	return bindAttestationConfigOptions(cmd.PersistentFlags(), v)
+}
+
+func bindAttestationConfigOptions(flags *pflag.FlagSet, v *viper.Viper) error {
```
spiffcs, line 22: Todo: I think the packages command `password` for anchore enterprise conflicts with `password` for the attest command options. I need to check and see which overrides the new decorator's behavior and find a solution to distinguish between the two.

cmd/syft/cli/packages.go
```diff
@@ -0,0 +1,98 @@
...
+
+	err := po.AddFlags(cmd, v)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	return cmd
+}
+
+func newLogWrapper(app *config.Application) {
```
spiffcs, line 84: probably should move this to its own file so it can be shared among CLI constructors.

cmd/syft/cli/packages/packages.go
```diff
@@ -0,0 +1,240 @@
...
+	}()
+
+	// could be an image or a directory, with or without a scheme
+	userInput := args[0]
+	si, err := source.ParseInput(userInput, app.Platform, true)
+	if err != nil {
+		return fmt.Errorf("could not generate source input for packages command: %w", err)
+	}
+
+	eventBus := partybus.NewBus()
```
spiffcs, line 49: Same comment about just in time initialization for this part of the code

cmd/syft/cli/packages/packages.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if err := c.Import(context.Background(), importCfg); err != nil {
+		return fmt.Errorf("failed to upload results to host=%s: %+v", app.Anchore.Host, err)
+	}
+
+	return nil
+}
+
+func GenerateSBOM(src *source.Source, errs chan error, app *config.Application) (*sbom.SBOM, error) {
```
spiffcs, line 150: Move this to top since it's exported

internal/config/application.go
```diff
@@ -29,13 +32,15 @@ type parser interface {
 
 // Application is the main syft application configuration.
 type Application struct {
-	ConfigPath         string             `yaml:",omitempty" json:"configPath"`                                                         // the location where the application config was read from (either from -c or discovered while loading)
```
spiffcs, line 32: Maybe keep `ConfigPath`

internal/config/application.go
```diff
@@ -29,13 +32,15 @@ type parser interface {
...
 	Outputs            []string           `yaml:"output" json:"output" mapstructure:"output"`                                           // -o, the format to use for output
 	File               string             `yaml:"file" json:"file" mapstructure:"file"`                                                 // --file, the file to write report output to
-	Quiet              bool               `yaml:"quiet" json:"quiet" mapstructure:"quiet"`                                              // -q, indicates to not show any status output to stderr (ETUI or logging UI)
 	CheckForAppUpdate  bool               `yaml:"check-for-app-update" json:"check-for-app-update" mapstructure:"check-for-app-update"` // whether to check for an application update on start up or not
 	Anchore            anchore            `yaml:"anchore" json:"anchore" mapstructure:"anchore"`                                        // options for interacting with Anchore Engine/Enterprise
-	CliOptions         CliOnlyOptions     `yaml:"-" json:"-"`                                                                           // all options only available through the CLI (not via env vars or config)
```
spiffcs, line 38: TODO: There is a small conflict here between the input as indicated by the user and the translation used for the application. EX: `scope string ==> scope internal usage` `verbosity uint ==> verbosity logging config`

Maybe we can reconcile this

internal/config/config.go
```diff
@@ -0,0 +1 @@
+package config
```
spiffcs, line 1: TODO: reconcile changes to internal/config/application.go and this file

schema/cyclonedx/Makefile
```diff
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := validate-schema
 .PHONY: validate-schema
 validate-schema:
-	go run ../../main.go ubuntu:latest -vv -o cyclonedx > bom.xml
```
spiffcs, line 4: TODO check makefile for other instances of program root changing

internal/config/application.go
```diff
@@ -7,17 +7,20 @@ import (
 	"reflect"
```
spiffcs, line 7: This file needs a more through review before going from WIP ==> Review Ready

spiffcs (commented):
Self Review Comments

cmd/syft/cli/options/packages.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/anchore/syft/internal/formats/table"
+	"github.com/anchore/syft/syft"
+	"github.com/anchore/syft/syft/pkg/cataloger"
+	"github.com/anchore/syft/syft/source"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+type PackagesOptions struct {
```
spiffcs, line 15: refactor and remove these =)

internal/config/application.go
```diff
@@ -49,72 +54,50 @@ type Application struct {
...
-	}
-}
-
-func (cfg *Application) parseConfigValues() error {
+func (a *Application) parseConfigValues() error {
 	// parse application config options
 	for _, optionFn := range []func() error{
-		cfg.parseUploadOptions,
-		cfg.parseLogLevelOption,
-		cfg.parseFile,
```
spiffcs, line 107: This needs to be added back

test/cli/power_user_cmd_test.go
```diff
@@ -21,13 +21,13 @@ func TestPowerUserCmdFlags(t *testing.T) {
...
 			},
 		},
-		{
-			name: "json-output-flag-fails",
-			args: []string{"power-user", "-o", "json", "docker-archive:" + getFixtureImage(t, "image-pkg-coverage")},
-			assertions: []traitAssertion{
-				assertFailingReturnCode,
-			},
-		},
+		//{
```
spiffcs, line 24: Because poweruser is being decorated with the packages flags output is now a valid option. 

TODO: update the `writer` function so poweruser can do multiple formats. This test becomes invalid.

cmd/syft/cli/commands.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/wagoodman/go-partybus"
+)
+
+const indent = "  "
+
+// New constructs the `syft packages` command, aliases the root command to `syft packages`,
+// and constructs the `syft power-user` and `syft attest` commands. It is also responsible for
+// organizing flag usage and injeting the application config for each command.
```
jonasagx, line 28:
```suggestion
// organizing flag usage and injecting the application config for each command.
```

cmd/syft/cli/options/attest.go
luhring: Curious what the resolution was for this 😃

cmd/syft/cli/options/options.go
```diff
@@ -0,0 +1,11 @@
+package options
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+type Interface interface {
```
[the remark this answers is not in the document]
luhring: There's precedent for naming types `Interface` in the std lib: https://pkg.go.dev/sort#Interface

I think this is okay, if the `Interface` is **_the canonical type_** for which the `options` package exists.

test/cli/power_user_cmd_test.go
luhring: I want to make sure I follow what's happening here... I know we plan to remove the `power-user` command in the future. And IMHO it's not so bad if we support `-o json` for the command in the meantime (curious for your thoughts on this!).

Should we remove this test case if we don't plan to use it either now or later?

luhring (approved):
This is fantastic @spiffcs — what an upgrade for the code... 🚀 

Approving this! ✅ 

It looks like there's still some follow-up needed here. I asked a couple of questions, and I saw there were a few existing questions/nits that aren't answered yet. But nothing that's left seems like a big lift at this point.

I also have two overall questions re: this PR:

1. Since `main.go` has been moved to another package, this breaks installation for users that build from source using `go install github.com/anchore/syft@latest`. We don't list this installation method in our README documentation, so no change is needed here. But it is a nice feature of Go 1.16+, and I've used it several times. Could we do a cursory check of uses of this installation method just so we have a heads up on potential broken workflows? (My guess is we're not using this internally.)

2. I see some PR comments that have "TODO", and the comments are resolved, and the code hasn't changed. I want to make sure I parse that correctly. Does that mean that no code change is needed at this point? Such as it was just a thing to look into, but you've done that outside of the code?

cmd/syft/cli/options/attest.go
spiffcs: This is a bug that exists on main at the moment. I have not found a good solution here besides renaming the flags. Sorry for resolving this comment that was an oversight on my part.

cmd/syft/cli/options/packages.go
spiffcs: We still need the hard values here for decoding the flags as well as a receiver for the method. There is a future refactor that possibly removes these, but it's not for this PR.

test/cli/power_user_cmd_test.go
spiffcs: Good callout here as I did not circle back on this comment. The solution here is to just make -o json the ONLY valid format for power-user. We can then remove this test.

It's currently hard coded here:
https://github.com/anchore/syft/blob/c270ee2a023f2ed3bacbd847bbeb6583f7db9ddc/cmd/power_user.go#L76-L81

spiffcs:
> This is fantastic @spiffcs — what an upgrade for the code... 🚀
> 
> Approving this! ✅
> 
> It looks like there's still some follow-up needed here. I asked a couple of questions, and I saw there were a few existing questions/nits that aren't answered yet. But nothing that's left seems like a big lift at this point.
> 
> I also have two overall questions re: this PR:
> 
> 1. Since `main.go` has been moved to another package, this breaks installation for users that build from source using `go install github.com/anchore/syft@latest`. We don't list this installation method in our README documentation, so no change is needed here. But it is a nice feature of Go 1.16+, and I've used it several times. Could we do a cursory check of uses of this installation method just so we have a heads up on potential broken workflows? (My guess is we're not using this internally.)

> 2. I see some PR comments that have "TODO", and the comments are resolved, and the code hasn't changed. I want to make sure I parse that correctly. Does that mean that no code change is needed at this point? Such as it was just a thing to look into, but you've done that outside of the code?

There are some I reconciled outside the code and realized they were not as big of a deal as when I made the comment. I think I responded to each conversation you commented on so we can follow up on those threads. 

TLDR: you're right and thanks for pointing those out. I think they got resolved accidentally as I was going through some of my previous self review.

cmd/syft/cli/options/attest.go
spiffcs: Sorry @luhring I misspoke here. Attest does NOT provide a flag for password and there is no conflict. Passwords can only be set by ENV variable, passed in via the TUI, or set in the config.

We do not allow for plaintext on the CLI.

PR #967, opened 2022-04-21
Fix `github-json` output option

StevenMaude:
This PR fixes the following problem in Syft 0.44.1:

```
$ syft packages . -o github-json
1 error occurred:
	* bad output format: 'github-json'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ Where the `formats` available are:
 - `cyclonedx-json`: A JSON report conforming to the [CycloneDX 1.4 specification](https://cyclonedx.org/specification/overview/).
 - `spdx-tag-value`: A tag-value formatted report conforming to the [SPDX 2.2 specification](https://spdx.github.io/spdx-spec/).
 - `spdx-json`: A JSON report conforming to the [SPDX 2.2 JSON Schema](https://github.com/spdx/spdx-spec/blob/v2.2/schemas/spdx-schema.json).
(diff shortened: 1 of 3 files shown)


StevenMaude:
Two things:

1. I'm testing Syft out and noticed this issue while trying out all the output options. I've not looked closely at the codebase at all. Feel free to tell me this is entirely the wrong approach :smiley:
2. I added a note to the docs, but couldn't find a suitable link/reference for the `github` format in #836 (nor from a quick search around). If there is a suitable link, I'm happy to add it to f82bcdedbd434116471a573c68d72f470ea057fe.

kzantow:
@StevenMaude this format isn't completely finalized. I'm not sure we should treat it as such until it is finalized, but at such time `github-json` sounds like an appropriate option 👍

jonasagx:
@StevenMaude thanks for fixing this. Your PR has one last thing to pass our tests, [there is a code formatting issue](https://github.com/anchore/syft/runs/6281624806?check_suite_focus=true), to fix it just run the following command from the project directory in your local machine:
```sh
make lint-fix
```
Then commit the changes and it should be good. I will be here if you need help.

StevenMaude:
@jonasagx: thanks for your patience! Hopefully I've wrangled it, I think it was just a `gofmt` issue in the end.

jonasagx (approved):

PR #985, opened 2022-05-03
unblock timeout for power-user select CLI tests

wagoodman:
The power user secrets CLI tests we're blocking and timing out when trying to access the same fixture contents. Creating a separate secret fixtures directory with different data unblocks the timeout and allows the tests to operate independently of each other.


test/cli/utils_test.go
```diff
@@ -103,15 +103,15 @@ func runCosign(t testing.TB, env map[string]string, args ...string) (*exec.Cmd,
...
+	stdout, stderr := runCommand(t, cmd, env)
 	return cmd, stdout, stderr
 }
 
 func getCosignCommand(t testing.TB, args ...string) *exec.Cmd {
 	return exec.Command(filepath.Join(repoRoot(t), ".tmp/cosign"), args...)
 }
 
-func runCommand(cmd *exec.Cmd, env map[string]string) (string, string) {
+func runCommand(t testing.TB, cmd *exec.Cmd, env map[string]string) (string, string) {
```
kzantow, line 114: I think this `t testing.TB` was just added for debugging, and isn't being used? should it be removed again?

luhring (approved):

test/cli/utils_test.go
spiffcs: removed

kzantow (approved):

PR #981, opened 2022-05-03
read Go main module version as is - (devel)

jonasagx:
closes https://github.com/anchore/syft/issues/959

Signed-off-by: Jonas Galvão Xavier <jonasx@anchore.com>

PR #986, opened 2022-05-04
golang cataloger - main module version as is

jonasagx:
https://github.com/anchore/syft/pull/981 overwrites the main version, which doesn't follow Syft's design philosophy of reading data fields as they are. It just so happens that we know the version and it is a constant.

This PR fixes that behavior by reading the main module version as provided by go's BuildInfo.


luhring (approved):
LGTM!

PR #989, opened 2022-05-08
add context timeout to long chain operations

jonasagx:
Fix for integration test timeout.

By adding context cancelation, to operations that _might_ get stuck, this PR fixes the timeout problem. The remaining question is what exactly is stuck? We might need more test coverage to handle that.

Signed-off-by: Jonas Xavier <jonasx@anchore.com>

PR #991, opened 2022-05-09
Update SPDX license list to v3.17

kzantow:
All tests will start failing until this is updated - note this may need to be merged with #989 due to a different test failure.


kzantow:
This was merged as part of #989

PR #993, opened 2022-05-09
Fix potential race condition during event subscription

kzantow:
I hastily merged #989 to try to work around a test timeout issue we've been seeing, but it did not seem to entirely solve the problem - only somewhat reduce it. This PR reverts those changes (which you could [compare to d2d532f](https://github.com/anchore/syft/compare/d2d532f4a852067c9ccba2329a8105ac5479e2ae...kzantow-anchore:root-cause-analysis-of-test-timeouts#diff-d71da09dda3dcb00b18044c8915e29aefc4ef51518792b7ce084924940e0f6cb) to see a more "accurate" diff of this PR) and instead adds a timeout that calls `cmd.Process.Signal(syscall.SIGABRT)`, which kills the process but ALSO includes a stack trace, so it is possible to determine where a CLI call is actually hung. It also adds a retry for these failed calls, which adds `-vv` to attempt to get more information. It was also identified that the most likely culprit is `eventBus.Subscribe()` being called _after_ an `exec*` call; this PR also corrects this order of operations for all `eventloop`s.

### DIFF
--- a/test/cli/utils_test.go
+++ b/test/cli/utils_test.go
@@ -109,8 +109,8 @@ func runSyftSafe(t testing.TB, env map[string]string, args ...string) (*exec.Cmd
 }
 
 func runSyftCommand(t testing.TB, env map[string]string, expectError bool, args ...string) (*exec.Cmd, string, string) {
-	timeout := 1000 * time.Millisecond
-	ctx, cancel := context.WithTimeout(context.Background(), timeout+100)
+	timeout := 90 * time.Second
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 
 	timer := make(chan bool, 1)
 	defer func() {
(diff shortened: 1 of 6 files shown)


test/cli/utils_test.go
```diff
@@ -95,23 +81,71 @@ func runSyftInDocker(t testing.TB, env map[string]string, image string, args ...
...
 
 	// we should not have tests reaching out for app update checks
 	env["SYFT_CHECK_FOR_APP_UPDATE"] = "false"
 
-	stdout, stderr := runCommand(cmd, env)
+	timeout := func() {
+		select {
+		case <-cancel:
+			return
+		case <-time.After(60 * time.Second):
```
jonasagx, line 114:
```suggestion
		case <-time.After(time.Minute):
```

jonasagx: Total nit

kzantow: Yeah, I intentionally did this so it's easier to adjust the number of seconds.

PR #996, opened 2022-05-10
Read Description from dpkg status files

ryanmoran:
There are a number of fields that might appear in the status file for a Debian package. This PR adds the `Description` field to the set of fields parsed when cataloging Debian operating systems. This change does not include the addition of this field in any of the SBOM output formats (as far as I can tell), and is initially meant for consumption via library users taking advantage of Syft's package cataloging features. I'm not sure whether it makes sense to include it in SBOM output in the future. I don't specifically have a need for it, but I would like to see it parsed and returned within the `sbom.SBOM` object.

### DIFF
--- a/syft/pkg/cataloger/deb/cataloger_test.go
+++ b/syft/pkg/cataloger/deb/cataloger_test.go
@@ -43,6 +43,10 @@ func TestDpkgCataloger(t *testing.T) {
 						Architecture:  "all",
 						Maintainer:    "Steve Langasek <vorlon@debian.org>",
 						InstalledSize: 1016,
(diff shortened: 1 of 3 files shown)


spiffcs:
Thanks for the PR @ryanmoran! Would you be able to merge the latest changes from main and push to your branch? We recently fixed a CI test and those changes are not present in this PR which is why I think the validations are failing.

spiffcs (approved):

spiffcs:
🟢 Thanks for adding the field will merge this now. If there are test issues I will jump in to try and resolve them

PR #1004, opened 2022-05-18
Malformed licenses field in package json warn not skip

houdini91:
PR includes small fix to the npm cataloger -
**Bug**: Packages with unstandard `licenses` field are dropped.

**Root**: Package json structure unmarshaling error.
```
		if err := dec.Decode(&p); err == io.EOF {
			break
		} else if err != nil {
			return nil, nil, fmt.Errorf("failed to parse package.json file: %w", err)
		}
```
**Suggestion**: Unmarshal `licences` field in raw bytes and unmarshal it sepritly so that the package still gets added to the resulted sbom.


jonasagx (approved):
LGTM - thank you @houdini91 🎉

Issue #950, opened 2022-04-13
Deprecated SPDX license (GPL-2.0+)


vargenau:
**What happened**:

The PackageLicenseConcluded and PackageLicenseDeclaredfor contain a deprecated license "GPL-2.0+".

```
PackageName: python-debian
SPDXID: SPDXRef-Package-python-python-debian-f14b246cf6138fa4
PackageVersion: 0.1.38
PackageDownloadLocation: NOASSERTION
FilesAnalyzed: false
PackageLicenseConcluded: GPL-2.0+
PackageLicenseDeclared: GPL-2.0+
```

**What you expected to happen**:

"GPL-2.0+" should be replaced by "GPL-2.0-or-later"

**How to reproduce it (as minimally and precisely as possible)**:

```
syft docker:fsfe/reuse:latest -o spdx-tag-value
```

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`:

```
Application:        syft
Version:            0.44.0
JsonSchemaVersion:  3.2.2
BuildDate:          2022-04-12T18:59:04Z
GitCommit:          b46d044d7e3afc2ee09fd5ff26635f783628126e
GitDescription:     v0.44.0
Platform:           linux/amd64
GoVersion:          go1.18
Compiler:           gc
```

- OS (e.g: `cat /etc/os-release` or similar):
```
PRETTY_NAME="Ubuntu 21.10"
NAME="Ubuntu"
VERSION_ID="21.10"
VERSION="21.10 (Impish Indri)"
VERSION_CODENAME=impish
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=impish
```

jonasagx:
Thank you for the suggestion @vargenau I wrote an implementation for that at https://github.com/anchore/syft/pull/1009

vargenau:
This was only an example of deprecated license.

"GPL-1.0+" should be replaced by "GPL-1.0-or-later"
"GPL-2.0+" should be replaced by "GPL-2.0-or-later"
"GPL-3.0+" should be replaced by "GPL-3.0-or-later"

"LGPL-2.0+" should be replaced by "LGPL-2.0-or-later"
[...]

PR #1009, opened 2022-05-20
Overwrite deprecated SPDX licenses automatically

jonasagx:
SPDX deprecated `GPL-2.0+` as a license identifier, as suggested by https://github.com/anchore/syft/issues/950 we can replace it with `GPL-2.0-or-later`, in this PR we do that while also adding a way to overwrite other licenses in the future.

Fix https://github.com/anchore/syft/issues/950


internal/spdxlicense/license_test.go
```diff
@@ -20,8 +20,8 @@ func TestIDParse(t *testing.T) {
 			"GPL-2.0",
 		},
 		{
-			"GPL-2+",
-			"GPL-2.0+",
+			"GPL-1+",
```
jonasagx, line 23: With GPL-2+ replaced this test started to fail, therefore this change

vargenau:
Please check my comment in https://github.com/anchore/syft/issues/950

There are many more deprecated licenses.

internal/spdxlicense/license_test.go
luhring: Can you elaborate more? What makes this change the _right_ change, beyond the fact that the tests pass now?

jonasagx: With this PR `GPL-2.0+` is overwritten by `GPL-2.0-or-later`, the test above is meant to validate that the code handles `GPL-1+` -> `GPL-1.0+`, where the code should imply the zero digit, which might be omitted in a package.

The case for `GPL-2.0+` -> `GPL-2.0-or-later` got its own test at https://github.com/jonasagx/syft/blob/deprecate-gpl-2%2B-license/internal/spdxlicense/generate/generate_license_list_test.go#L149

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -37,16 +37,40 @@ var licenseIDs = map[string]string{
...
+
+	if l.OSIApproved != other.OSIApproved {
+		return false
+	}
+
+	if len(l.SeeAlso) != len(other.SeeAlso) {
+		return false
+	}
+
+	for i := 0; i < len(l.SeeAlso); i++ {
```
luhring, line 62: nit: This could be a more idiomatic `for`/`range` loop

internal/spdxlicense/license_test.go
luhring: (That link to `generate_license_list_test.go` doesn't test `GPL-2.0+` -> `GPL-2.0-or-later`.)

>where the code should imply the zero digit

Isn't this tested by the prior test case as well? (it looks like this is exercised in both the 2nd and 3rd test case, if I'm following correctly)

internal/spdxlicense/generate/generate_license_list_test.go
```diff
@@ -144,3 +145,50 @@ func TestFindLicenseVersion(t *testing.T) {
...
+		"abc-1.0.0+":         "ABC-1.0-Or-later",
+		"abc-1-or-later":     "ABC-1.0-Or-later",
+		"abc-1.0-or-later":   "ABC-1.0-Or-later",
+		"abc-1.0.0-or-later": "ABC-1.0-Or-later",
+	}
+
+	licenses := processSPDXLicense(results)
+	for k, v := range licenses {
+		e := expected[k]
+		// t.Logf("%s --> %s, e: %s", k, v, e)
```
luhring, line 191: This looks like dead code, can it be removed?

internal/spdxlicense/generate/generate_license_list_test.go
```diff
@@ -144,3 +145,50 @@ func TestFindLicenseVersion(t *testing.T) {
 		})
 	}
 }
+
+func TestReplaceDeprecatedLicenses(t *testing.T) {
```
luhring, line 149: Question: Should this also test the case where a license is deprecated but no other license should be used to replace it?

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -139,9 +175,26 @@ func processSPDXLicense(result LicenseList) map[string]string {
 		}
 	}
 
+	for k, v := range licenseIDs {
+		cleanID := strings.ToLower(v)
```
luhring, line 179: Why do we recalculate the `cleanID`? I thought the `k` value was the clean ID, based on looking above at line 151

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -102,24 +126,36 @@ func run() error {
...
 			log.Fatalf("duplicate license ID found: %q", cleanID)
 		}
 		licenseIDs[cleanID] = l.ID
+
+		if l.Deprecated {
+			replacement := findEquivalentLicense(l, result.Licenses)
+			if replacement != nil {
+				overwrite[cleanID] = replacement.ID
+			}
+		}
```
luhring, line 158: Is there a reason we're building this `overwrite` collection here and waiting to use it until later? What's the reason for not replacing deprecated licenses right away before we generate more permutations below?

jonasagx: Good question: The order of variations/permutations of a license ID matters because of we how shuffle its digits, that is because the permutation code can generate the same value for two difference licenses, for example: The licenses `ABC-1.0` and `ABC-1.1` can both map to `ABC-1`, so we need to guarantee the order they are created to avoid mapping them wrongly. So we use a sorted list.

To overwrite deprecated licenses during the first pass we would later on rely on map order, [which in go is not consistent by design](https://stackoverflow.com/a/55925880).

I think we should revisit how we permute license IDs, but that is for another PR, as this code generates our global list of license IDs and that is delicate.

internal/spdxlicense/generate/generate_license_list_test.go
jonasagx: Good call, adding it

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -139,9 +175,26 @@ func processSPDXLicense(result LicenseList) map[string]string {
...
+			licenseIDs[k] = remap
+		}
+	}
+
 	return licenseIDs
 }
 
+func findEquivalentLicense(deprecated License, allLicenses []License) *License {
+	for _, l := range allLicenses {
+		if l.isEquivalent(deprecated) {
```
wagoodman, line 196: we want to return an equivalent license that is not deprecated, this is only guaranteed to return a equivalent license. Should there be another condition here?

jonasagx: Good catch, yes I will add the extra condition that it shouldn't be deprecated either. Thank you.

internal/spdxlicense/generate/generate_license_list.go
jonasagx: Because `v` is used as "primary key", that is: if `k0`, `k1`, `kn` point to v0, and in the overwrite map `v0` should point to `v1`, than the loop aboce will apply that to `k0`, `k1`, `kn`.

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -37,16 +37,46 @@ var licenseIDs = map[string]string{
...
+	Text        string   `json:"licenseText"`
+	Deprecated  bool     `json:"isDeprecatedLicenseId"`
+	OSIApproved bool     `json:"isOsiApproved"`
+	SeeAlso     []string `json:"seeAlso"`
+}
+
+func (l License) isEquivalent(other License) bool {
+	// With this equality check we guarantee the call order doesn't matter:
+	// a.isEquivalent(b) AND b.isEquivalent(a) yield the same value
+	if l.Deprecated == other.Deprecated {
```
wagoodman, line 52: this is really a concern of the caller and not within a function named "isEquivelent" (where deprecation has no barring on equivalency)

internal/spdxlicense/generate/generate_license_list.go
wagoodman: I think dan is right in the sense that `k` is no different than `cleanID`; take:
https://github.com/anchore/syft/pull/1009/files#diff-122a77148a7f1f5f9c1e3e3715c303dd5ce8d0ba75d9348ec00a1bfd0a46c305L117-L122
```golang
	for _, l := range result.Licenses {
		cleanID := strings.ToLower(l.ID)
		...
		licenseIDs[cleanID] = l.ID
                ...
         }
```

So later when doing:
```
for k, v := range licenseIDs {
   ...
}
```
`k` is indeed `cleanID`. This could be written like so for more clarity too:
```
for cleanID, id := range licenseIDs {
   ...
}
```

Shout out if I'm missing something

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -139,9 +187,26 @@ func processSPDXLicense(result LicenseList) map[string]string {
...
+		cleanID := strings.ToLower(v)
+		if remap, exists := overwrite[cleanID]; exists {
+			licenseIDs[k] = remap
+		}
+	}
+
 	return licenseIDs
 }
 
+func findEquivalentLicense(deprecated License, allLicenses []License) *License {
```
wagoodman, line 200: this seems important enough to have an explicit unit test for this (especially when adding the extra condition in)

internal/spdxlicense/generate/generate_license_list.go
jonasagx: Alright this can be rewritten as 
```go
	for cleanID, realID := range licenseIDs {
		if replacementID, exists := overwrite[strings.ToLower(realID)]; exists {
			licenseIDs[cleanID] = replacementID
		}
	} 
```
This code is replacing `cleanIDs` based on what value they point to. so if permutation `abc-1.0.0.0+` points to `ABC-1.0+`, and the overwrite map says `ABC-1.0+` should point to `ABC-1.0-Or-Later` that will happen for `abc-1.0.0.0+`, but also for all of its permutations, since they all point to `ABC-1.0+`.

Mind that `ABC-1.0+` is stored in the overwrite map as `abc-1.0+` (lower case),  so the use of `strings.ToLower(realID)` to find it in the map.

I suppose the reuse of the cleanID name, as in sanitized id for map indexation, caused more confusion than clarification.

internal/spdxlicense/generate/generate_license_list.go
jonasagx: done

internal/spdxlicense/license_test.go
jonasagx: You're right. Sorry about the broken link above. I've updated the test to [[1](https://github.com/anchore/syft/pull/1009/files#diff-ef65afb470a7847837074d4821ae6f10bdd7afcb46641ad73cdaf132b418e1beL20-R25)].
```diff
		{
 			"GPL-2",
 			"GPL-2.0-only",
 		},
 		{
 			"GPL-2+",
 			"GPL-2.0-or-later",
 		},
```

spiffcs (approved):

Issue #1014, opened 2022-05-26
Panic from Syft cyclonedx format method


cpendery:
**What happened**:
Syft's `toSyftModel` produces a panic on an empty components list

**What you expected to happen**:
It should report and error instead like it does for all other missing fields that it requires to run properly

**How to reproduce it (as minimally and precisely as possible)**:
```go
package main

import (
	"bytes"
	"encoding/json"

	"github.com/CycloneDX/cyclonedx-go"
	"github.com/anchore/syft/syft"
)

func main() {
	f := syft.FormatByID(syft.CycloneDxJSONFormatID)
	bom := cyclonedx.BOM{
		BOMFormat:   "CycloneDX",
		SpecVersion: "1.4",
		Version:     1,
	}
	bomBytes, _ := json.Marshal(&bom)
	f.Decode(bytes.NewReader(bomBytes))
}
```

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: v0.46.3
- OS (e.g: `cat /etc/os-release` or similar): Windows 11 Pro, 21H2

spiffcs:
Thanks for reporting the error and a way to reproduce it @cpendery. Let me take a quick look here and get a fix in for this bug.

PR #1015, opened 2022-05-26
fix: add component list to prevent cyclone-dx panic

cpendery:
## 📝 Description
Add check for empty components in `toSyftFormat` method to catch potential panic

fixes https://github.com/anchore/syft/issues/1014

### DIFF
--- a/internal/formats/common/cyclonedxhelpers/decoder.go
+++ b/internal/formats/common/cyclonedxhelpers/decoder.go
@@ -31,7 +31,9 @@ func GetValidator(format cyclonedx.BOMFileFormat) sbom.Validator {
 
 func GetDecoder(format cyclonedx.BOMFileFormat) sbom.Decoder {
 	return func(reader io.Reader) (*sbom.SBOM, error) {
-		bom := &cyclonedx.BOM{}
+		bom := &cyclonedx.BOM{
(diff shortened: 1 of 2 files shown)


internal/formats/common/cyclonedxhelpers/decoder.go
```diff
@@ -45,8 +45,8 @@ func GetDecoder(format cyclonedx.BOMFileFormat) sbom.Decoder {
 }
 
 func toSyftModel(bom *cyclonedx.BOM) (*sbom.SBOM, error) {
-	if bom == nil {
```
spiffcs, line 48: Thanks for adding the nil check! 

Does it make sense here to error on no components?

https://github.com/cpendery/syft/blob/7d24424b58bc78e2fa354c76466a90dbd3ac88ac/internal/formats/common/cyclonedxhelpers/decoder.go#L34-L39

If we constructed a bom initially with an empty list of components we would not see the nil pointer here and coule return/cover cases to users where no comonents were found.

If we instead

cpendery: That makes sense! I updated the code to reflect the suggestion

spiffcs (approved):

Issue #780, opened 2022-01-28
Python package Name and pURL have errors in CycloneDX output


37b:
When generating a bom.xml in CycloneDX format the following entries were created

 ``` <components>
    <component type="library">
      <name></name>
      <version></version>
      <cpe>cpe:2.3:a:python-:python-:*:*:*:*:*:*:*:*</cpe>
      <purl>pkg:pypi/</purl>
      <properties>
        <property name="foundBy">python-package-cataloger</property>
        <property name="language">python</property>
        <property name="type">python</property>
        <property name="metadataType">PythonPackageMetadata</property>
        <property name="path">/usr/lib/python3/dist-packages/.wh.setuptools-52.0.0.egg-info</property>
        <property name="layerID">sha256:501fee28c6e4a55d3b43d4194b23128b14dee7b16025e629f601ff562ee98c18</property>
      </properties>
    </component>
    <component type="library">
      <name></name>
      <version></version>
      <cpe>cpe:2.3:a:python-:python-:*:*:*:*:*:*:*:*</cpe>
      <purl>pkg:pypi/</purl>
      <properties>
        <property name="foundBy">python-package-cataloger</property>
        <property name="language">python</property>
        <property name="type">python</property>
        <property name="metadataType">PythonPackageMetadata</property>
        <property name="path">/usr/lib/python3/dist-packages/.wh.wheel-0.34.2.egg-info</property>
        <property name="layerID">sha256:501fee28c6e4a55d3b43d4194b23128b14dee7b16025e629f601ff562ee98c18</property>
      </properties>
    </component>
```

Here it is in JSON

```
   {
      "type": "library",
      "name": "",
      "version": "",
      "cpe": "cpe:2.3:a:python-:python-:*:*:*:*:*:*:*:*",
      "purl": "pkg:pypi/",
      "properties": [
        {
          "name": "foundBy",
          "value": "python-package-cataloger"
        },
        {
          "name": "language",
          "value": "python"
        },
        {
          "name": "type",
          "value": "python"
        },
        {
          "name": "metadataType",
          "value": "PythonPackageMetadata"
        },
        {
          "name": "path",
          "value": "/usr/lib/python3/dist-packages/.wh.setuptools-52.0.0.egg-info"
        },
        {
          "name": "layerID",
          "value": "sha256:501fee28c6e4a55d3b43d4194b23128b14dee7b16025e629f601ff562ee98c18"
        }
      ]
    },
    {
      "type": "library",
      "name": "",
[...]

PR #1017, opened 2022-05-27
Drop whiteout/opaque files and warn when pkg lacks Name/Version

jonasagx:
This PR addresses two issues from https://github.com/anchore/syft/issues/780

## Syft indexes [whiteout and opaque files](https://github.com/opencontainers/image-spec/blob/main/layer.md#whiteouts)
Addressed by:
* Filtering those files out during directory resolver walk
* Making use of stereoscope support for identifying whiteout and opaque files.
* Added cli tests and integration tests 

### Questions
* Should the filtering pattern be extended to `ImageScheme`?
* By excluding files from file trees, and the final SBOM, does it create a security issue? 
  * We can let the users decide what to do via syft's config files, using an exclusion pattern

## Packages created without Name or Version don't give warnings
Name and Version are important fields, later used to generate CPEs, pURLs. We should warn when they are missing, since that will produce broken SBOMs

Addressed by:
* Added warn logs to package `SetID` (that function already had some warnings, so it looks like an acceptable place)
* Added integration tests to the warn messages  
* Removed log from GitHub format encoder in favor of `SetID`'s warn log.

Fix: https://github.com/anchore/syft/issues/780


internal/formats/github/encoder.go
```diff
@@ -148,7 +147,6 @@ func toGithubManifests(s *sbom.SBOM) Manifests {
 func dependencyName(p pkg.Package) string {
 	purl, err := packageurl.FromString(p.PURL)
 	if err != nil {
-		log.Warnf("Invalid PURL for package: '%s' PURL: '%s' (%w)", p.Name, p.PURL, err)
```
jonasagx, line 151: This log line seems unnecessary with the new warning that works for all formats

syft/pkg/package.go
```diff
@@ -33,6 +33,14 @@ func (p *Package) OverrideID(id artifact.ID) {
 }
 
 func (p *Package) SetID() {
+	if p.Name == "" {
```
jonasagx, line 36: Another approach is to create a  package constructor and refactor it all over the code base.

test/cli/package_missing_data_test.go
```diff
@@ -0,0 +1,37 @@
...
+			tb.Helper()
+			assert.Contains(tb, stderr, "python-package-cataloger: missing package name, that is necessary for further metadata extraction, such as CPEs and pURLs")
+			assert.Contains(tb, stderr, "python-package-cataloger: missing package version, that is necessary for further metadata extraction, such as CPEs and pURLs")
+		},
+		assertSuccessfulReturnCode,
+	}
+
+	for _, o := range formats {
+		t.Run(fmt.Sprintf("format:%s", o), func(t *testing.T) {
+			// TODO: does it make sense to add empty metadata for all catalogers?
```
jonasagx, line 25: Metatada here as in data fixtures, simulating empty description files, like PKG-INFO for python projects.

syft/pkg/package.go
```diff
@@ -33,6 +33,14 @@ func (p *Package) OverrideID(id artifact.ID) {
...
+		log.Warnf("%s: missing package name, that is necessary for further metadata extraction. Please take a look at %s", p.FoundBy, p.Locations.ToSlice())
```
jonasagx, line 37: adding `p.Locations.ToSlice()` to the warning log to inform user where the problem is.

jonasagx:
Closing this PR in favor of https://github.com/anchore/syft/pull/1021

PR #1025, opened 2022-06-03
bump cosign to v1.9.0 to resolve reporting of GHSA-66x3-6cw3-v5gj

westonsteimel:


spiffcs:
Looks like this is a large change to their library code! I'll take a look here as soon as I have time to see what we can refactor so we can stay in sync.

Thanks for the nudge that this should be updated

spiffcs (approved):

Issue #908, opened 2022-03-22
Specify the "main module" in Go binary metadata for packages


luhring:
### Background

As of #827, Syft captures a large amount of data in Go packages when cataloging Go binaries (built with Go 1.18+). For reference, here's an excerpt of Syft's analysis of Syft:

```json
"metadata": {
    "goBuildSettings": {
      "-compiler": "gc",
      "-ldflags": "-w -s -extldflags '-static' -X github.com/anchore/syft/internal/version.version=0.42.1 -X github.com/anchore/syft/internal/version.gitCommit=283db88dc4d40502219c80af36d073b2f1556007 -X github.com/anchore/syft/internal/version.buildDate=2022-03-21T19:32:33Z -X github.com/anchore/syft/internal/version.gitDescription=v0.42.1 ",
      "CGO_ENABLED": "0",
      "GOAMD64": "v1",
      "GOARCH": "amd64",
      "GOOS": "darwin",
      "vcs": "git",
      "vcs.modified": "false",
      "vcs.revision": "283db88dc4d40502219c80af36d073b2f1556007",
      "vcs.time": "2022-03-21T18:59:10Z"
    },
    "goCompiledVersion": "go1.18",
    "architecture": "amd64"
}
```

Still, not all packages will show this extra data. A package won't show this extra data if either (A) the package isn't the **[main module](https://go.dev/ref/mod#glos-main-module)** (i.e. it's a dependency of a main module), or (B) the binary was built with Go 1.17 or earlier.

This has a couple of notable adverse consequences. This output doesn't explicitly communicate to users:

1. whether a given package is a main module or not, and
2. for any given package that's not a main module, which of the listed Go packages represents the main module that depends on the given package

### Proposed solution

We can solve for both of the above concerns by adding a new field to the [Go binary metadata struct](https://github.com/anchore/syft/blob/main/syft/pkg/golang_bin_metadata.go#L3-L9), called something like `mainModule`, for which we'd specify the main module detected when scanning the current Go binary, for all packages, whether or not a given package is the main module.

With this in place, consumers of Syft's output can determine if a given package is the binary's main module (because the package `name` vlaue would match the `mainModule` value), and for non-main modules, which package in Syft's output is the corresponding main module (because the `mainModule` value would match another package's `name` value).

jonasagx:
I can look into that.

PR #1026, opened 2022-06-03
add main module field to go bin metadata

jonasagx:
This PR adds `mainModule` to Go binary metadata implementing [Dan's suggestion](https://github.com/anchore/syft/issues/908): all golang packages will get a `mainModule`, including the main module itself (where `name` == `metadata.mainModule`).

JSON output schema was updated to `3.2.4`. Minor version bump because of new field added to golang's metadata.

Closes: https://github.com/anchore/syft/issues/908


luhring (commented):
This looks pretty good, but also this is a schema change for our output, so we need to update that.

luhring (approved):
Looks good!

PR #1033, opened 2022-06-06
Add announcement for Anchore OSS Virtual Meetup

briankoe741:
We would like to leave the announcement up until after the meetup on 6/15.


spiffcs:
@briankoe741 can you just sign the DCO here and make sure your commit is signed off
https://probot.github.io/apps/dco/

You can update your commit with the -s flag and it should take

spiffcs (approved):

PR #843, opened 2022-02-23
Option to enable specific language or ecosystem cataloger

ramanan-ravi:
Fixes #840 

If user knows the languages / ecosystem in the image or file system being scanned, user can run syft to generate sbom only for those specific languages / ecosystem.

Example command to enable a language or ecosystem cataloger
```
syft packages alpine:latest --enable-cataloger apkdb-cataloger
```

PR #1038, opened 2022-06-09
Add catalogers configuration

### LINKED ISSUE #465: Add ability to enable/disable package catalogers
There are two sets of package catalogers today: image and directory (see https://github.com/anchore/syft/blob/v0.19.1/syft/pkg/cataloger/cataloger.go#L33-L61). We should be able to control which package catalogers are enabled via application configuration. In this way a user can customize what catalogers are allowed for image scans vs directory scans.

This could open the door to having per-cataloger configuration as well (but doesn't need to be done for this issue).

### LINKED ISSUE #840: Enable/disable SBOM generation for specific language types

The syft command line tool generates SBOM's for all types of languages - JS, Ruby, Python, Php, Golang, Java, Rust & base operating system packages. 

However, it would be nice to provide some command line options to syft, so that we can enable/disable the SBOM generation for specific language types

**Why is this needed**:

We at Deepfence (www.deepfence.io) are big fans of syft & gripe, and we would like to integrate them into our open source offering, ThreatMapper (https://github.com/deepfence/ThreatMapper).

We get requests from community that they know the language types of their images, and would like to run vulnerability scans for those language types only.

If we provide such an option, it would improve the speed of the vulnerability scans.

**Additional context**:

We are glad to provide a pull request for this feature.

wagoodman:
This pulls in @ramanan-ravi s work from https://github.com/anchore/syft/pull/843 with a few adjustments (see https://github.com/anchore/syft/pull/843#issuecomment-1151214352 as for why the separate PR).

Closes #840
Closes #465


spiffcs:
@wagoodman I think there are some conflicts on this one. After those are resolved I can give it a review

README.md
```diff
@@ -360,6 +360,28 @@ exclude: []
 # same as --platform; SYFT_PLATFORM env var
 platform: ""
 
+# set the list of package catalogers to use when generating the SBOM
+# default = empty (cataloger set determined automatically by the source type [image or file/directory])
+# catalogers:
+#   - "ruby-gemfile"
```
spiffcs, line 366: nit: do these have to be in quotes or can we just remove that part?

spiffcs (approved):
Tentative approval from over here! If we fix the conflicts and I can give it a run on my local I think 👍

README.md
wagoodman: I don't think so, I'll update 👍

PR #1043, opened 2022-06-13
feat: add docker-compose workflow

cpendery:
## 📝 Description
Closes https://github.com/anchore/syft/issues/1042. Only difference between this and the grype version https://github.com/anchore/grype/pull/783 is that Syft requires install `git` and `zip`. 

Syft uses symlinks in its tests for `syft/syft/source`, but those fail because Docker resolves the links due to the mounted volume in compose [ref here](https://github.com/moby/moby/issues/13382). Could use some guidance/advice or if we just want to add this to the README as a point of failure

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -2,11 +2,15 @@
 
 ## Getting started
 
+### Native Development
+
 In order to test and develop in this repo you will need the following dependencies installed:
+
 - docker
 - make
 
 After cloning do the following:
+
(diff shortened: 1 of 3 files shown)


cpendery:
I'm going to close this pr and issue. I don't see a fix for the symlink issue or the fact that certain files exact size is checked, but on Windows this will be off by a few bytes due to the change in line endings. If I have a better idea for how to support this, I'll make a new pr

Issue #1044, opened 2022-06-14
Add support for Mariner distroless images


tofay:
**What would you like to be added**:
syft can identify RPM packages in [Mariner](https://github.com/microsoft/CBL-Mariner/)'s distroless container image

**Why is this needed**:
Enable scanning and subsequent CVE checking of Mariner distroless images

**Additional context**:
Mariner produces distroless images similar to https://github.com/GoogleContainerTools/distroless but containing Mariner packages.
These images don't contain a full RPM database, but they do contain a manifest file detailing the RPMs in the image. https://github.com/microsoft/CBL-Mariner/blob/main/toolkit/docs/how_it_works/5_misc.md#rpm-manifest-file

Syft could read this to determine the contained RPM. syft wouldn't be able to extract as much information as it could from an RPM db, e.g per-file data is not present, but it could extract enough information to enable subsequent CVE checking.

PR #1045, opened 2022-06-14
Add support for CBL-Mariner distroless images

tofay:
Fixes https://github.com/anchore/syft/issues/1044

Mariner distroless images don't contain an RPM database, but contain a file containing the
output of an `rpm --query --all ...` command: https://github.com/microsoft/CBL-Mariner/blob/main/toolkit/docs/how_it_works/5_misc.md#rpm-manifest-file

This PR adds support to syft for detecting that file and extracting package information from it. This is not as much information as syft extracts from a normal RPM database (a couple of examples being file information and signing algorithm) but it is enough information for subsequent CVE analysis.

### DIFF
--- a/syft/pkg/cataloger/rpmdb/parse_rpmdb.go
+++ b/syft/pkg/cataloger/rpmdb/parse_rpmdb.go
@@ -14,7 +14,7 @@ import (
 	rpmdb "github.com/knqyf263/go-rpmdb/pkg"
 )
 
-// parseApkDb parses an "Packages" RPM DB and returns the Packages listed within it.
+// parseRpmDb parses an "Packages" RPM DB and returns the Packages listed within it.
 func parseRpmDB(resolver source.FilePathResolver, dbLocation source.Location, reader io.Reader) ([]pkg.Package, error) {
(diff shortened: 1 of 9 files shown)


syft/pkg/cataloger/rpmdb/parse_rpmdb.go
```diff
@@ -14,7 +14,7 @@ import (
 	rpmdb "github.com/knqyf263/go-rpmdb/pkg"
 )
 
-// parseApkDb parses an "Packages" RPM DB and returns the Packages listed within it.
+// parseRpmDb parses an "Packages" RPM DB and returns the Packages listed within it.
```
wagoodman, line 17: 😎

wagoodman (approved):
TIL about RPM manifests; thanks for the contribution! (I added a comment for a small update, but that looks to be the only blocker from me)

tofay:
Thanks for the quick review! I've made the suggested change and one more thing I saw in passing (making parsing resilient to fields being added in future).

spiffcs (approved):

Issue #1049, opened 2022-06-15
add flag for image scanning to use all catalogers rather than just some


cpendery:
**What would you like to be added**:
I'd like to have a flag added where an image can be scanned by Syft using all catalogers rather than just a subset. 

**Why is this needed**:
When I create images, I scan them for any vulnerabilities before the build process. I'd like to be able to detect if any files exist that could potentially bring in vulnerable artifacts during the build process.

**Additional context**:
This is the enhancement required based on https://github.com/anchore/syft/issues/1046 as I had assumed this was default behavior

LOLCATATONIA:
The new 0.50 version should allow --use-all-catalogers, but it doesn't seem to work on my end?

```
$ syft version
Application:        syft
Version:            0.50.0
JsonSchemaVersion:  3.3.0
BuildDate:          2022-07-06T13:43:38Z
GitCommit:          69134ed3b54bc8b1d86d868611f7d069ce3290a8
GitDescription:     v0.50.0
Platform:           linux/amd64
GoVersion:          go1.18.3
Compiler:           gc
```

```
$ syft --use-all-catalogers
2022/07/07 08:31:33 error during command execution: unknown flag: --use-all-catalogers
```

PR #1050, opened 2022-06-15
feat: add use-all-catalogers flag

cpendery:
## Description
Add a new switch `--use-all-catalogers` that allows for the use of both image and file catalogers to run on the subject. This is useful when scanning an image built to use as a build environment, where artifacts may be installed using the files present soon.

Closes https://github.com/anchore/syft/issues/1049


cpendery:
Unnecessary if https://github.com/anchore/syft/pull/1038 gets merged

spiffcs (commented):
Added small comment about use of new flag vs extending a newly merged flag

cpendery:
> Added small comment about use of new flag vs extending a newly merged flag

I think extending the existing flag makes a lot of sense. I'll switch over to using that flag and part of the config :+1:

spiffcs (approved):

Issue #152, opened 2020-08-18
Allow user-defined output formats


luhring:
Some CLI tools, including `docker`, provide a command line option for the user to specify exactly in what form they would like to receive the tool's output.

<img width="793" alt="Screen Shot 2020-08-18 at 3 30 33 PM" src="https://user-images.githubusercontent.com/5199289/90557039-e189b180-e167-11ea-82a8-62c1c0b32218.png">

This is incredibly useful, as it allows users to hand pick which fields they are interested in, and it allows users to consume data using formats that the developers haven't yet considered or implemented.

For example, users could get CSV data out of the tool without us having to create an explicit output option for CSV:

```
$ syft myimage --format '{{.Name}},{{.Version}},{{.Type}}`
adduser,3.118ubuntu2,deb
alsa-topology-conf,1.2.2-1,deb
alsa-ucm-conf,1.2.2-1ubuntu0.1,deb
apt,2.0.2ubuntu0.1,deb
...
```

Go makes this really easy with its built-in [template](https://golang.org/pkg/text/template/#pkg-overview) functionality. Given the user-specified template string and the struct from which the data comes, it's trivial to loop through structs in a collection and output text in the format the user wants.

**Note:** This output feature can apply equally to other tools besides syft, but this is just a place to start.

This issue is related to https://github.com/anchore/grype/issues/42, but takes a different, more generalized approach. Both approaches might be desired, and they wouldn't inherently conflict with each other.

nicosto:
Hi,
This will be really useful to have this to report to application that doesn't support cyclondx/spdx.

luhring:
I'd like us to move forward on this now. But one update from the description — let's make sure to implement this functionality so that **it's exactly the same experience for users as Grype's templating functionality**, which is documented here: https://github.com/anchore/grype#using-templates

jonasagx:
Nice, I can take a look at it.

PR #1051, opened 2022-06-16
add template output

jonasagx:
Allow syft to output an SBOM based on a given [Go template](https://pkg.go.dev/text/template) like Grype currently does. I added a new output format called `template` that receives as param the template file path. I extended the `sbom.Format` to accept format options that formats can use if they implement it the interface themselves. 

Closes: #152 


cmd/syft/cli/options/writer.go
```diff
@@ -55,6 +56,7 @@ func parseOutputs(outputs []string, defaultFile string) (out []sbom.WriterOption
 			errs = multierror.Append(errs, fmt.Errorf("bad output format: '%s'", name))
 			continue
 		}
+		format = format.WithOptions(options.Format{TemplateFilePath: templateFilePath})
```
wagoodman, line 59: another way to achieve this without changing the Format interface is to type assert to `template.Format`, and if the assertion is ok, then add on the additional information or create a new instance of the format with the new information injected via a constructor.

syft/format-options/options.go
```diff
@@ -0,0 +1,6 @@
+package options
+
+// Format has options to be used by SBOM formats
+type Format struct {
```
wagoodman, line 4: I see why this is here, however, it is a little awkward (it's only purpose is for the state of template and isn't clear how to use this for more formats in the future). Is there a better way to do this?

wagoodman (commented):
I left a couple of comments around the` format-options` and downstream effects. The ultimate problem with the code (on `main`) is that there isn't a path to inject dependencies, which is the thing this PR is contending with.

I think there is a compromise that can be made without needing to change the `Format` interface, which is a pattern that affects all formats. That compromise can be made around https://github.com/anchore/syft/pull/1051/files#r900169470 . In the future we can pay down the improvement of injecting all formats from the application config (which solves this whole problem a lot better, but is out of scope for this PR).

cmd/syft/cli/options/writer.go
jonasagx: yes, I like that idea. I used and that helped drop `options.Format`.

syft/sbom/format.go
```diff
@@ -23,6 +25,7 @@ type Format interface {
 	Encode(io.Writer, SBOM) error
 	Decode(io.Reader) (*SBOM, error)
 	Validate(io.Reader) error
+	WithOptions(options.Format) Format
```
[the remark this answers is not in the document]
jonasagx: I see, I dropped it in favor of type casting to template format.

syft/format-options/options.go
jonasagx: Right, I removed it. Now the template output receives the param directly.

README.md
```diff
@@ -230,6 +230,40 @@ Where the `formats` available are:
 - `spdx-json`: A JSON report conforming to the [SPDX 2.2 JSON Schema](https://github.com/spdx/spdx-spec/blob/v2.2/schemas/spdx-schema.json).
 - `github`: A JSON report conforming to GitHub's dependency snapshot format.
 - `table`: A columnar summary (default).
+- `template`: Lets the user specify the output format. See "Using templates" below.
```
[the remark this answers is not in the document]
jonasagx: I added the link 👍🏼

internal/formats/template/format.go
```diff
@@ -0,0 +1,46 @@
...
+	doc := syftjson.ToFormatModel(s)
+	return tmpl.Execute(output, doc)
+}
+
+func (f OutputFormat) Validate(reader io.Reader) error {
+	return sbom.ErrValidationNotSupported
+}
+
+func (f OutputFormat) WithTemplate(filePath string) sbom.Format {
+	return OutputFormat{templateFilePath: filePath}
```
wagoodman, line 45: it's a little odd that this is worded like a functional option, but creates a new object like a builder pattern.

I think this fits a little better as a setter for a field on the instance it was called on, something like
```
f.SetTemplatePath(path)
``` 
(side steps the need to create a new instance of the format).

jonasagx: I see now. I've changed the func signature to that.

cmd/syft/cli/options/writer.go
```diff
@@ -55,6 +56,11 @@ func parseOutputs(outputs []string, defaultFile string) (out []sbom.WriterOption
 			errs = multierror.Append(errs, fmt.Errorf("bad output format: '%s'", name))
 			continue
 		}
+		tmpl, ok := format.(template.OutputFormat)
+		if ok {
```
luhring, line 60: nit: This is often done on the same line in Go

luhring (approved):
Looks good!

Issue #676, opened 2021-12-13
Catalog packages from source pom.xml during directory scans


luhring:
**What would you like to be added**:

During directory scans (e.g. w/ the `dir:` scheme), let's scan `pom.xml` files on disk as a means of learning about dependencies, and let's surface those dependencies as packages. Today Syft does parse `pom.xml` files, but only within the context of `.jar`/`.war`/etc. files.

**Why is this needed**:

This allows for source code detection of dependencies, even when no dependencies have been installed yet.

**Additional context**:

For an example repo that has dependencies listed in `pom.xml`, check out: https://github.com/nedenwalker/spring-boot-app-with-log4j-vuln

**Note:** Once this issue is implemented, we should pull this into Grype to improve Grype's scans of local code repositories.

luhring:
Related to #690 (the Gradle equivalent of this issue)

jonasagx:
I can look into that.

nedenwalker:
Sweet! Is there a particular release we should expect this functionality to be added in?

spiffcs:
@nedenwalker this should be in the current release of syft! Let me know if you find any problems with the implementation. Happy to accept PR for bug fixes or general feedback.

PR #1055, opened 2022-06-21
add pom.xml cataloger

jonasagx:
Add a parser to extract dependencies from pom.xml. 

Dependencies have dependencies themselves that might not be listed in the original pom and will only appear after you run `mvn install`. These deps of deps won't be included in the final sbom.

Fix: #676


syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -6,34 +6,79 @@ import (
...
+		URL:         p.URL,
+	}
+}
+
+func newPackageFromPom(dep gopom.Dependency) *pkg.Package {
+	p := &pkg.Package{
+		Name:         dep.ArtifactID,
+		Version:      dep.Version,
+		Language:     pkg.Java,
+		Type:         pkg.JavaPkg, // TODO: should we differentiate between packages from jar/war/zip versus packages from a pom.xml that were not installed yet?
```
jonasagx, line 63: ?

spiffcs:
Nice Addition @jonasagx! Noticed that there were no new files or updates under the `integration` folders for this feature. Should we add a `pom.xml` or some complex pom project to our integration images to make sure we're able to do this detection outside of the context of `.jar/.war/etc.`

syft/pkg/cataloger/java/parse_pom_xml.go
spiffcs: I don't have a good answer here and think it can go both ways cc @tool

jonasagx:
yes @spiffcs good call, it will probably be another test because the current ones rely on  `.jar/.war/etc.`

syft/pkg/cataloger/java/parse_pom_xml.go
wagoodman: the package type isn't meant to portray the installation status of the package (for instances we have two kinds of python catalogers, one that looks for installed packages and the other for declared packages... but they always return the same python package type)

wagoodman (approved):

PR #1063, opened 2022-06-24
Add support for dependency relationships for alpine (apk)

luhring:


spiffcs:
@tgerla I took this out of draft - if you push your changes I can review

tgerla:
Hi @spiffcs (and anyone else interested) -- I think this one is ready to go. Can you please review my last commit, "adjust test conditions"? I *think* I've updated the test correctly, but I am Go noob.

syft/pkg/apk_metadata.go
```diff
@@ -32,7 +32,8 @@ type ApkMetadata struct {
 	Description      string          `mapstructure:"T" json:"description"`
 	Size             int             `mapstructure:"S" json:"size" cyclonedx:"size"`
 	InstalledSize    int             `mapstructure:"I" json:"installedSize" cyclonedx:"installedSize"`
-	PullDependencies string          `mapstructure:"D" json:"pullDependencies" cyclonedx:"pullDependencies"`
+	PullDependencies []string        `mapstructure:"D" json:"pullDependencies" cyclonedx:"pullDependencies"`
```
kzantow, line 35: The nit here is that this ends up being a schema change, so I think we'll need to publish a new schema.

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -161,3 +181,42 @@ func parseApkDBEntry(reader io.Reader) (*pkg.ApkMetadata, error) {
...
+	lookup := make(map[string][]*pkg.Package)
+	// read "Provides" (p) and add as keys for lookup keys
+	for _, p := range pkgs {
+		apkg, ok := p.Metadata.(pkg.ApkMetadata)
+		if !ok {
+			// TODO warn
+			continue
+		}
+		for _, provides := range apkg.Provides {
+			// TODO: parse provides to remove =*
```
kzantow, line 196: It looks like there are still some TODOs here

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -161,3 +181,42 @@ func parseApkDBEntry(reader io.Reader) (*pkg.ApkMetadata, error) {
...
+		}
+
+		for _, dep := range apkg.PullDependencies {
+			// use the lookup to find what pkg we depend on
+			for _, depPkg := range lookup[dep] {
+				// this is a pkg that package "p" depends on... make a relationship
+				relationships = append(relationships, artifact.Relationship{
+					From: p,
+					To:   depPkg,
+					Type: "", // TODO... I think "RUNTIME_DEPENDENCY_OF" ... but really HAS_RUNTIME_DEPENDENCY
```
kzantow, line 216: Another TODO here might be important

kzantow (commented):
It looks like there are a couple `// TODO`s that might be important to fill out on this one -- what do you think?

syft/pkg/cataloger/apkdb/parse_apk_db.go
spiffcs: cc @wagoodman

syft/pkg/cataloger/apkdb/parse_apk_db.go
spiffcs: cc @wagoodman for context from our gardening meeting

wagoodman: while building this lookup, each string isn't necessarily a package reference. It could be a package name at a version: https://github.com/anchore/syft/blob/ed5c9c9b250615608da27a553e2ba404311b3d7c/syft/pkg/cataloger/apkdb/test-fixtures/extra-file-attributes#L14
```
so:openjdk8:libfontmanager.so=0
```
we need to extract out the names and ignore the versions:
```
so:openjdk8:libfontmanager.so
```

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -162,3 +182,42 @@ func parseApkDBEntry(reader io.Reader) (*pkg.ApkMetadata, error) {
...
 }
+
+func discoverPackageDependencies(pkgs []*pkg.Package) (relationships []artifact.Relationship) {
+	// map["provides" string] -> packages that provide the "p" key
+	lookup := make(map[string][]*pkg.Package)
+	// read "Provides" (p) and add as keys for lookup keys
+	for _, p := range pkgs {
+		apkg, ok := p.Metadata.(pkg.ApkMetadata)
+		if !ok {
+			// TODO warn
```
wagoodman, line 193: we don't want to stop processing here, but we should log that something unexpected happened here

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -162,3 +182,42 @@ func parseApkDBEntry(reader io.Reader) (*pkg.ApkMetadata, error) {
...
+			// TODO: parse provides to remove =*
+			lookup[provides] = append(lookup[provides], p)
+		}
+	}
+
+	// read "Pull Dependencies" (D) and match with keys
+	for _, p := range pkgs {
+		apkg, ok := p.Metadata.(pkg.ApkMetadata)
+		if !ok {
+			// TODO warn
```
wagoodman, line 206: ditto https://github.com/anchore/syft/pull/1063/files#r964110343

syft/pkg/cataloger/apkdb/parse_apk_db.go
wagoodman: we didn't land on a relationship type yet here, though it's most likely https://github.com/anchore/syft/blob/ed5c9c9b250615608da27a553e2ba404311b3d7c/syft/artifact/relationship.go#L13-L14 . The comment is referring to the direction of the edge... the relationship should read like a sentence:
```
<FROM> <TYPE> <TO>
```
so, for example:
```
[python package "requests"] is a [RUNTIME DEPENDENCY OF] [python package "anchore engine"]
```
or
```
[python package "anchore engine"] [HAS RUNTIME DEPENDENCY] [python package "requests"]
```
We should pick the direction that makes the most sense and try to stay consistent across all ecosystems.

syft/pkg/cataloger/apkdb/parse_apk_db_test.go
```diff
@@ -2,12 +2,14 @@ package apkdb
...
+
+	"os"
+	"testing"
+
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/source"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
-	"os"
-	"testing"
```
tgerla, line 13: Sorry, VS Code did this and I don't know why yet.

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -190,20 +190,20 @@ func discoverPackageDependencies(pkgs []*pkg.Package) (relationships []artifact.
...
 		apkg, ok := p.Metadata.(pkg.ApkMetadata)
 		if !ok {
-			// TODO warn
+			log.Warnf("cataloger failed to extract apk 'provides' metadata for package %+v", p.Name)
 			continue
 		}
 		for _, provides := range apkg.Provides {
-			// TODO: parse provides to remove =*
-			lookup[provides] = append(lookup[provides], p)
+			k := strings.Split(provides, "=")[0]
```
tgerla, line 197: This is how I'd do it in Python, not sure if there is a more idiomatic way in Go.

spiffcs (commented):
Added a question about schema change procedure as well as separator being used. When we settle on the right separator we might want to add a quick test to be sure the split is behaving correctly. 

Also small nit on the strings usage
```
splits := strings.Split(provides, "=")
if strings.Contains(splits[0], "=") {
    // protect against unsuccesful split
}
```

syft/pkg/apk_metadata.go
wagoodman: yup, that's correct, this is a breaking change to the schema -- to V4 we go! There are instructions of how to create a new schema and bump it in `schema/json/README.md`. I think we should also add a test that asserts we can decode examples of APK metadata from a syft JSON with the format decoders without issue (decoding V4 works in the case with APK metadata).

syft/pkg/cataloger/apkdb/parse_apk_db.go
wagoodman: the `=` is correct (as it is in the code now). Take for example: 

```
P:ssl_client
V:1.35.0-r17
D:so:libc.musl-x86_64.so.1 so:libcrypto.so.1.1 so:libssl.so.1.1.       # provides libcrypto
```
and
```
P:libcrypto1.1
V:1.1.1q-r0
D:so:libc.musl-x86_64.so.1
p:so:libcrypto.so.1.1=1.1           # depends on libcrypto v1.1 specifically
```

Where `ssl_client` depends on `libcrypto`. The libcrypto package specifies the exact version provided and the consumer specifies the loose requirements that would suffice (in this case there is no `=1.1`). A better example is one where the provided specifies a more specific version, and the consumer locks to the major version (and not the minor version).

But, let's take a look at another example:
```
C:Q1VFFFWMKjB9aRkehIATc5kwgAhlU=
P:apk-tools
V:2.12.9-r3
A:x86_64
D:musl>=1.2 ca-certificates-bundle so:libc.musl-x86_64.so.1 so:libcrypto.so.1.1 so:libssl.so.1.1 so:libz.so.1
p:so:libapk.so.3.12.0=3.12.0 cmd:apk=2.12.9-r3
```
where `musl` is referring to a package name (`P:` field) and not a provider field (`p:` field):
```
C:Q1aCu0LmUDoAFSOX49uHvkYC1WasQ=
P:musl
V:1.2.3-r0
# note, there is no D: or p: field on this entry
```

We want the "dependencies" and "provides" to match up, like with the previous example, but this shows that we need "dependencies" and the "package" field to also be sometimes related.

This hints at more things needing to be captured by this parser.

wagoodman:
@tgerla it looks like there may be more things that were not originally in scope that need to be considered as part of this PR. It looks like the https://wiki.alpinelinux.org/wiki/Apk_spec page got a really healthy update over the last few months to help interpret an APK index vs installed DB file, which will be really helpful for determining the missing bits. I'd be happy to pair on this.

kzantow (approved):
Just a few non-blocking questions...

syft/pkg/apk_metadata.go
```diff
@@ -23,20 +23,21 @@ var (
...
+	OriginPackage string          `mapstructure:"o" json:"originPackage" cyclonedx:"originPackage"`
+	Maintainer    string          `mapstructure:"m" json:"maintainer"`
+	Version       string          `mapstructure:"V" json:"version"`
+	License       string          `mapstructure:"L" json:"license"`
+	Architecture  string          `mapstructure:"A" json:"architecture"`
+	URL           string          `mapstructure:"U" json:"url"`
+	Description   string          `mapstructure:"T" json:"description"`
+	Size          int             `mapstructure:"S" json:"size" cyclonedx:"size"`
+	InstalledSize int             `mapstructure:"I" json:"installedSize" cyclonedx:"installedSize"`
+	Dependencies  []string        `mapstructure:"D" json:"pullDependencies" cyclonedx:"pullDependencies"`
```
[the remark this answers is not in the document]
wagoodman: I ended up leaving it since it would affect hoe to decode existing SBOMs. I updated the variable since the APK documentation has been updated a couple of months ago with much better details.

test/integration/encode_decode_cycle_test.go
```diff
@@ -36,24 +37,47 @@ func TestEncodeDecodeEncodeCycleComparison(t *testing.T) {
...
+				start := bytes.Index(in, []byte("  <dependencies>")) // important: mind the prefix whitespace
+				endVal := "</dependencies>\n"                        // important: mind the postfix whitespace
+				stop := bytes.Index(in, []byte(endVal))
+				if start != -1 && stop != -1 {
+					stopAfterVal := stop + len(endVal)
+					inCopy := make([]byte, 0)
+					inCopy = append(inCopy, in[:start]...)
+					inCopy = append(inCopy, in[stopAfterVal:]...)
+					in = inCopy
+				}
```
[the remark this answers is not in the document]
wagoodman: I tried that at first and it did not work as expected... but got it working with the final regex: `(?m:(\n\s+)*<dependencies>[\s\S]*?</dependencies>)`

test/integration/encode_decode_cycle_test.go
```diff
@@ -36,24 +37,47 @@ func TestEncodeDecodeEncodeCycleComparison(t *testing.T) {
...
+				// unstable values
+				in = regexp.MustCompile(`"(timestamp|serialNumber|bom-ref)": "[^"]+",`).ReplaceAll(in, []byte{})
+
+				// dependencies are not supported (edge types cannot be encoded or inferred during decoding)
+				var det map[string]interface{}
+				require.NoError(t, json.Unmarshal(in, &det))
+				delete(det, "dependencies")
+				inCopy, err := json.Marshal(det)
+				require.NoError(t, err)
+				in = inCopy
```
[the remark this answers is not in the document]
wagoodman: it seems that investigating this raised another larger issue with how cataloger relationship IDs are crafted (and the pointer from where they are sourced from is derived). I'll need to dig deeper and fix in another PR. I'll hold off on merging this PR until this fix is in.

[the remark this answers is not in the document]
kzantow: @wagoodman I noticed this, too, in working on the multiple sources cyclonedx stuff. I can make a PR to fix this if you'd like.

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -159,6 +179,61 @@ func parseApkDBEntry(reader io.Reader) (*pkg.ApkMetadata, error) {
...
+
+		for _, depSpecifier := range apkg.Dependencies {
+			// use the lookup to find what pkg we depend on
+			dep := stripVersionSpecifier(depSpecifier)
+			for _, depPkg := range lookup[dep] {
+				// this is a pkg that package "p" depends on... make a relationship
+				relationships = append(relationships, artifact.Relationship{
+					From: depPkg,
+					To:   p,
+					Type: "runtime-dependency-of", // depPkg IS A runtime dependency OF p
```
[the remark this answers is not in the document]
wagoodman: done 👍

test/integration/encode_decode_cycle_test.go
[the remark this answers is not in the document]
wagoodman: I've removed the need for this after our larger conversation about package IDs and relationships

luhring:
This one should not have been squashed 😄

Issue #1066, opened 2022-06-25
Syft generates too loose of cpes which will/have never occurred


cpendery:
**What happened**:
Syft generates cpes with the vendor and product being identical

**What you expected to happen**:
This shouldn't happen as that leads to false flagging and has never occurred in a cpe

**How to reproduce it (as minimally and precisely as possible)**:
See linked issue below for replication

**Anything else we need to know?**:
Related to https://github.com/anchore/grype/issues/800

**Environment**:
- Output of `syft version`: `38.1 (from Grype 40.0)`
- OS (e.g: `cat /etc/os-release` or similar):
```
System Version: macOS 11.6 (20G165)
Kernel Version: Darwin 20.6.0
Model Name: MacBook Pro
Model Identifier: MacBookPro16,1
Processor Name: 6-Core Intel Core i7
```

PR #1067, opened 2022-06-25
fix: dropping generation of excessively loose cpes

cpendery:
## 📝 Description
A cpe where vendor and product are identical has never occurred in any data source Grype or Syft uses and it's not in the 2.3 cpe database from NVD. It can lead to false matching, so this PR removes it from being generated

Closes https://github.com/anchore/grype/issues/800 and closes https://github.com/anchore/syft/issues/1066

### DIFF
--- a/syft/pkg/cataloger/common/cpe/generate.go
+++ b/syft/pkg/cataloger/common/cpe/generate.go
@@ -39,6 +39,10 @@ func Generate(p pkg.Package) []pkg.CPE {
 	cpes := make([]pkg.CPE, 0)
 	for _, product := range products {
 		for _, vendor := range vendors {
+			// avoid generating incorrect cpe...
+			if product == vendor {
+				continue
+			}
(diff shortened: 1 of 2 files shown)


westonsteimel:
Thanks Chapman!  I'm currently seeing 5,774 unique instances where product and vendor are identical for a CPE from the [NVD CPE match dateset](https://nvd.nist.gov/feeds/json/cpematch/1.0/nvdcpematch-1.0.json.gz) though

cpendery:
> Thanks Chapman! I'm currently seeing 5,774 unique instances where product and vendor are identical for a CPE from the [NVD CPE match dateset](https://nvd.nist.gov/feeds/json/cpematch/1.0/nvdcpematch-1.0.json.gz) though

Resolved offline, I didn't realize the cpe dictionary wasn't cumulative. PR isn't a fix for the issue

PR #1068, opened 2022-06-27
feat: add new login cmd

### LINKED ISSUE #1060: new login cmd

Syft requires to do "docker login" first for private registries. But what if I don't have a docker CLI installed on my environment? In this case, I should be able to log in with syft. So, we (w/@dentrax) are proposing to add a new login command for Syft, actually, it is so easy to add[^1]. Because crane has some global commands that we can easily use. 

https://github.com/google/ko/blob/8e4587e93e309f615482e97d3ec5a402dd263b8c/pkg/commands/root.go#L60

[^1]: https://github.com/sigstore/cosign/pull/1398

**Why is this needed**:

**Additional context**:

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #1060

### DIFF
--- a/cmd/syft/cli/commands.go
+++ b/cmd/syft/cli/commands.go
@@ -15,6 +15,7 @@ import (
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/internal/version"
 	"github.com/anchore/syft/syft/event"
+	cranecmd "github.com/google/go-containerregistry/cmd/crane/cmd"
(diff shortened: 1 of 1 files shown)


spiffcs:
Nice thanks for the quick fix here @developer-guy! I'll give this a test locally and clean up the static-analysis stuff

spiffcs (approved):

PR #1070, opened 2022-06-28
fix: add manual vendor/product removal to fix false flags

cpendery:
## 📝 Description
Not a great solution, but it at least removes these false flagging events and matches the current practices of manual vendor/product additions by adding support for manual vendor/product removals

Closes https://github.com/anchore/syft/issues/1066
Closes https://github.com/anchore/grype/issues/800
Closes https://github.com/anchore/grype/issues/491

### DIFF
--- a/syft/pkg/cataloger/common/cpe/candidate_by_package_type.go
+++ b/syft/pkg/cataloger/common/cpe/candidate_by_package_type.go
@@ -11,6 +11,12 @@ type candidateComposite struct {
 	candidateAddition
 }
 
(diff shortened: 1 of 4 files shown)


spiffcs:
@westonsteimel should we even introduce this into the app layer with `vulnerability_match_exclusion` now being a table where anchore can start managing these at the data layer?

cc @wagoodman also for his thoughts on corrections for cpe at the app layer vs data layer

spiffcs (approved):

Issue #1071, opened 2022-06-28
feat: add Erlang/Elixir (rebar, mix) support


cpendery:
**What would you like to be added**:
I'd like to have my Erlang/Elixir projects where I used Hex be able to produce sboms

**Why is this needed**:
Erlang and Elixir are programming languages that have CVEs reported and are well established, so we should be able to produce sboms for them.

**Additional context**:

cpendery:
happy to take on this contribution if its not being worked on internally

westonsteimel:
@cpendery , thanks, that would be fantastic!

PR #1073, opened 2022-06-29
feat: add Elixir mix support [WIP]

cpendery:
## 📝 Description
Adds support for parse `mix.exs` and `mix.lock` files to add cataloguing support for elixir projects. Placed under the `beam` cataloger as Erlang projects can use both Mix and Rebar. I'll add support for Rebar to the cataloger in a following PR

Completes task 1: https://github.com/anchore/syft/issues/1071


westonsteimel:
Thanks for another great contribution @cpendery!  I do think having `beam` as the language is going to be confusing.  My suggestion would be to break it up into a separate `erlang` cataloger for `rebar.lock` and `elixir` cataloger for `mix.lock`, and both having the `hex` package type.  I'm not sure if sharing the same package type between two languages will cause any other problems though.

cpendery:
> Thanks for another great contribution @cpendery! I do think having `beam` as the language is going to be confusing. My suggestion would be to break it up into a separate `erlang` cataloger for `rebar.lock` and `elixir` cataloger for `mix.lock`, and both having the `hex` package type. I'm not sure if sharing the same package type between two languages will cause any other problems though.

With splitting the languages into Elixir and Erlang, when we see a purl for the Hex `pkg:hex/` package manager (supports both languages), I wouldn't know what language to resolve it to. I get using the ecosystem as a language is confusing. Any ideas @westonsteimel

westonsteimel:
> With splitting the languages into Elixir and Erlang, when we see a purl for the Hex pkg:hex/ package manager (supports both languages), I wouldn't know what language to resolve it to. I get using the ecosystem as a language is confusing. Any ideas @westonsteimel

Hmm, can we encode the language as a parameter to the package url when syft creates it?  @spiffcs , any thoughts?

seabass-labrax:
> Hmm, can we encode the language as a parameter to the package url when syft creates it? @spiffcs, any thoughts? Do we ever need to decode language from a package url (perhaps this is something that happens from one of the non syft-native formats)?

The purl specification doesn't have a `language` property for the [Hex purl type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#hex), unfortunately. However, it seems like an advantageous thing to add, so maybe @pombredanne would be up for adding this! :smile:

cpendery:
It could be useful for Conan (C/C++) and Cocoapods (Objective-C/Swift) since both support dual languages like Hex

spiffcs:
@cpendery this looks really good. I'll wait on merging or updating in anyway until we hear back on the PR you made for the purl-spec.

spiffcs:
TODO: update cataloger to new generic cataloger pattern

wagoodman:
I think we can leave language as blank / unknown in these circumstances -- the cataloger is more valuable than resolving the language from the pURL IMHO.

I can help rebase what is here and update the patterns some based on the drift.

wagoodman:
The main changes I made were:
- Split the cataloger into `erlang` and `elixir`. The conflict for parsing the language from the pURL has been kicked down the road.
- Split the `HexMetadata` into `MixLockMetadata` and `RebarLockMetadata`. Why do this since they contain essentially the same information? If we intend to support extracting more information from these sources, they aren't exactly the same, so we want future room to grow here without having to make a breaking change. The package type may map to many different metadatas, and the metadatas should most closely represent the source in which you are parsing from --this is the general rule of thumb.
- Updated the cataloger to use the new generic cataloger and surrounding patterns.

I'll push shortly, and I think this will be good to go!

wagoodman (approved):
Thanks for your work and patience on this!

Issue #1077, opened 2022-06-29
unintended artifactRelationship records of type ownership-by-file-overlap are being reported in SBOMs generated against current fedora container imges


nurmi:
**What happened**:

Noticed that when generating and SBOM against latest fedora containers (e.g. docker.io/fedora:36, docker.io/fedora:37), there are many 'artifactRelationships' records of type 'ownership-by-file-overlap' specifically for '/usr/lib/sysimage/rpm/rpmdb.sqlite'.  Essentially, it appears that every RPM has such a relationship record where the parent is the 'rpm' package itself.  This is a change from previous fedora containers where those ownership-by-file-overlap records do not exist in the output of syft.

**What you expected to happen**:

In:

https://github.com/anchore/syft/blob/1e3ffbebb9de2dccadd76b2d283c2f64fb3e40ea/syft/pkg/relationships_by_file_ownership.go#L10

there is an explicit filter to ensure that the rpmdb.sqlite (matched by the global RpmDBGlob regex) is filtered out when determining whether there is a ownership-by-file-overlap type relationship, and the expectation is that the rpmdb.sqlite location in current fedora images would be filtered like is against older fedora container images.

**How to reproduce it (as minimally and precisely as possible)**:

since fedora:36 and fedora:35 don't have substantially different composition, one can see the difference by inspecting the number and nature of the ownership-by-file-overlap records between the syft output of the two images, but a shortcut way is

```
# syft -o json fedora:36 | grep ownership-by-file-overlap | wc
    149     298    5960

# syft -o json fedora:35 | grep ownership-by-file-overlap|wc
      3       6     120
```

**Anything else we need to know?**:

I believe the fundamental reason for this new behavior is that fedora:36 (and 37, perhaps others moving forward) have introduced a change where /var/lib/rpm is a softlink to /usr/lib/sysimage/rpm, whereas in prior fedora builds /var/lib/rpm was a real directory. 

```
# docker run -t -i fedora:35 ls -ald /var/lib/rpm
drwxr-xr-x 2 root root 4096 Feb  2 12:40 /var/lib/rpm

# docker run -t -i fedora:36 ls -ald /var/lib/rpm
lrwxrwxrwx 1 root root 26 May  6 10:10 /var/lib/rpm -> ../../usr/lib/sysimage/rpm

# docker run -t -i fedora:36 ls -ald /usr/lib/sysimage/rpm
[...]

PR #1079, opened 2022-06-30
fix: add glob to filter list to ensure rpm metadata files are matched…

nurmi:
… when var/lib/rpm is a softlinked (#1077)

Signed-off-by: Daniel Nurmi <nurmi@anchore.com>


spiffcs:
@nurmi retriggered so that the failing check was removed

PR #1106, opened 2022-07-19
move from scratch to distroless base

### LINKED ISSUE #833: Replace scratch base image with distroless static
I'd like to change the base image to distroless static from scratch. And also publish a debug version.

**Why is this needed**:
For GitLab CI the presence of shell is needed for Kubernetes based runners. If the project would publish such images, there could be easily a variant for `debug` image, that would contain shell. It is the approach that other project followed that runs on top of shell-less images, e.g. kaniko.

Additionally it'd save manually copying CA certificates during image build, as currently present in Dockerfile. The image size difference would be negligible with distroless static base image.

**Additional context**:
If you agree with the approach, I'm happy to contribute this.

spiffcs:
Move base image from scratch to distroless as discussed during the community meeting.

Closes #833 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM alpine:latest AS build
 
 RUN apk --no-cache add ca-certificates
 
-FROM scratch
+FROM gcr.io/distroless/base
 # needed for version check HTTPS request
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
@@ -27,6 +27,5 @@ LABEL org.opencontainers.image.licenses="Apache-2.0"
 LABEL io.artifacthub.package.readme-url="https://raw.githubusercontent.com/anchore/syft/main/README.md"
 LABEL io.artifacthub.package.logo-url="https://user-images.githubusercontent.com/5199289/136844524-1527b09f-c5cb-4aa9-be54-5aa92a6086c1.png"
 LABEL io.artifacthub.package.license="Apache-2.0"
-      
 
 ENTRYPOINT ["/syft"]


spiffcs:
After reading #833 more in-depth it looks like this still does not add a shell which was the requested change. I'll revert this commit and add this via the image_template block with a separate file in `.goreleaser.yaml`

06kellyjac:
Only the distroless debug images have a shell. Also for this reason I would strongly request there be a scratch & distroless debug image

https://github.com/GoogleContainerTools/distroless#debug-images

Or provide a simple dockerfile that can copy the binary from syft scratch container into distroless debug or alpine etc

---

Also as a note, distroless static comes with ca-certs, so doesnt need them copied in, so is probably the best option for a static go bin

https://github.com/GoogleContainerTools/distroless/tree/main/base#image-contents

spiffcs:
Thanks a million @06kellyjac! I'll update the goreleaser config so we have the scratch & distroless debug image.

debug.Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM gcr.io/distroless/static-debian11:debug
```
wagoodman, line 1: nit: mind if we rename this file to `Dockefile.debug` to keep the files organized next to one another?

spiffcs: flipped

.goreleaser.yaml
```diff
@@ -98,14 +98,22 @@ brews:
...
   - image_templates:
-      - "anchore/syft:latest"
-      - "anchore/syft:{{ .Tag }}-amd64"
-      - "anchore/syft:v{{ .Major }}-amd64"
-      - "anchore/syft:v{{ .Major }}.{{ .Minor }}-amd64"
-      - "ghcr.io/anchore/syft:latest"
-      - "ghcr.io/anchore/syft:{{ .Tag }}-amd64"
-      - "ghcr.io/anchore/syft:v{{ .Major }}-amd64"
-      - "ghcr.io/anchore/syft:v{{ .Major }}.{{ .Minor }}-amd64"
+      - anchore/syft:{{.Tag}}-debug
```
wagoodman, line 101: should we be providing a `anchore/syft:latest-debug` ? Or I guess it makes more sense as `anchore/syft:debug`

spiffcs: I can add `anchore/syft:debug` as the rolling latest version here

.goreleaser.yaml
```diff
@@ -98,14 +98,22 @@ brews:
...
+      - "--platform=linux/amd64"
+      - "--build-arg=BUILD_DATE={{.Date}}"
+      - "--build-arg=BUILD_VERSION={{.Version}}"
+      - "--build-arg=VCS_REF={{.FullCommit}}"
+      - "--build-arg=VCS_URL={{.GitURL}}"
+
+  - image_templates:
+      - anchore/syft:latest
+      - ghcr.io/anchore/syft:latest
+      - ghcr.io/anchore/syft:{{.Tag}}-amd64
```
wagoodman, line 116: is this `-amd64` intentional. It looks to be covered in the manifests section (but we should double check this)

spiffcs: ^ Manifests pull from the `dockers.image_templates` section

So if you did a `make build` on your local and then `docker manifest inspect anchore/syft:latest`

you would see 2 distributions. 

See: https://goreleaser.com/customization/docker_manifest/#example-config for the example setup for manifests and why we specify amd64 here.

wagoodman: That makes sense, thank for info / double checking 👍

wagoodman (approved):

Issue #1108, opened 2022-07-19
Add support for auditable Rust binaries


tofay:
**What would you like to be added**:
[rust-audit](https://github.com/Shnatsel/rust-audit) is a tool/cargo plugin for embedding dependency information in Rust binaries, for subsequent scanning/inspection (similar to go binaries).

I'd like syft to be able to detect Rust crate dependencies in Rust binaries built with rust

**Why is this needed**:
To account for statically linked Rust binaries in SBOMs/scanning.

**Additional context**:
I've been discussing rust-audit with the Rust secure code working group recently - the format is currently a simplified version of Cargo.lock but there's a desire to embed the information in a standardized way and we're looking into options.
Because of this if syft were to support this it should likely be marked as experimental, or off by default in some way.

PR #1116, opened 2022-07-23
add a cataloger for binaries built with rust-audit

tofay:
Fixes https://github.com/anchore/syft/issues/1108

https://github.com/Shnatsel/rust-audit is a tool for embedding Rust dependency information into binaries for subsequent scanning/detection.
This PR adds support for detecting dependencies in binaries created with rust-audit. Per https://github.com/anchore/syft/issues/1108 it's not enabled by default. (A motivation for syft integration is to get greater experience with rust-audit with a view to adding it to Rust itself).

The new cataloger is similar to the golang binary one - I modularized some of the golang binary cataloger code. Detection/parsing of the dependency information is done via https://github.com/microsoft/go-rustaudit, a go library for extracting Rust audit dependency information.


spiffcs:
Thank you so much for the PR @tofay! I'll take a look this morning and see if we can get this into the next syft release.

spiffcs:
PR looks good from a functionality perspective - 

I just had a question when running it against binaries in the wild.

 Is this the constraint that this cataloger will only work for static assets built with `https://github.com/Shnatsel/rust-audit` and with `auditable_build::collect_dependency_list();` in a build script?

The integration test you added passed, but when I rant his against other binaries I found no packages being detected.

If this is the case can we rename the file `binary_cataloger` ==> `audit_binary_cataloger` just for keep it clear. 

Thanks for updating the docs as well that points to this condition so we're not just advertising we can do static analysis for any binary!

I also noticed `Getting some real-world experience with this before committing to a stable data format` as a part of their roadmap for moving this into cargo.

Are there any competing projects or other contenders that we might have to flip to in the future?

spiffcs:
cc @wagoodman for a 2nd review here

spiffcs (commented):
See above comments - waiting for second review

tofay:
Thanks for looking at this. I appreciate it's not as straightforward as "add support for widely used language ecosystem X" type PRs :)

> Is this the constraint that this cataloger will only work for static assets built with https://github.com/Shnatsel/rust-audit and with auditable_build::collect_dependency_list(); in a build script?

Correct. Would it help set expectations for this cataloger if the cataloger name was prefixed with `experimental-`? I can change the file name as you suggest.

> I also noticed Getting some real-world experience with this before committing to a stable data format as a part of their roadmap for moving this into cargo. Are there any competing projects or other contenders that we might have to flip to in the future?

There aren't other competing projects that I'm aware of. In terms of data formats - I'm looking into language-agnostic options to embed dependency information into binaries. e.g https://github.com/ossf/wg-vulnerability-disclosures/issues/76 and https://github.com/spdx/spdx-spec/issues/739

spiffcs:
Cool! No need for the experimental name change unless @wagoodman has opinions there.

I'm happy with its current state since it's not in the default path and is something we can toggle/experiment with. Approved and thanks again so much for the contribution.

PR #1123, opened 2022-07-28
bump goreleaser to v1.10.3

westonsteimel:


spiffcs:
Looking at the CI here now to help get this merged

westonsteimel:
Wow, only one of the three broke though.  I was figuring all of them would

westonsteimel:
Also, likely need same bumps for grype.  I'm out for a bit but can submit those when I get back unless someone else gets to it before then

spiffcs:
> Also, likely need same bumps for grype. I'm out for a bit but can submit those when I get back unless someone else gets to it before then

I'll get this fixes over in grype after looking at this one.

wagoodman (approved):

test/cli/utils_test.go
```diff
@@ -225,7 +225,11 @@ func getSyftBinaryLocationByOS(t testing.TB, goOS string) string {
 	// note: there is a subtle - vs _ difference between these versions
 	switch goOS {
 	case "darwin", "linux":
-		return path.Join(repoRoot(t), fmt.Sprintf("snapshot/%s-build_%s_%s/syft", goOS, goOS, runtime.GOARCH))
+		archPath := runtime.GOARCH
```
wagoodman, line 228: nit: i feel it would be better to put the arch addition before the switch case in case we add more cases to the switch case in the future.

PR #1133, opened 2022-08-01
remove docker assets from main goreleaser configuration to reduce mac-os runner friction

spiffcs:

This PR removes syft's docker assets from the main `.gorelease.yaml` configuration.

Because of the removal of the docker dependency from our core release process we'll shave almost 20 minutes off of our current release time.

### Context:

Currently, the release process needs to run on the `mac-os` github runner. This constraint is necessary because the binary assets published for darwin need to be signed and notarized on an xcode-compatible system.

This presents challenges in the form of installing and wrangling the dependencies for docker builds to be possible on the `mac-os` runner. Syft previously used `https://github.com/docker-practice/actions-setup-docker` to meet this challenge. Recently the action has taken 15-21 minutes to get a working docker installation up and running. In some cases, it would fail outright and kill any potential release that had been triggered for publishing.

This change removes the need for the docker dependency on the `mac-os` runner, thus removing `actions-setup-docker` from our release workflow.

With this change a release will be processed as follows:
- Release commit is tagged which triggers the quality gate 
- If the quality gate passes the `make release` command is run on a `mac-os` runner (no longer using any docker dependencies)
- If the release is published successfully then the docker assets are triggered to be built and pushed to their respective registries

### Changes
This PR introduces a separate `.goreleaser_docker` configuration that is used to build and publish images after a GitHub release has been published. It's labeled as `release-docker-assets` and can be re-triggered without having to re-run the `quality gate` or `release` steps should it fail independently.

The Makefile has been updated to add `snapshot-docker-assets` and `release-docker-assets`. These are analogous to their `snapshot` and `release` counterparts for the main `.goreleaser.yaml` file.


Makefile
```diff
@@ -263,6 +263,18 @@ snapshot-with-signing: ## Build snapshot release binaries and packages (with dum
 	# remove the keychain with the trusted self-signed cert automatically
 	.github/scripts/apple-signing/cleanup.sh
 
+snapshot-docker-assets: # Build snapshot images of docker images that will be published on release
+	$(call title,Building snapshot docker release assets)
+
+	# create a config with the dist dir overridden
+	echo "dist: $(DISTDIR)" > $(TEMPDIR)/goreleaser.yaml
+	cat .github/.goreleaser_docker.yaml >> $(TEMPDIR)/goreleaser.yaml
```
kzantow, line 271: Why does there need to be a different directory than the default when this is running on its own runner?

kzantow (approved):

Makefile
spiffcs: I wanted to keep the makefile consistent. 

We're setting the path for a consumed `goreleaser.yaml` as `$(TEMPDIR)/gorelease.yaml` for the other make commands that invoke `goreleaser`. 

Reading it from `.github/.goreleaser_docker.yaml` would also be fine, just different from how we approach setting up the config in other commands.

.github/workflows/release.yaml
```diff
@@ -132,8 +130,6 @@ jobs:
         env:
           GHCR_USERNAME: ${{ github.actor }}
           GHCR_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
spiffcs: Removing - we don't upload assets anywhere but the release page which is not covered by these variables

.github/.goreleaser_docker.yaml
```diff
@@ -0,0 +1,190 @@
+# Separate docker configuration to isolate docker dependency away from
```
wagoodman, line 1: nit: I feel that it would be more apparent for all go-releaser things to be in the same spot. In this case the root of the repo.

.github/workflows/release.yaml
```diff
@@ -164,3 +157,19 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Build & Publish docker images
+        run: make release-docker-assets
+        env:
+          GHCR_USERNAME: ${{ github.actor }}
+          GHCR_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+          DOCKER_USERNAME: ${{ secrets.TOOLBOX_DOCKER_USER }}
```
wagoodman, line 174: though this step is given the credentials, this wouldn't work until we explicitly run `docker login`. We couldn't do this in a sane way with the signing step (we needed to replace the whole keychain, so the creds would be wiped), but now that there are no signing concerns and this is just a workflow concern (not a step that would ever need to be done locally / in the makefile) then we can use the docker provided action: https://github.com/docker/login-action

wagoodman: (same answer for the ghcr credentials)

.goreleaser.yaml
```diff
@@ -97,160 +97,3 @@ brews:
     homepage: *website
     description: *description
     license: "Apache License 2.0"
```
wagoodman, line 99: nit: we should probably remove the `DOCKER_CLI_EXPERIMENTAL` var since it's no longer used

.github/workflows/release.yaml
```diff
@@ -164,3 +157,19 @@ jobs:
         with:
           name: artifacts
           path: dist/**/*
+
+  release-docker-assets:
```
wagoodman, line 161: not a blocking comment: if this fails we'd need to republish the entire release. We could change this section to a post-release workflow:
```
# .github/workflows/publish-docker.yaml

name: "Release follow-up"
on:
  release:
    types: [published]
  workflow_dispatch:
    inputs:
      tag:
        description: 'The tag to base publication from'
        required: true 
        type: string

jobs:
 ...same...

```
In which case, if the release succeeds but the docker publish fails then we could modify and re-run the publishing in a workflow-dispatch (manual) fashion (thus, we wouldn't need to re-tag and re-release).

spiffcs: I think we're actually able to re-run this as a JOB on the actions page if it fails. So if `quality-gate` and `release` pass, but `release-docker-assets` fails we have to option to re-run all failed jobs. which in this case would just be the `release-docker-assets` job.

wagoodman: I think this makes sense, as long as the failure doesn't require a code update 👍

wagoodman (approved):
Soooo looking forward to getting 15 minutes back from the release process `['hip!', 'hip!']`

PR #1134, opened 2022-08-01
remove ENV checks from siging script

spiffcs:


wagoodman (approved):
ouch

PR #1136, opened 2022-08-02
improve docker release bootstrap

spiffcs:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -180,5 +180,18 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Restore go cache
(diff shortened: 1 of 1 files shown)


.github/workflows/issues.yaml
```diff
@@ -1,17 +0,0 @@
-# Uses https://github.com/actions/add-to-project example to add issues to Anchore OSS project
```
kzantow, line 1: Deleting this file seems unrelated

kzantow (approved):
Caveat about deleted `issues.yaml`

.github/workflows/issues.yaml
spiffcs: Yep! It's not doing anything right now so was cleaning up where I could while I was in the workflows file.

Issue #1141, opened 2022-08-04
Can't configure off-by-default cataloger without using --all


tofay:
**What happened**:
I'm trying to use the rust-audit-binary-cataloger, which is off by default, on a container image.
When I run
```
syft --catalogers rust-audit-binary <my-image>
```
the default set of image catalogers is used.
**What you expected to happen**:
The rust-audit-binary-cataloger to be a) used b) the only cataloger that is used.

**How to reproduce it (as minimally and precisely as possible)**:
Run
```
syft -v --catalogers rust-audit-binary ubuntu:20.04
[0000]  INFO syft version: 0.53.0
[0000]  INFO new version of syft is available: 0.53.4 (current version is 0.53.0)
[0001]  INFO identified distro: Ubuntu 20.04.4 LTS
[0001]  INFO cataloging image
[0001]  INFO skipping cataloger "alpmdb-cataloger"
[0001]  INFO skipping cataloger "ruby-gemspec-cataloger"
[0001]  INFO skipping cataloger "python-package-cataloger"
[0001]  INFO skipping cataloger "php-composer-installed-cataloger"
[0001]  INFO skipping cataloger "javascript-package-cataloger"
[0001]  INFO skipping cataloger "dpkgdb-cataloger"
[0001]  INFO skipping cataloger "rpmdb-cataloger"
[0001]  INFO skipping cataloger "java-cataloger"
[0001]  INFO skipping cataloger "apkdb-cataloger"
[0001]  INFO skipping cataloger "go-module-binary-cataloger"
[0001]  INFO skipping cataloger "dotnet-deps-cataloger"
[0001]  INFO skipping cataloger "portage-cataloger"
No packages discovered
```
**Anything else we need to know?**:
I can enable the cataloger with the use of `--catalogers all`, so this isn't blocking me from using that cataloger.

**Environment**:
- Output of `syft version`:
- OS (e.g: `cat /etc/os-release` or similar):

PR #1142, opened 2022-08-04
cataloger configuration is respected regardless of source

tofay:
If the user specifies catalogers via configuration, then use the
requested catalogers regardless of source type.

This makes it possible to enable an off-by-default cataloger on an image or directory without
using "--all" catalogers.

Signed-off-by: Tom Fay <tomfay@microsoft.com>

Fixes https://github.com/anchore/syft/issues/1141

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ registry:yourrepo/yourimage:tag          pull image directly from a registry (no
 - hackage
 
 #### Non Default:
-- rust-audit-binary
+- cargo-auditable-binary
 
 ### Excluding file paths
 
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ nix-shell -p syft
 
 ## Getting started
 
-#### SBOM
+### SBOM
 
 To generate an SBOM for a container image:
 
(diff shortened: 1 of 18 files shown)


tofay:
One thing to flag is that if one cataloger name is the substring of another, excluding -cataloger suffix, then it's not possible to select the former without also selecting the latter. This PR potentially exposes that, in that filtering all catalogers is likelier to lead to collisions e.g the rust cataloger name is a substring of the rust-audit cataloger.

spiffcs:
> One thing to flag is that if one cataloger name is the substring of another, excluding -cataloger suffix, then it's not possible to select the former without also selecting the latter. 

I think because the filtering is non-obvious when passing an option that isn't a specific cataloger `python => {PythonIndex, PythonPackage}` we should maybe update the filtering function to only explicitly match on cataloger names rather than sub strings. So if it sees a string that is not `all` or an official cataloger name it does not act on the filtering.

WDYT @tofay ?

tofay:
> WDYT @tofay? I'd be happy to contribute a commit on this branch that makes the filterCataloger function a little more restrictive in that way.


That sounds sensible. I imagine we'll still want to allow the  -cataloger suffix to be omitted?

spiffcs:
Awesome - yea I'll get that commit added and then this merged!

README.md
```diff
@@ -193,6 +135,47 @@ file:path/to/yourproject/file            read directly from a path on disk (any
...
+- go-mod-file
+- rust-cargo-lock
+- dartlang-lock
+- dotnet-deps
+- cocoapods
+- conan
+- hackage
+
+#### Non Default:
+- rust-audit-binary
```
tofay, line 177: Whilst there's renaming going on, we could switch this to `cargo-auditable-binary`?

Since I added syft integration we've been working on making it easier to generate binaries with audit information, and have renamed the project to cargo-auditable (to reflect the new approach which is to simply install the cargo plugin and run `cargo auditable build/install ...` rather than `cargo build/install`.

tofay: ok I can do that tomorrow

spiffcs: Doing it all right now - about to finish the enum work for the catalogers so I'll stick the rename with that

spiffcs (approved):

PR #1148, opened 2022-08-04
Add modularitylabel metadata to RPM type records generated by syft

### LINKED ISSUE #1145: Add 'rpm modularity' to rpm records generated by syft

Rpm records include a field 'modularity' that indicates which module/appstream an rpm belongs to (if any, otherwise its an empty string).  This can be important metadata that is currently not surfaced within the metadata fields associated with RPM type records generated by syft.

**Why is this needed**:

A good example for where this information will be needed is mapping RPM records against vulnerability data records that scope affected packages by modularity - see https://github.com/anchore/grype/issues/820

**Additional context**:

The rpm parsing upstream that syft uses has support for this field, so the task here is to add extraction of 'modularitylabel' and including the data in the syft generated rpm records.

nurmi:
Fixes #1145


syft/pkg/rpmdb_metadata.go
```diff
@@ -26,16 +26,17 @@ var (
...
+	Name            string            `json:"name"`
+	Version         string            `json:"version"`
+	Epoch           *int              `json:"epoch"  cyclonedx:"epoch" jsonschema:"nullable"`
+	Arch            string            `json:"architecture"`
+	Release         string            `json:"release" cyclonedx:"release"`
+	SourceRpm       string            `json:"sourceRpm" cyclonedx:"sourceRpm"`
+	Size            int               `json:"size" cyclonedx:"size"`
+	License         string            `json:"license"`
+	Vendor          string            `json:"vendor"`
+	ModularityLabel string            `json:"modularitylabel"`
```
wagoodman, line 38:
```suggestion
	ModularityLabel string            `json:"modularityLabel"`
```

Multi word fields should be camel cased

wagoodman (approved):
made a small adjustment to the schema to match the casing update

PR #1152, opened 2022-08-09
Associate node package licenses from node_modules

### LINKED ISSUE #845: No licenses included in scan with yarn.lock
**What happened**:

Get SBOM from a project result in an SPDX-JSON file with all licenses NONE.

SPDX-JSON file with licenses concluded differently than NONE

As we discussed in Slack, you only need to have a dummy nodeJS project with `yarn.lock` and `node_modules`

Nope.

**Environment**:
- Output of `syft version`: 0.38.0
- OS (e.g: `cat /etc/os-release` or similar): OS X (tested locally)

kzantow:
This PR adds a hook in the `GenericCataloger` which allows post-processing such as searching for licenses, which is also used by the `javascript` lockfile cataloger to search for `package.json` files and extract license information, if found.

Closes #845


syft/pkg/cataloger/common/generic_cataloger.go
```diff
@@ -31,6 +33,16 @@ func NewGenericCataloger(pathParsers map[string]ParserFn, globParsers map[string
 	}
 }
 
+// NewPostProcessingGenericCataloger if provided path-to-parser-function and glob-to-parser-function lookups creates a GenericCataloger
+func NewPostProcessingGenericCataloger(pathParsers map[string]ParserFn, globParsers map[string]ParserFn, upstreamCataloger string, postProcess PostProcessFunc) *GenericCataloger {
```
wagoodman, line 37: open question: we could allow for one or more of these processors and make it variadic. This would also allow for a common constructor:
```
func NewGenericCataloger(pathParsers map[string]ParserFn, globParsers map[string]ParserFn, upstreamCataloger string, processors ...PostProcessFunc) *GenericCataloger {
...
}
```

syft/pkg/cataloger/javascript/cataloger.go
```diff
@@ -23,5 +31,47 @@ func NewJavascriptLockCataloger() *common.GenericCataloger {
...
+
+func addLicenses(resolver source.FileResolver, location source.Location, p *pkg.Package) {
+	dir := path.Dir(location.RealPath)
+	pkgPath := []string{dir, "node_modules"}
+	pkgPath = append(pkgPath, strings.Split(p.Name, "/")...)
+	pkgPath = append(pkgPath, "package.json")
+	pkgFile := path.Join(pkgPath...)
+	locations, err := resolver.FilesByPath(pkgFile)
+	if err != nil || len(locations) == 0 {
+		log.Debugf("no package.json found at: %s", pkgFile)
```
wagoodman, line 45: this could be a little noisy, is it necessary to log that there is no package.json?

Also, should we put out a warning when there is an error? I feel that these cases should be handled separately.

wagoodman (approved):

martin-langhoff:
Yay!!!

PR #1155, opened 2022-08-10
Give the contributing guide a substantial rework

joshbressers:
This changes the entire flow of the contributing guide. Please consider this a first draft (it still needs more details).

I tried to put the guide into a logical order for completing and understanding how to best start to contribute to Syft. I would value any feedback on the structure and if the current order makes sense.

The setup instructions are only for Debian at the moment. I will include at least OSX at some point in the near future (it takes a lot longer than one would expect to setup a clean machine).

Signed-off-by: Josh Bressers <josh@bress.net>


spiffcs (approved):

spiffcs:
No notes on this! We can always change it or break it out in future PR, but this is a substantial update with really good information so approved!

PR #1156, opened 2022-08-11
Correctly populate hasFiles field for spdx-json output

jedevc:
Heya :wave: Have been playing around with Syft recently, and came across some odd behavior! It looks like the `hasFiles` field isn't populating correctly, despite the relationships being in-place as expected, and there being logic to extract for this field.

Previously, extracting relationships between packages and files was not completing correctly, as SPDXRef- ElementIDs were being compared to raw IDs, and so never matched. This PR fixes this behavior, by ensuring that we always compare
ElementIDs.

I think the original behavior was introduced in https://github.com/anchore/syft/pull/634, and while the unit tests pass, I can't find any integration tests that check the behavior - but would be happy to add one, if you have any pointers on where might be the best place! :tada:

Let me know if there's anything else you need, happy to contribute!

### DIFF
--- a/internal/formats/spdx22json/to_format_model.go
+++ b/internal/formats/spdx22json/to_format_model.go
@@ -112,8 +112,10 @@ func fileIDsForPackage(packageSpdxID string, relationships []artifact.Relationsh
 			continue
 		}
 
-		if string(relationship.From.ID()) == packageSpdxID {
-			fileIDs = append(fileIDs, string(relationship.To.ID()))
+		from := model.ElementID(relationship.From.ID()).String()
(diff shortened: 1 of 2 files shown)


kzantow:
Good catch @jedevc ! As for a test, maybe just add another unit test case in the `to_format_model_test.go` in the same package?

jedevc:
Actually, on second thoughts, fixing the test case that I broke from the change is hopefully good enough!

kzantow:
@jedevc agree -- looks like that test case should cover it just fine 👍

kzantow (approved):

Issue #1115, opened 2022-07-22
Online verification of artifacts


wagoodman:
This issue is meant to be a spot to host discussion on a couple of related topics:

- should syft **gather** information from external sources (e.g. maven.org, pypi.org, rubygems.org, etc.) in order in enrich the information provided in the SBOM with select point-in-time external package data? For example, for all jars found, use the SHA1 of the Jar to search for an authoritative ArtifactID and GroupID for the Jar (since sometimes the packaged data is inaccurate or missing). 

- should syft **verify** information found within a scanned artifacts against external sources? And if so, list verification claims directly in the SBOM?

This issue has intentionally been left open-ended to gather feedback and specific use cases from the community.

joshbressers:
These are great questions @wagoodman 

My knee jerk reaction to this was "why wouldn't it!"

But upon further thought, I'm less certain.

Today Syft is functionally a tool for taking a snapshot of some collection of artifacts (I know it does some other things like convert between SBOM formats, but that's a different discussion, let's just pretend it only collects details right now).

I think you describe in both of these cases is a second pass operation. Pass one is to collection details, pass two is to enrich the data. If we make Syft do both of these, we will be adding A LOT of new functionality to Syft to increase complexity.

Maybe the real question is should Syft do one thing, or should Syft do many things?

There would be massive value to use this information to enrich an SBOM, I suspect we all agree we want a tool to do this. Should Syft do it or should a new tool do it?

westonsteimel:
I had always envisioned some other tool after syft doing the enrichment.  Then you have a trail from the original sbom and an option to use the enrichment or not without complicating syft with more options to do everything

xtreme-conor-nosal:
While it looks like a decision is already made via #1158 I wanted to link to https://github.com/anchore/syft/issues/1129 as a sample for this discussion. As some ecosystems don't commit full dependency tree info (e.g. spring boot pom.xml) into the git repo, an un-enriched scan isn't able to produce a complete and accurate sbom from the source repo.

While there's a few directions enrichment can go in, they all have tradeoffs:
- wait for the build process (e.g. the jar manifests are more complete), but this is a slower feedback loop
- resolve external sources similar to how the build would, adding a larger footprint to Syft and requiring user configuration
- require the project directory to be "initialized" or have the build toolchain present, adding overhead to CI integration

wagoodman:
[...]

PR #1158, opened 2022-08-11
External sources configuration

mdeicas:
The use of external sources is new to Syft, and they must be managed thoughtfully (i.e. configurability, clear to users what has been used and how). This PR proposes a way to do so. It is motivated by https://github.com/anchore/syft/issues/1159 and https://github.com/anchore/syft/issues/1115. 

It adds a new external sources configuration, an additional function that catalogers must implement, and a cli flag to shut off the use of external sources. This approach assumes that external sources will only come into Syft through catalogers. 

When catalogers are implemented that use external sources, they must add their own configuration options to the application-wide external sources configuration created here.

### DIFF
--- a/syft/pkg/cataloger/cataloger.go
+++ b/syft/pkg/cataloger/cataloger.go
@@ -41,6 +41,8 @@ type Cataloger interface {
 	Name() string
 	// Catalog is given an object to resolve file references and content, this function returns any discovered Packages after analyzing the catalog source.
 	Catalog(resolver source.FileResolver) ([]pkg.Package, []artifact.Relationship, error)
+	// UsesExternalSources returns if the cataloger uses external sources, such as querying a database
+	UsesExternalSources() bool
 }
 
 // ImageCatalogers returns a slice of locally implemented catalogers that are fit for detecting installations of packages.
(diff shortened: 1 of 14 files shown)


lumjjb (approved):
LGTM!

spiffcs (approved):

spiffcs:
Brilliant! I'll get this one merged in first. Thanks a million!

syft/pkg/cataloger/cataloger.go
```diff
@@ -41,6 +41,8 @@ type Cataloger interface {
 	Name() string
 	// Catalog is given an object to resolve file references and content, this function returns any discovered Packages after analyzing the catalog source.
 	Catalog(resolver source.FileResolver) ([]pkg.Package, []artifact.Relationship, error)
+	// UsesExternalSources returns if the cataloger uses external sources, such as querying a database
+	UsesExternalSources() bool
```
wagoodman, line 45: I don't think we want to make this part of the cataloger interface, instead we should be passing in indication of this configuration to cataloger constructors, and constructors that return `nil` instances and `nil` error are discarded from the list of package catalogers.

wagoodman (commented):
I neglected get this in on time! @spiffcs can you tackle follow ups to this?

PR #1161, opened 2022-08-16
Capture if a node module is private

### LINKED ISSUE #1160: Capture package.json private field for npm modules

The [`private` field within package.json](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#private) for a node module when set to true indicates that the package cannot be published to a public registry. I'd like to see this field captured within a Syft SBOM along side other NpmPackageJsonMetadata metadata.

**Why is this needed**:

Node applications will often define themselves as private modules using generic names. When this name matches a published module with CVEs, these CVEs end up being reported by Grype against the private module that is in no way related to the public module of the same name. If we can identify the Syft artifact as a private modules, then we know we can ignore vulnerabilities reported for that artifact. In time we can also try to make Grype smarter and not report the vulnerabilities in the first place.

**Additional context**:

scothis:
The private field within a node module's package.json indicates when
true that the package cannot be published to a registry. This is a
strong indication that public CVE reported against a component of the
same name are likely do not affect this particular module.

This change introduce a Private bool to NpmPackageJSONMetadata that is
exposed in the Syft SBOM's metadata for an npm package artifact. It does
not directly consume the value in any way.

Resolves #1160 

Signed-off-by: Scott Andrews <andrewssc@vmware.com>


tgerla:
Hi @scothis, thank you very much for your contribution! We'll review it as soon as we can and get back to you with any comments or questions.

spiffcs (approved):

spiffcs:
Approved - adding private field to metadata LGTM!

spiffcs:
^ I'll update this test - looks like the license list has been updated behind the scenes.

PR #1167, opened 2022-08-18
workflow to create automated PRs to update bootstrap tools

westonsteimel:

### DIFF
--- a/.github/workflows/update-bootstrap-tools.yml
+++ b/.github/workflows/update-bootstrap-tools.yml
@@ -1,0 +1,75 @@
+name: PR for latest versions of bootstrap tools
+on:
+  schedule:
+    - cron: "0 8 * * *" # 3 AM EST
+
+  workflow_dispatch:
+
+env:
+  GO_VERSION: "1.18.x"
+  GO_STABLE_VERSION: true
+
+jobs:
+  update-bootstrap-tools:
+    runs-on: ubuntu-latest
+    if: github.repository == 'anchore/syft' # only run for main repo
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          stable: ${{ env.GO_STABLE_VERSION }}
+
+      - run: |
+          GOLANGCILINT_LATEST_VERSION=$(go list -m -json github.com/golangci/golangci-lint@latest 2>/dev/null | jq -r '.Version')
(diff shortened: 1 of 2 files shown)


.github/workflows/update-bootstrap-tools.yml
```diff
@@ -0,0 +1,75 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          stable: ${{ env.GO_STABLE_VERSION }}
+
+      - run: |
+          GOLANGCILINT_LATEST_VERSION=$(go list -m -json github.com/golangci/golangci-lint@latest 2>/dev/null | jq -r '.Version')
```
spiffcs, line 25: TIL `ubuntu-latest` comes with jq

.github/workflows/update-bootstrap-tools.yml
```diff
@@ -0,0 +1,75 @@
...
+          # export the versions for use with create-pull-request
+          echo "::set-output name=GOLANGCILINT::$GOLANGCILINT_LATEST_VERSION"
+          echo "::set-output name=BOUNCER::BOUNCER_LATEST_VERSION"
+          echo "::set-output name=CHRONICLE::CHRONICLE_LATEST_VERSION"
+          echo "::set-output name=GORELEASER::GORELEASER_LATEST_VERSION"
+          echo "::set-output name=YAJSV::YAJSV_LATEST_VERSION"
+          echo "::set-output name=COSIGN::COSIGN_LATEST_VERSION"
+        id: latest-versions
+
+      - uses: tibdex/github-app-token@v1
```
spiffcs, line 53: Interesting! Why do we need to use this action for the workflow?

spiffcs (approved):
Approved with comments! I guess the best way to check the correctness here is to let it run and smooth any edge cases

.github/workflows/update-bootstrap-tools.yml
westonsteimel: I don't actually know, that was something @kzantow had as part of the stereoscope update one, so I assumed it would need to be here also

.github/workflows/update-bootstrap-tools.yml
westonsteimel: I was just assuming because we have https://github.com/anchore/grype/blob/ae37eb4a05acfacf551138e15a2266a0894d4399/.github/workflows/update-syft-release.yml#L25

PR #1172, opened 2022-08-23
feat: enable formatters to take in external library structs

cpendery:
## Description
This pr enables the `spdx` and `cyclonedx` formatters to be called directly with those external libraries objects. Currently, if you build an sbom using these structs, you have to marshal it into json and then pass the reader to be un-marshaled and parsed into Syft's intermediate representation, causing wasted computations, especially when working with large SBOMs. 
The often comes into play when calling Syft to produces packages for Grype's api methods. 

Considerations:
**Don't use `interface{}`**: I agree, but generics got a bit out of hand considering the `Format` struct is quite widely used. Happy to refactor as desired

### DIFF
--- a/internal/formats/common/cyclonedxhelpers/parser.go
+++ b/internal/formats/common/cyclonedxhelpers/parser.go
@@ -1,0 +1,17 @@
+package cyclonedxhelpers
+
+import (
+	"github.com/CycloneDX/cyclonedx-go"
+	"github.com/anchore/syft/syft/sbom"
+)
+
+func GetParser() sbom.Parser {
+	return func(input interface{}) (*sbom.SBOM, error) {
+		switch input := input.(type) {
+		case *cyclonedx.BOM:
+			return toSyftModel(input)
+		default:
+			return nil, sbom.ErrParsingNotSupported
+		}
+	}
+}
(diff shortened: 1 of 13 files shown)


spiffcs:
I think I'm ok with interface here for now considering the overhead of the generic work.

cc @wagoodman since he usually has the final say when it comes to updating the interface types.

Code LGTM - I think just the consideration is worth the second pair of eyes.

spiffcs (approved):

kzantow:
@cpendery thanks for attending the community meeting to add context to this change -- do you have a specific format you're dealing with and would it just be easier to export the right conversion functions by moving these formatters to the top-level `syft` package and exporting the `ToSyftModel` functions?

cpendery:
> @cpendery thanks for attending the community meeting to add context to this change -- do you have a specific format you're dealing with and would it just be easier to export the right conversion functions by moving these formatters to the top-level `syft` package and exporting the `ToSyftModel` functions?

I'm just working with the cyclonedx and spdx models. Just exposing the ToSyftModel function outside internal would be great and is all I actually need

kzantow:
@cpendery would you like to take a stab at the refactoring? If not, I could probably do this in the next day or two.

cpendery:
> @cpendery would you like to take a stab at the refactoring? If not, I could probably do this in the next day or two.

I can refactor the pr and just pull out the formats package 👍

spiffcs (approved):

PR #1157, opened 2022-08-11
Add SBOM Discovery with Rekor

mdeicas:
This PR adds the ability to discover build-time SBOMs of binaries with the Rekor transparency log and to create external document references for them in SPDX JSON. Explained in more detail in https://github.com/anchore/syft/issues/1159

PR #1175, opened 2022-08-24
improvements to rekor cataloger code

mdeicas:
This PR, to `kubecon-draft`, addresses many of the comments from the original PR for [SBOM discovery with Rekor](https://github.com/anchore/syft/pull/1157). 

One of the bigger changes is returning a `subject` and `sbomEntry` from `parseAndValidateAttestation` instead of a `InTotoAttestation`. This is to reduce the amount of validations that are needed in the code. 

Another change is adding a configuration for `rekor-cataloger`. As the cataloger is off-by-default, the following must be specified in a configuration file:
```yaml
rekor-cataloger:
  cataloger:
    enabled: true
``` 

To avoid this, a CLI flag specifically for the rekor-cataloger could be added. Alternatively, a more general CLI flag that filters the file catalogers or the non-package catalogers could be implemented, similar to the cli flag for package catalogers. 

I had some difficulty enabling the `rekor-cataloger` via the configuration when there were other configuration fields also specified. If this is encountered, use a configuration file containing only the rekor-cataloger configuration. I will look at this issue further.  

Signed-off-by: Marco Deicas <mdeicas@google.com>

### DIFF
--- a/syft/file/rekor_cataloger.go
+++ b/syft/file/rekor_cataloger.go
@@ -1,4 +1,4 @@
-package rekor
+package file
 
 import (
 	"fmt"
@@ -6,8 +6,7 @@ import (
 	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
-	"github.com/anchore/syft/syft/pkg"
-	rekorLib "github.com/anchore/syft/syft/rekor"
+	"github.com/anchore/syft/syft/rekor"
 	"github.com/anchore/syft/syft/source"
 )
 
@@ -16,10 +15,12 @@ import (
 
 const catalogerName = "rekor-cataloger"
 
-type Cataloger struct{}
+type Cataloger struct {
+	client *rekor.Client
+}
 
-func NewRekorCataloger() *Cataloger {
-	return &Cataloger{}
+func NewRekorCataloger(client *rekor.Client) *Cataloger {
+	return &Cataloger{client: client}
 }
 
 func (c *Cataloger) Name() string {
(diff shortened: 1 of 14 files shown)


syft/file/rekor_cataloger.go
```diff
@@ -30,26 +31,25 @@ func (c *Cataloger) UsesExternalSources() bool {
 	return true
 }
 
-func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, []artifact.Relationship, error) {
+func (c *Cataloger) Catalog(resolver source.FileResolver) ([]artifact.Relationship, error) {
```
spiffcs, line 34: design: The return signature here is interesting and I think I need to explore the file package code a bit more before giving good feedback here.

I looked at the other catalogers in this package:
- classification
- contents
- digest
- metadata
- secrets

Each of those returns something akin to `map[source.Coordinates]someValue`. Even though this is not enforced by an interface implementation at a higher level, do we want the `rekor_cataloger` to be different in this case and return `[]Relationship` as the output for `Catalog`?

Personally, I think this is ok since we're describing a union to an external source and we're doing all of the work upfronts in the cataloger of creating the external ref and relationships without additional overhead. I just wanted to point out that this cataloger was a little different from its current sibblings.

syft/rekor/query.go
```diff
@@ -21,29 +20,19 @@ import (
...
-		log.Info("attestation found on Rekor with multiple SBOMS, which is not currently supported. Proceeding with the first SBOM.")
-	}
-	uri := att.Predicate.Sboms[0].URI
-	if uri == "" {
-		return nil, errors.New("uri of sbom is empty")
-	}
-
-	resp, err := client.Get(uri)
+func getSbom(sbomEntry sbomEntry, client *http.Client) (*[]byte, error) {
+	resp, err := client.Get(sbomEntry.URI)
```
spiffcs, line 27: Nice great up here

syft/rekor/query.go
```diff
@@ -148,12 +134,11 @@ func getAndVerifySbomFromUUID(uuid string, client *Client) (*sbomWithMetadata, e
 	}
 
 	sbomWrapped := &sbomWithMetadata{
-		executableSha256: att.Subject[0].Digest["sha256"],
+		executableSha256: subject.Digest["sha256"],
```
spiffcs, line 137: Nice!

syft/rekor/utils.go
```diff
@@ -93,10 +94,11 @@ func parseEntry(entry *models.LogEntryAnon) (*models.IntotoV001Schema, error) {
 	return intotoV001, nil
 }
 
-// validateAttestation does some checks that the attestation contains the necessary fields to proceed
+// validateAttestation validates that the attestation contains the necessary fields to proceed, and
```
spiffcs, line 97: nit: I think the function name changed here validateAttestation => validateAndGetFirsts

syft/rekor/utils.go
```diff
@@ -108,40 +110,42 @@ func validateAttestation(att *InTotoAttestation) error {
 	} else if len(att.Subject) > 1 {
 		err = errors.New("attestation found on rekor contains multiple subjects, which is not supported. Ignoring log entry")
 	} else if _, ok := att.Subject[0].Digest["sha256"]; !ok {
```
spiffcs, line 112: nit: Just to protect from any incomplete `att` that could be uploaded it might be best to verify the `Subject` at the zero index exists before doing this map check. There also might be an opportunity to exit early in the `len(att.Subject) == 0` case

spiffcs (commented):
Small comments - great job and thanks so much for the extra mile in fitting in these last-minute changes!

spiffcs:
cc @wagoodman

syft/file/rekor_cataloger.go
mdeicas: I agree that this is ok. I think that the ability to implement catalogers with different return types is one of the benefits of implementing a cataloger at the task level as opposed to the package-cataloger level. `syft.CatalogPackages` in `generateCatalogPackagesTask` also has a different return type.

PR #1188, opened 2022-08-31
Add RPM file scanning support

kzantow:
This PR adds support for scanning `*.rpm` files found in the filesystem. Part of this PR renames the `RpmdbMetadata` type to `RpmMetadata`, as well as supporting ingesting existing SBOMs with `RpmdbMetadata`.

Prerequisite for: https://github.com/anchore/grype/issues/570


.github/workflows/validations.yaml
```diff
@@ -95,6 +95,16 @@ jobs:
...
 
+      - name: Build cache key for rpm test-fixture blobs (for unit tests)
+        run: make rpm-binaries-fingerprint
+
+      - name: Restore RPM test-fixture cache
+        id: unit-rpm-cache
+        uses: actions/cache@v2.1.3
+        with:
+          path: syft/pkg/cataloger/rpm/test-fixtures/rpms
+          key: ${{ runner.os }}-unit-rpm-cache-${{ hashFiles( 'syft/pkg/cataloger/rpm/test-fixtures/rpms.fingerprint' ) }}
```
kzantow, line 106: This could be simplified to just hash the Makefile, I've left it this way for consistency

go.mod
```diff
@@ -79,6 +80,7 @@ require (
 	github.com/Azure/go-autorest/autorest/date v0.3.0 // indirect
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
+	github.com/DataDog/zstd v1.4.5 // indirect
```
wagoodman, line 83: this repo contains a lot of C code -- can you double check that we are not becoming reliant on CGO for this functionality? (our build process requires that CGO is disabled, so if there was an issue I would presume that our snapshot build would fail... so I don't think there is an issue)

syft/pkg/cataloger/rpm/db_cataloger.go
```diff
@@ -1,39 +1,39 @@
...
 
 	"github.com/anchore/syft/internal"
-
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/source"
 )
 
-const catalogerName = "rpmdb-cataloger"
+const dbCatalogerName = "rpmdb-cataloger"
```
wagoodman, line 16: should we rename to `rpm-db-cataloger`?

syft/pkg/cataloger/rpm/file_cataloger.go
```diff
@@ -0,0 +1,141 @@
...
+	return &FileCataloger{}
+}
+
+// Name returns a string that uniquely describes a cataloger
+func (c *FileCataloger) Name() string {
+	return "rpm-file-cataloger"
+}
+
+// UsesExternalSources indicates that the rpm file cataloger does not use external sources
+func (c *FileCataloger) UsesExternalSources() bool {
```
wagoodman, line 31: @spiffcs Did we revert this interface change?

syft/pkg/cataloger/rpm/file_cataloger_test.go
```diff
@@ -0,0 +1,106 @@
...
+			expected: map[string]pkg.Package{
+				"abc": {
+					Name:         "abc",
+					Version:      "1.01",
+					Locations:    source.NewLocationSet(),
+					FoundBy:      "rpm-file-cataloger",
+					Type:         pkg.RpmPkg,
+					MetadataType: pkg.RpmdbMetadataType,
+					Licenses:     []string{"MIT"},
+					Metadata: pkg.RpmdbMetadata{
```
wagoodman, line 30: we should be creating a new distinct `RpmMetadata` even if the type ends up being dup'd:
```golang
type RpmMetadata RpmdbMetadata
```

syft/pkg/cataloger/cataloger.go
```diff
@@ -54,7 +54,8 @@ func ImageCatalogers(cfg Config) []Cataloger {
 		php.NewPHPComposerInstalledCataloger(),
 		javascript.NewJavascriptPackageCataloger(),
 		deb.NewDpkgdbCataloger(),
-		rpmdb.NewRpmdbCataloger(),
+		rpm.NewRpmdbCataloger(),
+		rpm.NewFileCataloger(),
```
wagoodman, line 58: The image catalogers should only include catalogers that detect installed packages. The RPMDB is a good indicator of this, however, an RPM file in the image is not.

go.mod
kzantow: I'm not quite sure how to check all possible permutations for this, but explicitly running `CGO_ENABLED=0 go build -o main ./cmd/syft` works fine

syft/pkg/cataloger/rpm/file_cataloger.go
kzantow: `UsesExternalSources` is part of the `Cataloger` interface: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/cataloger.go#L45

go.mod
kzantow: A deeper dive resulted in this being included because the `go-rpmutils` has a conditional if CGO enabled, if not enabled, then a pure go library is used: https://github.com/sassoftware/go-rpmutils/blob/e5cf0f808585c10c36b3c133f0a524b93afd34d1/uncompress_nocgo.go#L1-L9

syft/pkg/cataloger/rpm/file_cataloger_test.go
kzantow: From offline discussion: we will rename `RpmdbMetadata` -> `RpmMetadata`, and make sure to handle decoding of existing SBOMs with a `metadataType` of `RpmdbMetadata`

wagoodman (approved):
🚀

Issue #292, opened 2020-12-15
Compare two SBOMs


wagoodman:
It would be interesting to enable the ability to provide a previous SBOM result and natively "diff" the results without having to result in scripting. Something like:

```
syft -o json my-image:latest > sbom-1.json
# ... time passes ...
syft -o json my-image:latest > sbom-2.json

# what are the differences between these two SBOMs?
syft diff sbom-1.json sbom-2.json

Packages added:
- package name 1 (java)
- package name 2 (java)

Packages removed:
- package name 3 (python)

Packages changed:
- package name 4 (ruby)
```

The output should probably support both a summary and JSON output formats, the example output format is only for illustration.

In this way a user can get a feeling for what's changing across your product SBOM overtime.

coderpatros:
We have a similar issue logged for the CycloneDX CLI tool https://github.com/CycloneDX/cyclonedx-cli/issues/31

It would be great to collaborate and align the output format if possible.

For CycloneDX I was thinking of having two output formats. The first, basic text output that is more human friendly. The second, JSON format that is more machine friendly.

ckotzbauer:
Maybe this is interesting for one of you: https://github.com/ckotzbauer/sbom-diff-action
This action creates a Diff from two Syft-SBOMs and attaches the diff to PRs as comment.

PR #1194, opened 2022-09-07
Normalize syft-json output

scothis:
There are numerous places within syft where a map is iterated over to create a list. Since map iteration is non-deterministic, the order of items within the list is also non-deterministic. While this doesn't impact the semantic meaning of the output, it makes it harder to compare the output of two independent runs of syft.

```sh
syft ghcr.io/jenkins-x/jx-scm -o syft-json > a.json
syft ghcr.io/jenkins-x/jx-scm -o syft-json > b.json
diff a.json b.json
```

Without this PR there is a diff, with this PR the diff is empty.

This PR is a band-aid to address the most common sources of drift. There are certainly other sources that are not addressed.

Related #331 #292

Signed-off-by: Scott Andrews <andrewssc@vmware.com>

### DIFF
--- a/internal/formats/syftjson/to_format_model.go
+++ b/internal/formats/syftjson/to_format_model.go
@@ -215,6 +215,15 @@ func toRelationshipModel(relationships []artifact.Relationship) []model.Relation
 			Metadata: r.Data,
 		}
 	}
+	sort.Slice(result, func(i, j int) bool {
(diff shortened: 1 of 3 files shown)


kzantow (approved):
Thanks! This looks 👍

scothis:
Looks like a GitHub flake caused a test to fail https://github.com/anchore/syft/runs/8230778367?check_suite_focus=true#step:6:66

PR #1206, opened 2022-09-15
feat: Add support for npm lockfile version 3

### LINKED ISSUE #1203: NPM package-lock.json version 3
Add support for parsing `package-lock.json` version 3.

**Why is this needed**:
There is an incompatible change from version 1 and 2 which Syft should support.

**Additional context**:
Version 1 has a structure like:
```
    "dependencies": {
        "@types/prop-types": {
            "version": "15.7.5",
            "resolved": "https://registry.npmjs.org/@types/prop-types/-/prop-types-15.7.5.tgz",
            "integrity": "sha1-XxnSuFqY6VWANvajysyIGUIPBc8="
        },
```
Version 2 added a duplication of data under the `packages` field, and version 3 removes the `dependencies` field altogether, with a structure like:
```
    "packages": {
        "node_modules/@types/prop-types": {
            "version": "15.7.5",
            "resolved": "https://registry.npmjs.org/@types/prop-types/-/prop-types-15.7.5.tgz",
            "integrity": "sha1-XxnSuFqY6VWANvajysyIGUIPBc8=",
            "license": "MIT"
        },
```

See the NPM docs here: https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json

robcresswell:
This PR adds support for npm lockfile version 3, which drops the "dependencies" key and uses "packages" instead. I've refactored the lockfile parser to make the distinction between the versions explicit rather than the implicit behaviour before. It _might_ be worth splitting into separate files at some point, but the logic is so minimal that I haven't done it.

Some open questions;

- Does the code look vaguely correct? I don't know Go well at all
- I can't find good documentation around the presence of the "license" key under the "packages" entries. It seems to be present in the v2 fixture, but I couldn't recreate that locally.
- Are there other places that I need to add / update tests?

Fixes #1203

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_package_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_package_lock.go
@@ -32,10 +32,11 @@ type lockDependency struct {
 }
 
 type lockPackage struct {
+	Name      string `json:"name"` // only present in the root package entry (named "")
 	Version   string `json:"version"`
 	Resolved  string `json:"resolved"`
 	Integrity string `json:"integrity"`
-	License   string `json:""`
+	License   string `json:"license"`
 }
 
 // parsePackageLock parses a package-lock.json and returns the discovered JavaScript packages.
@@ -49,27 +50,41 @@ func parsePackageLock(resolver source.FileResolver, _ *generic.Environment, read
 	var pkgs []pkg.Package
 	dec := json.NewDecoder(reader)
 
+	var lock packageLock
 	for {
-		var lock packageLock
 		if err := dec.Decode(&lock); errors.Is(err, io.EOF) {
 			break
 		} else if err != nil {
 			return nil, nil, fmt.Errorf("failed to parse package-lock.json file: %w", err)
 		}
-		licenseMap := make(map[string]string)
-		for _, pkgMeta := range lock.Packages {
-			var sb strings.Builder
-			sb.WriteString(pkgMeta.Resolved)
-			sb.WriteString(pkgMeta.Integrity)
-			licenseMap[sb.String()] = pkgMeta.License
-		}
+	}
 
+	if lock.LockfileVersion == 1 {
 		for name, pkgMeta := range lock.Dependencies {
-			pkgs = append(pkgs, newPackageLockPackage(resolver, reader.Location, name, pkgMeta, licenseMap))
+			pkgs = append(pkgs, newPackageLockV1Package(resolver, reader.Location, name, pkgMeta))
+		}
+	}
+
+	if lock.LockfileVersion == 2 || lock.LockfileVersion == 3 {
+		for name, pkgMeta := range lock.Packages {
+			if name == "" {
+				if pkgMeta.Name == "" {
+					continue
+				} else {
+					name = pkgMeta.Name
+				}
+			}
+
+			pkgs = append(pkgs, newPackageLockV2Package(resolver, reader.Location, getNameFromPath(name), pkgMeta))
 		}
 	}
 
 	pkg.Sort(pkgs)
 
 	return pkgs, nil, nil
+}
+
+func getNameFromPath(path string) string {
+	parts := strings.Split(path, "node_modules/")
+	return parts[len(parts)-1]
 }
--- a/syft/pkg/cataloger/javascript/test-fixtures/pkg-lock/package-lock-3.json
+++ b/syft/pkg/cataloger/javascript/test-fixtures/pkg-lock/package-lock-3.json
@@ -1,0 +1,40 @@
+{
+  "name": "lock-v3-fixture",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "lock-v3-fixture",
+      "version": "1.0.0",
+      "dependencies": {
+        "@types/react": "^18.0.9"
+      }
+    },
+    "node_modules/@types/prop-types": {
+      "version": "15.7.5",
+      "resolved": "https://registry.npmjs.org/@types/prop-types/-/prop-types-15.7.5.tgz",
+      "integrity": "sha512-JCB8C6SnDoQf0cNycqd/35A7MjcnK+ZTqE7judS6o7utxUCg6imJg3QK2qzHKszlTjcj2cn+NwMB2i96ubpj7w=="
+    },
+    "node_modules/@types/react": {
+      "version": "18.0.20",
+      "resolved": "https://registry.npmjs.org/@types/react/-/react-18.0.20.tgz",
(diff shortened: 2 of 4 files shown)


syft/pkg/cataloger/javascript/parse_package_lock.go
```diff
@@ -47,38 +47,51 @@ func parsePackageLock(path string, reader io.Reader) ([]*pkg.Package, []artifact
...
 				Language: pkg.JavaScript,
 				Type:     pkg.NpmPkg,
-				Licenses: licenses,
 			})
 		}
 	}
 
+	if lock.LockfileVersion == 2 || lock.LockfileVersion == 3 {
+		for name, pkgMeta := range lock.Packages {
+			if name == "" {
```
kzantow, line 72: I wonder if we should include this? We'd need to read the `name` field, it's an indication what this package itself is...

syft/pkg/cataloger/javascript/parse_package_lock.go
```diff
@@ -47,38 +47,51 @@ func parsePackageLock(path string, reader io.Reader) ([]*pkg.Package, []artifact
...
+				continue
+			}
+
+			newPackage := &pkg.Package{
+				Name:     getNameFromPath(name),
```
kzantow, line 77: I think we might need to add a custom `NodeMetadata` or something to capture the full "name" so the license lookup will be accurate if there isn't a license defined in the `package-lock` file

wagoodman: If we do that we should distinguish that from the existing `NpmPackageJsonMetadata` with something like `NpmPackageLockMetadata`.  If we go that directory it would be a good opportunity to add some of the useful lock information that we're parsing but not capturing today: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/javascript/parse_package_lock.go#L28-L29

syft/pkg/cataloger/javascript/test-fixtures/pkg-lock/package-lock-3.json
```diff
@@ -0,0 +1,40 @@
+{
```
wagoodman, line 1: nit/question: it seems that there is no license examples in this fixture, is that intentional? or does the V3 not include this?

wagoodman: disregard, I understand your PR comment now 🤦

kzantow: I was also thinking it would be nice to have licenses in the test fixture... it's documented on [the packages](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#packages) field:
> Package descriptors have the following fields:
> ...
> * bin, license, engines, dependencies, optionalDependencies: fields from package.json

I'm not sure offhand how to get a file like this generated, though...

robcresswell: Yeah, this is the same issue I had. Given this isn't urgent, let me do some digging and see if I can figure out why these keys appear (or not)

syft/pkg/cataloger/javascript/parse_package_lock.go
robcresswell: Can you elaborate on the license gathering? v1 lockfiles only use name and version, not path, so I'm a little confused as to what could've broken here. Also a good opportunity to write a new test

syft/pkg/cataloger/javascript/parse_package_lock.go
robcresswell: Is there a place it should be read to? One of the return types?

kzantow: I think just added to the `Dependency` struct above. And then `if name == "" && pkgMeta.Name != "" { name = pkgMeta.Name }` or so

syft/pkg/cataloger/javascript/parse_package_lock.go
kzantow: Yeah, I should have noted this earlier -- there's a secondary lookup for licenses from `node_modules`: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/javascript/cataloger.go#L38 this has the assumption that the package name is found like `node_modules/<package-name>`, but in the case of a nested module like `node_modules/<different-package>/node_modules/<package-name>` we should retain that information so the license lookup is accurate

robcresswell: Ah okay, I think I follow now. Would the new metadata type be correct, or is the Locations data better? I'm a little confused on what Locations are

syft/pkg/cataloger/javascript/parse_package_lock.go
robcresswell: Just following up on this. I'm not sure I follow the solution; how does Syft make use of the name of the root package? Do I return it somewhere?

syft/pkg/cataloger/javascript/test-fixtures/pkg-lock/package-lock-3.json
robcresswell: https://github.com/npm/cli/issues/5532 << Tracking the `license` issue here

spiffcs:
Just to get this on the main thread this PR is BLOCKED: https://github.com/npm/cli/issues/5532

syft/pkg/cataloger/javascript/parse_package_lock.go
kzantow: Let's say we `git clone github.com/facebook/react` and then we `syft dir:react`, we wouldn't have the _react_ package anywhere, even though we were directly scanning `react`. I think we should make packages from this "root"

kzantow: _Packages_ aren't dependencies -- they are _the things found when scanning_.

robcresswell: Well, sure, but if you say "scan my application" do we typically list the application as a package?

robcresswell: (Thats not rhetorical, I have no idea how Syft works currently)

kzantow: If you say: "scan this node binary" we return "we found a node binary at version X", if we can. There is some inconsistence here, to be frank, but I think we should definitely include this stuff if we have the information. For a `go` binary, for example: it may not actually include the root project name, so we only have dependencies to surface.

robcresswell: Okay, I've updated to include the root name

kzantow (approved):
LGTM

Mikcl:
made some follow up comments 
https://github.com/anchore/syft/commit/9d8244bae6aab260d5ddb504bc1de278d0f4bf10 

addressing in PR:
https://github.com/anchore/syft/pull/1349

PR #1211, opened 2022-09-16
refactor: replace ioutil=>io; update linter

spiffcs:
## Summary

Syft's `DEVELOPING.md` instructs the user to boostrap tools and run `make`

The default make step has a couple of errors when running out of the box.

The first is `make lint`:
Screen Shot 2022-09-16 at 1 29 53 PM

These errors are correct by changing the directives at the top of these files to the newer `nolint:all`

The second is where `io/util` is considered deprecated in the latest Golang release. These instances have been updated to their latest `os` or `io` equivlants.

Abandoned linters have also been disabled per the warnings from the `golangci` tool:
Screen Shot 2022-09-16 at 1 32 08 PM

### Changes
- remove deprecated linters that have been abandoned
- replace usage of ioutil => io/os upgrades
- update std library copied files to new lint directive

### Follow up
- updating developer.md with better flow diagrams, explanation of what default  `make` does, and shorter introduction commands to get new users up to speed

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -10,7 +10,6 @@ linters:
   enable:
     - asciicheck
     - bodyclose
-    - deadcode
     - depguard
     - dogsled
     - dupl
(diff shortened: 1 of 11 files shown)


.golangci.yaml
```diff
@@ -10,7 +10,6 @@ linters:
   enable:
     - asciicheck
     - bodyclose
-    - deadcode
```
kzantow, line 13: should the other linters be added to the commented section below?

kzantow (approved):
LGTM just one question about adding the removed linters to the comments

.golangci.yaml
spiffcs: I move the ones that were relevant to the generic issue. The ones that have been deleted are: 

`[WARN] x is deprecated. Replaced by unused.`

I can include them commented out if we want to ever incorporate them again if they are picked up by a maintainer

kzantow: Ah no need if they're deprecated 👍

PR #1029, opened 2022-06-05
SBOM cataloger

patrikbeno:
Supports ad hoc binary application layouts:

CI saves CycloneDX SBOM JSON in docker image or application deployment/installation folder. 

During image/directory scanning, Syft does not need to rely on specific layouts supported by specific catalogers.

Instead, application may use whatever layout and format it needs, and provides SBOM JSON artifact for integration with Syft or other tools.

Typical scenario is a node/react/webpack build folder with static content for web server. This particular use case can be supported by providing npm/yarn lock file and enabling javascript-lock-cataloger (#1022). 

SBOM cataloger provides more generic approach to this problem.

PR #1218, opened 2022-09-19
refactor!: Add support for multiple sources and distros

kzantow:


kzantow:
Closing this PR as we do not feel this is the right direction at this time, but rather only supporting additional package data within an SBOM (as implemented in PR #1029), not multiple sources within the same SBOM. We will investigate adopting or introducing a format that allows for multiple _SBOMs_ as we feel this more closely aligns with the needs today.

PR #1221, opened 2022-09-20
Follow symlinks when searching for globs in all-layers scope

kzantow:
This PR modifies the behavior when using `all-layers` scope such that symlinks from prior layers work properly.

There was an issue reported where a SUSE RPM database was changed but the change was not identified because the symlink did not change and the SUSE location did not match the recognized location for the RPM database. Now, regardless of the layer a symlink is introduced in, it should work as expected to match globs.

Precursor for https://github.com/anchore/grype/issues/894

### DIFF
--- a/syft/source/all_layers_resolver.go
+++ b/syft/source/all_layers_resolver.go
@@ -120,8 +120,6 @@ func (r *allLayersResolver) FilesByPath(paths ...string) ([]Location, error) {
 }
 
 // FilesByGlob returns all file.References that match the given path glob pattern from any layer in the image.
-//
-//nolint:gocognit
 func (r *allLayersResolver) FilesByGlob(patterns ...string) ([]Location, error) {
 	uniqueFileIDs := file.NewFileReferenceSet()
 	uniqueLocations := make([]Location, 0)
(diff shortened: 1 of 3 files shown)


syft/source/all_layers_resolver.go
```diff
@@ -120,18 +120,28 @@ func (r *allLayersResolver) FilesByPath(paths ...string) ([]Location, error) {
...
-			results, err := r.img.Layers[layerIdx].Tree.FilesByGlob(pattern, filetree.FollowBasenameLinks, filetree.DoNotFollowDeadBasenameLinks)
+			results, err := r.img.Layers[layerIdx].SquashedTree.FilesByGlob(pattern, filetree.FollowBasenameLinks, filetree.DoNotFollowDeadBasenameLinks)
 			if err != nil {
 				return nil, fmt.Errorf("failed to resolve files by glob (%s): %w", pattern, err)
 			}
 
 			for _, result := range results {
+				found, _, err := r.img.Layers[layerIdx].Tree.File(result.RealPath)
+				if err != nil {
+					return nil, fmt.Errorf("failed to resolve file during glob resolution %s: %w", result.RealPath, err)
```
wagoodman, line 139: question: should we warn, return an error, or continue without note here?

I'm curious about what the error conditions are in `File()` and what that might hint at.

kzantow: I'll dig in to this, but seeing as there's a `bool` return, I maybe incorrectly assumed a "not found" should just be that...

kzantow: It turns out one cannot look in the layer to determine if a file exists, it just needs to use the `SquashedTree`, so removed this bit anyway.

spiffcs (approved):

PR #1228, opened 2022-09-26
Support exclude patterns on Windows

### LINKED ISSUE #1024: Excluding a directory does not work on Windows
When I use syft with option packages, the --exclude argument seems to be ignored.
The output tells me that there is a problem to access a zip file in %APPDATA%/Local/Temp/2/syft-archive-contents-... (not a valid zip file) or (the process cannot access the file because it is being used by another process).
So it seems that syft is analyzing its own temporal archive.

[0012]  WARN cataloger 'java-cataloger' failed to parse entries at location=Location<id=3767 RealPath="C:\\Users\\asd\\AppData\\Local\\Temp\\2\\syft-archive-contents-2361904858\\archive-archive-archive-archive-archive-arch
ive-archive-archive-archive-archive-archive-python-3.8.zip">: unable to read files from java archive: unable to open zip archive (C:\Users\asd\AppData\Local\Temp\2\syft-archive-contents-2300921872\archive-archive-archive-a
rchive-archive-archive-archive-archive-archive-archive-archive-archive-python-3.8.zip): cannot find beginning of zip archive="C:\\Users\\asd\\AppData\\Local\\Temp\\2\\syft-archive-contents-2300921872\\archive-archive-archi
ve-archive-archive-archive-archive-archive-archive-archive-archive-archive-python-3.8.zip" : zip: not a valid zip file
[0012] ERROR unable to cleanup archive tempdir: remove C:\Users\asd\AppData\Local\Temp\2\syft-archive-contents-3815765297\archive-archive-archive-archive-archive-archive-archive-archive-archive-archive-archive-python-3.8.z
ip: The process cannot access the file because it is being used by another process.

Hence I tried to add the Temp directory to the exclusion list, but no success.

from powershell
cd /
.\syft\syft.exe packages --exclude '**/Temp' dir:c:\\

Also when I try '**/* everything will be checked, too.

BTW: When I try to specify an absolute directory, it tells me that I need to use one of ./, */ or **/. But I just want to use c:\ or /c/Users/*/AppData/Local/Temp or similar.

Second question:
It seems that the output does not contain any exe file or program-name of an installed application. Is this correct? It only finds some python files at c:\cygwin64 for example.

Thank you very much for your help.

**Environment**:

Application:        syft
Version:            0.46.3
JsonSchemaVersion:  3.2.3
BuildDate:          2022-05-26T06:46:34Z
GitCommit:          7cb8e1fc14a278ec5afce379623a47577aba9917
GitDescription:     v0.46.3
Platform:           windows/amd64
GoVersion:          go1.18.2
Compiler:           gc

Windows Server 2019 Datacenter Evaluation

kzantow:
This PR fixes `--exclude` patterns on Windows along with adding some additional testing. However, since we are not utilizing any Windows runners, the test will not run as part of CI currently, so there is an approximation of it that runs on Linux.

Fixes #1024

### DIFF
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -392,6 +392,9 @@ func getDirectoryExclusionFunctions(root string, exclusions []string) ([]pathFil
 		return nil, err
 	}
 
+	// this handles Windows file paths by converting them to C:/something/else format
+	root = filepath.ToSlash(root)
+
 	if !strings.HasSuffix(root, "/") {
 		root += "/"
 	}
@@ -414,6 +417,8 @@ func getDirectoryExclusionFunctions(root string, exclusions []string) ([]pathFil
 	return []pathFilterFn{
 		func(path string, _ os.FileInfo) bool {
 			for _, exclusion := range exclusions {
+				// this is required to handle Windows filepaths
+				path = filepath.ToSlash(path)
 				matches, err := doublestar.Match(exclusion, path)
 				if err != nil {
 					return false
--- a/syft/source/source_test.go
+++ b/syft/source/source_test.go
@@ -579,6 +579,75 @@ func TestImageExclusions(t *testing.T) {
 	}
 }
 
+func Test_crossPlatformExclusions(t *testing.T) {
+	testCases := []struct {
+		desc    string
+		root    string
+		path    string
+		exclude string
+		match   bool
+	}{
+		{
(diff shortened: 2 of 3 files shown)


spiffcs:
Tried this on my windows machine - LGTM!

spiffcs (approved):

Issue #1222, opened 2022-09-21
Support for Conan lockfiles


kzantow:
Hi syft community,

conanfile.txt is a primary/declarative use of conan to only install dependencies.
Most conan users use conanfile.py, the imperative way to define dependencies, build and packaging.
Do you plan to support conanfile.py? For example using lock files (https://docs.conan.io/en/latest/versioning/lockfiles.html)?

Regards.

_Originally posted by @EstebanDugueperoux2 in https://github.com/anchore/syft/issues/1083#issuecomment-1253742862_

kzantow:
Hi @EstebanDugueperoux2, I created this issue to from your comment on #1083. When looking at this feature, it states:
> This is an experimental feature subject to breaking changes in future releases.

Maybe it would be wise to wait until the feature has stabilized? I do not think we will prioritize this at least until the feature is stable, but do welcome PRs!

spiffcs:
@EstebanDugueperoux2 with `conan.lock` support committed I believe we are now cataloging a superset of the packages that would be located in `conan.py`. Do you still see any value in parsing `conan.py` and including that info?

tgerla:
We'll go ahead and close this ticket but please feel free to re-open if conan.py support is still important now that we have support for the lockfiles. Thanks!

PR #1083, opened 2022-07-01
feat: add support for conan packages (C/C++)

cpendery:
## 📝 Description
Adds support for indexing Conan `conanfile.txt` files

Closes https://github.com/anchore/syft/issues/1082

PR #1230, opened 2022-09-28
Add Conan (C/C++) conan.lock file support

hkwi:
extend #1083 and implement #1222


syft/pkg/cataloger/cpp/parse_conanlock.go
```diff
@@ -0,0 +1,52 @@
...
+// parseConanlock is a parser function for conan.lock contents, returning all packages discovered.
+func parseConanlock(_ string, reader io.Reader) ([]*pkg.Package, []artifact.Relationship, error) {
+	pkgs := []*pkg.Package{}
+	var graphLock struct {
+		GraphLock struct {
+			Nodes map[string]struct {
+				Ref string
+			}
+		} `json:"graph_lock"`
+	}
```
kzantow, line 25: It might be nice to define this struct at the top level of this file and also be explicit about each field using `json:""` tags.

kzantow (commented):
This is looking pretty good -- just one small suggestion

syft/pkg/cataloger/cpp/parse_conanlock.go
```diff
@@ -0,0 +1,54 @@
...
+				continue
+			}
+			pkgName, pkgVersion := splits[0], splits[1]
+			pkgs = append(pkgs, &pkg.Package{
+				Name:         pkgName,
+				Version:      pkgVersion,
+				Language:     pkg.CPP,
+				Type:         pkg.ConanPkg,
+				MetadataType: pkg.ConanaMetadataType,
+				Metadata: pkg.ConanMetadata{
```
wagoodman, line 50: capturing the name and version isn't really necessary. There is one or two other catalogers that do this for implementation-specific limitations, however, in this case it's redundant.

That being said, there is other information in a `conan.lock` file that could be useful to capture (such as `options`).  It's probably worth taking a look at what other information could be provided and adding that to the metadata.

If we don't want to add any more elements, my recommendation now would be to not have a metadata struct at all.

kzantow (approved):
Agree with [@wagoodman 's comment](https://github.com/anchore/syft/pull/1230#discussion_r982684026) about not needing redundant name and version information, but 👍 from me

syft/pkg/cataloger/cpp/parse_conanlock.go
spiffcs: @wagoodman I can add `options` `path` and `context` as metadata fields and remove `name` and `version` here

Then I'll update the `conanfile` to include `nil` for the metadata

spiffcs: @wagoodman it also looks like `name` and `version` are tied to the method for generating a PURL for a given `conan` package. It might be useful to keep those fields so we can agnostically generate the PURL for either a `lockfile` or `conanfile` even if the information is redundant for the package struct:

https://github.com/anchore/syft/blob/main/syft/pkg/conan_metadata.go

Is there another space you would want this method?

spiffcs:
Talked offline with @wagoodman. @kzantow I enhanced the metadata struct with more fields from the lockfile. We'll keep `name` and `version` for now to not disturb the PURL generation for cpp packages.

spiffcs:
One more small refactor incoming

syft/pkg/cataloger/cpp/parse_conanlock.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/common"
+)
+
+// integrity check
+var _ common.ParserFn = parseConanlock
+
+type conanLock struct {
+	GraphLock struct {
+		Nodes map[string]struct {
```
wagoodman, line 18: nit: this could be `map[string]pkg.ConanMetadata` instead of redefining these. That is, if the metadata is meant to raise up the raw packaging metadata, if we remove name/version then there would be no difference between this struct def and `pkg.ConanMetadata`.

wagoodman: I could be wrong about this if `options string` should really be transformed into `options []string` (split by newline) or `options map[string]string` if split by newline and `=`. (same comment for similar fields)

syft/pkg/conan_metadata.go
```diff
@@ -8,6 +8,9 @@ import (
 type ConanMetadata struct {
 	Name    string `mapstructure:"name" json:"name"`
 	Version string `mapstructure:"version" json:"version"`
+	Options string `mapstructure:"options" json:"options"`
```
wagoodman, line 11: + json schema regeneration due to change (might as well wait on this until we know what the final change is)

syft/pkg/cataloger/cpp/parse_conanlock.go
wagoodman: ahh, we're reusing the same metadata for logically a different things. I don't think we should do this. That is, the `Metadata` is reserved for things that are specific from what this package was parsed from (I didn't see the reuse on the first review pass). We should probably be introducing a new metadata type to capture this information.

syft/pkg/cataloger/cpp/parse_conanlock.go
spiffcs: For those following along here are all possible fields for the conan.lock structure
https://github.com/conan-io/conan/blob/develop/conans/model/graph_lock.py#L101-L276

spiffcs:
Quick status on this one

Currently separating out the metadata types per this comment:
https://github.com/anchore/syft/pull/1230#discussion_r982727569

Updating the CLI tests and adding more fields to the new metadata object per this comment: 
https://github.com/anchore/syft/pull/1230#discussion_r982779423

spiffcs:
Last change here will be to regenerate the schema, I'm still reviewing and trying to find examples of a conan lock file is all possible fields populated so we're sure the new metadata struct if constructed correctly

spiffcs:
CLI passed on my local @kzantow so this 🤞 should be ready

internal/constants.go
```diff
@@ -6,5 +6,5 @@ const (
 
 	// JSONSchemaVersion is the current schema version output by the JSON encoder
 	// This is roughly following the "SchemaVer" guidelines for versioning the JSON schema. Please see schema/json/README.md for details on how to increment.
-	JSONSchemaVersion = "3.3.2"
+	JSONSchemaVersion = "4.0.0"
```
kzantow, line 9: Is this actually a _breaking_ change? I thought we were just adding _more_ information, so this would just be `3.3.3`?

spiffcs: We also modified the Conan datatype to be Ref instead of name/version which is breaking

spiffcs: Screen Shot 2022-09-29 at 2 13 31 PM

syft/pkg/conan_lock_metadata_test.go
```diff
@@ -0,0 +1 @@
+package pkg
```
kzantow, line 1: Is this necessary?

spiffcs: It would be if the test had made it in there! Let me see where that got stashed apologies

spiffcs: Test added back Thanks for the review!

PR #1234, opened 2022-09-29
update requires to use list; remove field

spiffcs:
Incorrect data type being used for requires. Was string ==> now slice of string.

`BuildRequires` was also not a part of the conan.lock spec anymore so removing


wagoodman (approved):
The json schema will need to be bumped too. If there hasn't been a release yet then you can use the same version and you just need to regenerate.

PR #1243, opened 2022-10-04
feat: add identifiable field to source object

spiffcs:
Allow source.Source struct to set reproducible id for the following scheme:

- DirectoryScheme
- FileScheme
- ImageScheme
- UnknownScheme 

This ID is calculated in the following ways: 
- a digest from the given directory path
- the digest of the file contents or file path 
- the digest of the image manifest 
    - if this cannot be found it is calculated as a ChainID: https://github.com/opencontainers/image-spec/blob/main/config.md#layer-chainid
- If the scheme is unknown then the ID is calculated as a Hash of the source struct.

Followups coming in separate smaller PR:
- [ ] Add ID field to source and generate a relationship to packages
- [ ] Update SPDX format to reflect this relationship of source --> package via [OCI PURL](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci)

### DIFF
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/bmatcuk/doublestar/v4"
 	"github.com/mholt/archiver/v3"
+	digest "github.com/opencontainers/go-digest"
 	"github.com/spf13/afero"
 
 	"github.com/anchore/stereoscope"
@@ -25,6 +26,7 @@ import (
 // Source is an object that captures the data source to be cataloged, configuration, and a specific resolver used
 // in cataloging (based on the data source and configuration)
 type Source struct {
+	id                string
 	Image             *image.Image // the image object to be cataloged (image only)
 	Metadata          Metadata
 	directoryResolver *directoryResolver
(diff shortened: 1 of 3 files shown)


syft/source/source.go
```diff
@@ -304,6 +306,50 @@ func NewFromImage(img *image.Image, userImageStr string) (Source, error) {
...
+	s.id = calculateChainID(s.Image)
+
+	if s.id == "" {
+		// TODO what happens here if image has no layers?
+		s.id = digest.FromString(s.Metadata.ImageMetadata.UserInput).String()
+	}
+	return
+}
+
+func calculateChainID(img *image.Image) string {
```
lumjjb, line 331: hmm, trying to understand a bit more,.. have to preface this with i don't think i've fully grasped the proposal.. 
i think the hash of the container (well the one is used to download the container and identify it), isn't the layer chain ID? - but the sha256 of the OCI manifest

spiffcs: Yea good point - probably will use this as the fallback if syft can determine the image was not pulled and is locally built. Thanks for the comment!

syft/source/source.go
```diff
@@ -304,6 +307,73 @@ func NewFromImage(img *image.Image, userImageStr string) (Source, error) {
 	}, nil
 }
 
+func (s *Source) ID() artifact.ID {
+	return s.id
+}
+
+func (s *Source) SetID() {
```
kzantow, line 314: Should this PR include a call to this function?

kzantow (approved):
👍

syft/source/source.go
spiffcs: ^ Second PR is going to do that so we can link it into relationships - Just wanted to keep this PR small, but I can add the call so we're setting it on construction

spiffcs: Looking again, there also isn't ONE great place to put the single call since we have functions like `New`, `NewFromImage` `NewFromDirectory` `NewFromFile`.

Let me see all the places it needs to be included so we have full coverage of all the source creation

spiffcs: I could just have it get set on the first call to `ID` if s.id = ""

PR #1246, opened 2022-10-05
Fix #1179 Deprecated SPDX licenses

### LINKED ISSUE #1179: Deprecated SPDX license (GFDL* and BSD-2-Clause-NetBSD)
**What happened**:

This is a follow-up of #950.
The GPL family has been solved, but not all deprecated licenses. Please check the whole list of deprecated licenses at https://spdx.org/licenses/#Deprecated%20License%20Identifiers

"GFDL-1.1" should be replaced by "GFDL-1.1-only"
"GFDL-1.2" should be replaced by "GFDL-1.2-only"
"GFDL-1.3" should be replaced by "GFDL-1.3-only"
"GFDL-1.1+" should be replaced by "GFDL-1.1-or-later"
"GFDL-1.2+" should be replaced by "GFDL-1.2-or-later"
"GFDL-1.3+" should be replaced by "GFDL-1.3-or-later"
"BSD-2-Clause-NetBSD " should be replaced by "BSD-2-Clause"

```
syft docker:apache/airflow:2.3.0b1-python3.10 -o spdx-tag-value > airflow-syft-0.54.0.spdx
```

**Environment**:
- Output of `syft version`:

```
Application:        syft
Version:            0.54.0
JsonSchemaVersion:  3.3.2
BuildDate:          2022-08-17T18:56:00Z
GitCommit:          21eb77206020c04c9ca5316562e32179c73926a9
GitDescription:     v0.54.0
Platform:           linux/amd64
GoVersion:          go1.18.5
Compiler:           gc
```

```
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
UBUNTU_CODENAME=jammy
```

vargenau:
Fix all SPDX deprecated licenses including GFDL*, BSD-2-Clause-NetBSD listed on https://spdx.org/licenses/#Deprecated%20License%20Identifiers

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>


internal/spdxlicense/license_list.go
```diff
@@ -109,18 +109,18 @@ var licenseIDs = map[string]string{
 	"bsd-1.0-clause":                         "BSD-1-Clause",
 	"bsd-1.0.0-clause":                       "BSD-1-Clause",
 	"bsd-2-clause":                           "BSD-2-Clause",
-	"bsd-2-clause-freebsd":                   "BSD-2-Clause-FreeBSD",
-	"bsd-2-clause-netbsd":                    "BSD-2-Clause-NetBSD",
+	"bsd-2-clause-freebsd":                   "BSD-2-Clause-Views",
+	"bsd-2-clause-netbsd":                    "BSD-2-Clause",
```
kzantow, line 113: I don't think we should be changing the semantics of `freebsd` and `netbsd` in this way and below.

internal/spdxlicense/license_list.go
```diff
@@ -170,9 +170,9 @@ var licenseIDs = map[string]string{
 	"busl-1":                                   "BUSL-1.1",
 	"busl-1.1":                                 "BUSL-1.1",
 	"busl-1.1.0":                               "BUSL-1.1",
-	"bzip2-1":                                  "bzip2-1.0.5",
-	"bzip2-1.0":                                "bzip2-1.0.5",
-	"bzip2-1.0.5":                              "bzip2-1.0.5",
+	"bzip2-1":                                  "bzip2-1.0.6",
+	"bzip2-1.0":                                "bzip2-1.0.6",
+	"bzip2-1.0.5":                              "bzip2-1.0.6",
```
kzantow, line 175: I don't think we should be relabeling these licenses. If something is explicitly `bzip-1.0.5`, we probably shouldn't say it is `1.0.6`

internal/spdxlicense/license_list.go
```diff
@@ -472,49 +472,49 @@ var licenseIDs = map[string]string{
 	"fsfullr":                                  "FSFULLR",
 	"ftl":                                      "FTL",
 	"gd":                                       "GD",
-	"gfdl-1":                                   "GFDL-1.1",
+	"gfdl-1":                                   "GFDL-1.1-only",
```
kzantow, line 475: Does adding `-only` here change the semantics? This seems like a reasonable thing to assume, but it would be good to be sure. Same question for all the other `-only` additions above and below.

internal/spdxlicense/license_list.go
```diff
@@ -984,7 +984,7 @@ var licenseIDs = map[string]string{
 	"sspl-1":                                   "SSPL-1.0",
 	"sspl-1.0":                                 "SSPL-1.0",
 	"sspl-1.0.0":                               "SSPL-1.0",
-	"standardml-nj":                            "StandardML-NJ",
+	"standardml-nj":                            "SMLNJ",
```
kzantow, line 987: This is already covered in the `smlnj` case, isn't it?

kzantow (changes requested):
Hi @vargenau -- thank you for the contribution. I've left some specific questions, but what is the motivation for getting rid of the deprecated identifiers? These are still valid until they have been removed, right?

There is a problem, however: we generate the license list using `make generate-license-list` which pulls all the license identifiers from https://spdx.org/licenses/licenses.json and it will just overwrite the changes you've made here. If we were to make some changes, it would need to be done in the [generate_license_list.go](https://github.com/anchore/syft/blob/main/internal/spdxlicense/generate/generate_license_list.go#L105) or similar.

internal/spdxlicense/license_list.go
vargenau: Yes, we should.

This is explained at:
- https://spdx.org/licenses/BSD-2-Clause-FreeBSD.html
- https://spdx.org/licenses/BSD-2-Clause-NetBSD.html

Using BSD-2-Clause-FreeBSD or BSD-2-Clause-NetBSD is invalid SPDX and is reported as such by https://tools.spdx.org/app/validate/

internal/spdxlicense/license_list.go
vargenau: Yes we must.

See https://spdx.org/licenses/bzip2-1.0.5.html

internal/spdxlicense/license_list.go
```diff
@@ -637,11 +637,11 @@ var licenseIDs = map[string]string{
 	"lgpl-2.0.0-only":                          "LGPL-2.0-only",
 	"lgpl-2.0.0-or-later":                      "LGPL-2.0-or-later",
 	"lgpl-2.1":                                 "LGPL-2.1-only",
-	"lgpl-2.1+":                                "LGPL-2.1+",
+	"lgpl-2.1+":                                "LGPL-2.1-or-later",
```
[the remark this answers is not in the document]
vargenau: Yes "+" is still valid, but LGPL-2.1+ is explicitly deprecated and leads to invalid SPDX.

See: https://spdx.org/licenses/LGPL-2.1+.html

spiffcs:
This suggestion from Dan seems to point to the correct place where we could start identifying the map from deprecated licenses to their newer versions: https://github.com/anchore/syft/issues/950#issuecomment-1132874627

I think there is a way where we can update [generate_license_list.go](https://github.com/anchore/syft/blob/main/internal/spdxlicense/generate/generate_license_list.go?rgh-link-date=2022-10-05T15%3A59%3A45Z#L105) to:

1) see if a license is deprecated
2) if so check the fields between its deprecated object and the newer object it points to
3) if those fields match, update the map programmatically so that we can auto-generate this file rather than manually curate it

vargenau:
Thank you for your comments.

Who will implement this? I am not sure to understand enough [generate_license_list.go](https://github.com/anchore/syft/blob/main/internal/spdxlicense/generate/generate_license_list.go?rgh-link-date=2022-10-05T15%3A59%3A45Z#L105) to be able to do it myself, but I can help.

It is very important that we generate valid SPDX.

kzantow:
The licenses in question are _deprecated_, but they should still be valid SPDX licenses, shouldn't they?

vargenau:
> The licenses in question are _deprecated_, but they should still be valid SPDX licenses, shouldn't they?

From the standard:

> When a license identifier is "deprecated" on the SPDX License List, it effectively means that there is an updated license identifier and the deprecated license identifier, _while remaining valid_, **should no longer be used**.

The SPDX validator https://tools.spdx.org/app/validate/ gives a warning.

So technically, it is valid, but it should not be used, so it still think it would be beneficial to have it fixed.

vargenau:
Yes, the replacement of the deprecated license is _exactly_ the same as the deprecated license.

I had first created #950 that has been implemented in #1009.

This replaces e.g . GPL-2.0 by GPL-2.0-only and GPL-2.0+ by GPL-2.0-and-later.

Why not doing the same with GFDL family, LGPL family, etc?

We are now in the middle of the river. #1179 is about the complete the job with all deprecated licenses.

PR #1248, opened 2022-10-05
feat: add RelationshipsBySourceOwnership to syft json output

spiffcs:
## feat: add new optional source id to json schema

Update json schema to include the source id field.
This new field allows users to view relationships between a source and its packages

Currently src --> packages contain all cataloged packages.
This will become more nuanced as multiple source support is added.

Eventually, syft will be able to generate an SBOM with a single root node that contains multiple
sources, which then contain multiple packages, which then contain some
amount of files. 

These entities are also not strictly confined to the described tree structure.
They can behave more like a graph where other directional edges may be applied across entities.

### TODO:
- [x] Integration test updates encode/decode cycle


syft/formats/syftjson/to_syft_model.go
```diff
@@ -78,6 +81,14 @@ func toSyftRelationships(doc *model.Document, catalog *pkg.Catalog, relationship
 	return out
 }
 
+func toSyftSource(s model.Source) *source.Source {
+	newSrc := &source.Source{
+		Metadata: *toSyftSourceData(s),
+	}
+	newSrc.SetID()
```
spiffcs, line 88: This method will generate a consistent ID across encode/decode

syft/source/metadata.go
```diff
@@ -2,6 +2,7 @@ package source
 
 // Metadata represents any static source data that helps describe "what" was cataloged.
 type Metadata struct {
+	ID            string        // the id generated from the parent source struct
```
wagoodman, line 5: is this value part of the hash function? (since metadata is not ignored explicitly on the parent struct). If so, then I think this may need to be hash ignored

syft/source/source.go
```diff
@@ -353,27 +355,28 @@ func (s *Source) SetID() {
 	}
 
 	s.id = artifact.ID(strings.TrimPrefix(d, "sha256:"))
+	s.Metadata.ID = strings.TrimPrefix(d, "sha256:")
 }
 
-func calculateChainID(img *image.Image) string {
-	if len(img.Layers) < 1 {
+func calculateChainID(m ImageMetadata) string {
```
wagoodman, line 361: nit: it looks like this doens't need the full metadata, could we just pass the layers to keep the concerns smaller for this function?

syft/lib.go
```diff
@@ -74,9 +74,24 @@ func CatalogPackages(src *source.Source, cfg cataloger.Config) (*pkg.Catalog, []
...
 	return catalog, relationships, release, nil
 }
 
+func newSourceRelationshipsFromCatalog(src *source.Source, c *pkg.Catalog) []artifact.Relationship {
+	relationships := make([]artifact.Relationship, 0) // Should we pre-allocate this by giving catalog a Len() method?
+	for p := range c.Enumerate() {
+		relationships = append(relationships, artifact.Relationship{
+			From: src,
+			To:   p,
+			Type: artifact.ContainsRelationship,
```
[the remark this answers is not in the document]
wagoodman: I think the question is fair, though I don't think `Source[Relationship]` is a good edge name. That is, the edges should act like verbs or phrases to help qualify the nouns (nodes). In more detailed/nuanced edges the phrases might describe how a noun is impacted, but the nouns shouldn't be the primary edge word/phrase.

I think `Contains` is well suited, but I understand that semantically we'd have an overlap of usage elsewhere. Would the fact that the nodes are of different types be enough to distinguish these use cases here? or maybe we could capture the specific use cases within the relationship `.data` section to help distinguish these cases.

syft/formats/syftjson/to_syft_model.go
spiffcs: We need to call `SetID` since `id` is a private field. Even if I carried doc into this function signature `newSrc.id = doc.Source.ID`

It's important that our `SetID` call is consistent across serialization/deserialization. As long as the source metadata remains unchanged between the ser/deser calls then this method will consistently produce the same ID.

I could also update `SetID` for this case to take (id string) as an argument which would override the field rather than calculate the hash.

syft/lib.go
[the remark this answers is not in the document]
spiffcs: `BelongsTo`?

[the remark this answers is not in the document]
kzantow: I think `Contains` is fine. As @wagoodman mentioned, we can determine the type by casting the `From`/`To` types. I've had to do this for CycloneDX, anyway, as there are no relationship type qualifiers anyway.

syft/source/source.go
```diff
@@ -96,6 +96,7 @@ func NewFromRegistry(in Input, registryOptions *image.RegistryOptions, exclusion
 	source, cleanupFn, err := generateImageSource(in, registryOptions)
 	if source != nil {
 		source.Exclusions = exclusions
+		source.SetID()
```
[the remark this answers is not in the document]
spiffcs: There are some cases where `source.ID()` is not called before the document is serialized which can lead to a blank string.  Example: if you remove these calls you'll see documents in our integration tests where `source.ID: ""`

Both these cases have been added as part of the source constructor to make sure a constructed source that does not call `ID()` still has the field set.

syft/formats/syftjson/test-fixtures/snapshot/TestDirectoryEncoder.golden
```diff
@@ -67,6 +67,7 @@
  ],
  "artifactRelationships": [],
  "source": {
+  "id": "",
```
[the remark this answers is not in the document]
spiffcs: Resolved by making SetID more explicit when generating new sources

syft/formats/syftjson/test-fixtures/snapshot/TestEncodeFullJSONDocument.golden
```diff
@@ -139,6 +139,7 @@
   }
  ],
  "source": {
+  "id": "",
```
spiffcs, line 142: This is still blank while others are getting and ID

syft/formats/syftjson/test-fixtures/snapshot/TestImageEncoder.golden
```diff
@@ -64,10 +64,11 @@
  ],
  "artifactRelationships": [],
  "source": {
+  "id": "00afea0209d754683fdfcdd47cfea94ec9f2e81286be444e297a8c776c4accbf",
```
spiffcs, line 67: Also Blank

syft/formats/syftjson/to_format_model.go
```diff
@@ -17,8 +17,6 @@ import (
 )
 
 // ToFormatModel transforms the sbom import a format-specific model.
-// note: this is needed for anchore import functionality
-// TODO: unexport this when/if anchore import functionality is removed
```
spiffcs, line 21: This is still needed to be exported since the`template` package is using it in its `Encode` function

wagoodman (approved):

Issue #1162, opened 2022-08-16
Similar Packages Should Be Aggregated


cpendery:
**What happened**:
Producing an sbom with Syft is creating almost duplicate packages, where they are derived from different sources, but all of the information about the dependency is the same. We see this primarily using the `java-cataloger`

Ex:
1. A pom.xml and an archive file exist for the same package. Scanning these will yield duplicates for every dependency in this, even though it's the same package, because they have different sources (path, virtualPath).
2. Two archive files exist, each for different packages sharing a dependency. Scanning these will yield duplicates for the shared dependency as it will be produced for each, but again the only difference is the sources (path, virtualPath)

**What you expected to happen**:
I'd expect these two packages in the sbom to be aggregated into a single object where all the possible sources are included as an array, rather than having multiple of the same package where this is the only difference. 

**How to reproduce it (as minimally and precisely as possible)**:
```shell
docker pull elasticsearch:8.3.3 && syft docker:elasticsearch:8.3.3 -o cyclonedx-json | jq '.components[].purl' | grep 'pkg:maven/org.slf4j/slf4j-api@1.6.2'
```
This shows that the same purl is appearing multiple times, but looking at the full sbom output shows the separate items that are all the same except for some metadata attributes which could be combined

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: v0.53.2
- OS (e.g: `cat /etc/os-release` or similar): MacOS

spiffcs:
Wow great find thanks so much @cpendery!

wagoodman:
Changing `VirtualPath` on the java metadata from a string to a slice of string is very tempting, however, doing this would be inconsistent with how we merge packages today (which is by considering location). I was looking forward to the overall paradigm change in your PR #1249 , however, after more consideration I don't think we should make this change (see https://github.com/anchore/syft/pull/1249#issuecomment-1276144170 for more details).

Can I ask more about the use cases that motivated package merging? Maybe we can find a good workaround for your needs.

PR #1249, opened 2022-10-05
fix: duplicate packages when identical except location 

cpendery:
## 📝 Description
This pr fixes the issue of the same package being found in multiple locations producing multiple separate entries in the sbom.  

Closes https://github.com/anchore/syft/issues/1162


syft/pkg/package_test.go
```diff
@@ -99,16 +99,6 @@ func TestIDUniqueness(t *testing.T) {
 			},
 			expectedIDComparison: assert.NotEqual,
 		},
-		{
-			name: "location is reflected",
```
cpendery, line 103: not needed as the location shouldn't be reflected in the ID anymore

cpendery:
@spiffcs to polish the unit/integration tests snapshots. Please ping me if its a bigger issue than just re-snapping them

syft/pkg/java_metadata.go
```diff
@@ -20,7 +20,7 @@ var jenkinsPluginPomPropertiesGroupIDs = []string{
 
 // JavaMetadata encapsulates all Java ecosystem metadata for a package as well as an (optional) parent relationship.
 type JavaMetadata struct {
-	VirtualPath    string         `json:"virtualPath" cyclonedx:"virtualPath"` // we need to include the virtual path in cyclonedx documents to prevent deduplication of jars within jars
```
spiffcs, line 23: I don't think we can ignore this since it can help distinguish between two logically separate packages.

Example:
- We have a jenkins plugin that is tested where they are duplicates EXCEPT for the virtual path they are installed. For the first one `VirtualPath: /java/example-jenkins-plugin.hpi`. For the second one `VirtualPath: /java/example-jenkins-plugin.hpi:WEB-INF/lib/example-jenkins-plugin.jar` <-- This is what allows syft to detect separate packages nested within their parents with the same name.

I think ignoring the Locations list is ok as far as duplicates are concerned, but this field shows is similarly named packages are located within each other or at separate locations.

spiffcs (commented):
Thanks for the PR @cpendery and sorry I didn't catch my comment when we spoke earlier today. I ran into it when doing the integration testing updates. In your testing what situation did you see where this field being hashed was causing duplicates that were semantically the same?

spiffcs:
Here's the latest commit message I wrote, but I wanted to throw it out here on the PR as well since it feels like there is some contention here:

```
virtualPath needs to be considered as a part of ID generation for
packages that use the JavaMetadata struct. Consider a package that has a
virtualPath of "/java/example-jenkins-plugin.hpi" vs one that has the
virtualPath of "/java/example-jenkins-plugin.hpi:WEB-INF/lib/example-jenkins-plugin.jar".
Syft views these as separate packages and thus they should be cataloged
and sorted separately and in a stable fashion. 

Conversely, if an apkdb package is found to have exactly the same metadata, it is considered the
same package and the location lists are to be merged.
```

^ That's the current behavior, but it looks to me like we're treating the two ecosystems differently. 

In the example provided by #1162  `VirtualPath` is the only field that is different among the 4 packages listed as "slf4j-api". The sha1 digests are identical, with only the `metadata.VirtualPath` field showing any difference.

In the example given above for our `example-jenkins-plugin` scenario, the sha1 digests for the packages are different values, which demonstrates the value in separating them based on their virtual paths.

Given VirtualPath is location data, should we be merging these packages under one object and showing all locations it's installed, or should we keep duplicating this data and generating 4 separate objects in our different SBOM outputs?

spiffcs:
cc @kzantow and cc @wagoodman for other maintainer 👀 
cc @zhill

kzantow:
My 2c here: all other things the same, I would consider two packages with the same ecosystem, name, and version to be the same(-ish) and we should try to merge all locations where we find evidence of said package together. I think the rub here is for example we might find one with RPM metadata and another with Java metadata, and merging these would lose some information from one of the catalogers, even if realistically the packages might be referring to the same thing (e.g. apache tomcat found via rpmdb but also due to jars on the filesystem). It seems like to do this effectively we might want to account for merging things like this, too (which seems like potentially a bigger change).

spiffcs:
Thanks @kzantow for the feedback! I spoke with @wagoodman offline as well.

"It seems like to do this effectively we might want to account for merging things like this (cross ecosystem), too (which seems like potentially a bigger change)."

I think we can speak about solving this one as its own PR but really good to highlight here.

Specifically, to solve for #1162, we're going to ignore `VirtualPath`, but include `ArchiveDigests` in the generation of the ID.
This should reduce the output from 4 > 1 for `pkg:maven/org.slf4j/slf4j-api@1.6.2`.

We'll also update `VirtualPath` from `string` => `[]string` on the JavaMetadata so when we call `merge` we can account for all `VirtualLocations` for a single Java package, with the same digest, in the same way we treat locations for other packages.

spiffcs:
I've pushed two fixes here that update `JavaMetadata.VirtualPath` from `string` => `[]string`. The merge function for the catalog has been updated to reflect this change.

I tested this with the example provided in the issue. Instead of 4 separate packages `syft-json` is now a single package that looks like this:
```
  {
   "id": "6711fde2ffdba0f0",
   "name": "slf4j-api",  "version": "1.6.2",
   "type": "java-archive",
   "foundBy": "java-cataloger",
    ................
   "metadataType": "JavaMetadata",
   "metadata": {
    "virtualPath": [
     "/usr/share/elasticsearch/modules/vector-tile/slf4j-api-1.6.2.jar",
     "/usr/share/elasticsearch/modules/x-pack-security/slf4j-api-1.6.2.jar",
     "/usr/share/elasticsearch/modules/x-pack-identity-provider/slf4j-api-1.6.2.jar",
     "/usr/share/elasticsearch/modules/repository-azure/slf4j-api-1.6.2.jar"
    ],
    "manifest": {
     "main": {
      "Archiver-Version": "Plexus Archiver",
      "Build-Jdk": "1.6.0_16",
```

@wagoodman if this is what we want here then I'll update the schema and tests to cover this new behavior

syft/pkg/package.go
```diff
@@ -66,6 +67,15 @@ func (p *Package) merge(other Package) error {
 	if p.PURL == "" {
 		p.PURL = other.PURL
 	}
+
+	if p.MetadataType == JavaMetadataType {
+		virtualPaths := p.Metadata.(JavaMetadata).VirtualPath
+		virtualPaths = append(virtualPaths, other.Metadata.(JavaMetadata).VirtualPath...)
+		if t, ok := p.Metadata.(JavaMetadata); ok {
```
kzantow, line 74: If we're doing this check `if t, ok := p.Metadata.(JavaMetadata); ok {` anyway, why not do this first instead of a possible panic here `virtualPaths := p.Metadata.(JavaMetadata).VirtualPath` (albeit very unlikely). And we should probably test `other.Metadata.(JavaMetadata)`, too.

kzantow (commented):
Just a couple questions about hash ignores

wagoodman:
Unfortunately this issue/PR is in conflict with some upcoming work, specifically https://github.com/anchore/syft/issues/572 .

Today syft reports out a rather "flat" package list, with some package relationships to files to convey which packages own which files. The infrastructure for allowing catalogers to discover and report up package-to-package relationships is done and #572 is meant to utilize this to allow SBOMs generated by Syft to convey package-to-package dependency graphs.

Before picking up #572 I was very much in favor of deduplicating packages based on the metadata we find out about that package (and ignore where the package is from... in this way the "same bytes" found in multiple places would resolve to a single logical package). This method lends itself well when describing lists of unrelated packages, however, in the context of a graph, deduplicating (or merging) nodes means that the graph topology could be changing in unexpected ways.

Take, for instance, two different and unrelated python projects in the same container image installed in two different locations, I think it would be surprising if the dependency graph for both projects had edges to the same node. Logically speaking the projects may share the same dependency at the same version, but in reality they do so in separate contexts, so the graph should reflect this. This is further complicated if there are installation details that result in different child dependencies that are not easily visible in the packaging metadata (I think python package `extras` might be a good example of this -- the same package is installed with different options and could result in different dependencies needed... though I'd need to dig to come to a solid conclusion on this).

I'm going to close this PR in the meantime and follow up in the issue with other options. I'd be happy to talk about this further here, in the issue, or in the community meetings.

PR #1258, opened 2022-10-11
fix: reset merged package into map; 

spiffcs:
The current map access pattern for the catalog returned a package by copy and not by ref.

This PR updates the catalog map after merge so that merged values can be included back into the core catalog.

A test with CPE's has been added to display this behavior pre/post change.


kzantow (approved):
👍

PR #1259, opened 2022-10-12
Fix #1245 Update SPDX license list to 3.18

vargenau:
File licenses.json taken from https://github.com/spdx/license-list-data/releases/tag/v3.18

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>

Fixes: #1245

PR #1263, opened 2022-10-13
Fixes #1179 Deprecated SPDX license

### LINKED ISSUE #1179: Deprecated SPDX license (GFDL* and BSD-2-Clause-NetBSD)

vargenau:
Fixes #1179 Deprecated SPDX license (GFDL* and BSD-2-Clause-NetBSD)

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>


kzantow:
Hi @vargenau, I made a bit of a change to this PR and regenerated the license list -- could you have a look through the result and confirm it is what you are expecting?

Thanks!

internal/spdxlicense/generate/generate_license_list_test.go
```diff
@@ -22,18 +22,18 @@ func TestReplaceDeprecatedLicenses(t *testing.T) {
...
-		"abc-1.0.0":          "ABC-1.0-Only",
-		"abc-1.0-only":       "ABC-1.0-Only",
-		"abc-1.0.0-only":     "ABC-1.0-Only",
-		"abc-1+":             "ABC-1.0-Or-later",
-		"abc-1.0+":           "ABC-1.0-Or-later",
-		"abc-1.0.0+":         "ABC-1.0-Or-later",
-		"abc-1-or-later":     "ABC-1.0-Or-later",
-		"abc-1.0-or-later":   "ABC-1.0-Or-later",
-		"abc-1.0.0-or-later": "ABC-1.0-Or-later",
+		"abc-1":              "ABC-1.0-only",
```
spiffcs, line 25: Interesting that the casing has changed here, but just wanted to confirm for others who come across this that these are correct as of 3.18:

https://spdx.org/licenses/ <-- "or" and "only" are lower cased here

internal/spdxlicense/generate/license.go
```diff
@@ -43,7 +60,7 @@ func (l License) canReplace(other License) bool {
 		}
 	}
 
-	return l.ID != other.ID
+	return l.ID == other.ID
```
spiffcs, line 63: Why did we have to flip this here? I admit it's a little hard to reason about this case since it's the last one checked with so many other chances of an early return

spiffcs (commented):
Just quick clarification comment -thanks @vargenau for the PR and @kzantow for the follow up here!

internal/spdxlicense/generate/license.go
kzantow: Thanks for the look @spiffcs -- this function is `canReplace`, I think we haven't seen this bug because the other negative test conditions failed first. Basically: this function is trying to find any license that could be replaced by another license by checking equality in almost all the other fields. However, when it gets to the end here, the condition previously here is "if the ID is _NOT_ the same, we can consider it equal". Two comments about this: 1) this is not ever triggered based on the current document, and 2) I really don't think this is right -- if at the end of all the checks, the IDs are the same, it would return `false`... which means _maybe_ we have different `Text`...

internal/spdxlicense/generate/generate_license_list_test.go
kzantow: Yes, I believe  these should be lower cased. I don't see any license with upper case variants.

spiffcs (approved):

vargenau:
Thank you all for accepting and improving my patch.

I confirm it works as expected for me.

Can you also have a look at at pull request #1259 ? Thank you

vargenau:
Please check also pull request #1266 One more deprecated SPDX license: StandardML-NJ

PR #1265, opened 2022-10-14
Add go binary h1 digest to SPDX

### LINKED ISSUE #1261: Include go binary h1 digests in SPDX
Syft captures some metadata when cataloging go binaries, including the h1 digests. This information is currently output in Syft and CycloneDX (as a property) but not SPDX. There is a request to include the h1 digest information in SPDX format.

Based on the [go mod docs](https://go.dev/ref/mod#go-sum-files), it appears we can put this information in the [package checksum](https://spdx.github.io/spdx-spec/v2.3/package-information/#710-package-checksum-field) field after translation: `h1` -> `SHA256` and `hex.Encode(base64.Decode(value))`.

**Why is this needed**:
To align data between the formats and provide digest information for SPDX users.

**Additional context**:
This was originally reported by @deitch and has a little more context here: https://github.com/anchore/syft/pull/546#issuecomment-1276082318

kzantow:
This PR adds support for go binary h1 digest support being output to the SPDX checksum field.

Fixes #1261


syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -352,6 +356,25 @@ func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interf
 		return pkg.JavaMetadataType, pkg.JavaMetadata{
 			ArchiveDigests: digests,
 		}
+	case pkg.GoModulePkg:
+		var h1Digest string
```
kzantow, line 360: This should probably consolidated into a utility function somehow, but it's very specific to `h1` -> `sha256`

syft/formats/spdx22json/to_format_model.go
```diff
@@ -53,32 +55,15 @@ func toPackages(catalog *pkg.Catalog, relationships []artifact.Relationship) []m
 	for _, p := range catalog.Sorted() {
 		license := spdxhelpers.License(p)
 		packageSpdxID := model.ElementID(p.ID()).String()
-		filesAnalyzed := false
```
kzantow, line 56: NOTE: `FilesAnalyzed` is _NOT_ related to package checksums, it's related to [Package Verification Code](https://spdx.github.io/spdx-spec/v2.3/package-information/#79-package-verification-code-field)

spiffcs: I think `FilesAnalyzed` is generally applicable.
```
Indicates whether the file content of this package has been available for 
or subjected to analysis when creating the SPDX document.
If false, the package shall not contain any files. 
The metadata for the files analyzed field is shown in Table 20.
```

By generating a `checksum` we are confirming that the file content of this package was available/subjected to analysis.

The Cardinality for `Package Verification Code` is
```
0..1 if FilesAnalyzed ([7.8](https://spdx.github.io/spdx-spec/v2.3/package-information/#7.8)) is true or omitted.
```

So `PVC` can be `0 or 1` if `FilesAnalyzed` is true.

`FileAnalyzed` will be true, IF `PVC` is set, but `FilesAnalyzed` being true, has direct dependency on `PVC` being present

syft/formats/spdx22json/to_format_model.go
```diff
@@ -53,32 +54,15 @@ func toPackages(catalog *pkg.Catalog, relationships []artifact.Relationship) []m
...
 		// note: the license concluded and declared should be the same since we are collecting license information
 		// from the project data itself (the installed package files).
 		packages = append(packages, model.Package{
-			Checksums:        checksums,
+			Checksums:        toPackageChecksums(p),
 			Description:      spdxhelpers.Description(p),
 			DownloadLocation: spdxhelpers.DownloadLocation(p),
 			ExternalRefs:     spdxhelpers.ExternalRefs(p),
-			FilesAnalyzed:    filesAnalyzed,
+			FilesAnalyzed:    false, // this directly relates to packageVerificationCode
```
spiffcs, line 65: See the earlier comment. I believe this still holds that it can be true given that the file content of this package has been available for or subjected to analysis when creating the SPDX document.

spiffcs (commented):
Just a small comment about `FilesAnalyzed`

syft/formats/spdx22json/to_format_model.go
kzantow: Thanks for the review @spiffcs -- the problem with `FilesAnalyzed`, which we were only setting `true` when providing Java checksums [here](https://github.com/anchore/syft/pull/1265/files#diff-e0a0d8eccb8d2ea7bdbff4619151963a07c5ed32de52a12407219def486c039dL65) is that according to the SPDX spec, [we now need to _also_ provide a _Package Verification Code_](https://spdx.github.io/spdx-spec/v2.3/package-information/#79-package-verification-code-field):
> Required | Yes
> 0..1 if FilesAnalyzed ([7.8](https://spdx.github.io/spdx-spec/v2.3/package-information/#7.8)) is true or omitted, 0..0 (must be omitted) if FilesAnalyzed is false.

I think I've misinterpreted this and there's a bug in the documentation that `PackageVerificationCode` is actually _not_ required.

But we should actually be returning `true` for everything -- is that your opinion?

kzantow: Offline conversation: FilesAnalyzed should actually be true in the one case it was, corrected!

spiffcs (approved):

deitch:
Thanks @spiffcs .

How do I find that output? I cloned the repo, ran `make bootstrap` and then ran the binary against a go project directory, using both `spdx-json` and `spdx` output. Where should I find the hashes?

kzantow:
@deitch -- you can find the hashes in the `checksums` field, e.g.:
```json
      "checksums": [
        {
          "algorithm": "SHA256",
          "checksumValue": "33d5c1b2232c970da56f66738250b44ce9013cae4d422d3f9e852b767a05c148"
        }
```
These have been converted from the `h1:<base64>` versions.

deitch:
Huh, you are right, it is there. I guess I didn't find it on the larger one I ran. I had better figure out why. Likely my error, but if not, I will open an issue.

deitch:
It looks like it is finding some but not all. Is that possible?

And I cloned latest (checked that it has this commit) ran `go run cmd/syft/main.go -o spdx-json test > /tmp/test.json` but only has checksums for 2 of the packages. Allin [this gist](https://gist.github.com/deitch/f179f525659bb44bb34639b36cb15355)

Am I doing something wrong?

deitch:
It didn't even occur to me that this was for compiles binaries. How funny!

So I should run it against an actual binary? 

Also, curiously, what happens if I run it against an OCI image? Does it find the go binaries in there and scan them too? Or do I need to be explicit?

kzantow:
@deitch it does find go binaries when you do container scans, you shouldn't need to do anything special to enable this 👍 -- it should be noted, only binaries compiled using go mod will have this information, though.

If you `go build` that project, and scan the resulting `test` executable, you'll see checksum entries for the 2 dependencies.

deitch:
> If you go build that project, and scan the resulting test executable, you'll see checksum entries for the 2 dependencies

Actually, I see 2 `checksums` when scanning both source and binary, so that is equivalent already.

What I don't see is all of the dependencies. Take a look at the gist. 

* `go.mod`: 3 dependencies:
   * github.com/spf13/cobra (direct)
   * github.com/inconshreveable/mousetrap (indirect)
   * github.com/spf13/pflag (indirect)
* `go.sum`: 7 dependencies
* `spdx.json`: checksums for 2 of the (3 or 7) dependencies: cobra and pflag

kzantow:
@deitch this is a little nuanced -- the information included in the go binary is for _everything included in the binary_. What that means is: depending on the actual code paths the binary includes, it may or may not include all the dependencies in the `go.mod/sum`. Given how simple the test example is, I assume it actually isn't including a code path that causes `mousetrap` or the other dependencies to get included, and thus only `pflag` and `cobra` are included in the binary metadata. This is good -- it is a better indicator of what the binary actually contains!

As for scanning the _source_, this is a different question! I don't see any digests included in the source scan (did you scan the directory with the binary still present?). This is something we should probably add, but it explains the discrepancy between only seeing 2 dependencies with the binary and 3 in the `go.mod`. It looks like currently, `go.sum` is not being used in the cataloging process for go source (but it _could_ be at least to get the h1digests).

deitch:
> did you scan the directory with the binary still present?

Huh, maybe I did. I just reran it and am getting no checksums. Go figure. PEBKAC.

> but it explains the discrepancy between only seeing 2 dependencies with the binary and 3 in the go.mod.

meaning that the actual binary only uses 2 even if go.mod has 3 and go.sum has 7?

> I assume it actually isn't including a code path that causes mousetrap or the other dependencies to get included, and thus only pflag and cobra are included in the binary metadata. This is good -- it is a better indicator of what the binary actually contains!

That is somewhat surprising. Wouldn't you expect the `go mod` tools to be more efficient about it? Or are you saying they are moderately efficient, while the final `go build` is more efficient?

deitch:
btw, I just ran it on a complex binary, the results showed ` ✔ Cataloged packages      [106 packages]`, and the output does indeed show 106 `pkg:golang`, but not a single one has checksums.

deitch:
How interesting. I ran it for my other binary, it shows go 1.18.6 and all of the `dep`, all of them with the semver or other module version from go.mod, but not a single one with the `h1` hash. 🤷‍♂️

PR #1269, opened 2022-10-17
Add `--name` option to override name in output

jedevc:
A small pain point for me when generating SBOMs is that I want to have fine-grained control over the output, and specifically attempt to cleanup the recorded source from which the SBOM was produced. For example, I might mount a container filesystem to `/mnt` and then scan it - or I might unpack it into the current directory and perform a scan - or I might scan it as part of a build step (and then push an image later to a different production location). For my use case, I'd like that to remain an implementation detail, and not be exposed to the end user - a detail as to where the original path was on my machine isn't relevant to my user, so ideally I'd like to be able to hide those details.

As a (very) naive attempt, I've added a new `--name` flag to the CLI, which is propagated through to the formatters, to then control the various "name" fields, which would otherwise be automatically derived from the path name, or container path passed.

E.g. in SPDX-JSON:

```json
{
 "SPDXID": "SPDXRef-DOCUMENT",
 "name": "<NAME GOES HERE>",
 "spdxVersion": "SPDX-2.2",
 ...
```

E.g. in CycloneDX-JSON:

```json
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.4",
  "serialNumber": "urn:uuid:d34c78cc-412b-4887-9db0-21450a667f5a",
  "version": 1,
  "metadata": {
    "timestamp": "2022-10-17T13:54:03+01:00",
    "tools": [
      {
        "vendor": "anchore",
        "name": "syft",
        "version": "[not provided]"
      }
    ],
    "component": {
      "bom-ref": "371840ed01a6c73f",
      "type": "file",
      "name": "<NAME GOES HERE>"
    }
  },
  ...
```

I'm opening this more as a proposal, with an attempt to find a solution :smile: -- I'm not particularly tied to the approach in here (it's quite messy), but I thought I'd show what I quickly hacked together, even though it's likely the actual solution would probably look quite different :eyes:. I can always modify the SBOM for now, but I'd love to have something in Syft

### DIFF
--- a/cmd/syft/cli/attest/attest.go
+++ b/cmd/syft/cli/attest/attest.go
@@ -124,7 +124,7 @@ func parseAttestationOutput(outputs []string) (format string) {
 }
 
 func parseImageSource(userInput string, app *config.Application) (s *source.Input, err error) {
-	si, err := source.ParseInput(userInput, app.Platform, false)
+	si, err := source.ParseInputWithName(userInput, app.Platform, false, app.Name)
 	if err != nil {
 		return nil, fmt.Errorf("could not generate source input for attest command: %w", err)
 	}
--- a/cmd/syft/cli/options/packages.go
+++ b/cmd/syft/cli/options/packages.go
@@ -21,6 +21,7 @@ type PackagesOptions struct {
 	Platform           string
 	Exclude            []string
 	Catalogers         []string
+	Name               string
 }
 
 var _ Interface = (*PackagesOptions)(nil)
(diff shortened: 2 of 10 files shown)


spiffcs:
👋 @jedevc thanks for the PR! I saw this is still in draft and has some unit tests that need to be updated. Do you want help at all with that? Should I wait before giving a review?

As to the proposal I don't see a problem with adding a flag that lets you override this field rather than let syft choose out of the box for the user. 

As for the attempt - I think it's awesome because it gives us some clear vision on what can be centralized or refactored a bit to clean up the execution path of syft which would ultimately make a change like this easier.

Let me ask some of our team and see what they think and we can move this forward!

jedevc:
Woops, meant to get back to this a bit sooner :tada:

> As to the proposal I don't see a problem with adding a flag that lets you override this field rather than let syft choose out of the box for the user.

Aha, this is the reason it's in draft :smile: I wasn't sure whether this was the right approach, so if that general approach sounds roughly correct I'll update the unit tests and bring it out of draft so it's ready for review.

kzantow (approved):
LGTM -- thanks for the contribution @jedevc!

Issue #1268, opened 2022-10-17
syft attest --output cyclonedx-json incompatible with cosign


lehors:
**What happened**:
Produced an attestation with syft but cosign verify-attestation couldn't find it.

**What you expected to happen**:

cosign verify-attestation should successfully verify the attestation produced by syft

**How to reproduce it (as minimally and precisely as possible)**:

```
$ syft attest --output cyclonedx-json localhost:5000/hello:latest

$ COSIGN_EXPERIMENTAL=1 cosign verify-attestation --allow-insecure-registry --type cyclonedx localhost:5000/hello:latest
Error: none of the attestations matched the predicate type: cyclonedx
main.go:62: error during command execution: none of the attestations matched the predicate type: cyclonedx

```

**Anything else we need to know?**:

The problem comes from syft using ["https://cyclonedx.org/bom"](https://github.com/anchore/syft/blob/main/cmd/syft/cli/attest/attest.go#L361) as the predicate while cosign uses ["https://cyclonedx.org/schema" defined in-toto-golang](https://github.com/in-toto/in-toto-golang/blob/master/in_toto/model.go#L85)

**Environment**:
- Output of `syft version`:
Application:        syft
Version:            0.58.0
JsonSchemaVersion:  4.0.0
BuildDate:          2022-09-29T19:59:22Z
GitCommit:          b9b13d5525df89194d332467f692bc28bc68d07f
GitDescription:     v0.58.0
Platform:           darwin/amd64
GoVersion:          go1.18.6
Compiler:           gc
- OS (e.g: `cat /etc/os-release` or similar):
MacOs

lehors:
See related comment: https://github.com/in-toto/attestation/issues/82#issuecomment-1280759933

lehors:
It turns out that syft is using the predicate the CycloneDX project has selected (see https://github.com/CycloneDX/specification/issues/132) so the issue ends up being more on the cosign side. I submitted a PR to update in-toto after which the incompatibility will disappear.

It would still be good for syft to pull the definition from the in-toto project directly rather than have its own definition so I suggest keeping this open until that change is made with PR #1270

lehors:
Fixed with PR #1270

spiffcs:
Thank you @lehors for all the help on this one!

PR #1270, opened 2022-10-17
Use in-toto CycloneDX predicate to be compatible with cosign

lehors:
This addresses Issue #1268

As noted in https://github.com/in-toto/attestation/issues/82#issuecomment-1280759933 it is not clear what the predicate value should really be but this change leaves that question to in-toto and ensures alignment with other tools such as cosign.

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

### DIFF
--- a/cmd/syft/cli/attest/attest.go
+++ b/cmd/syft/cli/attest/attest.go
@@ -359,8 +359,7 @@ func formatPredicateType(format sbom.Format) string {
 	case spdx22json.ID:
 		return in_toto.PredicateSPDX
 	case cyclonedxjson.ID:
-		// Tentative see https://github.com/in-toto/attestation/issues/82
-		return "https://cyclonedx.org/bom"
+		return in_toto.PredicateCycloneDX
 	case syftjson.ID:
 		return "https://syft.dev/bom"
 	default:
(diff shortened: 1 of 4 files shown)


lehors:
I'm told the CycloneDX project has discussed the matter and decided to use the same value as what syft is currently using. So, I'm changing this to a draft PR. We should wait for the in-toto definition to be updated before applying this PR.

kzantow:
Thanks for the follow-up @lehors!

lehors:
The decision by the CycloneDX project is now official:
https://cyclonedx.org/specification/overview/#recognized-predicate-type

And in-toto-goland has been updated accordingly:
https://github.com/in-toto/in-toto-golang/pull/188

I think it makes sense to merge this PR. It will make no difference in how syft works but it makes the code cleaner.

spiffcs:
Yea Happy to merge this so we're using the value from the in-toto library. Really appreciate the follow up on this @lehors!

spiffcs:
I think we just need to update this PR with the latest in-toto-golang

lehors:
> I think we just need to update this PR with the latest in-toto-golang

Indeed, I had just thought about that: I also need to update go.mod to get the newer version. I don't know what your policy is with regard to dependencies. Would you rather wait for the next official release of in-toto-golang so we can simply use the new tag? I think that's fine, there is no hurry.

spiffcs:
Just pushed a commit that has the latest from `main` @anchore/tools let me know if you think we should just wait for the next release

lehors:
Looks good. I think you can go ahead and merge this now. Thanks!

PR #1276, opened 2022-10-20
Add Support for Native Image Executables

wdblair:
GraalVM allows developers to compile Java programs into Native Image executables. Recently, developers have been able to embed a software bill of materials (SBOM) into a Native Image executable in order to document the dependencies utilized by the Native Image. In this PR, we extend syft to extract these packages and include them in SBOMs for any container image or archive that includes a Native Image executable.


syft/pkg/cataloger/java/native_image_cataloger.go
```diff
@@ -0,0 +1,309 @@
+/*
+ * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
```
kzantow, line 2: I don't think we'd be able to accept a PR with a conflicting copyright

syft/pkg/type.go
```diff
@@ -17,6 +17,7 @@ const (
 	PythonPkg        Type = "python"
 	PhpComposerPkg   Type = "php-composer"
 	JavaPkg          Type = "java-archive"
+	NativeImagePkg   Type = "native-image"
```
kzantow, line 20: Would a better name for this be `java-native-image`?

syft/pkg/cataloger/java/native_image_cataloger.go
```diff
@@ -0,0 +1,309 @@
...
+
+type NativeImageElf struct {
+	File *elf.File
+}
+
+type NativeImageMachO struct {
+	File *macho.File
+}
+
+type Cataloger struct{}
```
kzantow, line 58: Maybe `NativeImageCataloger`?

syft/pkg/language.go
```diff
@@ -13,6 +13,7 @@ const (
 	// the full set of supported programming languages
 	UnknownLanguage Language = ""
 	Java            Language = "java"
+	JvmBytecode     Language = "jvm-bytecode"
```
kzantow, line 16: Is this decidedly a separate language? Would we typically be getting Java dependencies in the embedded information? I have a suspicion that this could result in missed java vulnerabilities if it's not the `java` language.

kzantow (changes requested):
Thanks for the contribution! This sounds like a really nice addition.

We'd like to see some sort of testing added for this -- would you be able to add some?

wdblair:
Thanks for the quick review! I would be happy to add tests to the PR. Is it alright if I store small native-image executables for the tests in `syft/pkg/cataloger/java/test-fixtures/native-image`?

wdblair:
> (Just to reiterate a previous comment, it might be better to name this java-native-image or jvm-native-image)

I agree, referencing Java or the JVM in the name will make the source of packages and the cataloger's use case clearer.

> We to avoid putting binaries in the codebase. For Java artifacts we run some builds -- could you make a simple build to generate a binary?

That would be much cleaner than maintaining binaries in the codebase. I'll try adding a simple build to generate the test binary with Maven.

syft/pkg/cataloger/java/native_image_cataloger.go
wdblair: We want this change to be in line with the project's license. Is it alright if we keep the following copyright header in this file to acknowledge the change comes from Oracle?

Copyright © 2022 Oracle and/or its affiliates.

syft/pkg/language.go
wdblair: We would typically see Java dependencies in the SBOM, so we can change the language to `java` in order to avoid missing relevant vulnerabilities.

syft/pkg/cataloger/java/native_image_cataloger.go
kzantow: Thanks @wdblair, I've asked for some more clarity on this and will get back to you with a definitive answer.

But, I'm curious if there might be some alternate solution where you / Oracle can get credit for this contribution without starting to add copyright headers in the codebase (as you can see, we are not doing this currently). What if we added an `AUTHORS.txt` or `CONTRIBUTORS.txt`, which we could be explicit about people / organizations who have contributed to this codebase? Would that be an acceptable method to retain credit for contributing things?

wdblair: Thanks for following up on this, I see that the copyright headers would deviate from your project's style conventions. I'll check whether a designated file such as `AUTHORS.txt` or `CONTRIBUTORS.txt` file will be alright and update you here. Would it be alright if we include the copyright statement only within this designated file?

wdblair: Hi @kzantow, we have updated this PR to include a `CONTRIBUTORS.txt` file to document the change, and have removed the copyright header from the source file. Would this be alright, or should we change the format of the contributors file or place it somewhere else in the repo?

wdblair:
@kzantow The latest version of this PR now includes tests to confirm that `syft` can parse Native Image executables, and that the Native Image cataloger can recover packages from a sample Native Image SBOM. I also updated `syft/pkg/type.go` to avoid the conflict raised by GitHub, but let me know if there are any other changes I can make.

syft/pkg/cataloger/java/native_image_cataloger.go
kzantow: Thank you @wdblair -- I asked about this `CONTRIBUTORS.txt`, and I think this is a lot closer to what we could take. The response I have for this is:

> We are okay with a notice of attribution along the following lines:
>
> _GraalVM Native Image - A cataloger contributed by Oracle, Inc. that extracts packages given within GraalVM Native Image SBOMs._
>
> But not okay with including in the attribution notice any copyright notice of Oracle or any other such external entity, because we have no way of knowing, other than taking the word of the contributing entity, whether and what extent, that entity is actually owner of copyright in the contribution.

Sorry this is challenging to get to the finish line -- it's the first time we've had a request like this in our open source projects.

wdblair: No worries, thanks for sharing the proposed change and for helping improve the PR. I will check to see if an attribution in this form is alright and will update you here.

wdblair: Hi @kzantow, this attribution works well for us. I have updated the `CONTRIBUTORS.txt` to follow this format.

syft/pkg/cataloger/java/graalvm_native_image_cataloger_test.go
```diff
@@ -0,0 +1,188 @@
...
+			if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+				if status.ExitStatus() != 0 {
+					t.Fatalf("failed to generate fixture: rc=%d", status.ExitStatus())
+				}
+			}
+		} else {
+			t.Fatalf("unable to get generate fixture result: %+v", err)
+		}
+	}
+}
```
kzantow, line 83: This is typically handled by the Makefile: https://github.com/anchore/syft/blob/main/Makefile#L249 (which also runs on CI), so you don't need to shell out here, just make sure it runs as part of the Java builds (https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/test-fixtures/java-builds/Makefile#L16-L29) and assume it has been run here. There also is a cache of these builds on CI: https://github.com/anchore/syft/blob/main/.github/workflows/validations.yaml#L91-L96 which is why there's the whole fingerprint deal.

kzantow (changes requested):
This looks good -- almost to the finish line, only thing is to probably remove `generateNativeImageBuildFixture` and add it to the Makefile(s) mentioned

syft/pkg/cataloger/java/graalvm_native_image_cataloger_test.go
wdblair: Thanks for the clarification, the layout of the test fixtures is more clear now. I have removed `generateNativeImageBuildFixture` and incorporated the build script into the Java builds to simplify building the Native Image test program. Do you see any other issues with the test fixtures?

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,316 @@
...
+// Catalog attempts to find any native image executables reachable from a resolver.
+func (c *NativeImageCataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, []artifact.Relationship, error) {
+	var pkgs []pkg.Package
+	fileMatches, err := resolver.FilesByMIMEType(internal.ExecutableMIMETypeSet.List()...)
+	if err != nil {
+		return pkgs, nil, fmt.Errorf("failed to find binaries by mime types: %w", err)
+	}
+
+	for _, location := range fileMatches {
+		readerCloser, err := resolver.FileContentsByLocation(location)
```
kzantow, line 301: The only other concern I have with this change is that it looks like every executable file is going to be read, contents loaded and scanned. I wonder how much this could slow down scans for large images.

Is there other discriminator we could use to identify "this is a graalvm  file" before reading all the contents?

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,316 @@
...
+		if err != nil {
+			log.Infof("native-image cataloger: opening file: %v", err)
```
kzantow, line 303: nit: should this be "error opening file"? and either `log.Warnf` (would be displayed) or `log.Debugf` (would only be displayed with `-vv` flag)?

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,316 @@
...
+			continue
+		}
+		log.Infof("Found an executable file %v", location)
```
kzantow, line 306: This should probably be `log.Trace`.

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,316 @@
...
+		p := getPackage(component)
+		pkgs = append(pkgs, p)
+	}
+	return pkgs, nil
+}
+
+// fileError logs an error message when an executable cannot be read.
+func fileError(filename string, err error) (NativeImage, error) {
+	// We could not read the file as a binary for the desired platform, but it may still be a native-image executable.
+	log.Infof("native-image cataloger: unable to read executable (file=%q): %v", filename, err)
```
kzantow, line 151: probably `log.Debugf`, since we're scanning all executable files and expect some of them not to be ELF or MachO (e.g. executable scripts, etc.)

kzantow (commented):
Just a few more small comments

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,316 @@
...
+		case "_" + nativeImageSbomSymbol:
+			sbom = s
+		case "_" + nativeImageSbomLengthSymbol:
+			sbomLength = s
+		case "_" + nativeImageSbomVersionSymbol:
+			svmVersion = s
+		}
+	}
+	if sbom.Value == 0 || sbomLength.Value == 0 || svmVersion.Value == 0 {
+		log.Infof("native-image cataloger: %v.", nativeImageMissingSymbolsError)
```
wagoodman, line 244:
```suggestion
		log.Warnf("native-image cataloger: %v.", nativeImageMissingSymbolsError)
```

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// NOTE: multiple readers are returned to cover universal binaries, which are files
+	// with more than one binary
+	readers, err := unionreader.GetReaders(reader)
+	if err != nil {
+		log.Warnf("native-image cataloger: failed to open a binary: %v", err)
+		return nil
+	}
+	for _, r := range readers {
+		for _, make := range imageformats {
```
wagoodman, line 276: since `make` overrides a built-in function it's a bit awkward to read, can you change the name of this variable?

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,316 @@
...
+		case nativeImageSbomSymbol:
+			sbom = s
+		case nativeImageSbomLengthSymbol:
+			sbomLength = s
+		case nativeImageSbomVersionSymbol:
+			svmVersion = s
+		}
+	}
+	if sbom.Value == 0 || sbomLength.Value == 0 || svmVersion.Value == 0 {
+		log.Infof("native-image cataloger: %v", nativeImageMissingSymbolsError)
```
wagoodman, line 206:
```suggestion
		log.Warnf("native-image cataloger: %v", nativeImageMissingSymbolsError)
```

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
wagoodman: agreed with this, there are a lot of Info log lines for handling error cases. Warn is great in cases where there is something wrong that warrants the user to take a look and report an issue in all of cases. Debug is better when the error may occur occasionally or doesn't warrant the user to take action.

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
wagoodman: I can see the concern here -- the short answer is that I'm not certain if this will be impactful "enough" to take action or not. I don't know of an easier way to do the upfront "is this a graalvm file" check than what is done here since a graalvm file produces native binaries.

wagoodman: I did an initial test with a few images and so far the difference run-time-wise is pretty nominal. Right now I don't have a large concern with this (could change in the future though).

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
wdblair: This makes more sense, we can rely on Debug when we want to log errors that do not need to reach users, but may be helpful for development/understanding the cataloger's logic.

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
wdblair: Of course, maybe `makeNativeImage` would be clearer.

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
wdblair: It looks like Go's packages for consuming binary files (i.e., ELF, Mach-O, and PE) use a `ReaderAt` to read data from the file. I think this implies that the Native Image cataloger will only read parts of the file relevant for detecting a valid executable (i.e. a correct header) and GraalVM Native Image symbols. That is, if the GraalVM version symbol is absent, the cataloger will stop reading the file.

Below, I `strace` `syft` while fetching the GraalVM SBOM from `ls` and a Native Image application. The cataloger appears to read the first part of `ls`, and then enumerates the symbol table. The cataloger reads less than the full size of the `ls` binary (which is 140kb) before concluding that there is no SBOM. As expected, the cataloger reads more data when reading the SBOM from the Native Image.

Based on this small experiment, it looks like you could expect our cataloger to make <50 small reads on each valid executable binary file in the container image. Of course, we can try to implement a more optimal check if this overhead is too high.

ls:

    openat(AT_FDCWD, "/path/to/test-container/ls", O_RDONLY|O_CLOEXEC) = 6
    pread64(6, "\177ELF", 4, 0)             = 4
    pread64(6, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0", 16, 0) = 16
    pread64(6, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0", 16, 0) = 16
    pread64(6, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\320g\0\0\0\0\0\0"..., 64, 0) = 64 x 43
    pread64(6, "\0.shstrtab\0.interp\0.note.gnu.pro"..., 285, 139932) = 285
    pread64(6, "\260h\0\0\0\0\0\0ph\0\0\0\0\0\0@s\0\0\0\0\0\0\240x\0\0\0\0\0\0"..., 4696, 135184) = 4696
    close(6)                                = 0
   
Native Image:

    openat(AT_FDCWD, "/path/to/test-container/native-application", O_RDONLY|O_CLOEXEC) = 6
    pread64(6, "\177ELF", 4, 0)             = 4
    pread64(6, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0", 16, 0) = 16
    pread64(6, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0", 16, 0) = 16
    pread64(6, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\200\2W\0\0\0\0\0"..., 64, 0) = 64 x 39
    pread64(6, "\0.symtab\0.strtab\0.shstrtab\0.inte"..., 277, 138508034) = 277
    pread64(6, "`\3W\0\0\0\0\0\220\0W\0\0\0\0\0 \3W\0\0\0\0\0\360\323\31\5\0\0\0\0"..., 29612, 138474344) = 29612
    close(6)

wdblair:
In the latest version of this PR, we added support for extracting SBOMs from GraalVM Native Images represented as Windows Portable Executable (PE) files. To limit the amount of read system calls performed on other executables in a container image, we first confirm that the SBOM symbols are present by enumerating the binary's exported symbols. When the SBOM symbols are present, we only read from the `.data` section to retrieve the SBOM. Do you think this approach is reasonable with respect to performance?

wdblair:
In addition to extracting SBOMs from Windows Native Images, we have also implemented the changes suggested in previous reviews. Are there any other improvements we can make to this PR?

kzantow:
Apologies @wdblair, there have been some holidays recently -- I'll get to this first thing tomorrow!

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,538 @@
...
+		log.Debugf("native-image cataloger: cannot obtain buffer from data segment.")
+		return nil, nil
+	}
+	sbomLocation := sbom.Value - dataSegment.Addr
+	lengthLocation := sbomLength.Value - dataSegment.Addr
+
+	return decompressSbom(databuf, sbomLocation, lengthLocation)
+}
+
+type ExportTypesPE struct {
```
kzantow, line 288: similar export naming comment as above (could be not exported, lowercase names), although here the fields do not need to be exported, either. the same applies to nearly all exported names following, too

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
kzantow: I also retested this on a somewhat large image (~550MB compressed), and the results were:

Syft main:
go run ./cmd/syft cloudbees/cloudbees-core-cm  36.64s user 14.22s system 93% cpu 54.441 total

This branch:
go run ./cmd/syft cloudbees/cloudbees-core-cm  38.29s user 15.00s system 100% cpu 53.254 total

... so an almost negligible difference from what I can tell at this point. 👍

kzantow (commented):
I had a fairly thorough review and only found a few more things that I think would be good to update. I could also help get these changes done, if you'd like -- just let me know!

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -0,0 +1,538 @@
...
+	AddressOfFunctions uint32
+	AddressOfNames     uint32
+	// Locations of SBOM symbols in the .data section
+	AddressOfSbom       uint32
+	AddressOfSbomLength uint32
+	AddressOfSvmVersion uint32
+}
+
+var t ExportTypesPE
+var header ExportPrefixPE
```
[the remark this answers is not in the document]
wdblair: Thanks for pointing this out, keeping these in global variables looked messy. In the latest revision, I refactored these types a little bit and made them a part of the `nativeImagePE` type. This way, I can access this information through a variable that represents the native image which is available in the relevant functions.

wdblair:
Thanks @kzantow for the quick review! I think I addressed all the points by keeping relevant types, fields, and functions as internal, declaring variables in a consistent manner, and keeping data in function parameters as opposed to global variables. Let me know if there are any other style/idiomatic items we can improve.

kzantow:
Hi @wdblair there was a refactoring of the `pkg.CPE` type to `cpe.CPE` (moved to the `github.com/anchore/syft/syft/cpe` package) just recently, it looks like that's where the failures are stemming from -- would you mind updating? (Again, I'm happy to help here, too, just shout)

wdblair:
Thanks @kzantow for catching that, I probably should have rebased before making my changes. We now create CPEs using the refactored `cpe` package.

kzantow:
... and one more gofmt issue -- just run `make lint-fix` to test/correct the static analysis stuff.

kzantow (approved):
Thanks for sticking with this contribution @wdblair, very much appreciated! 👍

wdblair:
Thanks @kzantow and @wagoodman for the helpful feedback, I have a much better understanding of Go idioms now.

PR #1278, opened 2022-10-20
fix: check pom properties for version (#1251)

### LINKED ISSUE #1251: Versions not printed out properly from maven pom.xml
Note, I would love to help with submitting fixes here if possible. Though I don't know your codebase all that well, so I might need a little direction.

**What happened**:
When I run `syft .` on the Apache Project [`commons-text`](https://github.com/apache/commons-text) (note this is after having run `mvn clean install`), I get the following output from `syft`:
```
NAME                      VERSION                     TYPE         
assertj-core              3.23.1                      java-archive  
commons-io                2.11.0                      java-archive  
commons-lang3             3.12.0                      java-archive  
commons-rng-simple        ${commons.rng.version}      java-archive  
jmh-core                  ${jmh.version}              java-archive  
jmh-generator-annprocess  ${jmh.version}              java-archive  
js                        ${graalvm.version}          java-archive  
js-scriptengine           ${graalvm.version}          java-archive  
junit-jupiter                                         java-archive  
mockito-inline            ${commons.mockito.version}  java-archive 
```

```
NAME                      VERSION                   TYPE         
assertj-core              3.23.1                    java-archive  
commons-io                2.11.0                    java-archive  
commons-lang3             3.12.0                    java-archive  
commons-rng-simple        1.4                       java-archive  
jmh-core                  1.35                      java-archive  
jmh-generator-annprocess  1.35                      java-archive  
js                        22.0.0.2                  java-archive  
js-scriptengine           22.0.0.2                  java-archive  
junit-jupiter             5.9.1                     java-archive  
mockito-inline            4.8.0                     java-archive 
```

**How to reproduce it (as minimally and precisely as possible)**: Running syft on any maven java project where in the pom.xml, the versions are declared in the properties section as opposed to directly in `<version></version>` line of the dependency declaration.

**Anything else we need to know?**: I think this was working in version 48 of `syft`, and now isn't. Also worth noting is `mvn dependency:tree` generates a vastly different list of dependencies.

```
commons-text % mvn dependency:tree
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------------< org.apache.commons:commons-text >-------------------
[INFO] Building Apache Commons Text 1.10.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:3.3.0:tree (default-cli) @ commons-text ---
[INFO] org.apache.commons:commons-text:jar:1.10.1-SNAPSHOT
[INFO] +- org.apache.commons:commons-lang3:jar:3.12.0:compile
[INFO] +- org.junit.jupiter:junit-jupiter:jar:5.9.1:test
[INFO] |  +- org.junit.jupiter:junit-jupiter-api:jar:5.9.1:test
[INFO] |  |  +- org.opentest4j:opentest4j:jar:1.2.0:test
[INFO] |  |  +- org.junit.platform:junit-platform-commons:jar:1.9.1:test
[INFO] |  |  \- org.apiguardian:apiguardian-api:jar:1.1.2:test
[INFO] |  +- org.junit.jupiter:junit-jupiter-params:jar:5.9.1:test
[INFO] |  \- org.junit.jupiter:junit-jupiter-engine:jar:5.9.1:test
[INFO] |     \- org.junit.platform:junit-platform-engine:jar:1.9.1:test
[INFO] +- org.assertj:assertj-core:jar:3.23.1:test
[INFO] |  \- net.bytebuddy:byte-buddy:jar:1.12.10:test
[INFO] +- commons-io:commons-io:jar:2.11.0:test
[INFO] +- org.mockito:mockito-inline:jar:4.8.0:test
[INFO] |  \- org.mockito:mockito-core:jar:4.8.0:test
[INFO] |     +- net.bytebuddy:byte-buddy-agent:jar:1.12.14:test
[INFO] |     \- org.objenesis:objenesis:jar:3.2:test
[INFO] +- org.graalvm.js:js:jar:22.0.0.2:test
[INFO] |  +- org.graalvm.regex:regex:jar:22.0.0.2:test
[INFO] |  +- org.graalvm.truffle:truffle-api:jar:22.0.0.2:test
[INFO] |  \- org.graalvm.sdk:graal-sdk:jar:22.0.0.2:test
[INFO] +- org.graalvm.js:js-scriptengine:jar:22.0.0.2:test
[INFO] +- org.apache.commons:commons-rng-simple:jar:1.4:test
[INFO] |  \- org.apache.commons:commons-rng-core:jar:1.4:test
[INFO] |     \- org.apache.commons:commons-rng-client-api:jar:1.4:test
[INFO] +- org.openjdk.jmh:jmh-core:jar:1.35:test
[INFO] |  +- net.sf.jopt-simple:jopt-simple:jar:5.0.4:test
[INFO] |  \- org.apache.commons:commons-math3:jar:3.2:test
[INFO] \- org.openjdk.jmh:jmh-generator-annprocess:jar:1.35:test
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.255 s
[INFO] Finished at: 2022-10-06T12:59:22-04:00
[INFO] ------------------------------------------------------------------------
```

**Environment**:
- Output of `syft version`: 0.58.0
- OS (e.g: `cat /etc/os-release` or similar): MacOS 12.6

chtompki:
I think this closes #1251. I'm not sure but there may be an analogous issue with gradle. I'll see if I can find and example and submit a similar fix.


syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -24,6 +24,10 @@ func parserPomXML(path string, content io.Reader) ([]*pkg.Package, []artifact.Re
 
 	var pkgs []*pkg.Package
 	for _, dep := range pom.Dependencies {
+		if strings.HasPrefix(dep.Version, "${") {
+			versionProperty := dep.Version[2 : len(dep.Version)-1]
+			dep.Version = pom.Properties.Entries[versionProperty]
+		}
```
kzantow, line 30: It might be useful to move this to a utility function, something like `resolveValue(pom, string) string`, which would do something like this and could be used to resolve every value when reading the pom -- e.g. a user _could_ use variables for `groupId` or `artifactId`. If, at some point in the future, we are able to resolve parent POMs, then this function would also be able to handle variable lookups in parents and the rest of the code would be unchanged. What do you think?

Also, I think this algorithm should probably be a little different -- the user _could_ have a variable as only a part of the string like `<version>1.2.${minor.version}-beta</version>` -- maybe using regex, something like (totally untested):

```go
var propertyMatcher = regexp.MustCompile("[$][{][^}]+[}]")

func resolveValue(pom gopom.Project, s string) string {
  return propertyMatcher.ReplaceAllStringFunc(s, func(match string) string {
    prop := match[:len(match)-1][2:]
    pom.Properties.Entries[prop]
  })
}
```

I could take a stab at this small refactoring, if you'd like.

chtompki: Great suggestions. This was my first cut at using `golang`

chtompki: I'll see if I can get to it before tomorrow

chtompki:
@kzantow - I think that last commit accomplishes what you're looking for. Let me know

syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -109,3 +113,11 @@ func cleanDescription(original string) (cleaned string) {
 	}
 	return strings.TrimSpace(cleaned)
 }
+
+func propertyReplacer(pomProperties gopom.Properties, stringWithVariablesToReplace string) string {
+	var propertyMatcher = regexp.MustCompile("[$][{][^}]+[}]")
```
kzantow, line 118: Quick note: this would be maybe more idiomatic go to initialize a variable in a function like `propertyMatcher := regexp.MustCompile("[$][{][^}]+[}]")` (instead of the `var` definition). However, this can be moved out of the function so the regex is only compiled once, e.g.:
```go
var propertyMatcher = regexp.MustCompile("[$][{][^}]+[}]")

func propertyReplacer(pomProperties gopom.Properties, stringWithVariablesToReplace string) string {
```

syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -109,3 +113,11 @@ func cleanDescription(original string) (cleaned string) {
 	}
 	return strings.TrimSpace(cleaned)
 }
+
+func propertyReplacer(pomProperties gopom.Properties, stringWithVariablesToReplace string) string {
+	propertyMatcher := regexp.MustCompile("[$][{][^}]+[}]")
+	return propertyMatcher.ReplaceAllStringFunc(stringWithVariablesToReplace, func(match string) string {
+		prop := match[2 : len(match)-1]
+		return pomProperties.Entries[prop]
```
kzantow, line 121: Thinking about this a little more, it might be good to only return this value if it's found in the map... so we can see which properties are not found more obviously. This would be something like:
```go
value, ok := pomProperties.Entries[prop]
if ok {
  return value
}
return match
```
The main reason I'd prefer to do this is because variables might be coming from a parent pom or another expression like `${project.groupId}`, which we're not currently processing and if these values end up blank, it is less obvious that there's an expression we might need to support.

kzantow:
@chtompki -- if you'd like, I can make a couple small updates to this that I think will get it to the finish line

chtompki:
Go for it. I'm curious to see what you do. I'll learn from that and try to use that style for any future PRs I submit

kzantow:
@chtompki thanks -- I went down a little bit of a rabbit hole that I wouldn't expect most people to do in order to support project properties like `${project.parent.groupId}`, which I've seen used enough to justify adding support for this while we're touching the property resolution. I've used some reflection for this because the `gopom` library has appropriate `xml` tags to make this fairly generic processing. We _still_ need a solution for resolving parent and dependency poms, but that's outside the scope of this change.

chtompki:
Cool...I see what you've done here, and that helps me think about the way I might make an analogous change for gradle. I don't have a good example for gradle yet, and I've only seen it once. But, I'm thinking about trying to make that my next contribution

spiffcs (approved):

PR #1279, opened 2022-10-21
Replace logger interface with anchore/go-logger

wagoodman:
This PR replaces the existing logger interface in syft with a superset defined in `anchore/go-logger`. This gives syft:
- a common logger interface across the anchore ecosystem
- trace logging
- centralized spot for logging adapters (not required though)
- field-logger support 🎉 
- less code to do the same thing 🎉 🌮


cmd/syft/cli/commands.go
```diff
@@ -118,7 +119,7 @@ func validateArgs(cmd *cobra.Command, args []string) error {
 }
 
 func checkForApplicationUpdate() {
-	log.Debugf("checking if new vesion of %s is available", internal.ApplicationName)
+	log.Debugf("checking if a new version of %s is available", internal.ApplicationName)
```
kzantow, line 122: 👏

internal/config/application.go
```diff
@@ -140,33 +140,25 @@ func (cfg *Application) parseLogLevelOption() error {
...
-		switch v := cfg.Verbosity; {
-		case v == 1:
-			cfg.Log.LevelOpt = logrus.InfoLevel
-		case v >= 2:
-			cfg.Log.LevelOpt = logrus.DebugLevel
-		default:
-			cfg.Log.LevelOpt = logrus.ErrorLevel
-		}
+		// TODO: there is a panic in this function when specifying more verbosity than whats available
+		cfg.Log.Level = logger.LevelFromVerbosity(int(cfg.Verbosity), logger.WarnLevel, logger.InfoLevel, logger.DebugLevel, logger.TraceLevel)
```
kzantow, line 147: This `TODO` is really for the `anchore/go-logger`, no?

internal/config/application.go
```diff
@@ -140,33 +140,25 @@ func (cfg *Application) parseLogLevelOption() error {
...
 			cfg.Verbosity = 1
 		}
 	default:
-		cfg.Log.LevelOpt = logrus.WarnLevel
-	}
-
-	if cfg.Log.Level == "" {
-		cfg.Log.Level = cfg.Log.LevelOpt.String()
+		// TODO: set default warn
+		cfg.Log.Level = logger.WarnLevel
```
kzantow, line 161: This `TODO` looks to be doing what it says?

kzantow (approved):
LGTM

internal/config/application.go
wagoodman: actually, this TODO isn't true anymore (it was fixed in go-logger, I'll remove here)

internal/config/application.go
wagoodman: ha! also defunct... or I'm being really really tricky somehow ... (I'll remove the todo)

cmd/syft/cli/commands.go
```diff
@@ -138,22 +139,33 @@ func checkForApplicationUpdate() {
...
 	log.Debugf("application config:\n%+v", color.Magenta.Sprint(app.String()))
 }
 
 func newLogWrapper(app *config.Application) {
-	cfg := logger.LogrusConfig{
+	cfg := logrus.Config{
 		EnableConsole: (app.Log.FileLocation == "" || app.Verbosity > 0) && !app.Quiet,
-		EnableFile:    app.Log.FileLocation != "",
-		Level:         app.Log.LevelOpt,
-		Structured:    app.Log.Structured,
```
spiffcs, line 150: Does this field have to be set later if `app.Log.Structured` is `true`

internal/config/logging.go
```diff
@@ -30,4 +30,5 @@ func (cfg *logging) parseConfigValues() error {
 func (cfg logging) loadDefaultValues(v *viper.Viper) {
 	v.SetDefault("log.structured", false)
 	v.SetDefault("log.file", "")
+	v.SetDefault("log.level", string(logger.WarnLevel))
```
spiffcs, line 33: Nice! Was about to ask about the todo that went away.

cmd/syft/cli/commands.go
spiffcs: This might have been force pushed away - let me check again

wagoodman: it does -- we only key off of a boolean from the app config, but the logger config needs a different formatter for this to work (the adapter doesn't have an opinion about the formatters, only allows folks to override them)

wagoodman: unless we moved this "structured" boolean to the logging adapter to `anchore/go-logger`. I'm trying to leave those adapters as opinionless as possible (to make it more flexible to the caller)

internal/config/logging.go
wagoodman: yeah, it felt quite nice to get rid of the multiple-fields to represent a logger level in the app config. And since the upstream has this field as essentially a string, we can use it directly like this 🎉 .

cmd/syft/cli/commands.go
wagoodman: > So under app.Log.Structured on line 153 should we also update cfg.Sructured to be true?

Today there is no `Structured` element on the config (unless we do https://github.com/anchore/syft/pull/1279#discussion_r1001884788):
https://github.com/anchore/go-logger/blob/main/adapter/logrus/logger.go#L23-L29

PR #1283, opened 2022-10-24
Port apk cataloger to new generic cataloger pattern

wagoodman:
Follow up to #1281 , porting the APK cataloger to the new `generic.Cataloger`


kzantow (approved):
LGTM

PR #1288, opened 2022-10-25
Update deb to new generic cataloger pattern

wagoodman:
Follow up to #1281 , porting the DEB cataloger to the new `generic.Cataloger`.

This PR also makes the following adjustments:
- updates the new `internal/pkgtest` helper with a builder pattern approach (the DEB cataloger needs these extra use cases and a multi-constructor based approach would become untenable here)
- With the last point, to test catalogers without introducing a package cycle, moving the `cataloger.Cataloger` interface to `pkg.Cataloger` was necessary. This is additionally consistent with the plan in https://github.com/anchore/syft/issues/558 .


kzantow (approved):
It looks like this includes #1287?

wagoodman:
re: 1287. yeah, I'm starting to chain PRs... I need to rebase this one so it won't include it

PR #1293, opened 2022-10-27
docs: update attestation directions with new cosign changes

pipern:
…based on cosign v1.13.1 and syft 0.59.0 usage

I've found that `cosign verify-attestation` defaults to `custom` type; the only pairing of syft and cosign I've found works for me is to use SPDX JSON


spiffcs:
Thanks @pipern - Would you also be able to update the branch with the DCO check?

https://github.com/anchore/syft/pull/1293/checks <--

spiffcs (approved):

PR #1294, opened 2022-10-28
feat: prefer known CPE vendors over other candidates

raboof:
All ASF projects will be under the `apache` vendor in CPE, and indeed this is already one of the candidates, but the logic for selecting the 'most specific' CPE string would select for example `apache_software_foundation` or `commons-text`.

This is not necessarily 'wrong' in the CPE candidate selection logic: there is no way to reliably determine the right candidate. I think it makes sense to use specific data around the vendor candidate generation, somewhat similar to
'defaultCandidateAdditions'.

Unfortunately there are still a few CVE's for old (pre-5.x, long unsupported) tomcat versions that are actually tagged with `apache_software_foundation`, but I'm not sure those are worth spending time on.


spiffcs:
Pushed a merge upstream for this @raboof -- I think this is something we can consider, but I would like to hear @anchore/tools opinion on returning such a specific list for the specific case

If this vendor is already considered for CPE generation then what is the gain for selecting such a specific value?

raboof:
> If this vendor is already considered for CPE generation then what is the gain

I suppose I should have turned the case I encountered into a test case ;)

I don't have the reproducer handy right now, but I saw situations where `apache` was considered for the vendor, but also `apache_software_foundation` or `commons-text`, and the logic picked `apache_software_foundation` or `commons-text` in those cases instead of the correct `apache`.

wagoodman:
I incorporated this PR (rebased) in grype and ran against the quality gate... the results showed no drop in matching quality 🙌 . I also ran some manual tests with it and overall it looks like a good change (we may add more vendors in the future). I do think there is some risk of pruning potentially better/more accurate CPEs for a non-apache product that for some reason has "apache" in one of the fields we are keying off of, however, I could not find an example of this happening.

wagoodman (approved):
I rebased and made one tweak (swapping the array for a set) but otherwise looks great! Thanks for the fix here 🙌

tbroyer:
Updating to 0.90.0 from 0.89.0 seems to have reintroduced this. Here's an excerpt from a generated sbom:
```diff
@@ -249,7 +249,7 @@
           <name>https://www.apache.org/licenses/LICENSE-2.0.txt</name>
         </license>
       </licenses>
-      <cpe>cpe:2.3:a:apache:commons-codec:1.15:*:*:*:*:*:*:*</cpe>
+      <cpe>cpe:2.3:a:apache-software-foundation:commons-codec:1.15:*:*:*:*:*:*:*</cpe>
       <purl>pkg:maven/commons-codec/commons-codec@1.15</purl>
       <externalReferences>
         <reference type="build-meta">
@@ -264,8 +264,15 @@
         <property name="syft:package:language">java</property>
         <property name="syft:package:metadataType">JavaMetadata</property>
         <property name="syft:package:type">java-archive</property>
-        <property name="syft:cpe23">cpe:2.3:a:apache:commons_codec:1.15:*:*:*:*:*:*:*</property>
-        <property name="syft:cpe23">cpe:2.3:a:apache:commons:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:apache-software-foundation:commons_codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:apache_software_foundation:commons-codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:apache_software_foundation:commons_codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:commons-codec:commons-codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:commons-codec:commons_codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:commons_codec:commons-codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:commons_codec:commons_codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:commons:commons-codec:1.15:*:*:*:*:*:*:*</property>
+        <property name="syft:cpe23">cpe:2.3:a:commons:commons_codec:1.15:*:*:*:*:*:*:*</property>
         <property name="syft:location:0:path">/0/keycloak-config-cli.jar</property>
         <property name="syft:metadata:-:artifactID">commons-codec</property>
         <property name="syft:metadata:-:groupID">commons-codec</property>
```

raboof:
> Updating to 0.90.0 from 0.89.0 seems to have reintroduced this

The other way around I guess?

> Here's an excerpt from a generated sbom:

Hmm, it seems `apache` isn't even generated as a candidate anymore (and thus not selected). I guess it would be worth looking into what changed in the candidate generation. Interesting to note that this apparently wasn't caught by `./syft/pkg/cataloger/common/cpe/generate_test.go`, which does have a case similar to this.

PR #1296, opened 2022-10-28
feat: add nodejs-binary package classifier

spiffcs:
## Add nodejs-binary classifier to package cataloger

This PR allows for the generation of a `node.js` package if the following conditions are met:
- syft considers all files found for the executable mime type list here https://github.com/anchore/syft/blob/bd9007fc0e3de88685a55d3aafedd7af6e233922/internal/mimetype_helper.go#L55-L63
- If a files real path or virtual path matches this pattern its contents are considered: `regexp.MustCompile(`(.*/|^)node$`),`
- The contents are matched against the following: `(?m)node\.js\/v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`

If the 3 above checks are fulfilled a node.js package is entered into the catalog with the captured version from the file contents.

Here is an example of the potential record we could now find via `grype`
```
package: node.js 
namespace: nvd:cpe 
constraints: >= 8.9.0, < 8.11.4 \|\| >= 8.0.0, < 8.8.1 \|\| >= 10.0.0, < 10.9.0 \|\| >= 6.9.0, < 6.14.4 \|\| >= 6.0.0, < 6.8.1 
cpe: ["cpe:2.3:a:nodejs:node.js:*:*:*:*:lts:*:*:*","cpe:2.3:a:nodejs:node.js:*:*:*:*:-:*:*:*"]
```

### With the current change here is the output of syft => grype:
`go run cmd/syft/main.go node:16.11.0-alpine -o json | grype`
Note the new detections of `node` as type `binary`:

Screen Shot 2022-10-31 at 11 15 11 AM


syft/file/classifier.go
```diff
@@ -111,11 +122,15 @@ func (c Classifier) Classify(resolver source.FileResolver, location source.Locat
 			continue
 		}
 
+		// TODO: it looks like we're getting two duplicate classificationns in some cases where
+		// VirtualPath differs from RealPath. The classifications don't show this difference
+		// Should we update the classifications to identify the virtual path that linked to the realpath?
```
spiffcs, line 127: I'll remove this `TODO` if we like the VirtualPath change. It's also a reminder for me to update the schema if we go this route.

internal/config/file_classification.go
```diff
@@ -11,7 +11,7 @@ type fileClassification struct {
 }
 
 func (cfg fileClassification) loadDefaultValues(v *viper.Viper) {
-	v.SetDefault("file-classification.cataloger.enabled", catalogerEnabledDefault)
```
spiffcs, line 14: File classifiers aren't that expensive and probably should be removed or enabled by default. This PR attempts to start a general `binary` cataloger

syft/pkg/cataloger/binary/cataloger.go
```diff
@@ -0,0 +1,11 @@
+package binary
+
+import (
+	"github.com/anchore/syft/internal"
+	"github.com/anchore/syft/syft/pkg/cataloger/generic"
+)
+
+func NewBinaryCataloger() *generic.Cataloger {
+	return generic.NewCataloger("binary-cataloger").
+		WithParserByMimeTypes(parseBinary, internal.ExecutableMIMETypeSet.List()...)
```
spiffcs, line 10: Only run the cataloger against `ExecutuableMIMETypeSet`

spiffcs:
We're detecting node and busy box for other distributions, but `go run cmd/syft/main.go busybox:1.34.1 -o json` still reveals no packages - trying to see why the MIMEType for `/bin/busybox` is not matching correctly for that image

syft/pkg/cataloger/binary/parse_binary.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+
+			matchMetadata := internal.MatchNamedCaptureGroups(pattern, string(contents))
+			if binPkg == nil {
+				binPkg = &pkg.Package{
+					Name:         binParser.Package,
+					Version:      matchMetadata["version"],
+					Language:     pkg.UnknownLanguage,
+					Locations:    source.NewLocationSet(reader.Location),
+					Type:         pkg.UnknownPkg,
```
westonsteimel, line 99: Should we break this out into a parser file per binary or per ecosystem?  So for example we could have one for `rust` that handles parsing for any of the tools that are part of the rust ecosystem (cargo, rustc, etc) or a file per binary.  And then have an actual package type here `pkg.Binary` rather than just using `Unknown`

spiffcs: Yea breaking it out seems better the more I work over in grype - will do that right now

syft/pkg/cataloger/javascript/parse_node_binary.go
```diff
@@ -0,0 +1,38 @@
+package javascript
```
kzantow, line 1: I wonder if this belongs in something more like `syft/pkg/cataloger/binary/node.go` or similar, since it's not _exactly_ javascript but rather a mix of lots of C and some javascript.

syft/pkg/cataloger/javascript/parse_node_binary.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var nodeLookup = generic.Lookup{
+	Package: "node.js", // Note: this purposely matches the "node.js" string to aid nvd vuln matching
+	FilepathPatterns: []*regexp.Regexp{
+		// note: should we just parse all files resolved with executable mimetypes
+		// regexp that matches node binary
+		regexp.MustCompile(`(.*/|^)node$`),
+	},
+	EvidencePatternTemplates: []string{
```
kzantow, line 20: Would this be better as a callback function against the matched file contents if things are a bit more complicated than a regex?

syft/pkg/cataloger/javascript/parse_node_binary.go
```diff
@@ -0,0 +1,38 @@
...
+
+func parseNodeBinary(_ source.FileResolver, _ *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	p, _, err := nodeLookup.Find(reader)
+	if err != nil {
+		log.Trace("failed to find node.js package: %+v", err)
+		return nil, nil, nil // we can silently fail here to reduce warning noise
+	}
+	// TODO add node specific metadata to the packages to help with vulnerability matching
+	if p != nil {
+		p.Language = pkg.JavaScript
```
kzantow, line 34: Is this accurate? As noted before, this is predominantly C code with some javascript... but perhaps it matches things in the ecosystem in GHSA/etc.? Without digging in to this I would think the `javascript` ecosystem would be for javascript packages but _not_ the node binaries

syft/pkg/cataloger/generic/lookup.go
```diff
@@ -0,0 +1,85 @@
...
+// Lookup is a generic package lookup that can be used to match a package definition
+// To a file that meets the given content criteria of the EvidencePatternTemplates.
+type Lookup struct {
+	Package string
+	// FilepathPatterns is a list of regular expressions that will be used to match against the file path of a given
+	// source location. If any of the patterns match, the file will be considered a candidate for parsing.
+	FilepathPatterns []*regexp.Regexp
+	// EvidencePatternTemplates is a list of regular expressions that will be used to match against the file contents of a
+	// given file in the source location. If any of the patterns match, the file will be considered a candidate for parsing.
+	EvidencePatternTemplates []string
```
kzantow, line 28: Just noting the comment from below that maybe this should be a callback function?

syft/pkg/cataloger/generic/lookup.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/internal/unionreader"
+	"github.com/anchore/syft/syft/source"
+)
+
+// Lookup is a generic package lookup that can be used to match a package definition
+// To a file that meets the given content criteria of the EvidencePatternTemplates.
+type Lookup struct {
```
kzantow, line 21: Would this be more of a `Binary` cataloger instead of `Lookup`? I might be misunderstanding the purpose. Ok, I think I understand now.

This seems like it's essentially doing what the catalogers are doing already, is this abstraction needed? It might be a little simpler -- since we're already adding a node binary cataloger -- to just make the `Find` function below a utility function for the `Package` and `EvidencePatternTemplates` and have it called by the node binary cataloger (which it already is doing) and specify the file path patterns in the `NewNodeBinaryCataloger`

kzantow (commented):
This is looking pretty good -- a few questions (also in the comments):
* Is the `Lookup` abstraction needed, since it's sorta doing what the `Cataloger` is doing already?
* is `pkg.JavaScript` the right type for this, since it's mostly C code? (I don't really know where to expect to find the vulnerabilities for node itself)
* I don't see the `VirtualPath` really being used -- is this something that is needed downstream by Grype?

syft/pkg/cataloger/javascript/parse_node_binary.go
spiffcs: I organized it this way since when we're trying to identify an ecosystem-specific tool I believe it's easier to reason about the coverage of that grouping without spreading it too far across other packages:

- Example if we wanted to discover vulnerable versions of `npm` it would also be `package javascript`
- Lookups for things like `cargo, rustc` would fall under `package rust` 

Also see weston's comment here for the rust example
https://github.com/anchore/syft/pull/1296#discussion_r1008656419

syft/pkg/cataloger/javascript/parse_node_binary.go
spiffcs: ^ The language might be a little loose - we could default this to `Unknown`, but if we're able to enter a namespace in the grype db where `node.js` has a vulnerability filed against it that was not caught or listed incorrectly in NVD this gives us a bit more of a shot at this package being compared against the correct data. I do agree it's awkward.

syft/pkg/cataloger/generic/lookup.go
spiffcs: I think it's helpful to talk about some of the newer code that was recently merged to help explain the reason `Lookup` exists and why this still is a `BinaryCataloger`.

https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/generic/cataloger.go <-- Here is the upgraded `Generic` cataloger that we're utilizing for `node-binary-cataloger` construction.

https://github.com/anchore/syft/blob/4fea2fa71e926acc27947f8b697706e9efdbc85d/syft/pkg/cataloger/javascript/cataloger.go#L40-L43

`parseNodeBinary` is the `Parser` that was added here which is associated with a location:
https://github.com/anchore/syft/blob/4fea2fa71e926acc27947f8b697706e9efdbc85d/syft/pkg/cataloger/generic/cataloger.go#L83-L92

In our case, the parser is going to look at every location that has a MIMEType in this list:
https://github.com/anchore/syft/blob/bd9007fc0e3de88685a55d3aafedd7af6e233922/internal/mimetype_helper.go#L55-L62

It is not matching on the glob patterns from the get-go.

Rather than write `nodeBinaryParser` `goBinaryParser` `cargoBinaryParser`, etc with the duplicated logic of further refinement via `FilepathPatterns` and then content analysis with`checkEvidenceInContent`, I wanted to condense that in a single place like the classification does for the `file` package ==> `Lookup` was born.

So this is not as a cataloger that's initially matching on `glob` in the first place, we should instead think of it as a cataloger that pulls in all files with the executable mime-type and then has a mechanism to further refine that executable list via glob/regex for a list of FilepathPatterns.

Of note! If FilepathPatterns is empty, we should just do a content analysis of `EvidencePatternTemplate` against all files returned by the cataloger. I'll make that update now!.

Here is an example of this where we are doing this with the new `GoBinary` parsing:
https://github.com/anchore/syft/blob/6826d7603bfcec9f3d2654609987ffb27f9eb7dd/syft/pkg/cataloger/golang/parse_go_binary.go#L41

So it's not `doing what the catalogers are doing already`, since we need some way to match against `EvidencePatternTemplates` across ecosystems that are not duplicated

The result of this generic `EvidencePatternTemplate` matching should then be returned to the parsers to assign package/ecosystem-specific metadata for the given lookup. The package in the parser is still mutable after coming back from the `Lookup`

`Lookup` also serves the purpose in giving a common way to configure the `Package Name` and potentially other domain-specific details you're searching for that will be constructed on a HIT against those pattern templates.

syft/pkg/cataloger/generic/lookup.go
spiffcs: Added a longer explanation to the main comment

syft/pkg/cataloger/javascript/parse_node_binary.go
spiffcs: Added a longer explanation to the main comment

syft/pkg/cataloger/generic/lookup.go
kzantow: Got it -- thank you for the detailed response here! So the way I see this is: we have a multi-factor match we're looking for: _executable_ AND _matching the node regex_, essentially a new cataloger builder function that would be something like `WithParserByMimeTypesAndPattern`, following the current pattern (I won't get into the naming or combinatorial explosion here, which could both be valid)? I also don't have _any_ problem with a configuration struct, as you noted, or whichever direction (utility function or method) it's used for the `Locator`. My main question is around conflating file lookups in the midst of it all, which seem to be a `Cataloger` concern.

syft/pkg/cataloger/javascript/parse_node_binary.go
kzantow: I'm definitely on board with `pkg.JavaScript` being better than `pkg.Unknown`, if that's the only other option 👍

syft/pkg/cataloger/generic/lookup.go
spiffcs: We don't have to call it lookup then. `Matcher` `Classifier` `Identifier` <--? pick you're favorite

syft/pkg/cataloger/javascript/parse_node_binary.go
kzantow: Definitely no strong opinion here... organizing by _ecosystem_ seems to make sense. 👍

syft/pkg/cataloger/generic/lookup.go
kzantow: It's not the naming as much, but the fact it's doing file matching, which the `Cataloger` is already doing and could be expanded upon to do both _executable_ and _pattern_ matches together. In this case I'd probably rename it from `Locator` as well, to something like `ContentMatcher`

syft/pkg/cataloger/javascript/parse_node_binary.go
spiffcs: That was for the pkg.Type which is being set in `Find` as `pkg.Binary`

`Find` sets the language as unkown and then leaves the decision to the caller on if want's to enhance it with a known language or ecosystem. Definitely awkward

westonsteimel:
One other thing that seems kinda weird from a UX perspective is the package name being `node.js` in the output.  I think I'd expect that to match the actual name of the binary from the path?

westonsteimel:
I'm not really sure if I fully agree on my comment above, just something feels off here.  I think I'd expect the name raised by a binary cataloger to match the name of the binary, but there to be some metadata associated with it with maybe a slice of classifications, one of which in this case would be `node.js`?  Because I guess it's possible we classify a binary as possibly being more than one thing?

westonsteimel:
Also, in the future I think it would be cool to include every binary we find as an artifact, but some would be unclassified or classified as unknown.  That could give a way to measure the completeness of an SBOM

westonsteimel:
Hmm actually this is going to have implications for the labelling as well since the labels are meant to be tool agnostic and the `name` surfaced here currently won't necessarily be

westonsteimel:
Although perhaps we can adjust the yardstick logic to account for the binary type differently from other artifact types, so instead of using name it'd use the full path to the binary or something like that

spiffcs:
> One other thing that seems kinda awkward from a UX perspective is the package name being `node.js` in the output. I think I'd expect that to match the actual name of the binary from the path? So probably `nodejs` in this case? I'm guessing maybe it's this way currently to try and force the correct CPE-generation logic? I wonder if it's better to just be explicit about which CPEs should be created for binary catalogers until the point where we can maybe relocate some of that to the grypedb and make it not be syft's job to try and determine that?

I can make the update that gives us the name of the binary and make a `BinaryMetadataType` that has the list of classifications we think to apply. Those classifications for now will have explicit CPE.

syft/pkg/cataloger/javascript/parse_node_binary.go
```diff
@@ -0,0 +1,43 @@
...
+var nodeClassifier = generic.Classifier{
+	Package: "node.js", // Note: this purposely matches the "node.js" string to aid nvd vuln matching
+	FilepathPatterns: []*regexp.Regexp{
+		// note: should we just parse all files resolved with executable mimetypes
+		// regexp that matches node binary
+		regexp.MustCompile(`(.*/|^)node$`),
+	},
+	EvidencePatternTemplates: []string{
+		// regex that matches node.js/vx.y.z
+		`(?m)node\.js\/v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`,
```
kzantow, line 22: Could this just be a regex to keep this change more targeted to what it's doing? I don't _see_ any templating that's needed here and it would simplify this bit here: https://github.com/anchore/syft/pull/1296/files#diff-4ed9ad02a509d1b6975f51c8902c2bda275ee91011f73ef48bca5b3249358724R53-R72

kzantow (approved):
👍

PR #1311, opened 2022-11-02
SPDX 2.3 support

### LINKED ISSUE #1292: Support spdx 2.3

The latest version of spdx specification is 2.3. SBOMs produced by Syft are still on version 2.2.
It would be nice to add support for 2.3. Other sbom authoring tools like ko, apko and kubernetes sbom tool already support producing sbom with version 2.3 as the default.

https://spdx.github.io/spdx-spec/v2.3/

kzantow:
This PR makes a few notable changes:

* Updates the `spdx/tools-golang` library to the current `main` (which is not released)
* Updates the basic SPDX encoding/decoding to version 2.3 (there are some more enhancements to come later)
* Consolidates the SPDX JSON and TagValue code paths to use the same conversion logic, which also:
  * Significantly improves both SPDX JSON and TagValue outputs

Fixes #1292 

NOTE to reviewers: make sure to look at `syft/formats/common/spdxhelpers/to_format_model.go`, which is a consolidation of the tag-value and json converters, one of the more important things in this PR but it's collapsed by default in the GitHub UI.


syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -0,0 +1,500 @@
...
+		id = SanitizeElementID(fmt.Sprintf("Package-%+v-%s-%s", v.Type, v.Name, v.ID()))
+	case artifact.Identifiable:
+		id = string(v.ID())
+	case artifact.ID:
+		id = string(v)
+	case string:
+		id = v
+	default:
+		// FIXME don't panic here
+		panic(fmt.Sprintf("Invalid ID type: %+v", v))
```
kzantow, line 151: NOTE: need to fix this before merging

kzantow:
NOTE: this is blocked on a fix in the `tools-golang` repo: https://github.com/spdx/tools-golang/pull/170

syft/formats/common/spdxhelpers/to_format_model_test.go
```diff
@@ -169,93 +168,6 @@ func Test_toFileChecksums(t *testing.T) {
 	}
 }
 
-func Test_fileIDsForPackage(t *testing.T) {
```
kzantow, line 172: This has been deprecated in favor of using the `CONTAINS` relationship, which we are already including. See: http://spdx.org/rdf/terms#relationshipType_contains

kzantow, line 172: The `hasFiles` field, which was using this function is not documented in the SPDX spec but is in the RDF and JSON schemas. However, it has been deprecated since 2.0 in favor of using the `CONTAINS` relationship, which we are already including. See: https://spdx.org/rdf/terms/#d4e4761

wagoodman: is there an equivalent functionality that needs to tested now that this test is being removed?

wagoodman (approved):
😎

PR #1317, opened 2022-11-03
Add java virtualpath to metadata

VandanaAkhramka:
When the components come from an spdx-json file, the source_path contains valuable info that just gets thrown away when its converted to the syft sbom format. Keep that info, very useful for debugging.

### DIFF
--- a/syft/formats/common/spdxhelpers/to_syft_model.go
+++ b/syft/formats/common/spdxhelpers/to_syft_model.go
@@ -353,7 +353,7 @@ func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interf
 		}
 		return pkg.JavaMetadataType, pkg.JavaMetadata{
 			ArchiveDigests: digests,
-			VirtualPath: p.PackageSourceInfo,
+			VirtualPath:    p.PackageSourceInfo,
 		}
 	case pkg.GoModulePkg:
 		var h1Digest string
--- a/syft/formats/common/spdxhelpers/to_syft_model.go
+++ b/syft/formats/common/spdxhelpers/to_syft_model.go
@@ -355,7 +355,6 @@ func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interf
 			ArchiveDigests: digests,
 			VirtualPath: p.PackageSourceInfo
 		}
-		
 	case pkg.GoModulePkg:
 		var h1Digest string
 		for _, value := range p.PackageChecksums {
--- a/syft/formats/common/spdxhelpers/to_syft_model.go
+++ b/syft/formats/common/spdxhelpers/to_syft_model.go
@@ -355,6 +355,7 @@ func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interf
 			ArchiveDigests: digests,
 			VirtualPath: p.PackageSourceInfo
 		}
+		
 	case pkg.GoModulePkg:
 		var h1Digest string
 		for _, value := range p.PackageChecksums {
--- a/syft/formats/common/spdxhelpers/to_syft_model.go
+++ b/syft/formats/common/spdxhelpers/to_syft_model.go
@@ -353,7 +353,7 @@ func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interf
 		}
 		return pkg.JavaMetadataType, pkg.JavaMetadata{
 			ArchiveDigests: digests,
-			VirtualPath: p.PackageSourceInfo
+			VirtualPath: p.PackageSourceInfo,
 		}
 	case pkg.GoModulePkg:
 		var h1Digest string
--- a/syft/formats/common/spdxhelpers/to_syft_model.go
+++ b/syft/formats/common/spdxhelpers/to_syft_model.go
@@ -353,6 +353,7 @@ func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interf
 		}
 		return pkg.JavaMetadataType, pkg.JavaMetadata{
 			ArchiveDigests: digests,
+			VirtualPath: p.PackageSourceInfo
 		}
 	case pkg.GoModulePkg:
 		var h1Digest string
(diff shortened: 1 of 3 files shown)


kzantow:
Thanks for the contribution @VandanaAkhramka -- would it be possible to add a simple test for this change?

VandanaAkhramka:
@kzantow , I've added a small unit test. Could you please review and release a patched tag if you're happy with it?

syft/formats/spdx22json/decoder_test.go
```diff
@@ -80,6 +80,15 @@ func TestSPDXJSONDecoder(t *testing.T) {
...
 							continue packages
 						}
+						switch p.Type {
+						case pkg.JavaPkg:
+							if p.Metadata.(pkg.JavaMetadata).VirtualPath != "acquired package info from installed java archive: /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/ext/nashorn.jar" {
+								assert.NoError(t, fmt.Errorf("Mismatch in jar for package: %s. "+
+									"Expected `acquired package info from installed java archive: "+
+									"/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/ext/nashorn.jar` and "+
+									"received VirtualPath: %s ", pkgName, p.Metadata.(pkg.JavaMetadata).VirtualPath))
+							}
```
kzantow, line 90: You could just do an `assert.Equal(t, "acquired package info from installed java archive: /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/ext/nashorn.jar", p.Metadata.(pkg.JavaMetadata).VirtualPath)`

syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -353,6 +353,7 @@ func extractMetadata(p *spdx.Package2_2, info pkgInfo) (pkg.MetadataType, interf
 		}
 		return pkg.JavaMetadataType, pkg.JavaMetadata{
 			ArchiveDigests: digests,
+			VirtualPath:    p.PackageSourceInfo,
```
kzantow, line 356: Looking at the test added here, I am concerned that this is changing the semantics of the `VirtualPath` -- in other words: setting values here that do not directly represent a path. This won't _necessarily_ cause a problem with things unless, for example, SPDX output needs to find a file to generate a SHA-1 hash from. Is there a way that we could search for file path patterns in the `p.PackageSourceInfo` and only include a path if we find exactly 1?

kzantow (commented):
I left a little feedback here, about maybe trying to search for file paths instead of just dumping the entire text into `VirtualPath`

syft/formats/common/spdxhelpers/to_syft_model.go
VandanaAkhramka: That uncovers a potential bug; according to section 7.12.1 in the spdx spec (https://spdx.github.io/spdx-spec/v2.2.2/package-information/), the source information field has a cardinality of 0..1. The logic in syft/formats/common/spdxhelpers/source_info.go just concatenates all the paths it finds. 

In our use case, we use syft to create an SPDX-json SBOM. We then do some manipulation on it, so we need the 2 to be compatible. 
Can we merge this and raise another issue to make syft only add 1 path to source_info and then add the check when creating `VirtualPath`?

syft/formats/spdx22json/decoder_test.go
VandanaAkhramka: Thanks!

syft/formats/spdx22json/decoder_test.go
```diff
@@ -80,6 +80,12 @@ func TestSPDXJSONDecoder(t *testing.T) {
 						if p.Name == pkgName {
 							continue packages
 						}
+						switch p.Type {
+						case pkg.JavaPkg:
+							if p.Metadata.(pkg.JavaMetadata).VirtualPath != "acquired package info from installed java archive: /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/ext/nashorn.jar" {
+								assert.Equal(t, "acquired package info from installed java archive: /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/ext/nashorn.jar", p.Metadata.(pkg.JavaMetadata).VirtualPath)
+							}
```
kzantow, line 87: This doesn't need an `if` statement, just use `assert.Equal`

syft/formats/common/spdxhelpers/to_syft_model.go
kzantow: Looking at a Syft-generated SPDX SBOM, it's not including the java metadata `VirtualPath` at all, only the location RealPath: https://github.com/anchore/syft/blob/main/syft/formats/common/spdxhelpers/source_info.go#L52

There are 2 potential paths here:
1. importing a Syft-generated SPDX document
2. importing a non-Syft-generated SPDX document

For case 1, we'd be setting `VirtualPath` to a totally different value, which is not something we would want to do. For case 2, we'd possibly like to gather as much information as we can, but this is probably less important to be the Java metadata VirtualPath but would be better suited as locations.

In both of these cases we could match all paths and add them as locations, which would reverse the current Syft encoding logic but would leave VirtualPath empty.

I don't think we're going to merge a change that is setting this value to something arbitrary, but let me ask more specifically: what is the problem you're seeing by this value not being imported?

wagoodman:
From @kzantow points:
> There are 2 potential paths here:
>
> 1. importing a Syft-generated SPDX document
>
> 2. importing a non-Syft-generated SPDX document
>
> For case 1, we'd be setting VirtualPath to a totally different value, which is not something we would want to do. For case 2, we'd possibly like to gather as much information as we can, but this is probably less important to be the Java metadata VirtualPath but would be better suited as locations.

This resonates with me -- we shouldn't be overloading existing fields to mean semantically different things. If there is a useful field we should be capturing we should capture it separately as a new field.

> let me ask more specifically: what is the problem you're seeing by this value not being imported?

@VandanaAkhramka I'm also interested in the use case here, as that might lead to a different or more broad solution. 

In the meantime I'm going to close this PR but am still interested in a conversation about this; please feel free to open an issue or another draft/concept PR!

PR #1319, opened 2022-11-03
Port python cataloger to new generic cataloger pattern

wagoodman:
Follow up to https://github.com/anchore/syft/pull/1281 , porting the python cataloger to the new generic.Cataloger.


syft/pkg/cataloger/python/package.go
```diff
@@ -0,0 +1,81 @@
...
+		Language:  pkg.Python,
+		Type:      pkg.PythonPkg,
+	}
+
+	p.SetID()
+
+	return p
+}
+
+func newPackageForPackage(m pkg.PythonPackageMetadata, sources ...source.Location) pkg.Package {
```
kzantow, line 26: Fun name :)

wagoodman: yeah, I couldn't think of another 👎

PR #1321, opened 2022-11-04
Port RPM cataloger to new generic cataloger pattern

wagoodman:
Follow up to https://github.com/anchore/syft/pull/1281 , porting the RPM cataloger to the new generic.Cataloger.


kzantow (approved):
🎉 -140 LoC

PR #1322, opened 2022-11-04
Port ruby cataloger to new generic cataloger pattern

wagoodman:
Follow up to https://github.com/anchore/syft/pull/1281 , porting the ruby cataloger to the new generic.Cataloger.

### DIFF
--- a/syft/pkg/cataloger/ruby/parse_gemspec.go
+++ b/syft/pkg/cataloger/ruby/parse_gemspec.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
-	"io"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


syft/pkg/cataloger/ruby/parse_gemspec.go
```diff
@@ -36,14 +35,14 @@ var patterns = map[string]*regexp.Regexp{
 	"homepage": regexp.MustCompile(`.*\.homepage\s*=\s*["']{1}(?P<homepage>.*)["']{1} *`),
 
 	// match example:       files = ["exe/bundle".freeze, "exe/bundler".freeze]    --->    "exe/bundle".freeze, "exe/bundler".freeze
-	"files": regexp.MustCompile(`.*\.files\s*=\s*\[(?P<files>.*)\] *`),
+	"files": regexp.MustCompile(`.*\.files\s*=\s*\[(?P<files>.*)] *`),
```
spiffcs, line 38: Why did we have to remove matching on the `\` here?

spiffcs (approved):

syft/pkg/cataloger/ruby/parse_gemspec.go
wagoodman: IDE flagged this as a redundant escape (which appears correct)

PR #1325, opened 2022-11-04
# DO NOT MERGE GITHUB ACTION WIP#

spiffcs:


spiffcs:
Quick check for the delta between `act` and the actual runner

PR #1327, opened 2022-11-06
feat: Add the origin field to the output format of syftjson

asi-cider:
This PR adds the origin field to the output format of syftjson when parsing “package-lock.json” or “Pipfile.lock”. The origin field can be useful in tracking the origin of downloaded packages in case of a security incident. The origin field contains the URL used to download an artifact according to the appropriate lock file. For example, package-lock.json has the resolved field to indicate the url used to download the package. 

I wasn’t able to find a way to update the golden files of the integration tests similarly to the argument in the unit test’s golden files. Could you please guide me on how to update them?

Changes to the golden files were done using the appropriate parameters of the test files.


kzantow (changes requested):
Thanks for the contribution @asi-cider! It looks like this `Origin` field is specific to NPM packages and as such should not be added to the main `pkg.Package` but rather the appropriate _metadata_ type. In this case we probably need a _new_ metadata type, something like `PackageLockPackageMetadata`, which is set [when creating the package](https://github.com/anchore/syft/blob/10f43d75e0585ecb09bbb1fca629bd071375e015/syft/pkg/cataloger/javascript/package.go#L46-L98), which contains the Origin field. [An example metadata type is here](https://github.com/anchore/syft/blob/10f43d75e0585ecb09bbb1fca629bd071375e015/syft/pkg/cataloger/javascript/package.go#L29-L38).

I think this change would also reduce some of the conflicts, as it stands there are a number of conflicts to resolve if you are able to.

asi-cider:
> Thanks for the contribution @asi-cider! It looks like this `Origin` field is specific to NPM packages and as such should not be added to the main `pkg.Package` but rather the appropriate _metadata_ type. In this case we probably need a _new_ metadata type, something like `PackageLockPackageMetadata`, which is set [when creating the package](https://github.com/anchore/syft/blob/10f43d75e0585ecb09bbb1fca629bd071375e015/syft/pkg/cataloger/javascript/package.go#L46-L98), which contains the Origin field. [An example metadata type is here](https://github.com/anchore/syft/blob/10f43d75e0585ecb09bbb1fca629bd071375e015/syft/pkg/cataloger/javascript/package.go#L29-L38).
> 
> I think this change would also reduce some of the conflicts, as it stands there are a number of conflicts to resolve if you are able to.

Hey @kzantow, thanks for the response. This field exists in some form or another in all kinds of lock files. Currently I've add this for npm and pipfile but I intend to add this for more ecosystems. Do you still think it should be added as metadata field?

wagoodman:
@asi-cider I agree with @kzantow  on this one. Though there are some similarities between different package managers on this field, at this time I think keeping this kind of field in the `pkg.Metadata` makes the most sense.

asi-cider:
> @asi-cider I agree with @kzantow on this one. Though there are some similarities between different package managers on this field, at this time I think keeping this kind of field in the `pkg.Metadata` makes the most sense.

Sure i'll add that in `pkg.Metadata`

asi-cider:
Hey @kzantow and @wagoodman,
I've made the changes I would appreciate your approval.

syft/pkg/cataloger/javascript/package.go
```diff
@@ -91,6 +92,7 @@ func newPackageLockV2Package(resolver source.FileResolver, location source.Locat
 			Language:  pkg.JavaScript,
 			Type:      pkg.NpmPkg,
 			Licenses:  licenses,
+			Metadata:  Metadata{Resolved: u.Resolved},
```
kzantow, line 95: There are a few more changes I think we need here:
This should be named something more unique like `NpmMetadata`, and we also need to set a corresponding `MetadataType` in the package struct here. These currently all defined in: [syft/pkg](https://github.com/anchore/syft/blob/main/syft/pkg/)

We also need to add any new metadata types to the appropriate spots in [syft/pkg/metadata.go](https://github.com/anchore/syft/blob/main/syft/pkg/metadata.go)

Similar comments for the `python.Metadata` below.

kzantow (changes requested):

syft/pkg/cataloger/javascript/package.go
asi-cider: done

asi-cider:
Hey @kzantow, 

Could you guide me on how to fix the unit tests?
Regarding the CLI test, to what version should I bump the schema?

asi-cider:
> 

Hey @kzantow @wagoodman, could you please help?

kzantow:
@asi-cider I'll have a look today, but there's a conflict that needs to be dealt with -- are you able to merge the latest `main` changes and take care of this?

wagoodman:
I can help out with the conflicts

wagoodman:
I rebased on main and made the following changes:
- make the metadata struct names as specific as possible relative to the parsed data sources they represent
- move the metadata structs to the `syft.pkg` package
- regenerated the JSON schema to account for the new fields. (since 6.1.0 hasn't been released yet, it can safely be regenerated)

wagoodman (approved):
Thanks for the improvements @asi-cider !

syft/pkg/cataloger/javascript/package.go
wagoodman: the latest commit accounts for this (see https://github.com/anchore/syft/pull/1327#issuecomment-1380925363 and https://github.com/anchore/syft/pull/1327/commits/8cecacfbfd6fc6c16be985cbbecdffffb1dd6884)

PR #1336, opened 2022-11-10
Generic Binary Cataloger

kzantow:
This PR moves the file classifiers to a binary file cataloger, which surfaces packages based on binary files. This currently requires knowledge about a curated set of binaries including how to determine version information, which may mean scanning strings within the binary. This currently supports:

* python
* go
* node (which was already being detected)
* busybox

TODO:
- [x] get feedback from the team
- [x] clean up commented out catalogers
- [x] ~add new schema~


kzantow:
cc: @jedevc

wagoodman (commented):
we should probably unexport some of the resources within the `binary` package such as `DefaultClassifiers` and `Classifier` since the only thing we intend for users to leverage is the cataloger itself

syft/pkg/cataloger/binary/cataloger.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/source"
+)
+
+const catalogerName = "binary-cataloger"
+
+func NewBinaryCataloger() pkg.Cataloger {
```
wagoodman, line 11: nit: since the package name is already `binary`

```suggestion
func NewCataloger() pkg.Cataloger {
```

syft/pkg/cataloger/binary/cataloger.go
```diff
@@ -0,0 +1,78 @@
...
+	return &binaryCataloger{}
+}
+
+// binaryCataloger is the cataloger responsible for surfacing evidence of a very limited set of binary files,
+// which have been identified by the classifiers. The binaryCataloger is _NOT_ a place to catalog any and every
+// binary, but rather the specific set that has been curated to be important, predominantly related to toolchain-
+// related runtimes like Python, Go, Java, or Node. Some exceptions can be made for widely-used binaries such
+// as busybox.
+type binaryCataloger struct{}
```
wagoodman, line 20:
```suggestion
type Cataloger struct{}
```
same reason as https://github.com/anchore/syft/pull/1336/files#r1035187073

syft/pkg/cataloger/binary/cataloger.go
wagoodman, line 11: as a rule of thumb we should return the concretion not the abstraction
```suggestion
func NewBinaryCataloger() *Cataloger {
```

syft/pkg/cataloger/binary/classifier.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/internal/unionreader"
+	"github.com/anchore/syft/syft/source"
+)
+
+var emptyPURL = packageurl.PackageURL{}
+
+// binaryClassifier is a generic package classifier that can be used to match a package definition
+// to a file that meets the given content criteria of the evidenceMatcher.
+type binaryClassifier struct {
```
wagoodman, line 22: nit: should just be `classifier` since this is within the `binary` package

test/integration/package_deduplication_test.go
```diff
@@ -21,7 +22,7 @@ func TestPackageDeduplication(t *testing.T) {
 	}{
 		{
 			scope:        source.AllLayersScope,
-			packageCount: 172, // without deduplication this would be 618
+			packageCount: 173, // without deduplication this would be 618
```
wagoodman, line 25: why did these package counts change?

wagoodman: nvm -- facepalm

wagoodman (approved):

PR #1343, opened 2022-11-16
chore: clean up linting configuration

spiffcs:
## Summary
Edits to the configuration based on quick team feedback on unused or obsolete linters

### Remove
- dogsled
- goprintffuncname
- nakedret

### Add
- errorlint
- importas
- tparallel 
	- (note: we do not run linters on tests by default, but we can do a small one time cleanup here in a future PR)

### Updates
- Update error fmt and usage of `Is` and `As` to match suggestions from `errorlint`
- Funlen defaults bumped a little to give some room for >,< 1 issues being encountered in certain files

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -41,7 +41,6 @@ linters-settings:
 output:
   uniq-by-line: false
 run:
-  issues-exit-code: 1
   timeout: 10m
 
 # do not enable...
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -11,7 +11,6 @@ linters:
     - asciicheck
     - bodyclose
     - depguard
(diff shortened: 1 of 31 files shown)


.golangci.yaml
```diff
@@ -11,23 +11,23 @@ linters:
     - asciicheck
     - bodyclose
     - depguard
-    - dogsled
```
wagoodman, line 14: nit: for the items removed, mind adding them in the commented out section at the bottom with a quick reason why we aren't using them anymore?

wagoodman (approved):

.golangci.yaml
spiffcs: Yep!

PR #1345, opened 2022-11-16
Fix decoding package groups from cyclonedx sboms

wagoodman:
This PR is work extracted from #1029 specifically relating to comment https://github.com/anchore/syft/pull/1029#issuecomment-1239317225 from @wurstbrot (cc: @patrikbeno). This fixes the decoding of cyclonedx packages to account for information that may be stored in the `group` attribute. However, what work has not been completed yet is considering how that should affect the encoding of a package into the `group` attribute. Without this we cannot pass the encode-decode-encode integration test that attempts to prove stability in our encoding and decoding methods.

### DIFF
--- a/syft/formats/common/cyclonedxhelpers/component.go
+++ b/syft/formats/common/cyclonedxhelpers/component.go
@@ -2,6 +2,7 @@ package cyclonedxhelpers
 
 import (
 	"reflect"
+	"strings"
 
 	"github.com/CycloneDX/cyclonedx-go"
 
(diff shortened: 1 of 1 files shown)


spiffcs:
I'll update the conflicts, but notes here from sync review. We probably want to enhance the `package` struct with a `FormatProperties` attribute that can reserve these format specific values across encode and decode for `spdx/cyclonedx`. I'll give a quick stab at that and post it here for further review.

wagoodman:
This is being superseded by https://github.com/anchore/syft/pull/3853

PR #1347, opened 2022-11-17
1111 clean name bug

### LINKED ISSUE #1111: Don't replace : with - in docker SPDX namespaces
In https://github.com/anchore/syft/pull/649, https://github.com/anchore/syft/blob/main/internal/formats/common/spdxhelpers/document_name.go#L26 was introduced which replaces `:` with `-` in namespaces. This impacts the output name of docker SPDX.
Ideally these docker names like `IMAGE[:TAG|@DIGEST]` would be the same as outputted by syft.

**Why is this needed**:
We are having to manually parse this on ingestion time, back to : so we can correlate docker images with their syft scans.

spiffcs:
## Summary 
Closes #1111

Cleaning the document name was aggressively changing the user inputs to not match users expectations. This PR removes this function from the name generation process as there are currently no designations as to what is vs isn't a valid name:
https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#64-document-name-field

This reverts a previous change that was intended only to sanitize name-spacing URL in the specification.

Tests have been confirmed via snapshot that no namespace changes are affected as of this PR


kzantow (approved):
LGTM -- sorry for giving you a handful of conflicts!

PR #1348, opened 2022-11-18
fix: spdx java checksum correctness

spiffcs:
## Summary
SPDX expects a `ToUpper` checksum algorithm for all package syntax. Previously java would not behave this way and returned `sha1` lowercase causing downstream consumer issues.

- add ToUpper for java packages being converted
- add additional coverage for Golang and niche test cases
- add decoder smoke test with the future notes for developers on e2e conversion coverage


kzantow (approved):
🎉

PR #1361, opened 2022-11-22
fix: return unicode version only letter/num

### LINKED ISSUE #1360: Python cataloging fails to remove some non-version characters from version string
1. Create a `requirements.txt` with the following content:

   ```ini
   argh==0.26.2 \
     --hash=sha256:a9b3aaa1904eeb78e32394cd46c6f37ac0fb4af6dc488daa58971bdc7d7fcaf3 \
     --hash=sha256:e9535b8c84dc9571a48999094fda7f33e63c3f1b74f3e5f3ac0105a58405bb65
   ```

2. Run `syft requirements.txt`. Note the version is `0.26.2 \`, where it should be `0.26.2`

`syft version` output:

```
$ syft version
Application:        syft
Version:            0.62.0
JsonSchemaVersion:  6.0.0
BuildDate:          2022-11-18T19:09:50Z
GitCommit:          da4b2df57640e03f273a2e7e9b04eca40555e139
GitDescription:     v0.62.0
Platform:           darwin/amd64
GoVersion:          go1.18.7
Compiler:           gc
```

spiffcs:
## Summary
Closes #1360 - Trim space fails to cut the end of a string given a blocking `\`

This fix updates the string parser function to only return strings with valid Unicode letter/number.

cc @robcresswell 

### DIFF
--- a/syft/pkg/cataloger/python/parse_requirements.go
+++ b/syft/pkg/cataloger/python/parse_requirements.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"strings"
+	"unicode"
 
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
@@ -48,7 +49,9 @@ func parseRequirementsTxt(_ source.FileResolver, _ *generic.Environment, reader 
 			continue
 		}
 		name := strings.TrimSpace(parts[0])
-		version := strings.TrimSpace(parts[1])
+		version := strings.TrimFunc(parts[1], func(r rune) bool {
+			return !unicode.IsLetter(r) && !unicode.IsNumber(r)
+		})
 		packages = append(packages, newPackageForIndex(name, version, reader.Location))
 	}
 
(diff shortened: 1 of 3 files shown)


kzantow:
@spiffcs yeah, exactly -- I'm mostly wondering if _all_ `requirements.txt` will have this exact split or some _might_ put them on a single line, so if we pre-processed the continued lines, in this case as you say to:
```
argh==0.26.2 --hash=sha256:a9b3aaa1904eeb78e32394cd46c6f37ac0fb4af6dc488daa58971bdc7d7fcaf3 --hash=sha256:e9535b8c84dc9571a48999094fda7f33e63c3f1b74f3e5f3ac0105a58405bb65
```
And then extracted the version, this _may_ make processing more consistent for other cases.

I'm fine with the change as is, though, especially if this suggestion would make it a significantly larger change (e.g. we're always looking for version at the end of a line and now we need to do other stuff).

spiffcs:
I added a case to cover the negative here:
`if a string exists where the line continuation is not used we are still able to select the version`

I'm going to read more about `requirements.txt` to understand the edge cases of what else could be affected if we remove `\\\n` and flatten the continuations. I want to be sure if we go that route that we handle other common line constructs correctly

PR #1365, opened 2022-11-27
Fix parsing of apk databases with large entries

### LINKED ISSUE #1354: Syft finds no apks for some images with apks

```shell
syft -q registry:cgr.dev/chainguard/sdk@sha256:871b75ddd7b91a29ea0ed8695e271f59056c21af6a2f937f9224e81d6030dbec | grep 'apk\s'
```

**What happened**:

No `apk` type packages are found by Syft for this image.

Since the image's `/lib/apk/db/installed` does contain several package entries, Syft should have reported those packages.

I did a small amount of debugging. It seems like Syft does find the `installed` db and starts to find packages, but then hits a scanner error (`ErrTooLong`) and so it returns a `nil` slice of packages.

Package entries in the apk db can be pretty large. I think Syft's current parsing approach isn't able to handle these larger entries, because the entry's byte count exceeds the scanner's buffer size.

This may be solvable by increasing the size of the buffer, but it may be better to try an approach that doesn't require the scanner to hold an entire package entry in the buffer at once. I haven't thought through this idea, but would be happy to chat through it on this issue if that helps!

luhring:
This PR is a suggested solution to #1354. An alternative implementation could have been to increase the scanner buffer size even further, but the chosen implementation avoids the need to find a perfect buffer size since whole entries no longer need to fit within the buffer. Consequently, it also removes the custom configuration applied to the scanner.

This PR also updates existing tests that were testing `parseApkDBEntry` to test `parseApkDB` instead, since `parseApkDBEntry` no longer exists in the suggested implementation.

**Caveat:** The test fixture used to demonstrate the failure and prove the fix is `syft/pkg/cataloger/apkdb/test-fixtures/very-large-entries`. This is the APK installed DB taken from the image where the failure was first noticed. **This file is large (~4.6 MB)!** Let me know if you think it makes sense to approach this test in a different way.

Closes #1354

### DIFF
--- a/syft/pkg/cataloger/apkdb/parse_apk_db.go
+++ b/syft/pkg/cataloger/apkdb/parse_apk_db.go
@@ -3,12 +3,9 @@ package apkdb
 import (
 	"bufio"
 	"fmt"
-	"io"
 	"path"
 	"strconv"
 	"strings"
-
-	"github.com/mitchellh/mapstructure"
 
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -22,157 +19,266 @@ import (
...
-	var entry pkg.ApkMetadata
-	pkgFields := make(map[string]interface{})
+func (f apkField) apply(p *pkg.ApkMetadata, ctx *apkFileParsingContext) {
+	switch f.name {
+	// APKINDEX field parsing
 
-	// We want sane defaults for collections, i.e. an empty array instead of null.
-	pkgFields["D"] = []string{}
-	pkgFields["p"] = []string{}
-	files := make([]pkg.ApkFileRecord, 0)
```
spiffcs, line 82: These are now defaulted in `parseListValue`

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -22,157 +19,266 @@ import (
...
+		// save updated file
+		ctx.files[i] = latest
+	case "Z":
+		i := ctx.indexOfLatestRegularFile
+		latest := ctx.files[i]
+		latest.Digest = processChecksum(f.value)
+
+		// save updated file
+		ctx.files[i] = latest
+	}
```
spiffcs, line 224: Do we want to add a default case here to indicate if f.name is unknown or is that overkill?

spiffcs (commented):
small nit on reading `entryParsingInProgress`

Thank you for the updated parsing logic for `apply` - it's pretty easy to follow how the metadata is built over time and then reset for a new entry.

Thanks also for making sure everything that could be nil is guarded with empty values.

I think I'm happy to merge this and then maybe follow up by lifting the `large` case to the integration level. If you have an image in mind that this is currently failing against that we should add a regression case against that's probably the ideal image.

spiffcs:
cc @anchore/tools for a second review since the parsing logic has changed in more instances than just parsing larger DB entries

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -22,157 +19,266 @@ import (
...
+		apks = append(apks, p)
+	}
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if line == "" {
+			// i.e. apk entry separator
+
+			if entryParsingInProgress {
```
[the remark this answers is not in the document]
luhring: >"If entry parsing In Progress (true)" then the entry is complete and we append?

Yes (almost) — "If entry parsing has been in progress AND we've now encountered an empty line, then call the entry complete _NOW_".

We don't know that parsing has completed for a given package entry until we enter the conditional branch on line 57, where the conditions on both line 53 and line 56 are evaluated to `true`.

The idea here is that each successive line of the file has a field that's applied to the **_same_** package entry unless the entry has been completely parsed, i.e. an empty line has been found. And only in that case would we start applying data to a _new_ package entry (if there's more package data left in the file).

The `if` check on line 56 there is needed because if we encounter an empty line, but we hadn't been working on an entry, we **shouldn't** add anything to the package collection.

Very unopinionated on the variable name, so if you have something that makes more sense here, I can go ahead and update it!

syft/pkg/cataloger/apkdb/parse_apk_db.go
luhring: Yeah good catch, there's a bit of overlap here. Maybe I should adjust `parseListValue` to return `nil` for an empty list, and leave the nil-to-empty-slice conversion in `nilFieldsToEmptySlice`, since only the latter is handling an empty `files` list currently... 🤔 

Also, **question** (that maybe I knew the answer to at one point?):

Why is it necessary to use empty slices (e.g. `[]string{}` instead of `nil`) here in the parsing logic? I know sometimes we've wanted to ensure JSON output uses `[]` instead of `null`, but that's more of a presentation concern, and I'm not remembering other reasons why this is important...

syft/pkg/cataloger/apkdb/parse_apk_db.go
luhring: We can! Is an extra/unknown field something we'd want to warn users about? (And if so, why? Just curious)

luhring:
> Thank you for the updated parsing logic for `apply` - it's pretty easy to follow how the metadata is built over time and then reset for a new entry.

Thanks! ❤️ 

> I think I'm happy to merge this and then maybe follow up by lifting the large case to the integration level. If you have an image in mind that this is currently failing against that we should add a regression case against that's probably the ideal image.

I put one in the linked issue (#1354), we could use that?

syft/pkg/cataloger/apkdb/parse_apk_db.go
spiffcs: Ahh sorry - I added this comment for myself to track where this had been moved to keep track of the diff. 

I have no opinion either way on the overlap `nilFieldsToEmptySlice` and `parseListValue` and defer to you where the default of `[]string{}` should live.

The presentation concern is valid, but I was not around when this was initially written. Another concern could be guarding against any kind of exceptions that may occur downstream for any functions that access these fields after assembly.

syft/pkg/cataloger/apkdb/parse_apk_db.go
spiffcs: You might have more context on the why - How often if at all do new field names get added or removed here? 

If it's super stable and no additional information will be added in the future then it's probably not worth the default case. 

The why from my side is the proactive => "Hey! This new type of entry has been added and syft doesn't handle it" kind of message.

luhring: I gotcha, yeah.

AFAIK the format of the installed DB is very stable. I believe most of the current efforts on `apk` itself are focused on the upcoming _apkv3_, which has an entirely new encoding of this data altogether. The single letter field thing (e.g. `P:`, `Z:`, etc.) is an _apkv2_ thing (not v3) — so while the apkv2 format itself is probably not going to change, eventually apkv2 will probably be phased out. And when apkv3 is eventually rolled out, Syft will need an entirely new parser in order to support it.

So based on your thoughts, I'd suggest not handling extra fields explicitly right now, since I don't think that's very likely to happen?

syft/pkg/cataloger/apkdb/parse_apk_db.go
luhring: Adjusted for the overlap in f9cf2fcd.

The "why `[]string{}`?" question can be a lingering curiosity of mine 😄  — not needed for this PR in particular

spiffcs (approved):

PR #1371, opened 2022-11-29
Recover from bad parsing of golang binary

wagoodman:
The llvm project has some great examples of malformed binaries which currently panic in the go stdlib:

Screen Shot 2022-11-29 at 10 09 34 AM

file: https://github.com/llvm/llvm-project/blob/llvmorg-15.0.6/llvm/test/Object/Inputs/macho-invalid-dysymtab-bad-size

This PR adjusts the golang binary cataloger to at least survive these cases. Additionally more trace logging has been added to help expose what syft is parsing when bad crashes occur.


kzantow (approved):
👍

PR #1377, opened 2022-11-30
feat: add catalog delete

spiffcs:
## Summary

Allows consumers of syft as a library to delete packages from the catalog based on their own filter or gating mechanisms.

### DIFF
--- a/syft/pkg/catalog_test.go
+++ b/syft/pkg/catalog_test.go
@@ -16,6 +16,31 @@ type expectedIndexes struct {
 	byPath map[string]*strset.Set
 }
 
+func TestCatalogDeleteRemovesPackages(t *testing.T) {
+	c := NewCatalog()
+	pkg1 := Package{
+		id:   artifact.ID("pkg1"),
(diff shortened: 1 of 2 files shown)


syft/pkg/catalog_test.go
```diff
@@ -16,6 +16,31 @@ type expectedIndexes struct {
...
+		id:   artifact.ID("pkg3"),
+		Name: "pkg3",
+		Type: RpmPkg,
+	}
+	c.Add(pkg1, pkg2, pkg3)
+
+	c.Delete(pkg1.ID(), pkg2.ID())
+
+	assert.Equal(t, 1, c.PackageCount())
+	assert.Equal(t, []Package{pkg3}, c.Sorted())
```
wagoodman, line 41: I think adding a few more assertions to ensure that the indexes are alright would be helpful. Additionally, some tests that test that the right logic is being applied for real path and virtual path for the location index would be good.

wagoodman (approved):

PR #1383, opened 2022-12-02
Replace core SBOM-creation API with builder pattern

### LINKED ISSUE #1039: Improve cataloger selection capabilities
With #843 (via #1038 ) now you can set the specific set of catalogers to run when generating an SBOM:
```
syft myimg:latest --catalogers go-mod-cataloger
# run the go.mod package cataloger

syft myimg:latest --catalogers python,ruby
# run ALL python and ruby catalogers
```

This is a huge step forward!

Here are the next steps that could be next:
- [x] allow for the user to configure named groups of catalogers (this is implemented in #888  but I'm mentioning it here for completeness)
- [x] allow for cataloger labels which can be selected from and combined, e.g. `python+installed` would select any catalogers that are `python`-based and have the `installed` label.
- [x] expand the set of catalogers beyond packages and allow for any artifact cataloger (e.g. file-metadata, digests, etc...)
- [x] additive only option; something like `syft ... --catalogers +file-metadata` would use the default set of catalogers for the given input type and add file-metadata (today you can only set the entire list of catalogers, not append)

### LINKED ISSUE #1731: Add ability to enable or disable individual catalogers
The ability to enable or disable individual catalogers.

**Why is this needed**:
Today Syft has two different default lists of catalogers: one for image scans and one for directory scans. These have some overlap, but also result in certain catalogers not being run in certain contexts. Syft _also_ has the ability to specify _all the catalogers_ using the `catalogers:` yaml option, `SYFT_CATALOGERS` env var, and `--catalogers` command-line flag, but this requires the user to _know all the catalogers they want to run_. If a user sets this and Syft adds more catalogers, these do not get run until the user updates the Syft config/invocation they are using.

**Additional context**:
I believe it's at least as useful to be able to be able to explicitly _disable_ a cataloger (for example, one causing a problem) or _enable_ a cataloger that isn't part of the default set.

My current proposal is to update the `--catalogers` flag and other configurations to allow each cataloger to be prefixed by a `+` or `-`, which would enable and disable the cataloger, respectively. If all catalogers specified in the configuration are additions or removals, the default set of catalogers is used for the scan while having the set supplemented by the `+` additions and the `-` entries removed.

This should make configuration significantly easier for a large number of use cases.

### LINKED ISSUE #2136: When specify java-cataloger, java-pom-cataloger will also be selected
**What happened**:

I need to scan a directory, but I want to use only image catalogers, so I tried to use `--catalogers` options, like:

```
$syft <path-to-scan> --catalogers=...,java-cataloger,graalvm-native-image-cataloger,...
```

As you can see, I only specified "java-cataloger", but scan result always contained packaged found by "java-pom-cataloger". So I debugged a little bit, and found the problem is at https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/cataloger.go#L186C56-L186C56. According to the function, "java-pom-cataloger" will match to "java-cataloger", so that it is selected as well.

I wonder if the behavior is intentional? If yes, why? @wagoodman 

**Environment**:
- Output of `syft version`: main branch

wagoodman:
Adds a top-level replacement for the syft API. The idea is to allow for encapsulation of more kinds of cataloging without the need to share a data interface. This allows for file-base cataloging and package-based cataloging to share the same approach to selection and configuration.

The existing cataloging functions have been removed, which is why this is a breaking change. I initially attempted to keep both schemes in place, however, the configuration management became ultimately confusing.

This PR adds high-level configuration:
- `syft/cataloging/*.go`: cross-cutting configuration that could affect all catalogers, the artifacts they produce, or add downstream artifacts based on these descriptions. This is a set of configurations NOT capabilities (e.g. behavior, such as catalogers themselves).
- `syft/cataloging/pkgcataloging/*.go`: wires up configurations for all package catalogers
- `syft/cataloging/filecataloging/*.go`: wires up all configurations for file catalogers

This PR removes the existing configurations:
- `syft/pkg/cataloger/config.go` 

From a high-level, the `CreateSBOMConfig` is the entrypoint to all cataloging. The configuration itself describes what should be done. Ultimately all capabilities (file cataloging, pkg cataloging, linux distro identification, and cross-cutting relationship additions) are expressed as "tasks". Tasks act like a facade, similar to the command pattern, and encapsulates pre-configured behavior that ultimately writes to an SBOM.  The notion of "tasks" has not been exported to the public API. 

Secondarily, this PR makes the following adjustments:
- migrates all cross-cutting relationship functions to an internal `relationship` package
- leverages the recent [UI enhancements](https://github.com/anchore/syft/pull/2369) to show catalogers in a tree on the CLI
- migrates some application configurations in a breaking fashion:
  - `exclude-binary-overlap-by-ownership` has been moved to `package.exclude-binary-overlap-by-ownership`
  - `default-image-pull-source` has been moved to `source.image.default-pull-source`

Minimal example of using the new API:
```golang
src := source.New(...)
sbom, err := syft.CreateSBOM(ctx, src, nil)
```
or
```golang
src := source.New(...)
cfg := syft.DefaultCreateSBOMConfig()
sbom, err := syft.CreateSBOM(ctx, src, cfg)
```

Leveraging a little more of the API:
```golang
cfg := syft.DefaultCreateSBOMConfig().
	WithTool("my-tool", "v1.0.2").
	WithParallelism(5).
	WithCatalogerSelection("+sbom-cataloger", "-rpm").
	WithRelationshipsConfig(
		cataloging.RelationshipsConfig{
			FileOwnership:        false,
			FileOwnershipOverlap: false,
			ExcludeBinaryPackagesWithFileOwnershipOverlap: false,
		},
	).
	WithSearchConfig(
		cataloging.SearchConfig{
			Scope: source.SquashedScope,
		},
	).
	WithDataGenerationConfig(
		cataloging.DataGenerationConfig{
			GenerateCPEs:          false,
		},
	).
	WithFilesConfig(
		filecataloging.Config{
			Selection: file.OwnedFilesSelection,
			Hashers: []crypto.Hash{
				crypto.SHA256,
				crypto.SHA1,
			},
		},
	).
	WithPackagesConfig(
		pkgcataloging.Config{
			Golang: golang.CatalogerConfig{
				SearchLocalModCacheLicenses: true,
				SearchRemoteLicenses:        true,
			},
			LinuxKernel: kernel.LinuxKernelCatalogerConfig{
				CatalogModules: true,
			},
			Python: python.CatalogerConfig{
				GuessUnpinnedRequirements: true,
			},
			Java: java.CatalogerConfig{
				ArchiveSearchConfig: cataloging.ArchiveSearchConfig{
					IncludeIndexedArchives:   true,
					IncludeUnindexedArchives: false,
				},
				UseNetwork: true,
			},
		},
	)

sbom, err := syft.CreateSBOM(ctx, src, cfg)
```

Today when the cataloging process is run, the application configuration is captured to show the exact input. I've changed this some to instead capture an API-level construct instead of a construct that is in the `cmd` package. Here is an example of the `syft-json` `descriptor` section:
```json
{
  "name": "syft",
  "version": "v0.99.0",
  "configuration": {
    "catalogers": {
      "requested": {
        "default": [
          "binary"
        ],
        "selection": []
      },
      "used": [
        "binary-cataloger",
        "cargo-auditable-binary-cataloger",
        "dotnet-portable-executable-cataloger",
        "go-module-binary-cataloger"
      ]
    },
    "data-generation": {
      "generate-cpes": true
    },
    "extra": null,
    "files": {
      "hashers": [
        "sha-1",
        "sha-256"
      ],
      "selection": "owned-files"
    },
    "packages": {
      "golang": {...},
      "java": {...},
      "javascript": {... },
      "linux-kernel": {... },
      "python": {...}
    },
    "relationships": {
      "exclude-binary-packages-with-file-ownership-overlap": true,
      "file-ownership": true,
      "file-ownership-overlap": true
    },
    "search": {
      "scope": "squashed"
    }
  }
}

```
<details>
  <summary>click to see all options in an example</summary>

```json
{
  "name": "syft",
  "version": "[not provided]",
  "configuration": {
    "catalogers": {
      "requested": {
        "default": [
          "binary"
        ],
        "selection": []
      },
      "used": [
        "binary-cataloger",
        "cargo-auditable-binary-cataloger",
        "dotnet-portable-executable-cataloger",
        "go-module-binary-cataloger"
      ]
    },
    "data-generation": {
      "generate-cpes": true
    },
    "extra": null,
    "files": {
      "hashers": [
        "sha-1",
        "sha-256"
      ],
      "selection": "owned-files"
    },
    "packages": {
      "golang": {
        "local-mod-cache-dir": "/Users/wagoodman/.local/share/rtx/installs/go/1.21.1/packages/pkg/mod",
        "proxies": [
          "https://proxy.golang.org",
          "direct"
        ],
        "search-local-mod-cache-licenses": false,
        "search-remote-licenses": false
      },
      "java-archive": {
        "include-indexed-archives": true,
        "include-unindexed-archives": false,
        "maven-base-url": "https://repo1.maven.org/maven2",
        "max-parent-recursive-depth": 5,
        "use-network": false
      },
      "javascript": {
        "npm-base-url": "https://registry.npmjs.org",
        "search-remote-licenses": false
      },
      "linux-kernel": {
        "catalog-modules": true
      },
      "python": {
        "guess-unpinned-requirements": false
      }
    },
    "relationships": {
      "exclude-binary-packages-with-file-ownership-overlap": true,
      "file-ownership": true,
      "file-ownership-overlap": true
    },
    "search": {
      "scope": "squashed"
    }
  }
}

```

</details>

This beats the current approach of using the catalogers: https://gist.github.com/wagoodman/57ed59a6d57600c23913071b8470175b

## PRs broken off of this one
- #2363
- #2365
- #2369
- #2370
- #2439
- #2438
- #2437 
- #2436

## Follow up PRs
- https://github.com/anchore/syft/pull/2366

Partially implements #558 

Fixes #2136
Closes #1731 
Closes #1039


tommyknows:
I'm wondering whether it makes sense to have a "global" config for all catalogers, or if it's possible to move the config to a per-cataloger basis. Additionally, why not configure the `CreateSBOM` command directly, instead of having to setup a config struct first?

Some examples / ideas:
```go
syft.CreateSBOM(src, 
    syft.SearchConfig(...),
    // could also leverage config-structs instead of functional options.
    // I don't know how nice it would be to implement such an API though.
    syft.WithDefaultCatalogers(src.Metadata, 
        filedigest.Hash(crypto.SHA256), 
        java.OtherConfigOption(true),
    ),
    syft.FilterCatalogers(userInput),
)

// other example of such an option, on a per-cataloger level instead.
// have a separate method that allows to overwrite a cataloger's config. Could also be embedded directly
// into the `WithDefaultCatalogers` function. I could imagine that this also allows adding additional 
// catalogers through another method on `WithDefaultCatalogers`.
// Could also use functional options.
WithDefaultCatalogers(src.Metadata).Configure(filedigest.Cataloger(filedigest.Config{}))
```

I'm not trying to say that this is any better than what you're proposing, just different. Maybe something to think about :) 
I'm also not super familiar with the current code, so maybe such a refactor would be too big to implement.

Also, let me know if I should have a look at the code :)

wagoodman:
> I'm wondering whether it makes sense to have a "global" config for all catalogers, or if it's possible to move the config to a per-cataloger basis

The goal here is to allow for both, though, in the current implementation there isn't a nice execution path to pass a configuration override to a cataloger. Today there happens to be only one cataloger that needs specific configuration (java) however I think that will be changing over time. _I'll update this PR to account for this._ 🙌 

Why have a global-like configuration? Moving forward it seems very likely that there will be a body of shared configuration options that will be shared amongst multiple catalogers. Instead of requiring the user to feed in configs for each duped field, a single "how-you-search" config is provided, which hydrates other cataloger configs is the current approach.

The high-level "global" config is meant to be declarative (e.g. purely data flags that is used by cataloger to make decisions), where as items on the builder config can be imperative (e.g. a list of cataloger objects, which directly represents behavior).

> `WithDefaultCatalogers(src.Metadata).Configure(filedigest.Cataloger(filedigest.Config{...}))`

I think the one downside with this is that the catalogers get constructed with the `WithDefaultCatalogers()` call and then one cataloger (the filedigest cataloger) gets reconstructed on the `Configure()` call. Ideally catalogers should be constructed once.

> why not configure the CreateSBOM command directly, instead of having to setup a config struct first?

This was the original proposal in #558 , however, it became apparent that having options for a single function in a top-level API would pollute the namespace for other functions that could in the future take options. For example:

```golang
package syft

// cataloging

type CatalogingOption func(*source.Source, *CatalogingConfig) error

func WithDefaultCatalogers(...) CatalogingOption {
   ...
}

func CreateSBOM(s *source.Source, options ...CatalogingOption) (*sbom.SBOM, error) {
   ...
}

// formatting... this is made up just to illustrate a conflict

type FormattingOption func(format.Format, *FormatConfig) error

func WithSomeSPDXOption(...) FormattingOption {
   ...
}

func Format(f format.Format, s sbom.SBOM, options ...FormattingOption) ([]byte, error) {
   ...
}

```

From a type system point of view, this works out great: `CreateSBOM` can only take `CatalogingOption`s and `Format` only takes `FormattingOption`s. 

However, it's not easy to see what options are available for the call you are using while developing. Say I'm in my IDE of choice, And type:
```golang
// assume "src" is a source object
syft.CreateSBOM(src, syft.With...
```
I should see a dropdown in my IDE that has:
```
syft.WithDefaultCatalogers
syft.WithSomeSPDXOption
```
This is because most IDEs will list suggestions only based on lexical names and not by types (though it may order the list by types). This is what I mean by "the namespace get's cluttered quickly".

A builder pattern naturally helps with this, by constraining options that belong to a specific configuration struct with receiver methods on that struct... so the namespace is semantically constrained.

@tommyknows thanks for the feedback / your opinions on this... more eyes helps with the API design 🙌 . I'll incorporate the update for the cataloger-specific configs and report back.

syft/tasks.go
```diff
@@ -0,0 +1,307 @@
...
+		newTaskDescriptor(cfg, php.NewPHPComposerLockCataloger(), declaredTag, directoryTag, packageTag, languageTag, "php", "composer"),
+		newTaskDescriptor(cfg, python.NewPythonIndexCataloger(), declaredTag, directoryTag, packageTag, languageTag, "python"),
+		newTaskDescriptor(cfg, ruby.NewGemFileLockCataloger(), declaredTag, directoryTag, packageTag, languageTag, "ruby", "gem"),
+		newTaskDescriptor(cfg, rust.NewCargoLockCataloger(), declaredTag, directoryTag, packageTag, languageTag, "rust", "cargo"),
+		newTaskDescriptor(cfg, swift.NewCocoapodsCataloger(), declaredTag, directoryTag, packageTag, languageTag, "swift", "cocoapods"),
+
+		// language-specific package declared & installed catalogers
+		newTaskDescriptor(cfg, python.NewPythonPackageCataloger(), directoryTag, installedTag, imageTag, packageTag, languageTag, "python"),
+		newTaskDescriptor(cfg, golang.NewGoModuleBinaryCataloger(), directoryTag, installedTag, imageTag, packageTag, languageTag, "go", "golang", "gomod", "binary"),
+		newTaskDescriptor(cfg, java.NewJavaCataloger(cfg.Java()), directoryTag, installedTag, imageTag, packageTag, languageTag, "java", "maven"),
```
wagoodman, line 131: note: we should consider renaming the `java-cataloger` to `java-archive-cataloger`

cmd/syft/cli/options/pkg.go
```diff
@@ -1,23 +1,18 @@
...
-		Cataloger: scope{
-			Enabled: true,
-			Scope:   c.Scope.String(),
-		},
+func defaultPkg() packageConfig {
+	c := cataloging.DefaultArchiveSearchConfig()
+	return packageConfig{
+		SearchIndexedArchives:           c.IncludeIndexedArchives,
+		SearchUnindexedArchives:         c.IncludeUnindexedArchives,
+		ExcludeBinaryOverlapByOwnership: true,
```
wagoodman, line 16: **open question**: this logically affects packages, so belongs in the package config. However, in terms of when it's needed, this is really used after packages are created and uses the relationships to determine how to prune packages (so functionally belongs in the relationships config).

These two things are at odds and are also at odds in app config vs API config.

What do folks think about this?

cmd/syft/cli/cli.go
```diff
@@ -91,8 +91,10 @@ func create(id clio.Identification, out io.Writer) (clio.Application, *cobra.Com
 		cranecmd.NewCmdAuthLogin(id.Name), // syft login uses the same command as crane
 	)
 
-	// explicitly set Cobra output to the real stdout to write things like errors and help
-	rootCmd.SetOut(out)
+	// note: we would direct cobra to use our writer explicitly with rootCmd.SetOut(out) , however this causes
+	// deprecation warnings to be shown to stdout via the writer instead of stderr. This is unfortunate since this
+	// does not appear to be the correct behavior on cobra's part https://github.com/spf13/cobra/issues/1708 .
+	// In the future this functionality should be restored.
```
wagoodman, line 97: I wanted to call this change out explicitly...

README.md
```diff
@@ -155,69 +155,114 @@ If docker is not present, then the Podman daemon is attempted next, followed by
...
+Catalogers can be referenced in two different ways:
+- *by name*: the exact cataloger name (e.g. `java-pom-cataloger` or `java-archive-cataloger`)
+- *by tag*: a tag that is associated with a cataloger (e.g. `java`)
+
+Syft can take lists of references on the CLI or in the application configuration to define which catalogers to use.
+
+You can **Set** the list of catalogers explicitly to use with the `--override-default-catalogers` CLI flag, accepting a comma-separated list of cataloger names or tags.
+
+You can also **Add** to, **remove** from, or **sub-select** catalogers to use within the default set of catalogers by using the `--select-catalogers` CLI flag.
+  - To sub-select catalogers simply provide a tag (e.g. `--select-catalogers TAG`). This is always interpreted from the default set of catalogers. If used, the union of these catalogers becomes the new default set.
```
willmurphyscode, line 193: An example would really help here.

README.md
```diff
@@ -468,21 +513,18 @@ output: "syft-table"
 # same as -q ; SYFT_QUIET env var
 quiet: false
```
willmurphyscode, line 514: Is the yaml section of the README generated or programmatically validated? If not, should it be?

README.md
willmurphyscode: (I didn't see the examples below before I posted this.)

README.md
```diff
@@ -155,69 +155,114 @@ If docker is not present, then the Podman daemon is attempted next, followed by
...
+- `all-files`: capture all files from the search space
+- `owned-files`: capture only files owned by packages (default)
+- `no-files`: disable capturing any file information
+
+
+### Package cataloger selection
+
+#### Concepts
+
+Syft uses different sets of catalogers depending on *what is being scanned*: a container image or a directory on disk.
```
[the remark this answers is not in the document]
wagoodman:
```suggestion
Syft uses different sets of catalogers depending on *what is being scanned*: a container image or a files on the host. 
```

As well as further on changing to:
```
The catalogers used for files on the host will look for installed software as well as declared dependencies that are not necessarily installed
```

How about something like this?

README.md
wagoodman: when we started doing this we didn't have a way to generate this. Today we do, but we don't have any automation to patch the README.md in the right section with the correct information. (down for adding that, but it should be it's own issue/PR)

README.md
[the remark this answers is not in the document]
kzantow: Maybe something more like:

> [!IMPORTANT]  
> Syft uses different sets of catalogers when files directly than it does when scanning images

?

syft/file/selection.go
```diff
@@ -0,0 +1,9 @@
+package file
+
+const (
+	NoFilesSelection    Selection = "no-files"
+	OwnedFilesSelection Selection = "owned-files"
```
willmurphyscode, line 5: Does `owned-files` mean "files owned by a package?" (Out of context I thought it meant files owned by a group or user for a second.)

syft/cataloging/relationships.go
```diff
@@ -0,0 +1,30 @@
+package cataloging
+
+type RelationshipsConfig struct {
+	FileOwnership                                 bool `yaml:"file-ownership" json:"file-ownership" mapstructure:"file-ownership"`
```
willmurphyscode, line 4: `FileOwnership` and `FileOwnershipOverlap` don't seem to mean anything by themselves. What does it mean to say `file-ownership: true`? (I think it means, "Syft should report relationships of the type where a package owns a file" and `file-ownership-overlap: true` means, "Syft should report relationships where multiple packaages own a file?" but I'm not sure I'm right.)

syft/cataloging/relationships.go
```diff
@@ -0,0 +1,30 @@
...
+	FileOwnershipOverlap                          bool `yaml:"file-ownership-overlap" json:"file-ownership-overlap" mapstructure:"file-ownership-overlap"`
+	ExcludeBinaryPackagesWithFileOwnershipOverlap bool `yaml:"exclude-binary-packages-with-file-ownership-overlap" json:"exclude-binary-packages-with-file-ownership-overlap" mapstructure:"exclude-binary-packages-with-file-ownership-overlap"`
```
willmurphyscode, line 6: I think this name is super long, but it's clearer than the names above. It means, "If Syft finds a binary, say `java`, and also an RPM that installed `java`, it should only report the RPM," right?

cmd/syft/cli/options/pkg.go
willmurphyscode: I thought this was already in the relationships config? I'm missing something.

cmd/syft/cli/options/relationships.go
```diff
@@ -0,0 +1,13 @@
+package options
+
+type relationshipsConfig struct {
+	FileOwnership        bool `mapstructure:"file-ownership" json:"file-ownership" yaml:"file-ownership"`
```
willmurphyscode, line 4: Same comment as the other place this is used. What does `file-ownership: true` or `file-ownership: false` mean?

cmd/syft/cli/options/source.go
```diff
@@ -12,16 +12,40 @@ type sourceConfig struct {
...
+			DefaultPullSource: "",
+		},
+	}
+}
+
+func (c imageSource) PostLoad() error {
+	return checkDefaultSourceValues(c.DefaultPullSource)
+}
+
+var validDefaultSourceValues = []string{"registry", "docker", "podman", ""}
```
willmurphyscode, line 41: What does `""` mean as a default source?

cmd/syft/cli/options/catalog.go
```diff
@@ -6,41 +6,47 @@ import (
...
-	Parallelism                     int          `yaml:"parallelism" json:"parallelism" mapstructure:"parallelism"`                                                                         // the number of catalog workers to run in parallel
-	DefaultImagePullSource          string       `yaml:"default-image-pull-source" json:"default-image-pull-source" mapstructure:"default-image-pull-source"`                               // specify default image pull source
-	BasePath                        string       `yaml:"base-path" json:"base-path" mapstructure:"base-path"`                                                                               // specify base path for all file paths
-	ExcludeBinaryOverlapByOwnership bool         `yaml:"exclude-binary-overlap-by-ownership" json:"exclude-binary-overlap-by-ownership" mapstructure:"exclude-binary-overlap-by-ownership"` // exclude synthetic binary packages owned by os package files
+	// high-level cataloger configuration
+	Catalogers        []string            `yaml:"-" json:"catalogers" mapstructure:"catalogers"` // deprecated and not shown in yaml output
```
willmurphyscode, line 28: Does this also keep it from being _parsed_ from yaml output?

cmd/syft/cli/options/catalog.go
```diff
@@ -77,6 +158,20 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
...
+		}
+	} else {
+		panic("unable to mark flags as deprecated")
+	}
+
+	flags.StringArrayVarP(&cfg.DefaultCatalogers, "override-default-catalogers", "",
+		"set the catalogers to use explicitly as the final cataloger set (defaults to 'image' or 'directory' dynamically depending on the scan source)")
+
+	flags.StringArrayVarP(&cfg.SelectCatalogers, "select-catalogers", "",
+		"add to, remove from, or select a subset of catalogers from the final cataloger set")
```
willmurphyscode, line 173: I'd like to talk about the wording of this a bit. Maybe "*for* the final cataloger set" instead of "*from* the final cataloger set"?

cmd/syft/cli/options/catalog.go
```diff
@@ -91,91 +186,45 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
...
-				},
-				cfg.Java.MaxParentRecursiveDepth),
-		Javascript: javascriptCataloger.DefaultCatalogerConfig().
-			WithSearchRemoteLicenses(cfg.Javascript.SearchRemoteLicenses).
-			WithNpmBaseURL(cfg.Javascript.NpmBaseURL),
-		Python: pythonCataloger.CatalogerConfig{
-			GuessUnpinnedRequirements: cfg.Python.GuessUnpinnedRequirements,
-		},
-		ExcludeBinaryOverlapByOwnership: cfg.ExcludeBinaryOverlapByOwnership,
+	if len(cfg.Catalogers) > 0 && (len(cfg.DefaultCatalogers) > 0 || len(cfg.SelectCatalogers) > 0) {
```
willmurphyscode, line 201: I'd love to see some parentheses so I don't have to remember operator precedence.

internal/relationship/by_file_ownership.go
```diff
@@ -29,19 +32,35 @@ type ownershipByFilesMetadata struct {
...
+func byFileOwnershipOverlap(catalog *pkg.Collection) []artifact.Relationship {
 	var relationships = findOwnershipByFilesRelationships(catalog)
 
 	var edges []artifact.Relationship
 	for parentID, children := range relationships {
 		for childID, files := range children {
 			fs := files.List()
 			sort.Strings(fs)
+
+			parent := catalog.Package(parentID) // TODO: this is potentially expensive
```
willmurphyscode, line 58: I'm reviewing commit-by-commit. Are these TODOs addressed later?

cmd/syft/cli/commands/packages.go
```diff
@@ -205,51 +207,228 @@ func getSource(opts *options.Catalog, userInput string, filters ...func(*source.
 }
 
 func generateSBOM(id clio.Identification, src source.Source, opts *options.Catalog) (*sbom.SBOM, error) {
-	tasks, err := eventloop.Tasks(opts)
+	s, err := syft.CreateSBOM(src, opts.ToSBOMConfig(id))
 	if err != nil {
+		expErrs, err := filterExpressionErrors(err, 0)
+		notifyExpressionErrors(expErrs)
```
kzantow, line 213: The current output looks something like this (the section "Found 1 invalid cataloger selection expression" is purple while the error section starting with "1 error occurred:" is red):
```
$ syft alpine:latest --override-default-catalogers js
...
Found 1 invalid cataloger selection expression:            
                                                           
 ❖ Given expression "--override-default-catalogers js"     
   However, "js" is not a recognized cataloger name or tag.
                                                           
1 error occurred:
        * invalid cataloger selection expression provided
```
I think it _may_ be more clear if the errors remain just as they are and we supplement the messaging to the user with suggestions in the purple section. I'd suggest something along these lines (the purple section is the Hints or Suggestions, and the red is the 1 error occurred, like it is today:

```
$ syft alpine:latest --override-default-catalogers js
...
1 error occurred:
        * "js" is not a recognized cataloger name or tag. default catalogers must be specified by name or tag
```

```
$ syft alpine:latest --select-catalogers sbom-cataloger
...
Hints:
        - to add the sbom-cataloger to the defaults use: --select-catalogers +sbom-cataloger

1 error occurred:
        * "sbom-cataloger" is not a known cataloger tag. filtering catalogers must use tags
```

```
$ syft alpine:latest --select-catalogers +javascript
...
Suggestions:
        - to add all javascript catalogers to the defaults use: --override-default-catalogers default,javascript

1 error occurred:
        * "javascript" is not a recognized cataloger name. adding catalogers must be specified by name
```

... as discussed, this would also mean all the code to remove/alter the errors can go away and just check for the presence of the known errors to populate the optional suggestion section.

wagoodman:
@kzantow I took your feedback regarding:
- the top-level API taking strings with embedded operators instead of rich objects
- `WithCataloger()` should result in the passed catalogers being used by default (regardless of the selection)

The changes have been incorported in https://github.com/anchore/syft/pull/1383/commits/098fbd71cad8671f6c2146e70e836c20fe4eb18c

The specifics are:
- a new `pkgcataloging.SelectionRequest` has been created and replaces the existing `task.SelectionRequest` and is expected to be passed to the `CreateSBOMConfig` if a non-default option is used. Empty `SelectionRequest` results in good default behavior.
- `WithCataloger()` has been amended to always include catalogers in the selection in which no tags were provided.

cmd/syft/cli/cli.go
wagoodman: fixed in another PR

cmd/syft/cli/options/catalog.go
wagoodman: I'll take @kzantow s wording here

cmd/syft/cli/options/catalog.go
wagoodman: regarding sets of operators for || and &&, this is already done -- am I missing something here? (maybe an extra parenthesis around the first check?)

cmd/syft/cli/options/catalog.go
```diff
@@ -50,13 +56,88 @@ var _ interface {
...
-		FileContents:                    defaultFileContents(),
-		Source:                          defaultSourceCfg(),
-		Parallelism:                     1,
-		ExcludeBinaryOverlapByOwnership: true,
+		Scope:         source.SquashedScope.String(),
+		Package:       defaultPkg(),
+		LinuxKernel:   defaultLinuxKernel(),
+		File:          defaultFile(),
+		Relationships: defaultRelationships(),
+		Source:        defaultSourceCfg(),
```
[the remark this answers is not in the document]
wagoodman: I'll make them consistent with other spots:`Config`

cmd/syft/cli/options/source.go
wagoodman: this is the default value and defers to the underlying lib to control this behavior

cmd/syft/cli/options/output_file.go
```diff
@@ -13,13 +17,13 @@ var _ interface {
 
 // Deprecated: OutputFile supports the --file to write the SBOM output to
 type OutputFile struct {
-	Enabled bool   `yaml:"-" json:"-" mapstructure:"-"`
-	File    string `yaml:"file" json:"file" mapstructure:"file"`
+	Enabled    bool   `yaml:"-" json:"-" mapstructure:"-"`
+	LegacyFile string `yaml:"-" json:"-" mapstructure:"legacyFile"`
```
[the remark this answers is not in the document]
wagoodman: This is the same as the already deprecated `--file` option:
```
--file string              file to write the default report output to (default is STDOUT) (DEPRECATED: use: output)
```

This will break any user configuration that is using `file:` . Given that this is an ideal time to make this breaking change and there isn't a really good semantic equivalent for file cataloging configuration, I think this is the right time to make this breaking change.

syft/cataloging/data_generation.go
```diff
@@ -0,0 +1,23 @@
...
+
+type DataGenerationConfig struct {
+	GenerateCPEs          bool `yaml:"generate-cpes" json:"generate-cpes" mapstructure:"generate-cpes"`
+	GuessLanguageFromPURL bool `yaml:"guess-language-from-purl" json:"guess-language-from-purl" mapstructure:"guess-language-from-purl"`
+}
+
+func DefaultDataGenerationConfig() DataGenerationConfig {
+	return DataGenerationConfig{
+		GenerateCPEs:          true,
+		GuessLanguageFromPURL: true,
```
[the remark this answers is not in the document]
wagoodman: when we generate a PURL for a package, we use that as a way to fill in the `language` field on the package (when this setting is enabled). This doesn't mean that there are purls-as-input.

syft/cataloging/archive_search.go
```diff
@@ -0,0 +1,23 @@
...
+
+type ArchiveSearchConfig struct {
+	IncludeIndexedArchives   bool `yaml:"include-indexed-archives" json:"include-indexed-archives" mapstructure:"include-indexed-archives"`
+	IncludeUnindexedArchives bool `yaml:"include-unindexed-archives" json:"include-unindexed-archives" mapstructure:"include-unindexed-archives"`
+}
+
+func DefaultArchiveSearchConfig() ArchiveSearchConfig {
+	return ArchiveSearchConfig{
+		IncludeIndexedArchives:   true,
+		IncludeUnindexedArchives: false,
```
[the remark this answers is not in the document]
wagoodman: an example of an indexed archive is zip, since there is a table at the end of the file that can be used to quickly search for matching paths that are contained within the zip without needing to decompress anything. A tar.gz is an example of an unindexed archive, where we need to decompress the contents and seek through the entire file (over each tar header) to understand what paths are contained within the archive. 

This distinction in the configuration allows for "speedy" vs "deep" archive searches. This is not supported by most catalogers yet.

syft/file/selection.go
wagoodman: > Does owned-files mean "files owned by a package?"

correct 👍

cmd/syft/cli/options/pkg.go
wagoodman: there is a CLI-level configuration and an API-level configuration -- currently this option is on the Packages config on the CLI-level configuration and the Relationships config in the API-configuration. There seems to be logic to reason about it either way. Consistency is better than correctness in my opinion (if this is a situation where they are really mutually exclusive... but I'm not certain if that's true). Looking for options here.

syft/cataloging/relationships.go
wagoodman: Ultimately this configuration holds which kinds of relationships should be captured or not. Happy for suggestions on better names!

> `file-ownership: true` ... means, "Syft should report relationships of the type where a package owns a file" 

correct 👍 

> `file-ownership-overlap: true` means, "Syft should report relationships where multiple packages own a file?"

correct 👍

syft/cataloging/relationships.go
wagoodman: yeah, this name is terrible... but couldn't find a shorter name that wasn't potentially confusing. Side note: this is the option that is at odds between the Relationships and Packages config in the CLI/API levels .

cmd/syft/cli/commands/packages.go
wagoodman: I've updated the code with the following changes:
- original errors are persisted, which makes the code much easier to follow
- only suggestions that are actionable are shown, the original errors will enumerate any non-actionable issues

I didn't rework the copy to be exactly what you had above. Here's an example:


```
❯ syft alpine:3.12 --override-default-catalogers image --select-catalogers javascript --select-catalogers python,sbom-cataloger --select-catalogers +js,+javascript
...

Suggestions:                                                                                              
                                                                                                          
 ❖ Given expression "--select-catalogers sbom-cataloger"                                                  
   However, names are not allowed with this operation (must use tags).                                    
   It seems like you are intending to add a cataloger in addition to the default set.                     
   ... Did you mean "--select-catalogers +sbom-cataloger" instead?                                        
                                                                                                          
 ❖ Given expression "--select-catalogers javascript"                                                      
   However, tags are not allowed with this operation (must use exact names).                              
   Adding groups of catalogers may result in surprising behavior (create inaccurate SBOMs).               
   If you are certain this is what you want to do, use "--override-default-catalogers javascript" instead.
                                                                                                          
                                                                                                          
3 errors occurred:
        * invalid expression: "sbom-cataloger": names are not allowed with this operation (must use tags)
        * invalid expression: "javascript": tags are not allowed with this operation (must use exact names)
        * invalid expression: "js": unknown name or tag given

```

wagoodman:
TODO: consider using `context.Context` for the top level entrypoint

syft/cataloging/pkgcataloging/config.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/anchore/syft/syft/pkg/cataloger/java"
+	"github.com/anchore/syft/syft/pkg/cataloger/javascript"
+	"github.com/anchore/syft/syft/pkg/cataloger/kernel"
+	"github.com/anchore/syft/syft/pkg/cataloger/python"
+)
+
+type Config struct {
+	Golang      golang.CatalogerConfig            `yaml:"golang" json:"golang" mapstructure:"golang"`
+	JavaArchive java.ArchiveCatalogerConfig       `yaml:"java-archive" json:"java-archive" mapstructure:"java-archive"`
+	Javascript  javascript.CatalogerConfig        `yaml:"javascript" json:"javascript" mapstructure:"javascript"`
```
wagoodman, line 14: from @willmurphyscode consider using `JavaScript`

syft/create_sbom_config.go
```diff
@@ -0,0 +1,312 @@
...
+			return nil, nil, fmt.Errorf("unable to determine default cataloger tag: %w", err)
+		}
+
+		if defaultTag != "" {
+			c.CatalogerSelection.DefaultNamesOrTags = append(c.CatalogerSelection.DefaultNamesOrTags, defaultTag)
+		}
+
+		c.CatalogerSelection.RemoveNamesOrTags = replaceDefaultTagReferences(defaultTag, c.CatalogerSelection.RemoveNamesOrTags)
+		c.CatalogerSelection.SubSelectTags = replaceDefaultTagReferences(defaultTag, c.CatalogerSelection.SubSelectTags)
+	}
```
wagoodman, line 237: comment from @willmurphyscode this should probably be decomposed into a separate function (since it's a lower level of abstraction)

wagoodman (commented):
from @willmurphyscode  : (another issue/PR) add `syft config` (or similar) to help get from an SBOM `.descriptor` to a real application configuration. This would help with the transition for the new descriptor block where folks can copy-paste that as-is today and have a valid app config.

syft/create_sbom.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/anchore/syft/internal/sbomsync"
+	"github.com/anchore/syft/internal/task"
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/event/monitor"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/sbom"
+	"github.com/anchore/syft/syft/source"
+)
+
+func CreateSBOM(src source.Source, cfg CreateSBOMConfig) (*sbom.SBOM, error) {
```
wagoodman, line 23: from a conversation with @willmurphyscode 
```suggestion
func CreateSBOM(src source.Source, cfg *CreateSBOMConfig) (*sbom.SBOM, error) {
```

maybe move to a pointer here? then folks can use `nil` and we'd call `DefaultCreateSBOMConfig` on their behalf. There may be good alternatives here.

syft/file/selection.go
```diff
@@ -0,0 +1,9 @@
...
+	AllFilesSelection   Selection = "all-files"
+)
+
+type Selection string
```
wagoodman, line 9: from @willmurphyscode : for the future, we could add an options for "non-owned" files... create an issue for this

syft/create_sbom_config.go
```diff
@@ -0,0 +1,312 @@
...
+// user-provided catalogers (without the default syft-provided catalogers).
+func (c CreateSBOMConfig) WithoutCatalogers() CreateSBOMConfig {
+	c.packageTaskFactories = nil
+	c.persistentPackageTaskFactories = nil
+	return c
+}
+
+// WithCatalogers allows for adding user-provided catalogers to the final set of catalogers that will always be run
+// regardless of the source type or any cataloger selections provided.
+func (c CreateSBOMConfig) WithCatalogers(catalogers ...pkg.Cataloger) CreateSBOMConfig {
```
wagoodman, line 125: from @willmurphyscode : consider removing this and adding to the signature for `CreateSBOM()`. This would make the nuance a little more clear. This also implies `WithCataloger ` would not make a distinction between 0-tags (tags are required?).

wagoodman: consider this as well:
```
WithCataloger(cat pkg.Cataloger, forceEnabled bool, tags ...string)
```

this is harder to misuse given that the persistence vs selectability of the cataloger is explicitly given.

wagoodman: another approach, we could make this a rich struct that has all of these options.
```golang
type CatalogerConfig struct {
  Cataloger pkg.Cataloger
  ForceEnabled bool
  Tags []string
}

func WithCatalogers(CatalogerConfig ...) CreateSBOMConfig {

}

// maybe not?
func WithSelectableCataloger(pkg.Cataloger, tags []string) CreateSBOMConfig {

}

func WithAlwaysEnabledCatalogers(pkg.Cataloger ...) CreateSBOMConfig {

}
```

wagoodman: I went with this last option

syft/create_sbom.go
```diff
@@ -20,7 +21,10 @@ import (
 	"github.com/anchore/syft/syft/source"
 )
 
-func CreateSBOM(src source.Source, cfg CreateSBOMConfig) (*sbom.SBOM, error) {
+func CreateSBOM(ctx context.Context, src source.Source, cfg *CreateSBOMConfig) (*sbom.SBOM, error) {
```
wagoodman, line 24: missing docstring, should at least contain hint for `nil` config for default behavior

syft/create_sbom_config.go
```diff
@@ -30,12 +31,12 @@ type CreateSBOMConfig struct {
...
 
-	packageTaskFactories           task.PackageTaskFactories // syft default + user-provided catalogers with tags to select with
-	persistentPackageTaskFactories task.PackageTaskFactories // user-provided catalogers without tags to select with
+	packageTaskFactories       task.PackageTaskFactories
+	packageCatalogerReferences []pkgcataloging.CatalogerReference
 }
 
-func DefaultCreateSBOMConfig() CreateSBOMConfig {
-	return CreateSBOMConfig{
+func DefaultCreateSBOMConfig() *CreateSBOMConfig {
```
wagoodman, line 38: revisit convention for `Default*`, possibly move to `NewWithDefaults()`

internal/task/file_tasks.go
```diff
@@ -0,0 +1,102 @@
...
+						continue
+					}
+					if c, ok := r.To.(file.Coordinates); ok {
+						coordinates = append(coordinates, c)
+					}
+				}
+			}
+		})
+
+		result, err := digestsCataloger.Catalog(resolver, coordinates...)
```
wagoodman, line 46: from @willmurphyscode , blocking: we need to explicitly pass all coordinates, since there is no guarantee to have any results from a owned-files indication

internal/sbomsync/builder.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Builder provides a simple facade for simple additions to the SBOM
+type Builder interface {
+	// nodes
+
+	AddPackages(...pkg.Package)
+	AddFileMetadata(file.Coordinates, file.Metadata)
+	AddFileDigests(file.Coordinates, ...file.Digest)
+	AddFileContents(file.Coordinates, string)
+	AddFileLicenses(file.Coordinates, ...file.License)
```
wagoodman, line 26: delete these, they are not used

syft/create_sbom_config.go
wagoodman: I think this would be a larger change -- there are ~40 `Default*` functions using this convention. If we want to change course on this convention we should break that into another PR.

internal/task/file_tasks.go
wagoodman: I'm going to fix the functional problem in this PR, but to address the signature and generator issue I really should break that into a separate PR that I follow up with after this PR

syft/file/selection.go
wagoodman: I've also incorporated the additional verbal comment about the names of these values and variables... a recent commit adjusts these

internal/task/file_tasks.go
wagoodman: I think this will have some play into the solution here https://github.com/anchore/syft/issues/2487

cmd/syft/cli/options/pkg.go
wagoodman: we'll leave this as is

syft/file/selection.go
wagoodman: renamed 👍

syft/cataloging/data_generation.go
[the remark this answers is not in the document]
wagoodman: I'll remove the configurable but not the functionality for now (since it was here before this PR)

internal/task/file_tasks.go
```diff
@@ -100,3 +74,64 @@ func NewFileMetadataCatalogerTask(selection file.Selection) Task {
...
+
+		return nil
+	}
+
+	return NewTask("file-content-cataloger", fn)
+}
+
+// TODO: this should be replaced with a fix that allows passing a coordinate or location iterator to the cataloger
+// Today internal to both cataloger this functions differently: a slice of coordinates vs a channel of locations
+func coordinatesForSelection(selection file.Selection, builder sbomsync.Builder) ([]file.Coordinates, bool) {
```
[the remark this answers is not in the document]
wagoodman: added

cmd/syft/cli/options/relationships.go
wagoodman: I've renamed by prepepending `package` to each element and adding descriptions.

syft/cataloging/relationships.go
wagoodman: TODO: write and example in the docs

cmd/syft/cli/options/catalog.go
wagoodman: break this up into other variables

willmurphyscode (approved):
Thanks for all the careful thought in making the API and configs easier to use for the future.

Issue #1391, opened 2022-12-07
syft produces different output when run with sudo


Hritik14:
**Please provide a set of steps on how to reproduce the issue**
On an ``arm`` machine (Technically, any machine other than x86 will do).
```
syft rust:1.42-slim-buster | grep x86
 ✔ Parsed image
 ✔ Cataloged packages      [113 packages]
binutils-x86-64-linux-gnu  2.31.1-16               deb
```
```
sudo syft rust:1.42-slim-buster | grep x86
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [111 packages]

sudo syft rust:1.42-slim-buster | grep aarch
 ✔ Pulled image
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [111 packages]
binutils-aarch64-linux-gnu  2.31.1-16               deb
```

**What happened**:
``syft`` running with ``sudo`` yields different results than without sudo.

**What you expected to happen**:
Both runs should have same output.

**Anything else we need to know?**:
When run with ``sudo``,  docker daemon is available and is used. Without sudo, syft uses OCI registry via stereoscope.
Stereoscope takes the following approach while determining image pull methodology:
```go
// DetermineDefaultImagePullSource takes an image reference string as input, and
// determines a Source to use to pull the image. If the input doesn't specify an
// image reference (i.e. an image that can be _pulled_), UnknownSource is
// returned. Otherwise, if the Docker daemon is available, DockerDaemonSource is
// returned, and if not, OciRegistrySource is returned.
func DetermineDefaultImagePullSource(userInput string) Source {
```
Source: https://github.com/anchore/stereoscope/blob/3b80d983223f6e6fc2d33b0ffa003d30268418e9/pkg/image/source.go#L134-L139

and because there are inconsistencies between OCI source and docker daemon in stereoscope when respective ``Providers`` are run without supplying a ``platform``, it bubbles out to syft yielding different images (one x86 and other arm) and thus different SBOMs.
See:
- https://github.com/anchore/stereoscope/issues/149

This can be avoided by having a correct default value for ``platform``

**Environment**:
- Output of `syft version`:
```
Application:        syft
Version:            0.62.3
JsonSchemaVersion:  6.0.0
BuildDate:          2022-11-30T19:01:12Z
GitCommit:          247b054ab56f848c477495218754166272f470b5
GitDescription:     v0.62.3
Platform:           linux/arm64
GoVersion:          go1.18.8
[...]

PR #1394, opened 2022-12-07
Use build architecture as default platform

Hritik14:
One possible solution to https://github.com/anchore/syft/issues/1391

Build is taking a lot of time, **Not yet tested**

### DIFF
--- a/internal/config/application.go
+++ b/internal/config/application.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"path"
 	"reflect"
+	"runtime"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


Hritik14:
Can someone approve the workflows ?

wagoodman:
Hey, I've tried this out, however, if we go this direction we still need to change some behavior. Specifically, specifying a platform override will break input sources which do not allow for specifying platform:
```
        2023/01/12 15:15:12 error during command execution: 1 error occurred:
                * could not fetch image "test-fixtures/cache/stereoscope-fixture-image-secrets-6dbc1b7867bf6e50c7c4f681fb9d0ac9af3fd291aae2bb643f7d268789f85c77.tar": specified platform="linux/amd64" however image source="DockerTarball" does not support selecting platform

```
(the CLI tests show this)

Thanks for doing all of the upfront work with https://github.com/anchore/stereoscope/issues/149 , https://github.com/anchore/syft/issues/1391 , and this PR. I can make a few mods to this and see how it work as a PoC?

wagoodman:
On further thought, I don't think it makes sense to use the user override here to specify a default platform. I think it makes more sense to put in a fix in stereoscope conditionally based on if the user provided a platform or not (and if not, set the default based off of the host arch).

I'll open up a PR in stereoscope shortly.

wagoodman:
See https://github.com/anchore/stereoscope/pull/152

Issue #1401, opened 2022-12-13
Syft does not generate valid SPDX documents


rnjudge:
**Please provide a set of steps on how to reproduce the issue**
When I ran Syft on three minimal container base images (`photon:3.0`, `debian:latest` and `alpine:latest`), none of the output from Syft validated in tag-value or json output (using https://tools.spdx.org/app/validate/ to validate).

**What happened**:
Errors seen when validating the JSON output include:

```
[instance type (null) does not match any allowed primitive type 
(allowed: ["array"]) for {"pointer":"/files/<*0..n*>/checksums"}

Document must have at least one relationship of type DOCUMENT_DESCRIBES]
```

Errors seen for tag-value documents include:

```
[Package at line 25461 invalid: Warning: Found analyzed files for package libc-bin when analyzedFiles is set to false.,
Package at line 25461 invalid: Missing required SHA1 hashcode value for <insert_filename_here>
```

**What you expected to happen**:
SPDX documents generatd by Syft should be valid SPDX documents.

kzantow:
I'm pretty sure this is a duplicate of: https://github.com/anchore/syft/issues/1396 -- please reopen if it's not!

rnjudge:
This issue includes more than just the files analyzed issue mentioned in #1396. It also includes issues with the formatting of the JSON format and missing fields in the JSON format.

rnjudge:
I'm not able to re-open this issue.. but I think it should remain open :)

kzantow:
Reopening this for the `null` array issue.

spiffcs:
@rnjudge if you want to check out #1404 I've added a path that generates valid SPDX documents for the images you listed

spiffcs:
https://github.com/spdx/tools-golang/pull/176 <-- This issue is blocked until this is merged upstream with the correct field for the tools-golang library

spiffcs:
After https://github.com/anchore/syft/pull/1418 is merged you should have success generating VALID spdx using the commands from the PR.

The digest cataloger that produces sha1s for regular files is not currently default syft behavior, but can be enabled with the following env:
```
SYFT_FILE_METADATA_CATALOGER_ENABLED=true SYFT_FILE_METADATA_DIGESTS=sha1 
```

PR #1404, opened 2022-12-14
bug: spdx checksum empty array; allow syft to generate SHA1 for spdx-tag-value documents

spiffcs:
## Summary

Syft currently generates SPDX documents that do not contain SHA1 digests for the files by default.

For a user to obtain a document that meets the current specification the following has to be true:
- `app.FileMetadata.Cataloger.Enabled`
- `app.FileMetadata.Digests` must add `sha1` to the list of desired digests (currently defaults to sha256)

Here is a link to the reference that shows this requirement:
https://spdx.github.io/spdx-spec/v2.3/file-information/#84-file-checksum-field

## How to test ---

I've provided a docker image that wraps the `spdx-java-tools`: `caphill4/java-spdx-tools`

Three images are supplied by #1401
- alpine:latest
- debian:latest
- photon:3.0

You can follow these steps to see the newly valid spdx documents. Make sure to replace `<IMAGE>` with one of the above.
```
SYFT_FILE_METADATA_CATALOGER_ENABLED=true SYFT_FILE_METADATA_DIGESTS=sha1 go run cmd/syft/main.go <IMAGE> -o spdx > test.spdx

docker run -v (pwd)/test.spdx:/test.spdx caphill4/java-spdx-tools Verify /test.spdx 
```
Screenshot 2022-12-19 at 4 08 59 PM

### Context
Running the following command `syft -o spdx-json alpine:latest > spdx.json` generates an SPDX json document.

1) Using https://github.com/spdx/tools-python a user can run:
`pyspdxtools_parser --file spdx.spdx `

This results in a python error:
`TypeError: 'NoneType' object is not iterable`

The fix in this PR updates the checksum array from `null` ==> `[]`

2) After an empty array is replaced the following errors are seen:
```
alpine:latest: /usr/share/apk/keys/alpine-devel@lists.alpinelinux.org-616ac3bc.rsa.pub: At least one file checksum algorithm must be SHA1
alpine:latest: /usr/share/apk/keys/alpine-devel@lists.alpinelinux.org-616adfeb.rsa.pub: At least one file checksum algorithm must be SHA1
alpine:latest: /usr/share/apk/keys/alpine-devel@lists.alpinelinux.org-616ae350.rsa.pub: At least one file checksum algorithm must be SHA1
alpine:latest: /usr/share/apk/keys/alpine-devel@lists.alpinelinux.org-616db30d.rsa.pub: At least one file checksum algorithm must be SHA1
alpine:latest: /usr/share/udhcpc/default.script: At least one file checksum algorithm must be SHA1
Errors while parsing:  True
```

Given that a user has enabled the file digest cataloger: 
`SYFT_FILE_METADATA_CATALOGER_ENABLED=true SYFT_FILE_METADATA_DIGESTS=sha1`

This PR adds the ability to generate a valid SPDX document by enhancing the package entity with a `PackageVerificationCode`.

The `PackageVerificationCode` is generated by obtaining all of the `Contains` relationships for a package and performing the following algorithm:
```
verificationcode = 0
filelist = templist = ""
for all files in the package {
    if file is an "excludes" file, skip it /* exclude SPDX analysis file(s) */

        append templist with "SHA1(file)/n"
    }
sort templist in ascending order by SHA1 value
filelist = templist with "/n"s removed. /* ordered sequence of SHA1 values with no separators */
verificationcode = SHA1(filelist)
```
https://spdx.github.io/spdx-spec/v2.3/package-information/#79-package-verification-code-field

#### Note: 
We use the `Contains` relationship here since this is the value the SPDX validator will key off of if a package verification code is missing.

Example:
Validator warns: `At least one file checksum algorithm must be SHA1 for package x files`

After providing a SHA1 digest for the file:

Validator warns: `Found analyzed files for package x when analyzedFiles is set to false`

After settting analyzedFiles to true:

Validator warns: `Missing Package Verification Code for package x`

#### Note
Empty-size checks were removed from the digest cataloger in order to produce a valid document where empty files like `/lib/apk/db/lock` were not being summed. 

If this is incorrect or needs to be reworked so we're not missing some edge case regarding the removed tar file note I'm happy to refactor.


syft/file/digest_cataloger.go
```diff
@@ -92,15 +92,11 @@ func DigestsFromFile(closer io.ReadCloser, hashes []crypto.Hash) ([]Digest, erro
 		writers[idx] = hashers[idx]
 	}
 
-	size, err := io.Copy(io.MultiWriter(writers...), closer)
+	_, err := io.Copy(io.MultiWriter(writers...), closer)
```
spiffcs, line 95: We should discuss this change -- since we would be producing invalid SPDX documents even given `digest` enabled and `SHA1` being produced

test/cli/spdx_json_schema_test.go
```diff
@@ -35,11 +34,6 @@ func TestSPDXJSONSchema(t *testing.T) {
 			args:       []string{"-o", "spdx-json"},
 			fixture:    imageFixture,
 		},
-		{
-			name:       "power-user:image:docker-archive:pkg-coverage",
```
spiffcs, line 39: The poweruser command is no longer defaulting to spdx and instead only outputs syft-json. This test is no longer needed as we do not produce spdx anymore for this given path

syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -414,6 +414,7 @@ func toFiles(s sbom.SBOM) (results []*spdx.File) {
 }
 
 func toFileChecksums(digests []file.Digest) (checksums []common.Checksum) {
+	checksums = make([]common.Checksum, 0, len(digests))
```
spiffcs, line 417: Empty array `[]` instead of `null`

syft/file/digest_cataloger.go
spiffcs: @wagoodman I don't have a ton of context on why 0 size was not getting a digest previously. Let me know if this change is incorrect and breaks another assumption we were making.

Note: we do need to produce digests for files size 0 for a valid SPDX document

spiffcs:
Waiting on https://github.com/anchore/syft/pull/1383 to merge as there is some overlap here. Changes in that PR will need to be pulled in and updated on this PR

syft/file/digest_cataloger.go
wagoodman: you're change here is the correct one 👍

cmd/syft/cli/eventloop/tasks.go
```diff
@@ -83,6 +83,8 @@ func generateCatalogFileMetadataTask(app *config.Application) (Task, error) {
 	return task, nil
 }
 
+// SPDX requires that a SHA1 digest be present for all files in the SBOM.
+// This task checks for the config, but also cases where the format is spdx
```
spiffcs, line 87: Note: I should remove this comment since we're no longer doing this for the user

syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -462,3 +485,56 @@ func toFileTypes(metadata *source.FileMetadata) (ty []string) {
...
+	})
+
+	// filelist = templist with "/n"s removed. /* ordered sequence of SHA1 values with no separators
+	var b strings.Builder
+	for _, digest := range digests {
+		b.WriteString(digest.Value)
+	}
+
+	//nolint:gosec
+	// verificationcode = SHA1(filelist)
```
wagoodman, line 532: nit: dead code

spiffcs: apologies - this is not deadcode but extracted from the SPDX algorithm docs. I'll just remove it since it's unclear

syft/sbom/sbom.go
```diff
@@ -65,6 +67,36 @@ func (s SBOM) AllCoordinates() []source.Coordinates {
...
+// If no types are provided, all relationships for the package are returned.
+func (s SBOM) RelationshipsForPackage(p pkg.Package, rt []artifact.RelationshipType) []artifact.Relationship {
+	if len(rt) == 0 {
+		rt = artifact.AllRelationshipTypes()
+	}
+
+	var relationships []artifact.Relationship
+	for _, relationship := range s.Relationships {
+		// check if the relationship is one we're searching for; rt is inclusive
+		idx := slices.IndexFunc(rt, func(r artifact.RelationshipType) bool { return relationship.Type == r })
```
wagoodman, line 80: using `exp/slices` is contingent on our current stance on introducing generic-dependent code into syft -- we should follow up on this

test/cli/test-fixtures/image-java-spdx-tools/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM cgr.dev/chainguard/jdk
+
+# Not committed to repository - can be found here: https://github.com/spdx/tools-java/releases
+ADD tools-java-1.1.3-jar-with-dependencies.jar /usr/local/bin/tools-java.jar
```
kzantow, line 4: Could we just curl/wget this in the Dockerfile from a specific release artifact? (Probably not latest, but also maybe latest)

syft/sbom/sbom.go
```diff
@@ -65,6 +67,36 @@ func (s SBOM) AllCoordinates() []source.Coordinates {
...
+			relationships = append(relationships, relationship)
+		}
+	}
+
+	return relationships
+}
+
+// CoordinatesForPackage returns all coordinates for the provided package for provided relationship types
+// If no types are provided, all relationship types are considered.
+func (s SBOM) CoordinatesForPackage(p pkg.Package, rt []artifact.RelationshipType) []source.Coordinates {
```
kzantow, line 91: It looks like we're only passing a single RelationshipType at present -- would this be better just as a single RelationshipType or maybe a slightly easier consumer signature of `CoordinatesForPackage(p pkg.Package, rt ...artifact.RelationshipType)`?

kzantow (commented):
This is great -- I can't wait to be able to have completely valid SPDX files!

Only real question is about `image-java-spdx-tools/Dockerfile`, primarily because I believe I saw that you might like to add some tests for this.

test/cli/test-fixtures/image-java-spdx-tools/Dockerfile
wagoodman: +1 on making the pull in the dockerfile and not needed any files in the docker context

spiffcs: updated to pull in the docker file and added cli tests for tag value validation

syft/sbom/sbom.go
spiffcs: Followed up internally and it looks like we can add generic code now to syft

syft/formats/common/spdxhelpers/to_format_model.go
wagoodman: oh! that makes sense now, having a comment here about this is useful, feel free to leave it, just maybe with a touch more context is all

test/cli/spdx_tooling_validation_test.go
```diff
@@ -0,0 +1,85 @@
...
+				}
+
+				f, err := os.CreateTemp(path.Join(cwd, "test-fixtures", "image-java-spdx-tools"), rename)
+				if err != nil {
+					log.Fatal(err)
+				}
+
+				// spdx tooling only takes a file with suffix spdx
+				err = os.Rename(f.Name(), filepath.Join(cwd, "test-fixtures", "image-java-spdx-tools", rename))
+				if err != nil {
```
wagoodman, line 64: nit: replace with `require.NoError(t, err)` (same comment in a few other places)

test/cli/test-fixtures/image-java-spdx-tools/Makefile
```diff
@@ -0,0 +1,8 @@
+all: build validate
+
+.PHONY: build
+build:
+	docker build -t spdx-java-tools:latest .
+
+validate:
+	docker run --rm -v ${FILE}:/home/build/${BASE} spdx-java-tools:latest Verify ${BASE}
```
kzantow, line 8: Is this just a utility to run the validator manually?

kzantow (approved):
LGTM 👍

test/cli/spdx_tooling_validation_test.go
```diff
@@ -0,0 +1,85 @@
...
+				}
+
+				var rename string
+				if strings.Contains(test.name, "json") {
+					rename = "test.json"
+				} else {
+					rename = "test.spdx"
+				}
+
+				f, err := os.CreateTemp(path.Join(cwd, "test-fixtures", "image-java-spdx-tools"), rename)
```
wagoodman, line 57: this is creating a temp directory within the project tree? In case anything goes wrong, I'd feel better about having temp dirs being created in the OS standard spot... otherwise we could have temp dirs that end up in the project tree and accidentally get checked in or accumulated over time.

test/cli/test-fixtures/image-java-spdx-tools/Makefile
spiffcs: the test calls it here:
https://github.com/anchore/syft/blob/026428f53d62350d1053f36af03c05655e1b346e/test/cli/spdx_tooling_validation_test.go#L74-L81

wagoodman (approved):
I have a couple of nit comments on the new test (feel free to take it or leave it). Overall a quite-worthy improvement to the SPDX validation --thanks for your time on this 🚀

test/cli/spdx_tooling_validation_test.go
spiffcs: Yea let me take a second pass here.

This is creating a temporary file. That file is then mounted into the docker container with the spdx tooling. I would also prefer this probably to exist at the OS standard spot - I think I can get the mount to work that way rather than write the file to this exact directory

test/cli/spdx_tooling_validation_test.go
```diff
@@ -0,0 +1,85 @@
...
+				}
+
+				// validate against spdx java tooling
+				fixturesPath := filepath.Join(cwd, "test-fixtures", "image-java-spdx-tools")
+				fileArg := fmt.Sprintf("FILE=%s", filepath.Join(fixturesPath, rename))
+				mountArg := fmt.Sprintf("BASE=%s", rename)
+				cmd = exec.Command("make", "validate", fileArg, mountArg)
+				cmd.Dir = fixturesPath
+				runAndShow(t, cmd)
+				os.Remove(filepath.Join(fixturesPath, rename))
```
wagoodman, line 81: this should probably use `t.Cleanup()` to ensure that this would always run regardless of test failures.

test/cli/test-fixtures/image-java-spdx-tools/Makefile
spiffcs: The weirdness with the renaming is because the spdx validation tools only take a file with the `.spdx` suffix so the temp files as they are made will fail if consumed by the validator

test/cli/spdx_tooling_validation_test.go
wagoodman: related nit: Also, I see why `os.CreateTemp` (file) is used over `t.TempDir`, though the upside of the latter is that cleanup is done for free (though it does take one extra `os.Open` to use... so somewhat of a tradeoff).

spiffcs: moved this so that `os.CreateTemp` uses t.TempDir so we get the free cleanup

test/cli/spdx_tooling_validation_test.go
spiffcs: removed `t.Cleanup()` and `os.Remove` in favor of `t.TempDir`

PR #1412, opened 2022-12-19
Add NetBSD support.

0-wiz-0:
Signed-off-by: Thomas Klausner <wiz@gatalith.at>


kzantow (approved):
Thanks for the contribution. Definitely seems to match the docs I can find on this:

> As an example, the build tag found at the top of a source file
> 
> `// +build darwin freebsd netbsd openbsd`

PR #1417, opened 2022-12-20
Clean package names found in python catalogers

wagoodman:
This PR fixes a couple of bugs:
- When there are formatting directives in `setup.py` package names (for dynamically crafted require lines, such as `%s` or `{}`) we should skip these lines
- When there are empty name or version fields for `requirements.txt` files (such as `==1.2.3`) we should skip these lines

### DIFF
--- a/syft/pkg/cataloger/python/parse_requirements.go
+++ b/syft/pkg/cataloger/python/parse_requirements.go
@@ -56,6 +56,11 @@ func parseRequirementsTxt(_ source.FileResolver, _ *generic.Environment, reader 
 		version = strings.TrimFunc(version, func(r rune) bool {
 			return !unicode.IsLetter(r) && !unicode.IsNumber(r)
 		})
+
+		if name == "" || version == "" {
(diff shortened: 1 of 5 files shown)


syft/pkg/cataloger/python/parse_requirements.go
```diff
@@ -56,6 +56,11 @@ func parseRequirementsTxt(_ source.FileResolver, _ *generic.Environment, reader
 		version = strings.TrimFunc(version, func(r rune) bool {
 			return !unicode.IsLetter(r) && !unicode.IsNumber(r)
 		})
+
+		if name == "" || version == "" {
+			log.WithFields("path", reader.RealPath).Warnf("found empty package in requirements.txt line: %q", line)
```
kzantow, line 61: Should this be `Debugf`? I presume this will only be actionable if the user is scanning their own code

kzantow (approved):

syft/pkg/cataloger/python/parse_requirements.go
wagoodman: good catch!

PR #1418, opened 2022-12-20
feat: update spdx format model to produce valid spdx json documents

spiffcs:
## Summary - add small adjustments so default spdx documents are valid

- Add Describes Relationship to spdx relationships
- Add DOCUMENT describes DOCUMENT relationship to final spdx format model
-  Add empty digests for files that are not regular files which cannot be linked back to their regular file

The above changes get syft to a point where it can produce SPDX documents that pass the different tooling validators found under [spdx-org](https://github.com/spdx?type=source). There are future enhancements we can make that allow syft to generate better digest information for symlinks and irregular files. 

### Future improvments
1) Symlinks can be updated in the SBOM to resolve to their real path and lookup the related digest built for that real path by the digest cataloger

2) The `DOCUMENT` `describes` relationship can be pointed to a specific package constructed by the SBOM. One proposal surrounding this is using the [primary package purpose field](https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field). For a container scan, we can use the `source` information to construct a package that references the scanned container with the primary package purpose of `CONTAINER`. The SPDX `DOCUMENT` `describes` relationship can then be pointed at that package.

#### Image showing new CLI tests for spdx validity against candidate images
Screenshot 2022-12-20 at 5 17 21 PM

### To Test
```
SYFT_FILE_METADATA_CATALOGER_ENABLED=true SYFT_FILE_METADATA_DIGESTS=sha1 go run cmd/syft/main.go alpine:latest -o spdx > test.spdx
```
```
SYFT_FILE_METADATA_CATALOGER_ENABLED=true SYFT_FILE_METADATA_DIGESTS=sha1 go run cmd/syft/main.go alpine:latest -o spdx-json > test.json
```

^ The above outputs can be submitted to the SPDX validator:
https://tools.spdx.org/app/validate/

Note: To produce valid SPDX we still need to enable to digest cataloger because of the files analyzed field. Running this with no sha1 digest capability will result in `Found analyzed files for package xxx when analyzedFiles is set to false`

There is also automated testing now in syft to check against the tooling:
Screenshot 2022-12-20 at 11 12 19 PM


syft/artifact/relationship.go
```diff
@@ -15,6 +15,9 @@ const (
 
 	// DescribedByRelationship is a proxy for the SPDX 2.2.2 DESCRIBED_BY relationship.
 	DescribedByRelationship RelationshipType = "described-by"
+
+	// DescribesRelationship is a proxy for the SPDX 2.3.0 DESCRIBES relationship.
+	DescribesRelationship RelationshipType = "describes"
```
wagoodman, line 20: I've been trying to keep all of the relationships that syft outputs "one-way" by design and not express "both ways". That is, we have a `described-by` relationship which `describes` expresses the same edge type in a different direction. Can we use the existing `described-by` relationship we have today (and flip the node order in the relationship) instead of duping the kind in order to express a different direction?

spiffcs: Great question! I think we can do that on the syft side and then have some formatting specific logic in the spdx format model code that gets the `DESCRIBES` wording into the final document.

^ Does this work for you? Apologies for including one-way edges that mirrored effort here - tunnel-visioned on the validation a little too hard.

spiffcs: Removed

syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -27,12 +27,30 @@ const (
 	noAssertion = "NOASSERTION"
 )
 
+var DocumentDescribesRelationship = &spdx.Relationship{
```
wagoodman, line 30: since this isn't being used elsewhere we shouldn't be exporting it. (it could probably be inlined within `ToFormatModel`)

wagoodman (approved):

PR #1427, opened 2022-12-27
docs: migrate to homebrew-core

chenrui333:
I have [created a PR to include the formula](https://github.com/Homebrew/homebrew-core/pull/118929) into the homebrew-core, the installation experience would be better (I think). Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,6 @@ choco install syft -y
 
 ### Homebrew
 ```bash
-brew tap anchore/syft
 brew install syft
 ```
 


spiffcs (approved):

spiffcs:
Thanks a million, @chenrui333! This is a lot better.

spiffcs:
@chenrui333, with syft now contributed into core should we follow this part of the contribution guide when bumping versions
https://github.com/Homebrew/homebrew-core/blob/HEAD/CONTRIBUTING.md#to-submit-a-version-upgrade-for-the-foo-formula

chenrui333:
sorry for the late reply (my email would be better reach, rui@chenrui.dev)

we have maintainers autobump to ensure the are up-to-date. And also you can use these two actions to keep it to up to date:
- https://github.com/mislav/bump-homebrew-formula-action
- https://github.com/dawidd6/action-homebrew-bump-formula

Let me know if that works for you. Thanks!

PR #1435, opened 2023-01-03
Add purls for binary classifiers

noqcks:
I noticed that purls were empty for binary matched packages

```
  {
      "id": "42128c6f2809fa6c",
      "name": "go",
      "version": "1.19.4",
      "type": "binary",
      "foundBy": "binary-cataloger",
      ...
      "purl": "",
      ...
  },
```

And after 

```
  {
     "id": "42128c6f2809fa6c",
     "name": "go",
     "version": "1.19.4",
     "type": "binary",
     "foundBy": "binary-cataloger",
     ...
     "purl": "pkg:generic/go@1.19.4",
     ...
  },
```

Reference: [purl-spec generic type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic)

cc: @kzantow


kzantow (approved):
Good catch!

Issue #1437, opened 2023-01-05
Add binary matcher for redis


noqcks:
**What would you like to be added**:

Add a new binary classifier for `redis-server` in the [binary classifiers](https://github.com/anchore/syft/blob/bc1edb9c8a2fb4824bfdcac6147edc2bbf47aaf6/syft/pkg/cataloger/binary/default_classifiers.go)

**Why is this needed**:

Redis is often installed from source https://redis.io/docs/getting-started/installation/install-redis-from-source/, it would be nice to have a cataloger for a `redis-server` binary.

**Additional context**:

PR #1438, opened 2023-01-05
Add redis binary classifier

noqcks:
Fixes https://github.com/anchore/syft/issues/1437

EDIT: WIP. Revising this regex.
EDIT2: Unfortunately this doesn't seem to work for some earlier redis versions. Seems like there's no stable way to get a version from parsing the redis-server binary. Closing for now.
EDIT3: I found that I can parse redis-cli with stability across versions. Reopened this PR.


syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -38,6 +38,15 @@ var defaultClassifiers = []classifier{
 		PURL:    mustPURL("pkg:generic/go@version"),
 		CPEs:    singleCPE("cpe:2.3:a:golang:go:*:*:*:*:*:*:*:*"),
 	},
+	{
+		Class:    "redis-binary",
+		FileGlob: "**/redis-cli",
```
kzantow, line 43: Can the Redis CLI be used to connect to remote Redis systems? If so, I don't think this will be a strong enough indication that Redis itself is present.

kzantow: Right, I understand it's likely that _if_ Redis is installed `redis-cli` also will be installed, but the inverse is not true: _if_ `redis-cli` is installed, there is no guarantee that Redis is.

noqcks: I would argue that the redis package includes all of these binaries together: `redis-server`, `redis-cli` and `redis-sentinel`. Although someone could install redis-cli or redis-server manually. There is no distinction for the CPE on the difference between `redis-server` or `redis-cli`, both are listed in CVEs [here](https://nvd.nist.gov/vuln/search/results?form_type=Basic&results_type=overview&query=cpe%3A2.3%3Aa%3Aredislabs%3Aredis%3A*%3A*%3A*%3A*%3A*%3A*%3A*%3A*&search_type=all&isCpeNameSearch=true). So matching on a single binary `redis-server` or `redis-cli` is still of value in vulnerability matching.

kzantow: A couple more comments about this -- first, I should note that we are currently being very selective about what binaries to include in this cataloging process. The [note on the binary cataloger](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/binary/cataloger.go#L15-L19) tries to explain the primary purpose of it:

> [The Binary] Cataloger is the cataloger responsible for surfacing evidence of a very limited set of binary files, which have been identified by the classifiers. The Cataloger is _NOT_ a place to catalog any and every binary, but rather the specific set that has been curated to be important, predominantly related to *toolchain-related* runtimes like Python, Go, Java, or Node. Some exceptions can be made for widely-used binaries such as busybox.

I think Redis seems to be a quite popular and widely used, which in my opinion probably lets it fall into that exception category -- I'll bring this up to the team.

Second, back to the discussion of `redis-cli` being an acceptable method of identifying Redis as a whole. I spent a little time digging into this and see your point that to install either Redis or `redis-cli` they are often installed together and then one would have to remove something. But this isn't always true, for example in Ubuntu you can `apt install redis-tools`, which only provides the `redis-cli`. I also looked around the CVEs and, while you are right that there isn't a distinction between Redis and `redis-cli`, there are only 2 or 3 things I found that were actually applicable to `redis-cli`, which tells me it's much better to identify Redis server and not introduce false positives because we've only found the CLI. As such, I think we'd definitely want to figure out some reliable ways to identify Redis server. Could you provide some docker images with various redis server versions installed and maybe I could help looking through the binaries?

noqcks: I found a way to match redis-server versions. Added some commits to this PR. 

I only uploaded snippets of the redis-server binarys to this repo to keep things small. But you can download the full versions like this for your own verification. 

```
#!/bin/bash

versions=(5.0.0 2.8.23 4.0.11 6.0.16 7.0.0)

for version in "${versions[@]}"; do
  container_id=$(docker run -d redis:$version | awk '{print $1}')
  docker cp $container_id:/usr/local/bin/redis-server redis-server-$version
done
```

kzantow (approved):
Thanks for the contribution, this LGTM! 👍

syft/pkg/cataloger/binary/default_classifiers.go
kzantow: Thanks, I did some testing with these images and everything seems to work well -- nice job!

PR #1442, opened 2023-01-09
fix: update attestation code to remove library dependencies and shellout for keyless flow

spiffcs:
## Summary

When importing `cosign` library dependencies for attestation support the syft binary grows from 29MB -> 63MB.

This PR shrinks the final binary size back to 29MB by incorporating a shell out alternative for the keyless flow. A second PR will follow that implements custom PKI provided by the user.

### Fixes
- Moves `syft attest` to use locally available cosign binary
- Moves attestation flow to `device` [flow](https://github.com/sigstore/cosign/blob/8dc365ae43931c8695f4ffab164e62395e108926/KEYLESS.md#oauth-flows)
- Adapts shell out of cosign command to only use the `custom` predicate format for sbom attestation: see [predicate type](https://github.com/sigstore/cosign/issues/2494)
- Any use of a predicate type other than custom for attestations will result in the following when trying to verify the attestation:
```
Error: none of the attestations matched the predicate type: custom
main.go:62: error during command execution: none of the attestations matched the predicate type: custom
```
- Adapts to use keyless non-interactive mode - Users using `COSIGN_EXPERIMENTAL=1` will need to interact with the provided URL in the output of the command to create the t-log entry for rekor

## Current Example Workflow
Currently, users of syft can use cosign to attest sbom outputs by running the following two commands:
```
syft -o json caphill4/test:latest > test.json && COSIGN_EXPERIMENTAL=1 cosign attest caphill4/test:latest --type
custom --predicate test.json

# view verification of the attestation
COSIGN_EXPERIMENTAL=1 cosign verify-attestation caphill4/test:latest > validation.json
Verification for caphill4/test:latest --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - Any certificates were verified against the Fulcio roots.
Certificate subject:  christopher.phillips@anchore.com
Certificate issuer URL:  https://github.com/login/oauth

# view sbom that attestation was made against
COSIGN_EXPERIMENTAL=1 cosign verify-attestation caphill4/test:latest | jq -r .payload | base64 --decode | jq -r
.predicate.Data
```

Note in the above example, the image `caphill4/test:latest` should be an image the user has write permissions for uploaded in some image registry.

## New Syft Attest

Note: off-screen is the device flow of using the generated URL to provide keyless credentials for the attestation

The same verification flow listed above can be used to verify the same attestation
Jan-09-2023 14-41-51

## TODO:

- [x] - Update options to recognize different sbom formats
- [x] - Validation of cosign installed before running
- [ ] Removal of completion might have been accidental from another wip branch - I'll add it back
- [ ] Old integration tests will no loner work for HARD PKI - keyless was harder to test


cmd/syft/cli/attest.go
```diff
@@ -4,74 +4,41 @@ import (
...
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
 	"github.com/anchore/syft/cmd/syft/cli/attest"
 	"github.com/anchore/syft/cmd/syft/cli/options"
-	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/internal/config"
 )
 
-const (
```
spiffcs, line 17: Adding this back with the update examples

cmd/syft/cli/attest.go
```diff
@@ -4,74 +4,41 @@ import (
...
-	attestSchemeHelp = "\n" + indent + schemeHelpHeader + "\n" + imageSchemeHelp
-
-	attestHelp = attestExample + attestSchemeHelp
-)
-
-func Attest(v *viper.Viper, app *config.Application, ro *options.RootOptions) *cobra.Command {
-	ao := options.AttestOptions{}
+func Attest(v *viper.Viper, app *config.Application, ro *options.RootOptions, po *options.PackagesOptions) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "attest --output [FORMAT] --key [KEY] [SOURCE]",
```
spiffcs, line 31: Updating this back with new correct examples

cmd/syft/cli/attest.go
```diff
@@ -4,74 +4,41 @@ import (
...
-		Short: "Generate a package SBOM as an attestation for the given [SOURCE] container image",
-		Long:  "Generate a packaged-based Software Bill Of Materials (SBOM) from a container image as the predicate of an in-toto attestation",
-		Example: internal.Tprintf(attestHelp, map[string]interface{}{
-			"appName": internal.ApplicationName,
-			"command": "attest",
-		}),
+		Use:   "attest",
+		Short: "Generate an SBOM and sign it with a private/keyless key",
```
spiffcs, line 18: This is wrong - there is no signing of the sbom itself - updating the verbage here

cmd/syft/cli/attest/attest.go
```diff
@@ -1,398 +1,195 @@
...
+			fmt.Printf("unable to close report destination: %+v", err)
 		}
+	}()
 
-		ko.PassFunc = passFunc
-	}
-
-	sv, err := sign.SignerFromKeyOpts(ctx, "", "", ko)
+	// could be an image or a directory, with or without a scheme
+	// TODO: validate that source is image
```
spiffcs, line 43: Adding validation that source is of type image here

cmd/syft/cli/attest/attest.go
```diff
@@ -1,398 +1,195 @@
...
 			return
 		}
 
-		sbomBytes, err := syft.Encode(*s, format)
-		if err != nil {
-			errs <- err
-			return
-		}
+		// TODO: add multi writer support
+		for _, o := range app.Outputs {
```
spiffcs, line 102: The `sbom.Writer` needs its own PR so that user workflows with multiple formats/destinations can individually have `Bytes` called to get a copy of the sbom (format specific) in memory for the attestation

cmd/syft/cli/attest/attest.go
```diff
@@ -1,398 +1,195 @@
...
+				errs <- fmt.Errorf("unable to find cosign in PATH; make sure you have it installed")
+				return
+			}
 
-func generateAttestation(si source.Input, predicate []byte, src *source.Source, sv *sign.SignerVerifier, predicateType string) ([]byte, error) {
-	var h v1.Hash
+			args := []string{"attest", si.UserInput, "--type", "custom", "--predicate", f.Name()}
+			execCmd := exec.Command(cmd, args...)
+			execCmd.Env = os.Environ()
+			execCmd.Env = append(execCmd.Env, "COSIGN_EXPERIMENTAL=1")
```
spiffcs, line 127: Currently supports keyless - a separate PR will follow here to support user provided PKI

cmd/syft/cli/attest/attest.go
```diff
@@ -1,398 +1,195 @@
...
-	default:
-		ref, err := name.ParseReference(src.Metadata.ImageMetadata.UserInput)
-		if err != nil {
-			return err
-		}
-
-		digest, err := ociremote.ResolveDigest(ref)
-		if err != nil {
-			return err
+func ValidateOutputOptions(app *config.Application) error {
```
spiffcs, line 157: Todo: I think this function is duplicated - I'll move it so it's in a common spot all cmd can use it

ui/event_handlers.go
```diff
@@ -570,3 +530,80 @@ func ImportStartedHandler(ctx context.Context, fr *frame.Frame, event partybus.E
...
+	title := tileFormat.Sprintf("Running %s shell command...", source)
+	s := bufio.NewScanner(reader)
+	l := list.New()
+
+	formatFn := func(_ progress.Progress) {
+		spin := color.Magenta.Sprint(spinner.Next())
+		_, _ = io.WriteString(titleLine, fmt.Sprintf(statusTitleTemplate, spin, title))
+	}
+
+	var tlogEntry string
```
spiffcs, line 557: Because we have to capture the t-log entry this isn't as generic as I would like it. This function will grow with matches for aux text if more shell out commands are used in syft. Probably fine for now, but needs a refactor in the future if we shell out again

cmd/syft/cli/attest.go
```diff
@@ -15,63 +14,48 @@ import (
...
 
 const (
-	attestExample = `  {{.appName}} {{.command}} --output [FORMAT] --key [KEY] alpine:latest
-  Supports the following image sources:
-    {{.appName}} {{.command}} --key [KEY] yourrepo/yourimage:tag     defaults to using images from a Docker daemon. If Docker is not present, the image is pulled directly from the registry.
-    {{.appName}} {{.command}} --key [KEY] path/to/a/file/or/dir      only for OCI tar or OCI directory
+	attestExample = `  {{.appName}} {{.command}} --output [FORMAT] <IMAGE>
+  Attest supports the following image sources:
+    {{.appName}} {{.command}} yourrepo/yourimage:tag     defaults to using images from a Docker daemon. If Docker is not present, the image is pulled directly from the registry.
+    {{.appName}} {{.command}} yourrepo/yourimage:tag     users of this command must have write access to the image repository for the attestation upload
```
wagoodman, line 20: if there a reason why the same command and args are listed twice? If not intentional, I would change this to only whitespace.

wagoodman (approved):

Issue #1446, opened 2023-01-11
syft memory usage is high 


gnmahanth:
**What happened**:
when catalogers are specified, syft indexes whole file system even when its not required 

**What you expected to happen**:
index only required files when catalogers are specified 

**Environment**:
- Output of `syft version`: latest version
- OS (e.g: `cat /etc/os-release` or similar): Linux

kzantow:
@gnmahanth the title says memory usage is high but the description is referring to Syft scanning the whole filesystem, are both of these problems? Could you provide some more information such as _how much memory_ Syft is using for you, what Syft command you are executing, any public containers we could use to reproduce the problem, etc?

gnmahanth:
From #1447 with commands 
`syft -vv packages confluentinc/cp-kafka`
> improvements while scanning confluentinc/cp-kafka image when all catalogers are enabled
syft memory usage: ~350MB
syft memory usage with this PR: ~125MB

`syft -vv packages confluentinc/cp-kafka --catalogers java`
> with only java cataloger enabled
syft memory usage: ~290MB
syft memory usage with this PR: ~80MB

PR #1447, opened 2023-01-11
Optimise resolver

gnmahanth:
Fixes: https://github.com/anchore/syft/issues/1446

Depends On PR: https://github.com/anchore/stereoscope/pull/151

when dpkdb-cataloger is selected syft indexes whole filesystem even when cataloger is looking for files in `**/var/lib/dpkg/{status,status.d/**}`, to fix this this PR adds path filter function which indexes files from paths required by catalogers when provided

improvements while confluentinc/cp-kafka image when all catalogers are enabled 
syft memory usage: ~350MB
syft memory usage with this PR: ~125MB

with only java catalogers enabled
syft memory usage: ~290MB
syft memory usage with this PR: ~80MB

### DIFF
--- a/cmd/syft/main.go
+++ b/cmd/syft/main.go
@@ -2,6 +2,9 @@ package main
 
 import (
 	"log"
+	"os"
+	"runtime"
+	"runtime/pprof"
 
 	"github.com/anchore/syft/cmd/syft/cli"
 )
@@ -14,5 +17,14 @@ func main() {
 
 	if err := cli.Execute(); err != nil {
 		log.Fatalf("error during command execution: %v", err)
+	}
+	f, err := os.Create("mem.pprof")
+	if err != nil {
+		log.Fatal("could not create memory profile: ", err)
+	}
+	defer f.Close() // error handling omitted for example
+	runtime.GC()    // get up-to-date statistics
+	if err := pprof.WriteHeapProfile(f); err != nil {
+		log.Fatal("could not write memory profile: ", err)
 	}
 }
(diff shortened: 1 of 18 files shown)


cmd/syft/main.go
```diff
@@ -15,4 +18,13 @@ func main() {
 	if err := cli.Execute(); err != nil {
 		log.Fatalf("error during command execution: %v", err)
 	}
+	f, err := os.Create("mem.pprof")
```
wagoodman, line 21: note: this should be removed.

Also there should already be memory profiling built into syft. If you use `SYFT_DEV_PROFILE_MEM` --that being said, it seems to have been inadvertently removed. I'll try and restore this functionality.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/anchore/syft
 
 go 1.18
 
+replace github.com/anchore/stereoscope => github.com/deepfence/stereoscope v0.0.0-20230111090613-ee0c9713629f
```
wagoodman, line 5: just a reminder for reviewers: this should be removed before merging (keep this comment unresolved for now)

cmd/syft/main.go
gnmahanth: remove in commit f4df5a2e962d93e4ef4aa4defa393ac28b1938ba

wagoodman:
We don't want to remove files from the resolver index since we don't exhaustively know all possible calls about possible content. That is, we may use the configured globs to find packages but the contents of glob results may hint at other locations which we add as evidence on to the package. These new evidence paths aren't known in advance, so removing them from the index would break existing functionality.

PR #1450, opened 2023-01-12
Bump github.com/spdx/tools-golang to v0.4.0

lucacome:
Adds support for SPDX 2.3


kzantow:
Hi @lucacome we already have SPDX 2.3 support -- we've been using the main branch of spdx/tools-golang for a while now, but it would definitely be good to update to the release tag!

kzantow (approved):

lucacome:
any idea why the tests are failing with `needs to be valid json` ?

kzantow:
@lucacome there is some redaction of values when comparing JSON to snapshots, I think there were fields added which are making the redaction ([which is just a few regular expressions](https://github.com/anchore/syft/blob/main/syft/formats/spdxjson/encoder_test.go#L49-L58)) cause the JSON to be invalid. This should be a simple fix, I can help with if you need. It _probably_ is something like this:

```go
func spdxJsonRedactor(s []byte) []byte {
	// each SBOM reports the time it was generated, which is not useful during snapshot testing
	s = regexp.MustCompile(`"created":\s+"[^"]*"`).ReplaceAll(s, []byte(`"created":""`))

	// each SBOM reports a unique documentNamespace when generated, this is not useful for snapshot testing
	s = regexp.MustCompile(`"documentNamespace":\s+"[^"]*"`).ReplaceAll(s, []byte(`"documentNamespace":""`))

	// the license list will be updated periodically, the value here should not be directly tested in snapshot tests
	return regexp.MustCompile(`"licenseListVersion":\s+"[^"]*"`).ReplaceAll(s, []byte(`"licenseListVersion":""`))
}
```

... and then the snapshots need to be updated by [setting this flag to true](https://github.com/anchore/syft/blob/main/syft/formats/spdxjson/encoder_test.go#L11) (you can just temporarily update the code)

kzantow:
@lucacome I went ahead and fixed the tests, etc. -- thanks for the contribution!

lucacome:
Thanks for taking care of this, unfortunately I got busy with other stuff 👍

PR #1451, opened 2023-01-12
Add dependabot

lucacome:
Adds dependabot to keep dependencies up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


lucacome:
I see there's a lot of noise coming from GitHub Actions because they're out of date and using deprecated commands (they're fixed in newer versions). Can anybody take a look at this?

wagoodman (approved):
we've tried this before when we had a lot more dependencies and it introduced a lot of noise. That being said, we'd be up for trying it again since the number of syft dependencies has dramatically decreased.

Issue #1479, opened 2023-01-18
ZERO npm dependencies discovered if any npm dependency has an array as a license


Mikcl:
**Please provide a set of steps on how to reproduce the issue**

1. Create a `package.json` with the following a package that has a deprecated license:

> [pause stream](https://www.npmjs.com/package/pause-stream) for example has 3 million weekly downloads, and a depreacted license it seems. 

```
{
  "name": "tmp",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "pause-stream": "0.0.11"
  }
}
```

2. generate the `package-lock.json` via `npm install --package-lock-only`

You will see something like the following in the `package-lock.json` (note the deprecated license evaluated):

```
requires": true,
  "packages": {
    "": {
      "name": "tmp",
      "version": "1.0.0",
      "license": "ISC",
      "dependencies": {
        "pause-stream": "0.0.11"
      }
    },
    "node_modules/pause-stream": {
      "version": "0.0.11",
      ....
      "license": [
        "MIT",
        "Apache2"
      ],
      "dependencies": {
        "through": "~2.3"
      }
    },
    ....
```

3. run `syft .` in the same directory.

**What happened**:

```
✔ Indexed .
 ✔ Cataloged packages      [0 packages]
No packages discovered
```

This happens even when there are numerous packages in the dependencies. i.e ZERO packages are discovered. Not `pause-stream`, nor its dependencies, or any other packages specified in `package.json/package-lock.json` prior to running syft. 

**What you expected to happen**:

I would expect syft to find all the packages regardless of whether the licencse section is deprecated or not (possibly even if it is invalid?) 

https://docs.npmjs.com/cli/v9/configuring-npm/package-json#license

**Anything else we need to know?**:

This is particualrly high importance as this would mean syft misses all of the npm dependencies if only one is "misformatted". No error is raised, only a warning when running in verbose mode.

```
WARN cataloger failed cataloger=javascript-lock-cataloger error=failed to parse package-lock.json file: json: cannot unmarshal array into Go struct field lockPackage.Packages.license of type string location=/home/[...redacted...]/package-lock.json
```

[...]

PR #1482, opened 2023-01-19
npm: package-lock license decoding to accept string or array

Mikcl:
Fixes https://github.com/anchore/syft/issues/1479

Introduces new decoding behaviour for the `license` field in `package-lock.json` packages. To accept a string and array as the datatype. 

Previously: only a string was an expected value. This would mean the decoding step would fail in the event of an array being present, and all the remaining dependencies in the package-lock.json would not be cataloged.

This PR:
1. Accepts either a string or an array.
2. If its neither string or array and unable to parse the license field. Sets license to nil and does not error to allow us to capture the remaining package dependencies (at the expense of perhaps an unexpected license field not being captured).

- Have created the new  `packageLockLicense` type with overriden `UnMarshallJSON` behaviour to create this change.
- Added test cases from an example `package-lock.json`

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_package_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_package_lock.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"strings"
 
+	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/pkg/cataloger/generic"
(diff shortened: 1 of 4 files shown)


syft/pkg/cataloger/javascript/parse_package_lock.go
```diff
@@ -24,6 +24,34 @@ type packageLock struct {
...
+
+	// 2. A string
+	var str string
+	if err = json.Unmarshal(data, &str); err == nil {
+		*licenses = make([]string, 1)
+		(*licenses)[0] = str
+		return nil
+	}
+
+	// 3. Unexpected
```
kzantow, line 48: maybe add something like `if len(data) > 2 {  // debug log contents }`?

kzantow (commented):
This looks great @Mikcl just one suggestion

syft/pkg/cataloger/javascript/parse_package_lock.go
Mikcl: added a debug log, did it for `len(data) > 0`. 

Note the log here would not be so informative as it is just the license value and knowing it happended in a `package-lock.json`. Though better than complete silence, sufficent for a first pass fix imo.

Mikcl:
Thanks for the quick review @kzantow 

Updated with the feedback :)

Mikcl:
Note the "CLI tests" [failing](https://github.com/anchore/syft/actions/runs/3962848151/jobs/6790053297#step:6:173) are not from changes in the PR but from `upstream/main`.

kzantow (approved):

Issue #1359, opened 2022-11-21
Symlink traversal confusion when scanning unpacked filesystems


jedevc:
Assume we have a directory containing an unpacked container image in the `target` directory:

```bash
$ echo "FROM alpine" | docker build -o type=local,dest=target
```

We could also create `target.tar` out of this same `target` directory:

```bash
$ tar cf target.tar target/*
```

If we scan this as a directory source, we can note a couple warning message that indicate that syft is scanning directories that weren't intended by the user:

```bash
$ syft -o spdx-json=out.json dir:./target -v
[0000]  INFO syft version: [not provided]
[0000]  WARN unable to access path="/bin/busybox": lstat /bin/busybox: no such file or directory
[0000]  WARN unable to access path="/run/brltty": open /run/brltty: permission denied
...
```

These warnings start to appear because of the symlinks `/bin/tar` to `/bin/busybox` and `/var/run` to `/run`. I would expect these contents to be resolved entirely within the provided path, but it seems like this behavior is intentional, though I'm not sure of the context: see https://github.com/anchore/syft/blob/main/syft/source/directory_resolver.go#L530-L532 introduced in https://github.com/anchore/syft/pull/442.

What's even weirder to me is this behavior is identical even when scanning `file:target.tar` (since it shares the same directory resolver code, after unpacking it into a tmp directory):

```bash
$ syft -o spdx-json=out.json file:./target.tar -v
[0000]  INFO syft version: [not provided]
[0000]  WARN unable to access path="/bin/busybox": lstat /bin/busybox: no such file or directory
[0000]  WARN unable to access path="/run/brltty": open /run/brltty: permission denied
```

After [raising on the community slack](https://anchorecommunity.slack.com/archives/C019BUXV7R6/p1669042269693499), @kzantow suggested a possible new source type, `root-dir` (or similar), where all content could be resolved internally, additionally, with the same cataloguers as the `image:` source (since those also differ).

kzantow:
We still need to figure out the right solution here, but this definitely seems like something we want to support.

kzantow:
I'm going to leave this open until it gets implemented in Syft proper, not just the API.

kzantow:
This was completed by #1867 -- there is a `--base-path` flag and matching config option to set this. The complication with symlinks is that there are different scenarios you want different behavior, here's an explanation of the current Syft behavior:

### if you are scanning a source code type filesystem, you probably DO NOT want to set base-path

For example you if you have:
```
/src
[...]

PR #1485, opened 2023-01-19
Allow scanning unpacked container filesystems

jedevc:
:bug: Fixes https://github.com/anchore/syft/issues/1359 :tada:

This PR is an attempt to explore how to modify the existing directory resolver to allow for resolving symlinks in an unpacked container filesystem. To do this, we add a new optional `base` parameter for the directory resolver that resolves all symlinks relative to this base. There are two intended use cases:

- `base = "/"`. The previous behavior, symlinks are resolved relative to the root filesystem.
- `base = path`. Symlinks are resolved relative to the target filesystem, allowing correct behavior when scanning unpacked container filesystems on disk.

Currently, this only modifies the API surface for the source, which would probably be sufficient for [my use case in buildkit-syft-scanner](https://github.com/docker/buildkit-syft-scanner/blob/fd906b1a34eac4dac9edc7b614d6463ce7943f87/internal/target.go#L52-L55).

I've marked this as in-draft since I've only done some preliminary testing, but the following things probably need working out before this could be ready:
- [x] Extend this to the CLI with a `root-dir` input type, or similar? Not quite sure on naming, I don't have any particular preference here.
  - *Update*: planned for follow-up
- [x] Add tests for the new directory resolver behavior. The logic seems sound to me, but I may have missed something - thankfully, the symlink resolution seems isolated to a couple of places.
  - *Update*: tests added
- [x] I've temporarily replaced [`filepath.EvalSymlinks`](https://pkg.go.dev/path/filepath?utm_source=godoc#EvalSymlinks) with [`cfs.RootPath`](https://pkg.go.dev/github.com/containerd/continuity/fs#RootPath) from [containerd/continuity](https://github.com/containerd/continuity). The logic of these functions is fairly similar, however, the containerd equivalent allows bounding the result to a specified directory.
   This logic is only needed for 1 of the symlink resolution components - the other is much simpler, since syft does the heavy lifting. I'm not sure how the maintainers feel about adding this as a new dependency? Maybe there's a simpler solution I'm not seeing :eyes:
   *Update*: I can remove this dependency, since we can avoid this second instance of doing symlink resolution - we can just use the file tree that we've already indexed over, which should be more efficient anyways. I *think* that this is a valid use? It seems to work in the cases I tried it on.

cc @wagoodman @kzantow


syft/source/directory_resolver.go
```diff
@@ -244,10 +259,21 @@ func (r directoryResolver) addSymlinkToIndex(p string, info os.FileInfo) (string
...
+		linkTarget = filepath.Join(r.base, linkTarget)
+	} else {
+		// if the link is not absolute (e.g, /dev/stderr -> fd/2 ) we need to
+		// resolve it relative to the directory in question (e.g. resolve to
+		// /dev/fd/2)
 		linkTarget = filepath.Join(filepath.Dir(p), linkTarget)
+
+		// we then need to ensure that the relative link cannot escape from the
+		// base directory
+		if !strings.HasPrefix(linkTarget, r.base) {
```
jedevc, line 274: Note to self: this is probably not a great idea - it feels like we could confuse `/foo` and `/foobar` here?

wagoodman (approved):
nice improvement 👍  this additionally further leverages the already built indexes which helps some upcoming PRs regarding performance boosts

cburgess:
Would love to the see this feature make it to the CLI. We have a use case for wanting to run syft on a root filesystem and have it work just like syft does against an existing docker image today.

PR #1484, opened 2023-01-19
fix: skip filling in empty fields in APK metadata

Nirusu:
Attempt to fix #1483. This should skip empty fields in the definition, leaving the Dependencies and Provides slices empty.
Added a test for it. 

The debug log message might be a bit noisy, so not sure whether to keep it? It also lacks the package name when printing. However, since the package name is parsed from another field in the same context, I didn't want to rely on it.

Feedback and suggestions appreciated!

PR #1494, opened 2023-01-19
Fix panic in APK version specifier handling 

luhring:
A bug was introduced in https://github.com/anchore/syft/pull/1063 where Syft would panic if a package listed in the installed DB doesn't "provide" anything.

This PR ~reworks the logic so that Syft allows for empty "provides" values~ (edit: no longer part of this PR, see https://github.com/anchore/syft/pull/1494#issuecomment-1398096945). It also adjusts the `stripVersionSpecifier` helper function to remove the possibility of panicking (index out of range).

### DIFF
--- a/syft/pkg/cataloger/apkdb/parse_apk_db.go
+++ b/syft/pkg/cataloger/apkdb/parse_apk_db.go
@@ -321,9 +321,6 @@ func discoverPackageDependencies(pkgs []pkg.Package) (relationships []artifact.R
 		}
 		lookup[p.Name] = append(lookup[p.Name], p)
 		for _, provides := range apkg.Provides {
-			if provides == "" {
-				continue
-			}
 			k := stripVersionSpecifier(provides)
 			lookup[k] = append(lookup[k], p)
 		}
--- a/syft/pkg/cataloger/apkdb/parse_apk_db.go
+++ b/syft/pkg/cataloger/apkdb/parse_apk_db.go
@@ -364,7 +364,7 @@ func stripVersionSpecifier(s string) string {
 	// examples:
 	// musl>=1                 --> musl
 	// cmd:scanelf=1.3.4-r0    --> cmd:scanelf
-	
+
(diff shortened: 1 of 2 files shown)


luhring:
Having trouble making sense of the failure logs: https://github.com/anchore/syft/actions/runs/3963383936/jobs/6791170964#step:6:159

Help appreciated 😃

spiffcs:
@luhring we've updated this in main now where this CLI test is fixed - if you're able to merge from upstream you should no longer see this failure

Nirusu:
I basically submitted the same fix for #1484, just in a different way.
Don't mind in which way this is fixed, though this also applies to "dependencies", not just "provides" from what I remember correctly.

luhring:
@Nirusu Ah! I hadn't seen your PR, sorry — I'm glad you caught this. I actually like your solution better! I say we merge it, and I can close mine. My PR was just trying to stop the bleeding.

Yours might need a rebase, and I like the idea of incorporating the changes to `stripVersionSpecifier` to prevent panics in the future — but to me the first priority is preventing panics, which your PR does elegantly. 👏

wagoodman:
@luhring  I'm going to merge the PR that @Nirusu  provided for this, however I do think keeping the `stripVersionSpecifier` fix also make sense -- if you'd like you could rescope this PR to just that enhancement or open another PR.

luhring:
> @luhring I'm going to merge the PR that @Nirusu provided for this, however I do think keeping the `stripVersionSpecifier` fix also make sense -- if you'd like you could rescope this PR to just that enhancement or open another PR.

Sounds good, thanks @wagoodman!

wagoodman:
note: the force push was to rebase on the latest in main (fixes the failing CLI test)

changes made:
- added some explicit tests for `stripVersionSpecifier`
- removed the empty value check before calling `stripVersionSpecifier`

wagoodman (approved):

PR #1501, opened 2023-01-20
Add release trigger

wagoodman:
This PR proposes to make the following changes to the release process. Specifically, replace manually pushing a tag with `make trigger-release`, which will:
- check if you're authenticated into github in order to trigger a release
- speculate the next release version
- generate a changelog to preview before release
- kick off the release workflow (the release workflow would tag the release instead)

This removes the need for the developer kicking off the release to pick the release and offers an opportunity to see the effective changelog to double check the issue and PR labels.

The "break glass" approach (incase chronicle is wrong and manual intervention is needed) is to run the release workflow via workflow_dispatch and supply the intended version.

Screen Shot 2023-02-08 at 10 20 15 AM

TODO:
- [x] update chronicle to additionally include all remaining closed issues and merged PRs in the preview list to show what might not be labeled or associated correctly [update: chronicle v0.5.1 has this feature on by default]
- [ ] ~add `gh` tooling to the bootstrap? (or similar)~ update: there isn't a way to install this nicely in .tmp without a lot of work (`go install` will not work either due to their usage of go mod replace lines). The trigger script now checks to see if gh is installed before continuing.
- [x] add `gh` auth check up front before release (can you kick off the workflow before attempting to do so)
- [x] incorporate https://github.com/anchore/chronicle/pull/65 to not calculate the changelog/version twice


.github/workflows/release.yaml
```diff
@@ -1,12 +1,10 @@
...
-    # take no actions on push to any branch...
-    branches-ignore:
-      - "**"
-    # ... only act on release tags
-    tags:
-      - "v*"
+  workflow_dispatch:
+    inputs:
+      version:
+        description: tag the latest commit on main with the given version (prefixed with v)
```
kzantow, line 6: It might be nice if this workflow validates the correct tag format, I don't see this below

Makefile
```diff
@@ -16,6 +18,7 @@ GORELEASER_VERSION := v1.15.1
 YAJSV_VERSION := v1.4.1
 COSIGN_VERSION := v1.13.1
 QUILL_VERSION := v0.2.0
+GLOW_VERSION := v1.4.1
```
kzantow, line 21: Should this get added to the update bootstrap dependency workflow?

kzantow (approved):
A couple nit comments, but LGTM otherwise 👍

Makefile
wagoodman: good call -- updated

PR #1509, opened 2023-01-24
add sbom artifacts to model

houdini91:
Hi,
I noticed using the sbom cataloger the relations that are read from the sbom are not aggregated in to the Syft model sbom.

I have a small fix suggestion, not sure if there was a reason behind it or simply a typo.

Are there any tests you would like me to add for this PR?


kzantow:
@houdini91 this looks good but are you able to fix the tests?

syft/pkg/cataloger/internal/pkgtest/test_generic_parser.go
```diff
@@ -179,9 +180,29 @@ func (p *CatalogTester) assertPkgs(t *testing.T, pkgs []pkg.Package, relationshi
 		t.Errorf("unexpected packages from parsing (-expected +actual)\n%s", diff)
 	}
 
-	if diff := cmp.Diff(p.expectedRelationships, relationships, p.compareOptions...); diff != "" {
-		t.Errorf("unexpected relationships from parsing (-expected +actual)\n%s", diff)
+	if err := SubsetRelationship(p.expectedRelationships, relationships, p.compareOptions...); err != nil {
+		diff := cmp.Diff(p.expectedRelationships, relationships, p.compareOptions...)
+		t.Errorf("unexpected %s relationships from parsing (-expected +actual)\n%s", err, diff)
```
kzantow, line 185: I'm not entirely sure this is ideal -- if there is a diff, the diff will be very noisy and not represent what the check is actually doing. Could we just update the individual tests to have the right expected relationships?

houdini91: Np, do you want me to as simply 150 relationship to the static go code? Should we remove some from the sbom?

houdini91: Fyi the diff line your referring is simply to output for failed subset check above it was simpler for me to use it's full diff output for test outpu
But I understand you concern about the subset comparison

kzantow: I'd be good with updating the source SBOM to make a manageable set of relationships to add explicitly to the tests.

kzantow:
@houdini91 I think you may also just need to run `make lint-fix` for the static analysis failure

houdini91:
@kzantow Hi, sorry for the delay, i admittedly forgot about this PR.
Fixed your suggestions, rebased and squashed (i forgot to sign some commit).
Have a look seem like the tests pass.

kzantow (approved):
LGTM 👍

kzantow:
Thanks @houdini91 !

PR #1526, opened 2023-01-31
feat: update golang to 1.19

bradleyjones:
Go 1.18 will become EOL with the upcoming release of Go 1.20

PR #1528, opened 2023-01-31
chore: refresh git index before checking diff

robcresswell:


spiffcs:
👋 thanks for the great find @robcresswell!

We incorporated it into the script in this PR:
https://github.com/anchore/syft/pull/1526/files

Closing this for now and linking it to the go update

Issue #1532, opened 2023-02-02
Syft hardcodes `custom` attestation type 


Nirusu:
**Please provide a set of steps on how to reproduce the issue**
(see: https://github.com/edgelesssys/constellation/actions/runs/4076041100/jobs/7023239500)
Using Syft 0.69.0, if you have a workflow like:

```bash
syft attest --key cosign.key ghcr.io/edgelesssys/constellation/joinservice:v2.6.0-pre.0.20230131140552-27cae81bd7bc@sha256:cd33aacb5733f6cb7c9d9694a8d6c5337c78e019274ea1d531e25114f86b537b -o cyclonedx-json

cosign attach attestation ghcr.io/edgelesssys/constellation/key-service:v2.6.0-pre.0.20230202153131-e91a34d4ad99@sha256:e82624bc9889497589672f56680a9b5e4f8f6ead5673d7fcc6507ff534218b3d --attestation container-image.att.json

cosign verify-attestation ghcr.io/edgelesssys/constellation/key-service:v2.6.0-pre.0.20230202153131-e91a34d4ad99@sha256:e82624bc9889497589672f56680a9b5e4f8f6ead5673d7fcc6507ff534218b3d --type 'https://cyclonedx.org/bom'
```

It will fail:
```bash
Error: none of the attestations matched the predicate type: https://cyclonedx.org/bom
main.go:62: error during command execution: none of the attestations matched the predicate type: https://cyclonedx.org/bom
```

**What happened**:
Syft hardcodes `--type custom` in the call to `cosign` as an argument.

**What you expected to happen**:
Syft passes my defined type to `cosign`.

**Anything else we need to know?**:
Judging from #1442, this hardcoding seems to be intentional, but in my opinion, for a completely flawed reason.

Cosign does not auto-detect the predicate type, as described in https://github.com/sigstore/cosign/issues/2264.

@spiffcs noticed this issue and reported this in https://github.com/sigstore/cosign/issues/2494.

However, the issue here is that Cosign also requires the type to be set for verification, due to the lack of auto-detection. It defaults to `custom` whereas @spiffcs has defined a type in the `syft attest` command in https://github.com/sigstore/cosign/issues/2494, but not in the `cosign attest` command.

Now, I am not sure why #1442 does hard code the type for,  given that seems to me like either user error (defining a type on one command but not on another), or an upstream issue in Cosign (detection does not work, wrong types between different command).

[...]

PR #1533, opened 2023-02-02
Don't hardcode Cosign attest type

Nirusu:
My proposed fix for #1532.

Since this seems to be intentional, I also raised the issue to describe the issue and maybe keep the discussion there?

Depending on how the discussion goes, maybe merge or close this PR. I don't know - the fix is trivial though :)


kzantow:
@spiffcs probably has the most context here, but I believe the issue is we can no longer verify attestations if they are not using the `custom` type

Nirusu:
Yes, see the issue :)

From what I gathered is that the issue here is 
a) a missing `type` in the verification command 
and/or 
b) Cosign itself - not Syft.

Cosign default to `custom` if no type is specified, so I don't know why this has been hard coded?

spiffcs:
👋 Hi @Nirusu see this issue we filed with cosign upstream:
https://github.com/sigstore/cosign/issues/2494

As of this feature the only way I could get `verify-attestation` to behave correctly was by using a custom type - this is why it's been hardcoded this way.

spiffcs:
I read the issue you linked - I do agree that we probably should just expose the type at this point and let the downstream commands work as expected giving the user the most freedom

spiffcs:
I'll add that change - thanks for the report!

Nirusu:
I saw your issue (see my chain of thoughts in #1532), but that's what I don't get.

You ran:
```bash
$COSIGN_EXPERIMENTAL=1 cosign attest --predicate test_spdx_scratch.json --type spdxjson caphill4/scratch:latest 
COSIGN_EXPERIMENTAL=1 cosign verify-attestation caphill4/scratch:latest                                                                                                                                    
```

The attest is specified with a predicate, the verify-attestation **without**.

That doesn't make sense to me. 
If you specify the predicate, why add not add it to the verify-attestation? It does not auto-detect the predicate, that's actually broken.
If you want `custom` as a type anyway, why specify it? Cosign defaults to `custom` if nothing is specified (hence this PR).

So I don't really see what you were going here for - not specifying it should already give you the same `custom` behavior.
Specifying it as a predicate in both commands - attest and verify-attestation - should also work (not 100% sure now, though). 

But only in one command? Yes, auto-detection is broken, but I would still pass that as user error if you define another type but ultimately don't care about the type.

spiffcs (approved):

PR #1538, opened 2023-02-03
fix: update config struct to not decode password/key

spiffcs:
Affects syft versions >= v0.69.0

### DIFF
--- a/internal/config/attest.go
+++ b/internal/config/attest.go
@@ -3,6 +3,7 @@ package config
 import "github.com/spf13/viper"
 
 type attest struct {
+	// IMPORTANT: do not show the attestation key/password in any YAML/JSON output (sensitive information)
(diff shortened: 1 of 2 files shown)


internal/config/attest.go
```diff
@@ -3,8 +3,8 @@ package config
 import "github.com/spf13/viper"
 
 type attest struct {
-	Key      string `yaml:"key" json:"key" mapstructure:"key"`
-	Password string `yaml:"password" json:"password" mapstructure:"password"`
+	Key      string `yaml:"-" json:"-" mapstructure:"key"`
+	Password string `yaml:"-" json:"-" mapstructure:"password"`
```
wagoodman, line 7: nit: mind adding comments similar to https://github.com/anchore/syft/blob/main/internal/config/registry.go#L13 ?

wagoodman (approved):

Issue #933, opened 2022-03-31
Licenses missing in most report format


WhyJee:
**What happened**:

Scanning the same image leads to different results depending on the output format.

| Type |  Components | cpe | purl | Versions | Licenses | Notes |
| -- | -- | -- | -- | -- | -- | -- |
| cyclonedx | 154 | 154 | 153 | 154 |  0 | **WARN** unable to convert relationship from CycloneDX 1.3 JSON, dropping... |
| cyclonedx-json | 154 | 154 | 153 | 154 |  0 | **WARN** unable to convert relationship from CycloneDX 1.3 JSON, dropping... |
| json | 153 | 806 | 153 | 153 | 153 | with 2 UNKNOWN License for python artifacts |
| spdx-json | 153 | 806 | 153 | 153 | 0 | |
| spdx-tag-value | 153 | 806 | 153 | 153 | 0 | |
| text | 153 | 0 | 0 | 153 |  0 |  |
| table | 153 | 0 | 0 | 153 |  0 |  |

Scanning the same image using `tern`

Type | Components | cpe | purl | Versions | Licenses | Notes
-- | -- | -- | -- | -- | -- | --
cyclonedxjson|149|0|149|149|149|
html|149|0|0|149|149|
json|149|0|0|149|149|
spdxtagvalue|149|0|0|149|149|
spdxjson|149|0|0|149|149|
yaml|149|0|0|149|149|

Thus the presence or absence of the license is not a format problem as for the common spdx or cyclonedx, tern is able to get this field correctly filled. As in json Syft is able to have all the information, this is probably in the converter that the loss occur (which is reflected I think by the WARN logs).

**What you expected to happen**:

Expectation is that content is independent of the format (if we except of course table and text) and everything that the format may accept shall be in the output.

**How to reproduce it (as minimally and precisely as possible)**:

```
docker run \
            --rm \
            -it \
            -v /var/run/docker.sock:/var/run/docker.sock \
            -v $PWD:/tmp/workdir \
            anchore/syft:latest \
            -v \
            packages \
            -s Squashed \
            -o <format> \
            --file /tmp/workdir/bom.format \
            docker:almalinux:latest
```

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: 0.42.4
- OS (e.g: `cat /etc/os-release` or similar):

spiffcs:
Tagging @cpendery

cpendery:
[...]

PR #1540, opened 2023-02-05
add support for licenses not found on list

### LINKED ISSUE #1529: apk packages with simplified license show NOASSERTION

syft reports the license concluded as `NOASSERTION` when a license that may not be complete but is nonetheless intelligible is found.

For example, use the following dockerfile:

```dockerfile
FROM alpine:3.17
RUN apk add openssh-client-common
```

Running syft on the above will give the following output for the package `openssh-client-common` (in spdx-json format):

```json
  {
   "name": "openssh-client-common",
   "SPDXID": "SPDXRef-Package-apk-openssh-client-common-931d8f7c42c54694",
   "versionInfo": "9.1_p1-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://www.openssh.com/portable.html",
   "sourceInfo": "acquired package info from APK DB: lib/apk/db/installed",
   "licenseDeclared": "NOASSERTION",
   "description": "OpenBSD's SSH client common files",
```

The relevant part of `/lib/apk/db/installed` is:

```
C:Q1YcVPyM0QbevnRbzdyPGn4hQmAjk=
P:openssh-client-common
V:9.1_p1-r1
A:aarch64
S:835516
I:3186688
T:OpenBSD's SSH client common files
U:https://www.openssh.com/portable.html
L:BSD
```

I stepped through syft's code, and found that it carried the `"BSD"` license in the `Licenses` field all the way through until [license.ID()](https://github.com/anchore/syft/blob/fb388c0f2525e0e4af8399ba44e5f1fe5b9de9ae/internal/spdxlicense/license.go#L17), where it lowercases it ("BSD" -> "bsd") and then checks in the table [here](https://github.com/anchore/syft/blob/fb388c0f2525e0e4af8399ba44e5f1fe5b9de9ae/internal/spdxlicense/license_list.go#L8). The table has lots of variants of BSD with versions and such, but nothing that is just plain "bsd".

I get that the license itself really _should_ be "BSD with a version". My question is how we should support this in the real world, where these exist. Perhaps there should be an option `--accept-unknown-licenses=true`, to say, "if I do not recognize the license, instead of `NOASSERTION`, I will show the text as is, maybe with some warning".

For what it is worth, the `APKBUILD` with license is [here](https://git.alpinelinux.org/aports/tree/main/openssh/APKBUILD?h=3.17-stable).

Note: may be related to #933 ; I chose to open a new issue as I have as simple replication and am focused on one particular problem, rather than the "license output differs between files". The cause, I believe, is the same.

deitch:
Fixes #1529 

As discussed with @kzantow , if a license identifier is not found on the official list, rather than returning `""` and `false`, it creates `LicenseRef-<id>` per the [other licenses section of the spdx spec](https://spdx.github.io/spdx-spec/v2.3/other-licensing-information-detected/#102-extracted-text-field).

Also adds handling an empty identifier, and adds tests.

The approach may need some tweaking, but it seems a simple enough solution.


kzantow (commented):
Hey @deitch -- thanks for submitting this PR! I think this is half of what needs to happen -- the `LicenseRef-<id>` getting put in the appropriate license field. The other half is that we need to add these licenses with a matching `LicenseRef-<id>` to the **Other Licenses section** so there is something to refer to

deitch:
I was kind of wondering about that. Thanks for the pointer to the right struct.

I saw that [`ID()`](https://github.com/anchore/syft/pull/1540/files#diff-82757e250ab469859dc510032987e691e114a9518412ac68ebeb83d4133b3fd3) is used by both spdx and cyclonedx, even though it is part of spdx. Do I need to find a parallel struct for CycloneDX?

kzantow:
@deitch CycloneDX is easier -- we just need to include the _name_ instead of ID. [See the spec here](https://cyclonedx.org/docs/1.4/json/#tab-pane_components_items_licenses_items_license_oneOf_i1), optionally including `text` if we have it

deitch:
Ok, I think I got it, both for spdx and cyclonedx, but have a look. I updated tests so they handle the cases and pass, and I also ran some tests locally.

syft/formats/common/spdxhelpers/license.go
```diff
@@ -22,16 +22,24 @@ func License(p pkg.Package) string {
...
+	parsedLicenses := ParseLicenses(p.Licenses)
 
 	if len(parsedLicenses) == 0 {
 		return NOASSERTION
 	}
 
 	return strings.Join(parsedLicenses, " AND ")
 }
+
+func ParseLicenses(raw []string) (parsedLicenses []string) {
```
kzantow, line 34: I don't think this needs to be exported (e.g. it can start with lowercase `parseLicenses`)

syft/formats/common/spdxhelpers/license.go
```diff
@@ -22,16 +22,24 @@ func License(p pkg.Package) string {
...
+	for _, l := range raw {
+		if value, other, exists := spdxlicense.ID(l); exists {
+			parsed := value
+			if other != "" {
+				parsed = "LicenseRef-" + other
```
kzantow, line 39: This should use the constant `LicenseRefPrefix`

syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -511,6 +513,28 @@ func toFileTypes(metadata *source.FileMetadata) (ty []string) {
 	return ty
 }
 
+func toOtherLicenses(catalog *pkg.Catalog) []*v2_3.OtherLicense {
```
kzantow, line 516: This should return `[]*spdx.OtherLicense` and we shouldn't import `v2_3` -- everything in the `spdx` package is an alias to the latest version (in this case, v2_3)

kzantow (commented):
Thanks @deitch -- I think this is pretty close, just a few items I've noted

kzantow:
I think this may fix #933 as well.

@deitch it also looks like the `TestEncodeDecodeEncodeCycleComparison` is failing for CycloneDX, which means we need to add decoding for the licenses with `name` [probably around here](https://github.com/anchore/syft/blob/main/syft/formats/common/cyclonedxhelpers/licenses.go#L31).

syft/formats/common/spdxhelpers/license.go
deitch: Easy enough. Done.

syft/formats/common/spdxhelpers/license.go
deitch: Yeah, I set it up, but clearly didn't use it everywhere. Done.

syft/formats/common/spdxhelpers/to_format_model.go
deitch: I was wondering about that. OK, I can get that.

deitch: Done

kzantow:
@deitch it looks like the last hold-up is the CycloneDX decoding noted above ☝️ -- I could help with this, if you like

deitch:
I assume it has something to do with [this](https://github.com/anchore/syft/blob/main/syft/formats/common/cyclonedxhelpers/licenses.go#L31) assuming that it should be `l.License.ID`, when it could be `.ID` _or_ `.Name`.

Yeah, fixing that gets the tests passing locally. I will push it out and see if CI is happy.

kzantow (approved):

kzantow:
Thanks much @deitch! This is a great addition!

deitch:
Excellent. Thanks for helping get it in. I am going to try and tackle a different issue soon with gopkg licenses.

deitch:
I will flag you on that issue, so I can try and submit for that as well.

kzantow:
Right -- properly parsing complex license expressions is not something Syft is doing today, I think it's _okay_ to just include these as LicenseRefs for now (it's better than excluding them)

deitch:
But you wouldn't object to it doing so properly, would you? I am close to having a PR ready for it.

PR #1542, opened 2023-02-06
source: when base is set, responsePath should be absolute

jedevc:
:arrow_up: Follow up to #1485 

When base is set, it should appear identically to when we scan the root filesystem - and as a result, the path should begin with the path separator.

E.g. when scanning the root `./target/` with the same base, `target/bin/busybox` should appear in the output as `/bin/busybox`, not as previously as `bin/busybox`.

Signed-off-by: Justin Chadwell <me@jedevc.com>


spiffcs:
Reran the AC tests since the failure was a connection issue and not a test failure

PR #1543, opened 2023-02-06
Allow specific versions of formats to be specified

### LINKED ISSUE #1519: Export specific format versions (SPDX)
Ability to export formats at a specific version, for example SPDX version 2.2.

**Why is this needed**:
Improve interoperability with other tools.

**Additional context**:
The leading thought is that this will be an extension of the format string, with an `@<version-matcher>` at the end, so you could run:
```
syft <something> -o spdx-json@2.2
```
which would output an SPDX 2.2 document version, or:
```
syft <something> -o spdx-tag-value@2
```
which would output the latest SPDX v2 document type (in this case would be 2.3).

Also: we will probably only be able to implement this for SPDX at the moment, since the [spdx/tools-golang](https://github.com/spdx/tools-golang) library now has support for conversions between different versions but other libraries do not (including Syft JSON itself!).

kzantow:
This PR adds support for including a version number along with formats, for example: `spdx@2.1` or `syft@6`.

Fixes #1519


cmd/syft/cli/options/packages.go
```diff
@@ -30,8 +30,8 @@ func (o *PackagesOptions) AddFlags(cmd *cobra.Command, v *viper.Viper) error {
 	cmd.Flags().StringVarP(&o.Scope, "scope", "s", cataloger.DefaultSearchConfig().Scope.String(),
 		fmt.Sprintf("selection of layers to catalog, options=%v", source.AllScopes))
 
-	cmd.Flags().StringArrayVarP(&o.Output, "output", "o", FormatAliases(table.ID),
-		fmt.Sprintf("report output format, options=%v", FormatAliases(syft.FormatIDs()...)))
+	cmd.Flags().StringArrayVarP(&o.Output, "output", "o", []string{string(table.ID)},
+		fmt.Sprintf("report output format, options=%v", formats.AllIDs()))
```
kzantow, line 34: Lingering question: should this function (`FormatAliases`) still exist to customize the format output to the client?

syft/formats.go
```diff
@@ -1,49 +0,0 @@
...
-	"github.com/anchore/syft/syft/formats/spdxtagvalue"
-	"github.com/anchore/syft/syft/formats/syftjson"
-	"github.com/anchore/syft/syft/formats/table"
-	"github.com/anchore/syft/syft/formats/template"
-	"github.com/anchore/syft/syft/formats/text"
-	"github.com/anchore/syft/syft/sbom"
-)
-
-// these have been exported for the benefit of API users
-// TODO: deprecated: now that the formats package has been moved to syft/formats, will be removed in v1.0.0
```
wagoodman, line 18: It seems like these could still remain in place until 1.0, honoring the current promise from the deprecation notice

kzantow: There isn't a nuance, I'll add these back in. 👍

syft/sbom/format.go
```diff
@@ -18,20 +19,48 @@ func (f FormatID) String() string {
 	return string(f)
 }
 
+type FormatVersion string
```
wagoodman, line 22: Why have a specific type for version?

kzantow: I'll update this to a string 👍

syft/formats/spdxjson/encoder.go
```diff
@@ -18,3 +21,20 @@ func encoder(output io.Writer, s sbom.SBOM) error {
...
+	var out v2_2.Document
+	err := convert.Document(doc, &out)
+	if err != nil {
+		return err
+	}
+
+	enc := json.NewEncoder(output)
+	// prevent > and < from being escaped in the payload
+	enc.SetEscapeHTML(false)
+	enc.SetIndent("", " ")
```
wagoodman, line 37: more of a side note for later, not necessarily for this PR, but we should start collecting instances of these and have a common utility function somewhere in internal.

kzantow: I moved this to a common function in the `spdxjson/encoder` for now, we can always refactor more if needed.

syft/sbom/format.go
```diff
@@ -18,20 +19,41 @@ func (f FormatID) String() string {
 	return string(f)
 }
 
+const AnyVersion = ""
```
wagoodman, line 22: I feel that this belongs in the `format` package, so it would read `format.AnyVersion` instead of `sbom.AnyVersion`

wagoodman: I forgot there is `formats` which is about capabilities and not definitions. I take back this comment

syft/sbom/format.go
```diff
@@ -18,20 +19,41 @@ func (f FormatID) String() string {
...
+
 type Format interface {
 	ID() FormatID
+	IDs() []FormatID
+	Version() string
 	Encode(io.Writer, SBOM) error
 	Decode(io.Reader) (*SBOM, error)
 	Validate(io.Reader) error
+	String() string
```
wagoodman, line 31: nit: `fmt.Stringer` could be embedded here instead

kzantow: Done 👍

PR #1567, opened 2023-02-13
fix: improve CPE and upstream generation logic for Alpine packages

westonsteimel:
CPE-based matching is particularly important for the Alpine and Wolfi ecosystems since the NVD does serve as the authoritative advisory source for vulnerabilities.  This improves the generation of CPE's and `upstream` names which are later used to provide matches in grype.  In particular, it uses some seemingly well-adopted prefixes for python and ruby packages to ensure syft tailors the generation to specific logic for those ecosystems.


kzantow (approved):
LGTM 👍

PR #1570, opened 2023-02-14
Enable SBOM conversions from STDIN

wagoodman:
A small addition to the `convert` command to allow for taking SBOMs from stdin:
```
cat my.sbom | syft convert - -o spdx-json
```


kzantow (approved):
👍

PR #1580, opened 2023-02-17
Add ruby binary classifier

noqcks:
Adding a ruby binary pkg classifier

I think this would be nice to have, however there is one caveat.... 

For older versions of ruby, matching is precise, we're matching on this text and return version 2.0.0

```
ruby 2.0.0p648 (2015-12-16 revision 53162)
```

For newer versions of ruby, we can't match the patch version, so even if ruby 2.5.1 is installed, we only return version 2.5. This is because we're matching on the library file libruby.so from this text string

```
__deregister_frame_info_ITM_deregisterTMCloneTable__cxa_finalize__register_frame_info_ITM_registerTMCloneTableruby_run_node_finiruby_initruby_optionsruby_init_stackruby_sysinitsetlocale__libc_start_mainlibruby.so.2.5libc.musl-x86_64.so.1_start_c__bss_start_edata_end/usr/local/
```

If the latter is unacceptable, we can just match on versions we can find the full semver


kzantow:
Thanks for the contribution @noqcks! Could you link to a few images that this should work with?

noqcks:
```
ruby:2.0
ruby:2.5
ruby:3.2
ruby:3.2.1-buster
```

PR #1582, opened 2023-02-17
Fix cataloger selection to be more specific

### LINKED ISSUE #1573: Cataloger filtering cross matches wrong catalogers
Adding the following test to https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/cataloger_test.go#L27 shows that when requesting only `java-cataloger`, the `javascript-package-cataloger` and `javascript-lock-cataloger` are both enabled, which IMHO is wrong.

```
                {
			name: "don't cross match",
			patterns: []string{
				"java-cataloger",
			},
			catalogers: []string{
				"javascript-package-cataloger",
				"java-cataloger",
			},
			want: []string{
				"java-cataloger",
			},
		},
```

wagoodman:
Today the cataloger selection allows for partial name matches, so input such as `go` will select all go catalogers:
- **go**-mod-file-cataloger
- **go**-module-binary-cataloger

Unfortunately the current implementation takes a simplistic approach (a simple substring check) which causes additional false selections across ecosystems, which would most likely be unintentional. For instance, the same `go` search would additionally select:
- rust-car**go**-lock-cataloger
- car**go**-auditable-binary-cataloger

This PR applies a workaround to make this behavior a little better before https://github.com/anchore/syft/pull/1383 lands in a release (which fixes this behavior entirely).

The new behavior restricts the substring check to only allow for full words, separated by `-`. This still allows for the current selection behavior today but fixes all of the (incorrect) cross selections I could find.

Closes #1573

### DIFF
--- a/syft/pkg/cataloger/cataloger.go
+++ b/syft/pkg/cataloger/cataloger.go
@@ -161,9 +161,29 @@ func contains(enabledPartial []string, catalogerName string) bool {
 		if partial == "" {
 			continue
 		}
-		if strings.Contains(catalogerName, partial) {
+		if hasFullWord(partial, catalogerName) {
 			return true
 		}
 	}
 	return false
+}
+
+func hasFullWord(targetPhrase, candidate string) bool {
+	if targetPhrase == "cataloger" || targetPhrase == "" {
+		return false
+	}
+	start := strings.Index(candidate, targetPhrase)
+	if start == -1 {
+		return false
+	}
+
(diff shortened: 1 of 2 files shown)


spiffcs (approved):

spiffcs:
@wagoodman code looks good - there is a weird error in CI whee the the AT won't pass

wagoodman:
CI issue workaround applied, see https://github.com/anchore/syft/pull/1584

PR #1598, opened 2023-02-21
feat: set cosign attest predicate type based on Syft output type

Nirusu:
In #1533, I wanted to remove the hardcoded "custom" predicate type passed to Cosign. This seemed to fix some of the issues I was seeing, though to be honest at this point I don't know exactly why 😅 

However, what I did not anticipate is that there's still no way to manually set a predicate type. Generally, I think it makes sense to set the predicate type on the type we are passing as predicate to Cosign.

Thus, here my attempt at this. 

Please mind, this is untested so far. Wanted to get general feedback on this first (or if someone can quickly test this and verify the attestation, that'd be cool).

Also, if someone can tell me how I can properly reference the output type in the switch statement rather than just using string comparison based on the user input here without major refactoring, I'd be happy to hear about that :)

### DIFF
--- a/cmd/syft/cli/attest/attest.go
+++ b/cmd/syft/cli/attest/attest.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/wagoodman/go-partybus"
 	"github.com/wagoodman/go-progress"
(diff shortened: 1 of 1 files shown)


kzantow:
Hi @Nirusu -- could you update to the latest `main`? It should fix the failing test for you.

Nirusu:
Done.

spiffcs (dismissed):

spiffcs (approved):

spiffcs:
@Nirusu I tested it and think we can merge as is -- I'll follow this up with maybe a small refactor that uses the correct values for the input types. Had an issue checking out the branch using `gh` command so no harm in merge and then follow up =)

PR #1601, opened 2023-02-21
feat: rust toolchain binary cataloger

westonsteimel:
Adds binary classifier for the rust toolchain for linux and mac


syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -187,4 +187,24 @@ var defaultClassifiers = []classifier{
...
+		Package: "rust",
+		PURL:    mustPURL("pkg:generic/rust@version"),
+		CPEs:    singleCPE("cpe:2.3:a:rust-lang:rust:*:*:*:*:*:*:*:*"),
+	},
+	{
+		Class:    "rust-standard-library-macos",
+		FileGlob: "**/libstd-*.dylib",
+		EvidenceMatcher: fileContentsVersionMatcher(
+			// c 1.48.0 (7eac88abb 2020-11-16)
+			`(?m)c (?P<version>[0-9]+(\.[0-9]+)?(\.[0-9]+)) \(\w+ \d{4}\-\d{2}\-\d{2}\)`),
```
kzantow, line 205: Is this the most specific we can be with the file globs? I wonder if `libstd-*.dylib` and `libstd-*.so` might match more than we really want? For the linux case, I think the `rustc version` is probably pretty specific, but this macOS pattern seems like it _might_ match things unexpectedly?

kzantow (approved):
It might be nice to cut down the size of some of the test fixtures (some are in the MB range)

syft/pkg/cataloger/binary/default_classifiers.go
westonsteimel: No, that is the most specific possible

westonsteimel: The filenames are things like `libstd-d05561573e32b48f.dylib`

westonsteimel: The whole point being you can have many toolchain versions installed simultaneously

westonsteimel: I have a test image at https://github.com/anchore/test-images/blob/main/containers/analyzer-rust-toolchain/Dockerfile

westonsteimel: @kzantow , I tried `**/libstd-[0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z].dylib` but it doesn't seem to work

westonsteimel: It seems like that is valid for globstar but some logic in stereoscope search indexing is causing it to fail.  I'll look into it some more later, but perhaps @wagoodman might have ideas on it.

wagoodman: that may be worth capturing as an issue in stereoscope, it should have at least fell through to the expensive glob case if it didn't know how to handle it (which is seems like it didn't?).

wagoodman (changes requested):
> ...some are in the MB range....

it seems like there may be an extra addition to the gitignore needed for the busybox image fixture tar

westonsteimel:
> > ...some are in the MB range....
> 
> it seems like there may be an extra addition to the gitignore needed for the busybox image fixture tar

I think it might already be excluded - I had to force add my test cases since they'd normally be excluded and must have accidentally picked up that as well

kzantow (approved):
The `/rust-1.67.1/toolchains/stable-x86_64-unknown-linux-musl/lib/libstd-86aefecbddda356d.so` is still 1.25 MB, it would be nice to just copy the small (maybe < 100 bytes) binary bits that are required for matching. Other than that, 👍

syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -196,4 +196,24 @@ var defaultClassifiers = []classifier{
 		Package: "postgresql",
 		PURL:    mustPURL("pkg:generic/postgresql@version"),
 	},
+	{
+		Class:    "rust-standard-library-linux",
+		FileGlob: "**/libstd-????????????????.so",
```
wagoodman, line 201: good call, that is more specific and not too crazy for the index parsing in stereoscope 🙌

wagoodman (approved):

PR #1612, opened 2023-02-23
Add support for SUPPORT_END in distro

noqcks:
`SUPPORT_END` is an optional parameter in `/etc/os-release` that describes "The date at which support for this version of the OS ends." https://www.freedesktop.org/software/systemd/man/os-release.html

This would be useful to have for a project I'm building around finding EOL dates for software, which uses Syft to generate SBOMs - https://github.com/noqcks/xeol


kzantow (approved):
LGTM 👍

PR #1615, opened 2023-02-23
read relative etc/apk/repositories for alpine version when no OS provided

### LINKED ISSUE #1572: purl for apk packages missing when installed db file is not in root
**What happened**:

* If `lib/apk/db/installed` is in the _root_ of the filesystem being scanned, then syft includes everything, including the purl.
* If `lib/apk/db/installed` is _below_ the root of the filesystem being scanned, then syft includes everything, except the purl

In both cases it reads the `installed` file and parses it, yet for some reason, if not in the root, it misses the purl.

Interestingly, if the package is _also_ in the root, i.e. in addition to the above, it is in `lib/apk/db/installed`, then it adds the `purl` for all of them.

Add the apk purl for all packages, wherever it finds them.

1. Create a Dockerfile with a simple package, e.g. curl:
```dockerfile
FROM alpine:3.17
RUN apk add curl
```
1. Build it to a tmpdir, e.g. `/tmp/oci`:
```sh
$ docker buildx build -t spdx-test --output type=local,dest=/tmp/oci .
```
1. Run syft on the single dir, it works: `syft packages -o spdx-json /tmp/oci` - WORKS
1. Make a directory for it one layer down and copy it: `mkdir -p /tmp/single/one && tar -C /tmp/oci -cvf - . | (cd /tmp/single/one; tar -xvf - )`
1. Run syft on the upper layer, see that it all works except missing the purl: `syft packages -o spdx-json /tmp/single` - FAILS
1. Repeat with two parallel deeper directories: `mkdir -p /tmp/deep/one && mkdir -p /tmp/deep/two && tar -C /tmp/oci -cvf - . | (cd /tmp/deep/one; tar -xvf - ) && tar -C /tmp/oci -cvf - . | (cd /tmp/deep/two; tar -xvf - )
1. Run syft on the upper layer, see that it captures both of the instances of the curl package, but still no purl: `syft packages -o spdx-json /tmp/deep` - FAILS
1. Create a copy on the root level in addition to deeper, see that it works: `mkdir -p /tmp/staggered && mkdir -p /tmp/staggered/one && mkdir -p /tmp/staggered/two &&  tar -C /tmp/oci -cvf - . | (cd /tmp/staggered/one; tar -xvf - ) &&  tar -C /tmp/oci -cvf - . | (cd /tmp/staggered/two; tar -xvf - ) &&  tar -C /tmp/oci -cvf - . | (cd /tmp/staggered; tar -xvf - )`
1. Run syft on the root level, see that it works: `syft packages -o spdx-json /tmp/stagged`

Output from the first (limited to curl package for readability):

```json
  {
   "name": "curl",
   "SPDXID": "SPDXRef-Package-apk-curl-262ddcbe57a87ed8",
   "versionInfo": "7.87.0-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://curl.se/",
   "sourceInfo": "acquired package info from APK DB: lib/apk/db/installed",
   "licenseConcluded": "curl",
   "licenseDeclared": "curl",
   "description": "URL retrival utility and library",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:curl:curl:7.87.0-r1:*:*:*:*:*:*:*"
    },
    {
     "referenceType": "purl",
     "referenceLocator": "pkg:apk/alpine/curl@7.87.0-r1?arch=aarch64&upstream=curl&distro=alpine-3.17.2"
    }
   ]
  },
```

From the single but one layer deep:

```json
  {
   "name": "curl",
   "SPDXID": "SPDXRef-Package-apk-curl-d6e41f455208de3",
   "versionInfo": "7.87.0-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://curl.se/",
   "sourceInfo": "acquired package info from APK DB: oci/lib/apk/db/installed",
   "licenseConcluded": "curl",
   "licenseDeclared": "curl",
   "description": "URL retrival utility and library",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:curl:curl:7.87.0-r1:*:*:*:*:*:*:*"
    }
   ]
  },
```

From two parallel but deep:

```json
  {
   "name": "curl",
   "SPDXID": "SPDXRef-Package-apk-curl-a9c664463a5e13cd",
   "versionInfo": "7.87.0-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://curl.se/",
   "sourceInfo": "acquired package info from APK DB: one/lib/apk/db/installed",
   "licenseConcluded": "curl",
   "licenseDeclared": "curl",
   "description": "URL retrival utility and library",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:curl:curl:7.87.0-r1:*:*:*:*:*:*:*"
    }
   ]
  },
  {
   "name": "curl",
   "SPDXID": "SPDXRef-Package-apk-curl-f9087650e668eed7",
   "versionInfo": "7.87.0-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://curl.se/",
   "sourceInfo": "acquired package info from APK DB: two/lib/apk/db/installed",
   "licenseConcluded": "curl",
   "licenseDeclared": "curl",
   "description": "URL retrival utility and library",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:curl:curl:7.87.0-r1:*:*:*:*:*:*:*"
    }
   ]
  },
```

From staggered, i.e. two deep and one in root:

```json
  {
   "name": "curl",
   "SPDXID": "SPDXRef-Package-apk-curl-262ddcbe57a87ed8",
   "versionInfo": "7.87.0-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://curl.se/",
   "sourceInfo": "acquired package info from APK DB: lib/apk/db/installed",
   "licenseConcluded": "curl",
   "licenseDeclared": "curl",
   "description": "URL retrival utility and library",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:curl:curl:7.87.0-r1:*:*:*:*:*:*:*"
    },
    {
     "referenceType": "purl",
     "referenceLocator": "pkg:apk/alpine/curl@7.87.0-r1?arch=aarch64&upstream=curl&distro=alpine-3.17.2"
    }
   ]
  },
  {
   "name": "curl",
   "SPDXID": "SPDXRef-Package-apk-curl-a9c664463a5e13cd",
   "versionInfo": "7.87.0-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://curl.se/",
   "sourceInfo": "acquired package info from APK DB: one/lib/apk/db/installed",
   "licenseConcluded": "curl",
   "licenseDeclared": "curl",
   "description": "URL retrival utility and library",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:curl:curl:7.87.0-r1:*:*:*:*:*:*:*"
    },
    {
     "referenceType": "purl",
     "referenceLocator": "pkg:apk/alpine/curl@7.87.0-r1?arch=aarch64&upstream=curl&distro=alpine-3.17.2"
    }
   ]
  },
  {
   "name": "curl",
   "SPDXID": "SPDXRef-Package-apk-curl-f9087650e668eed7",
   "versionInfo": "7.87.0-r1",
   "originator": "Person: Natanael Copa \u003cncopa@alpinelinux.org\u003e",
   "downloadLocation": "https://curl.se/",
   "sourceInfo": "acquired package info from APK DB: two/lib/apk/db/installed",
   "licenseConcluded": "curl",
   "licenseDeclared": "curl",
   "description": "URL retrival utility and library",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:curl:curl:7.87.0-r1:*:*:*:*:*:*:*"
    },
    {
     "referenceType": "purl",
     "referenceLocator": "pkg:apk/alpine/curl@7.87.0-r1?arch=aarch64&upstream=curl&distro=alpine-3.17.2"
    }
   ]
  },
```

**Anything else we need to know**

It was @dautovri who first figured this out and explained it. Credit where it is due.

**Environment**:
- Output of `syft version`: latest commit
- OS (e.g: `cat /etc/os-release` or similar): ran on macOS and Linux

deitch:
fixes #1572 

Sort of

The way apkdb works now, if it cannot find an `/etc/os-release` (or similar files on a fixed list), and if the ID is not `"alpine"`, it still reports on the package, but no purl is provided.

This has several problems.

* not everything will have an `etc/os-release`. This especially runs true when you are looking at a filesystem with multiple containers, or perhaps you have modified the `os-release` because your OS isn't alpine, but the packages in the embedded container are.
* there are multiple potential sources for packages; `os-release` is not necessarily the best place to find out where it came from.
* packages can come from multiple parallel locations, i.e. `/etc/apk/repositories` can have alpinelinux.org but also other places
* the version used for the packages might entirely be different from the version of the OS

This tries to address some of it. It does so very cautiously. If it cannot find an `/etc/os-release`, i.e. `linux.Release` is nil, then then it tries to read the `etc/apk/repositories` relative to where it found `lib/apk/db/installed` (so if it is embedded deep in the filesystem, it still can find it). If it finds `alpinelinux.org` in there, it sets it up as if running on alpine, and takes the version from there.

It is an improvement over the current design, but could be improved further.

* it should _prefer_ `etc/apk/repositories` over `os-release`; this uses it only if release still is nil
* it should read multiple options from `repositories`, but that will have to wait, since there is no way to correlate those with a particular package in `installed`

This is a real if imperfect improvement over the current state.

apk team is looking at improving this chain-of-custody, including providing a purl in the `installed` file. Things will get much easier then. See [this issue](https://gitlab.alpinelinux.org/alpine/apk-tools/-/issues/10875). Until then, this can help.

### DIFF
--- a/syft/pkg/cataloger/apkdb/cataloger_test.go
+++ b/syft/pkg/cataloger/apkdb/cataloger_test.go
@@ -24,6 +24,7 @@ func TestCataloger_Globs(t *testing.T) {
 			pkgtest.NewCatalogTester().
 				FromDirectory(t, test.fixture).
 				ExpectsResolverContentQueries(test.expected).
+				IgnoreUnfulfilledPathResponses("etc/apk/repositories").
 				TestCataloger(t, NewApkdbCataloger())
 		})
 	}
--- a/syft/pkg/cataloger/apkdb/parse_apk_db.go
+++ b/syft/pkg/cataloger/apkdb/parse_apk_db.go
@@ -23,7 +23,7 @@ import (
 var _ generic.Parser = parseApkDB
 
 var (
-	repoRegex = regexp.MustCompile(`(?m)^https://.*\.alpinelinux\.org/alpine/v([^\/]+)/([a-zA-Z0-9_]+)$`)
+	repoRegex = regexp.MustCompile(`(?m)^https://.*\.alpinelinux\.org/alpine/v([^/]+)/([a-zA-Z0-9_]+)$`)
 )
 
 // parseApkDB parses packages from a given APK installed DB file. For more
@@ -31,33 +31,6 @@ var (
 //
 //nolint:funlen,gocognit
 func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
-	// find the repositories file from the the relative directory of the DB file
-	var releases []linux.Release
-	if resolver != nil {
-		reposLocation := path.Clean(path.Join(path.Dir(reader.Location.RealPath), "../../../etc/apk/repositories"))
-		paths, err := resolver.FilesByPath(reposLocation)
-		if err != nil {
-			log.Debugf("unable to find repositories file at %s: %w", reposLocation, err)
-		}
-		if err == nil && len(paths) > 0 {
-			reposContent, err := resolver.FileContentsByLocation(paths[0])
-			if err == nil {
-				reposB, err := io.ReadAll(reposContent)
-				if err == nil {
-					parts := repoRegex.FindAllStringSubmatch(string(reposB), -1)
-					for _, part := range parts {
-						if len(part) >= 3 {
-							releases = append(releases, linux.Release{
-								Name:      "Alpine Linux",
-								ID:        "alpine",
-								VersionID: part[1],
-							})
-						}
-					}
-				}
-			}
-		}
-	}
 	scanner := bufio.NewScanner(reader)
 
 	var apks []pkg.ApkMetadata
@@ -139,8 +112,13 @@ func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader s
 	// but we now have no way of handling different repository sources. On the other hand,
 	// we never could before this. At least now, we can handle some.
 	// This should get fixed with https://gitlab.alpinelinux.org/alpine/apk-tools/-/issues/10875
-	if r == nil && len(releases) > 0 {
-		r = &releases[0]
+	if r == nil {
+		// find the repositories file from the relative directory of the DB file
+		releases := findReleases(resolver, reader.Location.RealPath)
+
+		if len(releases) > 0 {
+			r = &releases[0]
+		}
 	}
 
 	pkgs := make([]pkg.Package, 0, len(apks))
@@ -149,6 +127,58 @@ func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader s
 	}
 
 	return pkgs, discoverPackageDependencies(pkgs), nil
+}
+
+func findReleases(resolver source.FileResolver, dbPath string) []linux.Release {
+	if resolver == nil {
+		return nil
+	}
+
+	reposLocation := path.Clean(path.Join(path.Dir(dbPath), "../../../etc/apk/repositories"))
+	locations, err := resolver.FilesByPath(reposLocation)
+	if err != nil {
+		log.Tracef("unable to find APK repositories file %q: %+v", reposLocation, err)
+		return nil
+	}
+
+	if len(locations) == 0 {
+		return nil
+	}
+	location := locations[0]
+
+	reposReader, err := resolver.FileContentsByLocation(location)
+	if err != nil {
+		log.Tracef("unable to fetch contents for APK repositories file %q: %+v", reposLocation, err)
+		return nil
+	}
+
+	return parseReleasesFromAPKRepository(source.LocationReadCloser{
+		Location:   location,
+		ReadCloser: reposReader,
+	})
+}
+
+func parseReleasesFromAPKRepository(reader source.LocationReadCloser) []linux.Release {
+	var releases []linux.Release
+
+	reposB, err := io.ReadAll(reader)
+	if err != nil {
+		log.Tracef("unable to read APK repositories file %q: %+v", reader.Location.RealPath, err)
+		return nil
+	}
+
+	parts := repoRegex.FindAllStringSubmatch(string(reposB), -1)
+	for _, part := range parts {
+		if len(part) >= 3 {
+			releases = append(releases, linux.Release{
+				Name:      "Alpine Linux",
+				ID:        "alpine",
+				VersionID: part[1],
+			})
+		}
+	}
+
+	return releases
 }
 
 func parseApkField(line string) *apkField {
--- a/syft/pkg/cataloger/apkdb/parse_apk_db.go
+++ b/syft/pkg/cataloger/apkdb/parse_apk_db.go
@@ -3,7 +3,9 @@ package apkdb
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -20,11 +22,42 @@ import (
 // integrity check
 var _ generic.Parser = parseApkDB
 
+var (
+	repoRegex = regexp.MustCompile(`(?m)^https://.*\.alpinelinux\.org/alpine/v([^\/]+)/([a-zA-Z0-9_]+)$`)
+)
+
 // parseApkDB parses packages from a given APK installed DB file. For more
 // information on specific fields, see https://wiki.alpinelinux.org/wiki/Apk_spec.
 //
-//nolint:funlen
-func parseApkDB(_ source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+//nolint:funlen,gocognit
+func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	// find the repositories file from the the relative directory of the DB file
+	var releases []linux.Release
+	if resolver != nil {
+		reposLocation := path.Clean(path.Join(path.Dir(reader.Location.RealPath), "../../../etc/apk/repositories"))
+		paths, err := resolver.FilesByPath(reposLocation)
+		if err != nil {
+			log.Debugf("unable to find repositories file at %s: %w", reposLocation, err)
+		}
+		if err == nil && len(paths) > 0 {
+			reposContent, err := resolver.FileContentsByLocation(paths[0])
+			if err == nil {
+				reposB, err := io.ReadAll(reposContent)
+				if err == nil {
+					parts := repoRegex.FindAllStringSubmatch(string(reposB), -1)
+					for _, part := range parts {
+						if len(part) >= 3 {
+							releases = append(releases, linux.Release{
+								Name:      "Alpine Linux",
+								ID:        "alpine",
+								VersionID: part[1],
+							})
+						}
+					}
+				}
+			}
+		}
+	}
 	scanner := bufio.NewScanner(reader)
 
 	var apks []pkg.ApkMetadata
@@ -100,6 +133,14 @@ func parseApkDB(_ source.FileResolver, env *generic.Environment, reader source.L
 	var r *linux.Release
 	if env != nil {
 		r = env.LinuxRelease
+	}
+	// this is somewhat ugly, but better than completely failing when we can't find the release,
+	// e.g. embedded deeper in the tree, like containers or chroots.
+	// but we now have no way of handling different repository sources. On the other hand,
+	// we never could before this. At least now, we can handle some.
+	// This should get fixed with https://gitlab.alpinelinux.org/alpine/apk-tools/-/issues/10875
+	if r == nil && len(releases) > 0 {
+		r = &releases[0]
 	}
 
 	pkgs := make([]pkg.Package, 0, len(apks))
--- a/syft/pkg/cataloger/apkdb/parse_apk_db_test.go
+++ b/syft/pkg/cataloger/apkdb/parse_apk_db_test.go
@@ -1,8 +1,10 @@
 package apkdb
 
 import (
+	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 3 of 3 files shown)


syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -20,11 +22,41 @@ import (
...
-//nolint:funlen
-func parseApkDB(_ source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+//nolint:funlen,gocognit
+func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	// find the repositories file from the the relative directory of the DB file
+	var repos []string
+	if resolver != nil {
+		reposLocation := path.Clean(path.Join(path.Dir(reader.Location.RealPath), "../../../etc/apk/repositories"))
+		paths, err := resolver.FilesByPath(reposLocation)
+		if err == nil && len(paths) > 0 {
```
kzantow, line 45: If `err != nil` it might be good to at least `log.Debugf("%v", err)` -- there should never be an error for file not found, but if something else happens it would be good to know when running with `-vv`, at least

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -20,11 +22,41 @@ import (
...
 
+var (
+	repoRegex         = regexp.MustCompile(`^https://.*\.alpinelinux\.org/alpine/v([^\/]+)/([a-zA-Z0-9_]+)$`)
+	newlineSplitRegex = regexp.MustCompile(`[\r \s\n]+`)
+)
+
+func init() {
+	repoRegex.Longest()
+	newlineSplitRegex.Longest()
+}
```
kzantow, line 33: I think this could be simplified to use a multiline modifier (`(?m)`) in the original regex, e.g:

```go
repoRegex = regexp.MustCompile(`(?m)^https://.*\.alpinelinux\.org/alpine/v([^\/]+)/([a-zA-Z0-9_]+)$`)
```

... and since it reads the entire line (the `^` and `$`), it probably doesn't need the `.Longest()` modifier.

see: https://regex101.com/r/goXXNl/1

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -20,11 +22,41 @@ import (
...
 // parseApkDB parses packages from a given APK installed DB file. For more
 // information on specific fields, see https://wiki.alpinelinux.org/wiki/Apk_spec.
 //
-//nolint:funlen
-func parseApkDB(_ source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+//nolint:funlen,gocognit
+func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	// find the repositories file from the the relative directory of the DB file
+	var repos []string
+	if resolver != nil {
```
kzantow, line 42: I think this whole block can be moved down into the `r == nil` check to only occur if there isn't a linux release identified. As noted in your assessment of this issue, it might be valid to read the repositories in order to generate a correct PURL, but since there is no way to correlate these entries to the apkdb currently, and we are just looking for `alpinelinux.org`, we could avoid doing this if we already identified a release for now.

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -20,11 +22,41 @@ import (
...
+			reposContent, err := resolver.FileContentsByLocation(paths[0])
+			if err == nil {
+				reposB, err := io.ReadAll(reposContent)
+				if err == nil {
+					reposDirect := newlineSplitRegex.Split(string(reposB), -1)
+					for _, repo := range reposDirect {
+						if repo != "" {
+							repos = append(repos, repo)
+						}
+					}
```
kzantow, line 55: If simplified to the single regex, this would just match and extract the version, like what is happening in the section below

kzantow (changes requested):
Overall, I'm good with this approach; I left some implementation detail notes (consolidating into the `r == nil` section, simpler regex handling). Also, could you add some sort of test for this?

syft/pkg/cataloger/apkdb/parse_apk_db.go
deitch: OK, we can add that. My thought to all of this was, "if we find the file, we use it, if not, we just keep going." Logs don't violate that principle, though. Will add it.

syft/pkg/cataloger/apkdb/parse_apk_db.go
deitch: _That_ is a good approach. Pity that I deleted the play.golang where I was testing all of it. Oh well.

deitch:
Addressed all of your points, take a look please.

syft/pkg/cataloger/apkdb/parse_apk_db.go
deitch: I am still holding back on the thought that this is valuable on its own, and at some point, we will _prefer_ this over the OS. If it doesn't hurt, let's leave it up here?

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -20,11 +22,47 @@ import (
...
-func parseApkDB(_ source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+//nolint:funlen,gocognit
+func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	// find the repositories file from the the relative directory of the DB file
+	var repos []linux.Release
+	if resolver != nil {
+		reposLocation := path.Clean(path.Join(path.Dir(reader.Location.RealPath), "../../../etc/apk/repositories"))
+		paths, err := resolver.FilesByPath(reposLocation)
+		if err != nil {
+			log.Debugf("unable to find repositories file at %s: %w", err)
```
kzantow, line 45: This needs another variable for the `%s`, I think it would be:

```go
log.Debugf("unable to find repositories file at %s: %w", reposLocation , err)
```

kzantow (dismissed):
Sorry for the lengthy comment, but I think I've outlined how we can make this change a bit cleaner without the need to make "fake" linux.Release objects, WDYT?

syft/pkg/cataloger/apkdb/parse_apk_db.go
deitch: Ah oops! Fixing.

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -20,11 +22,47 @@ import (
 // integrity check
 var _ generic.Parser = parseApkDB
 
+var (
+	repoRegex         = regexp.MustCompile(`(?m)^https://.*\.alpinelinux\.org/alpine/v([^\/]+)/([a-zA-Z0-9_]+)$`)
+	newlineSplitRegex = regexp.MustCompile(`[\r \s\n]+`)
```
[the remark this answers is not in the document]
deitch: Yup, done.

syft/pkg/cataloger/apkdb/cataloger_test.go
```diff
@@ -24,6 +24,7 @@ func TestCataloger_Globs(t *testing.T) {
 			pkgtest.NewCatalogTester().
 				FromDirectory(t, test.fixture).
 				ExpectsResolverContentQueries(test.expected).
+				IgnoreUnfulfilledPathResponses("etc/apk/repositories").
```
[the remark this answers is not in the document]
deitch: Correct. Right now, it is to ensure that the test knows, "I do not care that this path is unresolved."

syft/pkg/cataloger/apkdb/parse_apk_db.go
```diff
@@ -20,11 +22,47 @@ import (
...
+
 // parseApkDB parses packages from a given APK installed DB file. For more
 // information on specific fields, see https://wiki.alpinelinux.org/wiki/Apk_spec.
 //
-//nolint:funlen
-func parseApkDB(_ source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+//nolint:funlen,gocognit
+func parseApkDB(resolver source.FileResolver, env *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	// find the repositories file from the the relative directory of the DB file
+	var repos []linux.Release
```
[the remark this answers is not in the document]
deitch: Good point. Fixed.

deitch:
Oh, yeah, I like the namespace approach. Going to give it a shot quickly now.

deitch:
OK, I tried. The problem is that `apkdb.parseApkDB()` calls `apkdb.newPackage()`, which calls `apkdb.packageURL()`, which calls `syft/pkg.PURLQualifiers()`, which is more generically shared, and expects to receive a `*linux.Release`. It then uses that to receive the ID and check for a version as well. So replacing `*linux.Release` with `namespace` not only doesn't work, but it breaks the ability to pass more information out.

deitch:
For now, I left is as `var releases []linux.Release`, but open to what you have in mind.

If you ask me, the correct thing to do here is to have something entirely different passed to [`PURLQualifiers()`](https://github.com/anchore/syft/blob/main/syft/pkg/url.go#L24):

```go
type PackageRelease interface {
  func ID() string
  func VersionID() string
  func BuildID() string
}

func PURLQualifiers(vars map[string]string, release PackageRelease) (q packageurl.Qualifiers) {
```

Some packages are tied to a linux release, others not so much.

That requires changing everything that calls it, so it might be a bit much for this. I would sooner have this in, and then we can worry about the right abstraction for `PURLQualifiers()`

wagoodman:
@deitch solid improvement 💯 . 

I pushed a small refactor that decomposed the additions into smaller functions and swapped out multiple nesting of conditions with guard clauses instead. You can treat this commit as a suggestion, if you really don't like it or would prefer a different approach feel free to force-push over it / revert it.

The upside of the decomposition is to allow for easier unit testing of just the apk repository file parsing (now the `parseReleasesFromAPKRepository` function). Can you add a unit test that exercises parsing the repository file?

deitch:
Yeah, that's a fine breakdown; smaller is easier whenever possible.

deitch:
Added a test of that. Back at you.

wagoodman (approved):

PR #1610, opened 2023-02-23
fix: determine upstream for apk version streams

westonsteimel:
Determines better upstream package name for version-stream apk packages:

Examples:

- postgresql-13 -> postgresql
- postgresql15 -> postgresql
- go-1.19 -> go
- perl100.55 -> perl

PR #1618, opened 2023-02-24
fix: stricter apk upstream parsing regex

westonsteimel:
This is a follow-up to https://github.com/anchore/syft/pull/1610.  I had updated the regex match string locally to fix up some edge cases but apparently forgot to push them.


westonsteimel:
Superseded by https://github.com/anchore/syft/pull/1619

PR #1630, opened 2023-02-28
Support scanning license files in golang packages over the network

### LINKED ISSUE #1056: Unable to identify license on Golang packages imported by URL
**What happened**:

Given a Golang package that imports from a URL, licenseConcluded in the associated SPDX file is "none". The remote package has a clearly identifiable license file.

The remote repo (https://github.com/aws/aws-sdk-go) has a LICENSE.txt. I expected that to be deferenced and used.

I have a golang file:
`./test.go`

Containing a remote import:
```
package test
import (
        awsSDK "github.com/aws/aws-sdk-go/aws"
```

I process the file using the syft cli and generate spdx. The spdx code generated is:
```

 "packages": [
  {
   "SPDXID": "SPDXRef-320fa0ac8fde1360",
   "name": "github.com/aws/aws-sdk-go",
   "licenseConcluded": "NONE",

```

**Environment**:
- Output of `syft version`:

syft 0.46.3

- 
OS X 11.6

** Screenshot**

image

deitch:
Fixes #1056 

As discussed with @kzantow 

This checks in `$GOPATH/pkg/` for a package@version, and, if it finds it, reads the licenses from there. If not found, it falls back to reading the package from the internet via the official go proxy https://proxy.golang.org

In addition, it has two new CLI flags: `go-fetch bool`  and `--go-proxy string`. The `fetch` is supposed to enable fetching over the Internet, default to `false`; the `proxy` is supposed to override the default.

The CLI flags are not wired up, as I could not quite figure out how to pass them all the way through to the go cataloger. I managed to get them onto the `app *config.Application`, but there is a bunch of steps down. Some pointers would be helpful.


deitch:
I do not understand why the CLI tests fail. 🤷‍♂️

deitch:
At the request of @kzantow , I am splitting this one up into 2 PRs.

The first #1645 , only checks local `GOPATH/mod` for packages to find licenses. After that one is in, we will open a second PR, adding a CLI option to reach out to the Internet if a package is not found locally.

deitch:
Actually, I mostly managed to get it in place. Still needs some help.

README.md
```diff
@@ -504,6 +504,15 @@ golang:
    # SYFT_GOLANG_LOCAL_MOD_CACHE_DIR env var
    local-mod-cache-dir: ""
 
+   # search for go package licences by retrieving the package from a network proxy
+   # SYFT_GOLANG_SEARCH_REMOTE_LICENSES env var
+   search-remote-licenses: false
+
+   # remote proxy to use when retrieving go packages from the network,
+   # if unset this defaults to $GOPROXY followed by https://proxy.golang.org
```
kzantow, line 512: We may want to support the `GOPRIVATE` env var: https://github.com/goproxyio/goproxy.io/blob/master/content/docs/GOPRIVATE-env.md

deitch: I think we should, but I would leave it for a later iteration.

wagoodman: There are a few interactions with GOPROXY, GONOPROXY, and GOPRIVATE (https://go.dev/ref/mod#environment-variables).

1. GOPROXY not set or set to `direct` , which means download from the source repo
2. GOPROXY set to `off` which would disallow network access (effectively disabling this feature)
3. GOPROXY could have a `file:///` prefix which means to search the filesystem
4. GOPROXY could have multiple values specified in a comma separated list, though there is special processing where if `direct` in the list is found, then no further sources are processed
5. GOPRIVATE can select certain modules based off of a glob prefix such that those sources are always considered as `direct` and the remaining modules follow the standard GOPROXY value (also accepts comma delimited list)
6. GONOPROXY (note the "no") defaults to GOPRIVATE if not set, but can be explicitly overridden and even set to `none` (I think) to ensure everything is accessed via GOPROXY configuration.

I think just having `remote-proxy` configuration could lead to surprising results for those that are already using these configurations.

I agree with @kzantow , since we're keying off of the GO* configuration, we should try to replicate the same behavior. Even if we don't support the same functionality we should at least detect when we're configured in a way that we cannot support and output at least warnings to the log (if not error out entirely).

README.md
```diff
@@ -504,6 +504,15 @@ golang:
...
+   # SYFT_GOLANG_REMOTE_PROXY env var
+   remote-proxy: ""
```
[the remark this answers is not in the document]
deitch: I think your inclination for `goproxy` is correct; I took that for now. We always can adjust it.

README.md
deitch: This has me thinking that maybe we should just use the env vars? Meaning the setting in the config file (or env var `SYFT_GOLANG_GOPROXY` or `SYFT_GOLANG_PROXY`) really should be an _enabler_, rather than the actual value:

* `SYFT_GOLANG_GOPROXY` set to `""`? We do not get from the Internet
* `SYFT_GOLANG_GOPROXY` set to `true`? We get from the Internet, following the official rules for the various env vars as shown above.

Needless to say, all of the logic in the original `go` cmd is in `private/` [here](https://github.com/golang/go/blob/06264b740e3bfe619f5e90359d8f0d521bd47806/src/cmd/go/internal/modfetch/repo.go#L228) and similar spaces.

deitch: I added a separate commit that does this (rather than modifying the same commit, so it will be easier to roll back).

The part I left out is the `"direct"`, as that is a messy thing to implement. [This says](https://go.dev/ref/mod#vcs):

> The go command may download module source code and metadata directly from a version control repository. 
> Downloading a module from a [proxy](https://go.dev/ref/mod#communicating-with-proxies) is usually faster, but 
> connecting directly to a repository is necessary if a proxy is not available or if a module’s repository is not accessible 
> to a proxy (frequently true for private repositories). Git, Subversion, Mercurial, Bazaar, and Fossil are supported. A 
> version control tool must be installed in a directory in PATH in order for the go command to use it.

That is going to be some ugly mess to build in and support, whether natively or using external commands (which is what `go` uses). You cannot even rely on them being there, as `syft` very well could be running (likely is) outside of the build environment.

Suggestions?

kzantow:
@deitch in addition to local caching of downloaded modules, I've added git support and TUI feedback (when I ran this, it seemed like it may have been hung because there was no progress being made on the number of packages), It looks like this: 
image

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -1,29 +1,88 @@
...
+			SubStatus:          true,
+			RemoveOnCompletion: true,
+			Title:              "Downloading go mod",
+		},
+	}
+}
+
+func remoteProxies(proxy string, noProxy string, module string) (proxies []string) {
+	if proxy == "" {
+		proxy = os.Getenv("GOPROXY")
```
wagoodman, line 54: env variables should be evaluated as early as possible at runtime -- this also looks like it would be reevaluated several times during cataloging.

Could this get pushed all the way up to application config processing? Or somewhere above this level of processing ? 

(same comment for other env evaluations)

syft/pkg/cataloger/golang/billyfs_adapter.go
```diff
@@ -0,0 +1,84 @@
+package golang
+
+import (
+	"io/fs"
+	"os"
+
+	"gopkg.in/src-d/go-billy.v4"
```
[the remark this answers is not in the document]
kzantow: I had some weird incompatibilities when I tried to import github.com/go-git/go-git directly. I'll see if there's a way to get the most recent stuff working again.

syft/source/unindexed_directory_resolver.go
```diff
@@ -0,0 +1,565 @@
...
+	"github.com/spf13/afero"
+	"golang.org/x/exp/slices"
+
+	"github.com/anchore/syft/internal/log"
+)
+
+type WritableFileResolver interface {
+	FileResolver
+
+	Write(path string, reader io.Reader) error
```
wagoodman, line 24: resolvers tend to deal with `Locations` instead of string paths in the content fetching, which is the closest use case to this. Is this something that was considered during the design process? (it may not make sense given the use case)

syft/source/unindexed_directory_resolver.go
```diff
@@ -0,0 +1,565 @@
...
+	return out
+}
+
+func (u UnindexedDirectoryResolver) isSymlink(fi os.FileInfo) bool {
+	return fi.Mode().Type()&fs.ModeSymlink == fs.ModeSymlink
+}
+
+// ------------------------- fs.FS ------------------------------
+
+type fsFS struct {
```
wagoodman, line 371: the name of `fsFS` isn't really clear here as to what it is and what it's for

syft/source/unindexed_directory_resolver.go
```diff
@@ -0,0 +1,565 @@
...
+func (f fsFS) Stat(name string) (fs.FileInfo, error) {
+	fi, err := f.u.fs.Stat(f.u.absPath(name))
+	if err != nil {
+		return nil, err
+	}
+	return newFsFileInfo(f.u, name, fi.IsDir(), fi), nil
+}
+
+func (f fsFS) Open(name string) (fs.File, error) {
+	// parts := strings.Split(name, "/")
```
wagoodman, line 447: nit: dead code

syft/source/unindexed_directory_resolver.go
deitch: I would comment, but I expect this is more for @kzantow as I have no idea what that is 😁

syft/pkg/cataloger/golang/cataloger.go
```diff
@@ -5,28 +5,47 @@ package golang
...
-	return generic.NewCataloger("go-module-binary-cataloger").
-		WithParserByMimeTypes(c.parseGoBinary, internal.ExecutableMIMETypeSet.List()...)
+	return &ProgressingCataloger{
+		progress: c.licenses.progress,
+		cataloger: generic.NewCataloger("go-module-binary-cataloger").
+			WithParserByMimeTypes(c.parseGoBinary, internal.ExecutableMIMETypeSet.List()...),
+	}
+}
+
+type ProgressingCataloger struct {
```
wagoodman, line 39: this looks like this is trying to be a common abstraction, maybe belongs in `generic`?

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -138,3 +163,79 @@ func processCaps(s string) string {
...
+			fsys = os.DirFS(p)
+		}
+		if fsys != nil {
+			break
+		}
+	}
+	return
+}
+
+//nolint:gosec
```
wagoodman, line 189: only nolint the line that is failing, not the entire function (probably on `http.Get`)
```
	resp, err := http.Get(u)  //nolint:gosec
```

internal/config/application.go
```diff
@@ -72,10 +72,13 @@ func (cfg Application) ToCatalogerConfig() cataloger.Config {
 		},
 		Catalogers:  cfg.Catalogers,
 		Parallelism: cfg.Parallelism,
-		Golang: golangCataloger.GoCatalogerOpts{
-			SearchLocalModCacheLicenses: cfg.Golang.SearchLocalModCacheLicenses,
-			LocalModCacheDir:            cfg.Golang.LocalModCacheDir,
-		},
+		Golang: golangCataloger.NewGoCatalogerOpts(
```
wagoodman, line 75: we've been starting to lean into a struct builder pattern in places instead of package-wide functional options in a few places. They have the advantage of not polluting the package namespace with options for multiple structs. I think we should lean in favor of that for now:
```
golangVataloger.NewGoCatalogerConfig().
  WithSearchLocalModCacheLicenses(...).
  WithProxy(...).
  ...
```

syft/event/event.go
```diff
@@ -34,4 +34,7 @@ const (
 
 	// AttestationStarted is a partybus event that occurs when starting an SBOM attestation process
 	AttestationStarted partybus.EventType = "syft-attestation-started-event"
+
+	// GenericProgressStarted is a partybus event that occurs when a user wants to display arbitrary progress messages
+	GenericProgressStarted partybus.EventType = "syft-generic-event-started"
```
wagoodman, line 39: I don't think we're ready for a common generic progress event. I think we should keep this scoped for the cataloger for now, allowing us to know by the event type what the progress belongs under (always the cataloging task).

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -138,3 +163,79 @@ func processCaps(s string) string {
...
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("failed to get module zip: %s", resp.Status)
+		}
+	}
+	// read the zip
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	return zip.NewReader(bytes.NewReader(b), resp.ContentLength)
```
wagoodman, line 217: TIL zip.Reader implements `fs.FS`, very cool!

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -138,3 +163,79 @@ func processCaps(s string) string {
...
+func getModuleRepository(progress *event.GenericProgress, moduleName string, moduleVersion string) (fs.FS, error) {
+	repoName := moduleName
+	parts := strings.Split(moduleName, "/")
+	if len(parts) > 2 {
+		repoName = fmt.Sprintf("%s/%s/%s", parts[0], parts[1], parts[2])
+	}
+	progress.SetValue(fmt.Sprintf("git: %s", repoName))
+	f := memfs.New()
+	buf := &bytes.Buffer{}
+	_, err := git.Clone(memory.NewStorage(), f, &git.CloneOptions{
```
wagoodman, line 229: a fs.FS should be passed by the caller instead of `memory.NewStorage()` being used directly. (I've found these kinds of considerations are usually cheap to do during initial development but potentially harder to bolt on later)

kzantow: You mean a temp dir? The idea was for this to stay in memory until it's written out

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -68,38 +67,64 @@ func modCacheResolver(modCacheDir string) source.FileResolver {
...
 func (c *goLicenses) getLicenses(resolver source.FileResolver, moduleName, moduleVersion string) (licenses []string, err error) {
-	moduleName = processCaps(moduleName)
-
 	licenses, err = findLicenses(resolver,
-		fmt.Sprintf(`**/go/pkg/mod/%s@%s/*`, moduleName, moduleVersion),
+		fmt.Sprintf(`**/go/pkg/mod/%s@%s/*`, processCaps(moduleName), moduleVersion),
 	)
 
-	if c.searchLocalModCacheLicenses && err == nil && len(licenses) == 0 {
+	search := c.opts.searchLocalModCacheLicenses || c.opts.searchRemoteLicenses
```
wagoodman, line 82: this seems to control the local search, but we're keying off of the remote search option too, is this intentional?

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -68,38 +67,64 @@ func modCacheResolver(modCacheDir string) source.FileResolver {
...
+	dir := fmt.Sprintf("%s@%s", processCaps(moduleName), moduleVersion)
+	glob := fmt.Sprintf("%s/*", dir)
+
+	if search && err == nil && len(licenses) == 0 {
```
wagoodman, line 86: I think there are opportunities to make these code blocks a little more grokkable by separating some of the compound conditional, using more guard clauses, and decomposing functions into smaller pieces. For example:
```golang

func (c *goLicenses) getLicenses(resolver source.FileResolver, moduleName, moduleVersion string) (licenses []string, err error) {
	// look in the target resolver...
	licenses, err = findLicenses(resolver,
		fmt.Sprintf(`**/go/pkg/mod/%s@%s/*`, processCaps(moduleName), moduleVersion),
	)
	if err != nil || len(licenses) > 0 {
		return requireCollection(licenses), err
	}

	// look in the local host mod cache...
	licenses, err = c.getLicensesFromLocal(moduleName, moduleVersion)
	if err != nil || len(licenses) > 0 {
		return requireCollection(licenses), err
	}

	// fetch the module remotely and search...
	licenses, err = c.getLicensesFromRemote(moduleName, moduleVersion)
	if err != nil || len(licenses) > 0 {
		return requireCollection(licenses), err
	}

	return requireCollection(licenses), err
}

func (c *goLicenses) getLicensesFromRemote(moduleName, moduleVersion string) ([]string, error) {
	if !c.opts.searchRemoteLicenses {
		return nil, nil
	}

	if err := c.fetchModuleFromRemote(moduleName, moduleVersion); err != nil {
		log.Tracef("remote proxy walk failed for: %s@%s", moduleName, moduleVersion)
		return nil, err
	}

	return findLicenses(c.localModCacheResolver, moduleSearchGlob(moduleName, moduleVersion))

}

func (c *goLicenses) fetchModuleFromRemote(moduleName, moduleVersion string) error {
	proxies := remotesForModule(c.opts.proxies, c.opts.noProxy, moduleName)

	var fsys fs.FS
	fsys, err := getModule(c.progress, proxies, moduleName, moduleVersion)
	if err != nil {
		return err
	}

	// populate the mod cache with the results
	return fs.WalkDir(fsys, ".", func(filePath string, d fs.DirEntry, _ error) error {
		if d.IsDir() {
			return nil
		}
		f, err := fsys.Open(filePath)
		if err != nil {
			return err
		}
		return c.localModCacheResolver.Write(
			source.NewLocation(
				path.Join(
					moduleDir(moduleName, moduleVersion),
					filePath),
			),
			f)
	})

}

func (c *goLicenses) getLicensesFromLocal(moduleName, moduleVersion string) ([]string, error) {
	isEnabled := c.opts.searchLocalModCacheLicenses || c.opts.searchRemoteLicenses

	if !isEnabled {
		return nil, nil
	}

	// if we're running against a directory on the filesystem, it may not include the
	// user's homedir / GOPATH, so we defer to using the localModCacheResolver
	return findLicenses(c.localModCacheResolver, moduleSearchGlob(moduleName, moduleVersion))
}

func moduleDir(moduleName, moduleVersion string) string {
	return fmt.Sprintf("%s@%s", processCaps(moduleName), moduleVersion)
}

func moduleSearchGlob(moduleName, moduleVersion string) string {
	return fmt.Sprintf("%s/*", moduleDir(moduleName, moduleVersion))
}

func requireCollection(licenses []string) []string {
	if licenses == nil {
		return []string{}
	}
	return licenses
}

```

Consider this comment as optional (no action required), but it is encouraged.

syft/pkg/cataloger/golang/licenses.go
kzantow: this is intentional, because we're caching things in the same location

syft/pkg/cataloger/golang/licenses.go
wagoodman: sorry, retracting this, I'm looking at the wrong interface

wagoodman (approved):
approving in advance assuming the `NopResolver` is implemented. Nice work @deitch @kzantow !

deitch:
🥳

PR #1636, opened 2023-03-01
fix: handle individual cataloger panics

### LINKED ISSUE #1624: Handle cataloger panics
If any single cataloger panics, the entire Syft execution terminates. It would be nice to loudly log these as `ERROR` messages but allow Syft to continue cataloging and return a result.

**Why is this needed**:
Prevent Syft from crashing and give as much information as possible to users.

**Additional context**:
There was a recent ticket, https://github.com/anchore/syft/issues/1621, in which Syft panic'd and the entire run failed, where we could have cataloged everything except a single file that caused the panic, and reported that file. This would help Syft development as well, because it would be much easier for users to point developers to failed files.

kzantow:
This PR handles any panics that occur within individual catalogers, allowing cataloging to continue and an SBOM to be generated.

Fixes #1624

### DIFF
--- a/syft/pkg/cataloger/catalog.go
+++ b/syft/pkg/cataloger/catalog.go
@@ -108,7 +108,11 @@ func runCataloger(cataloger pkg.Cataloger, resolver source.FileResolver) (catalo
 func Catalog(resolver source.FileResolver, release *linux.Release, parallelism int, catalogers ...pkg.Cataloger) (*pkg.Catalog, []artifact.Relationship, error) {
 	catalog := pkg.NewCatalog()
 	var allRelationships []artifact.Relationship
+
 	filesProcessed, packagesDiscovered := newMonitor()
+	defer filesProcessed.SetCompleted()
+	defer packagesDiscovered.SetCompleted()
+
 	// perform analysis, accumulating errors for each failed analysis
 	var errs error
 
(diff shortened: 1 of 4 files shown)


syft/pkg/cataloger/catalog.go
```diff
@@ -168,14 +175,10 @@ func Catalog(resolver source.FileResolver, release *linux.Release, parallelism i
 
 	allRelationships = append(allRelationships, pkg.NewRelationships(catalog)...)
 
-	if errs != nil {
-		return nil, nil, errs
-	}
-
 	filesProcessed.SetCompleted()
 	packagesDiscovered.SetCompleted()
```
kzantow, line 179: The fact that this function was returning _before_ the `.SetCompleted()` calls below resulted in a hung TUI for any error that occurs.

wagoodman: oh duh! man, I sent you down the hard path too soon.

wagoodman: actually, can you add a little more defensiveness to this (protecting against future changes)? Can you add something to the affect of:
```golang
defer filesProcessed.SetCompleted()
defer packagesDiscovered.SetCompleted()
```

where these are created with `newMonitor()`?

kzantow: Good call -- done!

Issue #1634, opened 2023-03-01
Config parsing includes `config.yaml` in working dir


AidanDelaney:
**What happened**:
`syft` reads any `config.yaml` in the current working directory as Syft configuration.  If the `config.yaml` is not a valid `syft` config, then scanning fails.

**What you expected to happen**:
I expect `config.yaml` in the current working directory to not be read as Syft configuration.

**Steps to reproduce the issue**:
1. Create directory `foo`: `mkdir foo`
2. Create a `config.yaml` in the `foo` directory: `touch foo/config.yaml`
3. Place invalid Syft config in the `config.yaml`: 
    ```
    cat > foo/config.yaml << EOF
      output:
        example:
          value: value
    EOF
    ```
  Note that is is important that the config file uses a Syft configuration value like `output`
4. Run syft in the `foo` directory: `cd foo && syft packages .`
5. Observe the error `'output[0]' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[example:map[value:value]]'`

**Anything else we need to know?**:
The problem occurs on https://github.com/anchore/syft/blob/main/internal/config/application.go#L226.

On line 226 we add the `pwd` as a config path.  We then check to see if a `.syft` config file exists.  If it does not, we move to check the other documented config files.

On line 236 we add `.syft` as a config path.  Note that `pwd` is still in the search path for configs.  When we then search for config files with the name `config`, it picks up `./config.yaml` which was never intended as a `syft` configuration file.

**Environment**:
- Output of `syft version`: latest syft from `git`
- OS (e.g: `cat /etc/os-release` or similar): Falure observed on Linux and Macos

PR #1640, opened 2023-03-02
Add pwd to config search path iff it contains a config

AidanDelaney:
Only add the pwd to the config search path if and only if it contains a config file that we expect.  This avoids incorrectly finding config files that may be specific to applictions other than syft.

fixes: #1634

### DIFF
--- a/internal/config/application.go
+++ b/internal/config/application.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"os"
 	"path"
 	"reflect"
 	"sort"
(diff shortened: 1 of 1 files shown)


spiffcs:
👋 Thanks for the PR @AidanDelaney - do you mind if I make a quick update the keeps the current behavior, but adds a warning. 

We're looking to release syft v1.0.0 in the future and think this is a good candidate for that handover.

I don't want to merge this PR as is and break setups for people who are relying on discovering `config.yaml`.

Instead, if we introduce your changes, but still allow `v.AddConfigPath(".")` with a logged warning we can communicate to people this will be sunset in the `v1.0.0` release

I'll also make sure to add your issue to our v1 milestone so we don't miss removing the warning and sunsetting the wrong behavior

AidanDelaney:
Please feel free to make any update you'd like.  I *think* that adding `v.AddConfigPath(".")` will mean that this PR then does not fix #1634. Do we want to keep that issue open until `v1.0.0`?

kzantow (approved):

internal/config/application_test.go
```diff
@@ -0,0 +1,114 @@
+package config
+
+import (
+	"github.com/adrg/xdg"
```
kzantow, line 4: nit: gosimports

internal/config/application_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"os"
+	"path"
+	"testing"
+)
+
+// TODO: set negative case when config.yaml is no longer a valid option
+func TestApplicationConfig(t *testing.T) {
+	// config is picked up at desired configuration paths
+	// VALID: .syft.yaml, .syft/config.yaml, ~/.syft.yaml, <XDG_CONFIG_HOME>/syft/config.yaml
+	// DEPRECATED: .config.yaml is currently supported by
```
kzantow, line 16: nit: is it `.config.yaml` or `./config.yaml`?

internal/config/cataloger_options.go
```diff
@@ -16,6 +16,7 @@ type catalogerOptions struct {
 
 func (cfg catalogerOptions) loadDefaultValues(v *viper.Viper) {
 	v.SetDefault("package.cataloger.enabled", true)
+	v.SetDefault("package.cataloger.scope", "squashed")
```
wagoodman, line 19: isn't there a CLI default here? (in which case this wouldn't need to be set)

kzantow (approved):
👍

PR #1645, opened 2023-03-03
support for scanning license files in golang packages

deitch:
Partial replacement for #1630 

After discussions with @kzantow , it was decided to split #1630 into 2 PRs: one that scans local `$GOPATH/mod`, which is guaranteed to be safe, and another (after) that will take a CLI option to enable pulling down the golang package from the Internet via the GOPROXY, if it is not found in local GOPATH.

This is the first part: look for the go package in `GOPATH/mod`, and thus has no CLI changes, and does not reach out to the Internet.


deitch:
Note: I did not remove the remote package retrieval code, as it has been tested and works. Instead, I put in a `const disableRemotePackage`, which we can just remove when the future PR opens.

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -0,0 +1,112 @@
...
+
+func newGoLicenses(searchLocalGoModForLicenses bool) goLicenses {
+	return goLicenses{
+		searchLocalGoModForLicenses: searchLocalGoModForLicenses,
+		localGoModResolver:          deferredResolver,
+	}
+}
+
+// this needs to be shared between GoMod & GoBinary so it's only scanned once
+var deferredResolver = deferredResolverForLocalGoMod()
```
kzantow, line 28: FYI a nice bonus for using a standard resolver, is we get TUI output for free:

```
% SYFT_PACKAGE_CATALOGER_GOLANG_SEARCH_LOCAL_GO_MOD_LICENSES=true syft file:snapshot/darwin-build_darwin_amd64_v1/syft 
 ✔ Indexed snapshot/darwin-build_darwin_amd64_v1/syft 
 ⠹ Cataloging packages     [packages 0]
 ⠹ Indexing /Users/kzantow/go/pkg/mod [file: /Users/kzantow/go/pkg/mod/github.com/!azure/azure-...ces/network/mgmt/2018-06-01/network/virtualwans.go]
```

syft/pkg/cataloger/golang/package.go
```diff
@@ -6,18 +6,25 @@ import (
...
 	"github.com/anchore/syft/syft/source"
 )
 
-func newGoBinaryPackage(dep *debug.Module, mainModule, goVersion, architecture string, buildSettings map[string]string, locations ...source.Location) pkg.Package {
+func (c *goBinaryCataloger) newGoBinaryPackage(resolver source.FileResolver, dep *debug.Module, mainModule, goVersion, architecture string, buildSettings map[string]string, locations ...source.Location) pkg.Package {
 	if dep.Replace != nil {
 		dep = dep.Replace
 	}
 
+	licenses, err := c.licenses.getLicenses(resolver, dep.Path, dep.Version)
```
spiffcs, line 19: Do we want licenses to be nil here or, if there is an err, have it be an initialized empty array?

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		return source.NewFromDirectory(path.Join(goPath, "pkg", "mod"))
+	})
+}
+
+func (c *goLicenses) getLicenses(resolver source.FileResolver, moduleName, moduleVersion string) (licenses []string, err error) {
+	nameParts := strings.Split(moduleName, "/")
+
+	if len(nameParts) < 3 {
```
spiffcs, line 49: We might want to add a TODO comment here for the future if we find that `dep.Path` cases don't always conform to `host/org/repo` I think there might be some core packages that are just `domain/thing`, but I can't find them currently. Does anyone else have more concrete information on `dep.Path`variations?

spiffcs (commented):
Deferred resolver stuff for finding `resolver` outside of the source looks good to me. Thanks for updating this from the fs pattern to the internal resolver stuff. 

I just had some minor comments about licenses being nil vs empty and if the check for splitting on the go.mod ever runs into cases where we get 2 parts for a valid module name

deitch:
Thanks for pushing the rewrite onto it.

How does this work now, then? As far as I can tell, you pushed in a few key changes:

* enabling search for licenses outside of current tree using env var
* resolver to find those licenses
* nice test structure (I like what you did there)

Is that about correct? So would golang license searching happen only if the env var is set `SYFT_PACKAGE_CATALOGER_GOLANG_SEARCH_LOCAL_GO_MOD_LICENSES=true`?

And if so, would a similar env var enable network search, e.g. `SYFT_PACKAGE_CATALOGER_GOLANG_SEARCH_INTERNET_GO_MOD_LICENSES=true`?

Also, why is it `LOCAL_GO_MOD_LICENSES`? The licenses aren't in `go.mod`, they are in the actual packages? 

And does it work when scanning a go binary, rather than source? (in which case the `GO_MOD` in the env var is confusing)

internal/config/cataloger_options.go
```diff
@@ -11,11 +11,13 @@ import (
 type catalogerOptions struct {
 	Enabled  bool         `yaml:"enabled" json:"enabled" mapstructure:"enabled"`
 	Scope    string       `yaml:"scope" json:"scope" mapstructure:"scope"`
+	Golang   golang       `yaml:"golang" json:"golang" mapstructure:"golang"`
```
wagoodman, line 14: these are meant for options that apply to all catalogers, and also gives a pretty odd env var path to give: ``SYFT_PACKAGE_CATALOGER_GOLANG_...` .

I think there is another compromise here that elevates the importance of simplifying the user-facing values, probably trading off for short-term complexity in the config processing path.

syft/pkg/cataloger/golang/licenses.go
kzantow: Good call @spiffcs -- I think this could actually be simplified to just use the `moduleName` directly in the path, since we're using a resolver now

deitch:
Yup. All clear. I would request (as a user) that we document that capabilities somewhere? Whether CLI flag with `--help` explaining it, or docs in the repo (or both?). I guarantee I will hunt to find it in 2 months, and I am thinking the last thing maintainers need is people opening issues or slack messages asking how to do this. 😄

kzantow:
@deitch good call, I'll add it to the configuration section of the documentation. NOTE: you can always dump the application config by running `syft -vv`, and you'd see the golang section, like:
```
golang:
  search-local-mod-cache-licenses: false
```

The environment variables are always prefixed by `SYFT_` and then match the full path in the config file with underscores, so as noted above `golang.search-local-mod-cache-licenses` translates to `SYFT_GOLANG_SEARCH_LOCAL_MOD_CAHCE_LICENSES`

syft/pkg/cataloger/golang/package.go
kzantow: Good call -- this always returns a non-nil slice now

internal/config/golang.go
```diff
@@ -0,0 +1,11 @@
+package config
+
+import "github.com/spf13/viper"
+
+type golang struct {
+	SearchLocalModCacheLicenses bool `json:"search-local-mod-cache-licenses" yaml:"search-local-mod-cache-licenses" mapstructure:"search-local-mod-cache-licenses"`
```
wagoodman, line 6: "Local" could be ambiguous (local to where? the container image? the host?). What about "Host"?

```
search-host-mod-cache-licenses
```

(not trying to nit pick to death, but since it's on the user surface area this is easy to change now, hard to change later)

kzantow: If you're scanning a directory... is that still the "host" (not like host vs container is more obvious)?

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -0,0 +1,108 @@
...
+			contents, err := resolver.FileContentsByLocation(l)
+			if err != nil {
+				return nil, err
+			}
+			parsed, err := licenses.Parse(contents)
+			if err != nil {
+				return nil, err
+			}
+
+			if parsed != nil {
```
wagoodman, line 93: nit: nil check is not necessary

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -0,0 +1,108 @@
...
+func newDeferredModCacheResolver() source.FileResolver {
+	return source.NewDeferredResolverFromSource(func() (source.Source, error) {
+		goPath := os.Getenv("GOPATH")
+
+		if goPath == "" {
+			homeDir, err := homedir.Dir()
+			if err != nil {
+				log.Debug("unable to determine user home dir: %v", err)
+			}
+			goPath = path.Join(homeDir, "go")
```
wagoodman, line 41: it doesn't look like we should use the value of `homeDir` if error != nil... should we return an error instead (and remove the log statement)?

deitch: See above comment. I don't think we should try and recreate the `GOPATH`. Either it is set and we look there, or it is not, and we do not even try.

If we get requests later that _lots of people_ want us to determine this automatically, we always can add it later.

internal/config/golang.go
wagoodman: I guess it could be ambiguous any way you slice it:
- on host, dir scanning on the host: "host" makes sense
- on host, scanning an image: "host" makes sense
- in container, dir scanning in the container: "host" would be wrong in an absolute sense
- in container, scanning an image : "host" would be wrong in an absolute sense

I take back my comment -- "local" is good enough (still suffers from some ambiguity, but better than than being wrong in an absolute sense).

syft/pkg/cataloger/golang/licenses.go
wagoodman: actually, I think I like that better. If `goPath` is empty, then return don't return a resolver to search against at all.

kzantow: According [to the go docs](https://go.dev/doc/gopath_code), the default gopath is:

> The GOPATH environment variable specifies the location of your workspace. It defaults to a directory named go inside your home directory, so $HOME/go on Unix, $home/go on Plan 9, and %USERPROFILE%\go (usually C:\Users\YourName\go) on Windows.

Why wouldn't we default to this same location if GOPATH is not explicitly set?

wagoodman (approved):
@deitch and @kzantow nice work!

Future work for the config: as more parameters are added that equate to relatively the same thing we might want to think about "single switch" options that enable all options of that kind.

PR #1647, opened 2023-03-03
feat: disable cpe vendor wildcards to reduce false positives

westonsteimel:
Currently, syft is still generating `*` for the vendor for Ruby and javascript, but this seems unnecessary and is as source of many reported false positives in Grype.  I have added the URL -> vendor candidates parsing logic for both as well.  Both ecosystems are already well-covered by GitHub Security Advisories anyways, so CPE's are becoming increasingly irrelevant to both.  Also we've already disabled wildcard vendors for java and that was the ecosystem where we were most likely to be catching something unknown (though still also catching enormous numbers of false positives)

### DIFF
--- a/syft/pkg/cataloger/common/cpe/generate.go
+++ b/syft/pkg/cataloger/common/cpe/generate.go
@@ -93,13 +93,6 @@ func candidateVendors(p pkg.Package) []string {
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


kzantow (approved):
LGTM 👍

Issue #1646, opened 2023-03-03
Update haproxy binary matcher


witchcraze:
**What would you like to be added**:

Fix classifier which finds haproxy binary.

**Why is this needed**:

2.x series are not detected

**Additional context**:

```
$ syft --version
syft 0.74.0

$ syft -q haproxy:2.7 | grep haproxy
$

$ syft -q haproxy:2.6 | grep haproxy
$

$ syft -q haproxy:2.5 | grep haproxy
$

$ syft -q haproxy:2.4 | grep haproxy
$

$ syft -q haproxy:2.3 | grep haproxy
$

$ syft -q haproxy:2.2 | grep haproxy
$

$ syft -q haproxy:2.1 | grep haproxy
$

$ syft -q haproxy:2.0 | grep haproxy
$

$ syft -q haproxy:1.9 | grep haproxy
haproxy                 1.9.16                  binary

$ syft -q haproxy:1.8 | grep haproxy
haproxy                 1.8.31                   binary

$ syft -q haproxy:1.7 | grep haproxy
haproxy                 1.7.14                  binary

$ syft -q haproxy:1.6 | grep haproxy
haproxy                 1.6.15                 binary

$ syft -q haproxy:1.5 | grep haproxy
haproxy                 1.5.19                 binary
```

kzantow:
FYI this will be fixed when https://github.com/anchore/syft/pull/1648 is merged -- I overlooked [a detail that was needed in the evidenceMatchers function](https://github.com/anchore/syft/pull/1648/files#diff-3b4f786ed492b751281b58aa4ef14036cad70ba66e09ab625c8ac7c03d9e21eaR62-R68)

Issue #1643, opened 2023-03-02
Improve Python binary scanning


kzantow:
**What would you like to be added**:
Use shared libraries to determine python binary versions and remove `patchlevel.h` scanning.

**Why is this needed**:
Python binary files are not currently being detected when they reference shared libraries, because there is nothing in the binary bytes that contains the version.

**Additional context**:
Python binary scanning today is split up into 3 separate classifiers:
* python binary
* python binary library
* cpython source

Often the _python binary library_ classifier finds appropriate python binary information, but this does not have any reference to the actual python executable because the binary itself is just loading the shared library. Additionally, Syft sometimes surfaces the `patchlevel.h` files from the cpython source, without any location of the actual binary. We need to examine the shared libraries referenced in python binaries and look up the contents of the shared library to determine the version of python installed.

This is related to: https://github.com/anchore/syft/issues/661

witchcraze:
Maybe, that will also useful fot ruby

Docker official image ruby:latest
```
# which ruby
/usr/local/bin/ruby
# ruby -v
ruby 3.2.1 (2023-02-08 revision 31819e82c8) [x86_64-linux]

# strings /usr/local/bin/ruby | grep '3\.2'
libruby.so.3.2

# ls -al /usr/local/lib/libruby*
lrwxrwxrwx 1 root root       16 Mar  1 17:14 /usr/local/lib/libruby.so -> libruby.so.3.2.1
lrwxrwxrwx 1 root root       16 Mar  1 17:14 /usr/local/lib/libruby.so.3.2 -> libruby.so.3.2.1
-rwxr-xr-x 1 root root 18327896 Mar  1 17:13 /usr/local/lib/libruby.so.3.2.1

# strings /usr/local/lib/libruby.so.3.2 | grep '3\.2\.1'
/usr/local/include/ruby-3.2.0/x86_64-linux/rb_mjit_min_header-3.2.1.h
ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [x86_64-linux]
ruby 3.2.1 (2023-02-08 revision 31819e82c8) +MJIT [x86_64-linux]
ruby 3.2.1 (2023-02-08 revision 31819e82c8) [x86_64-linux]
3.2.1
```

related PR
https://github.com/anchore/syft/pull/1580

kzantow:
@witchcraze I think you're right:

```
# ldd /usr/local/bin/ruby 
[...]

PR #1648, opened 2023-03-03
fix: improved Python binary detection

kzantow:
Originally, when Python binary detection was added, it checked 3 sources:
* `python*` binaries directly
* `libpython*` shared libraries
* `patchlevel.h` source file

The reason for the latter 2 - especially the last one, is that some Python binaries do not actually contain version information but instead load a `libpython*` shared library, which contains this.

This has caused some confusion for users and did not necessarily prove to be the most accurate. This PR changes the behavior such that `patchlevel.h` is never looked for, as there is no guarantee this would be applicable to the particular python binary file found. Additionally, when inspecting Python binaries, the behavior is now:
1. look for version information in the file and use this if found
1. look for the corresponding `libpython*` shared library in the binary dynamic library list, and extract version information from that file

With this change, Syft _does_ still look for `libpython*` files independently, and these may or may not show up as the primary location for a package.

The result is a few different cases:

* A Python binary is found, with version information _and_ a libpython shared library is found. If the versions are the same, both locations will be merged to the same package. If the versions are different, there will be 2 distinct packages with the corresponding location information.

* A Python binary is found, without version information _and_ a libpython shared library is found. If the libpython library is referenced from the Python binary, the primary location is the Python binary, and secondary location information includes the libpython library.

* No Python binary is found but a libpython shared library is found. This will continue to surface Python findings, with the primary location of the libpython library.

The reason for the last described case is that there are situations a container may have libpython without a binary -- for example an Apache HTTP Server image with a Python module. In this case, it is important to surface the fact that there may be Python execution with library and version found.

Fixes: #1643
Also fixes: #1646


syft/pkg/cataloger/binary/classifier.go
```diff
@@ -91,26 +96,83 @@ func fileNameTemplateVersionMatcher(fileNamePattern string, contentTemplate stri
...
 		}
 
 		matchMetadata := internal.MatchNamedCaptureGroups(pat, string(contents))
-		return singlePackage(classifier, reader, matchMetadata), nil
+		return singlePackage(classifier, location, matchMetadata), nil
+	}
+}
+
+func isExecutable(mode os.FileMode) bool {
+	return mode&0111 != 0
```
wagoodman, line 123: when catalogers look for indications of binaries we've leaned away from using the mode as a way to determine if the file is what we're looking for. Instead MIME type is more preferable of a check here (or no check and let the sharedLibMatcher surface nothing).

syft/pkg/cataloger/binary/classifier.go
```diff
@@ -195,3 +262,52 @@ func singleCPE(cpeString string) []cpe.CPE {
 		cpe.Must(cpeString),
 	}
 }
+
+// sharedLibraries returns a list of all shared libraries found within a binary, currently
+// supporting: elf, macho, and windows pe
+func sharedLibraries(resolver source.FileResolver, location source.Location) ([]string, error) {
```
wagoodman, line 268: I can't think of a better way to write this in terms of "detect the binary format then use the full parser" other than peaking at the magic number, which I think would be overkill here anyway.

Though, since there are a few formats and they share the same `ImportedLibraries ` there could be an ad-hoc interface and a list of constructors of the interface to make this more of a list of the same operation being done instead of it being unraveled, but I don't think there'd be any huge benefit and would probably be the same LoC anyway. 

(this was more thinking "outloud", consider this whole comment non-blocking)

wagoodman (approved):
My only blocking comment is the note about `mode` as an indicator for parsing, other than that 💯 !

syft/pkg/cataloger/binary/classifier.go
kzantow: Right. Unfortunately `internal.IsExecutable` seems not to work, I've just removed this whole check, it was only added because there were a lot of results returned for `**/python*`, but I don't think it will be a huge performance issue, at least it isn't in my testing.

syft/pkg/cataloger/binary/classifier.go
```diff
@@ -195,3 +262,52 @@ func singleCPE(cpeString string) []cpe.CPE {
...
+	contents, err := getContents(resolver, location)
+	if err != nil {
+		return nil, err
+	}
+
+	r := bytes.NewReader(contents)
+
+	e, err := elf.NewFile(r)
+	if err != nil {
+		log.Debug(err)
```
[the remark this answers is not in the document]
kzantow: Agreed! Updated.

syft/pkg/cataloger/binary/classifier.go
kzantow: Agree, I don't think there would be a huge benefit to changing this at the moment, at least.

noqcks:
I think we lost some information on python binaries after this PR: 

syft v0.740
```
➜  ~ syft -q python:3.4 | grep binary
python                        2.7.13                          binary
python                        3.4.10                          binary
python                        3.5.3                           binary
```

syft v0.74.1
```
```

kzantow:
@noqcks could you give the json output for those packages between the versions?

noqcks:
- [v0.74.1 - python:3.4](https://gist.github.com/noqcks/ee2f1811cdc119c35f6e4746daa95454)
- [v0.74.0 - python:3.4](https://gist.github.com/noqcks/a211f8cd91c9ec2b8440769d5d306295)

Issue #1606, opened 2023-02-23
argocd binary classifier


y12studio:
**What would you like to be added**:

binary classifiers for argocd, helm and kustomize.

**Why is this needed**:

argocd is widely used and is usually not installed with a package manager inside docker images.
> pullCount : 100M+
    from https://hub.docker.com/r/argoproj/argocd

**Additional context**:
current ouput from syft 0.72.0

```sh
$ syft -q packages argoproj/argocd:v2.6.2 | grep -E 'argocd|helm|kustomize'
helm.sh/helm/v3                                                              v0.0.0-20221214143859-835b7334cfe2                 go-module  
sigs.k8s.io/kustomize/api                                                    v0.11.4                                            go-module  
sigs.k8s.io/kustomize/api                                                    v0.12.1                                            go-module  
sigs.k8s.io/kustomize/cmd/config                                             v0.10.9                                            go-module  
sigs.k8s.io/kustomize/kustomize/v4                                           v0.0.0-20220802162801-56d82a8378df                 go-module  
sigs.k8s.io/kustomize/kyaml                                                  v0.13.6                                            go-module  
sigs.k8s.io/kustomize/kyaml                                                  v0.13.9                                            go-module  

$ syft -q packages argoproj/argocd:v2.6.2 -o json | grep -E '2.3:a:[.]?argo|2.3:a:helm|2.3:a:kustomize'
    "cpe:2.3:a:argoproj:argo-cd\\/v2:v0.0.0-20230216145252-6e02f8b23201:*:*:*:*:*:*:*",
    "cpe:2.3:a:argoproj:argo_cd\\/v2:v0.0.0-20230216145252-6e02f8b23201:*:*:*:*:*:*:*"
    "cpe:2.3:a:argoproj:gitops-engine:v0.7.1-0.20221208230615-917f5a0f16d5:*:*:*:*:*:*:*",
    "cpe:2.3:a:argoproj:gitops_engine:v0.7.1-0.20221208230615-917f5a0f16d5:*:*:*:*:*:*:*"
    "cpe:2.3:a:argoproj:notifications-engine:v0.3.1-0.20221203221941-490d98afd1d6:*:*:*:*:*:*:*",
    "cpe:2.3:a:argoproj:notifications_engine:v0.3.1-0.20221203221941-490d98afd1d6:*:*:*:*:*:*:*"
[...]

PR #1663, opened 2023-03-10
feat: add argocd, helm, kustomize and kubectl binary classifiers

y12studio:
Add 4 binary classifiers  for https://github.com/anchore/syft/issues/1606


spiffcs:
Approved the workflows --- I'll take a look at this PR this afternoon for review. Thanks for the contribution!

syft/pkg/cataloger/binary/test-fixtures/classifiers/positive/argocd-2.5.11/README.md
```diff
@@ -0,0 +1,39 @@
+test-fixtures for https://github.com/anchore/syft/issues/1606
```
spiffcs, line 1: Thanks for including a README.md about the test fixtures. 

My suggestion here is to point you towards a recent PR that was merged regarding the Ruby classifier:
https://github.com/anchore/syft/pull/1665/files

Rather than have the README that can be built, can we incorporate the test fixtures in the same pattern used in that PR as seen in the `Makefile`? This would hook the testing of these specific binaries into the same pattern that we're using for python and ruby

syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -46,6 +46,42 @@ var defaultClassifiers = []classifier{
...
+			`(?m)common\.version=(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`),
+		Package: "argocd",
+		PURL:    mustPURL("pkg:golang/argocd@version"),
+		CPEs:    singleCPE("cpe:2.3:a:argoproj:argocd:*:*:*:*:*:*:*"),
+	},
+	{
+		Class:    "helm",
+		FileGlob: "**/helm",
+		EvidenceMatcher: fileContentsVersionMatcher(
+			`(?m)\x00v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)\x00`),
```
spiffcs, line 62: I'm going to just flag this for further testing on my part. 

It looks like this is doing `[NUL]v<version-string>[NUL]` which could be too broad. I'll `strings` the `helm` and `kubectl` binaries to see the different edge cases here. If you have any comments on why this is needed please feel free to list on this thread!

spiffcs (commented):
Comments Submitted! Thanks again for the PR. I think we're happy to merge this after some clean up of the PURL and testing strategy. I will note for future readers that we might be looking at a future improvement in the next week or so on how the go cataloger parses build flags. This could cause duplicates that would need to be removed if we're able to detect versions for the binaries listed here.

The current pattern is `package cataloger` takes precedence over `binary cataloger` so we should keep this PR in mind when those possible improvements land.

syft/pkg/cataloger/binary/cataloger_test.go
```diff
@@ -239,6 +239,114 @@ func Test_Cataloger_DefaultClassifiers_PositiveCases(t *testing.T) {
...
 			},
 		},
+		{
+			name:       "positive-argocd-2.5.11",
+			fixtureDir: "test-fixtures/classifiers/positive/argocd-2.5.11",
+			expected: pkg.Package{
+				Name:      "argocd",
+				Version:   "2.5.11",
+				Type:      "binary",
+				PURL:      "pkg:golang/argocd@2.5.11",
```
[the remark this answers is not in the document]
y12studio: agree, using the full go mod path is right way.

syft/pkg/cataloger/binary/test-fixtures/classifiers/positive/argocd-2.5.11/README.md
y12studio: Thanks for your recommendation. Shifting the `README.md` which deals with test fixtures into the `Makefile` is a superior choice.

spiffcs:
Approved the new workflows -- I'll check back but the updates took good

syft/pkg/cataloger/binary/default_classifiers.go
spiffcs: Resolved This testing was good and NULL check here is valid

spiffcs (approved):

PR #1667, opened 2023-03-13
fix: more python matching support

kzantow:
@noqcks pointed out a few more python matching cases in [the comment here](https://github.com/anchore/syft/pull/1648#issuecomment-1464954661). This PR corrects an issue where `fileNameTemplateVersionMatcher` was causing versions with `.` to be treated as a regex dot, which matched any character. Additionally, it adds support for libpython named with a letter following the version number, e.g. `libpython3.4m.so.1.0`.


syft/pkg/cataloger/binary/cataloger_test.go
```diff
@@ -292,8 +292,8 @@ func Test_Cataloger_DefaultClassifiers_PositiveCases(t *testing.T) {
 			fixtureDir: "test-fixtures/classifiers/positive/python-binary-lib-3.7",
 			expected: pkg.Package{
 				Name:      "python",
-				Version:   "3.7.4a-vZ9",
-				PURL:      "pkg:generic/python@3.7.4a-vZ9",
+				Version:   "3.7.4",
```
spiffcs, line 295: Was this something we wanted to no longer test as it's an invalid version number or was this part of the string causing the regular expression to fail to capture the version?

syft/pkg/cataloger/binary/cataloger_test.go
```diff
@@ -576,9 +604,9 @@ func Test_Cataloger_DefaultClassifiers_PositiveCases(t *testing.T) {
 			packages, _, err := c.Catalog(resolver)
 			require.NoError(t, err)
 
-			for _, p := range packages {
```
spiffcs, line 579: This was removed because we're only testing single packages correct?

syft/pkg/cataloger/binary/classifier.go
```diff
@@ -79,6 +80,11 @@ func fileNameTemplateVersionMatcher(fileNamePattern string, contentTemplate stri
 
 		filepathNamedGroupValues := internal.MatchNamedCaptureGroups(pat, location.RealPath)
 
+		// versions like 3.5 should not match any character, but explicit dot
```
spiffcs, line 83: Would it be possible to add an example to this comment? I'm a little unclear what case this should be covering.

Is it "3.5 should not be captured and instead the things that should be matched are 3\\.5\\.0"?

syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -12,7 +12,7 @@ var defaultClassifiers = []classifier{
 		EvidenceMatcher: evidenceMatchers(
 			// try to find version information from libpython shared libraries
 			sharedLibraryLookup(
-				`^libpython[0-9]+(?:\.[0-9]+)+\.so.*$`,
+				`^libpython[0-9]+(?:\.[0-9]+)+[a-z]?\.so.*$`,
```
spiffcs, line 15: I see we're adding `a-z` in this capture group, but the test case was removed for `3.7.4a-vZ9` <-- Is this a case we want to capture for the lib python tests?

spiffcs (commented):
Looks like there is a quality gate failure:
```
Summary: 
   Baseline Packages: 175
   New Packages:      176
   Baseline Packages Matched: 88.00 % (154/175 packages)
   Baseline Metadata Matched: 0.57 % (1/175 metadata)
   Quality Gate: FAILED (requires exact name & version match)
   ```

And same thing appears for AC Tests:

```
Summary: 
   Baseline Packages: 175
   New Packages:      176
   Baseline Packages Matched: 88.00 % (154/175 packages)
   Baseline Metadata Matched: 0.57 % (1/175 metadata)
   Quality Gate: FAILED (requires exact name & version match)
   ```

Also added some comments to help me understand the PR better. Sorry for the confusion on reviewing this one, just wanted to make sure I'm caught up and fully understand the updates

syft/pkg/cataloger/binary/cataloger_test.go
kzantow: As far as I know, this was an invalid version number, yes. The test was actually silently failing because it wasn't returning any packages.

syft/pkg/cataloger/binary/classifier.go
kzantow: I'm happy to try to make this comment more clear somehow but yes, it's saying when the version is passed to the regex template, `3.5` should be treated as `3\.5`

spiffcs (approved):
Approved!

PR #1675, opened 2023-03-16
fix: reduce logging for bad dpkg lines

kzantow:
When scanning DPKG with invalid lines, we should DEBUG log these vs. WARN.


syft/pkg/cataloger/deb/parse_dpkg_db.go
```diff
@@ -151,7 +151,7 @@ func extractAllFields(reader *bufio.Reader) (map[string]interface{}, error) {
 			var val interface{}
 			key, val, err = handleNewKeyValue(line)
 			if err != nil {
-				log.Warnf("parsing dpkg status: extracting key-value from line: %s err: %v", line, err)
+				log.Debugf("parsing dpkg status: extracting key-value from line: %s err: %v", line, err)
```
wagoodman, line 154: this should really be trace (opinion): reason being that this may fail even in nominal circumstances.

spiffcs (approved):

PR #1684, opened 2023-03-20
Remove more side effects from application config testing

wagoodman:
Follow up to #1640 with a few adjustments:
- remove the symlink during testing since this is writing to paths in the repo (which is not ideal)
- uses the t.Cleanup facility for each test instead of `defer` calls
- removes a side effect against the xdg package based on XDG_CONFIG_PATH env var that persists after the test in the xdg package
- the homedir package has its own cache which needs to be disabled during testing (since the definition of HOME changes)
- for users that have a config in the home directory we change `HOME` to always point to a different (bogus) location so that never enters the equation.
- nit: changes `t.Fatal` to `require.NoError` calls instead

### DIFF
--- a/internal/config/application_test.go
+++ b/internal/config/application_test.go
@@ -6,12 +6,21 @@ import (
 	"testing"
 
 	"github.com/adrg/xdg"
+	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // TODO: set negative case when config.yaml is no longer a valid option
 func TestApplicationConfig(t *testing.T) {
+	// disable homedir package cache for testing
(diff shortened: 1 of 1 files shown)


internal/config/application_test.go
```diff
@@ -96,25 +86,31 @@ func TestApplicationConfig(t *testing.T) {
 			assertions: func(t *testing.T, app *Application) {
 				assert.Equal(t, "test-home-XDG-config", app.File)
 			},
-			Cleanup: func(t *testing.T) {},
+			cleanup: func() {
+				require.NoError(t, os.Unsetenv("XDG_CONFIG_DIRS"))
```
kzantow, line 90: This might actually need to track the original value to accurately reverse the behavior

wagoodman: good call

spiffcs (approved):

PR #1688, opened 2023-03-22
Fix shell completion by adding missing usage message required by spf13/cobra

DanHam:
Attempts a fix for #962. 

All credit to @cschug - see [this comment](https://github.com/anchore/syft/issues/962#issuecomment-1477771492)

As pointed out in that comment, the spf13/cobra library writes out the shell completion script using a `fmt.Sprintf` [here](https://github.com/spf13/cobra/blob/v1.6.1/bash_completionsV2.go#L37-L365). 

The `%[1]s` in that string seem to be replaced with the first command in the `Use:` field of the `cobra.Command` struct. Currently this field is ***not*** set so the `%[1]s` expand to an empty string resulting in a broken shell completion script. 

Adding the field (following the same method used by grype - [here](https://github.com/anchore/grype/blob/main/cmd/root.go#L63)) gives the `spf13/cobra` library the info it needs to correctly generate the shell completion script.

Testing locally seems to show everything is working OK. The shell completion now works as expected after running `source <(syft completion bash)`

Fixes: #962

### DIFF
--- a/cmd/syft/cli/commands.go
+++ b/cmd/syft/cli/commands.go
@@ -55,6 +55,7 @@ func New() (*cobra.Command, error) {
 
 	// rootCmd is currently an alias for the packages command
 	rootCmd := &cobra.Command{
+		Use:           fmt.Sprintf("%s [SOURCE]", internal.ApplicationName),
 		Short:         packagesCmd.Short,
 		Long:          packagesCmd.Long,
 		Args:          packagesCmd.Args,


kzantow:
I'm having a bit of a hard time getting this working. If I try this on my mac with `zsh`, it simply seems to do nothing. If I try this in a `bash` container (`golang:latest`), when I type `syft p<tab>`, I get this:
```
bash: _get_comp_words_by_ref: command not found
```
Is there a particular container I could use to see this work properly?

Also: you will need to [sign-off your commits](https://github.com/anchore/syft/blob/main/CONTRIBUTING.md#sign-off-your-work).

cmd/syft/cli/commands.go
```diff
@@ -55,6 +55,7 @@ func New() (*cobra.Command, error) {
 
 	// rootCmd is currently an alias for the packages command
 	rootCmd := &cobra.Command{
+		Use:           fmt.Sprintf("%s package [SOURCE]", internal.ApplicationName),
```
kzantow, line 58: `package` is not a valid command, and this is the root command, so this should probably be:
```go
fmt.Sprintf("%s [SOURCE]", internal.ApplicationName)
```

kzantow: Even so this is the _root_ command, not the packages command (even though they are duplicates), so it should display usage text as the root command.

DanHam:
Hi Keith

> I'm having a bit of a hard time getting this working. If I try this on my mac with zsh...

The original issue (#962) was with shell completion for bash. I don't use zsh so I've never tried the completions that are generated for that particular shell...

> Is there a particular container I could use to see this work properly?

I can see this working using the `golang:latest` image. Note: my arch is amd64:

1. Start: `docker run -it -v ${PWD}:/usr/src/syft -w /usr/src/syft --name syft-build golang:latest /bin/bash`
2. Fix dubious ownership warning: `git config --global --add safe.directory /usr/src/syft`
3. Install build deps: `make bootstrap`
4. Build: `make build`
5. Install syft for my os/arch: `install snapshot/linux-build_linux_amd64_v1/syft /usr/local/bin`
6. Verify: `syft --version # Outputs: syft 0.0.0-SNAPSHOT-18656c1`
7. The `golang:latest` image does not have the bash-completion package installed so lets add that: `apt update && apt install bash-completion`
8. Source the main bash-completion script: `source /usr/share/bash-completion/bash_completion`
9. Typing `ls --<tab><tab>` should now show completions e.g. `--all` etc
10. Source the syft completions script: `source <(syft completion bash)`
11. Typing `syft <tab><tab>` should now show completions

cmd/syft/cli/commands.go
DanHam: Using `fmt.Sprintf("%s packages [SOURCE]",...` causes the following to be output for the 'Usage'

```
$ syft
Generate a packaged-based Software Bill Of Materials (SBOM) from container images and filesystems

Usage:
  syft packages [SOURCE] [flags]
  syft [command]

Examples:
  syft packages alpin...
```

Happy to change that to what every you prefer. The important bit (for our fix) is that we have `syft` at the start of the command. That is what the `spf13/cobra` library needs to generate a working shell completion script (for bash at least!!)

kzantow: Right, but since a user can run syft without the `packages` command, I think we should leave this as `"%s [SOURCE]"`

DanHam:
> Also: you will need to [sign-off your commits](https://github.com/anchore/syft/blob/main/CONTRIBUTING.md#sign-off-your-work).

Shall I fix that typo (package -> packages) and force push with a signed commit?

kzantow:
Ahh, this was the magic sauce:
```
apt update && apt install bash-completion
source /usr/share/bash-completion/bash_completion
install snapshot/linux-build_linux_amd64_v1/syft /usr/local/bin
```

I think we can move forward after the aforementioned changes are done, thanks @DanHam

cmd/syft/cli/commands.go
DanHam: Hmmm... So the `[SOURCE]` bit of that usage command is only really relevant when running `syft packages ...`

```
$ syft packages
Generate a packaged-based Software Bill Of Materials (SBOM) from container images and filesystems

Usage:
  syft packages [SOURCE] [flags]
```

Perhaps we should just have `fmt.Sprintf("%s [command] [flags]", ...` as the usage for the main syft command?

This would just give:

```
$ syft
Generate a packaged-based Software Bill Of Materials (SBOM) from container images and filesystems

Usage:
  syft [command] [flags]
  syft [command]

Examples:
  syft packages alpine:latest                                a summary of discovered packages
  syft packages alpine:la...
```

...which seems to make a bit more sense. Are you OK with that?

kzantow: There's a bit of a mismatch between the help text and the command itself (this should be fixed, but I would not ask you to change this as part of this PR).

You can run Syft like:

```
syft alpine:latest
```

_without_ the `packages` command keyword, because the root command duplicates the functionality of the packages command.

For this reason, I would prefer that the root help text indicates how to run the root command, rather than indicating `packages` or some sub-command is necessary. It's a very small detail.

DanHam:
Thanks Keith,

I've now pushed up a signed commit with changes. 

I've set the usage message to `fmt.Sprintf('%s [command] [flags]",...)` as this seems to make the most sense for the main syft command (see my comment above - https://github.com/anchore/syft/pull/1688#discussion_r1145140433)

Please let me know if you require any additional changes.

kzantow:
@DanHam I hope this explanation why I still feel this should be `"%s [SOURCE]"`: https://github.com/anchore/syft/pull/1688#discussion_r1145151597

cmd/syft/cli/commands.go
DanHam: Ahh... OK. That makes sense now. The penny finally dropped!! :smile:

kzantow (approved):
👍 thanks for sticking through this, @DanHam !

DanHam:
> I hope this explanation makes sense...

It does! Thank you for your patience!!!

Hopefully, that last push should be OK.

kzantow:
FYI -- I tested this out with bash based on [the instructions](https://github.com/anchore/syft/pull/1688#issuecomment-1479783611) @DanHam provided:

```
$ syft <tab>
attest      (Generate an SBOM as an attestation for the given [SOURCE] container image)
completion  (Generate the autocompletion script for the specified shell)
convert     (Convert between SBOM formats)
help        (Help about any command)
login       (Log in to a registry)
packages    (Generate a package SBOM)
version     (show the version)
```

```
$ syft completion <tab>
bash        (Generate the autocompletion script for bash)
fish        (Generate the autocompletion script for fish)
powershell  (Generate the autocompletion script for powershell)
zsh         (Generate the autocompletion script for zsh)
```

PR #1691, opened 2023-03-22
1577 spdxlicense generate

spiffcs:
## Summary

Update the `license_list.go` to have more permissible inputs for greater SPDXID matching.

The `spdxlicense` package contains the `ID` method which interacts with the generated file `license_list.go`

The current implementation contains `-` in the keys. This PR removes these `-` and sanitizes the inputs so that we can match on a wider range of inputs found in the wild.

### Example:
`GPL3` `gpl3` `gpl-3` and `GPL-3` can all map to `GPL-3.0-only`

By moving all strings to lower and removing the - we're able to return valid SPDX license ID for a greater diversity of input strings.


internal/spdxlicense/license_list.go
```diff
@@ -1,1130 +1,1148 @@
...
+	"apache1.1":                             "Apache-1.1",
+	"apache1.1.0":                           "Apache-1.1",
+	"apache2":                               "Apache-2.0",
+	"apache2.0":                             "Apache-2.0",
+	"apache2.0.0":                           "Apache-2.0",
+	"apafml":                                "APAFML",
+	"apl1":                                  "APL-1.0",
+	"apl1.0":                                "APL-1.0",
+	"apl1.0.0":                              "APL-1.0",
+	"apps2.0.0p":                            "App-s2p",
```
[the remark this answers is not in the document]
spiffcs: There seem to be a couple of these cases which has caused the list to grow by 18 lines --- potentially 6 cases of a string that are now expanding version permutations which were not doing it before this PR

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -35,7 +33,7 @@ var licenseIDs = map[string]string{
 }
 `))
 
-var versionMatch = regexp.MustCompile(`-([0-9]+)\.?([0-9]+)?\.?([0-9]+)?\.?`)
+var versionMatch = regexp.MustCompile(`([0-9]+)\.?([0-9]+)?\.?([0-9]+)?\.?`)
```
spiffcs, line 36: Causing us to match on numbers in licenses that would otherwise not be considered versions, but I think the trade off of being able to match on more strings out in the wild is a good one here.

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -103,58 +100,59 @@ func run() error {
...
-		licenseIDs[cleanID] = l.ID
-	}
-
-	// The order of variations/permutations of a license ID matters because of we how shuffle its digits,
-	// that is because the permutation code can generate the same value for two difference licenses,
-	// for example: The licenses `ABC-1.0` and `ABC-1.1` can both map to `ABC-1`,
-	// so we need to guarantee the order they are created to avoid mapping them wrongly. So we use a sorted list.
-	// To overwrite deprecated licenses during the first pass we would later on rely on map order,
-	// [which in go is not consistent by design](https://stackoverflow.com/a/55925880).
+	// The order of variations/permutations of a license ID matter.
```
spiffcs, line 103: Simplified this logic down to a single pass

spiffcs: I commented out the log messages to reduce the noise since it's just a script we run to generate our license list and not part of the larger syft program.

You can turn those on to see how the sort prevents things like `alpm1` from mapping to later versions

internal/spdxlicense/generate/license.go
```diff
@@ -20,71 +20,73 @@ type License struct {
...
+
+func (l License) canReplace(deprecated License) bool {
+	// don't replace a license with a deprecated license
 	if l.Deprecated {
 		return false
 	}
 
 	// We want to replace deprecated licenses with non-deprecated counterparts
 	// For more information, see: https://github.com/spdx/license-list-XML/issues/1676
-	if other.Deprecated {
```
spiffcs, line 30: `findReplacementLicense` already assumes a deprecated input

internal/spdxlicense/generate/license.go
```diff
@@ -20,71 +20,73 @@ type License struct {
...
-		if sa != other.SeeAlso[i] {
+		if sa != deprecated.SeeAlso[i] {
 			return false
 		}
 	}
 
-	return l.ID == other.ID
-}
-
-func (ll LicenseList) findReplacementLicense(deprecated License) *License {
```
spiffcs, line 68: moved above the function `canReplace` for readability

internal/spdxlicense/license.go
```diff
@@ -18,16 +18,20 @@ const (
...
-		return value, "", exists
+	if value, exists := licenseIDs[cleanLicenseID(id)]; exists {
+		return value, exists
 	}
 	// we did not find, so treat it as a separate license
-	return "", id, true
+	return "", false
+}
+
+func cleanLicenseID(id string) string {
```
spiffcs, line 33: Had to duplicate this between generate and the actual package. If there is a better way to share the code happy to update to that!

internal/spdxlicense/license_test.go
```diff
@@ -6,94 +6,80 @@ import (
...
 			true,
 		},
 		{
 			"CC-by-nc-3-de",
 			"CC-BY-NC-3.0-DE",
-			"",
 			true,
 		},
 		// the below few cases are NOT expected, however, seem unavoidable given the current approach
-		{
```
spiffcs, line 53: No longer returning true 🥳

spiffcs:
Things to look at in the AM:

- [ ] `bzip2-1` <-- This test case is odd when removing the `-`. We might have to make the capture groups a little better to handle this case
- [ ] Cyclonedx License regression for `made-up`
- [ ] CycloneDX test updates with name carried through
- [ ] Encoder updates with new license updates as a consequence of `ID` change
- [ ] CLI regressions

internal/spdxlicense/license_test.go
wagoodman: :chef-kiss:

PR #1694, opened 2023-03-24
Add Linux Kernel cataloger

deitch:
This adds a cataloger for kernel files. It looks for files in a particular set of filenames, then uses libmagic (well, a library in go that reproduces part of it, with more to come) to parse metadata.

Includes options to append additional file globs to parse.

TODO:

* [x] Way too many CPEs, not sure how to fix those, as I do not know where they are generated
* [x] kernel modules, which are different file names and require different handling
* [ ] the extended version vs the basic version, when and how to use which
* [x] listing files as part of the package vs just having the package vs just having files (answer: files opened during cataloging are in pkg.Package.Location, otherwise owned files are listed in the pkg.Package.Metadata)
* [x] handling duplicates, e.g. we find the kernel, but the kernel also is part of an install package (answer: there is a file-ownership-overlap relationship created already by syft in these cases)


deitch:
Some sample output from a directory with a kernel file:

```json
                                                                                                                                                                {
 "spdxVersion": "SPDX-2.3",
 "dataLicense": "CC0-1.0",
 "SPDXID": "SPDXRef-DOCUMENT",
 "name": "/home/ubuntu/eve/rootfs/boot/",
 "documentNamespace": "https://anchore.com/syft/dir/home/ubuntu/eve/rootfs/boot-b6c3a42e-a58e-4d40-baf6-0629dfea8599",                                           "creationInfo": {
  "licenseListVersion": "3.20",
  "creators": [
   "Organization: Anchore, Inc",
   "Tool: syft-[not provided]"
  ],                                                                                                                                                              "created": "2023-03-24T07:50:58Z"
 },
 "packages": [
  {
   "name": "linux-kernel",
   "SPDXID": "SPDXRef-Package-linux-kernel-linux-kernel-cbf2ab66dde87d2d",
   "versionInfo": "5.10.121-linuxkit (root@buildkitsandbox) #1 SMP Fri Dec 2 10:35:42 UTC 2022",
   "sourceInfo": "acquired package info from the following paths: ",
   "licenseConcluded": "NONE",
   "licenseDeclared": "NONE",
   "externalRefs": [
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:linux-kernel:linux-kernel:5.10.121-linuxkit_\\(root\\@buildkitsandbox\\)_\\#1_SMP_Fri_Dec_2_10\\:35\\:42_UTC_2022:*:*:*:*:*:
*:*"
    },
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:linux-kernel:linux_kernel:5.10.121-linuxkit_\\(root\\@buildkitsandbox\\)_\\#1_SMP_Fri_Dec_2_10\\:35\\:42_UTC_2022:*:*:*:*:*:
*:*"
    },
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:linux_kernel:linux-kernel:5.10.121-linuxkit_\\(root\\@buildkitsandbox\\)_\\#1_SMP_Fri_Dec_2_10\\:35\\:42_UTC_2022:*:*:*:*:*:
*:*"
    },
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:linux_kernel:linux_kernel:5.10.121-linuxkit_\\(root\\@buildkitsandbox\\)_\\#1_SMP_Fri_Dec_2_10\\:35\\:42_UTC_2022:*:*:*:*:*:
*:*"
    },
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:linux:linux-kernel:5.10.121-linuxkit_\\(root\\@buildkitsandbox\\)_\\#1_SMP_Fri_Dec_2_10\\:35\\:42_UTC_2022:*:*:*:*:*:*:*"
    },
    {
     "referenceType": "cpe23Type",
     "referenceLocator": "cpe:2.3:a:linux:linux_kernel:5.10.121-linuxkit_\\(root\\@buildkitsandbox\\)_\\#1_SMP_Fri_Dec_2_10\\:35\\:42_UTC_2022:*:*:*:*:*:*:*"
    },
    {
     "referenceType": "purl",
     "referenceLocator": "pkg:linux-kernel/linux-kernel@5.10.121-linuxkit"
    }
   ]
  }
 ],
 "relationships": [
  {
   "spdxElementId": "SPDXRef-DOCUMENT",
   "relatedSpdxElement": "SPDXRef-DOCUMENT",
   "relationshipType": "DESCRIBES"
  }
 ]
}
```

wagoodman:
Nice!

One comment on the version value: `5.10.121-linuxkit (root@buildkitsandbox) #1 SMP Fri Dec 2 10:35:42 UTC 2022` seems surprising. From a glance it looks like the version is `5.10.121-linuxkit` and the remaining fields would be aux fields on the metadata. That would also fix a lot of what's going on with the CPEs being generated.

wagoodman:
> handling duplicates, e.g. we find the kernel, but the kernel also is part of an install package

as long as the `pkg.Location` is correct on the package you're generating and the package manager claims owning that path, then there will be a `ownership-by-file-overlap` relationship created between the two packages automatically (this one https://github.com/anchore/syft/blob/81b87dd1085757af4cf07bfb3cf9c96e4e1b4ed4/syft/artifact/relationship.go#L8). We don't deduplicate these packages by convention since they inherently have unique properties... one would be an RPM/DPKG/etc that describes basic packaging metadata and the other would be a package that has detailed metadata about the kernel itself from your analysis.

wagoodman:
> listing files as part of the package vs just having the package vs just having files

I'm not certain I 100% follow yet, but if you are looking to have "owned files" as part of the concept of the kernel package then all you would need to do is to have the package metadata implement the FileOwner interface (https://github.com/anchore/syft/blob/main/syft/pkg/file_owner.go) and this should be automagically handled downstream in syft (example: https://github.com/anchore/syft/blob/main/syft/pkg/rpm_metadata.go#L50-L60).

deitch:
> One comment on the version value: 5.10.121-linuxkit (root@buildkitsandbox) #1 SMP Fri Dec 2 10:35:42 UTC 2022 seems surprising. From a glance it looks like the version is 5.10.121-linuxkit and the remaining fields would be aux fields on the metadata. That would also fix a lot of what's going on with the CPEs being generated (and I think eliminate 2 of them being generated).

I changed it to just use the "base" version.

> as far as I can tell there isn't a "kernel" (or "kernel"-like) pURL to use. This feels like it should be a generic pURL: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic ... so pkg:generic/linux-kernel@5.10.121-linuxkit. I'm not certain if there are any restrictions for the get params for the generic pURL, you might be able to add on a few more useful bits of identifying information there.

I changed it to `generic`, but even the language there says, "please use something else":

> When possible another or a new purl type should be used instead of using the generic type and eventually
> contributed back to this specification.

It has nothing to capture the additional meta information other than the usual known qualifiers, listed [here](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs), which basically are `repository_url`, `download_url`, `vcs_url`, `filename` and `checksum`.

So where else should we capture it?

In the kernel itself, the whole thing is a single string. To use my example above, `5.10.121-linuxkit (root@buildkitsandbox) #1 SMP Fri Dec 2 10:35:42 UTC 2022` is a single string at a known location in the kernel file.

When you run `uname`, it treats the first part as "kernel release" (`uname -r`) and the rest as "kernel version" (`uname -v`).  This is based on [utsname.h](https://man7.org/linux/man-pages/man2/uname.2.html).

deitch:
I updated this to be able to scan kernel modules as well. It is successfully finding them, parsing them, getting module information out of them.

But then it needs to add them to a package, including per-module metadata. I assumed that would be the same as the kernel package, but if I do anything different (e.g. `Locations`, which is different per kernel file or module), I end up with hundreds packages, one per module and kernel file, which wasn't what we wanted.

So how do I:

1. Add each of these files (kernel module) to the same pre-existing kernel package?
2. Include the file information for each such kernel and module?
3. Include the actual `Location` without accidentally creating a new package?
4. Include the per-file (i.e. per kernel module) metadata?

deitch:
Also, in doing the purl for the kernel as `generic/linux-kernel`, it messed up some of the tests. I had to do some small but interesting reconstruction of `TypeFromPURL()` to get it to work. Whether or not we want this is up for question.

wagoodman:
The short answer to some of your questions is that the cataloger your writing needs to use the `FileResolver` directly to look for and catalog supporting files (such as the kernel modules). That is, the vmlinuz file is the primary evidence location you're cataloging (the kernel itself) and the kernel modules are more supporting evidence. I think this structure meets what you're looking for (as an example):
```json
...
  {
   "id": "ffe5a39e57372695",
   "name": "linux-kernel",
   "version": "6.2.8-060208-generic",
   "type": "generic/linux-kernel",
   "foundBy": "linux-kernel-cataloger",
   "locations": [
    {
     "path": "vmlinuz-6.2.8-060208-generic"
    }
    {
     "path": "mods/crc32c-intel.ko"
    },
   ],
   "licenses": [],
   "language": "",
   "cpes": [
    "cpe:2.3:a:linux-kernel:linux-kernel:6.2.8-060208-generic:*:*:*:*:*:*:*",
    "cpe:2.3:a:linux-kernel:linux_kernel:6.2.8-060208-generic:*:*:*:*:*:*:*",
    "cpe:2.3:a:linux_kernel:linux-kernel:6.2.8-060208-generic:*:*:*:*:*:*:*",
    "cpe:2.3:a:linux_kernel:linux_kernel:6.2.8-060208-generic:*:*:*:*:*:*:*",
    "cpe:2.3:a:linux:linux-kernel:6.2.8-060208-generic:*:*:*:*:*:*:*",
    "cpe:2.3:a:linux:linux_kernel:6.2.8-060208-generic:*:*:*:*:*:*:*"
   ],
   "purl": "pkg:generic/linux-kernel@6.2.8-060208-generic",
   "metadataType": "KernelPackageMetadata",
   "metadata": {
    "architecture": "x86",
    "version": "6.2.8-060208-generic",
    "extendedVersion": "6.2.8-060208-generic (kernel@sita) #202303220943 SMP PREEMPT_DYNAMIC Wed Mar 22 13:50:04 UTC 2023",
    "format": "bzImage",
    "videoMode": "Video mode 65535",
    "modules": [
     {
      "kernelVersion": "4.18.0-448.el8.x86_64",
      "versionMagic": "4.18.0-448.el8.x86_64 SMP mod_unload modversions ",
      "sourceVersion": "1E7D107C937AAE2A22F9942",
      "author": "Austin Zhang <austin.zhang@intel.com>, Kent Liu <kent.liu@intel.com>",
      "license": "GPL",
      "name": "crc32c_intel",
      "description": "CRC32c (Castagnoli) optimization using Intel Hardware.",
      "path": "mods/crc32c-intel.ko"
     }
    ]
   }
  }
...
```

I generated something like this that uses your kernel cataloger but instead of adding another cataloger for the modules, have the kernel Cataloger continue to search for modules once a kernel is found. This gives you the most flexibility for the kind of capabilities you're looking for I think.

I just pushed some code, but consider it to be more of a "draft", as there are still some problems to solve. The biggest one is to make certain that the modules found really do correspond to the kernel that it's being paired with. (also just to be clear, feel free to throw away any code I'm pushing to your)

deitch:
Thanks, I will look at your push and see if I can update it.

deitch:
I see what you did, that makes sense. A new cataloger is only for a new package entry. Once you hit a cataloger, it should find all of its children.

I don't like having all of those hundreds of files in `sourceInfo`; kernel modules usually are at standard locations, so I am going to rationalize that a bit.

The part I don't get is about I get additional files. You wrote:

> By adding the module information as a child of the kernel metadata.

I don't see a standard for it. The definition of `Metadata` is `interface{}`, so it could be anything. How do I create information about files such that syft will know to use those as child files?

deitch:
Cleaned up the `sourceInfo` sanely. It should handle the trees correctly.

How I actually add the files, though, still is beyond me.

wagoodman:
> The only thing I am missing, then, is the metadata on those files. Is there any way to report on them? We extract them and make them part of the []Modules that is part pf the KernelMetadata, but not sure how to get that reported it.

Currently syft-json and CycloneDX formats support showing these attributes. For syft-json there is nothing left to do, it will be shown automatically. For CyloneDX you will need to map the specific parameters that are worth capturing with the `cyclonedx` struct tag (example https://github.com/anchore/syft/blob/main/syft/pkg/golang_metadata.go#L5).

>  It wants to report checksums on everything. Maybe [comment field](https://spdx.github.io/spdx-spec/v2.3/file-information/#812-file-comment-field)?

I don't think I entirely understand your comment about checksums. The file digest cataloger can capture checksums for files and can be enabled with the `SYFT_FILE_METADATA_CATALOGER_ENABLED=true` env var.

Capturing structured data in the comments field is something that we try to avoid when possible. We have very few exceptions to this. Capturing arbitrary metadata into SPDX is a current limitation of the format.

wagoodman:
Also, regarding the license check failure:
```
Unallowable license () from "github.com/deitch/magic/pkg/magic"
Unallowable license () from "github.com/deitch/magic/pkg/magic/internal"
Unallowable license () from "github.com/deitch/magic/pkg/magic/parser"
failed validation
```

Mind adding a license in your repo?

deitch:
Rebased to pull your changes, added the license to magic, updated the version of magic used here.

deitch:
And added the cyclonedx metadata.

wagoodman:
re: checksums: we'll be turning on checksumming of files that are owned by packages in the near future. For now it's an opt-in feature via the env var.

deitch:
Ran the latest. Still not giving any file metadata:

```json
 "files": [
  {
   "fileName": "lib/modules/5.10.121-linuxkit/kernel/crypto/asymmetric_keys/asym_tpm.ko",
   "SPDXID": "SPDXRef-f23cf25c784db099",
   "checksums": [
    {
     "algorithm": "SHA1",
     "checksumValue": "0000000000000000000000000000000000000000"
    }
   ],
   "copyrightText": ""
  },
  {
   "fileName": "lib/modules/5.10.121-linuxkit/kernel/crypto/asymmetric_keys/pkcs8_key_parser.ko",
   "SPDXID": "SPDXRef-30d8022d6fa94550",
   "checksums": [
    {
     "algorithm": "SHA1",
     "checksumValue": "0000000000000000000000000000000000000000"
    }
   ],
   "copyrightText": ""
  },
```

wagoodman:
> Ran the latest. Still not giving any file metadata

The output you provided is `spdx-json` and to be clear all SPDX formats don't officially support arbitrary structured fields (like CycloneDX). I was trying to point out that `syft-json` should work out of the box (not `spdx-json`).

wagoodman:
One open question I have: how do we know that the kernels we've found are positively associated with the modules we also discover? It can be normal to have multiple kernel archives found but the modules really only apply to the kernel you're using at the time.

deitch:
Ah, so only cyclone and soft-json. Got it.

> One open question I have: how do we know that the kernels we've found are positively associated with the modules we also discover? It can be normal to have multiple kernel archives found but the modules really only apply to the kernel you're using at the time.

In general, kernel modules will be at `lib/modules/<kernel-release>/`, so in the above example `lib/modules/5.10.121-linuxkit/`. The cataloger here doesn't check the actual path, but the version I originally had did check that the kernel version, as provided by the information in the module itself, matches that, which is where it would sit.

I will add that back in; TBH, I don't recall when it went away.

deitch:
Updated. It now checks that the kernel version in the module matches the one from the kernel, and that the path is correct as well.

wagoodman:
@deitch  after looking closer at some of the attributes and doing a little more research, I think it makes sense to capture the kernel modules as different distinct packages that have relationships back to the kernel package.

Also I think some tweaking of names such as `Kernel` should be `LinuxKernel` in case we start supporting other kernels to catalog. Leaving the package as `kernel` makes sense for now, future kernel catalogers could reside together.

I've made some of these adjustments locally and and push them to your branch if you don't mind?

wagoodman:
I pushed the rename change, lint fixes, and simplified the glob logic to lean on the resolver for modules. I could go further and push changes to break out modules into their own packages (similar to how they were before) and add the relationships logic if you'd like? I didn't want to pull the trigger on that until I heard back from you on your opinion here.

deitch:
My concern with separate packages is that the vast majority of kernel modules actually are part of the same kernel source. When you compile a kernel and install it, you compile dozens (actually, hundreds) of those modules and install them. So they really _are_ a single package. But then there are external modules you might have compiled and inserted.

I will look at the logic you have here.

deitch:
Ah, you didn't change it yet.

What would the output look like if you changed it? Can you take the sample from above (either the syft-json or the spdx-json) and mock up what it looks like?

Or, if that is a pain, push the change as an additional commit. I can then download and run it, and we can compare, and always choose to keep the commit or discard it?

syft/pkg/linux_kernel_package_metadata.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"sort"
+
+	"github.com/scylladb/go-set/strset"
+)
+
+// LinuxKernelPackageMetadata represents all captured data for a Linux kernel
+type LinuxKernelPackageMetadata struct {
+	Name            string                      `mapstructure:"name" json:"name" cyclonedx:"name"`
```
wagoodman, line 11: are we ever expecting a name here? So far I've only found blank values here (I haven't looked for that long though)

wagoodman:
Let me expose some of the reasoning for separating kernel packages from the modules and describe the modules as separate packages:
1. Modules are not guaranteed to originate from the kernel source. As long as you have the kernel headers you can write, build, and package your own kernel modules (still tailored for a target kernel version / flavor)
2. Modules are not distributed with the kernel. In the RHEL ecosystem, for example, you have the `kernel` package and the `kernel-modules` package that is installed separately.
4. Secondary reason (adds to the pile, but not enough of a reason alone to make a decision one way or another): There are several kernel modules that syft would find on a system, which would make the specific kernel package entry very large.

What might this look like? Here's some syft JSON with an example:
```json
{
 "artifacts": [
  {
   "id": "6de951060b851e8c",
   "name": "linux-kernel",
   "version": "4.18.0-448.el8.x86_64",
   "type": "linux-kernel",
   "foundBy": "linux-kernel-cataloger",
   "locations": [
      {
       "path": "lib/modules/4.18.0-448.el8.x86_64/vmlinuz"
      }
    ],
   "purl": "pkg:generic/linux-kernel@4.18.0-448.el8.x86_64",
   "metadataType": "LinuxKernelPackageMetadata",
   "metadata": {
    "name": "",
    "architecture": "x86",
    "version": "4.18.0-448.el8.x86_64",
    "extendedVersion": "4.18.0-448.el8.x86_64 (mockbuild@kbuilder.bsys.centos.org) #1 SMP Wed Jan 18 15:02:46 UTC 2023",
    "format": "bzImage",
    "videoMode": "Video mode 65535"
  },
  {
   "id": "aa1b8a51805e6c0",
   "name": "blowfish_x86_64",
   "version": "E495BAFA5CF2920A3048AA7",
   "type": "linux-kernel-module",
   "foundBy": "linux-kernel-cataloger",
   "license": "GPL",
   "locations": [
      {
       "path": "lib/modules/4.18.0-448.el8.x86_64/kernel/arch/x86/crypto/blowfish-x86_64.ko"
      }
    ],
   "purl": "pkg:generic/linux-kernel-module/blowfish_x86_64@E495BAFA5CF2920A3048AA7",
   "metadataType": "LinuxKernelModulePackageMetadata",
   "metadata":  {
      "name": "blowfish_x86_64",
      "sourceVersion": "E495BAFA5CF2920A3048AA7",
      "description": "Blowfish Cipher Algorithm, asm optimized",
      "kernelVersion": "4.18.0-448.el8.x86_64",
      "versionMagic": "4.18.0-448.el8.x86_64 SMP mod_unload modversions ",
      "parameters": {
       "force": {
        "description": "Force module load, ignore CPU blacklist"
       }
      }
    }
  }
],
"artifactRelationships": [
  {
    "parent":  "6de951060b851e8c",    # note: the linux-kernel
    "child": "aa1b8a51805e6c0",    # note: the kernel module
    "type": "RUNTIME_DEPENDENCY_OF",    # note: still searching for the right value here...
  }

]
}
```


So the kernel and the module would be sibling packages with a relationship that shows how the module is related to which kernel.

syft/pkg/linux_kernel_package_metadata.go
deitch: It is theoretically possible, IIRC.

deitch:
It kind of cuts both ways, doesn't it? It could be installed together or separately. It would be nice if there were a standard for it.

OK, let's try this as separate modules in another commit. I will run it and see what it looks like.

wagoodman:
> It kind of cuts both ways, doesn't it? It could be installed together or separately.

Right, so I think the safer representation is to have them separate by default then. If the relationship approach is untenable from the consumer point of view or we can't find the right relationship type then I'm ok with keeping them as 1 package. But I think we should try and shoot for a multi-package approach.

(@anchore/tools  curious for others thoughts on this too)

syft/pkg/cataloger/kernel/parse_linux_kernel_module_file.go
```diff
@@ -0,0 +1,182 @@
...
+	"strings"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/internal/unionreader"
+	"github.com/anchore/syft/syft/source"
+)
+
+func findLinuxKernelModules(resolver source.FileResolver, pkgLocations *source.LocationSet, kernelVersion string) ([]pkg.LinuxKernelModuleMetadata, error) {
+	locations, err := resolver.FilesByGlob(fmt.Sprintf("**/lib/modules/%s/**/*.ko", kernelVersion))
```
wagoodman, line 15: I've noticed that there are some kernel modules that are distributed as `.ko.xz` files. Should we attempt to catalog these? is this common of distributions? Or is `.ko` more common?

deitch:
> But I think we should try and shoot for a multi-package approach.

If you have it figured out, push it, and I will run a test?

syft/pkg/cataloger/kernel/parse_linux_kernel_module_file.go
deitch: `.ko` is more common, as far as I know, but compressed modules are valid as well.

I think that this works as is (subject to the kernel module packages, which I am awaiting your push to test it), so let's use it, and not delay it for `.ko.xz` files.  We always can have a separate small PR to add it, which will be easy to do .

wagoodman:
I can do that 👍 I'll push something on Monday

syft/pkg/cataloger/kernel/package.go
```diff
@@ -0,0 +1,71 @@
...
+	var licenses []string
+	if metadata.License != "" {
+		licenses = []string{metadata.License}
+	} else {
+		licenses = []string{}
+	}
+
+	p := pkg.Package{
+		Name:         metadata.Name,
+		Version:      metadata.Version,
```
wagoodman, line 39: I found that the version is usually always empty for modules... is there ever a case where this isn't empty? Should we be using the source version instead (which is never empty)?

deitch: It is. I don't see it all that often, but it is used. If we are to be complete, we need to leave it. The kernelVersion is different.

deitch:
> also removed the glob appends since I didn't see a strong motivation for them -- shout out if you feel otherwise

I like the idea that someone can say, "I use other kernel files or other kernel module files". Might as well have it, if it does not hurt.

syft/pkg/cataloger/kernel/package.go
wagoodman: if it is used that's good enough. I know we shouldn't use the kernel version, but the source version looked like another good candidate if the version was always empty.

wagoodman:
> I like the idea that someone can say, "I use other kernel files or other kernel module files".

This could be true, but at this time we haven't added application configuration options yet to tweak other cataloger globs. For consistency I think we should omit this for now. I do have a feeling that these globs (for all catalogers) will be available in the application configuration some day though.

wagoodman:
> It handles the relationships correctly for syft-json, but spdx-json is missing it. Do you know why? 

I think I do, I'll push that fix shortly

wagoodman:
Json schema diff for reviewers
```diff
# ❯ diff schema/json/schema-7.1.2.json schema/json/schema-7.1.3.json
749a750,843
>     "LinuxKernelMetadata": {
>       "properties": {
>         "name": {
>           "type": "string"
>         },
>         "architecture": {
>           "type": "string"
>         },
>         "version": {
>           "type": "string"
>         },
>         "extendedVersion": {
>           "type": "string"
>         },
>         "buildTime": {
>           "type": "string"
>         },
>         "author": {
>           "type": "string"
>         },
>         "format": {
>           "type": "string"
>         },
>         "rwRootFS": {
>           "type": "boolean"
>         },
>         "swapDevice": {
>           "type": "integer"
>         },
>         "rootDevice": {
>           "type": "integer"
>         },
>         "videoMode": {
>           "type": "string"
>         }
>       },
>       "type": "object",
>       "required": [
>         "name",
>         "architecture",
>         "version"
>       ]
>     },
>     "LinuxKernelModuleMetadata": {
>       "properties": {
>         "name": {
>           "type": "string"
>         },
>         "version": {
>           "type": "string"
>         },
>         "sourceVersion": {
>           "type": "string"
>         },
>         "path": {
>           "type": "string"
>         },
>         "description": {
>           "type": "string"
>         },
>         "author": {
>           "type": "string"
>         },
>         "license": {
>           "type": "string"
>         },
>         "kernelVersion": {
>           "type": "string"
>         },
>         "versionMagic": {
>           "type": "string"
>         },
>         "parameters": {
>           "patternProperties": {
>             ".*": {
>               "$ref": "#/$defs/LinuxKernelModuleParameter"
>             }
>           },
>           "type": "object"
>         }
>       },
>       "type": "object"
>     },
>     "LinuxKernelModuleParameter": {
>       "properties": {
>         "type": {
>           "type": "string"
>         },
>         "description": {
>           "type": "string"
>         }
>       },
>       "type": "object"
>     },
1027a1122,1127
>             },
>             {
>               "$ref": "#/$defs/LinuxKernelMetadata"
>             },
>             {
>               "$ref": "#/$defs/LinuxKernelModuleMetadata"

```

wagoodman (approved):

deitch:
🥳

deitch:
Thanks for helping walk me through this @wagoodman @kzantow !

PR #1696, opened 2023-03-24
Add Nix cataloger

### LINKED ISSUE #462: Add Nix Cataloger

wagoodman:
This builds on the original work from @juliosueiras on https://github.com/anchore/syft/pull/1107 . The changes include:

- [x] pulling the latest changes from main in
- [x] refactoring to use resolver.AllLocations
- [x] moving the set of owned locations to the pkg.NixStoreMetadata and leaving the pkg.Locations to still indicate the locations where the existence of the package is evident (just the nix store path directory, not the contents)
- [x] refactored the nix store path parsing and added unit tests
- [x] added integration tests

I'm not very familiar with the Nix ecosystem so review from someone more familiar with Nix would be very helpful! (cc: @flokli if you have the time 🙏 )

## Details

I want to be clear on how this is working and what the consequences are. This implementation looks at only what directories are found in `/nix/store` and will construct packages for each output directory found (which means derivations are not included). This means that packages that have multiple outputs will seemingly be unrelated. This is because there is not enough metadata locally in the scan target to determine what outputs should be related. 

For example, if you install python37 from nixpkgs, you will see two outputs, `out` (the default, so has no name) and `debug`:
```
$ nix-env -iA nixpkgs.python37
$ ls -al /nix/store | grep 'python3-3.7'
dr-xr-xr-x 3 root root     4096 Jan  1  1970 13nr17z3x46xknz27gn9p77wxnll2sxl-python3-3.7.16-debug
dr-xr-xr-x 7 root root     4096 Jan  1  1970 a1gzdpyn8bbp192cavnxl6sa3snjx563-python3-3.7.16
-r--r--r-- 1 root root    11859 Jan  1  1970 y72x7nz63cw29lzy239acwcc7gswljn2-python3-3.7.16.drv
```

When running syft, you will see two `python3@3.7.16` packages, one for each output:
```
$ syft -o json ... | jq '.artifacts[] | select(.name == "python3")'
{
  "id": "5225a10fc861ef0e",
  "name": "python3",
  "version": "3.7.16",
  "type": "nix",
  "foundBy": "nix-store-cataloger",
  "locations": [
    {
      "path": "/nix/store/a1gzdpyn8bbp192cavnxl6sa3snjx563-python3-3.7.16",
      "layerID": "sha256:01c6ae491e3fc339b93ccf883079fba571df6ddfdaa3ed5c2ac91578ce85a1f2"
    }
  ],
  "licenses": [],
  "language": "",
  "cpes": [
    "cpe:2.3:a:python3:python3:3.7.16:*:*:*:*:*:*:*"
  ],
  "purl": "pkg:nix/python3@3.7.16?outputhash=a1gzdpyn8bbp192cavnxl6sa3snjx563",
  "metadataType": "NixStoreMetadata",
  "metadata": {
    "outputHash": "a1gzdpyn8bbp192cavnxl6sa3snjx563",
    "files": [ ... snip ... ]
},
{
  "id": "8c1c788327e1f52e",
  "name": "python3",
  "version": "3.7.16",
  "type": "nix",
  "foundBy": "nix-store-cataloger",
  "locations": [
    {
      "path": "/nix/store/13nr17z3x46xknz27gn9p77wxnll2sxl-python3-3.7.16-debug",
      "layerID": "sha256:01c6ae491e3fc339b93ccf883079fba571df6ddfdaa3ed5c2ac91578ce85a1f2"
    }
  ],
  "licenses": [],
  "language": "",
  "cpes": [
    "cpe:2.3:a:python3:python3:3.7.16:*:*:*:*:*:*:*"
  ],
  "purl": "pkg:nix/python3@3.7.16?output=debug&outputhash=13nr17z3x46xknz27gn9p77wxnll2sxl",
  "metadataType": "NixStoreMetadata",
  "metadata": {
    "outputHash": "13nr17z3x46xknz27gn9p77wxnll2sxl",
    "output": "debug",
    "files": [ ... snip ... ]
}
```

The pURLS are nicely distinct:
- "pkg:nix/python3@3.7.16?outputhash=a1gzdpyn8bbp192cavnxl6sa3snjx563",
- "pkg:nix/python3@3.7.16?output=debug&outputhash=13nr17z3x46xknz27gn9p77wxnll2sxl"

But the package names are not unique, which is technically correct, but it would be more ideal to try and relate or merge these packages in the future.

Closes #462


wagoodman:
JSON schema diff:
```diff
# $ diff schema/json/schema-7.0.1.json schema/json/schema-7.0.2.json
848a849,868
>     "NixStoreMetadata": {
>       "properties": {
>         "hash": {
>           "type": "string"
>         },
>         "output": {
>           "type": "string"
>         },
>         "files": {
>           "items": {
>             "type": "string"
>           },
>           "type": "array"
>         }
>       },
>       "type": "object",
>       "required": [
>         "hash"
>       ]
>     },
1010a1031,1033
>             },
>             {
>               "$ref": "#/$defs/NixStoreMetadata"
```

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_findVersionIsh(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   string
+		wantIdx int
+		wantEx  string
+		wantPr  string
```
flokli, line 16: Is there a better name for this?

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
```diff
@@ -0,0 +1,284 @@
...
+			wantIdx: -1,
+			wantEx:  "",
+			wantPr:  "",
+		},
+		{
+			name:    "semver with overbite into output",
+			input:   "/nix/store/h0cnbmfcn93xm5dg2x27ixhag1cwndga-glibc-2.34-210-bin",
+			wantIdx: 50,
+			wantEx:  "2.34-210-bin",
+			wantPr:  "210-bin",
```
flokli, line 30: `bin` is just the name of the output, it's not part of the version. There can be multiple outputs with the same "version number", but different output names. I'd probably just drop the optional -{bin,lib,dev,doc,devdoc} (these are the most common ones) output name from the version number - or is there a notion of multiple package outputs in syft?

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		{
+			name:    "multiple versions",
+			input:   "5zzrvdmlkc5rh3k5862krd3wfb3pqhyf-perl5.34.1-TimeDate-2.33",
+			wantIdx: 53,
+			wantEx:  "2.33",
+			wantPr:  "",
```
flokli, line 37: Is there any checking of the package name that gets extracted? This is TimeDate at version 2.33, in a perl 5.34.1 environment.

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
```diff
@@ -0,0 +1,284 @@
...
+			wantPr:  "3.1",
+		},
+		{
+
+			name:    "long name",
+			input:   "0296qxvn30z9b2ah1g5p97k5wr9k8y78-busybox-static-x86_64-unknown-linux-musl-1.35.0",
+			wantIdx: 74,
+			wantEx:  "1.35.0",
+			wantPr:  "",
+		},
```
flokli, line 115: There's also a unstable-yyyy-mm-dd convention, if nixpkgs pulls in a version that hasn't had a release recently. I feel like it should also be part of the unit tests.

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
flokli: Ah, it's further down. Maybe this test function should have a comment at the top, stating that the names themselves are tested in `Test_parseNixStorePath(` further down.

syft/pkg/nix_store_metadata.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"sort"
+
+	"github.com/scylladb/go-set/strset"
+)
+
+type NixStoreMetadata struct {
+	// Hash is the prefix of the nix store basename path
+	Hash string `mapstructure:"hash" json:"hash"`
```
flokli, line 11: As written above, I'd rather call this OutputHash.

syft/pkg/nix_store_metadata.go
```diff
@@ -0,0 +1,24 @@
...
+
+	// Output allows for optionally specifying the specific nix package output this package represents (for packages that support multiple outputs)
```
flokli, line 13: The default output also doesn't have this. Should probably be added in the comment.

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
wagoodman: yeah, let me rename all of these (ex = version, pr = prerelease)

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
wagoodman: Question: is `glibc-2.34` really the "package" and the single package has multiple outputs? If that's the case, there is more work to be done -- the cataloger should be merging these into a single package, and the metadata for each needs to account for expressing multiple outputs (and surrounding metadata).

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
wagoodman: Just so I understand, is this the behavior that needs to be updated and tested for?
> If a package is not a release but a commit from a repository, then the version attribute must be the date of that (fetched) commit. The date must be in "unstable-YYYY-MM-DD" format.

https://nixos.org/manual/nixpkgs/stable/#sec-package-naming (I'll add this in a comment in the code for future folks to see)

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
flokli: This is the `bin` output for the glibc 2.34 build. It includes binaries produced by the build, like `bin/getent`. The glibc .so file would be in a separate output (`lib` or the default output, where the prefix is omitted).

It makes sense to distinguish between these outputs (so the info should not be merged), and it also makes sense to keep the output hash as an identifier.

We can have multiple glibc-2.34 in the same "closure" / filesystem (let's say one carrying a patch, the other not), and we can ship the `bin` outputs or not, depending on whether we want `getent` or not.

If there's a vulnerability, the output_path allows distinguishing between the patched version or unpatched version, and knowing which outputs you ship also helps determining whether you might be affected by a certain vulnerability and it's limited to the CLI parser, not the library.

I'm having a hard time calling this package, because it's nothing bundled up, and these are the output paths produced by a build. After évaluation, Nix uses "derivations" as build recipes. Each derivation is a single build, and it may produce multiple outputs (with output paths).

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
flokli: We should include an additional test for packages / paths that use unstable-yyyy-mm-dd.
I didn't try out piping such a path through the regex yesterday, but even if it works right now, i feel like it should be tested separately.

Example: https://github.com/NixOS/nixpkgs/blob/798e23beab9b5cba4d6f05e8b243e1d4535770f3/pkgs/servers/webdav-server-rs/default.nix#L11

In this case I'd expect the "package name" to be `webdav-server-rs`, and the version to be `unstable-2021-08-16`.

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+import (
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_findVersionIsh(t *testing.T) {
+	// note: only the package version fields are tested here, the name is tested for the test for parseNixStorePath below.
```
flokli, line 11:
```suggestion
	// note: only the package version fields are tested here, the name is tested in parseNixStorePath below.
```

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
wagoodman: It sounds like the best way forward is to collect these definitions into a single package, where the metadata would be able to track the multiple outputs:
```json
{
  "name": "glibc"
  "version": "2.34"
  "metadata": {
    "outputs": [
        {
          "output": "man"
          "outputhash": "cfmmhxh0c5nxnag1bixdg7d3g2c9wan2"
          "files": [...]
        }
        {
          "output": "bin"
          "outputhash": "h0cnbmfcn93xm5dg2x27ixhag1cwndga"
          "files": [...]
        }
      ]
  }
}
```

In this way there is still a single logical `glibc@2.34` package and grype downstream would be able to look at the `outputhash` for each output on the metadata.

> If there's a vulnerability, the output_path allows distinguishing between the patched version or unpatched version

it sounds like the `output_path` would be really important to try and capture on the metadata directly then? Just to double check... is this output path the path to the store (e.g. `/nix/store/hs0yi5n5nw6micqhy8l1igkbhqdkzqa1-foo`)? Or is this referring to another value?

> I'm having a hard time calling this package, because it's nothing bundled up, and these are the output paths produced by a build.

I can see what you're saying here (I think). I started with the assumption of what you might find on search.nixos.org ... say for glibc https://search.nixos.org/packages?channel=22.11&show=glibc&from=0&size=1&sort=relevance&type=packages&query=glibc 
Screenshot 2023-03-27 at 11 07 07 AM

I feel this orients the results in terms of the logical name + version and outputs available. I've been equating the name + version as the idea of a singular logical package and the list of outputs installed as something that is attached metadata to that (single) package. I see what you're saying with nothing is being bundled up but syft doesn't necessarily require that packaging ecosystems do any bundling of builds/files at all.

Is this a kosher way to think about "packages" in the Nix ecosystem? Or said another way, could representing singular packages with possibly multiple outputs be inaccurate or misleading in downstream analysis?

flokli: > it sounds like the output_path would be really important to try and capture on the metadata directly then? Just to double check... is this output path the path to the store (e.g. /nix/store/hs0yi5n5nw6micqhy8l1igkbhqdkzqa1-foo)? Or is this referring to another value?

Let me pick a better example:

If I take a current checkout of nixpkgs (4bb072f0a8b267613c127684e099a70e1f6ff106), there's multiple outputs of the `openssl` package, described in the "build recipe" (.drv):

```
$ nix repl .
nix-repl> openssl
«derivation /nix/store/3jbv3scpmj2kxhshfz544jq37dzywbss-openssl-3.0.8.drv»

nix-repl> openssl.outputs
[ "bin" "dev" "out" "man" "doc" "debug" ]

nix-repl> :b openssl

This derivation produced the following outputs:
  bin -> /nix/store/05qna7girdhnzx38jkhr4yyh7gmnizv3-openssl-3.0.8-bin
  debug -> /nix/store/spa8fwjwd6r4n6d4gqg4s1ylijmkc0x7-openssl-3.0.8-debug
  dev -> /nix/store/ggsmy3rwsr6q6ri45ck3jdqnz9cf6dm6-openssl-3.0.8-dev
  doc -> /nix/store/bsabnghysgdmfhjlj2pal3i25q9w2brf-openssl-3.0.8-doc
  man -> /nix/store/4ac38nfgh4whllggv4zwp14va4mfirr7-openssl-3.0.8-man
  out -> /nix/store/s8vg2h8xzqmjd72f3g3p1jqy2lbbapc6-openssl-3.0.8
```

So, all these output paths are produced by the same build recipe (`/nix/store/3jbv3scpmj2kxhshfz544jq37dzywbss-openssl-3.0.8.drv`).

 - The `bin` output of that derivation has an output hash of `05qna7girdhnzx38jkhr4yyh7gmnizv3` ([nixbase32](https://cs.tvl.fyi/depot/-/blob/tvix/nix-compat/src/nixbase32.rs?L1) encoding).
 - The `debug` output has an output hash of `spa8fwjwd6r4n6d4gqg4s1ylijmkc0x7`
 - The `dev` output has an output hash of `ggsmy3rwsr6q6ri45ck3jdqnz9cf6dm6`
 - ...

If I now go back in time to a previous nixpkgs checkout (2 weeks before, 1df7332c722cc8d235b979cfa8f1bbe949b722fb), I still have the same openssl version in that case, but the build recipe is a different one (and due to this, all the hashes).

```
$ nix repl .
nix-repl> openssl   
«derivation /nix/store/y43ka1zh6icrf515hc5qzb643wsscxmd-openssl-3.0.8.drv»

nix-repl> openssl.outputs
[ "bin" "dev" "out" "man" "doc" "debug" ]

nix-repl> :b openssl

This derivation produced the following outputs:
  bin -> /nix/store/b1rd646j6yc22qnhiydb5a1hi5hc4dky-openssl-3.0.8-bin
  debug -> /nix/store/8zw06pv5v5l7ggklyls77n5fw03vgdr7-openssl-3.0.8-debug
  dev -> /nix/store/gs1841l5x53q3r5l196ld9kxq45pvf82-openssl-3.0.8-dev
  doc -> /nix/store/42krirjplr0a4z3w2myhrawg5kbgmy40-openssl-3.0.8-doc
  man -> /nix/store/rrz6hdzxsxg3yadk02gcypib36bq2zqm-openssl-3.0.8-man
  out -> /nix/store/fgrj06y1x83fwh8hqgg02v9abc7a7b65-openssl-3.0.8
```

Now the `bin`output hash is `b1rd646j6yc22qnhiydb5a1hi5hc4dky`.

Both these packages can coexist in /nix/store.

In more complicated closures / "container contents", it's very well possible for, let's say, two slightly different flavors of libraries to be linked against two different binaries.

Key takeaways:
 - just the "name" (openssl-3.0.8) doesn't uniquely identify a package, the outputhash is crucial
 - The output hash uniquely identifies an output, as soon as the build recipe (or any of its dependencies) changes it'll differ
 - Just by looking at the paths, without any additional database (that's what syft is doing) it's not possible to map / group certain output paths to the .drv / build recipe producing it.

wagoodman: Thanks for the example / input here -- much appreciated! I think the takeaways make sense given the fact that syft is guessing from a directory structure instead of package metadata information. In the future I feel it would be ideal to improve this in some way to make it more obvious which outputs are related... but I don't see a safe way to do this given the information available and the non-guarentees on the sqlite cache that's there. Thanks for helping to navigate these traps @flokli !

Ok, back to your question (this was all a little bit of a tangent):

> bin is just the name of the output, it's not part of the version. There can be multiple outputs with the same "version number", but different output names. I'd probably just drop the optional -{bin,lib,dev,doc,devdoc} (these are the most common ones) output name from the version number - or is there a notion of multiple package outputs in syft?

This function first extracts the version as if it were a semver, then later in processing (in `parseNixStorePath`) the last non-numeric, dash-delimited string is assumed to be the nix output. That is, the package ultimately reflects your suggestion (to not include output names in the version).

There isn't a notion of a "package output" in syft, however, what it will look like to most users is that there are multiple instances of the package `name-version` installed, and only when they look at the `.metadata` or `.purl` will they see the specific output this package represents.

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
wagoodman: Updated 🙌  https://github.com/anchore/syft/pull/1696/commits/46bf1fc55a3a70f1c16c3b44c37ceff32edd9655

syft/pkg/cataloger/nix/cataloger_test.go
```diff
@@ -0,0 +1,55 @@
...
+		fixture  string
+		wantPkgs []pkg.Package
+		wantRel  []artifact.Relationship
+	}{
+		{
+			fixture: "test-fixtures/fixture-1",
+			wantPkgs: []pkg.Package{
+				{
+					Name:         "glibc",
+					Version:      "2.34-210",
```
kzantow, line 24: According to how I interpret [@flokli's comment](https://github.com/anchore/syft/pull/1696#discussion_r1148320063), the version here should be `2.34`, not including the `-210` (maybe it's the build number or something?)

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
```diff
@@ -0,0 +1,304 @@
...
+			wantIdx:        44,
+			wantVersion:    "006",
+			wantPreRelease: "",
+		},
+		{
+
+			name:           "prerelease with multiple dashes",
+			input:          "zkgyp2vra0bgqm0dv1qi514l5fd0aksx-bash-interactive-5.1-p16-man",
+			wantIdx:        50,
+			wantVersion:    "5.1-p16-man",
```
kzantow, line 98: similarly to the comment above, I think we at least don't want the `-man` suffix?

syft/pkg/cataloger/nix/cataloger.go
```diff
@@ -0,0 +1,101 @@
...
+		parentStorePath := findParentNixStorePath(location.RealPath)
+		if parentStorePath != "" {
+			if _, ok := filesByPath[parentStorePath]; !ok {
+				s := source.NewLocationSet()
+				filesByPath[parentStorePath] = &s
+			}
+			filesByPath[parentStorePath].Add(location)
+		}
+
+		if !matchesStorePath {
```
kzantow, line 48: shouldn't this happen before `s := source.NewLocationSet() ... filesByPath[parentStorePath] = &s`? and this would go away if using `FilesByGlob`?

syft/pkg/cataloger/nix/cataloger.go
```diff
@@ -0,0 +1,101 @@
...
+			continue
+		}
+		parentStorePath := locations[0].RealPath
+		files, ok := filesByPath[parentStorePath]
+		if !ok {
+			log.WithFields("path", parentStorePath, "nix-store-path", parentStorePath).Warn("found a nix store file for a non-existent package")
+			continue
+		}
+		appendFiles(p, files.ToSlice()...)
+		continue
```
kzantow, line 77: ?

syft/pkg/cataloger/nix/package.go
```diff
@@ -0,0 +1,59 @@
...
+		qualifiers = append(qualifiers,
+			packageurl.Qualifier{
+				Key:   "outputhash",
+				Value: storePath.outputHash,
+			},
+		)
+	}
+	pURL := packageurl.NewPackageURL(
+		// TODO: nix pURL type has not been accepted yet (only proposed at this time)
+		"nix",
```
kzantow, line 52: we have had issues raised when PURLs don't match the spec (e.g. the nuget/dotnet thing), I'm not sure we should output potentially invalid PURLs

syft/pkg/cataloger/nix/cataloger_test.go
wagoodman: That particular comment is referring to the `-bin` suffix, which refers to the specific "output" called `bin`. The version for glibc does appear to have the `-###` suffix (which is different than the output):
Screenshot 2023-04-03 at 1 37 57 PM

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
wagoodman: This is for testing `findVersionIsh`, which only does semver-like parsing. The parent function `parseNixStorePath` is responsible for dealing with removing suffixes which are related to the output name (there are a separate set of tests below that check this).

syft/pkg/cataloger/nix/cataloger.go
```diff
@@ -0,0 +1,101 @@
...
+func (c *Cataloger) Name() string {
+	return catalogerName
+}
+
+func (c *Cataloger) Catalog(resolver source.FileResolver) ([]pkg.Package, []artifact.Relationship, error) {
+	// we want to search for only directories, which isn't possible via the stereoscope API, so we need to apply the glob manually on all returned paths
+	var pkgs []pkg.Package
+	var filesByPath = make(map[string]*source.LocationSet)
+	for location := range resolver.AllLocations() {
+		matchesStorePath, err := doublestar.Match(nixStoreGlob, location.RealPath)
```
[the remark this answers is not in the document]
kzantow: Ah -- I missed that detail 🤦

syft/pkg/cataloger/nix/parse_nix_store_path_test.go
kzantow: Oh, I see 👍

syft/pkg/cataloger/nix/cataloger.go
wagoodman: >  and this would go away if using FilesByGlob:

the search is for the `/nix/store/<hash>-<name>-<version>-<output>` directory, not a file. We are collecting locations that belong to a particular store path within the same loop that we are looking for store paths to not make redundant request: why ask for `/nix/store/PATH/**` when I've already seen all of the locations in an earlier loop?

> shouldn't this happen before s := source.NewLocationSet() ... filesByPath[parentStorePath] = &s

I don't think so; we want to capture:
a) `/nix/store/PATH` : the location to capture as a package
b) `/nix/store/PATH/**`: all files/directories that are "owned" by that package

We first check if the path is a parent of a store path (e.g. when cataloging `/nix/store/PATH/a/b/c` it's parent store path is `/nix/store/PATH`). We only continue forth if we have a new store path to catalog, in which case we need to parse the directory name to understand what the package is.

kzantow: Right, sorry I understand now about capturing directories

wagoodman:
As a bit of a side note, this thread was interesting and helpful in understanding more about the future of Nix in terms of capturing the right details in an SBOM and how that would relate to vulnerability analysis https://discourse.nixos.org/t/the-future-of-the-vulnerability-roundups/22424 .

kzantow (approved):

wagoodman:
@flokli I think this is ready for merging, but I wanted to hold off and get your take first. Are there any other changes relative to the Nix ecosystem that should be considered before considering this done? (or any nuance that I might have missed in your previous comments that is not addressed?)

raboof:
Cool addition! I took the liberty of starting a Discourse thread about the use of purl for Nix packages at https://discourse.nixos.org/t/package-urls-purl-for-nix-packages

PR #1698, opened 2023-03-27
Add package-to-file location evidence relationships

### LINKED ISSUE #1534: Syft decode does not handle files
Today you can encode an `sbom.SBOM` into a syft json `model.Document`, which includes `Files`. However, you cannot decode the information from `Files` back into the specific `sbom.SBOM.Artifaacts` sections.

wagoodman:
Each package tracks which files contributed to syft being able to detect the package's existence via the `pkg.Package.Locations` field. This PR exposes these locations as package-to-file relationships to show that package X is "evident by" file Y. This is made possible with the new `evident-by` relationship type, which maps to the `OTHER` SPDX relationship type and depends on a comment value to support decoding effectively. 

The current implementation takes all locations from `pkg.Package.Location` with the "primary" evidence annotation and makes relationships for them. This is nice since there is a single, downstream point in the code that would be responsible for creating these relationships. 

In the future application configuration can be introduced which allows for supporting evidence to also be represented with an `evident-by` relationship type.

This PR also:
- Fixes a small problem where decoding SPDX relationships of the type OTHER meant for "ownership-by-file-overlap" was interpreted as "dependency-of" instead
- Fixes #1534 since this PR adds file nodes to the SBOM which would not be able to be interpreted on multiple encode-decode-encode cycles) which would fail the encode-decode-encode integration test (specifically for the syft json format).

CC: @jedevc @cdupuis


wagoodman:
For interested folks, this PR will be refactored to use location annotations being introduced in #1723 and allow for expressing only primary evidence.

wagoodman:
```diff
# ❯ diff schema/json/schema-7.1.3.json schema/json/schema-7.1.4.json
571a572,574
>         },
>         "size": {
>           "type": "integer"
580c583,584
<         "mimeType"
---
>         "mimeType",
>         "size"
```

spiffcs (approved):

PR #1703, opened 2023-03-30
feat: Add config option to allow user to select the default image source location

spiffcs:
## Add config option to allow user to select the default image source location

Syft is consumed by upstream tools for generating SBOM. These SBOM can be generated from a variety of sources including local and OCI registry images. 

https://github.com/anchore/grype/issues/1204 surfaces the need for allowing a user to express a preference over the default source to be used when building an SBOM for an image (and then subsequently generating a vulnerability scan).

This first step adds a config option into syft and changes the library behavior so this preference can be injected into the SBOM generation process. A follow up PR will be made for grype to adapt to this new ability.

##### Registry value pulls source from OCI registry
Screenshot 2023-03-30 at 12 44 33 PM

##### Docker value pulls source from
Screenshot 2023-03-30 at 12 45 02 PM
 local Docker Daemon

###### No value keeps the same current behavior of default order lookup
Screenshot 2023-03-30 at 12 45 25 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,8 +110,6 @@ The above output includes only software that is visible in the container (i.e., 
 syft <image> --scope all-layers
 ```
 
-
-
 ## Supported sources
 
 Syft can generate a SBOM from a variety of sources:
@@ -140,6 +138,10 @@ dir:path/to/yourproject                  read directly from a path on disk (any 
 file:path/to/yourproject/file            read directly from a path on disk (any single file)
 registry:yourrepo/yourimage:tag          pull image directly from a registry (no container runtime required)
 ```
+
+You can also update source location via the config file. See [Configuration](#configuration) for more details on setting this up.
+
+ Simply update the syft config with `default-image-pull-source` to specify which source to use by default.
 
 #### Default Cataloger Configuration by scan type
 
--- a/README.md
+++ b/README.md
@@ -436,6 +436,10 @@ file: ""
 # same as SYFT_CHECK_FOR_APP_UPDATE env var
 check-for-app-update: true
 
+# allows users to specify which image source should be used to generate the sbom
+# valid values are: registry, docker, podman
+image-pull-source: ""
+
 # a list of globs to exclude from scanning. same as --exclude ; for example:
 # exclude:
 #   - "/etc/**"
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ The above output includes only software that is visible in the container (i.e., 
 syft <image> --scope all-layers
 ```
 
-## Supported sources
+### Supported sources
 
 Syft can generate a SBOM from a variety of sources:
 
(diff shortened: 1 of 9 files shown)


spiffcs:
cc @luhring

luhring:
@spiffcs This looks amazing! Thank you so so much!!!

I just tried it out and it works great so far. 🤩 

Two cents:

1. It'd be nice to put a little section in the README about this functionality. It will make a lot of people happy, and it'll give me something to directly link to as I further evangelize Syft/Grype!
2. It also might be worth considering naming the config item with "default" in the name, so users know that it's not restricting down the set of sources that will eventually be tried, it's just choosing the first source to try.

spiffcs:
I'll get this cleaned up now with the readme section and behavior change - thanks for the quick feedback!

internal/config/application.go
```diff
@@ -291,3 +299,18 @@ func loadConfig(v *viper.Viper, configPath string) error {
 	}
 	return nil
 }
+
+func checkDefaultSourceValues(source string) error {
+	valid := map[string]struct{}{
```
wagoodman, line 304: nit: we're already using `strset.Set` elsewhere, that would be preferable

internal/config/application.go
```diff
@@ -291,3 +299,18 @@ func loadConfig(v *viper.Viper, configPath string) error {
...
+		"registry": {},
+		"docker":   {},
+		"podman":   {},
+		"":         {},
+	}
+
+	if _, ok := valid[source]; !ok {
+		return fmt.Errorf("%s is not a valid default source; please use one of ['registry', 'docker', podman', '']", source)
```
wagoodman, line 312: in case the list of valid values changes, the string here should be built from the same variable/source of truth to stay consistent.

syft/source/source.go
```diff
@@ -89,16 +89,28 @@ func ParseInputWithName(userInput string, platform string, detectAvailableImageS
...
+		UserInput:   userInput,
+		Scheme:      scheme,
+		ImageSource: source,
+		Location:    location,
+		Platform:    platform,
+		Name:        name,
 	}, nil
 }
 
+func defaultImageSourceFromConfig(defaultImageSource string) image.Source {
```
wagoodman, line 101: nit: this isn't really coming from a config at this layer. This feels more like `parseDefaultImageSource`.

README.md
```diff
@@ -141,6 +139,12 @@ file:path/to/yourproject/file            read directly from a path on disk (any
 registry:yourrepo/yourimage:tag          pull image directly from a registry (no container runtime required)
 ```
 
+If an image source is not provided and cannot be detected from the given reference it is assumed the image should be pulled from the Docker daemon.
+If docker is not present, then the Podman daemon is attempted next, followed by reaching out directly to the image registry last.
+This default behavior can be overridden with the default-image-pull-source configuration option (See Configuration for more details).
```
wagoodman, line 144:
```suggestion
This default behavior can be overridden with the `default-image-pull-source` configuration option (See [Configuration](https://github.com/anchore/syft#configuration) for more details).
```

README.md
```diff
@@ -141,6 +139,12 @@ file:path/to/yourproject/file            read directly from a path on disk (any
...
+
+Simply update the syft config with `default-image-pull-source` to specify which source to use by default.
```
wagoodman, line 146: this is now redundant given the above paragraph

wagoodman (approved):

luhring:
You all rock!!!

PR #1707, opened 2023-04-03
Refactor spdx tooling test to reduce intermittent failures

wagoodman:
We've seen several failures of the SPDX tooling test, but haven't been able to determine exactly why:

Screenshot 2023-04-03 at 1 33 24 PM

I have a couple educated guesses and made some targeted refactors, but have no conclusive evidence if this will fix the underlying issue (seems to only happen in CI every once in a while).

I've also made a couple of improvements:
- uses the existing test fixture facilities, as to leverage any preloaded CI cache without requiring a re build of the image.
- uses a temp dir per test Run() call instead of sharing directories.
- passes the explicit file for syft to write to and passes this file path to the verification tool

### DIFF
--- a/test/cli/spdx_tooling_validation_test.go
+++ b/test/cli/spdx_tooling_validation_test.go
@@ -10,82 +10,53 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/anchore/stereoscope/pkg/imagetest"
 )
 
 func TestSpdxValidationTooling(t *testing.T) {
+	img := imagetest.GetFixtureImage(t, "docker-archive", "image-java-spdx-tools")
+	require.NotEmpty(t, img.Metadata.Tags)
+	imgTag := img.Metadata.Tags[0]
+
+	images := []string{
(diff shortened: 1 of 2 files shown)


spiffcs (approved):

kzantow (approved):

test/cli/spdx_tooling_validation_test.go
```diff
@@ -10,82 +10,53 @@ import (
...
-				require.NoError(t, err)
-				fixturesPath := filepath.Join(cwd, "test-fixtures", "image-java-spdx-tools")
-				buildCmd := exec.Command("make", "build")
-				buildCmd.Dir = fixturesPath
-				buildCmd.Stdout = os.Stdout
-				buildCmd.Stderr = os.Stderr
-				err = buildCmd.Run()
-				require.NoError(t, err)
-			},
-			assertions: []traitAssertion{
```
wagoodman, line 44: also, I removed the use of trait assertions since this test really needs this to be 0, not just assert it. The triat assertions are meant for blackbox testing, not for prerequirements for validations (which should fail immediately if not met).

PR #1723, opened 2023-04-07
Add annotations for evidence on package locations

wagoodman:
Adds the concept of location annotations, allowing arbitrary key-value pairs to be listed onto `source.Location` objects. This PR also shows the first example of this with the dpkg cataloger:
```json
[
  {
    "id": "3e9282034226b93f",
    "name": "adduser",
    "version": "3.118",
    "type": "deb",
    "foundBy": "dpkgdb-cataloger",
    "locations": [
      {
        "path": "/usr/share/doc/adduser/copyright",
        "layerID": "sha256:ec09eb83ea031896df916feb3a61cefba9facf449c8a55d88667927538dca2b4",
        "annotations": {
          "evidence": "supporting"
        }
      },
      {
        "path": "/var/lib/dpkg/info/adduser.conffiles",
        "layerID": "sha256:ec09eb83ea031896df916feb3a61cefba9facf449c8a55d88667927538dca2b4",
        "annotations": {
          "evidence": "supporting"
        }
      },
      {
        "path": "/var/lib/dpkg/info/adduser.md5sums",
        "layerID": "sha256:ec09eb83ea031896df916feb3a61cefba9facf449c8a55d88667927538dca2b4",
        "annotations": {
          "evidence": "supporting"
        }
      },
      {
        "path": "/var/lib/dpkg/status",
        "layerID": "sha256:ec09eb83ea031896df916feb3a61cefba9facf449c8a55d88667927538dca2b4",
        "annotations": {
          "evidence": "primary"
        }
      }
    ],
    "licenses": [
      "GPL-2"
    ],
...
```

Notes:
- `source.LocationSet` requires that the existing `source.Location` is hashable as to enable it's use as a key in a map. This isn't possible since annotations are themselves maps and are directly attached to the Location struct. For this reason I've split out position information and metadata within the Location struct and added those as embeddings.
- The `source.LocationSet` merges metadata for any two locations where the position information is the same, essentially merging the locations. This seems like the right call relative to the package struct, but this does not account nicely for duplicate keys with differing values, which will log a warning.


wagoodman:
```diff
# ❯ diff schema/json/schema-7.1.1.json schema/json/schema-7.1.2.json
817,818c817,823
<         "virtualPath": {
<           "type": "string"
---
>         "annotations": {
>           "patternProperties": {
>             ".*": {
>               "type": "string"
>             }
>           },
>           "type": "object"
946c951
<             "$ref": "#/$defs/Coordinates"
---
>             "$ref": "#/$defs/Location"
```

syft/source/location_set.go
```diff
@@ -20,10 +23,20 @@ func NewLocationSet(locations ...Location) (s LocationSet) {
...
+		s.set = make(map[LocationData]struct{})
+		s.metadata = make(map[LocationData]LocationMetadata)
 	}
 	for _, l := range locations {
-		s.set[l] = struct{}{}
+		s.set[l.LocationData] = struct{}{}
+		if m, ok := s.metadata[l.LocationData]; ok {
+			err := m.merge(l.LocationMetadata)
+			if err != nil {
+				log.Warnf("partial merge of location metadata: %+v", err)
```
kzantow, line 34: Should this be Debug or Trace logging?

syft/source/location_set.go
```diff
@@ -4,10 +4,13 @@ import (
...
 
 	"github.com/mitchellh/hashstructure/v2"
+
+	"github.com/anchore/syft/internal/log"
 )
 
 type LocationSet struct {
-	set map[Location]struct{}
+	set      map[LocationData]struct{}
+	metadata map[LocationData]LocationMetadata
```
kzantow, line 13: Why not just modify this to be something like:
```go
type LocationSet struct {
	set map[Location]*Location
```
It seems a bit odd to specialize the metadata here, rather than finding the location from another equal location

kzantow (commented):
Overall 👍 just a few comments before ✅

syft/source/location.go
```diff
@@ -3,66 +3,135 @@ package source
...
+				errs = multierror.Append(errs, err)
+				continue
+			}
+		}
+		m.Annotations[k] = v
+	}
+	return errs
+}
+
+func (l Location) Annotate(key, value string) Location {
```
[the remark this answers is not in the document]
wagoodman: I should have explicitly called this out. I found instances where the mutation was surprising (such as annotating directly on a source.LocationReadCloser). For this reason I went for the "simpler" return-a-copy, which was only awkward to use in the `dpkg` cataloger, but natural everywhere else.

syft/source/location_set.go
wagoodman: probably, I'll adjust.

syft/source/location_set.go
wagoodman: I condensed it into a single map 👍

syft/source/location.go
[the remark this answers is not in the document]
wagoodman: changed to WithAnnotation (left as non-mutating)

syft/formats/syftjson/model/package.go
```diff
@@ -21,16 +21,16 @@ type Package struct {
...
-	Licenses  []string             `json:"licenses"`
-	Language  pkg.Language         `json:"language"`
-	CPEs      []string             `json:"cpes"`
-	PURL      string               `json:"purl"`
+	ID        string            `json:"id"`
+	Name      string            `json:"name"`
+	Version   string            `json:"version"`
+	Type      pkg.Type          `json:"type"`
+	FoundBy   string            `json:"foundBy"`
+	Locations []source.Location `json:"locations"`
```
spiffcs, line 29: Nice

spiffcs (approved):

spiffcs:
This one looks good to me - no other comments to add besides what already exists on the PR. I've started incorporating the location changes into the license evidence PR so 👍 thanks for making this clear and easy to follow

PR #1742, opened 2023-04-17
Fix kernel cataloger test fixtures

wagoodman:
This PR does a few things:
- pins the kernel test fixtures to a specific version that is not in the updates fedora channel (I needed to swap out the module under test since we couldn't pin the module being currently used)
- caches test images in CI between runs


spiffcs:
Waiting for the CI to run to double check the cache is working as we expected from pairing then 👍

spiffcs (approved):

PR #1748, opened 2023-04-18
fix: update cache.fingerprint file to java-builds dir

spiffcs:
### Summary

The `validations` workflow was looking under the `packages` directory for a fingerprint file that was not being created correctly. This caused tests to fail on PR that used the older cache after this update was made:

- https://github.com/anchore/syft/pull/1719/files#diff-1c5a22dae4b51bc7d60c3ec55409b0afe7ce708568c66a5d36854856d2b859e6

This PR syncs the `Makefile` for `java-build` fixtures with patterns that exist already in the codebase. It also links these changes into the main Makefile so that CI can correctly validate when the cache needs to be busted.

### DIFF
--- a/.github/workflows/validations.yaml
+++ b/.github/workflows/validations.yaml
@@ -36,8 +36,8 @@ jobs:
       - name: Restore Java test-fixture cache
         uses: actions/cache@v3
         with:
-          path: syft/pkg/cataloger/java/test-fixtures/java-builds/packages
-          key: ${{ runner.os }}-unit-java-cache-${{ hashFiles( 'syft/pkg/cataloger/java/test-fixtures/java-builds/packages.fingerprint' ) }}
+          path: syft/pkg/cataloger/java/test-fixtures/java-builds
+          key: ${{ runner.os }}-unit-java-cache-${{ hashFiles( 'syft/pkg/cataloger/java/test-fixtures/java-builds/cache.fingerprint' ) }}
 
       - name: Restore RPM test-fixture cache
         uses: actions/cache@v3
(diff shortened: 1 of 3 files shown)


kzantow (approved):

.github/workflows/validations.yaml
```diff
@@ -36,8 +36,8 @@ jobs:
       - name: Restore Java test-fixture cache
         uses: actions/cache@v3
         with:
-          path: syft/pkg/cataloger/java/test-fixtures/java-builds/packages
-          key: ${{ runner.os }}-unit-java-cache-${{ hashFiles( 'syft/pkg/cataloger/java/test-fixtures/java-builds/packages.fingerprint' ) }}
+          path: syft/pkg/cataloger/java/test-fixtures/java-builds
```
wagoodman, line 39: this cache would overwrite what's checked into git (all other locations under java-builds that is outside of packages)

syft/pkg/cataloger/java/test-fixtures/java-builds/Makefile
```diff
@@ -72,7 +72,7 @@ $(PKGSDIR)/gcc-amd64-darwin-exec-debug:
 	./build-example-macho-binary.sh $(PKGSDIR)
 
 # we need a way to determine if CI should bust the test cache based on the source material
-$(PKGSDIR).fingerprint: clean-examples
-	mkdir -p $(PKGSDIR)
-	find example-* build-* Makefile -type f -exec sha256sum {} \; | sort | tee /dev/stderr | tee $(PKGSDIR).fingerprint
-	sha256sum $(PKGSDIR).fingerprint
+.PHONY: cache.fingerprint
+cache.fingerprint:
```
wagoodman, line 76: I don't think this fundamentally changed anything -- that is, I think this PR now passes since the fingerprint has changed, but ultimately the cache won't be automatically busted when there is a source change.

.github/workflows/validations.yaml
spiffcs: ^ My fault - I thought this is what we wanted where we were checking for any changes under `java-builds` and rebuilding on changes to those files - thinking more segmenting out the packages folder is the right choice since it keeps the cache smaller, we probably just want to update the makefile to also key off of rebuilding if there are changes to any of the shell scripts `*.sh`

PR #1751, opened 2023-04-20
Migrate location-related structs to the file package

wagoodman:
Partially implements #558 by [migrating the following resources from the `source` package to the `file` package](https://github.com/anchore/syft/issues/558#issuecomment-1075716731):

- `Location`, `Locations`, `LocationSet`
- `Coordinates`, `CoordinateSet`
- `FileResolver`, `PathResolver`, `ContentResolver`, `WritableFileResolver`, `FileMetadataResolver`

In doing so, type aliases have been left behind with deprecation warnings, allowing existing lib users to port over to the new types/functions without introducing a breaking change yet (See [`syft/source/deprecated.go`](https://github.com/anchore/syft/blob/028c1a352ae951a6ba3ec0b4d0ae5a09c7066019/syft/source/deprecated.go)).

In order to do this, part of the plan in #558 needed to be executed, specifically to [split up file type definitions from capabilities](https://github.com/anchore/syft/issues/558#issuecomment-1075670017) (catalogers). This means introducing the same package organization structure that exists today with the package catalogers:
```
syft/
└── file/
    ├── cataloger/
    │   └── (collections of catalogers, organized in semantically-named packages)
    └── (basic definitions relating to files)
```

Note: this is a breaking change for any lib consumer that is directly importing the file catalogers (which is not recommended).

I've also updated the existing places in the codebase that were leveraging the newly deprecated `source.*` types/functions.

File resolver implementations have been moved to the `file/resolver` package and are all now exported for general use. This is under the same reasoning for splitting up definitions from capabilities, but in this case it's the first step in separating out source.Source definitions from general capabilities of the source object. Since no file resolvers have been exported in the past, this is not a breaking change.

Lastly, a sizable 1.7 MB binary test fixture was removed from the repo :tada: , but I had to remove an element from the tests that used it to do so.


wagoodman:
Comments from the team during the offline/sync review:
- do we need to export the resolver implementations? we should try not to
- do the resolver implementations need to be in file? no, we can keep this in source
- delete the hardlink test, leave the comment

spiffcs:
@wagoodman Finished and LGTM after sync review - I'll 🟢 when the comments/changes are pushed from our call

spiffcs (approved):

spiffcs (approved):

PR #1752, opened 2023-04-20
Improve pnpm support

### LINKED ISSUE #1535: Improve pnpm support
1. Transitive dependencies - in the [first example](https://github.com/anchore/syft/files/10576137/pnpm-lock.zip), `react` package is referenced and it has two transitive dependencies, that are not in the output (`loose-envify` => `js-tokens`)
2. Workspaces - Please add support for reading dependencies from [pnpm-lock.yaml](https://github.com/anchore/syft/files/10576068/pnpm-lock.zip) created in project that is using workspaces. Currently no deps are discovered for this file.

**Why is this needed**:
To improve support of pnpm.

**Additional context**:

Shanedell:
Improve pnpm support

- Parse packages section from pnpm-lock.yaml to get transitive dependencies.
- Should add support for getting dependencies from pnpm-lock.yaml files that use workspaces.
  - This is due to the fact all dependencies should get listed under the packages section.

Closes #1535

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
@@ -3,6 +3,7 @@ package javascript
 import (
 	"fmt"
 	"io"
+	"strings"
 
 	"gopkg.in/yaml.v3"
 
@@ -16,7 +17,8 @@ import (
 var _ generic.Parser = parsePnpmLock
 
 type pnpmLockYaml struct {
-	Dependencies map[string]string `json:"dependencies"`
+	Dependencies map[string]string      `json:"dependencies"`
+	Packages     map[string]interface{} `json:"packages"`
 }
 
 func parsePnpmLock(resolver source.FileResolver, _ *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/javascript/parse_pnpm_lock.go
```diff
@@ -36,6 +38,19 @@ func parsePnpmLock(resolver source.FileResolver, _ *generic.Environment, reader
 		pkgs = append(pkgs, newPnpmPackage(resolver, reader.Location, name, version))
 	}
 
+	// parse packages from packages section of pnpm-lock.yaml
+	for nameVersion := range lockFile.Packages {
+		nameVersionSplit := strings.Split(nameVersion, "/")
```
kzantow, line 43: maybe just `strings.TrimPrefix(nameVersion, "/")` and don't ignore the first element later? are there situations where there is something else at the beginning of these strings?

Shanedell: Oh yeah doing like this

```go
nameVersionSplit := strings.Split(strings.TrimPrefix(nameVersion, "/"), "/")
```

would probably be easier. I believe in the `packages:` section of the `pnpm-lock.yaml` every package will start with `/`.

Along with that, should I also do something like

```go
// last element in split array is version
version := nameVersionSplit[len(nameVersionSplit)-1]
nameVersionSplit = nameVersionSplit[:len(nameVersionSplit)-1] // remove version

// construct name from all array items
name := strings.Join(nameVersionSplit, "/")
```

instead of 

```go
// last element in split array is version
version := nameVersionSplit[len(nameVersionSplit)-1]

// construct name from all array items other than last item (version)
name := strings.Join(nameVersionSplit[:len(nameVersionSplit)-1], "/")
```

or which would be preferred?

kzantow (approved):
LGTM 👍

PR #1754, opened 2023-04-20
Add template func `hasField`, update README

jlehman9:
- Adds the template function hasField.
	- Allows template to check if a field is defined in a given JSON struct.
- Update README.md to document custom template functions.

This came from an issue I faced trying to write a CSV template function that grabbed vendor/author/publisher data. Unfortunately those fields aren't always defined for every package, and attempting to access them will cause Syft to exit with something like 
`* template: csv.tmpl:3:55: executing "csv.tmpl" at <.Metadata.Author>: can't evaluate field Author in type interface {}`

I figured it would also be good to document the where the functions are defined in the `README.md` so others can see they exist and offer new ones. 

Thanks to @kzantow for the help discovering this solution!

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,8 @@ Which would produce output like:
 
 Syft also includes a vast array of utility templating functions from [sprig](http://masterminds.github.io/sprig/) apart from the default Golang [text/template](https://pkg.go.dev/text/template#hdr-Functions) to allow users to customize the output format.
 
+Lastly, Syft has custom templating functions defined in `./syft/format/template/encoder.go` to help parse the passed-in JSON structs.
+
 ## Multiple outputs
 
 Syft can also output _multiple_ files in differing formats by appending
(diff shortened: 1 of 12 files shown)


kzantow:
Hey, thanks @jlehman9 ! would you be able to add a simple test that executes this function with a template?

jlehman9:
> Hey, thanks @jlehman9 ! would you be able to add a simple test that executes this function with a template?

I have no idea what I'm doing since I don't really know Go, but I _think_ I've got tests working that test the hasField function. Let me know if there is anything else I can do/fix/etc.

kzantow:
It looks like there are quite a few failing tests due to the changes, maybe just make a new test?

Also, you'll need to [sign-off your commits](https://github.com/anchore/syft/blob/main/CONTRIBUTING.md#sign-off-your-work).

jlehman9:
I believe I've undone my previous poor commit, created a separate test, re-signed the bad commit, and ensured all of the tests pass on my local machine. Hopefully all is good? 

Previously I had improperly assumed that the failing tests were all due to my local environment, mainly due to not having docker daemon installed since I usually run docker from minikube but the tests couldn't see that. Sorry about that!

kzantow:
Thanks for the contribution @jlehman9!

PR #1757, opened 2023-04-21
chore: go-rpmdb update

spiffcs:
## Summary 
- Bump go-rpmdb library to latest version

## Notes 
- this is not failing on the RPMDB tests because of the adapter import change, but I'm not sure if we still need to load the sqlite adapter into our `main.go` implementation for the rpmdb functionality to still work correctly
- I tested this against `almalinux` and found the correct packages for the RPMDB cataloger, but wanted to double check with the team before making this change


kzantow (approved):

spiffcs:
Thanks @kzantow! - I think this one is important enough too that we get @wagoodman also to give it some 👀 if there is something obvious I missed or if there is another dependency he might have been thinking about going with

wagoodman (approved):

wagoodman (changes requested):
Actually I need to check one more thing but don't have a laptop handy

westonsteimel:
Probably worth checking that grype works as expected when updated with this version of syft also just due to the SQL driver registration logic.  I remember there was something weird happening there and that's why we use our own sqlite fork in grype (to get around the registration in init logic)

test/integration/sqlite_rpmdb_test.go
```diff
@@ -3,6 +3,8 @@ package integration
 import (
 	"testing"
 
+	_ "modernc.org/sqlite"
```
wagoodman, line 6: this registers the sqlite driver only in test code, which wont get compiled into the final syft binary (which means the test will pass but syft will be unable to open sqlite RPMDBs since there will not be a registered driver)

wagoodman: I pushed code that keeps this import, but I've moved it to the main package, this way library users can choose their own sqlite driver (we shouldn't be registering drivers that have side effects on the stdlib within our lib)

wagoodman: I pushed another commit that keeps the import in main and this integration test (since this package does not use main it wouldn't have a sqlite driver).

PR #1759, opened 2023-04-25
Create python requirements metadata

Shanedell:
Create python requirements metadata

- Create new metadata struct and type for python requirements.
- Update parsing of python requirements to use python requirements metadata.
- Remove extras and url from line. Add them to metadata instead.
- Add unit test to test that extras are removed from package name.
- Update test to look at requirements metadata.
- Will need updated in future to support more than just == for the version constraint.
- Update JSON schema data

Closes anchore/grype#1246
Closes anchore/grype#1251

@kzantow I believe this is possibly what you referring to doing instead of anchore/grype#1251 

### Testing

Filename: `requirements.txt`
Content:

```txt
celery[redis, pytest]==4.4.7
starlette==0.17.1
```

Command: `go run cmd/syft/main.go requirements.txt`

**Output Before**

```bash
 ✔ Indexed requirements.txt 
 ✔ Cataloged packages      [2 packages]
NAME                   VERSION  TYPE   
celery[redis, pytest]  4.4.7    python  
starlette              0.17.1   python 
```

NOTE: when `celery[redis, pytest]` is returned to Grype it wouldn't scan this package

**Output After**

```bash
 ✔ Indexed requirements.txt 
 ✔ Cataloged packages      [2 packages]
NAME       VERSION  TYPE   
celery     4.4.7    python  
starlette  0.17.1   python 
```


syft/pkg/cataloger/python/parse_requirements.go
```diff
@@ -121,3 +122,12 @@ func removeEnvironmentMarkers(line string) string {
 
 	return parts[0]
 }
+
+// check if pip package has extras, if so remove it from package name
+func checkForExtras(packageName string) string {
```
kzantow, line 127: nit: this should probably be named something like `trimExtras`

kzantow (commented):
Thanks @Shanedell -- would you be able to add a test for this change?

kzantow:
Hey @Shanedell -- there's a suggestion from @wagoodman about adding metadata for the requirements.txt: https://github.com/anchore/syft/pull/1597#issuecomment-1499120540 -- would you be up for taking this PR a bit further to add this?

Shanedell:
Yeah I think I could get that added here

Shanedell:
@kzantow Should I basically add that PR to this one because I may need some the parsing?

Shanedell:
@kzantow @wagoodman The metadata for requirments.txt was added. The version constraint will need updated in the future when we support more that `==` for parsing, possibly inside of https://github.com/anchore/syft/pull/1597? Everything other than `URL` is currently being setup

wagoodman:
> Should I basically add that PR [1597] to this one because I may need some the parsing?
...
> If needed I can possibly add the changes from the previously mentioned PR to this one.

no, I don't think there is any feature overlap from 1597 with your PR, so we can treat your PR separately and get it merged first, then I can rebase 1597 afterwards.

> The version constraint will need updated in the future when we support more tha[n] == for parsing, possibly inside of https://github.com/anchore/syft/pull/1597 ?

Exactly -- you can ignore the extra constraint parsing in this PR that is being introduced in 1597. As long as this PR populates `versionConstraint` in the new metadata struct you are adding 👍 

Thanks for adding the metadata, its a really nice addition to the python catalogers 🙌

wagoodman:
Since you're adding a new metadata you'll need to update the JSON schema as well:

- bump the version, since this is an addition it would bump from 7.1.4 to 7.1.5 https://github.com/anchore/syft/blob/main/internal/constants.go#L9 (more details on the version scheme here https://github.com/anchore/syft/blob/main/schema/json/README.md)
- add the new metadata struct to this container struct https://github.com/anchore/syft/blob/main/schema/json/generate.go#L32
- run `make generate-json-schema`
- update snapshot test fixtures for the syft-json format with `go test ./syft/formats/syftjson --update-json`

shout out if you run into any trouble here!

Shanedell:
@wagoodman Did that hit everything or did I miss anything?

wagoodman:
JSON schema-wise, yup, it looks correct 🙌 

> Currently the URL isn't set, still trying to figure out best way to handle that one

Yeah the URL isn't super common, here is an example: (from https://stackoverflow.com/a/35998253)

```
# line in a requirements.txt file
package-two @ git+https://github.com/owner/repo@releases/tag/v3.7.1
```

It would be good to capture these lines as well, but we can always defer this addition if you'd like, you've already added a nice and complete feature.

Shanedell:
@wagoodman I did a bit of cleanup on the PR, mostly how some stuff was flowing. I also added the parsing of the URL, based it on splitting `@`, then the index of the first item starting with `git` was retrieved, then joined the splice from the index on using `@`. I updated the `trimRequirementsTxtLine` to remove the extras and url from the line. The parsing of the different metadata items are done through the original raw line, with comments removed.

wagoodman (approved):
looks great 💯  -- and thanks for taking on the extra suggested bit of the metadata, too!

I created https://github.com/anchore/syft/issues/1766 for a future enhancement. I didn't think it was right to try and add it now since it has some open questions.

PR #1772, opened 2023-04-29
Create codeql.yml

cyberoslab:


kzantow:
Thanks for the PR, but CodeQL is already run: https://github.com/anchore/syft/blob/main/.github/workflows/codeql-analysis.yml

PR #1779, opened 2023-05-04
feat: add scope to pom properties

jneate:
This lays the foundation for https://github.com/anchore/grype/issues/985


kzantow:
Thanks @jneate -- one thing, since you've modified the metadata, you'll need to bump the schema version (just a patch bump seems fine here) and regenerate the new schema. Info can be found here: https://github.com/anchore/syft/blob/main/schema/json/README.md. Essentially, It should just be update here: https://github.com/anchore/syft/blob/main/internal/constants.go#L9 and then run `cd schema/json && go run generate.go`

jneate:
Let me rebase, looks like someone bumped the schema before me

PR #1780, opened 2023-05-04
Update lists of catalogers

jsquyres:
Update the lists of catalogers to be current, and sort the lists for easier human readability, both in the README and in `syft/pkg/cataloger/cataloger.go`.


README.md
```diff
@@ -152,39 +152,53 @@ This default behavior can be overridden with the `default-image-pull-source` con
...
+- rpm-db
 - java
+- graalvm-native-image
+- apkdb
 - go-module-binary
 - dotnet-deps
+- portage
+- nix-store
+- sbom
+- binary
```
kzantow, line 169: It looks like this list isn't alphabetical -- would you mind re-sorting it, too?

kzantow: And maybe while you're at it resort the catalogers alphabetically in: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/cataloger.go ?

jsquyres: 🤦‍♂️ 

I sorted it, but apparently I goofed!  Let me fix and re-push.

kzantow (approved):
Thanks for the cleanup @jsquyres !

PR #1789, opened 2023-05-05
Update the CPE generation for spring-security-core

joshbressers:
This commit update the CPE generation to include vmware:spring_security in the CPE name which is what NVD uses for spring security vulnerabilities such as https://nvd.nist.gov/vuln/detail/CVE-2023-20862

### DIFF
--- a/syft/pkg/cataloger/common/cpe/generate_test.go
+++ b/syft/pkg/cataloger/common/cpe/generate_test.go
@@ -769,6 +769,14 @@ func TestCandidateProducts(t *testing.T) {
 			expected: []string{"spring_framework", "springsource_spring_framework" /* <-- known good names | default guess --> */, "springframework"},
 		},
 		{
+			name: "spring-security-core",
+			p: pkg.Package{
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/common/cpe/generate_test.go
```diff
@@ -768,6 +768,14 @@ func TestCandidateProducts(t *testing.T) {
...
 			expected: []string{"spring_framework", "springsource_spring_framework" /* <-- known good names | default guess --> */, "springframework"},
 		},
+		{
+			name: "spring-security-core",
+			p: pkg.Package{
+				Name: "spring-security-core",
+				Type: pkg.JavaPkg,
+			},
+			expected: []string{"spring-security-core", "spring_security", "spring_security_core"},
+		},
```
spiffcs, line 778: I saw in the above additional vendors is also using `vmware`. 

Do we want to add a case here under `TestCandidateVendor` to cover `spring-security-core` as well?

spiffcs (approved):

Issue #730, opened 2022-01-02
Support cataloging R packages


westonsteimel:
**What would you like to be added**:

Support for cataloging [R](https://www.r-project.org/) packages.  The package metadata spec is at https://r-pkgs.org/description.html#description and the package url spec is at https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#cran

**Why is this needed**:

Extends support for SBOM generation to the R ecosystem

westonsteimel:
As an example, this is what the `DESCRIPTION` file looks like for the [odbc](https://cran.r-project.org/package=odbc) package

```
Package: odbc
Title: Connect to ODBC Compatible Databases (using the DBI Interface)
Version: 1.3.3
Authors@R:
    c(person(given = "Jim",
             family = "Hester",
             role = "aut"),
      person(given = "Hadley",
             family = "Wickham",
             role = c("aut", "cre"),
             email = "hadley@rstudio.com"),
      person(given = "Oliver", "Gjoneski", role = "ctb", comment = "detule"),
      person(given = "lexicalunit",
             role = "cph",
             comment = "nanodbc library"),
      person(given = "Google Inc.",
             role = "cph",
             comment = "cctz library"),
      person(given = "RStudio",
             role = c("cph", "fnd")))
Description: A DBI-compatible interface to ODBC databases.
License: MIT + file LICENSE
URL: https://github.com/r-dbi/odbc, https://db.rstudio.com
BugReports: https://github.com/r-dbi/odbc/issues
Depends: R (>= 3.2.0)
Imports: bit64, blob (>= 1.2.0), DBI (>= 1.0.0), hms, methods, rlang,
        Rcpp (>= 0.12.11)
Suggests: covr, DBItest, magrittr, RSQLite, testthat, tibble
LinkingTo: Rcpp
ByteCompile: true
Encoding: UTF-8
RoxygenNote: 7.1.2
SystemRequirements: C++11, GNU make, An ODBC3 driver manager and
        drivers.
Collate: 'odbc.R' 'Driver.R' 'Connection.R' 'DataTypes.R'
        'RcppExports.R' 'Result.R' 'Table.R' 'Viewer.R' 'db.R'
        'hidden.R' 'utils.R' 'zzz.R'
NeedsCompilation: yes
Packaged: 2021-11-29 20:57:53 UTC; jhester
Author: Jim Hester [aut],
  Hadley Wickham [aut, cre],
  Oliver Gjoneski [ctb] (detule),
[...]

PR #1790, opened 2023-05-05
feat: Add R cataloger

willmurphyscode:
Add a cataloger that detects installed R packages by looking for DESCRIPTION files.

fixes: https://github.com/anchore/syft/issues/730


syft/pkg/cataloger/r/cataloger_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestRPackageCataloger(t *testing.T) {
+	expectedPkgs := []pkg.Package{
+		{
+			Name:         "base",
+			Version:      "4.3.0",
+			FoundBy:      "r-package-cataloger",
+			Locations:    source.NewLocationSet(source.NewLocation("base/DESCRIPTION")),
+			Licenses:     []string{"Part of R 4.3.0"},
+			Language:     "R",
+			Type:         "R-package",
```
wagoodman, line 21: nit: should probably be using the constant in `pkg` for this (same for `language` and `MetadataType`)

syft/pkg/cataloger/r/package.go
```diff
@@ -0,0 +1,38 @@
...
+		Locations:    locationSet,
+		Licenses:     []string{pd.License},
+		Language:     "R",
+		Type:         pkg.Rpkg,
+		PURL:         packageURL(pd),
+		MetadataType: pkg.RDescriptionFileMetadataType,
+		Metadata:     pd.RDescriptionFileMetadata,
+	}
+
+	result.Language = "R"
```
wagoodman, line 27: seems to be a dup from up above.

syft/pkg/cataloger/r/package.go
```diff
@@ -0,0 +1,38 @@
...
+	for _, loc := range locations {
+		locationSet.Add(loc.WithAnnotation(pkg.EvidenceAnnotationKey, pkg.PrimaryEvidenceAnnotation))
+	}
+	result := pkg.Package{
+		Name:         pd.Package,
+		Version:      pd.Version,
+		FoundBy:      catalogerName,
+		Locations:    locationSet,
+		Licenses:     []string{pd.License},
+		Language:     "R",
```
wagoodman, line 20: can you add a new language constant under `pkg/language.go` that represents this?

syft/pkg/cataloger/r/package.go
```diff
@@ -0,0 +1,38 @@
...
+
+func newPackage(pd parseData, locations ...source.Location) pkg.Package {
+	locationSet := source.NewLocationSet()
+	for _, loc := range locations {
+		locationSet.Add(loc.WithAnnotation(pkg.EvidenceAnnotationKey, pkg.PrimaryEvidenceAnnotation))
+	}
+	result := pkg.Package{
+		Name:         pd.Package,
+		Version:      pd.Version,
+		FoundBy:      catalogerName,
```
wagoodman, line 17: fun fact: the found-by field isn't set by the cataloger function itself, but instead by the caller. Specifically the generic cataloger object does this https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/generic/cataloger.go#L137 (which yours is using)

syft/pkg/cataloger/r/parse_description.go
```diff
@@ -0,0 +1,134 @@
...
+			NeedsCompilation: yesNoToBool(values["NeedsCompilation"]),
+			Author:           values["Author"],
+			Repository:       values["Repository"],
+			Built:            values["Built"],
+		},
+	}
+}
+
+func yesNoToBool(s string) bool {
+	return strings.EqualFold(s, "yes")
```
wagoodman, line 66: are there more cases to consider here? (e.g. `y`, `Yes`, `YES`, etc)

test/integration/catalog_packages_test.go
```diff
@@ -163,12 +163,14 @@ func TestPkgCoverageDirectory(t *testing.T) {
 	for _, l := range pkg.AllLanguages {
 		definedLanguages.Add(l.String())
 	}
+	definedLanguages.Remove(string(pkg.R))
```
wagoodman, line 166: nit: for consistency could you move this down below where other removals are being done? (or move those up here, either one... then they would all be done in one place which is easier to understand)

syft/pkg/cataloger/r/parse_description.go
willmurphyscode: That's a good question. Let me make a Docker image with a lot of R packages installed and take a look.

syft/pkg/cataloger/r/parse_description.go
```diff
@@ -0,0 +1,134 @@
...
+License: LGPL (>= 2.1)
+License: MIT + file LICENSE
+License: Part of R 4.3.0
+License: Unlimited
+*/
+
+func parseDescriptionFile(_ source.FileResolver, _ *generic.Environment, reader source.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	values := extractFieldsFromDescriptionFile(reader)
+	m := parseDataFromDescriptionMap(values)
+	return []pkg.Package{newPackage(m, []source.Location{reader.Location}...)}, nil, nil
```
wagoodman, line 34: we should probably add a check just before this line that makes certain that minimum values were found before creating a package for it (such as `pkg.Package.Name` and `pkg.Package.Version`). Since the glob for this cataloger is fairly broad this would somewhat defend against unintended overlap with other ecosystems that have `DESCRIPTION` files.

syft/pkg/cataloger/r/parse_description.go
willmurphyscode: Looking for the answer, I found https://github.com/r-lib/desc/blob/c8463062edf4dc5501c66d4e8f6019743fae3fa3/R/constants.R#L11-L111 (I believe this is the library that is often used to write these files.) Might be a good reference if we want to improve some validations.

I'll see if I can find where that library writes `NeedsCompilation` and what values it can include.

syft/pkg/cataloger/r/package.go
willmurphyscode: Should I remove this then?

willmurphyscode: Will remove.

wagoodman (approved):
nice addition! 🎉

wagoodman (changes requested):
I realized just after hitting approve that a JSON schema addition needs to be made

wagoodman:
To make a JSON schema update you can add your new metadata struct to https://github.com/anchore/syft/blob/main/schema/json/generate.go#L32 and follow the steps in https://github.com/anchore/syft/blob/main/schema/json/README.md#json-schema on how to bump .

PR #1794, opened 2023-05-08
Keep original FileInfo persisted on file.Metadata structs

wagoodman:
Incorporates changes from anchore/stereoscope#172 by swapping over to `file.ManualInfo` in test fixtures and SBOM decoding execution paths as necessary.


syft/file/metadata_cataloger_test.go
```diff
@@ -146,13 +165,14 @@ func TestFileMetadataCataloger(t *testing.T) {
 			l := source.NewLocationFromImage(test.path, *ref.Reference, img)
 
 			if _, ok := actual[l.Coordinates]; ok {
-				redact := actual[l.Coordinates]
-				redact.ModTime = time.Time{}
-				actual[l.Coordinates] = redact
+				// we're not interested in keeping the test fixtures up to date with the latest file modification times
+				// thus ModTime is not under test
+				fi := test.expected.FileInfo.(file.ManualInfo)
+				fi.ModTimeValue = actual[l.Coordinates].ModTime()
```
kzantow, line 171: Nit: maybe update the actual value here with the hardcoded one?

spiffcs (approved):
👍

kzantow (approved):
👍

syft/file/metadata_cataloger_test.go
wagoodman: I don't think I follow -- the comment above explicitly mentions this field isn't intended to be put under test. Are you saying the test should cover this?

PR #1797, opened 2023-05-08
Reduce log spam on unknown relationship type

willmurphyscode:
An unknown relationship type doesn't mean things are incorrect; it probably means that the version of syft that generated the SBOM is ahead of the version of syft that's parsing it.

Addresses https://github.com/anchore/grype/issues/1244, but won't fix it until the version of syft that's imported by grype is updated, at https://github.com/anchore/grype/blob/75e7ef43cd03baba4c76af3751274c7aa7452f36/go.mod#L56.

### DIFF
--- a/syft/formats/syftjson/to_syft_model.go
+++ b/syft/formats/syftjson/to_syft_model.go
@@ -194,7 +194,7 @@ func toSyftRelationship(idMap map[string]interface{}, relationship model.Relatio
 	case artifact.OwnershipByFileOverlapRelationship, artifact.ContainsRelationship, artifact.DependencyOfRelationship, artifact.EvidentByRelationship:
 	default:
 		if !strings.Contains(string(typ), "dependency-of") {
-			log.Warnf("unknown relationship type: %s", typ)
+			log.Tracef("unknown relationship type: %s", typ)
 			return nil
 		}
 		// lets try to stay as compatible as possible with similar relationship types without dropping the relationship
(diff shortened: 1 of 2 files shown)


kzantow (approved):

willmurphyscode:
Discussed this with @kzantow and @wagoodman, and it might be better to still warn about this, but warn only once, rather than once per lost relationship. I will send a revision.

syft/formats/syftjson/to_syft_model.go
```diff
@@ -35,10 +36,30 @@ func toSyftModel(doc model.Document) (*sbom.SBOM, error) {
...
+}
+
+func deduplicateErrors(errors []error) []string {
+	errorCounts := make(map[string]int)
+	var errorMessages []string
+	for _, e := range errors {
+		errorCounts[e.Error()] = errorCounts[e.Error()] + 1
+	}
+	for msg, count := range errorCounts {
+		errorMessages = append(errorMessages, fmt.Sprintf("%q occurred %d time(s)", msg, count))
```
kzantow, line 58: suggestion:
```go
if count > 1 {
  errorMessages = append(errorMessages, fmt.Sprintf("%s (occurred %d times)", msg, count))
} else {
  errorMessages = append(errorMessages, msg)
}
```

kzantow (approved):
🎉 nicely done! -- just a small formatting suggestion

wagoodman (approved):
looks great 👍

syft/formats/syftjson/to_syft_model.go
willmurphyscode: Thanks for the suggestion - I think I'll leave it as is for now, since I like that the whole block of de-duplicated errors in the logs all look the same, but I'd be open to changing this in a follow-up if it seems important.

PR #1798, opened 2023-05-08
1577 license set

spiffcs:


syft/pkg/license_set.go
```diff
@@ -21,77 +22,43 @@ func NewLicenseSet(licenses ...License) (s LicenseSet) {
 	return s
 }
 
-func (s *LicenseSet) get(license License) (uint64, int, *License, error) {
-	id, err := license.Hash()
+func (s *LicenseSet) get(license License) (id artifact.ID, merged bool, err error) {
```
wagoodman, line 25: This is no longer just fetching an item, it's actively merging it. A name change is warranted. 
```suggestion
func (s *LicenseSet) addToExisting(license License) (id artifact.ID, merged bool, err error) {
```

syft/pkg/license_set.go
```diff
@@ -21,77 +22,43 @@ func NewLicenseSet(licenses ...License) (s LicenseSet) {
...
-	if s.set == nil {
-		return
-	}
-	for _, l := range licenses {
-		id, idx, v, err := s.get(l)
-		if err != nil {
-			log.Debugf("license set failed to remove license %#v: %+v", l, err)
+			log.Trace("license set failed to add license %#v: %+v", l, err)
+		} else {
+			log.Trace("merged licenses")
```
wagoodman, line 59: no need to log this event

syft/pkg/license_set_test.go
```diff
@@ -5,6 +5,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"github.com/anchore/syft/internal"
+	"github.com/anchore/syft/syft/license"
 	"github.com/anchore/syft/syft/source"
 )
```
wagoodman, line 12: need a case under test for concluded type (or a difference in type) as well as how that interplays with url/location/value/expression.

syft/pkg/license.go
```diff
@@ -68,11 +70,15 @@ func NewLicense(value string) License {
 		Value:          value,
```
wagoodman, line 70: should there be empty checks for this value so we don't create licenses with "" as a value?

syft/pkg/package.go
```diff
@@ -65,6 +65,7 @@ func (p *Package) merge(other Package) error {
 	}
 
 	p.Locations.Add(other.Locations.ToSlice()...)
+	p.Licenses.Add(other.Licenses.ToSlice()...)
```
wagoodman, line 68: is there a test case that covers this change?

syft/pkg/cataloger/internal/pkgtest/test_generic_parser.go
```diff
@@ -139,6 +161,26 @@ func (p *CatalogTester) IgnoreLocationLayer() *CatalogTester {
 	p.locationComparer = func(x, y source.Location) bool {
 		return cmp.Equal(x.Coordinates.RealPath, y.Coordinates.RealPath) && cmp.Equal(x.VirtualPath, y.VirtualPath)
 	}
+
+	// we need to update the license comparer to use the ignored location layer
+	p.licenseComparer = func(x, y pkg.License) bool {
+		return cmp.Equal(x, y, cmp.Comparer(p.locationComparer), cmp.Comparer(
+			func(x, y source.LocationSet) bool {
```
wagoodman, line 168: should probably break out this function into `DefaultLocationSetComparer`

wagoodman (approved):
I have a few comments that should get addressed before the parent PR is merged, but otherwise looks great! 🎉

Issue #1569, opened 2023-02-14
The actual error is overridden and not returned from getImageWithRetryStrategy


akpsgit:
**Please provide a set of steps on how to reproduce the issue**
Try to scan a non existing image, e.g.:
`syft registry:docker.io/istio/examples-bookinfo-reviews-v1@sha256:b8f16a765eea1e53aadf59146b3ffede9d2a869d6edf7d1729fb83d5146bbef3`

**What happened**:
The actual error is being printed only as warning (requires the -v flag, otherwise now way of knowing about it):
```
syft registry:docker.io/istio/examples-bookinfo-reviews-v1@sha256:b8f16a765eea1e53aadf59146b3ffede9d2a869d6edf7d1729fb83d5146bbef3 -v
[0000]  INFO syft version: 0.71.0
[0002]  WARN scheme "registry" specified, but it coincides with a common image name; re-examining user input "registry:docker.io/istio/examples-bookinfo-reviews-v1@sha256:b8f16a765eea1e53aadf59146b3ffede9d2a869d6edf7d1729fb83d5146bbef3" without scheme parsing because image retrieval using scheme parsing was unsuccessful: unable to use OciRegistry source: failed to get image descriptor from registry: GET https://index.docker.io/v2/istio/examples-bookinfo-reviews-v1/manifests/sha256:b8f16a765eea1e53aadf59146b3ffede9d2a869d6edf7d1729fb83d5146bbef3: MANIFEST_UNKNOWN: manifest unknown; unknown manifest name=istio/examples-bookinfo-reviews-v1 revision=sha256:b8f16a765eea1e53aadf59146b3ffede9d2a869d6edf7d1729fb83d5146bbef3
2023/02/14 15:10:12 error during command execution: 1 error occurred:
	* failed to construct source from user input "registry:docker.io/istio/examples-bookinfo-reviews-v1@sha256:b8f16a765eea1e53aadf59146b3ffede9d2a869d6edf7d1729fb83d5146bbef3": could not fetch image "docker.io/istio/examples-bookinfo-reviews-v1@sha256:b8f16a765eea1e53aadf59146b3ffede9d2a869d6edf7d1729fb83d5146bbef3": unable determine image source
```

**What you expected to happen**:
That actual error will be returned as part of the error:
```
[...]

Issue #1695, opened 2023-03-24
Default image source does not validate url when interpreting scheme as image


xtreme-conor-nosal:
**What happened**:
Syft has a check to retry fetching an image if the docker: or registry: scheme was provided, in case of possible ambiguity and the user wants to scan the `docker` or `registry` image from dockerhub (or the environment's default search registry). 

If the first interpretation was correct, trying to reinterpret the scheme as image can lead to fetch attempts that can't succeed and misleading log messages.

e.g.
```
failed to get image descriptor from registry: GET http://myhost:5000/v2/test/app/manifests/sha256:digest: UNAUTHORIZED:
"https://registry:myhost:5000/v2/": dial tcp: lookup registry:myhost: no such host
```

**What you expected to happen**:
Syft should validate if interpreting the scheme as image results in a valid image.
e.g. parsing `https://registry:myhost:5000/test/app:sha` is not a valid image url, so the potential ambiguity can be ruled out, and a second fetch does not need to be attempted. 

**Steps to reproduce the issue**:
Run syft with a `registry:` scheme specified and an image url that will fail.

tgerla:
Hi @xtreme-conor-nosal, if we're reading this right, "registry:" needs to be the very first part of the image source. If you put it after the https:// it will indeed be interpreted as a URL. Can you try this scheme instead?

```
syft registry:myhost:5000/test/app:sha
```

Can you share the exact test command you are using? We might be misunderstanding the problem here.

xtreme-conor-nosal:
```
syft registry:index.docker.io:443/nginx2:latest # deliberate typo


[0000]  WARN scheme "registry" specified, but it coincides with a common image name; re-examining user input "registry:index.docker.io:443/nginx2:latest" without scheme parsing because image retrieval using scheme parsing was unsuccessful: unable to use OciRegistry source: failed to get image descriptor from registry: GET https://index.docker.io:443/v2/nginx2/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:nginx2 Type:repository]]
[0000] ERROR Failed to parse known_hosts:  from-lib=stereoscope
2023/04/06 18:56:33 error during command execution: 1 error occurred:
[...]

PR #1801, opened 2023-05-09
Return both failures when failed to retrive an image with a scheme

FrimIdan:
This PR is meant to solve an issue we've got when syft failed to print the "right" error upon fetch failure.

For example, i'm trying to scan an arm64 image with an amd64 syft CLI and i'm getting the following error:
```
$ syft docker:idanfrim/multi-arch:only-arm64
 ⠸ Pulling image

2023/05/09 10:29:44 error during command execution: 1 error occurred:
	* failed to construct source from user input "docker:idanfrim/multi-arch:only-arm64": could not fetch image "idanfrim/multi-arch:only-arm64": unable to determine image source to select platform
```

The error is not clear, `unable to determine image source` is wrong since i've provided `docker` scheme.
When running the same command with debug I can see the correct failure.
```
$ syft -v docker:idanfrim/multi-arch:only-arm64
[0000]  INFO syft version: 0.77.0
[0003]  INFO new version of syft is available: 0.80.0 (current version is 0.77.0)
[0006]  WARN scheme "docker" specified, but it coincides with a common image name; re-examining user input "docker:idanfrim/multi-arch:only-arm64" without scheme parsing because image retrieval using scheme parsing was unsuccessful: unable to use DockerDaemon source: image has unexpected architecture "arm64", which differs from the user specified architecture "amd64"
2023/05/09 12:04:11 error during command execution: 1 error occurred:
	* failed to construct source from user input "docker:idanfrim/multi-arch:only-arm64": could not fetch image "idanfrim/multi-arch:only-arm64": unable to determine image source to select platform
```

So in this PR I've returned both errors so it will be more clear:
```
$ ./snapshot/darwin-build_darwin_amd64_v1/syft docker:idanfrim/multi-arch:only-arm64
 ⠙ Pulling image

2023/05/09 10:29:09 error during command execution: 1 error occurred:
	* failed to construct source from user input "docker:idanfrim/multi-arch:only-arm64": could not fetch image "idanfrim/multi-arch:only-arm64": scheme "docker" specified; image retrieval using scheme parsing (idanfrim/multi-arch:only-arm64) was unsuccessful: unable to use DockerDaemon source: image has unexpected architecture "arm64", which differs from the user specified architecture "amd64"; image retrieval without scheme parsing (docker:idanfrim/multi-arch:only-arm64) was unsuccessful: unable to determine image source to select platform
```

Solves https://github.com/anchore/syft/issues/1569

### DIFF
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -216,13 +216,27 @@ func getImageWithRetryStrategy(in Input, registryOptions *image.RegistryOptions)
 	// We need to determine the image source again, such that this determination
 	// doesn't take scheme parsing into account.
 	in.ImageSource = image.DetermineDefaultImagePullSource(in.UserInput)
-	img, err = stereoscope.GetImageFromSource(ctx, in.UserInput, in.ImageSource, opts...)
+	img, userInputErr := stereoscope.GetImageFromSource(ctx, in.UserInput, in.ImageSource, opts...)
 	cleanup = func() {
 		if err := img.Cleanup(); err != nil {
 			log.Warnf("unable to cleanup image=%q: %w", in.UserInput, err)
 		}
 	}
-	return img, cleanup, err
+	if userInputErr != nil {
+		// Image retrieval failed on both tries, we will want to return both errors.
+		return nil, nil, fmt.Errorf(
+			"scheme %q specified; "+
+				"image retrieval using scheme parsing (%s) was unsuccessful: %v; "+
+				"image retrieval without scheme parsing (%s) was unsuccessful: %v",
+			scheme,
+			in.Location,
+			err,
+			in.UserInput,
+			userInputErr,
+		)
+	}
+
+	return img, cleanup, nil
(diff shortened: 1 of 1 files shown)


FrimIdan:
@kzantow @tgerla Do you mind to take a look at the PR?

tgerla:
Hi @FrimIdan, thank you for the pull request! We've been looking at similar issues recently, and we might want to move some of this processing into stereoscope itself. We will keep this ticket updated as we investigate. Thanks again.

tgerla:
I believe this issue is also related: https://github.com/anchore/syft/issues/1695

FrimIdan:
@tgerla ok. I hope it can be sorted quickly, the current situation is confusing and my pr might be a temp solution until a better will be integrated.

spiffcs (approved):
@anchore/tools This is approved - waiting on a 2nd before merge!

PR #1810, opened 2023-05-10
feat: warn if parsing newer SBOM

willmurphyscode:
If syft is asked to parse an SBOM that was written by a newer version of syft, emit a warning.

Suggested during https://github.com/anchore/grype/issues/1244 - it's likely more helpful to warn that syft is being asked to parse an SBOM written by newer syft than it is to warn every single instance of a new type.

Note that this pulls in https://github.com/Masterminds/semver. It could have used https://github.com/anchore/go-version, but the version structs here expose Major, Minor, and Patch directly, which I thought I would need. Also, it might be good to get off https://github.com/anchore/go-version, which is an old fork, so I didn't want to use it in a new place.


syft/formats/syftjson/decoder.go
```diff
@@ -18,5 +22,26 @@ func decoder(reader io.Reader) (*sbom.SBOM, error) {
...
+		log.Warn(err)
+	}
+
 	return toSyftModel(doc)
 }
+
+func outOfDateParser(documentVerion string, parserVersion string) error {
+	documentV, err := semver.NewVersion(documentVerion)
+	if err != nil {
+		return err
```
willmurphyscode, line 35: Maybe it's worth wrapping this err? It might not make sense if it gets logged directly.

wagoodman: wrapping would be great to give context to what version couldn't be parsed (where the error would hopefully fill in why)

syft/formats/syftjson/decoder.go
```diff
@@ -18,5 +22,26 @@ func decoder(reader io.Reader) (*sbom.SBOM, error) {
...
 	}
 
+	if err := outOfDateParser(doc.Schema.Version, internal.JSONSchemaVersion); err != nil {
+		log.Warn(err)
+	}
+
 	return toSyftModel(doc)
 }
+
+func outOfDateParser(documentVerion string, parserVersion string) error {
```
wagoodman, line 32: nit: naming-wise this is really asserting/checking/validating the date, so something like `checkSupportedSchema` or similar seems a little better.

wagoodman (approved):

Issue #1815, opened 2023-05-15
CycloneDX Dependencies Relationships Inverted


shanealv:
**What happened**:

(previously reported in anchore/grype#1294)

When running `syft convert` on a CycloneDX SBOM with a "dependencies"  section and outputting back to a CycloneDX, the dependency relationships appear to be inverted.

So this SBOM:

<details><summary>
cat bom.json
</summary>

```json
{
  "$schema": "http://cyclonedx.org/schema/bom-1.4.schema.json",
  "bomFormat": "CycloneDX",
  "specVersion": "1.4",
  "version": 1,
  "components": [
    {
      "bom-ref": "pkg:golang/github.com/inconshreveable/mousetrap@v1.1.0?type=module",
      "type": "library",
      "name": "github.com/inconshreveable/mousetrap",
      "version": "v1.1.0",
      "purl": "pkg:golang/github.com/inconshreveable/mousetrap@v1.1.0?type=module\u0026goos=linux\u0026goarch=amd64"
    },
    {
      "bom-ref": "pkg:golang/github.com/spf13/cobra@v1.7.0?type=module",
      "type": "library",
      "name": "github.com/spf13/cobra",
      "version": "v1.7.0",
      "purl": "pkg:golang/github.com/spf13/cobra@v1.7.0?type=module\u0026goos=linux\u0026goarch=amd64"
    },
    {
      "bom-ref": "pkg:golang/github.com/spf13/pflag@v1.0.5?type=module",
      "type": "library",
      "name": "github.com/spf13/pflag",
      "version": "v1.0.5",
      "purl": "pkg:golang/github.com/spf13/pflag@v1.0.5?type=module\u0026goos=linux\u0026goarch=amd64"
    }
  ],
  "dependencies": [
    {
      "ref": "pkg:golang/github.com/inconshreveable/mousetrap@v1.1.0?type=module"
    },
    {
      "ref": "pkg:golang/github.com/spf13/cobra@v1.7.0?type=module",
      "dependsOn": [
        "pkg:golang/github.com/inconshreveable/mousetrap@v1.1.0?type=module",
        "pkg:golang/github.com/spf13/pflag@v1.0.5?type=module"
      ]
    },
    {
      "ref": "pkg:golang/github.com/spf13/pflag@v1.0.5?type=module"
    }
  ]
}
```

</details>

Became this:

<details><summary>syft convert bom.json -o cyclonedx-json</summary>

```json
{
[...]

PR #1816, opened 2023-05-15
fix: cyclonedx depends-on relationship inverted

shanealv:
When reading CycloneDX boms, the relationships defined in the "dependencies" sections are translated into "artifact.Relationship" objects. Previously, relationships were inverted by this process. That is, a CycloneDx dependency "a depends on b" was being read as "a dependency of b". This caused downstream processing to write out the dependency relationship incorrectly. This commit fixes the issue by inverting the relationship during decoding and adding a regression test.

Fixes anchore/syft#1815


kzantow (approved):
Good catch @shanealv !

kzantow:
@shanealv I went ahead and ran `make lint-fix` to fix the imports/static analysis error and pushed. Apologies if that's an issue, happy to back this out if you want.

PR #1820, opened 2023-05-16
fix: update field plurality of 8.0.0 schema before release

spiffcs:
To keep things consistent across the schema we want `Locations` and `URLs` to be plural fields now that they are fields on the `License` struct


kzantow (approved):
👍

PR #1830, opened 2023-05-19
source: plumb sym link base through to syft CLI

jsquyres:
Continuing the work of b81c980, change the behavior of the syft command: the root for symbol link resolution of directory scans is now the directory root, not "/".

Do this pedantically, by separating the directory root to the scanned from the directory sym link base.  As of this commit, these two values will always be the same, but this strategy leaves the door open for them to be different someday, if that would ever be useful.

Also remove some unused functions and consolidate down to just the NewFromDirectoryBaseWithName() function.

Refs #1485, #1359

### DIFF
--- a/syft/source/source.go
+++ b/syft/source/source.go
@@ -46,6 +46,7 @@ type Input struct {
 	Location    string
 	Platform    string
 	Name        string
+	SymLinkBase string
 }
 
 // ParseInput generates a source Input that can be used as an argument to generate a new source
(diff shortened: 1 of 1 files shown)


jsquyres:
Looks like some of those "unused" functions I removed are used by tests.  I'll investigate...

kzantow:
Hi @jsquyres and thanks for the contribution here -- we have a bit of a different change that does this in PR https://github.com/anchore/syft/pull/1867. It should be noted the referenced PR has a prerequisite of some _other_ changes and we're looking at getting these done hopefully in a week-ish timeframe. That said, I'm going to close this PR due to the overlap. Please let me know if I've misunderstood something and would be happy to reopen it!

jsquyres:
@kzantow Sorry for my long silence on this -- I got distracted by vacation, holidays, and other shiny objects.  If #1867 implements effectively the same kind of functionality as I did in this PR, we can certainly close this PR.  I'd just like the functionality, regardless of whether it's my PR or someone else's. 😄

kzantow:
Thanks @jsquyres! I'm going to close this as noted -- we're making progress

Issue #1776, opened 2023-05-03
Why are different catalogers disabled (by default) in different scenarios?


jsquyres:
This is a (perhaps naïve) question, not a feature request.

We discovered what was admittedly documented (but we had missed this part in the docs): that container images and filesystem scans have a different default set of catalogers enabled.  It's an easy-enough issue to resolve: we can use the `--cataloger` CLI option to get the behavior that we want.

But we're curious: *why* is there a difference in the default set of catalogers between these two scenarios?  As a concrete example: the RPM file cataloger is disabled by default for images, but enabled by default for directory scans.

Granted, some catalogers are obviously not relevant in some scenarios.  But is there harm -- e.g., performance degradation -- in having them enabled for all scenarios?

Thanks for any enlightenment you can provide!

jsquyres:
Bump.  Can anyone share some knowledge here? Thanks!

kzantow:
@jsquyres apologies for the delay getting back to you. The reason there are different catalogers enabled is that there are some expectations about what we would find in a source scan vs. an image scan. An example is: during a source scan we'll find a `package-lock.json` and _also_ a lot of `package.json` files in the `node_modules` directory. In order to try to avoid adding a bunch of packages that don't make a lot of sense, we have tried to organize the catalogers in such a way that things unexpected to find in directory scans or image scans are not included _by default_.

That said, we are also actively pursing a couple of improvements. The first of which is adding a "tagging" mechanism (part of PR https://github.com/anchore/syft/pull/1383). The second of which is adding some functionality to the `--catalogers` flag/configuration that allows prefixing a cataloger name with `+` or `-` to include, or exclude catalogers respectively. Would this solve the problem(s) for you?

jsquyres:
> An example is: during a source scan we'll find a package-lock.json and also a lot of package.json files in the node_modules directory.

I hear what you're saying, but let me ask my question in a slightly different way: just because you don't expect to find things via specific catalogers, is there a reason to disable them (by default)?  E.g., is wall-clock execution time a concern?  My assumption is that you would want to find everything that is in the filesystem / image, especially the things that you do not expect to be there.  Is that a naïve perspective?

To be clear: I'm not debating the value of having a robust cataloger selection mechanism for those who want/need to have a specific set of catalogers.  Having such functionality seems to be an obvious Good Thing.

jsquyres:
@kzantow Ping.

tgerla:
[...]

PR #1831, opened 2023-05-19
Minor cataloger improvements for the CLI

jsquyres:
1. Add `ALL` (synonym for `all`), `ALL:IMAGE`, and `ALL:DIRECTORY` catalog selectors (somewhat in response to #1776)
1. Add some more INFO and DEBUG log messages about which catalogers were selected.
1. Add a sanity check: if 0 catalogers are selected, emit an error.
    * Also make the Syft CLI fail faster if 0 catalogers are selected

See the individual commits and their corresponding commit messages for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,31 @@ If docker is not present, then the Podman daemon is attempted next, followed by 
 
 This default behavior can be overridden with the `default-image-pull-source` configuration option (See [Configuration](https://github.com/anchore/syft#configuration) for more details).
 
-### Default Cataloger Configuration by scan type
+### Cataloger Configuration
 
-##### Image Scanning:
+By default, a set of catalogers are chosen depending on the type of
+source that is being scanned.  However, the `--catalogers LIST` CLI
+option can be specified to override this default selection, where
+`LIST` can be one of the following:
+
+- A comma-delimited list of cataloger names (run `syft` with the `-v`
+  or `-vv` option to see the list of catalogers available).
+
+  **Note:** Syft does not currently warn if a non-existent cataloger
+  name is used in the list.
+- The keyword `ALL`, indicating that Syft should use all available
+  catalogers.
+- The keyword `ALL:IMAGE`, indicating that Syft should use the default
+  set of catalogers when container images are scanned (see below).
+- The keyword `ALL:DIRECTORY`, indicating that Syft should use the
+  default set of catalogers when directory trees are scanned (see
+  below).
+
+#### Image Scanning
+
+The following catalogers are selected by default (i.e., if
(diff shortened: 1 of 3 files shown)


kzantow:
Hi @jsquyres thank you for this PR -- I'd like to note that there is another upcoming change (which is taking a while to get to completion) that offers some similar, but possibly better options for enabling certain sets of catalogers: https://github.com/anchore/syft/pull/1383 -- would this work for your use case?

jsquyres:
@kzantow Interesting.  I'm not deep enough into the zen of Syft to fully grok how #1383 will result in what you're saying, but you seemed to make reference to the same capability in https://github.com/anchore/syft/issues/1776#issuecomment-1555011570.

Our use case is that we have entirely scanned images up to this point, but now we need to scan a filesystem.  We were a little surprised when scanning an image and a corresponding equivalent filesystem turned up different results (which led to my question in #1776).  Reading the docs and digging in to the source code led to better understanding and resulted in this PR.  To be clear: by examining the source of Syft v0.79.0 and using the `--catalogs` CLI option, we were able to get the results to be the same.  We can certainly keep doing that, but it's a little tedious to have to keep examining the source code or closely examine all the output from Syft to see exactly which catalogers were used in a given run.

* I see that #1383 has had no public code activity in ~6 months.  Is there an estimated timeframe when it will become available in a release?
* #1383 is a big PR, and I'm still a relative newbie to the Syft code base.  It doesn't look like there are CLI updates corresponding to the API code changes in that PR.  If that is correct, is there any other ongoing work to include CLI corresponding updates?

jsquyres:
@kzantow Ping.

wagoodman:
@jsquyres though #1383 is stale, it's quite important to the syft 1.0 work, which I think will be landing within the next 3-ish months. The feature isn't completed in that PR yet, but the API will be accounting for this. There is a follow up task in the PR description to consider https://github.com/anchore/syft/issues/1039 and https://github.com/anchore/syft/issues/1731 . #1383 has implemented a tag-based cataloger selection approach in the API, the next step is to wire it up to the `--cataloger` CLI option.

jsquyres:
@wagoodman @kzantow Using a tag-based cataloger selection sounds fine to me.  If I remove the `ALL*` stuff from this PR, would the other improvements be helpful?  (e.g., the docs changes, logging messages, and fail fast)

kzantow:
Thanks very much for the contribution @jsquyres. It's really great that you've been improving Syft, but the unfortunate timing of these PRs just so happened to overlap with a couple other analogous but different changes. I don't think there's a lot from this PR that we will be able to use, but we _really_ appreciate the contribution and would encourage you to reach out if there are other things you'd like to work on!

Issue #1735, opened 2023-04-13
Remove MetadataType from the core package struct


wagoodman:
(reporting on the behalf of @kzantow) The only reason to have a `MetadataType` field is so that when marshalled to JSON, there is an identifier available to determine which type to unmarshal to. However, this can be entirely handled in the JSON encoding/decoding using something similar to:

```go
func (p Package) MarshalJSON() ([]byte, error) {
    type pk Package
    return json.Marshal(struct {
        pk
        MetadataType string
    }{
        pk: pk(p),
        MetadataType:    reflect.TypeOf(p.Metadata).Name(),
    })
}

func (b *Package) UnmarshalJSON(data []byte) error {
    var typ struct {
        MetadataType string
    }
    if err := json.Unmarshal(data, &typ); err != nil {
        return err
    }

    metadataType := pkg.MetadataTypeByName[typ.MetadataType]

    type tmp Package // avoids infinite recursion
    err := json.Unmarshal(data, (*tmp)(p))
    if err != nil { return err }

    // handle metadata

    return nil
}
```

wagoodman:
Should be done in conjunction with https://github.com/anchore/syft/issues/1844 . Also consider consolidating all instances where a user would need to make an addition when a new metadata struct is added to also be generated by the same method as done in https://github.com/anchore/syft/pull/1841 .

PR #1841, opened 2023-05-23
Add test to ensure package metadata is represented in the JSON schema

wagoodman:
Today we generate our JSON schema from our go types. We create a single `ArtifactMetadataContainer` type in the `schema/json` package that has all `github.com/anchore/syft/pkg.*Metadata` structs defined on it, and use this type to infer what can be expressed in the `pkg.Package.Metadata` field. This approach has worked well, but there is one shortcoming: when a new type is added someone needs to remember to add this type to the `ArtifactMetadataContainer` type. 

This PR improves this some by adding a unit test that will fail when there is a new type that appears to be a metadata type within the `pkg` package that is not represented on `ArtifactMetadataContainer`. 

Screenshot 2023-05-23 at 3 48 13 PM

Additionally, the `generate-json-schema` make target has been updated to generate the `ArtifactMetadataContainer` struct based on the latest analysis of the source code in the `pkg` package.

Screenshot 2023-05-23 at 3 54 42 PM

All of this will help ensure that we are not introducing data shape changes without updating the schema and are hyper aware of any breaking changes occurring to the schema within the PR making that change.


schema/json/internal/metadata_types.go
```diff
@@ -0,0 +1,150 @@
...
+	root, err := repoRoot()
+	if err != nil {
+		return nil, err
+	}
+	files, err := filepath.Glob(filepath.Join(root, "syft/pkg/*.go"))
+	if err != nil {
+		return nil, err
+	}
+	return findMetadataDefinitionNames(files...)
+}
```
kzantow, line 31: There is some related data [here](https://github.com/anchore/syft/blob/main/syft/pkg/metadata.go#L46), could you use any of these exported lists? These need to be kept up to date, used when decoding CDX.

wagoodman: I thought about using these, but the `pkg.MetadataType` will be something that is removed closer to syft 1.0 https://github.com/anchore/syft/issues/1735 which means this map would be a CycloneDX-only format concern. So instead I leaned towards the current ast approach, which has the added benefit of not needing to remember to add it to a central list/map, in which case the PR protection would be a little more brittle.

kzantow: Right, but the list of all metadata types is something needed by the _Syft format decoder_ once the pkg.MetadataType goes away, which is specifically a concern with the Syft schema, is it not?

wagoodman: I see -- that is, we should have a single source of truth for these sets. I'll try to account for that in this PR (warning, the scope will creep a bit, but I think it's warranted).

wagoodman: I still agree with trying to get down to a single source of truth, however, I feel that is better tackled in https://github.com/anchore/syft/issues/1735 and https://github.com/anchore/syft/issues/1844 since there are naming inconsistencies and decoding considerations to make here. Additionally, this PR doesn't add another source of truth, this container struct being created exists today and is now being generated. Hopefully with the other two issues they can expand on this generation to include more things.

kzantow (approved):

Issue #1783, opened 2023-05-04
Allow source type input via CLI flag (not scheme)


wagoodman:
**What would you like to be added**:

Today we allow for input like this:
```
syft registry:my/image:latest
```
Which forces syft to reach out to grab `my/image:latest` directly from the registry, not through the docker daemon.

It would be great to add an optional `--from` flag that takes what we allow today as a scheme but instead as a CLI flag.

Note: we could allow for a hold over period that, if --from is empty, we fall back to todays scheme parsing approach. If someone attempts to use a scheme and specify --from an error is returned.

**Why is this needed**:

There are known error cases in specific images that have always caused issues and complexities when parsing the scheme (more recently https://github.com/anchore/syft/issues/1695 ). This moves syft forward to act more like anchorectl while removing these odd cases and code complexity (over time).

willmurphyscode:
At this point, I assume it's not possible to accept the breaking change of making the source want a proper URI scheme like `file://some-directory` or `docker://docker.io/library/alpine:latest` or whatever? Would that remove enough ambiguity to solve all the use cases that `--from` is meant to solve? In other words, am I right in thinking that `--from` is a way to work around the fact that URI's with schemes were not used initially?

kzantow:
Just an observation about the URI idea: something like `git://https://github.com/anchore/syft` is a valid URI, but looks a little weird.

In another tool, have a keyword like `--from` to make the source very explicit and it seems to work pretty well. I think it fits a bit with `--platform` and other flags we already have that affect fetching from the source. I'm 👍 with the idea.

Issue #1844, opened 2023-05-24
Add convention for JSON metadata type names and port existing values to the new convention


wagoodman:
The `pkg.MetadataType` constants will be removed from the core API in syft 1.0, however, the JSON format will still be outputting the string representation. There are inconsistencies with how these are crafted:
```
...
	ConanMetadataType              MetadataType = "ConanMetadataType"
	DartPubMetadataType            MetadataType = "DartPubMetadata"
...
```
Some of these end in `Type` and others do not. The correct answer is that they should not end in type. Why? These string constants represent type hints for the JSON format as to how to interpret the `pkg.Metadata` field. They describe the type names, not a "type of types".

The metadata type names should always be the same names as the structs. So for instance:
```
	RustCargoPackageMetadataType   MetadataType = "RustCargoPackageMetadata"

```
is not correct since the struct this represents is called `CargoPackageMetadata`.

wagoodman:
Note that any changes we make should still be compatible in our syft decoders. That is, even if we change these names, we should still be able to convert old syft documents and their package metadata without issue.

wagoodman:
| Today | Proposed | Comments |
|--------|--------|--------|
| AlpmMetadata | arch-alpm-catalog-entry | |
| ApkMetadata | alpine-apk-catalog-entry | |
| BinaryMetadata | binary-signature | |
| CocoapodsMetadataType | cocapod-lockfile | |
| ConanLockMetadataType | c-conan-lockfile | | 
| ConanMetadataType | c-conan-guidefile | |
| DartPubMetadata | dart-pubspec-lockfile | |
| DotnetDepsMetadata | dotnet-deps-guidefile | |
| DpkgMetadata | debian-dpkg-catalog-entry | | 
| GemMetadata | ruby-gemspec-guide | |
| GolangBinMetadata (aka GolangMetadata) | go-binary-buildinfo | |
| GolangModMetadata | go-module-guidefile | |
| HackageMetadata | haskell-hackage-stack-guidefile, haskell-hackage-stack-lockfile, haskell-hackage-cabal-lockfile  | This might need more work | 
| JavaMetadata | java-archive | Though there are multiple use cases, they don't appear to be mutually exclusive |
| KbPackageMetadata | microsoft-kb-package | This breaks the rule of thumb by containing "package", but "kb" is odd since it doesn't necessarily mean a "package" (originally stood for "knowledge base") |
| LinuxKernelMetadata | linux-kernel-archive | |
[...]

PR #1846, opened 2023-05-25
Refactor source API

wagoodman:
Refactors the single uber-object of `source.Source` into multiple single responsibility objects (e.g. `source.DirectorySource`, `source.StereoscopeImageSource`, etc) that all fall under the same (new) `source.Source` abstraction:

```golang
type Source interface {
	artifact.Identifiable
	FileResolver(Scope) (file.Resolver, error)
	Describe() Description
	io.Closer
}
```

There are two ways to create a source object:
1. directly
2. via detection (considering scheme + a content analysis of the reference if it's a local file)

The first is new with this PR:
```go
src, err := NewFromDirectory(
	DirectoryConfig{
		Path: "/somewhere/to/look/at",
	},
)
```

The second was a simplification of the existing approach:
```golang
detection, err := source.Detect("/somewhere/to/look/at", app.DefaultImagePullSource)
if err != nil {
	return err
}

src, err := detection.NewSource(
	&source.Alias{
		Name:    app.SourceName,
		Version: app.SourceVersion,
	},
	app.Registry.ToOptions(),
	app.Platform,
	app.Exclusions,
)
if src != nil {
	defer src.Close()
}
```

**Open questions**:
1. Today we do NOT have the `source.target` section JSON output captured in the JSON schema -- should we start to do this?

**Breaking changes**:
- JSON schema v9 bump (see ***[1]***). Note: there are some structural changes to the `source.target` section in the JSON output (now called `source.metadata`), however, we have never made guarantees about this section of the JSON document.
- The new `source.Source` interface replaces the existing object, which breaks the existing API.

**Main changes**:
- Added new `FileSource`, `DirectorySource`, and `StereoscopeImageSource` concrete types that implement the new `source.Source` interface
- There is a lot more testing around how the source ID is generated and what is used to define that ID for each source object.
- Simplified much of the `source.Input` and detection approach to a `source.Detect()` call which you can create a source object from.
- ***[1]*** Overall attempts to move the serialized source object to act like the package object: there is a top level declaration of name and version and nested metadata, and all metadata is mapped 1:1 in the output. This PR removes much of the "select a field and map it with special logic on encoding/decoding" for the source descriptions. Now it is a much simpler pass through for all fields on `source.Description`. To keep patterns in line with how packages are done, the `target` field in the JSON output of the source block has been renamed to `metadata`. This means a new JSON schema has been generated (v9).
- Elevates source.Name and source.Version to be outside of the specific source metadata. Instead it is at the top level, next to ID, as this is common to all sources. 
- Adds a new `file.ChrootContext` for the root+cwd+base logic from the directory resolver (simplifies the dir resolver and allows for reuse of the chroot-like logic).
- `FileSource` used to not respond to FilesByPath, however, now it does (the parent dir is indexed with all other files excluded now)

**Auxiliary changes**:
- Testing snapshots for formats now store the redacted output instead of the original output
- A new `internal.AllSourceMetadataReflectTypes` collection to aid in testing, ensuring completion testing for `source.*Metadata` types
- Encapsulates the source name and version into an alias object

**Deferred work**:
- any more adjustments to `source.Detect` have been deferred until adding the new `--from` option #1783

Follow up PRs:
- [ ] remove `pkg` metadata type reflect type helpers and use `packagemetadata` helpers with what was introduced with `sourcemetadata`. Related to https://github.com/anchore/syft/issues/1735
- [ ] unexport source models if possible
- [ ] move JSON schema version to the `syftjson` package
- [ ] Can we figure a better source ID for the directory source? (Today it's based on path, can we make it content driven?)

JSON Schema Diff for reviewers:
```diff
# $ diff ./schema/json/schema-8.0.1.json ./schema/json/schema-9.0.0.json
<   "$id": "https://github.com/anchore/syft/syft/formats/syftjson/model/document",
---
>   "$id": "anchore.io/schema/syft/json/9.0.0/document",
1851a1852,1857
>           "type": "string"
>         },
>         "name": {
>           "type": "string"
>         },
>         "version": {
1857c1863
<         "target": true
---
>         "metadata": true
1861a1868,1869
>         "name",
>         "version",
1863c1871
<         "target"
---
>         "metadata"

```

Addresses https://github.com/anchore/syft/issues/558#issuecomment-1525832828

Closes: #1866


wagoodman:
note from sync review: this is a good time to make all breaking changes now, can we finalize `detection.NewSource` now?

wagoodman:
from sync review:
- are we using the file.Base? Should we allow this to be configuable for a file source?
- should we name syftjson.model.source.type to metadatatype?

syft/source/directory_source.go
```diff
@@ -0,0 +1,219 @@
...
+		mutex:  &sync.Mutex{},
+	}, nil
+}
+
+// deriveIDFromDirectory generates an artifact ID from the given directory config. If an alias is provided, then
+// the artifact ID is derived exclusively from the alias name and version. Otherwise, the artifact ID is derived
+// from the path provided with an attempt to prune a prefix if a base is given. Since the contents of the directory
+// are not considered, there is no semantic meaning to the artifact ID -- this is why the alias is preferred without
+// consideration for the path.
+func deriveIDFromDirectory(cwd string, cfg DirectoryConfig) artifact.ID {
```
kzantow, line 76: why is CWD part of this?

kzantow (commented):
Left some initial feedback, overall looking like a 👍 change

syft/file/chroot_context.go
```diff
@@ -0,0 +1,151 @@
...
+			return nil, fmt.Errorf("could not evaluate base=%q symlinks: %w", base, err)
+		}
+		cleanBase, err = filepath.Abs(cleanBase)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	chroot := &ChrootContext{
+		root: cleanRoot,
```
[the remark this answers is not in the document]
wagoodman: since this was a lift and shift we can handle a bug like this in another PR

syft/formats/syftjson/test-fixtures/snapshot/TestDirectoryEncoder.golden
```diff
@@ -72,9 +72,13 @@
  ],
  "artifactRelationships": [],
  "source": {
-  "id": "eda6cf0b63f1a1d2eaf7792a2a98c832c21a18e6992bcebffe6381781cc85cbc",
+  "id": "d1563248892cd59af469f406eee907c76fa4f9041f5410d45b93aef903bc4216",
```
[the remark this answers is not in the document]
wagoodman: maybe in a follow up? (I don't have a good answer here, but would prefer to keep it for now)

syft/formats/syftjson/to_format_model_test.go
```diff
@@ -10,53 +12,170 @@ import (
...
+}
+
 func Test_toSourceModel(t *testing.T) {
-	allSchemes := strset.New()
-	for _, s := range source.AllSchemes {
-		allSchemes.Add(string(s))
+	allSources := strset.New()
+	for _, s := range internal.AllSourceMetadataReflectTypes() {
+		allSources.Add(s.Name())
 	}
```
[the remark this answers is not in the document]
wagoodman: maybe another testutils addition

syft/source/directory_source.go
wagoodman: it looks like it was needed for the chroot object, but I'm not using it to determine the ID here. I've updated this to extract cwd entirely from this execution path.

syft/source/directory_source.go
```diff
@@ -0,0 +1,215 @@
...
+// consideration for the path.
+func deriveIDFromDirectory(cfg DirectoryConfig) artifact.ID {
+	var info string
+	if cfg.Alias != nil {
+		// don't use any of the path information -- instead use the alias name and version as the artifact ID.
+		// why? this allows the user to set a dependable stable value for the artifact ID in case the
+		// scanning root changes (e.g. a user scans a directory, then moves it to a new location and scans again).
+		info = fmt.Sprintf("%s@%s", cfg.Alias.Name, cfg.Alias.Version)
+	} else {
+		log.Warn("no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)")
```
wagoodman, line 77: I wanted to call out this change explicitly -- I still haven't been able to reconcile what a "good" ID is for a directory (that isn't too expensive to figure). But I've added this warning to at least let folks know that if you're doing a directory scan an are not setting `--source-name` or `--source-version` explicitly then this is most likely wrong. I didn't call out the specific options since this is the lib path and not in the cmd pkg.

syft/formats/syftjson/to_format_model_test.go
```diff
@@ -1,72 +1,184 @@
...
 	"testing"
 
-	"github.com/scylladb/go-set/strset"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	stereoscopeFile "github.com/anchore/stereoscope/pkg/file"
 	"github.com/anchore/syft/syft/file"
-	"github.com/anchore/syft/syft/formats/syftjson/model"
+	syftjson2 "github.com/anchore/syft/syft/formats/syftjson/model"
```
kzantow, line 12: nit: `syftjson2` seems like a less than ideal alias, could we just use `model`?

syft/internal/packagemetadata/generated.go
```diff
@@ -0,0 +1,10 @@
+// DO NOT EDIT: generated by syft/internal/packagemetadata/generate/main.go
+
+package packagemetadata
+
+import "github.com/anchore/syft/syft/pkg"
+
+// AllTypes returns a list of all pkg metadata types that syft supports (that are represented in the pkg.Package.Metadata field).
+func AllTypes() []any {
+	return []any{pkg.AlpmMetadata{}, pkg.ApkMetadata{}, pkg.BinaryMetadata{}, pkg.CargoPackageMetadata{}, pkg.CocoapodsMetadata{}, pkg.ConanLockMetadata{}, pkg.ConanMetadata{}, pkg.DartPubMetadata{}, pkg.DotnetDepsMetadata{}, pkg.DpkgMetadata{}, pkg.GemMetadata{}, pkg.GolangBinMetadata{}, pkg.GolangModMetadata{}, pkg.HackageMetadata{}, pkg.JavaMetadata{}, pkg.KbPackageMetadata{}, pkg.LinuxKernelMetadata{}, pkg.LinuxKernelModuleMetadata{}, pkg.MixLockMetadata{}, pkg.NixStoreMetadata{}, pkg.NpmPackageJSONMetadata{}, pkg.NpmPackageLockJSONMetadata{}, pkg.PhpComposerJSONMetadata{}, pkg.PortageMetadata{}, pkg.PythonPackageMetadata{}, pkg.PythonPipfileLockMetadata{}, pkg.PythonRequirementsMetadata{}, pkg.RDescriptionFileMetadata{}, pkg.RebarLockMetadata{}, pkg.RpmMetadata{}}
```
kzantow, line 9: nit: I'm pretty sure this is entirely controlled by Jennifer, but it would be nice to have these each on their own line

syft/internal/packagemetadata/names.go
```diff
@@ -0,0 +1,13 @@
+package packagemetadata
+
+import (
+	"reflect"
+)
+
+func AllNames() []string {
```
kzantow, line 7: question: I missed this in the sync review, but where the "incorrect" names being handled?

syft/source/directory_source.go
kzantow: should this be an Info or Debug message? This will show for _every_ directory scan otherwise, I think

kzantow (approved):

syft/source/directory_source.go
wagoodman: I think that's the problem with the directory source ID -- it's really not adequate for any use case where you're trying to cross correlate SBOMs historically since it's brittle and may overlap with entirely unrelated scans. It feels like it should be a warning for that reason.

syft/formats/syftjson/to_format_model_test.go
wagoodman: I thought I got all of these (goland auto refactor). I'll get it back to model.

syft/internal/packagemetadata/generated.go
wagoodman: correct about jennifer not enabling this, but I also elected to not muck with the generated code this go around.

syft/internal/packagemetadata/names.go
wagoodman: this PR doesn't go all the way with the `packagemetadata`, that's reserved for https://github.com/anchore/syft/issues/1844 . In short, this isn't used yet (except for a test), only AllTypes() is used and is consistent with the `sourcemetadata` package approach. The value add for adding this in this PR was to add the completion test around `packagemetadata.AllNames()` to detect when there is drift with the ast analysis and the generated code... a test that will be extended to include the JSON name mapping in #1844 (as done in the `sourcemetadata.AllNames()` completion test).

syft/source/directory_source.go
wagoodman: I did find that this wasn't showing up though, and I discovered why: we can't assume that a pointer to alias means that the alias is optional, we still need to check individual fields. I've updated the API to not take alias pointers in config objects (any spot on the exported API where they were function args it is still a pointer though). This allows for the zero value to be used in configs (or for the caller to fully specify like with the cmd package) and we need to check the name and version explicitly, which was already being done in some circumstances.

PR #1855, opened 2023-05-30
accept main.version ldflags even without vcs

deitch:
syft's workaround for the go binary `(devel)` issue is [here](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/golang/parse_go_binary.go#L85-L120).

It looks at `ldflags` and, if it cannot find a valid `main.version` (or similar), it tries to calculate the go pseudo-version, which needs vcs info.

The current design does not check `ldflags` unless it finds vcs info _first_. Of course, if it finds `ldflags` info, it doesn't actually need the `vcs` info.

This PR changes it around so that it first tries to read `ldflags` and only after it fails to do so, does it try vcs.

I also added tests for this specific case, as well as adding tests for `main.version` and `main.Version`.


deitch:
Fixed the lint issue

wagoodman (approved):
nice adjustment!

deitch:
Thanks! I will look for it in the next release, 0.83.0, I guess? When does that come out?

PR #1861, opened 2023-06-05
in directory indexer, handle outside symlinks

### LINKED ISSUE #1860: symlinks within a scanned directory tree are parsed outside the tree, failing if target does not exist
**What happened**:

Here is a simple test case:

```sh
$ mkdir /tmp/syft
$ touch /tmp/syft/a /tmp/syft/b
$ ln -s /foo/bar /tmp/syft/link
```

The `link` can point to anywhere, as long as it both starts with `/` and does not exist.

The error you get will be:

```
2023/06/01 13:43:01 error during command execution: 1 error occurred:
        * unable to determine resolver while cataloging packages: unable to create directory resolver: unable to index filesystem path="/foo/bar": lstat /foo: no such file or directory
```

Ideally, it would recognize that `/tmp/syft/link` points outside of the scanned directory and not follow it. At minimal, it should handle it and not fail.

See above.

The commit that caused the issue is [6afbffce](https://github.com/anchore/syft/commit/6afbffce283dd3f3f0372dc90c034f061b03dc5f)

**Root cause analysis**

Coming in an edit

deitch:
Fixes #1860 

Ensures that if a symlink in a directory points to something that does not exist, it logs a debug message and continues, rather than adding a new root which does not exist, which would cause a fatal error.

### DIFF
--- a/syft/internal/fileresolver/directory_indexer.go
+++ b/syft/internal/fileresolver/directory_indexer.go
@@ -370,6 +370,13 @@ func (r directoryIndexer) addSymlinkToIndex(p string, info os.FileInfo) (string,
 	metadata.LinkDestination = linkTarget
 	r.index.Add(*ref, metadata)
 
+	// if the target path does not exist, then do not report it as a new root, or try to send
+	// syft parsing there.
+	if _, err := os.Stat(targetAbsPath); err != nil && errors.Is(err, os.ErrNotExist) {
+		log.Debugf("link %s points to unresolved path %s, ignoring target as new root", p, targetAbsPath)
+		targetAbsPath = ""
+	}
+
 	return targetAbsPath, nil
 }
 


syft/internal/fileresolver/directory_indexer.go
```diff
@@ -370,6 +370,13 @@ func (r directoryIndexer) addSymlinkToIndex(p string, info os.FileInfo) (string,
 	metadata.LinkDestination = linkTarget
 	r.index.Add(*ref, metadata)
 
+	// if the target path does not exist, then do not report it as a new root, or try to send
+	// syft parsing there.
+	if _, err := os.Stat(targetAbsPath); err != nil && errors.Is(err, os.ErrNotExist) {
+		log.Debugf("link %s points to unresolved path %s, ignoring target as new root", p, targetAbsPath)
+		targetAbsPath = ""
+	}
```
kzantow, line 378: Looking how this is used, what if this was moved above, to line 337 and behaved similar to when `os.ReadLink` fails? e.g.:

```go
	if _, err := os.Stat(targetAbsPath); err != nil && errors.Is(err, os.ErrNotExist) {
		return "", fmt.Errorf("link %s points to unresolved path %s, ignoring target as new root", p, targetAbsPath)
	}
```

kzantow (approved):
I think this is good as it is 👍

PR #1867, opened 2023-06-06
CLI flag for directory base

deitch:
The ability to add a "root" or "base" directory already exists. However, it only was exposed via API and not CLI.

This adds a flag called `--base-path <base>` which would use that as the root. Defaults to `""`.

In order to wire it up into `Source`, and specifically the whole `ParseInput*()` without adding even more variants, I rewired them to just take variadic options funcs, and created the funcs for the ones we know exist. I recognize that this looks good to me, but may be a little off the wall for others. I am more than happy to take redirection on it.

I thought about doing the same in `source.go` with `NewFromDirectoryWithName()` and `NewFromDirectoryRootWithName()` and `NewFromDirectoryWithNameVersion()` and etc. etc. But that already had the `*Root*` variants, so I left it alone. I can add it, if wanted.

Either way, this does work.


deitch:
Clean CI always is nice 😁

deitch:
How can I help move this along @kzantow ?

wagoodman:
API-wise this is about to introduce a new pattern on an object that is about to be deleted in https://github.com/anchore/syft/pull/1846 (which I'm hoping to land for final review this week). I think it makes sense to hold off on this PR, get #1846 in, then I can rebase this on the changes there.

wagoodman (approved):

deitch:
Huzzah!

deitch:
So it is just `syft packages --base-path /path/to/base`?

Is this going into a release version?

kzantow:
@deitch yes either `--base-path` or env var `SYFT_BASE_PATH=...`, it'll be in the next release, of course. We can look at getting a release done this week, does that work for you?

deitch:
Sure. Can I bother you to comment here when it is in? I will start to adopt it upstream in various places then.

kzantow:
@deitch this has made it into a release: https://github.com/anchore/syft/releases/tag/v0.85.0

CLechleitner42:
Shouldn't the base-path default to `"/"` instead of `""`?

I solved a CI problem (that occured as of the latest `syft` release) by adding `--base-path /`.

FYI we are scanning a temp directory with symlinks to `*.jar` files, because we only want a report on those JARs used by our application.

kzantow:
@CLechleitner42 could you open an issue for this? Symlinks are hard. We probably need a broader discussion to change the behavior -- but absolutely can if it's determined to be the right thing to do!

CLechleitner42:
OK, I'll create an issues shortly, first I have to make sure our CI is fixed, it blocks a major project's acceptance ;-)

Yes symlinks are not trivial to handle once you start to think about link symlinks to symlinks, loops, and so on.

PR #1870, opened 2023-06-09
fix: pom properties not setting artifact id

jneate:
I noticed when fixing https://github.com/anchore/grype/issues/985 that Grype has a field in the metadata for Artifact ID but it was an empty string and wasn't being set by Syft.


kzantow (approved):
LGTM Thanks @jneate 👍

PR #1872, opened 2023-06-10
Use package ID from decoded SBOMs when provided

jneate:
Today if you use `syft convert`, the syft API (say via grype) with an SBOM, or use the SBOM cataloger, all packages raised up from the underlying SBOM get new IDs derived from the data discovered as opposed to using the ID found within the artifact. There are pros and cons with each approach, however, this PR is changing syft's opinion on this to prefer the ID of artifacts from the SBOM discovered.

This is done by using the existing `pkg.Package.OverrideID()` at decode-time when constructing the package. If the ID is blank then we fallback to the standard derived Syft artifact ID.

Note that this approach only affects package ID and not file IDs from decoded SBOMs --that will require further work since there could be drawbacks to adding an `id` field to `file.Coordinates` (which are heavily used as map keys).

Here's an example of before and after of a grype run with these changes integrated; now the artifact IDs in the grype JSON are the native SBOM ID:

Screenshot 2025-05-07 at 10 13 04 PM

Addresses https://github.com/anchore/grype/issues/1265 

## Design alternatives

There was another approach considered where we persist the upstream cyclonedx and spdx library types onto the package and the SBOM object. This would go a long way towards lossless conversion, however, this was a little too much to bite off for now and the current approach selected (using ID overrides) does not conflict with that future goal.

## Type of change

## Checklist


jneate:
The alternate is that the entire BOM-Ref field becomes the ID instead of the package-id suffix? Happy to make said change if needed.

syft/formats/common/cyclonedxhelpers/decoder.go
```diff
@@ -92,9 +92,11 @@ func collectPackages(component *cyclonedx.Component, s *sbom.SBOM, idMap map[str
 		syftID := extractSyftPacakgeID(component.BOMRef)
 		if syftID != "" {
 			idMap[syftID] = p
+			p.OverrideID(artifact.ID(syftID))
+		} else {
+			// TODO there must be a better way than needing to call this manually:
+			p.SetID()
 		}
-		// TODO there must be a better way than needing to call this manually:
-		p.SetID()
```
kzantow, line 97: The only way this `p.OverrideID` gets called is if the `component.BOMRef` is a PURL that contains the `package-id` qualifier (which is probably only set when Syft generated the SBOM in the first place, using a PURL as the BOMRef). Additionally, this `syftID` might not even match the ID generated by the _current version of Syft_, so the way it's being used is probably wrong.

It was originally a conscious decision to re-generate the IDs for components (by calling the `p.SetID()` instead of reusing whatever ID is provided by the format in order to get reasonably consistent output. For example: if we used the BOMRef as the Syft package ID, then re-encoded CycloneDX, we'd have something like: `pkg:thing/name@version?package-id=pkg:thing/name@version?package-id=2a7443b435345b2` (ignore the lack of URL encoding). Obviously we can handle this during encoding, for example, by comparing the ID to what `SetID` would set.

A few things here:
* we should probably get rid of `extractSyftPacakgeID`. I don't think this is providing any value -- at best it's extracting the same ID that `SetID()` would generate, but more likely it is setting some different ID that is not useful in any way
* we should adjust output formats to be aware when an ID is explicitly set vs generated based on the data (this would probably look like 1 or 2 utility methods on the package -- maybe a `GenerateID` that could be used for comparison and/or an `IsIDGenerated()`)
  * We would also need to update the CycloneDX encoder and possibly other encoders to handle cases with explicitly set IDs

If we want to support reading IDs from formats instead of always generating them based on the data (I'm not sure we always do -- this seems like a configurable option), I think your observation that we should be using the `BOMRef` is accurate, maybe something like this:
```go
		if component.BOMRef != "" {
			p.OverrideID(artifact.ID(component.BOMRef))
		} else {
			p.SetID()
		}
		idMap[string(p.ID())] = p
```

syft/formats/common/cyclonedxhelpers/decoder_test.go
```diff
@@ -336,3 +336,41 @@ func Test_missingComponentsDecode(t *testing.T) {
...
+
+	packageWithId := cyclonedx.Component{
+		BOMRef:     "pkg:maven/org.springframework.boot/spring-boot-starter-test?package-id=646a5a71a4abeee0",
+		Type:       cyclonedx.ComponentTypeLibrary,
+		Name:       "spring-boot-starter-test",
+		Version:    "",
+		PackageURL: "pkg:maven/org.springframework.boot/spring-boot-starter-test",
+	}
+
+	packageID := extractSyftPacakgeID(packageWithId.BOMRef)
```
hectorj2f, line 350: There is a typo here `extractSyftPacakgeID`

syft/formats/common/cyclonedxhelpers/decoder_test.go
```diff
@@ -336,3 +336,41 @@ func Test_missingComponentsDecode(t *testing.T) {
...
+
+	packageWithoutId := cyclonedx.Component{
+		BOMRef:     "pkg:maven/org.springframework.boot/spring-boot-starter-webflux",
```
hectorj2f, line 353: Question: Why don't we reuse the BomRef value ? I assume sboms use the BomRef as an unique reference, therefore I am wondering why this initial unique ID for this SBOM needs to be adapted as done on `sbom.Artifacts.Packages.PackagesByName(packageWithoutId.Name)` or ` sbom.Artifacts.Packages.Package(artifact.ID(packageID)))`.

kzantow:
Hi @jneate -- I had a discussion with the team about this and there are some definite issues with directly using IDs from SBOMs as Syft internal object IDs. Although this PR is specific to CycloneDX, if we were to do this with SPDX, without more intelligent behavior, we would output IDs that increasingly grow in size.

A directly related thing we are interested in is maintaining additional information from SBOM formats that Syft doesn't support -- for example VEX information in CycloneDX currently will just be dropped if we simply converted CycloneDX -> CycloneDX.

I think we'd like to try to do a couple things on this front:
a) include this unknown information as some sort of metadata that can be output at least by the same format
b) track IDs in this metadata somewhere in a specific data location such that we could reuse them when outputting the same format

Unfortunately, for this PR, we probably don't want to accept as-is where we're just setting the Syft ID directly, but instead with some sort of additional field like `FormatID`. If it would help move things forward, we could chat about this more synchronously on our [community Slack](https://anchore.com/slack) or even jump on a Zoom to discuss (or, of course one of [our community meetings](https://docs.google.com/document/d/1ZtSAa6fj2a6KRWviTn3WoJm09edvrNUp4Iz_dOjjyY8/edit#heading=h.s8nj4k878m27)).

wagoodman:
This has been implemented with a different approach on #3865 ; I was tempted to update this PR with the changes, but it was different enough and this PR was several commits behind main

wagoodman:
I've changed my mind on this -- if we can make this behavior change without introducing new fields then we should prefer that minimal approach. I'l get this across the finish line today.

syft/format/internal/testutil/redactor.go
```diff
@@ -39,7 +40,67 @@ func NewPatternReplacement(r *regexp.Regexp) PatternReplacement {
...
 func (p PatternReplacement) Redact(b []byte) []byte {
-	return p.Search.ReplaceAll(b, []byte(p.Replace))
+	if len(p.Groups) == 0 {
+		return p.Search.ReplaceAll(b, []byte(p.Replace))
+	}
+
+	return p.redactNamedGroups(b)
+}
+
+func (p PatternReplacement) redactNamedGroups(b []byte) []byte {
```
kzantow, line 50: Rather than adding more specialized redacting functionality here, could we just create different test(s) to validate the ID behavior? (effectively the same as an earlier comment)

kzantow (approved):
Nothing blocking, but I something doesn't sit well for me seeing hashes no longer redacted in the snapshots it seems like something slightly different could be done for testing here

syft/format/cyclonedxxml/encoder_test.go
```diff
@@ -90,16 +90,24 @@ func TestCycloneDxImageEncoder(t *testing.T) {
...
 				`([0-9]+)-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])[Tt]([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)(\.[0-9]+)?(([Zz])|([+|\-]([01][0-9]|2[0-3]):[0-5][0-9]))`: `redacted`,
 
-				// image hashes and BOM refs
+				// image hashes
 				`sha256:[A-Za-z0-9]{64}`: `sha256:redacted`,
 
-				// serial numbers and BOM refs
-				`(serialNumber|bom-ref)="[^"]+"`: `$1="redacted"`,
+				// serial numbers
+				`(serialNumber)="[^"]+"`: `$1="redacted"`,
```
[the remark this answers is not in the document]
wagoodman: if we redact package bom-refs, then all package will have the same ID of `redacted` and incorrectly dedup to a single package

syft/format/internal/testutil/redactor.go
wagoodman: this is necessary to isolate the source bomref (which is unstable between two runs) from package bomrefs (which is unstable between code changes)

PR #1873, opened 2023-06-11
fix: allow valid cyclonedx input with no components

jneate:
This resolves https://github.com/anchore/grype/issues/1005

The cyclonedx-go library returns Nil when presented with:

```xml
<components></components>
```

But returns an empty slice when presented with:

```json
"components": []
```

The fix is to check if the format is XML, then inspect the XMLNS field and for other formats inspect the Components field

### DIFF
--- a/syft/formats/common/cyclonedxhelpers/decoder.go
+++ b/syft/formats/common/cyclonedxhelpers/decoder.go
@@ -23,7 +23,7 @@ func GetValidator(format cyclonedx.BOMFileFormat) sbom.Validator {
 			return err
 		}
 		// random JSON does not necessarily cause an error (e.g. SPDX)
-		if (cyclonedx.BOM{} == *bom || bom.Components == nil) {
+		if (cyclonedx.BOM{} == *bom || ((format == cyclonedx.BOMFileFormatXML && bom.XMLNS == "") || bom.Components == nil)) {
 			return fmt.Errorf("not a valid CycloneDX document")
 		}
 		return nil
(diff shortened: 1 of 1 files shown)


syft/formats/common/cyclonedxhelpers/decoder.go
```diff
@@ -23,8 +26,14 @@ func GetValidator(format cyclonedx.BOMFileFormat) sbom.Validator {
...
-		if (cyclonedx.BOM{} == *bom || ((format == cyclonedx.BOMFileFormatXML && bom.XMLNS == "") || bom.Components == nil)) {
-			return fmt.Errorf("not a valid CycloneDX document")
+		if format == cyclonedx.BOMFileFormatXML {
+			if (!strings.Contains(bom.XMLNS, cycloneDXXmlSchema) || cyclonedx.BOM{} == *bom) {
+				return fmt.Errorf("not a valid CycloneDX document")
+			}
+		} else {
+			if (bom.Components == nil || cyclonedx.BOM{} == *bom) {
+				return fmt.Errorf("not a valid CycloneDX document")
+			}
```
kzantow, line 36: Having nested `if` statements like this duplicates some of the logic and makes things a little harder to reason about. What about a change like this?
```suggestion
		xmlWithoutNS := format == cyclonedx.BOMFileFormatXML && !strings.Contains(bom.XMLNS, cycloneDXXmlSchema)
		if cyclonedx.BOM{} == *bom || bom.Components == nil || xmlWithoutNS {
			return fmt.Errorf("not a valid CycloneDX document")
		}
```

kzantow (approved):

PR #1883, opened 2023-06-17
test

cyberoslab:


kzantow:
I don't think we would like these changes made to the workflows -- in part, this is downgrading multiple action versions. We already have golangci-lint running for all PRs/main commits. If there are some changes here you feel would be useful, please update this PR. In the meantime, I'm going to close this as I do not believe these changes are intended for Syft

PR #1888, opened 2023-06-22
Switch UI to bubbletea

wagoodman:
This PR adds a new [bubbletea](https://github.com/charmbracelet/bubbletea) UI. This deprecates the existing jotframe-base UI (will be removed before syft 1.0).

Additionally makes the following changes:
- fixes the attestation UI element to show up as soon as possible instead of when the first bytes arrive from the cosign execution.
- adds report and notification event types and swaps over to using these for command results.
- renames the CLI-only events to be prefixed with `CLI` to be more clear about the context of when the events are used (not part of the syft lib)
- adds snapshot-based testing for UI components tested through the shared UI handler
- adds the log redaction capabilities to allow for the UI to leverage filtering secrets in report output.


spiffcs (approved):

spiffcs:
🟢 after pairing on the race condition in tests - thanks for the info!

PR #1891, opened 2023-06-23
fix: add support for Dart SDK package dependencies

rufman:
SDK package dependencies such as Flutter cause yaml parsing errors. Update the parser so that lockfiles including these packages can be evaluated.

### DIFF
--- a/syft/pkg/cataloger/dart/parse_pubspec_lock.go
+++ b/syft/pkg/cataloger/dart/parse_pubspec_lock.go
@@ -1,11 +1,12 @@
 package dart
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"sort"
 
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/dart/parse_pubspec_lock.go
```diff
@@ -31,13 +32,39 @@ type pubspecLockPackage struct {
 }
 
 type pubspecLockDescription struct {
+	isString bool
+	str      string
+	object   pubspecLockDescriptionObject
+}
+
+type pubspecLockDescriptionObject struct {
```
kzantow, line 40: I don't think this change in structure is necessary -- this can all be handled in the UnmarshalYAML

syft/pkg/cataloger/dart/parse_pubspec_lock.go
```diff
@@ -31,13 +32,39 @@ type pubspecLockPackage struct {
...
+		var t pubspecLockDescriptionObject
+		if err := value.Decode(&t); err != nil {
+			return err
+		}
+		s.isString = false
+		s.object = t
+		return nil
+	}
+	return errors.New("Unexpected type: Expected string or Description map")
+}
```
kzantow, line 66: I think you could avoid the struct changes and somewhat hard to read changes with the conditions below using an alternate implementation here. Something like (completely untested):
```go
func (p *pubspecLockDescription) UnmarshalYAML(value *yaml.Node) error {
	if value.Decode(&p.Name) == nil {
		return nil
	}
	type pld pubspecLockDescription // create a new type to avoid infinite recursion
	var p2 pld
	if err := value.Decode(&p2); err != nil {
		return err
	}
	*p = pubspecLockDescription(p2)
	return nil
}
```
An example of this technique [is here](https://github.com/spdx/tools-golang/blob/4b477e88fcb3d02aa302d976bc3768f4e34c5ca3/spdx/v2/v2_3/document.go#L84-L99).

kzantow (approved):
LGTM -- thanks for the contribution @rufman !

PR #1892, opened 2023-06-23
Simplify the SBOM writer interface

wagoodman:
This PR attempts to simplify the `sbom.Writer` interface by removing required methods and consolidating helper functions for writers to the `cmd/options` package.

This reduces the current interface:
```golang
type Writer interface {
	// Write writes the provided SBOM
	Write(SBOM) error

	// Bytes returns the bytes of the SBOM that would be written
	Bytes(SBOM) ([]byte, error)

	// Closer a resource cleanup hook which will be called after SBOM
	// is written or if an error occurs before Write is called
	io.Closer
}
```

To this:
```golang
type Writer interface {
	Write(SBOM) error}
}
```

The primary reasons being:
- we don't support writing multiple SBOMs or streaming, so a `Close()` isn't necessary. The `Write()` call should implicitly close any destination.
- `Bytes(...) []bytes ...` is ultimately a reading call, not a writing call, which is the opposite meaning of this interface

Additionally, many of the helpers around making writers are in the same `sbom` package, however, these helpers seem primarily targeted towards the CLI argument parsing, which should reside under the `cmd` package. This is why many of the writer helpers have been moved to `cmd/options`.

This change cause downstream changes necessary in the `attest` command so took additional time simplifying sbom.Writer usage in that command.

Partially addresses syft 1.0 work.


cmd/syft/cli/options/writer.go
```diff
@@ -63,7 +85,113 @@ func parseOutputs(outputs []string, defaultFile, templateFilePath string) (out [
...
+	writers []sbom.Writer
+}
+
+type nopWriteCloser struct {
+	io.Writer
+}
+
+func (n nopWriteCloser) Close() error {
+	return nil
+}
```
kzantow, line 123: nit: organization -- it would probably be good to keep all the `sbomMultiWriter` stuff together

kzantow (approved):

cmd/syft/cli/options/writer.go
wagoodman: yeah, this file got a little awkward, but believe it or not the https://github.com/anchore/syft/pull/1888 PR is about to remove this since it will write the the event bus. I split out this work to make the bubbletea PR smaller :)

PR #1897, opened 2023-06-27
Introduce indexed embedded CPE dictionary

luhring:
This PR tries out a new approach to CPE generation (that was [discussed briefly in Slack](https://anchorecommunity.slack.com/archives/C027JE5M345/p1687360918565849)) in a limited capacity.

In this PR, we leverage the information in NVD's [CPE Dictionary](https://nvd.nist.gov/products/cpe) to understand what CPEs are actually defined — and, thus usable when querying NVD's CVE data — and then attempt to relate CPE dictionary entries to identifiable ecosystem packages that Syft's SCA logic detects.

We capture this knowledge in a concise JSON file that we embed into Syft at build time. (This embedded file is currently ~95 KB.)

Finally, when Syft is adding CPEs to each package it discovered, it first looks for entries in the embedded dictionary index, and uses the CPE from the dictionary if available. If no entry can be found, Syft falls back to today's existing CPE generation logic. The net effect is that Syft's existing generation logic is still used the majority of the time, but Syft's CPE quality greatly improves when a hit is found in the embedded dictionary.

### Example

Using Chainguard's Node.js image, here's what Syft does **before** this change:

```console
$ syft cgr.dev/chainguard/node -qo json | jq '.artifacts[] | select(.name == "ansi-regex").cpes'
[
  "cpe:2.3:a:ansi-regex:ansi-regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi-regex:ansi_regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi_regex:ansi-regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi_regex:ansi_regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi:ansi-regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi:ansi_regex:5.0.1:*:*:*:*:*:*:*"
]
```

... and **after** this change:

```console
$ go run ./cmd/syft cgr.dev/chainguard/node -qo json | jq '.artifacts[] | select(.name == "ansi-regex").cpes'
[
  "cpe:2.3:a:ansi-regex_project:ansi-regex:5.0.1:*:*:*:*:node.js:*:*"
]
```

It's important to note that in the "before" state, Syft makes several guesses at the CPE for the package, but none of its guesses are correct. And with the new approach, Syft gets the CPE right in the first try. This can be confirmed by setting the CPE version to `*` and querying NVD's CVE data.


luhring:
This is odd... 🤔  ([link](https://github.com/anchore/syft/actions/runs/5497306836/jobs/10017970119?pr=1897#step:4:11))

```console
Building snapshot artifacts
# create a config with the dist dir overridden
echo "dist: ./snapshot" > ./.tmp/goreleaser.yaml
cat .goreleaser.yaml >> ./.tmp/goreleaser.yaml
# build release snapshots
./.tmp/goreleaser release --clean --skip-publish --skip-sign --snapshot --config ./.tmp/goreleaser.yaml
make: ./.tmp/goreleaser: Command not found
make: *** [Makefile:329: snapshot] Error [12](https://github.com/anchore/syft/actions/runs/5497306836/jobs/10017970119?pr=1897#step:4:13)7
Error: Process completed with exit code 2.
```

luhring:
From a [suggestion in the community Slack](https://anchorecommunity.slack.com/archives/C027JE5M345/p1688052242736289?thread_ts=1687809279.223319&cid=C027JE5M345), I tried to run the **quality gate** locally to make sure this change wouldn't cause any regression in matching quality. I'm about 60% sure I did this correctly. 😆 

I used this in `.yardstick.yaml` in Grype:

```yaml
tools:

  - name: syft
    # note: we want to use a fixed version of syft for capturing all results (NOT "latest")
    version: v0.74.1
    produces: SBOM
    refresh: false

  - name: syft
    # note: we want to use a fixed version of syft for capturing all results (NOT "latest")
    version: cpe-dict
    produces: SBOM-new
    refresh: false

  - name: grype
    version: latest
    takes: SBOM

  - name: grype
    version: latest
    takes: SBOM-new
```

And I ran `make capture` and then `make validate`. I didn't notice any differences in the output, and got:

```console
Quality gate passed!
```

syft/pkg/cataloger/common/cpe/dictionary/index-generator/testdata/official-cpe-dictionary_v2.3.xml
wagoodman, line 1: mind trimming this test fixtures down to a smaller selection?

wagoodman: actually, I think a one-time 25k file is ok, probably won't attempt to update this fixture often (if at all?)

wagoodman (commented):
Overall looks great! Looking forward to more accurate CPEs being generated -- left one comment around adding unit tests around specific expectations during indexing.

syft/pkg/cataloger/common/cpe/dictionary/index-generator/generate.go
```diff
@@ -0,0 +1,228 @@
...
+	prefixForRubyGems       = "https://rubygems.org/gems/"
+	prefixForRubyGemsHTTP   = "http://rubygems.org/gems/"
+	prefixForNativeRubyGems = "https://github.com/ruby/"
+	prefixForPyPIPackages   = "https://pypi.org/project/"
+	prefixForJenkinsPlugins = "https://github.com/jenkinsci/"
+	prefixForRustCrates     = "https://crates.io/crates/"
+)
+
+// indexCPEList creates an index of CPEs by ecosystem.
+func indexCPEList(list CpeList) *dictionary.Indexed {
```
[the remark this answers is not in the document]
luhring: Added! --> https://github.com/anchore/syft/pull/1897/commits/1a2b4aecbc68d23709231a3af850db946df8c4d7

That also helped me realize a simplification to the addEntry funcs' parameters 🙌

wagoodman (approved):

Makefile
```diff
@@ -338,7 +343,7 @@ release:
 	@.github/scripts/trigger-release.sh
 
 .PHONY: ci-release
-ci-release: ci-check clean-dist $(CHANGELOG)
+ci-release: ci-check clean-dist $(CHANGELOG) cpe-index
```
wagoodman, line 346: actually, generating new data during the release step isn't the right place for this. Though it answers the question of how this gets refreshed, it introduces potentially breaking changes right at the release, bypassing testing.

wagoodman (changes requested):
I want to get this change in, but I'm going to take one more pass at validating the impact downstream with grype. Additionally I added one more comment about how the data is getting refreshed that will need to get addressed. I don't have a specific suggestion yet.

Makefile
luhring: Yeah, I think we can just take that step out, lean on the index that's checked into the repo, and periodically regenerate what's in the repo. I've seen other projects take a similar approach with generated/embedded data. How does that sound to you?

syft/pkg/cataloger/common/cpe/cpe-index.json
```diff
@@ -0,0 +1,1293 @@
...
+      "y18n": "cpe:2.3:a:y18n_project:y18n:*:*:*:*:*:node.js:*:*",
+      "yargs-parser": "cpe:2.3:a:yargs:yargs-parser:*:*:*:*:*:node.js:*:*",
+      "yttivy": "cpe:2.3:a:yttivy_project:yttivy:*:*:*:*:*:node.js:*:*",
+      "yyooopack": "cpe:2.3:a:yyooopack_project:yyooopack:*:*:*:*:*:node.js:*:*",
+      "yzt": "cpe:2.3:a:yzt_project:yzt:*:*:*:*:*:node.js:*:*",
+      "zjjserver": "cpe:2.3:a:zjjserver_project:zjjserver:*:*:*:*:*:node.js:*:*",
+      "zwserver": "cpe:2.3:a:zwserver_project:zwserver:*:*:*:*:*:node.js:*:*"
+    },
+    "pypi": {
+      "0.0.1": "cpe:2.3:a:pypi:pypi:*:*:*:*:*:*:*:*",
```
wagoodman, line 1017: this seems wrong, but I haven't looked at the underlying data yet to check this (wanted to at least raise it up even if I haven't dug in yet)

westonsteimel: Ugh, https://nvd.nist.gov/vuln/detail/CVE-2022-34500, https://nvd.nist.gov/vuln/detail/CVE-2022-34501 - those CPEs just look really wrong

westonsteimel: Guessing it's because of that entry

luhring: Yeah this was an interesting one -> https://pypi.org/project/0.0.1/

wagoodman:
It looks like there wasn't sufficient sample coverage in the labeled data we use in the grype quality gate to demonstrate performance improvement. But! the gate run did show that from the existing samples, there was no degradation in performance (which is half the battle to prove). 

I did an ad-hoc analysis with: 
 - python packages: 22
 - npm packages: 50
 - gem packages: 31
 
where each package was selected based on the CVEs that existed (which helped determine which versions of the package to select), so changes for matches against these packages was high. Here's what I found

- Additional TPs found due to cpe fix: 14
- Additional FNs due to cpe fix: 0
- FPs eliminated due to cpe fix: 0
- FPs introduced due to cpe fix: 0

Which is a clear win 🎉 ! I deferred looking specifically at the jenkins plugins and rust crates until we can incorporate this analysis approach into the quality gate. This is slightly different than our typical analysis since we want to ignore unmatched labels (partial evaluation) which is flagged as invalid by the current gates. In the future we'll allow for images under test with both modes of analysis.

luhring:
@wagoodman Thanks for the analysis. Glad it's looking all positive!

In case it helps with future coverage testing, one bucket of **false positives that this PR solves** (and I'm really looking forward to) is matching Ruby's bundled openssl gem with OpenSSL itself. The presence of this CPE in the index — `"cpe:2.3:a:ruby-lang:openssl:*:*:*:*:*:*:*:*"` — saves Grype from incorrectly matching the gem to loads of OpenSSL vulns that don't at all affect Ruby's wrapper.

What's the next step here?

wagoodman:
>  one bucket of false positives that this PR solves is matching Ruby's bundled openssl gem with OpenSSL itself

nice! The packages I selected were random samples so I happened to not run across that case was all (openssl wasn't in there for certain). That should be an obvious win here 🙌 .

Additionally I just updated the branch to account for how this data gets generated -- right now by a weekly workflow that would show up as a PR.

Let me try out that workflow as a sanity check then I think this is ready to be merged.

wagoodman (approved):
great work 🚀  I added one more test to check that the generated data is wired to the function that uses it.

luhring:
Thanks a million 🙇 ❤️  — very excited about this one

PR #1914, opened 2023-07-05
Add file source digest support

wagoodman:
With #1846 it became possible to express the digest of a scanned file (e.g. `syft ./go.mod -o json` can now support showing a digest in the `.source` section). This new feature was only added in the json schema but not wired up to start adding values -- this PR adds this wiring.

The default algorithm for file sources is sha-256, however, this is configurable with `SYFT_SOURCE_FILE_DIGEST` which can take one or more values (comma separated). This additionally expands the allowable digest algorithms in general for syft to include other sha224, sha384, and sha512. While working on this I attempted to move more `file` API surface area to under `internal`.

I've migrated the `source-name` and `source-version` configuration options to reside under the new `source` struct, which will break users depending on `source-name` and `source-version`. This does not affect users depending on the CLI option.


.gitignore
```diff
@@ -1,3 +1,5 @@
+go.work
```
spiffcs, line 1: Nice

spiffcs (approved):

PR #1915, opened 2023-07-05
fix(install): return right error code

### LINKED ISSUE #1566: Install script returns exit code 0 even if install fails

- Attempt to install syft via the recommended method (curl piped to sh)
- Have some issue that would cause the installation to fail (e.g. no permissions on the target directory)

**What happened**:

- The installation script exits with code 0 (verified this wasn't a pipe issue etc)

- The installation script exits with a nonzero code when it fails

- Since the install script does not set `-e`, perhaps [this](https://github.com/anchore/syft/blob/890fb3f0e8595c1bd4a746d50fa32ac44a0607d9/install.sh#L697) should store the exit code and then return it later?

**Environment**:
- Output of `syft version`: n/a
- OS (e.g: `cat /etc/os-release` or similar): Tested on Unbutu but it shouldn't really matter

lorsatti:
This resolves #1566.


spiffcs (approved):

spiffcs:
I'll take care of the rebase here and get this merged - thanks for the fix!

PR #1918, opened 2023-07-06
Remove erroneous Java CPEs from generation

luhring:
Per discussion w/ @wagoodman: filter out CPEs known to be incorrect, specifically as CPEs that match non-Maven components to Maven vulnerabilities.

Also adds one case for counting a "Gradle Enterprise" plugin as Gradle Enterprise itself.


luhring:
Linking to https://github.com/anchore/grype/issues/1179 since this will solve a decent portion of the false positives reported there

wagoodman (approved):
thanks for the fix!

PR #1926, opened 2023-07-10
Don't use the actual redis or grpc CPEs for gems

luhring:
Matches using CPEs `redis:redis` and `grpc:grpc` for **_Ruby Gems_** will always be incorrect.


kzantow (approved):

luhring:
that was fast

PR #1929, opened 2023-07-12
fix: background reader apart from global handler for testing

spiffcs:
Add individual WG to background reader to track/block on buffer being read for snapshot tests

This should fix the race condition in CI rather than bumping iterations n number of times like we had in the past


wagoodman (approved):
thanks for digging on this 🙌

PR #1932, opened 2023-07-12
Remove jotframe UI

wagoodman:
Now that syft has moved to bubbletea (in https://github.com/anchore/syft/pull/1888) and grype no longer needs the jotframe handler (in https://github.com/anchore/grype/pull/1385) we can remove the jotframe UI and dependency entirely 🎉


kzantow (approved):
🎉

PR #1934, opened 2023-07-12
feat: support top-level SPDX package and graph

### LINKED ISSUE #1241: SPDX element/relationships should be defined for the container image being described

At the moment, when scanning container image, SPDX documents contain the list of packages that the target contains.. However, there are no elements/relationships that link back to the container that is being described.. 

**Why is this needed**:

Without specifying the container image element/relationships to the packages, the ingestion of SPDX documents resulting in detached packages without being able to query for the containers that use the packages. It will be useful to be able to link them up so that queries can be made for finding entities that use certain packages.

**Additional context**:

Example graph without links to entity:

image

The SPDX element to be defined can reference the OCI PURL (https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci), and should include sha256 checksum of the container image when possible.

### LINKED ISSUE #1661: Create a package record for the artifact an SBOM described when creating a SPDX SBOM
Syft should create an SPDX Package to represent the artifact an SBOM describes when creating a SPDX SBOM.

**Why is this needed**:
The SPDX spec is silent on whether a package should be created within the SPDX document to represent the artifact or component associated with the SBOM. If the artifact is not modeled as a SPDX package at the root of the SPDX relationship tree, there is no place to record artifact metadata such as its digest, author, license, download location, etc.

For comparison, CycloneDX is explicit about this issue. The `metadata.component` field specifies the component that the SBOM describes, and it has the same schema as the dependent `components`, so you specify the artifact's digests (`hashes`), author, licenses, cpe, purl, etc.

Syft does not create a SPDX package for the artifact it analyzes. Thus, it does not record pertinent information, such as the author, digest, or license of the artifact. That means it can be difficult to identify what artifact an SBOM belong to and perform SBOM merging.

**Additional context**:
When creating a CyclonDX file Syft will fill out some of the `metadata.component` fields, although it fails to fill in the `hashes`. If the artifact is a container image, it will fill out `metadata.component.version` with the container digest (e.g., `sha256:<digest>`).

kzantow:
This PR adds support to SPDX reading and writing for a top-level element within the SBOM that represents the source which was scanned -- e.g. an image, directory, or file. Additionally, Syft does a best-effort to decode SPDX SBOMs with this information more accurately into Syft's internal data model.

Fixes #1661 
Fixes #1241


syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -123,13 +148,113 @@ func ToFormatModel(s sbom.SBOM) *spdx.Document {
...
+			RefB: spdx.DocElementID{
+				ElementRefID: p.PackageSPDXIdentifier,
+			},
+		})
+	}
+	return
+}
+
+//nolint:funlen
+func toRootPackage(s sbom.SBOM) *spdx.Package {
```
wagoodman, line 174: nit: scoping wise what's passed in could be just source.Source instead of the entire SBOM

syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -123,13 +148,113 @@ func ToFormatModel(s sbom.SBOM) *spdx.Document {
...
+		if ref, ok := ref.(reference.NamedTagged); ok {
+			versionIfUnset(ref.Tag())
+		}
+		if ref, ok := ref.(reference.Digested); ok {
+			versionIfUnset(ref.Digest().String())
+		}
+		versionIfUnset(m.ID)
+		purpose = spdxPrimaryPurposeContainer
+
+		c := toChecksum(m.ID)
```
wagoodman, line 217: should this be the digest relative to the container image manifest and not the ID? Though the ID is a checksum, it isn't semantic to the artifact, but instead derived within syft

syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -49,6 +51,125 @@ func ToSyftModel(doc *spdx.Document) (*sbom.SBOM, error) {
 	return s, nil
 }
 
+func isDirectory(name string) bool {
+	if name == "." || name == ".." || strings.HasSuffix(name, "/") || !strings.Contains(path.Base(name), ".") {
```
wagoodman, line 55: the check `!strings.Contains(path.Base(name), ".")` doesn't help in all cases to verify if a path is a file or a directory (e.g. `status.d` is a common convention for a directory).

syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -49,6 +51,125 @@ func ToSyftModel(doc *spdx.Document) (*sbom.SBOM, error) {
...
+				Layers:    nil, // TODO handle formats with nested layer packages like Tern and K8s BOM tool
+			},
+		}
+	case spdxPrimaryPurposeFile:
+		version := p.PackageVersion
+		var metadata any = source.DirectorySourceMetadata{
+			Path: p.PackageName,
+			Base: "",
+		}
+		if typeName != prefixDirectory && (typeName == prefixFile || !isDirectory(p.PackageName)) {
```
wagoodman, line 138: Why is this case needed if the prefix explicitly states file? ` !isDirectory(p.PackageName)`

That is, the `isDirectory` is trying to lexically get to an answer that I don't think is right in all cases but there is already an answer encoded next to `DocumentRoot-...`. I feel I'm missing something here.

syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -164,21 +289,21 @@ func toFileMetadata(f *spdx.File) (meta file.Metadata) {
...
 	for _, r := range doc.Relationships {
-		// FIXME what to do with r.RefA.DocumentRefID and  r.RefA.SpecialID
+		// FIXME what to do with r.RefA.DocumentRefID and r.RefA.SpecialID
 		if r.RefA.DocumentRefID != "" && requireAndTrimPrefix(r.RefA.DocumentRefID, "DocumentRef-") != string(doc.SPDXIdentifier) {
 			log.Debugf("ignoring relationship to external document: %+v", r)
 			continue
 		}
 		a := spdxIDMap[string(r.RefA.ElementRefID)]
 		b := spdxIDMap[string(r.RefB.ElementRefID)]
-		from, fromOk := a.(*pkg.Package)
```
wagoodman, line 177: is there a change in this PR that requires this? or have we been missing this all along? (I'm still reviewing, so I may get the answer to this if I keep reading)

wagoodman: ah, got my answer https://github.com/anchore/syft/pull/1934/files#diff-eef7721a95075dfa290c3772b34e9d84eb0ed3a79baebdf05cf1396f3cccc123R408

syft/formats/common/spdxhelpers/to_format_model_test.go
```diff
@@ -13,9 +15,100 @@ import (
...
+			},
+			expected: &spdx.Document{
+				SPDXIdentifier: "DOCUMENT",
+				SPDXVersion:    spdx.Version,
+				DataLicense:    spdx.DataLicense,
+				DocumentName:   "alpine:latest",
+
+				Packages: []*spdx.Package{
+					{
+						PackageSPDXIdentifier: "Package--pkg-1-pkg-1",
```
wagoodman, line 51: nit: should we be able to render `Package--...` if information is missing? (I feel like it should be `Package-...`)

syft/formats/common/spdxhelpers/to_syft_model_test.go
```diff
@@ -431,3 +434,121 @@ func Test_toSyftRelationships(t *testing.T) {
...
+			},
+			packages:      packages,
+			relationships: relationships,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			src := &test.source
+			//src.Metadata.(source.DirectorySourceMetadata)
```
wagoodman, line 529: nit: dead comment

syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -280,24 +405,32 @@ func extractPkgInfo(p *spdx.Package) pkgInfo {
...
 	}
 
 	sP.SetID()
 
-	return &sP
+	return *sP
+}
+
+func isValidPURL(purl packageurl.PackageURL) bool {
+	return purl.String() != "pkg:/"
```
lumjjb, line 433: Hmm, is this an edge case? kind of curious.

lumjjb (commented):
Thank you so much @kzantow for this PR! This is 💯 , have a couple comments/questions.

syft/formats/common/spdxhelpers/to_syft_model.go
kzantow: This is the default string if nothing is set on the purl struct, e.g.:
```
p := packageurl.PackageURL{}
p.String() // returns "pkg:/"
```

syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -49,6 +51,125 @@ func ToSyftModel(doc *spdx.Document) (*sbom.SBOM, error) {
...
+		relations = append(relations, r)
+	}
+	return
+}
+
+func findRootPackages(doc *spdx.Document) (out []*spdx.Package) {
+	for _, p := range doc.Packages {
+		for _, r := range doc.Relationships {
+			if r.RefA.ElementRefID != "DOCUMENT" ||
+				r.Relationship != spdx.RelationshipDescribes ||
```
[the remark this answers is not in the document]
kzantow: Good call. I'm glad these "same" relationships are going away in SPDX 3.0

syft/formats/common/spdxhelpers/to_syft_model.go
kzantow: Yes, I'm aware of this. It's a best-guess in some case where we have an SBOM that probably wasn't generated by syft. It's a best guess... happy for suggestions here...

syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -33,23 +45,37 @@ const (
...
+		relationships = append(relationships, toRootRelationships(rootPackage, packages)...)
+
+		// append the root package
+		packages = append(packages, rootPackage)
+	}
+
+	// add a relationship for the package the document describes
 	documentDescribesRelationship := &spdx.Relationship{
 		RefA: spdx.DocElementID{
 			ElementRefID: "DOCUMENT",
```
wagoodman, line 70: nit
```suggestion
			ElementRefID: describesID,
```

syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -123,19 +149,130 @@ func ToFormatModel(s sbom.SBOM) *spdx.Document {
...
+			})
+		}
+
+		c := toChecksum(m.ManifestDigest)
+		if c != nil {
+			checksums = append(checksums, *c)
+			purl = &packageurl.PackageURL{
+				Type:       "oci",
+				Name:       s.Name,
+				Version:    m.ManifestDigest,
```
wagoodman, line 210: should the version have the `sha256:` prefix or not? (it currently does but am trying to determine if that's intentional)

syft/formats/common/spdxhelpers/to_syft_model.go
wagoodman: Based on https://github.com/anchore/syft/pull/1934/files#r1263105669 I think we should try to only use the lexical detection in cases where we think this wasn't generated by syft. After looking at this section for a while I think it does do exactly that, but the code really doesn't make this obvious.

Something like this would make more apparent what higher-level logic this is representing:
```golang

isFromSyft := typeName == prefixDirectory || typeName == prefixFile

if isFromSyft {
  switch typeName {
    case prefixDirectory:
      // make directory metadata...
    case prefixFile:
      // make file metadata...
  }
} else {
  // lexically detect if it's a directory or file
}

```

this also motivates breaking some of the metadata creation sections into separate functions for reuse (and makes this function smaller without needing a linter ignore).

syft/formats/spdxtagvalue/test-fixtures/snapshot/TestSPDXJSONSPDXIDs.golden
```diff
@@ -8,10 +8,17 @@ Creator: Organization: Anchore, Inc
...
+PackageName: foobar/baz
+SPDXID: SPDXRef-DocumentRoot-Directory-foobar-baz
+PrimaryPackagePurpose: FILE
+FilesAnalyzed: false
+
 ##### Package: @at-sign
 
 PackageName: @at-sign
-SPDXID: SPDXRef-Package---at-sign-3732f7a5679bdec4
+SPDXID: SPDXRef-Package--at-sign-3732f7a5679bdec4
```
wagoodman, line 21: I see that this got better from `---` to `--` but is why is there the extra `-` here? (this might be out of scope for this PR, so consider this non-blocking)

syft/formats/common/spdxhelpers/to_format_model.go
wagoodman: I take this back -- I think the prefix is a common convention for this use case.

syft/formats/common/spdxhelpers/to_format_model_test.go
```diff
@@ -5,17 +5,120 @@ import (
...
 
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/file"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/sbom"
+	"github.com/anchore/syft/syft/source"
 )
 
-// TODO: Add ToFormatModel tests
+func Test_toFormatModel(t *testing.T) {
```
wagoodman, line 22: this should be a completion test, ensuring it's covering all source types https://github.com/anchore/syft/blob/e1c1832f8413de04f9c6a47c1f22b3a526473231/syft/internal/sourcemetadata/completion_tester.go#L8

syft/formats/spdxtagvalue/test-fixtures/snapshot/TestSPDXJSONSPDXIDs.golden
kzantow: It is because there is an `@` character, which is changed to `-` due to SPDX ID rules rather than just removing the character. This behavior has existed for a while, so I think it's out of scope.

syft/formats/common/spdxhelpers/to_syft_model.go
```diff
@@ -280,24 +415,32 @@ func extractPkgInfo(p *spdx.Package) pkgInfo {
...
 	metadataType, metadata := extractMetadata(p, info)
-	sP := pkg.Package{
+	sP := &pkg.Package{
 		Type:         info.typ,
 		Name:         p.PackageName,
 		Version:      p.PackageVersion,
 		Licenses:     pkg.NewLicenseSet(parseSPDXLicenses(p)...),
 		CPEs:         extractCPEs(p),
-		PURL:         info.purl.String(),
+		PURL:         purl,
```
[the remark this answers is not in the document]
kzantow: There is some use of the purl qualifiers to extract information, so I went with a variation of the first suggestion, hope that's ok.

syft/formats/spdxtagvalue/test-fixtures/snapshot/TestSPDXJSONSPDXIDs.golden
wagoodman: ah! thanks

wagoodman (approved):
nice job 🙌

Issue #1799, opened 2023-05-08
Syft reports the "minimum required version" of .NET assemblies rather than the "assembly version"


kaniini:
**What happened**:

We have been investigating the appearance of a few GHSA advisories when scanning the `cgr.dev/chainguard/dotnet-sdk:7` image.  This image contains the .NET 7.0.105 SDK, built and targeted for Wolfi, which is the latest released version.

When scanning this image, we get:

```
 ✔ Vulnerability DB        [no update available]
 ✔ Parsed image
 ✔ Cataloged packages      [1230 packages]
 ✔ Scanning image...       [6 vulnerabilities]
   ├── 0 critical, 6 high, 0 medium, 0 low, 0 negligible
   └── 6 fixed
NAME                            INSTALLED  FIXED-IN  TYPE    VULNERABILITY        SEVERITY
System.Net.Http                 4.3.0      4.3.4     dotnet  GHSA-7jgj-8wvc-jh57  High
System.Text.RegularExpressions  4.3.0      4.3.1     dotnet  GHSA-cmhx-cq75-c4mj  High
```

Both of these GHSA vulnerabilities were addressed in 2019.

The `4.3.0` is interesting.  Why is that happening?  Syft reports that it picked up that errant `System.Net.Http` assembly from parsing `/usr/share/dotnet/sdk/7.0.105/DotnetTools/dotnet-format/dotnet-format.deps.json`.  But what is the real version of `System.Net.Http`?  For that, we can reach into Powershell.  Yes, I know, forgive me:

```
PS /> $Assembly = [Reflection.Assembly]::LoadFile("/usr/share/dotnet/shared/Microsoft.NETCore.App/7.0.5/System.Net.Http.dll")
PS /> $Assembly.GetName().Version

Major  Minor  Build  Revision
-----  -----  -----  --------
7      0      0      0
```

As we can see, the version of System.Net.Http that is actually present is `7.0.0.0`, *not* `4.3.0`.  But what does the `4.3.0` mean?  Well, it turns out that the versions in those JSON files are the *compatibility* versions, or in other words, "I need at least 4.3.0 of this assembly *or later*."

**What you expected to happen**:

Syft to report version 7.0.0.0 for the System.Net.Http assembly, not version 4.3.0.

**Steps to reproduce the issue**:

Run syft on any image containing a .NET SDK.

**Anything else we need to know?**:

Not really.

**Environment**:
- Output of `syft version`:  

```
Application:        syft
Version:            0.80.0
[...]

PR #1943, opened 2023-07-15
Add support for parsing .NET assemblies

luhring:
Today, to learn about installed .NET software, Syft parses `*.deps.json` files it discovers on the filesystem. This approach is incorrect, and it leads to Syft surfacing incorrect data about what's installed, which also leads to false positives in Grype.

This PR adds a new cataloger and parser to parse .NET assemblies themselves — which are [Portable Executables](https://en.wikipedia.org/wiki/Portable_Executable) — to find the correct name, version, and other metadata for the software installed on disk.

Fixes: #1799 

Here's a new Syft scan of the image referenced in that bug report:

```console
$ go run ./cmd/syft -q cgr.dev/chainguard/dotnet-sdk:7 | grep 'System\.Net\.Http'
System.Net.Http                                                 7.0.923.36201     dotnet
...
```

Here we see a version that we'd expect (7.0.*), given that we know .NET 7 is what's installed in this image.


wagoodman:
Seems like there are a few CLI test failures (adding the snippet here since the log goes on for a while):
```
--- FAIL: TestPackagesCmdFlags (19.62s)
...
    --- FAIL: TestPackagesCmdFlags/squashed-scope-flag (0.11s)
...
    --- FAIL: TestPackagesCmdFlags/override-default-parallelism (0.16s)
    --- FAIL: TestPackagesCmdFlags/default-parallelism (0.17s)
    --- FAIL: TestPackagesCmdFlags/password_and_key_not_in_config_output (0.17s)
```

syft/pkg/cataloger/dotnet/test-fixtures/System.Buffers.dll
wagoodman, line 1: we try to not capture binaries directly in the source tree unless absolutely necessary. And when we capture them in the source tree we try to keep them as small as possible. Though 5K is pretty small, would the parser still function with a fraction of this? (say drop the last 4K?). Not a deal breaker, but needed to ask.

syft/pkg/cataloger/cataloger.go
```diff
@@ -45,7 +45,7 @@ func ImageCatalogers(cfg Config) []pkg.Cataloger {
 		apkdb.NewApkdbCataloger(),
 		binary.NewCataloger(),
 		deb.NewDpkgdbCataloger(),
-		dotnet.NewDotnetDepsCataloger(),
+		dotnet.NewDotnetPortableExecutableCalaloger(),
```
wagoodman, line 48: this can also be added to the DirectoryCatalogers as well

syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go
```diff
@@ -0,0 +1,85 @@
...
+		CompanyName:     versionResources["CompanyName"],
+		ProductName:     versionResources["ProductName"],
+		ProductVersion:  versionResources["ProductVersion"],
+	}
+
+	p := pkg.Package{
+		Name:         name,
+		Version:      version,
+		Locations:    file.NewLocationSet(f.Location),
+		Language:     pkg.Dotnet,
```
wagoodman, line 75: Is there something in particular about how the version resources are built that we can assume this is `.net`? The cataloger is selecting files that end in .dll and .exe, which don't guarantee this is written in a `.net` language.

syft/pkg/cataloger/dotnet/cataloger.go
```diff
@@ -4,10 +4,13 @@ import (
...
-const catalogerName = "dotnet-deps-cataloger"
-
 // NewDotnetDepsCataloger returns a new Dotnet cataloger object base on deps json files.
 func NewDotnetDepsCataloger() *generic.Cataloger {
-	return generic.NewCataloger(catalogerName).
+	return generic.NewCataloger("dotnet-deps-cataloger").
 		WithParserByGlobs(parseDotnetDeps, "**/*.deps.json")
 }
+
+func NewDotnetPortableExecutableCalaloger() *generic.Cataloger {
```
kaniini, line 13:
```suggestion
func NewDotnetPortableExecutableCataloger() *generic.Cataloger {
```

luhring: 😆

syft/pkg/cataloger/dotnet/test-fixtures/System.Buffers.dll
luhring: It's a good question. I'm not sure how much we can get this smaller. There are several test fixtures in the source tree larger than this DLL at the moment, in case that's helpful reasoning — but I don't think I have a more satisfying answer right now 😞

syft/pkg/cataloger/cataloger.go
luhring: Sounds good!

syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go
luhring: Good question, and I'm thinking you might be right. Incidentally my first draft left this field empty, which seems like a more correct representation of what's here. I'll make that adjustment.

syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	err = peFile.Parse()
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to parse PE file: %w", err)
+	}
+
+	versionResources, err := peFile.ParseVersionResources()
+	if err != nil {
+		return nil, nil, err
```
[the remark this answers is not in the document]
luhring: I'm not sure. I'm also not sure where a PE binary would be seen outside of the .NET ecosystem in the past 15+ years.

Are you thinking this isn't worthy of bubbling up an error? I could see that... 🤔 

(I'm also not sure if this `Parse...` function would error in this scenario or just return an empty map.)

syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go
luhring: ( --> https://github.com/anchore/syft/pull/1943/commits/368313fddac9160d8a06a01ebe8c5ac7990232f5 )

syft/pkg/cataloger/dotnet/cataloger.go
luhring: --> https://github.com/anchore/syft/pull/1943/commits/e42fd77b2f8b6d42e076a84f6cce386861260941

luhring: But Syft should **absolutely** revisit this `Calaloger` concept when it shifts its focus to Song Composition Analysis 🎵

build
wagoodman, line 1: we scrubbed this from the `main` branch yesterday which is how I think this made it in -- can you scrubbed this from your history?

Issue #931, opened 2022-03-31
Package duplicated by different cataloger


WhyJee:
**What happened**:

Scanning `almalinux:latest` image with various tools to compare the generated SBOM. At first sight Syft sounds better as total number of identified components was greater. But... making a deeper analysis showed that Syft had identified the same packages from rpm and from python.

Example Rpm cataloger finding:
```
   "name": "libcomps",
   "version": "0.1.16-2.el8",
   "type": "rpm",
   "foundBy": "rpmdb-cataloger",
   "locations": [
    {
     "path": "/var/lib/rpm/Packages",
    }
   ],
   "licenses": [],
   "language": "",
   "cpes": [
    "cpe:2.3:a:almalinux:libcomps:0.1.16-2.el8:*:*:*:*:*:*:*",
    "cpe:2.3:a:libcomps:libcomps:0.1.16-2.el8:*:*:*:*:*:*:*"
   ],
   "purl": "pkg:rpm/almalinux/libcomps@0.1.16-2.el8?arch=x86_64&upstream=libcomps-0.1.16-2.el8.src.rpm&distro=almalinux-8.5",
```

Example Python Cataloger finding:
```
   "name": "libcomps",
   "version": "0.1.16",
   "type": "python",
   "foundBy": "python-package-cataloger",
   "locations": [
    {
     "path": "/usr/lib64/python3.6/site-packages/libcomps-0.1.16-py3.6.egg-info",
    }
   ],
   ],
   "licenses": [
    "GPLv2+"
   ],
   "language": "python",
   "cpes": [
    "cpe:2.3:a:rpm_software_management:python-libcomps:0.1.16:*:*:*:*:*:*:*",
    "cpe:2.3:a:rpm_software_management:python_libcomps:0.1.16:*:*:*:*:*:*:*",
    "cpe:2.3:a:rpm_software_management:libcomps:0.1.16:*:*:*:*:*:*:*",
    "cpe:2.3:a:python-libcomps:python-libcomps:0.1.16:*:*:*:*:*:*:*",
    "cpe:2.3:a:python-libcomps:python_libcomps:0.1.16:*:*:*:*:*:*:*",
    "cpe:2.3:a:python_libcomps:python-libcomps:0.1.16:*:*:*:*:*:*:*",
    "cpe:2.3:a:python_libcomps:python_libcomps:0.1.16:*:*:*:*:*:*:*",
[...]

PR #1948, opened 2023-07-20
931: binary cataloger defaults

spiffcs:
WIP

### DIFF
--- a/syft/lib.go
+++ b/syft/lib.go
@@ -76,6 +76,14 @@ func CatalogPackages(src source.Source, cfg cataloger.Config) (*pkg.Collection, 
 
 	relationships = append(relationships, newSourceRelationshipsFromCatalog(src, catalog)...)
 
+	for _, r := range relationships {
+		for _, e := range cfg.PackageExclusions.Exclusions {
+			if e.Match(r, catalog) {
+				catalog.Delete(r.To.ID())
+			}
+		}
+	}
+
 	return catalog, relationships, release, err
 }
 
--- a/syft/pkg/cataloger/config.go
+++ b/syft/pkg/cataloger/config.go
@@ -17,6 +17,7 @@ type Config struct {
 	Parallelism       int
 }
 
+// TODO: this is a helper function that is only used in test code. It should be migrated to test code.
 func DefaultConfig() Config {
 	return Config{
 		Search:            DefaultSearchConfig(),
--- a/syft/pkg/cataloger/config.go
+++ b/syft/pkg/cataloger/config.go
@@ -9,11 +9,12 @@ import (
 // TODO: these field naming vs helper function naming schemes are inconsistent.
 
 type Config struct {
-	Search      SearchConfig
-	Golang      golang.GoCatalogerOpts
-	LinuxKernel kernel.LinuxCatalogerConfig
-	Catalogers  []string
-	Parallelism int
+	Search            SearchConfig
+	Golang            golang.GoCatalogerOpts
+	LinuxKernel       kernel.LinuxCatalogerConfig
+	Catalogers        []string
+	Parallelism       int
+	PackageExclusions PackageExclusionsConfig
 }
 
 func DefaultConfig() Config {
--- a/syft/pkg/cataloger/config.go
+++ b/syft/pkg/cataloger/config.go
@@ -12,16 +12,17 @@ type Config struct {
 	Search            SearchConfig
 	Golang            golang.GoCatalogerOpts
 	LinuxKernel       kernel.LinuxCatalogerConfig
+	PackageExclusions PackageExclusionsConfig
 	Catalogers        []string
 	Parallelism       int
-	PackageExclusions PackageExclusionsConfig
 }
 
 func DefaultConfig() Config {
 	return Config{
-		Search:      DefaultSearchConfig(),
-		Parallelism: 1,
-		LinuxKernel: kernel.DefaultLinuxCatalogerConfig(),
+		Search:            DefaultSearchConfig(),
+		Parallelism:       1,
+		LinuxKernel:       kernel.DefaultLinuxCatalogerConfig(),
+		PackageExclusions: DefaultPackageExclusionsConfig(),
 	}
 }
 
--- a/internal/config/application.go
+++ b/internal/config/application.go
@@ -85,6 +85,7 @@ func (cfg Application) ToCatalogerConfig() cataloger.Config {
 		LinuxKernel: kernel.LinuxCatalogerConfig{
 			CatalogModules: cfg.LinuxKernel.CatalogModules,
 		},
+		PackageExclusions: cataloger.DefaultPackageExclusionsConfig(),
 	}
 }
 
--- a/syft/pkg/cataloger/alpm/cataloger.go
+++ b/syft/pkg/cataloger/alpm/cataloger.go
@@ -5,9 +5,9 @@ import (
 	"github.com/anchore/syft/syft/pkg/cataloger/generic"
 )
 
-const catalogerName = "alpmdb-cataloger"
+const CatalogerName = "alpmdb-cataloger"
 
 func NewAlpmdbCataloger() *generic.Cataloger {
-	return generic.NewCataloger(catalogerName).
+	return generic.NewCataloger(CatalogerName).
 		WithParserByGlobs(parseAlpmDB, pkg.AlpmDBGlob)
 }
--- a/syft/pkg/cataloger/apkdb/cataloger.go
+++ b/syft/pkg/cataloger/apkdb/cataloger.go
@@ -8,10 +8,10 @@ import (
 	"github.com/anchore/syft/syft/pkg/cataloger/generic"
 )
 
-const catalogerName = "apkdb-cataloger"
+const CatalogerName = "apkdb-cataloger"
 
 // NewApkdbCataloger returns a new Alpine DB cataloger object.
 func NewApkdbCataloger() *generic.Cataloger {
-	return generic.NewCataloger(catalogerName).
+	return generic.NewCataloger(CatalogerName).
 		WithParserByGlobs(parseApkDB, pkg.ApkDBGlob)
 }
--- a/syft/pkg/cataloger/binary/cataloger.go
+++ b/syft/pkg/cataloger/binary/cataloger.go
@@ -7,7 +7,7 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
-const catalogerName = "binary-cataloger"
+const CatalogerName = "binary-cataloger"
 
 func NewCataloger() *Cataloger {
 	return &Cataloger{}
@@ -22,7 +22,7 @@ type Cataloger struct{}
 
 // Name returns a string that uniquely describes the Cataloger
 func (c Cataloger) Name() string {
-	return catalogerName
+	return CatalogerName
 }
 
 // Catalog is given an object to resolve file references and content, this function returns any discovered Packages
--- a/syft/pkg/cataloger/binary/package.go
+++ b/syft/pkg/cataloger/binary/package.go
@@ -31,7 +31,7 @@ func newPackage(classifier classifier, location file.Location, matchMetadata map
 		),
 		Type:         pkg.BinaryPkg,
 		CPEs:         cpes,
-		FoundBy:      catalogerName,
+		FoundBy:      CatalogerName,
 		MetadataType: pkg.BinaryMetadataType,
 		Metadata: pkg.BinaryMetadata{
 			Matches: []pkg.ClassifierMatch{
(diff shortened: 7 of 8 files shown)


spiffcs:
Quick update on this PR - after some discussion we're going to dial back the feature to not be exposed to the user yet and just fix this for the narrower cases of `name --> name` overlap in the case of `os --> binary(synthetic package)` matches

syft/lib.go
```diff
@@ -74,11 +74,31 @@ func CatalogPackages(src source.Source, cfg cataloger.Config) (*pkg.Collection,
...
+	}
 
+	// no need to consider source relationships for os -> binary exclusions
+	relationships = append(relationships, newSourceRelationshipsFromCatalog(src, catalog)...)
 	return catalog, relationships, release, err
 }
 
+func removeRelationshipsByID(relationships []artifact.Relationship, id artifact.ID) []artifact.Relationship {
+	// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
+	filtered := relationships[:0]
```
kzantow, line 93: TIL

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,68 @@
...
+
+type CategoryType string
+
+const (
+	OsCatalogerType     CategoryType = "os"
+	BinaryCatalogerType CategoryType = "binary"
+)
+
+var CatalogerTypeIndex = map[CategoryType][]string{
+	"os": {
```
kzantow, line 23: nit: could use the constants (`OsCatalogerType` and `BinaryCatalogerType`) from above for the keys here

kzantow (approved):
No blockers, but left some feedback.

kzantow:
One more question I forgot: should this PR include a boolean config option to revert this behavior?

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type CategoryType string
+
+const (
+	OsCatalogerType     CategoryType = "os"
+	BinaryCatalogerType CategoryType = "binary"
+)
+
+var CatalogerTypeIndex = map[CategoryType][]string{
```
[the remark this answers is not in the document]
spiffcs: Nice! Yea that would be a good simplification here.

My only hesitancy to change it back to that is the original config object we had discussed on the issue:
https://github.com/anchore/syft/issues/931#issuecomment-1642641417

I think keeping this as is has two advantages:
1) Is clear to future users/contributors that Os/Binary categorization types were an explicit choice as and additional condition. The parent child designation loses this nuance a little. 
2) It keeps us open to category based configuration options we may want to consider in the future

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,68 @@
...
+
+type PackageExclusionsConfig struct {
+	Exclusions []PackageExclusion
+}
+
+type PackageExclusion struct {
+	RelationshipType artifact.RelationshipType
+	ParentType       CategoryType
+	ExclusionType    CategoryType
+}
```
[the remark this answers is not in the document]
spiffcs: yep! These were missed from the pair down good catch

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var CatalogerTypeIndex = map[CategoryType][]string{
+	"os": {
+		apkdb.CatalogerName,
+		alpm.CatalogerName,
+		deb.CatalogerName,
+		nix.CatalogerName,
+		"rpm-db-cataloger",
+		"rpm-file-cataloger",
```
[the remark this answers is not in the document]
spiffcs: Ty! Resolved with [14263d3](https://github.com/anchore/syft/pull/1948/commits/14263d3d6efe77da9989aaf523462e730ca6718b)

syft/pkg/cataloger/package_exclusions.go
spiffcs: 👍 resolved with [14263d3](https://github.com/anchore/syft/pull/1948/commits/14263d3d6efe77da9989aaf523462e730ca6718b)

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	child := c.Package(r.To.ID())
+	if child == nil {
+		return false
+	}
+
+	parentInExclusion := slices.Contains(CatalogerTypeIndex["os"], parent.FoundBy)
+	childInExclusion := slices.Contains(CatalogerTypeIndex["binary"], child.FoundBy)
+
+	return artifact.OwnershipByFileOverlapRelationship == r.Type &&
```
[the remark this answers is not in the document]
spiffcs: Nice catch - resolved with [14263d3](https://github.com/anchore/syft/pull/1948/commits/14263d3d6efe77da9989aaf523462e730ca6718b)

test/integration/package_ownership_relationship_test.go
```diff
@@ -45,3 +45,39 @@ func TestPackageOwnershipRelationships(t *testing.T) {
...
+				if p.Type == pkg.BinaryPkg && p.Name == "busybox" {
+					binaryPackages = append(binaryPackages, p)
+				}
+				if p.Type == pkg.ApkPkg && p.Name == "busybox" {
+					apkPackages = append(apkPackages, p)
+				}
+			}
+
+			if len(binaryPackages) != 0 {
+				t.Errorf("expected to find no binary packages but found %d", len(binaryPackages))
```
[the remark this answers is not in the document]
spiffcs: Yea - that makes it way easier - resolved with [14263d3](https://github.com/anchore/syft/pull/1948/commits/14263d3d6efe77da9989aaf523462e730ca6718b)

spiffcs:
> One more question I forgot: should this PR include a boolean config option to revert this behavior?

Yea good call - this should now be added with [58f6d69](https://github.com/anchore/syft/pull/1948/commits/58f6d69d83cb3ada7fd43c5d3799abf71441de19)

I've opted for the new behavior of exclusions to be the default since we've identified the synthetic binary packages in some cases to be a mistake. Users can add the following to their configs to reenable the old flow:
```
exclude-binary-overlap-by-ownership: false
```

syft/pkg/cataloger/rpm/cataloger.go
```diff
@@ -11,6 +11,11 @@ import (
 	"github.com/anchore/syft/syft/pkg/cataloger/generic"
 )
 
+const (
+	DBCatalogerName   = "rpm-db-cataloger"
+	FileCatalogerName = "rpm-file-cataloger"
```
kzantow, line 16: nit: use the constants below in `NewRpmDBCataloger` and `NewFileCataloger`

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,63 @@
...
+		return false
+	}
+	child := c.Package(r.To.ID())
+	if child == nil {
+		return false
+	}
+
+	if artifact.OwnershipByFileOverlapRelationship != r.Type {
+		return false
+	}
```
kzantow, line 55: In the original comment, I was probably a little unclear. I meant to move this before line 44, so we don't bother with `c.Package` lookups at all for relationships of differing types, even though these lookups are probably just pretty fast map lookups. Not at all a blocker.

syft/pkg/cataloger/package_exclusions.go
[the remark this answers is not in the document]
kzantow: If the concern is that we want to be explicit about OS and binary cataloger types, these could be named
```golang
var osCatalogerTypes = []string { .... }
var binaryCatalogerTypes = []string { .... }
```
> keeps us open to category based configuration options we may want to consider in the future

I'm all for forward-thinking such as being open to more configuration. The suggestion was more that since we're _not_ doing that at the moment, we don't necessarily know what that would look like (although you had an option originally), so it might be better to just make whatever changes at such time as we do change the feature. Again, this is not a blocker and I'll leave it to your discernment.

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type CategoryType string
+
+const (
+	OsCatalogerType     CategoryType = "os"
+	BinaryCatalogerType CategoryType = "binary"
+)
+
+var CatalogerTypeIndex = map[CategoryType][]string{
```
kzantow, line 23: Do these need to be exported?

test/integration/encode_decode_cycle_test.go
```diff
@@ -26,7 +26,10 @@ import (
 // encode-decode-encode loop which will detect lossy behavior in both directions.
 func TestEncodeDecodeEncodeCycleComparison(t *testing.T) {
 	// use second image for relationships
-	images := []string{"image-pkg-coverage", "image-owning-package"}
+	images := []string{
+		//"image-pkg-coverage",
+		"image-owning-package",
+	}
```
kzantow, line 32: Was this an intentional change? I don't think we should remove the `image-pkg-coverage` test

kzantow (commented):
Note the comment about the change to `encode_decode_cycle_test`, I think this might be a blocker (and an accidental commit?)

test/integration/encode_decode_cycle_test.go
spiffcs: ^ nooo good catch - I'll bring that back it was just to isolate the image for testing

spiffcs: I'll update the tests with the new config option too for integration

kzantow (approved):
LGTM 👍 -- and definitely agree that having the default behavior as you noted to exclude these entries in the SBOM

README.md
```diff
@@ -482,6 +482,10 @@ default-image-pull-source: ""
 #   - "./out/**/*.json"
 exclude: []
 
+# allows users to exclude synthetic binary packages from the sbom
+# these packages are removed if an overlap with a non-synthetic package is found
+exclude-overlap-by-ownership: true
```
wagoodman, line 487: the code has `exclude-binary-overlap-by-ownership`  https://github.com/anchore/syft/pull/1948/files#diff-9dd8956cf9479ebf46ae7743d82d2d89bd81661bd13cd239651ff31f414f10b5R226

syft/pkg/cataloger/config.go
```diff
@@ -8,23 +8,14 @@ import (
...
 type Config struct {
-	Search      SearchConfig
-	Golang      golang.GoCatalogerOpts
-	LinuxKernel kernel.LinuxCatalogerConfig
-	Python      python.CatalogerConfig
-	Catalogers  []string
-	Parallelism int
-}
-
-func DefaultConfig() Config {
```
wagoodman, line 21: why delete the `DefaultConfig` method?

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,55 @@
...
+	binaryCatalogerTypes = []string{binary.CatalogerName}
+)
+
+// Exclude will remove packages from a collection given the following properties are true
+// 1) the relationship between packages is OwnershipByFileOverlap
+// 2) the parent is an "os" package
+// 3) the child is a synthetic package generated by the binary cataloger
+// 4) the package names are identical
+// This exclude was implemented as a way to help resolve: https://github.com/anchore/syft/issues/931
+func Exclude(r artifact.Relationship, c *pkg.Collection) bool {
```
wagoodman, line 34: this function seems very specific, but has a very generic name. I think the name should probably be tweaked to be a little more specific.

syft/pkg/cataloger/package_exclusions.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/anchore/syft/syft/pkg/cataloger/alpm"
+	"github.com/anchore/syft/syft/pkg/cataloger/apkdb"
+	"github.com/anchore/syft/syft/pkg/cataloger/binary"
+	"github.com/anchore/syft/syft/pkg/cataloger/deb"
+	"github.com/anchore/syft/syft/pkg/cataloger/nix"
+	"github.com/anchore/syft/syft/pkg/cataloger/rpm"
+)
+
+var (
+	osCatalogerTypes = []string{
```
wagoodman, line 17: I think the filtering should be based on the package type, not the cataloger names.

syft/pkg/cataloger/config.go
spiffcs: This function was only used as a part of `*_test.go` files. It was moved here:https://github.com/anchore/syft/blob/c7272fd6a5417f4181f72293a169e394013e5165/test/integration/utils_test.go#L57-L65

Apologies for the boy scout change on an unrelated PR - my IDE was yelling about this being `deadcode` and I could not figure out why - the refactor over to test resolved that issue

PR #1956, opened 2023-07-25
fix: update java cataloger to include similar child packages, correct PURL, and correct GroupID

spiffcs:
## Summary

This PR aims to improve 3 related aspects of syft's java cataloging.

1. Improve PURL generation. In some cases syft is appending the artifact ID to the middle section (GroupID) of a PURL
    a. EX: `pkg:maven/org.apache.xalan/xalan@2.7.2` should be `pkg:maven/org.apache/xalan@2.7.2`
2. Improve `GroupID` detection. Currently syft does not use any hierarchy for `GroupID` detection and treats all sources as equal. It does already treat fields from files with priority. We should try to obtain a GroupID answer from `Pom Properties`, then the `Pom Project` and finally, if a GroupID has not been found, the `Manifest`. `Manifest` should not return an answer if one was found in `PomProject` was, and `PomProject` should not return a GroupID answer if `PomProperties` found one.
3. Syft eliminates java packages as duplicates if the package names match. With the enhanced GroupID detection we can extend this duplicate elimination to make sure syft is not eliminating packages with similar names, but different GroupID.
    

    - Reviewers of this PR will notice that there has been an update in how the virtual path is expressed on the java metadata. This is a consequence of new java packages being added when before they were incorrectly deduped, . Rather than `vPathSuffix += ":" + pomProperties.ArtifactID`, which just uses the ArtifactID, SBOM consumers will see the full path of the package's properties used for package creation Ex: `META-INF/maven/org.glassfish.jaxb/jaxb-core/pom.properties`. This allows for better identification for child packages that look identical, but are actually forks or similar clones:
        - Previously: `/casb.war:WEB-INF/lib/jaxb-core-2.2.11.jar:jaxb-core`
        - Now:`/casb.war:WEB-INF/lib/jaxb-core-2.2.11.jar:META-INF/maven/org.glassfish.jaxb/jaxb-core`

### DIFF
--- a/syft/pkg/cataloger/common/cpe/java.go
+++ b/syft/pkg/cataloger/common/cpe/java.go
@@ -1,6 +1,7 @@
 package cpe
 
 import (
+	"sort"
 	"strings"
 
 	"github.com/scylladb/go-set/strset"
@@ -189,6 +190,7 @@ func GroupIDsFromJavaMetadata(metadata pkg.JavaMetadata) (groupIDs []string) {
 	groupIDs = append(groupIDs, groupIDsFromPomProject(metadata.PomProject)...)
 	groupIDs = append(groupIDs, groupIDsFromJavaManifest(metadata.Manifest)...)
 
+	sort.Strings(groupIDs)
 	return groupIDs
 }
 
--- a/syft/pkg/cataloger/common/cpe/java.go
+++ b/syft/pkg/cataloger/common/cpe/java.go
@@ -31,13 +31,6 @@ var (
 		"Bundle-Activator",
 	}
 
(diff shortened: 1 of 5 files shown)


syft/pkg/cataloger/common/cpe/java.go
```diff
@@ -30,6 +30,14 @@ var (
 		"Implementation-Title",
 		"Bundle-Activator",
 	}
+
+	/*
```
spiffcs, line 34: remove these notes

spiffcs:
Splitting this into two PR and merging in upstream fixes from #2032

PR #1971, opened 2023-07-28
Add ruby.NewGemSpecCataloger to DirectoryCatalogers.

evanchaoli:
I have a use case that needs to run Syft scan inside a Dockerfile. But I found that installed Ruby Gems couldn't be found. So I have to use `--catalogers=all`.

Then I noticed that `ruby.NewGemSpecCataloger()` is not included in `DirectoryCatalogers`.

But I also noticed that `python.NewPythonPackageCataloger()` is in the list, I think PythonPackage is the same place as RubyGemSpec, so I feel it's reasonable to add `ruby.NewGemSpecCataloger` to `DirectoryCatalogers`.


wagoodman:
Hey! I think doing this is a good idea, however, a little more work would be needed. Today the current gemspec cataloger looks at gemspecs that are installed within a specifications directory (which is done when installing a gem via the CLI). 

https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/ruby/catalogers.go#L19

Another constructor for the same cataloger would be needed that looks for gemspecs in any location, something like:

```golang 
// the existing one today, renamed...
func NewInstalledGemSpecCataloger() *generic.Cataloger {
	return generic.NewCataloger("ruby-gemspec-cataloger").
		WithParserByGlobs(parseGemSpecEntries, "**/specifications/**/*.gemspec")
}

// new one with a more broad glob...
func NewGemSpecCataloger() *generic.Cataloger {
	return generic.NewCataloger("ruby-gemspec-cataloger").
		WithParserByGlobs(parseGemSpecEntries, "**/*.gemspec")
}
```

evanchaoli:
@wagoodman Thanks for your comment. I wonder why do we need a broader glob? In my case, `--catalogers=all` has worked around. Will *.gemspec files exist under other directories than "specifications"?

willmurphyscode:
> Will *.gemspec files exist under other directories than "specifications"?

@evanchaoli yes, one example might be scanning the directory when a Gem's source code is checked out - we'd expected to have a `*.gemspec` in the root.

https://github.com/anchore/syft/issues/2128 has some more information. If you'd like, reworking this PR to fix #2128 would be greatly appreciated.

willmurphyscode:
@evanchaoli did I answer your question adequately? Please let me know if there's anything I can do to help here.

evanchaoli:
Hey @willmurphyscode I just haven't got bandwidth to get back to this PR. I will try to do something next week.

evanchaoli:
Hey @wagoodman, finally I got back to this PR. Please take a look at the new diff.

wagoodman (approved):

Issue #1972, opened 2023-07-28
Relationships section of CycloneDX is not outputting even when the data is present


markgalpin:
**What happened**:
I have been trying to generate syft cycloneDX outputs with a Relationships section.   After reviewing the issues and the code, it seemed like the only reliable way to do this would be to use alpine package manager.  (Since its not clear that dependency-of is generated in any other ecosystem yet, but alpine definitely creates it)
Even in the case of alpine packages that definitely have dependency-of relationships, we're not seeing any output.

**What you expected to happen**:
I expected alpine package-to-package dependencies to be listed.

**Steps to reproduce the issue**:
docker pull alpine
syft -v registry1.dso.mil/ironbank/opensource/apache/tinkerpop/gremlin -o cyclonedx-json > cyclonedx.json

look at the generated cyclonedx-json file, and there is no 'relationships' section

**Anything else we need to know?**:
I actually worked pretty hard at debugging this, this afternoon.   Finally I added logging messages in `syft/formats/common/cyclonedxhelpers/format.go` in the `toDependencies` function after lines 152 and 157, and ran a dev build to see what was happening.

When I did that, I got this output on the check to see if the FromPkg is valid:  
`bad fromPkg: {From:Pkg(name="alpine-baselayout-data" version="3.4.3-r1" type="apk" id="afce4445f1c5ffeb") To:Pkg(name="alpine-baselayout" version="3.4.3-r1" type="apk" id="4457aba38f428cff") Type:dependency-of Data:<nil>}`
for this and a bunch of other packages

so far as I can tell those IDs are valid (e.g. they show up in bomref in the cyclonedx json for the appropriate packages) but that's about as far as I'm able to get.  

I don't know go well enough to have confidence debugging what appears to be an issue with type casting/inheritance?  (e.g. as far as I can tell the code in syft/pkg/cataloger/apkdb/parse_apk_db.go line 419 is assigning an object of type pkg.Package to the relationship, so what I take to be a type assertion in line 152 of the format.go should be succeeding...)

That said, I have a PR that seems to fix the issue (will submit in a moment), but I genuinely don't know go well enough to know if it generated something worse, or how to test it properly beyond ItWorksOnMyMachineForThisCase.

**Environment**:
- Output of `syft version`:
- Application:        syft
Version:            0.85.0
JsonSchemaVersion:  9.0.0
BuildDate:          2023-07-12T17:14:54Z
[...]

PR #1973, opened 2023-07-28
fix for issue #1972: Fixed typecasting, and added more debug logging.

markgalpin:
I'll be honest, I'm not at all certain this is tested adequately, so feel free to discard this and just use it as a PoC for better code... but I THINK its a low-risk fix.


markgalpin:
I think I confused the automated request byscrewing up the DCO check twice.  I'm going to try to re-submit.

Issue #1439, opened 2023-01-07
Syft hangs when run in CircleCI


tri-adam:
**Please provide a set of steps on how to reproduce the issue**

Run Syft in a CircleCI pipeline without the quiet (`-q`) or verbose (`-v`) flags. Specifically, the job I'm running is:

```
  generate_sbom:
    docker:
      - image: alpine:3.17
    steps:
      - attach_workspace:
          at: /docker
      - run:
          name: Install Syft
          command: wget -q -O - https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
      - run:
          name: Generate SBOM
          command: syft packages -o cyclonedx-json=/docker/sbom.cdx.json "docker-archive:/docker/service.tar"
      - persist_to_workspace:
          root: /docker
          paths:
            - sbom.cdx.json
```

Eventually, the `Generate SBOM` step times out with:

```
Too long with no output (exceeded 10m0s): context deadline exceeded
```

If I add the quiet (`-q`) or verbose (`-v`) flags, the command completes in a few seconds.

**What happened**:

`syft packages` hangs when run in a CircleCI environment unless the quiet (`-q`) or verbose (`-v`) flags are added.

**What you expected to happen**:

Syft should complete regardless of whether the quiet/verbose flags are present.

**Anything else we need to know?**:

I suspect this may have the same root cause as https://github.com/anchore/grype/issues/1051.

**Environment**:
- Output of `syft version`:
```
Application:        syft
Version:            0.65.0
JsonSchemaVersion:  6.0.0
BuildDate:          2023-01-04T17:00:24Z
GitCommit:          bc1edb9c8a2fb4824bfdcac6147edc2bbf47aaf6
GitDescription:     v0.65.0
Platform:           linux/amd64
GoVersion:          go1.18.9
Compiler:           gc
```
- OS (e.g: `cat /etc/os-release` or similar):
```
# cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.17.0
PRETTY_NAME="Alpine Linux v3.17"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

kzantow:
Thanks, @tri-adam! It sounds like there is a simple workaround, so this isn't a critical issue for you, correct?

tri-adam:
[...]

PR #1977, opened 2023-07-29
fix: prevent hang if stderr is tty but stdout is not

willmurphyscode:
Previously, the rich TUI would be enabled if either stderr or stdout is a TTY, but this could cause syft to hang if stderr was a TTY and stdout was not, as in some cases of redirection.

Fixes https://github.com/anchore/syft/issues/1439

## Manual testing done:

```
# hangs forever:
❯ cat <(syft alpine:latest | head -n 1) <(syft alpine:latest | rg binary)
^C
# runs normally:
❯ cat <(go run cmd/syft/main.go alpine:latest | head -n 1) <(go run cmd/syft/main.go alpine:latest | rg binary)
NAME                    VERSION      TYPE
busybox                 1.36.1       binary

### DIFF
--- a/cmd/syft/internal/ui/select.go
+++ b/cmd/syft/internal/ui/select.go
@@ -21,7 +21,7 @@ import (
 func Select(verbose, quiet bool) (uis []clio.UI) {
 	isStdoutATty := term.IsTerminal(int(os.Stdout.Fd()))
 	isStderrATty := term.IsTerminal(int(os.Stderr.Fd()))
-	notATerminal := !isStderrATty && !isStdoutATty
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by willmurphyscode** [comment]
We might want to think about a subtler approach here. If this PR is merged as is, it will make things like this not have the nice spinning output to let the user know something is happening:

```
❯ syft alpine:latest | rg git
 ✔ Loaded image                                                                                                                          alpine:latest
 ✔ Parsed image                                                                sha256:5053b247d78b5e43b5543fec77c856ce70b8dc705d9f38336fa77736f25ff47c
 ✔ Cataloged packages              [16 packages]  
```

Issue #1961, opened 2023-07-26
Add support for determining supplier of packages


kzantow:
**What would you like to be added**:
The *supplier* field to SBOMs, in order to conform to the NITA minimum SBOM requirements.

**Why is this needed**:
Syft should generate SBOMs that include the NTIA minimum requirements. The *supplier* field is one aspect of the NITA minimum SBOM requirements which Syft is not currently populating.

**Additional context**:
This is an aspect of #632

spiffcs:
Part of the next step of this work after #1980 goes in is keying off of the example of when the originator and the supplier are different and how to read that into documents that are not handcrafted - what package manager fields/data exist to allow us to make this distinction across ecosystems on document generation

```
The SPDX document identifies the package as [glibc](https://www.gnu.org/software/libc/) 
and the Package Supplier as [Red Hat](https://www.redhat.com/), 
but the [Free Software Foundation](http://www.fsf.org/) is the Package Originator.
```

kzantow:
We also need to be able to provide a supplier to the "source", which ends up as root elements in SPDX and CycloneDX. One possibility is to add something like `--source-name` and `--source-version`, which could be `--source-supplier`, but we don't have this element in the Syft data model yet, so it would need to be added. It is _possible_ this supplier information should then be used for packages as well (e.g. who should a user contact when they need something done to the software to correct an issue/vulnerability/etc.). It might also be important to add something to [a "hints" file](https://github.com/anchore/syft/issues/31) to be able to be more specific about overriding/specifying information for specific packages.

spiffcs:
Dev Note: When implementing this feature the resulting SBOM should pass this validator:
`https://github.com/CycloneDX/cyclonedx-cli`

spiffcs:
https://github.com/anchore/syft/pull/4131 Get's us close - since this is a larger issue to update as we add more support for this I'm taking it off the board of WIP

MackJM:
Hello there, I've opened a [PR](https://github.com/anchore/syft/pull/4991) that I think will help with this. I left it as draft until I can merge my [go-rpmdb PR](https://github.com/anchore/go-rpmdb/pull/18). Please let me know if I'd need to change anything. Thanks!

MackJM:
Hey @spiffcs, I was just wondering if you guys were still interested in this and if you had any feedback on my PRs. Thanks!

PR #1980, opened 2023-07-31
feat: use originator logic to fill supplier

spiffcs:
## Summary
Syft should be filling in the supplier information to meet NTIA minimum standards for SPDX sbom generated by the tool.

Partially Addressing #1961 

There are additional refinements we can make where supplier can get it's own function when we determine a good fence for when one field should specify one value vs another:

A good example:
```
The SPDX document identifies the package as [glibc](https://www.gnu.org/software/libc/) 
and the Package Supplier as [Red Hat](https://www.redhat.com/), 
but the [Free Software Foundation](http://www.fsf.org/) is the Package Originator.
```

#### [Supplier](https://spdx.github.io/spdx-spec/v2-draft/package-information/#75-package-supplier-field)
```
Identify the actual distribution source for the package/directory identified in the SPDX document. 
This might or might not be different from the originating distribution source for the package. 
The name of the Package Supplier shall be an organization or recognized author and not a web site.
```

#### [Originator](https://spdx.github.io/spdx-spec/v2-draft/package-information/#76-package-originator-field)
```
If the package identified in the SPDX document originated from a different person or organization than identified as Package Supplier (see [7.5](https://spdx.github.io/spdx-spec/v2-draft/package-information/#7.5) above)

This field identifies from where or whom the package originally came. 
In some cases, a package may be created and originally distributed by a different third party than the Package Supplier of the package.
```

In this case `NOASSERTION` is returned when:
```
the SPDX document creator has attempted to but cannot reach a reasonable objective determination
```


syft/formats/common/spdxhelpers/to_format_model.go
```diff
@@ -514,6 +514,21 @@ func toPackageOriginator(p pkg.Package) *spdx.Originator {
...
 
+func toPackageSupplier(p pkg.Package) *spdx.Supplier {
+	// this uses the Originator function for now until
+	// a better distinction can be made for supplier
+	kind, supplier := Originator(p)
+	if kind == "" || supplier == "" {
+		return &spdx.Supplier{
+			Supplier: NOASSERTION,
+		}
+	}
```
kzantow, line 525: I think supplier is optional in SPDX, should we return `nil`?

spiffcs: I wrote in the PR description for why this case exists:
```
The SPDX document creator has attempted to but cannot reach a reasonable objective determination
```

I don't think it's a question of if it's required or optional by SPDX, but more so did we attempt to provide an answer for this field. 

I think including `NOASSERTION` for cases we failed to identify a supplier is a better indicator on improving the SBOMs quality rather than removing the field entirely. Returning nil here might communicate to some users that we didn't even attempt to identify the supplier of a given source.

spiffcs:
Update the root package to have supplier as noassertion since this is a manually synthesized package as part of the source object

syft/formats/spdxtagvalue/test-fixtures/snapshot/stereoscope-fixture-image-simple.golden
wagoodman, line 1: was this updated intentionally?

syft/formats/common/spdxhelpers/to_format_model.go
kzantow: After discussion this seems like the right thing to do for now 👍

PR #1987, opened 2023-08-02
Set precedence for root options

### LINKED ISSUE #1986: SYFT_CONFIG environment variable not supported
**What happened**:

Running `SYFT_CONFIG=foo syft package` does not pick up the application config from file `foo`.

Running `syft -c bar packages` results in the expected error `unable to load config: unable to read application config="bar"`.

I would expect the same error when the config flag is set using an environment variable `SYFT_CONFIG`. 

1. Run `SYFT_CONFIG=foo syft packages`

Notice that the error `rror during command execution: an image/directory argument is required` is produced.  It should fail when the application configuration file is not found.

The value of `configPath` is passed to `Application.LoadAllValues`.  The decision on the value of top-level happens _before_ we load values.  The values of root options should come from a viper lookup, rather than from a pflag binding.

This is related to #1143, but is a smaller scoped issue.

**Environment**:
- Output of `syft version`: latest `git`
- OS (e.g: `cat /etc/os-release` or similar): macos & linux

AidanDelaney:
Allows RootOptions to be set using command line flags or alternatively through an environmental variable.  Maintains the priority order described in Application.LoadAllValues.

Fixes #1986

### DIFF
--- a/cmd/syft/cli/options/root.go
+++ b/cmd/syft/cli/options/root.go
@@ -2,8 +2,9 @@ package options
 
 import (
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
+
+	"github.com/anchore/syft/internal"
 )
 
 type RootOptions struct {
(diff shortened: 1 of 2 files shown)


kzantow:
Thanks for this PR @AidanDelaney -- I'm currently working on a much larger change for configuration that fixes, among other things, the precedence issues. I assume this isn't a blocking issue for you, would you be okay to wait a couple weeks for the changes to make their way in and close this PR in the meantime?

AidanDelaney:
@kzantow Thanks for the context. I'll have a look at [clio](https://github.com/anchore/clio) for my own projects.

PR #1997, opened 2023-08-03
fix: update glob to use newer usr/lib/sysimage path

spiffcs:
## Summary

See the below link:
https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr

Photon does not soft link the old location of `var/lib` to the newer location like other distros. We should account for this in our RpmDBGlob search by including a search for the newer path

#### Fix ScreenShot

Note the second run of syft is the latest version. The first run has the current branches code
Screenshot 2023-08-03 at 7 08 14 PM


nurmi (approved):
LGTM!

Issue #1143, opened 2022-08-04
Command line flags not overriding configuration file values


spiffcs:
**What happened**:
When running `syft -vv alpine:latest` an error will occur if the configuration file is also set for verbosity.

**What you expected to happen**:
No error for syft to occur with command line flags and config set.

**How to reproduce it (as minimally and precisely as possible)**:

run `syft -vv alpine:latest` with a configuration file also configured with log settings

```
log:
  # use structured logging
  # same as SYFT_LOG_STRUCTURED env var
  structured: false

  # the log level; note: detailed logging suppress the ETUI
  # same as SYFT_LOG_LEVEL env var
  level: "error"

  # location to write the log file (default is not to have a log file)
  # same as SYFT_LOG_FILE env var
  file: ""
```

**Anything else we need to know?**:
Community Meeting origin 8/4

**Environment**:
- Output of `syft version`: v0.53.4
- OS (e.g: `cat /etc/os-release` or similar):
- OSX

kzantow:
NOTE: this looks like the command line precedence has an issue. It was just reported the same thing was happening for `--output` on the community Slack: https://anchorecommunity.slack.com/archives/C4PJFNEEM/p1667844831946429

kzantow:
There is also a problem with command line flags overriding _any_ config. An example is:
```
SYFT_OUTPUT=table syft alpine:latest -o json
```
... outputs the _table_ instead of _json_ but the command line parameter should override the environment variable (and config file).

PR #2001, opened 2023-08-04
chore: update CLI to CLIO

kzantow:
This PR updates the Syft CLI configuration to use CLIO, which simplifies configuration while also solving a number of problems, including:

* fixes: #1143 
* fixes: #1986
* fixes: #1634 (we originally "deprecated" this behavior, but it seems like a good time to fix it)
* fixes: #2074


cmd/syft/cli/options/registry.go
```diff
@@ -1,21 +1,30 @@
...
-	Token string `yaml:"-" json:"-" mapstructure:"token"`
+	Username  string `yaml:"username" json:"username" mapstructure:"username"`
+	Password  string `yaml:"password" json:"password" mapstructure:"password"`
+	Token     string `yaml:"token" json:"token" mapstructure:"token"`
+}
+
+var _ clio.PostLoader = (*RegistryCredentials)(nil)
+
+func (r *RegistryCredentials) PostLoad() error {
+	// TODO ensure that the list of RegistryCredentials has PostLoad called
```
kzantow, line 21: Do not merge without validating this!

kzantow: Needed a fangs update (represented with the go mod replace directive); this is working correctly

cmd/syft/internal/constants.go
```diff
@@ -0,0 +1,6 @@
+package internal
+
+const (
+	Syft        = "syft"
+	NotProvided = "[not provided]"
```
kzantow, line 5: I'm not sure it's best to have these constants, or the right location for them.

syft/event/event.go
```diff
@@ -6,12 +6,10 @@ package event
 
 import (
 	"github.com/wagoodman/go-partybus"
-
-	"github.com/anchore/syft/internal"
 )
 
 const (
-	typePrefix    = internal.ApplicationName
+	typePrefix    = "syft"
```
kzantow, line 12: Events originating from this library, regardless of the actual application name make sense to be `"syft"`, I think

test/cli/packages_cmd_test.go
```diff
@@ -294,7 +294,7 @@ func TestRegistryAuth(t *testing.T) {
 			args: args,
 			env: map[string]string{
 				"SYFT_REGISTRY_AUTH_AUTHORITY": host,
-				"SYFT_REGISTRY_AUTH_TOKEN":     "token",
+				"SYFT_REGISTRY_AUTH_TOKEN":     "my-token",
```
kzantow, line 297: Redactions apply to all output, so redacting `token` results in text being redacted containing this string.

cmd/syft/cli/commands/update.go
```diff
@@ -1,31 +1,64 @@
...
+	"github.com/anchore/syft/syft/event"
 )
 
 var latestAppVersionURL = struct {
 	host string
 	path string
 }{
 	host: "https://toolbox-data.anchore.io",
-	path: fmt.Sprintf("/%s/releases/latest/VERSION", internal.ApplicationName),
+	path: fmt.Sprintf("/%s/releases/latest/VERSION", internal.Syft),
```
wagoodman, line 24: this could probably be hardcoded now, then we can drop the very-tempting-to-reference `internal.Syft` variable entirely

cmd/syft/cli/eventloop/tasks.go
wagoodman, line 1: I think this entire eventloop package can be deleted. The only exception is this file. The SBOM builder PR was going to be promoting these task objects, but that's really out of scope for this PR. Mind at least deleting the event_loop.go(+tests) and signals.go even if tasks stays were it is for now.

cmd/syft/internal/constants.go
wagoodman: I think we should remove the syft one, could be convined to keep the not-provided one though (maybe renamed to something like `VersionMissing` or `EmptyVersion`)

syft/event/event.go
wagoodman: though, this might motivate keeping the `internal.Syft` constant

cmd/syft/cli/commands/attest.go
```diff
@@ -0,0 +1,258 @@
...
+//nolint:funlen
+func runAttest(app clio.Application, opts *attestOptions, userInput string) error {
+	_, err := exec.LookPath("cosign")
+	if err != nil {
+		// when cosign is not installed the error will be rendered like so:
+		// 2023/06/30 08:31:52 error during command execution: 'syft attest' requires cosign to be installed: exec: "cosign": executable file not found in $PATH
+		return fmt.Errorf("'syft attest' requires cosign to be installed: %w", err)
+	}
+
+	defer bus.Exit()
```
wagoodman, line 91: all of these `bus.Exit()` lines are subtle and have implications with initialization logic (so must always be run). I think it would be easier to think about this if they were moved to the wrapper function [here](https://github.com/anchore/syft/blob/abd1b43547bb4ff90cde9fcf880f57f47f4532c1/cmd/syft/cli/commands/attest.go#L74-L77). 

For instance, where it is now, it is possible to not have cosign installed and not call exit.

This also has the benefit of separating bus implementation concerns with business concerns.

cmd/syft/cli/commands/packages.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// nolint:funlen
+func runPackages(app clio.Application, opts *packagesOptions, userInput string) error {
+	err := validatePackageOutputOptions(&opts.Output)
+	if err != nil {
+		return err
+	}
+
+	writer, err := options.MakeSBOMWriter(opts.Outputs, opts.File, opts.OutputTemplatePath)
```
wagoodman, line 115: I know that `options/writer.go` didn't change, but should we consider make these expressible via only option structs? E.g. `options/output.go` on the `Output` struct add a new `SBOMWriters()` method that calls `MakeSBOMWriter` and returns a single writer? This way we could unexport `MakeSBOMWriter` entirely.

syft/event/event.go
kzantow: Note: I've removed the `internal.ApplicationName` and created the `internal.Syft` in the `cmd/internal` package, which isn't available here. There aren't any other locations where a hardcoded `"syft"` seems to be needed in the `syft/` packages from what I found.

cmd/syft/internal/constants.go
kzantow: I got rid of the Syft constant here. I'm happy to rename the `NotProvided` variable, but I'd note it's used for more than just version:
```
	version        = internal.NotProvided
	buildDate      = internal.NotProvided
	gitCommit      = internal.NotProvided
	gitDescription = internal.NotProvided
```

cmd/syft/cli/commands/packages.go
kzantow: I made a change for this. I think it's better, but it's also a tiny bit wonky because of how the `MakeSBOMWriter` was being called differently across multiple commands.

wagoodman: agreed, I like this better 👍

cmd/syft/cli/options/pkg.go
```diff
@@ -12,13 +10,14 @@ type pkg struct {
 	SearchIndexedArchives   bool             `yaml:"search-indexed-archives" json:"search-indexed-archives" mapstructure:"search-indexed-archives"`
 }
 
-func (cfg pkg) loadDefaultValues(v *viper.Viper) {
-	cfg.Cataloger.loadDefaultValues(v)
+func pkgDefault() pkg {
```
wagoodman, line 13: nit: rename to `defaultPkg` to follow existing conventions

wagoodman: (there are a bunch of theses)

cmd/syft/cli/options/packages.go
```diff
@@ -2,117 +2,167 @@ package options
...
-func (o *PackagesOptions) AddFlags(cmd *cobra.Command, v *viper.Viper) error {
-	cmd.Flags().StringVarP(&o.Scope, "scope", "s", cataloger.DefaultSearchConfig().Scope.String(),
-		fmt.Sprintf("selection of layers to catalog, options=%v", source.AllScopes))
-
-	cmd.Flags().StringArrayVarP(&o.Output, "output", "o", []string{string(table.ID)},
-		fmt.Sprintf("report output format, options=%v", formats.AllIDs()))
-
-	cmd.Flags().StringVarP(&o.File, "file", "", "",
-		"file to write the default report output to (default is STDOUT)")
+var _ interface {
```
wagoodman, line 43: nit: these assertions should be at the top of files (above the type definitions)

syft/event/parsers/parsers.go
```diff
@@ -144,30 +144,22 @@ func ParseAttestationStartedEvent(e partybus.Event) (io.Reader, progress.Progres
...
-		return nil, err
-	}
-
-	fn, ok := e.Value.(func() error)
-	if !ok {
-		return nil, newPayloadErr(e.Type, "Value", e.Value)
-	}
-
-	return fn, nil
+type UpdateCheck struct {
```
wagoodman, line 147: do we want to have type definitions for these kinds of things in the parsers package? We tend to put this kind of stuff into the `monitor` package, but that name doesn't line up very well.

cmd/syft/cli/commands/packages.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	if err := writer.Write(*s); err != nil {
+		return fmt.Errorf("failed to write SBOM: %w", err)
+	}
+
+	return nil
+}
+
+func generateSBOM(app clio.Application, src source.Source, opts *options.Packages) (*sbom.SBOM, error) {
```
wagoodman, line 187: this function really doesn't need the whole application object -- it would be more ideal to pass the clio.Identification instead in these kinds of cases (I think I saw a couple of these instances, could be wrong though)

cmd/syft/cli/commands/poweruser.go
```diff
@@ -0,0 +1,152 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			applicationUpdateCheck(app, opts.CheckForAppUpdate)
+
+			return runPowerUser(app, opts, args[0])
+		},
+	}, opts)
+}
+
+//nolint:funlen
+func runPowerUser(app clio.Application, opts *powerUserOptions, userInput string) error {
```
wagoodman, line 62: same as https://github.com/anchore/syft/pull/2001/files#r1303115339

cmd/syft/cli/options/cataloger_options.go
```diff
@@ -0,0 +1,27 @@
+package options
+
+import (
+	"fmt"
+
+	"github.com/anchore/clio"
+	"github.com/anchore/syft/syft/source"
+)
+
+type catalogerOptions struct {
```
wagoodman, line 10: nit: probably drop the `options` suffix like done elsewhere?

syft/formats/common/spdxhelpers/document_namespace.go
```diff
@@ -44,7 +43,7 @@ func DocumentNamespace(name string, src source.Description) string {
 	u := url.URL{
 		Scheme: "https",
 		Host:   "anchore.com",
-		Path:   path.Join(internal.ApplicationName, identifier),
+		Path:   path.Join(src.Name, identifier),
```
wagoodman, line 46: why use src.Name here? (seems like a non 1:1 changee)

syft/formats/common/spdxhelpers/document_namespace_test.go
```diff
@@ -24,6 +24,7 @@ func Test_documentNamespace(t *testing.T) {
 			name:      "image",
 			inputName: "my-name",
 			src: source.Description{
+				Name: "syft",
```
wagoodman, line 27: I'm not certain this is a realistic value

syft/formats/spdxtagvalue/encoder_test.go
```diff
@@ -115,7 +115,7 @@ func redactor(values ...string) testutils.Redactor {
 				`Created: .*`: "Created: redacted",
 
 				// each SBOM reports a unique documentNamespace when generated, this is not useful for snapshot testing
-				`DocumentNamespace: https://anchore.com/syft/.*`: "DocumentNamespace: redacted",
+				`DocumentNamespace: https://anchore.com/.*`: "DocumentNamespace: redacted",
```
wagoodman, line 118: are we intentionally changing the behavior to not include syft here?

syft/pkg/cataloger/golang/test-fixtures/go-sum-hashes/go.sum
wagoodman, line 1: was this update intentional?

kzantow: sorta -- this file keeps getting automatically updated by IntelliJ

syft/event/parsers/parsers.go
kzantow: This was definitely a question. Maybe a further discussion on the parsers is necessary

syft/formats/spdxtagvalue/encoder_test.go
kzantow: yes, it's a very specific field which is only set by syft when run as syft

syft/formats/common/spdxhelpers/document_namespace.go
kzantow: This isn't a 1:1 change, but if a different application is running, we probably shouldn't indicate it's _syft_, but also we should probably have a different way to ascertain a URL

cmd/syft/cli/options/packages.go
```diff
@@ -2,117 +2,167 @@ package options
...
-	Output             []string
-	OutputTemplatePath string
-	File               string
-	Platform           string
-	Exclude            []string
-	Catalogers         []string
-	SourceName         string
-	SourceVersion      string
-	BasePath           string
+type Packages struct {
```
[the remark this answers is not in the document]
kzantow: I renamed this to `Cataloging`, as it covers more than just package cataloging

wagoodman:
just a heads up: this PR introduced new config values that need to get incorporated here https://github.com/anchore/syft/pull/1734

cmd/syft/cli/commands/packages.go
```diff
@@ -0,0 +1,252 @@
...
+`
+	nonImageSchemeHelp = `    {{.appName}} {{.command}} dir:path/to/yourproject                  read directly from a path on disk (any directory)
+    {{.appName}} {{.command}} file:path/to/yourproject/file            read directly from a path on disk (any single file)
+`
+	packagesSchemeHelp = "\n  " + schemeHelpHeader + "\n" + imageSchemeHelp + nonImageSchemeHelp
+
+	packagesHelp = packagesExample + packagesSchemeHelp
+)
+
+type packagesOptions struct {
```
wagoodman, line 56: the original app config has a `ConfigPath` option. We can replicate that by adding a `Config` option which fangs automatically injects. (I think changing from ConfigPath to Config makes more sense anyway)

cmd/syft/cli/options/cataloger.go
```diff
@@ -0,0 +1,27 @@
+package options
+
+import (
+	"fmt"
+
+	"github.com/anchore/clio"
+	"github.com/anchore/syft/syft/source"
+)
+
+type Cataloger struct {
```
wagoodman, line 10: should this remain unexported since it's meant for only other options here?

cmd/syft/cli/commands/packages.go
kzantow: Why would we do this? it's already handled by fangs

cmd/syft/cli/options/cataloger.go
kzantow: yes, except it conflicts with the syft `cataloger` package. very open to naming suggestions here...

cmd/syft/cli/options/registry.go
```diff
@@ -1,24 +1,21 @@
...
-	// IMPORTANT: do not show the token in any YAML/JSON output (sensitive information)
-	Token string `yaml:"-" json:"-" mapstructure:"token"`
+	// IMPORTANT: do not show any credential information, use secret type to automatically redact the values
+	Username secret `yaml:"username" json:"username" mapstructure:"username"`
+	Password secret `yaml:"password" json:"password" mapstructure:"password"`
+	Token    secret `yaml:"token" json:"token" mapstructure:"token"`
 
-	TLSCert string `yaml:"tls-cert,omitempty" json:"tls-cert,omitempty" mapstructure:"tls-cert"`
-	TLSKey  string `yaml:"tls-key,omitempty" json:"tls-key,omitempty" mapstructure:"tls-key"`
+	TLSCert secret `yaml:"tls-cert,omitempty" json:"tls-cert,omitempty" mapstructure:"tls-cert"`
```
wagoodman, line 17: these are just paths, so they don't need to be secrets

syft/formats/common/spdxhelpers/document_namespace.go
kzantow: This should not have been _source_.Name but rather _sbom.descriptor_.Name

wagoodman (approved):
nice work 😎

PR #2011, opened 2023-08-09
1948-filter-pkg-by-type

spiffcs:
## Summary

After #1948 was merged a few more comments were made regarding certain aspects of the PR. 

This addition addresses those comments by updating the exclusion approach. We no longer filter based on the cataloger that found the pkg, but the pkg type in relation to the synthetic binary pkg type


wagoodman (approved):
thanks for the adjustments 🙏  probably should also use the `changelog-ignore` label on this PR

PR #2016, opened 2023-08-10
chore: ensure syft binary is up-to-date when running CLI tests locally

kzantow:
While attempting to debug a CLI test locally, I got dramatically different results, which took debugging to realize that these tests use syft binaries built with `make snapshot`. However, it is confusing when this happens and not especially convenient to run `make snapshot` each time a code change is made before running each tests. Additionally, `make snapshot` is not especially fast: ~17s on my local machine, so the dev-test cycle is a bit painful.

I implemented something that can build the syft binary fairly quickly (in the 2-4s range, depending on the amount changed from the last build) and also much more quickly if there are no changes -- in the < 1sec range. However, this is using the `go build` command directly, which is a bit less than ideal -- since we're using `goreleaser`, it would be nice to have this consistent.

It should be noted: in all cases, this runs **once** for each test run, whether it's a single test or the entire CLI test suite.

So I implemented a few different options for this and ran them under different scenarios (I've included build times with an up-to-date go build cache):

### Build using go directly

* Empty snapshot: 3.82s
* Up to date: 731ms

This populates the goreleaser build flags with [hopefully] obvious placeholder values.

Version output:
```
Application:     syft
Version:         VERSION
BuildDate:       DATE
GitCommit:       COMMIT
GitDescription:  SUMMARY
Platform:        darwin/amd64
GoVersion:       go1.19.5
Compiler:        gc
```

### Build using goreleaser directly

* Empty snapshot: 4.5s
* Up to date: 4.308s

Version output:
```
Application:     syft
Version:         0.86.1-SNAPSHOT-3ccbea96
BuildDate:       2023-08-10T19:55:06Z
GitCommit:       3ccbea960e41369e3c8d2c4fc578288c422925e0
GitDescription:  v0.86.1-31-g3ccbea96-dirty
Platform:        darwin/amd64
GoVersion:       go1.19.5
Compiler:        gc
```

### Build using `make snapshot`

* Empty snapshot: 16.509s
* Up to date: fails to update snapshots -- does not run because snapshot dir exists
    * If update to .PHONY: 16.82s

Version output is the same as building with goreleaser directly.

## Summary:

`make snapshot` builds _all_ binaries every time it's run. The 17-ish second turnaround would be painful to sit through when running an individual test with a dev-test cycle.

Building with `goreleaser` needs `--clean`, to remove the prior build artifact, so it rebuilds each time. ~4 seconds overhead for each dev-test cycle is a little painful, still, but not the worst.

Building with `go` directly is fast with no changes, but each change could result in each run adding approximately 4s if changing files between runs. However, testing this with a small change can still be in the <1s range:

```
    utils_test.go:310: built binary: ./snapshot/darwin-build_darwin_amd64_v1/syft in 740ms
        affected paths:
        github.com/anchore/syft/syft/cpe
```

Finally: since both `go` and `goreleaser` variants only build a single binary that matches the current platform and architecture, they _do not_ build the linux variant with any of these changes. This means the one test that runs syft on docker fails: `TestDirectoryScanCompletesWithinTimeout`. The `make snapshot` does not have this problem, since it's building _all_ platform/architectures.


test/cli/utils_test.go
```diff
@@ -271,7 +274,137 @@ func getSyftBinaryLocation(t testing.TB) string {
 		// SYFT_BINARY_LOCATION is the absolute path to the snapshot binary
 		return os.Getenv("SYFT_BINARY_LOCATION")
 	}
-	return getSyftBinaryLocationByOS(t, runtime.GOOS)
+	bin := getSyftBinaryLocationByOS(t, runtime.GOOS)
+
+	// only run on valid bin target, when not running in CI
+	if bin != "" && os.Getenv("CI") == "" {
```
kzantow, line 280: I'm unsure the best way to determine if this is being run locally. It would also be fine (for me, at least) if this only runs when running a single test.

test/cli/utils_test.go
```diff
@@ -271,7 +274,137 @@ func getSyftBinaryLocation(t testing.TB) string {
...
+}
+
+func buildSyft(t testing.TB, outfile string) {
+	dir := repoRoot(t)
+
+	start := time.Now()
+
+	var stdout, stderr string
+	var err error
+	switch "go" {
```
kzantow, line 296: Change this to try the different variants.

test/cli/utils_test.go
```diff
@@ -271,7 +274,137 @@ func getSyftBinaryLocation(t testing.TB) string {
...
+				Summary string
+			}{
+				Version: "VERSION",
+				Commit:  "COMMIT",
+				Date:    "DATE",
+				Summary: "SUMMARY",
+			})
+			break
+		}
+	}
```
kzantow, line 348: This just just for using the same `ldflags` the goreleaser build uses.

test/cli/utils_test.go
```diff
@@ -271,7 +274,137 @@ func getSyftBinaryLocation(t testing.TB) string {
...
+	case "goreleaser":
+		stdout, stderr, err = buildSyftWithGoreleaser(dir)
+	case "make":
+		stdout, stderr, err = buildSyftWithMake(dir)
+	}
+
+	took := time.Now().Sub(start).Round(time.Millisecond)
+	if err == nil {
+		if len(stderr) == 0 {
+			t.Logf("binary is up to date: %s in %v", outfile, took)
```
kzantow, line 308: Note this is not entirely accurate. This is reported if there are no code changes and the build cache hasn't changed, but the binary is rebuilt anyway.

test/cli/utils_test.go
```diff
@@ -271,7 +274,137 @@ func getSyftBinaryLocation(t testing.TB) string {
...
+}
+
+func buildSyftWithGoreleaser(dir string) (string, string, error) {
+	tmpDir := path.Join(dir, ".tmp")
+
+	goreleaserYaml := goreleaserYamlContents(dir)
+
+	// # create a config with the dist dir overridden
+	tmpGoreleaserYamlFile := path.Join(tmpDir, "goreleaser.yaml")
+	_ = os.WriteFile(tmpGoreleaserYamlFile, []byte("dist: snapshot\n"+goreleaserYaml), os.ModePerm)
```
kzantow, line 383: Emulating what happens in the Makefile

test/cli/utils_test.go
wagoodman: I like the idea of by default having the goreleaser approach done since it would be most consistent with how we're going to release. That being said, I also like that this is swappable strategy... we could simply leave this here and wire it up to an environment variable so the dev could choose between go and goreleaser (should probably removed the make approach though).

wagoodman (approved):
my vote is to get this change in, though should probably drop the `make` approach entirely and default to either go or goreleaser and have the strategy be accessible via environment variable and document it in the DEVELOPING.md.

PR #2021, opened 2023-08-11
Detect golang boring crypto and fipsonly modules

bathina2:
This PR adds functionality to Syft to derive the crypto settings of a go executable.

It imports a library that scans the symbols and determines whether an executable contains the boring crypto module as well as the fipsonly module.


bathina2:
I was hesitant as well. I would have liked to have the reader functionality merged upstream, however, the upstream project hasn't changed much in a while and we don't have real time frame. My thoughts were that we could get this PR merged in syft and once the `goversion` changes got merged upstream we could pull them in.

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -30,14 +37,42 @@ func scanFile(reader unionreader.UnionReader, filename string) ([]*debug.BuildIn
...
+
+		builds = append(builds, &ExtendedBuildInfo{bi, v})
 	}
 
 	archs := getArchs(readers, builds)
 
 	return builds, archs
 }
 
+func getCryptoInformation(reader io.ReaderAt) (string, error) {
```
wagoodman, line 55: this seems to be derived from https://github.com/rsc/goversion/blob/3a30cee7003ec5dc0903682e79b3059c8e101952/main.go#L171-L183 however, I feel that the output is a little different can be awkward:
```
boring crypto
standard crypto
boring crypto +crypto/tls/fipsonly
standard crypto +crypto/tls/fipsonly
boring AND standard crypto!!! +crypto/tls/fipsonly
```
I think it would be more straightforward with a comma delimited list of values:
```
boring
standard
boring,crypto/tls/fipsonly
standard,crypto/tls/fipsonly
boring,standard,crypto/tls/fipsonly
```

or even better, allow that to be a native string slice:
```
["boring-crypto"]
["standard-crypto"]
["boring-crypto", "crypto/tls/fipsonly"]
["standard-crypto", "crypto/tls/fipsonly"]
["boring-crypto", "standard-crypto"]
["boring-crypto", "standard-crypto", "crypto/tls/fipsonly"]
```

This last way is more machine readable.

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -6,12 +6,19 @@ import (
 	"io"
 	"runtime/debug"
 
+	version "github.com/kastenhq/goversion/version"
+
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/pkg/cataloger/internal/unionreader"
 )
 
+type ExtendedBuildInfo struct {
```
wagoodman, line 15: nit: this shouldn't be exported from this package

syft/pkg/cataloger/golang/scan_binary.go
bathina2: I agree, will do.

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -6,12 +6,19 @@ import (
 	"io"
 	"runtime/debug"
 
+	version "github.com/kastenhq/goversion/version"
```
wagoodman, line 9: nit: can be simplified
```suggestion
	"github.com/kastenhq/goversion/version"
```

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -30,14 +37,42 @@ func scanFile(reader unionreader.UnionReader, filename string) ([]*debug.BuildIn
...
 	return builds, archs
 }
 
+func getCryptoInformation(reader io.ReaderAt) (string, error) {
+	v, err := version.ReadExeFromReader(reader)
+	if err != nil {
+		return "", err
+	}
+
+	cryptoInfo := ""
```
wagoodman, line 61: it would be nice to have this kind of extraction logic into it's own function so that can be unit tested (given a `version.Version` instance assert that the correct []string value is returned)

wagoodman:
@kzantow is right that we try to keep references to upstreams before forking. That being said, I would be OK with getting the functionality under the fork then create an issue for removing the fork and tracking the upstream PR there.

wagoodman (approved):

PR #2027, opened 2023-08-14
Update github.com/Microsoft/go-rustaudit to have correct capitalization

### LINKED ISSUE #2026: Cant use syft with go mod vendor because invalid case of microsoft module
**What happened**:
in code you used github.com/microsoft/go-rustaudit instead of github.com/**Microsoft**/go-rustaudit

to use github.com/Microsoft/go-rustaudit

install syft as dependency of another go project
run
go mod tidy
go mod vendor
go build -mod vendor

**Environment**:
- Output of `syft version`:

barnuri:
fixes #2026

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/jinzhu/copier v0.3.5
 	github.com/mgutz/ansi v0.0.0-20200706080929-d51e80ef957d // indirect
 	github.com/mholt/archiver/v3 v3.5.1
-	github.com/microsoft/go-rustaudit v0.0.0-20220730194248-4b17361d90a5
+	github.com/Microsoft/go-rustaudit v0.0.0-20220730194248-4b17361d90a5
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/hashstructure/v2 v2.0.2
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/anchore/syft
 
 go 1.19
 
+replace github.com/microsoft/go-rustaudit => github.com/Microsoft/go-rustaudit v0.0.0-20220730194248-4b17361d90a5
```
kzantow, line 5: We definitely don't want to have a replace directive in the `go.mod` -- what is this needed for?

kzantow: This PR has updated the capitalization to be correct, which includes the `go.mod` entries. Why is the `replace` directive needed?

wagoodman: agreed, the replace directive is not required if we change the import in the source files that are using this module -- just pushed an update that does this

wagoodman:
I'm not certain I see the reason for this change, see https://github.com/anchore/syft/issues/2026#issuecomment-1690518945

spiffcs:
Closing this PR now as we cannot reproduce the issue on our end and have validated that the casing for this is correct at the moment - https://github.com/anchore/syft/issues/2026#issuecomment-1690518945

PR #2034, opened 2023-08-17
fix: properly parse conan ref and include user and channel

Pro:
The current implementation in syft is not parsing the user and channel from a conan ref.

Both are highly relevant for the resulting package version.

https://github.com/CycloneDX/cyclonedx-conan is parsing the user and channel and putting it in the resulting file.

With this MR the same functionality is also added to syft.


syft/pkg/cataloger/cpp/package.go
```diff
@@ -8,23 +8,70 @@ import (
...
+		cref.Revision = tokens[1]
+	} else {
+		cref.Revision = ""
+	}
 
-	if pkgName == "" || pkgVersion == "" {
+	// name and version are always given
+	tokens = strings.Split(ref, "@")
+	name_version := strings.Split(tokens[0], "/")
+	if name_version[0] == "" || name_version[1] == "" {
```
kzantow, line 49: this should probably first check `len(name_version) > 1`

syft/pkg/cataloger/cpp/package.go
```diff
@@ -8,23 +8,70 @@ import (
...
+	cref.Name = name_version[0]
+	cref.Version = name_version[1]
+	// user and channel
+	if len(tokens) == 2 && tokens[1] != "" {
+		tokens = strings.Split(tokens[1], "/")
+		cref.User = tokens[0]
+		cref.Channel = tokens[1]
+	} else {
+		cref.User = ""
+		cref.Channel = ""
```
kzantow, line 61: Unnecessary else statement

syft/pkg/cataloger/cpp/package.go
```diff
@@ -8,23 +8,70 @@ import (
...
+	if name_version[0] == "" || name_version[1] == "" {
 		return nil
 	}
+	cref.Name = name_version[0]
+	cref.Version = name_version[1]
+	// user and channel
+	if len(tokens) == 2 && tokens[1] != "" {
+		tokens = strings.Split(tokens[1], "/")
+		cref.User = tokens[0]
+		cref.Channel = tokens[1]
```
kzantow, line 58: This should also check `len(tokens) > 1` before accessing `tokens[1]`

syft/pkg/cataloger/cpp/package.go
```diff
@@ -8,23 +8,70 @@ import (
...
+
+	var cref ConanRef
+
+	// timestamp
+	tokens := strings.Split(ref, "%")
+	text := tokens[0]
+	if len(tokens) == 2 {
+		cref.Timestamp = tokens[1]
+	} else {
+		cref.Timestamp = ""
```
kzantow, line 34: This is else statement is unnecessary -- `""` is the default value for strings

syft/pkg/cataloger/cpp/package.go
```diff
@@ -8,23 +8,70 @@ import (
...
 	}
 
-	pkgName, pkgVersion := fields[0], fields[1]
+	// revision
+	tokens = strings.Split(text, "#")
+	ref = tokens[0]
+	if len(tokens) == 2 {
+		cref.Revision = tokens[1]
+	} else {
+		cref.Revision = ""
```
kzantow, line 43: Unnecessary else statement

kzantow (commented):
This looks like a great start! Would you be able to add a test with some realistic samples of conanfile and lockfile entries?

syft/pkg/cataloger/cpp/package.go
kzantow: This could still be a problem with unexpected input -- before accessing `tokens[1]` the length should be checked, since line 52 is doing another split operation

kzantow (commented):
Thanks @Pro, I think the changes to the tests look sufficient; just one minor issue that wasn't resolved

kzantow:
@Pro -- you should be able to fix the static analysis failure by running `make lint-fix`

kzantow:
@Pro it looks like the last commit was not signed-off; could you rebase against main and force push these changes with sign-off?

syft/pkg/cataloger/cpp/package.go
```diff
@@ -8,23 +8,67 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
-func newConanfilePackage(m pkg.ConanMetadata, locations ...file.Location) *pkg.Package {
-	fields := strings.Split(strings.TrimSpace(m.Ref), "/")
-	if len(fields) < 2 {
-		return nil
+type ConanRef struct {
```
kzantow, line 11: nit: this does not need to be exported, since it's only used in this package. `ConanRef` -> `conanRef`

kzantow:
Thanks for this PR @Pro -- there is one very small request to _not_ export the `ConanRef` type. That's the last hold up for this PR, I think.

Do you happen to know much about Conan lockfiles version 0.5? I've been trying to find information about the schema, but I've been unsuccessful. There is a report that the format may have changed significantly in an incompatible manner, and we'd like to make sure to support both older and newer versions.

syft/pkg/cataloger/cpp/package.go
wagoodman: I went ahead and pushed this change

wagoodman (approved):
I made the small tweak that @kzantow was referring to, otherwise it looks great! Thanks for the improvement here!

PR #2037, opened 2023-08-17
chore: more lenient java groupID lookups

kzantow:
This PR adjusts the way Java groupIDs are looked up to be:

1. pom.properties groupID
2. pom.xml groupID
3. pom.xml parent groupID
4. known groupIDs
5. something that looks like a groupID in a number of MANIFEST.MF entries


kzantow:
This PR was just a test to a change in groupID identification

Issue #1950, opened 2023-07-21
Errors when handling symlinks on Windows with syft v0.85.0


shamil-mubarakshin:
**What happened**:
We are using `syft` to create SBOM report against a `C:\` directory on Windows Servers.
Starting with `v0.85.0` a terminating error is returned when running `syft` against a directory with symlinks (file and directory symlinks pointing inside the scanned folder):
- <details>
  <summary>Output</summary>

  ```powershell
  [0000]  INFO syft version: 0.85.0
  [0000] DEBUG application config:
  verbosity: 3
  quiet: false
  output:
  - spdx-json=sbom.json
  output-template-path: ""
  file: ""
  check-for-app-update: true
  dev:
    profile-cpu: false
    profile-mem: false
  log:
    structured: false
    level: trace
    file: ""
  catalogers: []
  package:
    cataloger:
      enabled: true
      scope: Squashed
    search-unindexed-archives: false
    search-indexed-archives: true
  golang:
    search-local-mod-cache-licenses: false
    local-mod-cache-dir: ""
    search-remote-licenses: false
    proxy: ""
    no-proxy: ""
  linux-kernel:
    catalog-modules: true
  attest: {}
  file-metadata:
    cataloger:
      enabled: false
      scope: Squashed
    digests:
    - sha256
  file-classification:
    cataloger:
      enabled: false
      scope: Squashed
  file-contents:
    cataloger:
      enabled: false
      scope: Squashed
    skip-files-above-size: 1048576
    globs: []
  secrets:
    cataloger:
      enabled: false
      scope: AllLayers
    additional-patterns: {}
    exclude-pattern-names: []
    reveal-values: false
    skip-files-above-size: 1048576
  registry:
    insecure-skip-tls-verify: false
    insecure-use-http: false
    auth: []
  exclude: []
  platform: ""
  name: ""
  source:
    name: ""
    version: ""
    file:
      digests:
      - sha256
  parallelism: 1
  default-image-pull-source: ""
  base-path: ""
  
  [0000] DEBUG checking if a new version of syft is available
  [0000] DEBUG no new syft update available
  [0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
  [0000] TRACE indexing filetree path=C:\temp-folder
  [0000] TRACE indexing filetree path=C:\temp-folder\C:\temp-folder\parent.txt
  2023/07/21 10:14:47 error during command execution: 1 error occurred:
[...]

PR #2051, opened 2023-08-22
Do not double-prefix symlink paths with base directory

selzoc:
We observed that on Windows, symlinks from the `C:\` base directory such as `C:\Users` were being resolved as `C:\C:\Users`.  In this PR, we do not join the base directory to the resolved symlink if it is already there.

On Linux, having a path such as `//usr/local/bin` does not cause any issues (and the `//` is converted to `/` in `filepath.Clean`) but `C:\C:\` does not work, and is _not_ converted to `C:\` in `filepath.Clean` ([go playground example](https://go.dev/play/p/AbwQqu5hsyh))

Fixes https://github.com/anchore/syft/issues/1950 (we think)


wagoodman (dismissed):
Nice fix -- I don't see a good way to get in a test for this given that the expectations would be different for various platforms (and that can't be specified in testing since it's behind filepath package).

kzantow:
@wagoodman I'm going to give this a look through on my Windows machine

selzoc:
Great!  Yes, I couldn't see a good place to add a test for this.  I did verify that it works on a running Windows Server 2019 VM.

wagoodman:
Thanks @selzoc @kzantow 🙏

kzantow (approved):
I did debug through this and verified absolute links work on windows with this change, albeit not bounded to a base directory. Thanks a lot @selzoc 👍

PR #2055, opened 2023-08-24
feat: add bash classifier

### LINKED ISSUE #1963: Add support to detect bash binaries
**What happened**: Bash isn't detected on the official bash docker images.

Bash should be included in the returned list of packages.

`syft -q  packages library/bash:5|grep bash`

The output includes bash dependencies, but not bash itself:

`.bash-rundeps           20230615.045102   apk`

I re-checked all of the images mentioned in #1197, and this seems to be the only still non-functional.

**Environment**:
- Output of `syft version`: syft 0.85.0
- OS (e.g: `cat /etc/os-release` or similar):  "Ubuntu 20.04.5 LTS"

witchcraze:
This PR adds a binary cataloger for bash

- As I can not judge version of alpha/beta (alpha1 or alpha2) in version information, alpha/beta/rc is not desplayed

Fixes #1963

### DIFF
--- a/syft/pkg/cataloger/binary/default_classifiers.go
+++ b/syft/pkg/cataloger/binary/default_classifiers.go
@@ -269,6 +269,20 @@ var defaultClassifiers = []classifier{
 			cpe.Must("cpe:2.3:a:nginx:nginx:*:*:*:*:*:*:*:*"),
 		},
 	},
+	{
+		Class:    "bash-binary",
+		FileGlob: "**/bash",
+		EvidenceMatcher: fileContentsVersionMatcher(
+			// @(#)Bash version 5.2.15(1) release GNU
+			// @(#)Bash version 5.2.0(1) alpha GNU
+			// @(#)Bash version 5.2.0(1) beta GNU
+			// @(#)Bash version 5.2.0(1) rc4 GNU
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -269,6 +269,20 @@ var defaultClassifiers = []classifier{
...
 	},
+	{
+		Class:    "bash-binary",
+		FileGlob: "**/bash",
+		EvidenceMatcher: fileContentsVersionMatcher(
+			// @(#)Bash version 5.2.15(1) release GNU
+			// @(#)Bash version 5.2.0(1) alpha GNU
+			// @(#)Bash version 5.2.0(1) beta GNU
+			// @(#)Bash version 5.2.0(1) rc4 GNU
+			`(?m)@\(#\)Bash version (?P<version>[0-9]+\.[0-9]+\.[0-9]+)\([0-9]\) [a-z0-9]+ GNU`,
```
kzantow, line 280: Does this regex need the end bits ` [a-z0-9]+ GNU`?

wagoodman: the nice thing about having the suffix you highlighted is that it's more specific, which means we have less chance of matching something that is not bash. The possible downside is if there are other builds of bash that might not have this suffix. My vote is to leave it as specific as possible until we find evidence that we're missing other builds that don't have this suffix.

kzantow (approved):
LGTM, thanks for the contribution @witchcraze 👍

PR #2075, opened 2023-08-29
fix: correct group IDs for commons-codec, okhttp, okio, and add integration tests for Java PURL generation

willmurphyscode:
Future commits can hand fix PURLs made for particular packages, and then the map literal here can be edited.


test/integration/java_purl_test.go
```diff
@@ -0,0 +1,219 @@
...
+		purl := found[key]
+		assert.Equal(t, expectedPURL, purl, fmt.Sprintf("found wrong or missing PURL for %s want %s, got %s", key, expectedPURL, purl))
+	}
+	for key, foundPURL := range found {
+		expectedPURL := expectedPURLs[key]
+		assert.Equal(t, expectedPURL, foundPURL, fmt.Sprintf("found extra purl for %s want %s, got %s", key, expectedPURL, foundPURL))
+	}
+}
+
+func getCatalog(t *testing.T, image string) sbom.SBOM {
```
willmurphyscode, line 38: TODO: this should be converted to use the normal test fixture mechanism.

test/integration/java_purl_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"args4j@2.0.16":                                   "pkg:maven/args4j/args4j@2.0.16",
+	"asm-commons@2.2.3":                               "pkg:maven/asm-commons/asm-commons@2.2.3",
+	"asm-tree@2.2.3":                                  "pkg:maven/asm-tree/asm-tree@2.2.3",
+	"asm@2.2.3":                                       "pkg:maven/asm/asm@2.2.3",
+	"avalon-framework@4.1.3":                          "pkg:maven/avalon-framework/avalon-framework@4.1.3",
+	"bridge-method-annotation@1.2":                    "pkg:maven/com.infradna.tool/bridge-method-annotation@1.2",
+	"classworlds@1.1":                                 "pkg:maven/org.codehaus.classworlds/classworlds@1.1",
+	"cli@1.390":                                       "pkg:maven/org.jvnet.hudson.main/cli@1.390",
+	"commons-beanutils@1.8.0":                         "pkg:maven/commons-beanutils/commons-beanutils@1.8.0",
+	"commons-codec@1.2":                               "pkg:maven/org.apache.commons.codec.*/commons-codec@1.2",
```
willmurphyscode, line 84: TODO: `pkg:maven/org.apache.commons.codec.*/commons-codec@1.2` is wrong.

test/integration/java_purl_test.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// Constructed by:
+// syft anchore/test_images:java-56d52bc -o template -t /tmp/test.templ | grep 'pkg:maven' | sort | uniq >> test/integration/java_purl_test.go
+// where the template is:
+/*
+{{ range .Artifacts}}"{{.Name}}@{{.Version}}":"{{.PURL}}",
+{{ end }}
+*/
+var expectedPURLs = map[string]string{
```
willmurphyscode, line 62: TODO: we should hand-check these for other errors.

test/integration/java_purl_test.go
willmurphyscode: The Dockerfile in question is here: https://github.com/anchore/test-images/blob/main/containers/java/Dockerfile

I can probably just copy it into a test fixture.

test/integration/test-fixtures/image-test-java-purls/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM anchore/test_images:java-56d52bc
```
willmurphyscode, line 1: This is a bit of a hacky way to bring the test fixture into syft. Let me know if it's better to do something else, but I like that this pins the artifact.

wagoodman: this is the right approach, but should we also use a digest too?

test/integration/java_purl_test.go
wagoodman: want some help with this in a divide-and-conquer fashion?

wagoodman (approved):

Issue #2077, opened 2023-08-30
Virtual path changes to java cataloger causing creation of extra incorrect packages when jars are renamed


westonsteimel:
**What happened**:

The virtual path changes introduced as a part of https://github.com/anchore/syft/pull/2033 has a side effect of creating weird fake packages when a java archive is named differently.  An example is with https://repo1.maven.org/maven2/org/jvnet/hudson/main/hudson-war/2.2.1/hudson-war-2.2.1.war

If we run syft on the archive `hudson-war-2.2.1.war`, we get the results we'd expect to see with a single entry for `hudson-war` and a virtualPath of `/hudson-war-2.2.1.war` with the PURL correctly identified from the included pom.xml:

```json
{
  "id": "07d4f39cd55d02cc",
  "name": "hudson-war",
  "version": "2.2.1",
  "type": "java-archive",
  "foundBy": "java-cataloger",
  "locations": [
    {
      "path": "/hudson-war-2.2.1.war",
      "annotations": {
        "evidence": "primary"
      }
    }
  ],
  "licenses": [],
  "language": "java",
  "cpes": [
    "cpe:2.3:a:hudson-war:hudson-war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:hudson-war:hudson_war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:hudson_war:hudson-war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:hudson_war:hudson_war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:hudson-war:hudson:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:hudson:hudson-war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:hudson:hudson_war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:hudson_war:hudson:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:jvnet:hudson-war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:jvnet:hudson_war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:main:hudson-war:2.2.1:*:*:*:*:*:*:*",
    "cpe:2.3:a:main:hudson_war:2.2.1:*:*:*:*:*:*:*",
[...]

PR #2080, opened 2023-08-31
fix: in some cases, try to use pom info to guess name and version to top level jar

willmurphyscode:
Implements the following changes in the java archive parser's `discoverMainPackage()` method:
if there is exactly one pom.xml and pom.properties in the jar and if the filename is a prefix of the artifact ID on the pom.properties, prefer, in order, for name and version:
1. A non-empty string on pom.properties
2. A non-empty string on pom.xml
3. A name based on the filename of the top-level jar
4. Name info from the manifest.

Definitely open to feedback and refactoring here, but wanted to show a working fix before spending too much time making it look great.

Fixes https://github.com/anchore/syft/issues/2077

### DIFF
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -148,28 +148,8 @@ func (j *archiveParser) parse() ([]pkg.Package, []artifact.Relationship, error) 
 
 // discoverMainPackage parses the root Java manifest used as the parent package to all discovered nested packages.
 func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
-	pomPropertyMatches := j.fileManifest.GlobMatch(pomPropertiesGlob)
-	pomMatches := j.fileManifest.GlobMatch(pomXMLGlob)
-	var pomPropertiesObject pkg.PomProperties
-	var pomProjectObject pkg.PomProject
-	if len(pomPropertyMatches) == 1 && len(pomMatches) == 1 {
-		// we have exactly 1 pom.properties in the archive; assume it represents the
-		// package we're scanning
-		properties, _ := pomPropertiesByParentPath(j.archivePath, j.location, pomPropertyMatches)
-		projects, _ := pomProjectByParentPath(j.archivePath, j.location, pomMatches)
-
-		for parentPath, propertiesObj := range properties {
-			if propertiesObj.ArtifactID != "" && j.fileInfo.name != "" && strings.HasPrefix(propertiesObj.ArtifactID, j.fileInfo.name) {
-				pomPropertiesObject = propertiesObj
-				if proj, exists := projects[parentPath]; exists {
-					pomProjectObject = proj
-				}
-			}
-		}
-	}
 
 	// search and parse java manifest files
-	// TODO: do we want to prefer or check for pom files over manifest here?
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -204,6 +218,37 @@ func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
 	}, nil
 }
 
+func (j *archiveParser) guessMainPackageNameAndVersionFromPomInfo() (string, string) {
+	pomPropertyMatches := j.fileManifest.GlobMatch(pomPropertiesGlob)
+	pomMatches := j.fileManifest.GlobMatch(pomXMLGlob)
+	var pomPropertiesObject pkg.PomProperties
+	var pomProjectObject pkg.PomProject
+	if len(pomPropertyMatches) == 1 && len(pomMatches) == 1 {
```
wagoodman, line 226: it's possible that we could have only a pom.xml and no properties, I think we should account for that as well (allow for 0 to 1 pom properties)

test/integration/regression_java_virtualpath_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/source"
+)
+
+func TestWarCatalogedCorrectlyIfRenamed(t *testing.T) {
+	// install hudson-war@2.2.1 and renames the file to `/hudson.war`
+	// TODO: is this better expressed in syft/pkg/cataloger/java/archive_parser_test.go?
```
wagoodman, line 15: agreed, I'm not against having this integration test at all, so consider this comment non-blocking, however, testing specific business logic should be pushed as low as possible test-level-wise (unit is preferred for these kinds of assertions). We do have some regression tests at the integration level to test against the specific asset where we found the regression, so keeping this test here is ok.

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -186,9 +185,24 @@ func (j *archiveParser) discoverMainPackage() (*pkg.Package, error) {
...
+	if name == "" {
+		name = selectName(manifest, j.fileInfo)
+	}
+	if version == "" {
+		version = selectVersion(manifest, j.fileInfo)
+	}
 	return &pkg.Package{
-		Name:     selectName(manifest, j.fileInfo),
-		Version:  selectVersion(manifest, j.fileInfo),
+		// TODO: maybe select name should just have a pom properties in it?
```
wagoodman, line 203: or maybe rename `selectName` to something more specific like `selectNameFromManifest` (same for the version function)

wagoodman (approved):

syft/pkg/cataloger/java/archive_parser.go
willmurphyscode: Thanks for the suggestion! I think I'll do this as a follow up.

test/integration/regression_java_virtualpath_test.go
willmurphyscode: Thanks for the suggestion! I think I'll leave this here for now; it was reported as an end to end regression, and the parser was still finding the correct package, it was just also finding an incorrect extra package, so the end-to-end flow seems a good place of the test.

I'll push a commit to remove the TODO.

Issue #2069, opened 2023-08-26
Using syft as a go library to decode a syft json has incomplete data


jeason81:
When using the `formats.Decode()` function to decode a Syft generated SBOM. The `Artifacts` object has missing contents for `FileContents` and `FileLicenses`. To be exact, both return an empty `map[]`.



**What you expected to happen**:
I expected the returned value of `FileContents` to have data associated with the file and the file contents so that this could be parsed and used to identify potential security issues.

**Steps to reproduce the issue**:
```
file, err := os.Open("syft-sbom.json")
	if err != nil {
		panic(err)
	}

	sbom, format, err := formats.Decode(file)
	if err != nil {
		panic(err)
	}
	defer file.Close()

	fmt.Printf("SBOM schema: %+v\n", format)
	fmt.Printf("%+v\n", sbom.Artifacts.FileContents)
```

Results in:
```
go run main.go                                                                                                                                                                                                                                                                                                           
syft-json@10.0.1
map[]
```

However, the syft-sbom.json file contains an extracted file content (content is not sensitive, this is a default /etc/passwd file):
```
{
   "id": "d63497173ad453fb",
   "location": {
    "path": "/etc/passwd",
    "layerID": "sha256:7cd83e46b22234ac775ed33e7b0c18d697f2e124681fd0592e859d2ee17fbcd4"
   },
   "metadata": {
    "mode": 644,
    "type": "RegularFile",
    "userID": 0,
    "groupID": 0,
    "mimeType": "text/plain",
    "size": 767
   },
[...]

PR #2083, opened 2023-08-31
fix: encode and decode FileLicenses and FileContents in Syft JSON

kzantow:
This PR corrects an issue that the Syft SBOM `FileLicenses` and `FileContents` were not being encoded/decoded properly.

Fixes: #2069


syft/formats/syftjson/model/file.go
```diff
@@ -21,3 +23,18 @@ type FileMetadataEntry struct {
 	MIMEType        string `json:"mimeType"`
 	Size            int64  `json:"size"`
 }
+
+type fileLicenses []FileLicense
```
wagoodman, line 27: is it intentional that this is unexported?

wagoodman: (if there is a change, then a regen of the json schema is needed)

wagoodman (approved):

PR #2067, opened 2023-08-26
chore: bump to use go1.21

chenrui333:
fixes #2066
relates to https://github.com/Homebrew/homebrew-core/pull/140496

PR #2088, opened 2023-09-03
Bump the golang.org/x/exp dependency and fix a build breakage.

dlorenc:
This was causing a conflict for us in some code that imports syft.


spiffcs:
Thanks @dlorenc! Let me get the conflicts resolved here and I'll try and have this merged and ready for the next release ASAP

spiffcs:
Just a quick note on this PR - after we update the CI files with the correct Go version we'll still need to check current consumers of syft within our own tooling and what they look like when upgraded to `v1.21` if there are no issues then we can start upgrading each consumer to use `v1.21`. Then we can merge this PR into Syft. 

Thanks again for the contribution/fix!

spiffcs:
related PR - https://github.com/anchore/syft/pull/2067

dlorenc:
Ah whoops - I actually didn't intend for go to add the 1.21 stuff, that was a side effect of me using 1.21 to perform the change myself I think. Feel free to separate that part out if that helps.

spiffcs:
No worries - so far it's not too big of a lift. If I run into complications I'll be sure to separate this out as it's own separate thing to make sure it get's resolved and is not blocked by the update

spiffcs (approved):

PR #2091, opened 2023-09-04
fix(cdx): validate external refs before encoding

### LINKED ISSUE #2079: Cyclonedx external reference URLs are not validated when encoding
**What happened**:
When analyzing `docker.io/anchore/test_images:grype-quality-node-d89207b@sha256:f56164678054e5eb59ab838367373a49df723b324617b1ba6de775749d7f91d4` and outputing to cyclonedx I'm seeing values that are not valid URLs in the `externalReferences` section:
``` bash
# cat node-87.1.cxd.json| jq '.components[].externalReferences'
...
[
  {
    "url": "jsdom/webidl-conversions",
    "type": "distribution"
  }
]
[
  {
    "url": "jsdom/whatwg-url",
    "type": "distribution"
  }
]
[
  {
    "url": "git://github.com/isaacs/node-which.git",
    "type": "distribution"
  }
]
[
  {
    "url": "git://github.com/isaacs/node-which.git",
    "type": "distribution"
  }
]
...
```

Which results in invalid cyclonedx relative to the validator:
```bash
cyclonedx validate --input-file ./node-87.1.cxd.json  --input-version v1_4

Validating JSON BOM...
Validation failed: Value does not match format "iri-reference"
#/properties/components/items/$ref/properties/externalReferences/items/$ref/properties/url/format
BOM is not valid.
```

URLs encoded should be compliant relative to where they were parsed, so for node cataloging https://github.com/jsdom/webidl-conversions/blob/master/package.json#L15 will result in a repository value of `jsdom/webidl-conversions`. For node, this should be interpreted as `github.com/jsdom/webidl-conversions`, for other package types these assumptions might not be valid or have different assumptions that can be made about the URL.

Ultimately https://github.com/anchore/syft/blob/cfebae27f5216d042b87b7508fcf39086c6b0338/syft/formats/common/cyclonedxhelpers/external_references.go#L15 the cyclonedx external reference encoder will need to take this into consideration.

**Environment**:
Syft version v0.87.1

hainenber:
Fixes #2079

## PR description
Validate extracted metadata before encoding into output file as `external_references`.

## Notes
- [x] Test added

### DIFF
--- a/syft/formats/common/cyclonedxhelpers/external_references.go
+++ b/syft/formats/common/cyclonedxhelpers/external_references.go
@@ -2,6 +2,7 @@ package cyclonedxhelpers
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/CycloneDX/cyclonedx-go"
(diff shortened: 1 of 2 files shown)


kzantow (approved):
Thanks for the contribution @hainenber 👍

PR #2097, opened 2023-09-05
Expose cobra command in cli package

wagoodman:
This allows for consuming applications to use the cobra command directly, which allows for embedding the syft cli into another cli easily. We might do this in the future up in clio but decided to hold off on that for now (related PR https://github.com/anchore/clio/pull/26 ).


kzantow (approved):
👍

Issue #2072, opened 2023-08-28
Add User-Agent with Syft version during update check


kzantow:
We should add a `User-Agent` header with Syft version to the version update check.

kzantow:
Sure thing, done @hainenber !

PR #2100, opened 2023-09-06
feat(cmd/update): add UA header with current ver when check for update

hainenber:
Implements #2072 

## PR description
When check for version update during `syft` commands, specify UA header with current version of the binary.

## Notes
- [x] Test updated

### DIFF
--- a/cmd/syft/cli/commands/update.go
+++ b/cmd/syft/cli/commands/update.go
@@ -38,7 +38,7 @@ func applicationUpdateCheck(id clio.Identification, check *options.UpdateCheck) 
 
 func checkForApplicationUpdate(id clio.Identification) {
 	log.Debugf("checking if a new version of %s is available", id.Name)
-	isAvailable, newVersion, err := isUpdateAvailable(id.Version)
+	isAvailable, newVersion, err := isUpdateAvailable(id)
 	if err != nil {
 		// this should never stop the application
 		log.Errorf(err.Error())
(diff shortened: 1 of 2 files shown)


cmd/syft/cli/commands/update.go
```diff
@@ -89,11 +89,12 @@ func isProductionBuild(version string) bool {
 	return true
 }
 
-func fetchLatestApplicationVersion() (*hashiVersion.Version, error) {
+func fetchLatestApplicationVersion(currentVersion string) (*hashiVersion.Version, error) {
 	req, err := http.NewRequest(http.MethodGet, latestAppVersionURL.host+latestAppVersionURL.path, nil)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create request for latest version: %w", err)
 	}
+	req.Header.Add("User-Agent", fmt.Sprintf("%v %v", "Syft", currentVersion))
```
kzantow, line 97: I think we should instead modify this function to take the `id clio.Identification` as a parameter (and, by association the `isUpdateAvailable` function), then this line would be something like:
```
req.Header.Add("User-Agent", fmt.Sprintf("%v %v", id.Name, id.Version))
```

hainenber: Thanks for the prompt review! Sounds reasonable to me, I'll try pushing new commits to amend this

hainenber: Done pushing. PTAL again. Thanks!

kzantow (approved):
Thanks for the contribution @hainenber!

PR #2111, opened 2023-09-08
Fix: Correctly parse the right version of SPDX JSON in `Identify`

jamietanna:


spiffcs:
Thanks for adding this test case and starting on this work @jamietanna - I think this work might actually be addressed by #2186 so I'm linking this here for convenience

wagoodman:
https://github.com/anchore/syft/pull/2186 refactored this path and also added tests to ensure all supported versions are decoded and identified correctly.

Issue #2046, opened 2023-08-21
remove redundant warning regarding power user and dir


tomerse-sg:
**What happened**:
when I run syft power-user I see the below message:
![image](https://github.com/anchore/syft/assets/110396524/6e1048c2-ce7d-43dd-aa84-b2dca93b2417)
it is specified that dir is not supported which is not true
**What you expected to happen**:
remove this message
**Steps to reproduce the issue**:
run syft power-user
**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: 0.84.1
- OS (e.g: `cat /etc/os-release` or similar): macOS

kzantow:
NOTE: this is in the _help text_.

PR #2113, opened 2023-09-10
fix the help output of power-user

tomersein:
Cloeses #2046 
the stdout of power-user --help specifies that it doesn't support fs, which is wrong.
removed the line.


kzantow (approved):
Thanks @tomersein !

PR #2123, opened 2023-09-12
wip: add cyclonedx schema version selection

spiffcs:


syft/formats/cyclonedxxml/format.go
```diff
@@ -7,14 +7,66 @@ import (
 	"github.com/anchore/syft/syft/sbom"
 )
 
-const ID sbom.FormatID = "cyclonedx-xml"
+const ID sbom.FormatID = "cyclonedx"
```
wagoodman, line 10: I think this is a consequential change, can you expound on this a bit? (I think there is an alias for cyclonedx already, which I want to say defaults to json, but I haven't verified this)

wagoodman: actually, this doesn't seem right: this is changing the ID for this format to something more generic, but the ID should be as specific as possible. SPDX does this by having a list of IDs, where the first is the ID and the remaining are unique aliases https://github.com/anchore/syft/blob/9de4129638a60dbef0c87b964bdcee561b0ee12a/syft/formats/spdxtagvalue/format.go#L9

Then that list is used in the format object https://github.com/anchore/syft/blob/9de4129638a60dbef0c87b964bdcee561b0ee12a/syft/formats/spdxtagvalue/format.go#L18

This PR seems to be reversing this, where the aliases are more specific 

https://github.com/anchore/syft/pull/2123/files#diff-353e0a54e1d8afae37163dec0733b3e941580dc79824b40b9bf35e649efa28b0R20

Since all of these instances are the same can you make a list and share that similar to SPDX?

syft/formats/formats.go
```diff
@@ -71,6 +81,13 @@ func ByNameAndVersion(name string, version string) sbom.Format {
 	for _, f := range Formats() {
 		for _, n := range f.IDs() {
 			if cleanFormatName(string(n)) == name && versionMatches(f.Version(), version) {
+				if version == sbom.AnyVersion && strings.Contains(string(n), "cyclonedx") {
```
wagoodman, line 84: This is looking for the string `cyclonedx` but we might introduce an alias like `cdx`. I think this should more formally ask the format for the latest supported version. 

At the same time, I feel that this format object is about to be rethought anyway. I think there are strong signs to split up for encoding vs decoding use cases, which means trying to settle this detail in this PR wouldn't be worth it. I think it's worth at least adding a comment here about the limitations of this approach and a TODO to refactor later.

syft/formats/cyclonedxxml/format.go
spiffcs: Ahh - I think this was just an unexpected change on my part thanks for the catch - I looked back at the diff and the original format had the alias of `cyclonedx`. The original ID was `cyclonedx-xml`. I think a unit test was failing and to fix it I put the wrong alias back in the wrong spot. I'll fix this now.

syft/formats/cyclonedxxml/format.go
```diff
@@ -9,12 +9,64 @@ import (
...
+func Format1_5() sbom.Format {
 	return sbom.NewFormat(
-		sbom.AnyVersion,
-		encoder,
+		cyclonedx.SpecVersion1_5.String(),
+		encoderV1_5,
 		cyclonedxhelpers.GetDecoder(cyclonedx.BOMFileFormatXML),
 		cyclonedxhelpers.GetValidator(cyclonedx.BOMFileFormatXML),
-		ID, "cyclonedx", "cyclone",
+		ID, "cyclonedx", "cyclone-xml",
```
kzantow, line 70: I think this should remain `cyclone`. `cyclone-xml` could be added, though.

syft/formats/cyclonedxjson/format.go
```diff
@@ -9,10 +9,62 @@ import (
 
 const ID sbom.FormatID = "cyclonedx-json"
 
-func Format() sbom.Format {
+var Format = Format1_4
+
+func Format1_0() sbom.Format {
```
kzantow, line 14: It might be nice to programmatically build these formats, something like this could live in the `common/cyclonedxhelpers` package:
```go
func Formats(format cyclonedx.BOMFileFormat, ids ...sbom.FormatID) (out []sbom.Format) {
  versions := []cyclonedx.SpecVersion{
    cyclonedx.SpecVersion1_0,
    cyclonedx.SpecVersion1_1,
    cyclonedx.SpecVersion1_2,
    cyclonedx.SpecVersion1_3,
    cyclonedx.SpecVersion1_4,
    cyclonedx.SpecVersion1_5,
  }
  for i := range versions {
    version := versions[i]
    encoder := func(output io.Writer, s sbom.SBOM) error {
      return encodeCycloneDX(output, s, format, version)
    }
    if version == cyclonedx.SpecVersion1_4 {
      out = append(out, sbom.NewFormatDefault(
		version.String(),
		encoder,
		cyclonedxhelpers.GetDecoder(format),
		cyclonedxhelpers.GetValidator(format),
		ids...,
	))
    } else {
      out = append(out, sbom.NewFormat(
		version.String(),
		encoder,
		cyclonedxhelpers.GetDecoder(format),
		cyclonedxhelpers.GetValidator(format),
		ids...,
	))
    }
  }
}
```

This way there only needs to be a single `encodeCycloneDX()` function and adding new versions is just updating the list, which would update both XML and JSON variants. E.g. the XML variant would have:

```go
func Formats() []sbom.Format {
  return cyclonedxhelpers.Formats(cyclonedx.BOMFileFormatXML, ID, "cyclonedx", "cyclone")
}
```

syft/formats/formats.go
kzantow: I would probably like to add a `Default` field. At some point in the future, when we support SPDX 3, we will probably also want to continue to default to SPDX 2.3.

syft/formats/cyclonedxjson/format.go
spiffcs: This is a good suggestion, but I think we wanted to save this work for when we go back and take a look at Formats before syft v1.0. The purpose of this PR was to just keep this as close to the same pattern as we have with SPDX. Getting creative here and trying to build these programmatically (different from the current SPDX pattern) would eventually get swallowed by that new work.

syft/formats/cyclonedxxml/format.go
spiffcs: ^ Yea this was an accidental flip and has been changed back so cyclonedx-xml is the primary ID and all aliases are the same as the original format object that was split into the 5 versions

kzantow: This is changing the alias of `cyclone` to `cyclone-xml`. E.g. `syft thing -o cyclone` will no longer work

spiffcs: Thanks! Yea definitely want to keep this as it was in the original format. No alias additions or changes should be part of this PR. Thanks for the catch.

syft/formats/formats.go
spiffcs: I can follow this PR up with a stab at that. I really want to keep this PR small and focused on integrating the new cyclonedx library. Other changes on how we handle formats, their defaults or how the latest is selected can be their own change set. The reason this line had to be included was because we don't necessarily want to be on latest for cyclonedx at this moment. Rather than rewrite ByNameAndVersion into a totally new thing I thought a simple edge case with comment would be the correct change here

syft/formats/cyclonedxxml/format.go
spiffcs: @wagoodman this should now be reverted to where there are no ID changes and aliases are consistent across the new formats

syft/formats/cyclonedxxml/format.go
spiffcs: This should now be in the correct state where no aliases are changed and ID/Alias are consistent with main

syft/formats/formats.go
```diff
@@ -71,6 +81,16 @@ func ByNameAndVersion(name string, version string) sbom.Format {
 	for _, f := range Formats() {
 		for _, n := range f.IDs() {
 			if cleanFormatName(string(n)) == name && versionMatches(f.Version(), version) {
+				// if the version is not specified and the format is cyclonedx, then we want to return the most recent version up to 1.4
+				// If more aliases like cdx are added this will not catch those - we want to eventually provide a way for
+				// formats to inform this function what their default version is
+				// TODO: remove this check when 1.5 is stable or default formats are designed. PR below should be merged.
+				// https://github.com/CycloneDX/cyclonedx-go/pull/90
+				if version == sbom.AnyVersion && strings.Contains(string(n), "cyclonedx") {
```
kzantow, line 89: This check should probably be `strings.Contains(string(n), "cyclone")` to use the same default when `-o cyclone` is used.

kzantow (approved):

syft/formats/cyclonedxjson/format.go
wagoodman: agreed to the both of you -- it's a good suggestion, but it would be better to be consistent with the current pattern first, then refactor to "the next" pattern.

wagoodman (approved):

PR #2138, opened 2023-09-15
chore: pin and update all workflow dependencies; add permission scopes

spiffcs:
## Summary

This PR takes a few steps on the github actions side of our code to protect against potential security threats.

- Updates All Github Actions to be pinned by sha (please double check there are no shortened commit sha here)
- Updates container images to be pinned by sha256
- Updates top level security permissions of workflows to be default `READ` (writes can still be scoped to individual steps)

### Methodology
For the Audit I used https://github.com/ossf/scorecard#basic-usage with the `--show-details` flag. The sections I looked at for this PR were:

[Dangerous-Workflow](https://github.com/ossf/scorecard/blob/main/docs/checks.md#dangerous-workflow)
[Pinned-Dependencies](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)
[Token-Permissions](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions)

### Unofficial Actions we use (not managed by github) to consider keeping, moving, or changing from
https://github.com/fountainhead/action-wait-for-check
https://github.com/tibdex/github-app-token/releases/tag/v2.0.0


.github/actions/bootstrap/action.yaml
```diff
@@ -24,13 +25,13 @@ inputs:
...
   steps:
-    - uses: actions/setup-go@v4
+    - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe #v4.1.0
       with:
         go-version: ${{ inputs.go-version }}
 
     - name: Restore tool cache
       id: tool-cache
-      uses: actions/cache@v3
+      uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 #v3.3.2
```
willmurphyscode, line 34: It would be nice if there was a mechanism besides the comment; does any automated process ensure that the version in the comment and the version identified by the digest are the same?

spiffcs: ^ Dependabot will actually update the commit and the commented version - let me find the PR that shows that example

spiffcs: Grype example where the sha and version comment are updated:
https://github.com/anchore/grype/pull/1481/files

.github/workflows/codeql-analysis.yml
```diff
@@ -14,11 +14,18 @@ on:
...
+permissions:
+  contents: read
+
 jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
 
+    permissions:
+      # required for all workflow
```
willmurphyscode, line 26: I don't quite understand this comment. Do you mean `all workflows`? Is there an "all" workflow?

test/integration/test-fixtures/image-go-bin-arch-coverage/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:latest as builder
+FROM golang:latest@sha256:cffaba795c36f07e372c7191b35ceaae114d74c31c3763d442982e3a4df3b39e as builder
```
willmurphyscode, line 1: What does this do when `golang:latest` moves?

test/integration/test-fixtures/image-java-virtualpath-regression/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:latest@sha256:7144f7bab3d4c2648d7e59409f15ec52a18006a128c733fcff20d3a4a54ba44a
```
willmurphyscode, line 1: Same question: will this break if `alpine:latest` moves?

test/integration/test-fixtures/image-go-bin-arch-coverage/Dockerfile
spiffcs: That's a good question - I think this fails - let me pin this to a strict version

test/integration/test-fixtures/image-java-virtualpath-regression/Dockerfile
spiffcs: Same response for golang - I think we pin to a more specific version - I'm not sure if dockerhub keeps a running log of `latest` or if it updates it in place

test/integration/test-fixtures/image-os-binary-overlap/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:latest@sha256:7144f7bab3d4c2648d7e59409f15ec52a18006a128c733fcff20d3a4a54ba44a
```
willmurphyscode, line 1: Another sha + `latest`. (Sorry for not marking all of them in the first round of comments.)

I wonder if it makes sense to just us `alpine@sha...`? (Maybe I should have done that in #2137 as well)

.github/workflows/codeql-analysis.yml
spiffcs: Sorry - this was lifted from their docs - "All" I think workflows and job might be conflated here. 
https://github.com/github/codeql-action/blob/main/.github/workflows/codeql.yml

You can see where they dogfood their own action how the security events `write` permission is set for build and version checks.

test/integration/test-fixtures/image-rust-auditable/Dockerfile
```diff
@@ -1,2 +1,2 @@
 # An image containing the example hello-auditable binary from https://github.com/Shnatsel/rust-audit/tree/master/hello-auditable
-FROM docker.io/tofay/hello-rust-auditable:latest
+FROM docker.io/tofay/hello-rust-auditable:latest@sha256:1d35d1e007180b3f7500aae5e27560697909132ca9a6d480c4c825534c1c47a9
```
willmurphyscode, line 2: Another `latest` + sha.

.github/workflows/codeql-analysis.yml
spiffcs: I've removed the comment since you're right it is confusing using the "workflow" noun here

test/integration/test-fixtures/image-os-binary-overlap/Dockerfile
spiffcs: I think it's ok to specify the version here - it doesn't make a different on the Docker side of things and communicates more information to future reads what the exact version is that this sha pins to so they don't have to hunt for that.

willmurphyscode (approved):

wagoodman (approved):

PR #2140, opened 2023-09-15
Add GitHub actions and shared workflow usage catalogers

### LINKED ISSUE #1896: Support cataloging github workflow & github action usages
## What would you like to be added

I would like that `syft` was able to index `github workflow` & `github action` file to list github action dependencies when generating a SBOM.

I think 2 new type could be added:

- `github-actions`: A step that is executing during a job.

  <https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses>

- `github-worfklow`: A workflow that is call by another workflow

  <https://docs.github.com/en/actions/using-workflows/reusing-workflows>
  <https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses>

## Why is this needed

I think that pipeline should also be integrated in the SBOM generation as they're dependencies for a project (e.g: If a pipeline build the final application, generate files, ...).

### For `github action`:

```yaml
steps:
  # Reference a specific commit
  - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
  # Reference the major version of a release
  - uses: actions/checkout@v3
  # Reference a specific version
  - uses: actions/checkout@v3.2.0
  # Reference a branch
  - uses: actions/checkout@main
```

`Syft` would have generated the following output:

| Name               | Version                                  | Type           |
| ------------------ | ---------------------------------------- | -------------- |
| actions/checkout   | 8f4b7f84864484a7bf31766abe9204da3cbe65b3 | github-actions |
| actions/checkout   | v3                                       | github-actions |
| actions/checkout   | v3.2.0                                   | github-actions |
| actions/checkout   | main                                     | github-actions |

### For `github workflow`

```yaml
jobs:
  call-workflow-1-in-local-repo:
    uses: octo-org/this-repo/.github/workflows/workflow-1.yml@172239021f7ba04fe7327647b213799853a9eb89
  call-workflow-2-in-local-repo:
    uses: ./.github/workflows/workflow-2.yml
  call-workflow-in-another-repo:
    uses: octo-org/another-repo/.github/workflows/workflow.yml@v1
```

`Syft` would have parsed:

| Name                                                 | Version | Type |
| ---------------------------------------------------- | ---------------------------------------- | ---------------- |
| octo-org/this-repo/.github/workflows/workflow-1.yml  | 172239021f7ba04fe7327647b213799853a9eb89 | github-workflows |
| ./.github/workflows/workflow-2.yml                   | checksum of the local file :thinking: ?  | github-workflows |
| octo-org/another-repo/.github/workflows/workflow.yml | v1                                       | github-workflows |

wagoodman:
This PR adds catalogers that will look for usages of GitHub Action within workflow and composite action files as well as shared workflow calls within workflows. Note: this does NOT catalog github action yaml files that describe a GitHub Action itself (an action.yaml file in the root of a repo housing a GitHub action itself).

Two catalogers were added:
- `github-actions-usage-cataloger`: parses step usage statements, raising up the new `pkg.GithubActionPkg` packages
- `github-action-workflow-usage-cataloger` parses job usage statements, raising up the new `pkg.GithubActionWorkflowPkg` packages

For each new kind of package, `github` type pURLs are used, e.g.:
- `pkg:github/actions/setup-go@v4`
- `pkg:github/actions/cache@v3#restore`
- `pkg:github/octo-org/another-repo@v1#.github/workflows/workflow.yml`
- `pkg:github/octo-org/this-repo@172239021f7ba04fe7327647b213799853a9eb89#.github/workflows/workflow-1.yml`

Note: this will only apply to directory scans, not image scans.

Deferred work: 
- resolving a version of a local action (e.g. .github/actions/*/action.yaml). Right now the version has been left blank, however the issue suggests to use the hash of the file. I'm not certain that makes sense in all cases, but could be a good start. Until there is more discussion on the correct value, I've decided to leave version blank in this case.

Closes #1896


syft/pkg/cataloger/cataloger.go
```diff
@@ -46,6 +47,8 @@ func ImageCatalogers(cfg Config) []pkg.Cataloger {
 		binary.NewCataloger(),
 		deb.NewDpkgdbCataloger(),
 		dotnet.NewDotnetPortableExecutableCataloger(),
+		githubactions.NewActionUsageCataloger(),
+		githubactions.NewWorkflowUsageCataloger(),
```
kzantow, line 51: Are images a likely source for github actions?

kzantow (approved):
Nothing blocking, but just question about this running for _images_ rather than just source

syft/pkg/cataloger/cataloger.go
wagoodman: I figured there wasn't harm here, and though that's true, we try to only reference installed software for image catalogers, which this doesn't fit the bill. I'm going to back that out (keep it only to dir scans)

PR #2142, opened 2023-09-16
Benji/javascript deps

noqcks:


test/integration/node_packages_test.go
```diff
@@ -13,6 +13,8 @@ func TestNpmPackageLockDirectory(t *testing.T) {
 	sbom, _ := catalogDirectory(t, "test-fixtures/npm-lock")
 
 	foundPackages := internal.NewStringSet()
+	// root pkg
+	foundPackages.Add("npm-lock")
```
noqcks, line 17: note: new top level root pkg being added

test/integration/catalog_packages_cases_test.go
```diff
@@ -144,6 +144,8 @@ var dirOnlyTestCases = []testCase{
 		pkgType:     pkg.NpmPkg,
 		pkgLanguage: pkg.JavaScript,
 		pkgInfo: map[string]string{
+			"yarn":              "0.0.0",
```
noqcks, line 147: note: new top level root pkg being added

syft/pkg/cataloger/javascript/parser/yarn/parse.go
```diff
@@ -0,0 +1,313 @@
...
+		pkg.Resolved, err = getResolved(line)
+		return err
+	case strings.HasPrefix(line, "dependencies:"):
+		deps := parseDependencies(scanner)
+		pkg.Dependencies = deps
+		return nil
+	default:
+		return nil
+	}
+}
```
nikpivkin, line 234: Wouldn't it be cleaner if using a named return value?

```suggestion

func handleLinePrefixes(line string, pkg *PkgRef, scanner *LineScanner) (err error) {
	switch {
	case strings.HasPrefix(line, "version"):
		pkg.Version, err = getVersion(line)
	case strings.HasPrefix(line, "integrity"):
		pkg.Integrity, err = getIntegrity(line)
	case strings.HasPrefix(line, "resolved"):
		pkg.Resolved, err = getResolved(line)
	case strings.HasPrefix(line, "dependencies:"):
		pkg.Dependencies = parseDependencies(scanner)
	}
	return
}
```

nikpivkin (commented):
@noqcks Do you mind if I leave a couple comments?

noqcks:
@wagoodman I actually had some things to cleanup in this PR. I saw you just self-assigned. Mind if I push some commits over the next hour or so and then let you know when it's ready?

wagoodman:
have at it!  (and just to be transparent we assign internally to track who's taking point on reviewing on the community github project)

noqcks:
@wagoodman ready now

wagoodman (commented):
I've got a rebased version of this branch locally that I'm reviewing -- I can push up my changes (might not be today though)

syft/pkg/cataloger/javascript/cataloger.go
```diff
@@ -20,3 +20,13 @@ func NewLockCataloger() *generic.Cataloger {
 		WithParserByGlobs(parseYarnLock, "**/yarn.lock").
 		WithParserByGlobs(parsePnpmLock, "**/pnpm-lock.yaml")
 }
+
+// NewJavaScriptCataloger returns a new JavaScript cataloger object based on detection
+// of npm based packages and lock files to provide a complete dependency graph of the
+// packages.
+func NewJavaScriptCataloger() *generic.GroupedCataloger {
+	return generic.NewGroupedCataloger("javascript-cataloger").
+		WithParserByGlobColocation(parseJavaScript, "**/yarn.lock", []string{"**/package.json", "**/yarn.lock"}).
```
wagoodman, line 29: I think there is a fundamental issue with the cataloger changes here. You can specify a set of primary globs and a secondary set of required globs, but there isn't a way to say that the secondary results must come from a sibling path relative to the primary path. Instead the secondary match can be written relative to the root of the source.

In this case you could have a yarn.lock in a subdirectory `/opt/place/yarn.lock` and an unrelated javascript project in `/home/myuser/code/` that container `package.json` and `/yarn.lock` files... I think this cataloger would relate those findings together, mixing these two dependency trees together (which is not desirable)

syft/pkg/cataloger/javascript/cataloger.go
```diff
@@ -20,3 +20,13 @@ func NewLockCataloger() *generic.Cataloger {
 		WithParserByGlobs(parseYarnLock, "**/yarn.lock").
 		WithParserByGlobs(parsePnpmLock, "**/pnpm-lock.yaml")
 }
+
+// NewJavaScriptCataloger returns a new JavaScript cataloger object based on detection
+// of npm based packages and lock files to provide a complete dependency graph of the
+// packages.
+func NewJavaScriptCataloger() *generic.GroupedCataloger {
```
wagoodman, line 27: I think this would need to be reconciled with the `NewLockCataloger()` above. At surface glance this cataloger would replace the lock cataloger.

syft/pkg/cataloger/javascript/cataloger.go
noqcks: This should return a groupedParser for each of those locations

One for the yarn.lock and package.json at /opt/place/ and one for the package.json and yarn.lock at /home/myuser/code.

We generate a list of colocated files with `collectColocatedFiles` and then we determine if we have both the primary and secondary file in `isAllGlobsMatched`

syft/pkg/cataloger/javascript/cataloger.go
noqcks: yep, that's true, was looking for guidance here on how to deprecate an existing cataloger on favor of a new cataloger and what the process would be for that

syft/pkg/cataloger/javascript/cataloger.go
wagoodman: ahh, I see now -- thanks for the clarification 👍

syft/pkg/cataloger/javascript/cataloger.go
wagoodman: I've just picked this back up for re-review today. I'm going to try and help get the PR rebased and working again, so I'll be pushing some commits. I'll get back to this question while I'm trying to square it with the current state on main.

noqcks:
dont have the time to work on this anymore and get it over the line, but someone else can pick it up!

will leave it in your hands @wagoodman

henrysachs:
hey the efforts done here look quite nice. Would this be something that would be nice to rebase and pickup again?

kunaals:
hey all! i'm adding Syft support to Cartography here: https://github.com/cartography-cncf/cartography/pull/2345

i'd love to pickup this issue - i'll submit something in the next couple days if folks are open to it

kunaals:
opened #4618, it doesn't cover the full scope of this PR but addresses part of the problem: getting npm dependency relationships into container image scans by wiring up `node_modules/.package-lock.json` (the hidden lockfile npm v7+ creates) to the existing V3 parser

PR #2143, opened 2023-09-17
Parse donet dependency trees

noqcks:
Adds dependency information for the dotnet ecosystem

This allows one to see the relationships between packages such as with the dependencies field in CycloneDX https://cyclonedx.org/docs/1.5/json/#dependencies.


noqcks:
it looks like there's two completely unrelated unit tests that are failing on master and somehow failed in this PR `TestParseConanlock` and `TestAllLayersResolver_AllLocations`

tgerla:
Hey @noqcks, thanks much! We will need a little time to review this but we'll do that as soon as we can, and we'll let you know if we need anything from you. Thanks again for the patch.

wagoodman (approved):
I only made a few adjustments:
- add some comments to the test capturing the expectations
- unexported the RelationshipLess function and added caveat doc comments for the SortRelationships function (it will most likely be replaced in the future since there are limitations with the current implementation, however, the current function is necessary).

overall, nice work! 🙌

Issue #2145, opened 2023-09-17
handle CycloneDX application componentType


noqcks:
**What would you like to be added**:

For this PR #2143 dependency tree parsing is being added for .NET dependencies. This creates a root pkg for the application. The problem I found that the componentType being output in CycloneDX for the root pkg is a library, while it should be an application.

[These are the lines](https://github.com/anchore/syft/blob/main/syft/formats/common/cyclonedxhelpers/component.go#L30-L33) that handle the componentType creation for the CycloneDX format.

Here's an example of what `TestLibrary.deps.json` looks like for a dotnet project in the syft test-fixtures showing that certain deps would be a `application` in the CycloneDX format, while others would be a `library`.

```
  "libraries": {
    "TestLibrary/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },
    "AWSSDK.Core/3.7.10.6": {
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-kHBB+QmosVaG6DpngXQ8OlLVVNMzltNITfsRr68Z90qO7dSqJ2EHNd8dtBU1u3AQQLqqFHOY0lfmbpexeH6Pew==",
      "path": "awssdk.core/3.7.10.6",
      "hashPath": "awssdk.core.3.7.10.6.nupkg.sha512"
    },
```

How should we think about this? Should there be an additional field in pkg.Package?

There is already a Type being set on a dotnet pkg, but this doesn't abstract well to the CycloneDX format.
```
	DotnetPkg               Type = "dotnet"
```

**Why is this needed**:

Accurate representation of dependency trees in CycloneDX.

kzantow:
I think the best solution here is to add the appropriate field to the DotnetDeps metadata instead of the main syft package, here: https://github.com/anchore/syft/blob/main/syft/pkg/dotnet.go#L4-L10

PR #2146, opened 2023-09-17
WIP: Add component type to pkg

noqcks:
See issue https://github.com/anchore/syft/issues/2145

This PR is failing right now because adding a new field to Pkg recomputes packageIDs (I believe). If this PR were to be deemed appropriate, I would update all tests where needed. Looking for an approval first before I update all the tests.

### DIFF
--- a/syft/pkg/package.go
+++ b/syft/pkg/package.go
@@ -17,18 +17,19 @@ import (
 // Package represents an application or library that has been bundled into a distributable format.
 // TODO: if we ignore FoundBy for ID generation should we merge the field to show it was found in two places?
 type Package struct {
-	id           artifact.ID      `hash:"ignore"`
-	Name         string           // the package name
-	Version      string           // the version of the package
-	FoundBy      string           `hash:"ignore" cyclonedx:"foundBy"` // the specific cataloger that discovered this package
-	Locations    file.LocationSet // the locations that lead to the discovery of this package (note: this is not necessarily the locations that make up this package)
-	Licenses     LicenseSet       // licenses discovered with the package metadata
-	Language     Language         `hash:"ignore" cyclonedx:"language"` // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
-	Type         Type             `cyclonedx:"type"`                   // the package type (e.g. Npm, Yarn, Python, Rpm, Deb, etc)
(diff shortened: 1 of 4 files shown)


tgerla:
Hey @noqcks, sorry for the delay getting back to you. I will see if we can get a review on this soon, if you are still interested in working on it. Much appreciated!

noqcks:
Yep, I am!

syft/pkg/package.go
```diff
@@ -17,18 +17,19 @@ import (
...
-	MetadataType MetadataType     `cyclonedx:"metadataType"`           // the shape of the additional data in the "metadata" field
-	Metadata     interface{}      // additional data found while parsing the package source
+	id            artifact.ID      `hash:"ignore"`
+	Name          string           // the package name
+	Version       string           // the version of the package
+	FoundBy       string           `hash:"ignore" cyclonedx:"foundBy"` // the specific cataloger that discovered this package
+	Locations     file.LocationSet // the locations that lead to the discovery of this package (note: this is not necessarily the locations that make up this package)
+	Licenses      LicenseSet       // licenses discovered with the package metadata
+	Language      Language         `hash:"ignore" cyclonedx:"language"` // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
+	ComponentType ComponentType    `cyclonedx:"type"`                   // the type of component (e.g. application, library, framework, etc)
```
kzantow, line 27: Hey @noqcks -- sorry for the delay here, I also left a comment on the related issue, but this gist is that instead of modifying this package, we should modify the DotnetDepsMetadata with the appropriate field(s) here: https://github.com/anchore/syft/blob/main/syft/pkg/dotnet.go#L4-L10

noqcks: are we able to access DotnetDepsMetadata from the cyclonedx formatter? from my understanding you can only access the package Struct

kzantow: Yes, you can access the metadata from the formatter, you just need to type assert the metadata field [here is an example extracting the author in CycloneDX](https://github.com/anchore/syft/blob/main/syft/format/internal/cyclonedxutil/helpers/author.go#L10-L32)

wagoodman:
I don't think we want to have a cyclonedx specific field in the core type when there isn't any clear indication from the catalogers what the correct answer is in most cases. At a higher-level I do agree in cases where we're converting from a cyclonedx doc we should be attempting to preserve this information, but I think it should consider more than just these fields.

PR #2154, opened 2023-09-20
Update chronicle to v0.8.0

wagoodman:
This PR adjusts the release pipeline to use annotated tags over lightweight tags and updates chronicle to the latest version to leverage this.

While looking at the old release logs I noticed that the cache keys are not unique, thus are causing cache restore errors (which file silently). Additionally I also removed an unused env var.


.github/workflows/release.yaml
```diff
@@ -123,7 +120,9 @@ jobs:
 
       - name: Tag release
         run: |
-          git tag ${{ github.event.inputs.version }}
+          git config --global user.name "anchoreci"
```
willmurphyscode, line 123: Why pass `--global`? Is this setting correct for every job that runs on this CI worker? (Maybe it is, come to think of it.)

willmurphyscode (approved):

.github/workflows/release.yaml
wagoodman: I can leave off global and it would also still function (I added it out of habit), but leaving the global flag will not affect other jobs as the github hosted CI runners don't allow for persisted state (you get a clean state every time).

PR #2160, opened 2023-09-20
Require ordering of relationships when comparing parser output

wagoodman:
There were two problems found:
- relationships were not being sorted during test assertions, leading to unstable results
- IDs were not being set on packages for conan tests (which will affect the new sorting)

I also added an additional entry to the gitignore for a case that goland keeps generating (erroneously).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ VERSION
 /test/results
 coverage.txt
 *.log
+test/integration/test-fixtures/**/go.sum
 
 # probable archives
 .images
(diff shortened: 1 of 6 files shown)


willmurphyscode (approved):

.gitignore
```diff
@@ -30,6 +30,7 @@ VERSION
 /test/results
 coverage.txt
 *.log
+test/integration/test-fixtures/**/go.sum
```
wagoodman, line 33: I keep running into goland generating go.sums for these and we don't want them in the codebase

willmurphyscode (approved):

Issue #1984, opened 2023-07-31
Automate homebrew releases


chenrui333:
👋 Please integrate either [bump-homebrew-formula-action](https://github.com/mislav/bump-homebrew-formula-action) or [action-homebrew-bump-formula](https://github.com/dawidd6/action-homebrew-bump-formula) into the release pipeline to automate the homebrew releases. Thanks!

PR #2164, opened 2023-09-21
chore: add automated homebrew action

spiffcs:
Fixes: https://github.com/anchore/syft/issues/1984

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -158,14 +158,3 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_TOOLBOX_WEBHOOK_URL }}
         if: ${{ success() }}
-  
-  homebrew:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -158,3 +158,14 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_TOOLBOX_WEBHOOK_URL }}
         if: ${{ success() }}
+  
+  homebrew:
```
wagoodman, line 162: actually, if we add this as a separate post-release workflow (using I think `on: release`) , then we can safely re-run this without needing to worry about the release workflow itself. Thoughts?

wagoodman: I updated the branch to a separate post-release workflow (I have not run this workflow though)

wagoodman (approved):

PR #2188, opened 2023-10-02
Upgrade tool management

wagoodman:
This PR does two things primarily:
- Upgrades the Makefile to a [Taskfile](https://taskfile.dev/). A much smaller Makefile has been left behind for convenience, easy bootstrapping in CI, and for those of use that can't shake the muscle memory of typing `make ...` (*slowly raises hand*).
- Uses [binny](https://github.com/anchore/binny) to manage tools to utilize a caching strategy and improve quality of life while developing. This still keeps all tooling that is used in the repo in a local directory, however, that directory is now `.tool` instead of `.tmp`.

Example PR of the update bootstrap tools workflow:

Screenshot 2023-10-07 at 10 42 42 AM


willmurphyscode:
Is it expected that a github token needs to be in the environment to bootstrap after this change? I just checked out the branch and tried `make bootstrap`:

``` sh
❯ make bootstrap
bash: .tool/task: No such file or directory
bash: .tool/task: No such file or directory
make .tool/task
bash: .tool/task: No such file or directory
bash: .tool/task: No such file or directory
1 error occurred:
	* failed to install tool "task": unable to fetch github release go-task/task@v3.30.1: non-200 OK status code: 401 Unauthorized body: "{\"message\":\"Bad credentials\",\"documentation_url\":\"https://docs.github.com/graphql\"}"
make[1]: *** [.tool/task] Error 1
make: *** [bootstrap] Error 2
```

Am I doing something wrong with this? I checked that I don't have `GITHUB_TOKEN` set in the environment to rule out an expired one causing this error.

wagoodman:
With binny v0.6.2 you don't need a GITHUB_TOKEN for installing or even updating tools sourced from a github release page 🎉 (I've also improved the error message when a token is not provided and there is a need to fall back to the v4 github api)

spiffcs (approved):

PR #2213, opened 2023-10-11
Detect a license file in the root directory or META-INF of a jar

### LINKED ISSUE #2147: Find license by embedded license text
For Java archives, we have the option of retrieving licenses either directly from the META-INF/MANIFEST.MF if Bundle-License is defined (OSGi) or from the pom.xml if it's build with Maven (https://github.com/anchore/syft/pull/2115). However many projects are built with Gradle and are not OSGi bundles. Instead, these projects typically might include a license in the META-INF directory (for example the Spring framework includes the ASLv2 in META-INF/license.txt).

So what I'd like to do would be for Java archives to check for either a LICENSE or LICENSE.txt (case insensitive) in either the root directory or in META-INF and parse it for a license. I'm happy to contribute this - but do we have any existing code which can detect what license is in a license text file?

coheigea:
Closes #2147

If the jar doesn't have a Bundle-License in the Manifest, or a pom.xml in the archive with the declared license, it falls back to look for a valid license in either the root directory of the jar or in META-INF. I don't think it's necessary to check anywhere else for a license in a jar.

With this fix Syft correctly detects licenses for:

- https://repo1.maven.org/maven2/xerces/xercesImpl/2.12.2/ (META-INF/LICENSE)
- https://repo1.maven.org/maven2/org/apache/xmlbeans/xmlbeans/5.1.1/ (LICENSE.txt in root directory)

### DIFF
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	intFile "github.com/anchore/syft/internal/file"
+	"github.com/anchore/syft/internal/licenses"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/file"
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -310,6 +330,38 @@ func getDigestsFromArchive(archivePath string) ([]file.Digest, error) {
...
 
+func (j *archiveParser) getLicenseFromFileInArchive() ([]pkg.License, error) {
+	for _, filename := range licenses.FileNames {
+		licenseMatches := j.fileManifest.GlobMatch("/META-INF/" + filename)
+		if len(licenseMatches) == 0 {
+			// Try the root directory if it's not in META-INF
+			licenseMatches = j.fileManifest.GlobMatch("/" + filename)
+		}
+		// For now, we only support one license file per archive
+		if len(licenseMatches) == 1 {
```
wagoodman, line 341: why not raise up all of the licenses found?

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -310,6 +330,38 @@ func getDigestsFromArchive(archivePath string) ([]file.Digest, error) {
...
+			log.Debugf("Java license found in %s", filename)
```
wagoodman, line 342: nit: we should remove the two debug statements here

coheigea:
All done @wagoodman , please take another look

PR #2218, opened 2023-10-11
feat: Add Wordpress cataloger

### LINKED ISSUE #1911: Catalog wordpress plugins

Add a new cataloger to discover wordpress plugins

**Why is this needed**:
Wordpress is quite heavily used and there are quite a significant number of vulnerability records related to wordpress plugins, so being able to discover available plugins with syft would allow detection later.

**Additional context**:
Some information on what the expected metadata looks like: https://developer.wordpress.org/reference/functions/get_plugin_data/

Using the quickstart from https://hub.docker.com/r/bitnami/wordpress/, after execing into the wordpress image I get:

```
wp plugin list
+--------------------------------+----------+--------+---------+
| name                           | status   | update | version |
+--------------------------------+----------+--------+---------+
| akismet                        | inactive | none   | 5.2     |
| all-in-one-wp-migration        | inactive | none   | 7.76    |
| all-in-one-seo-pack            | inactive | none   | 4.4.0.1 |
| amp                            | inactive | none   | 2.4.1   |
| google-analytics-for-wordpress | inactive | none   | 8.17    |
| hello                          | inactive | none   | 1.7.2   |
| jetpack                        | inactive | none   | 12.2.1  |
| simple-tags                    | inactive | none   | 3.8.0   |
| w3-total-cache                 | inactive | none   | 2.3.3   |
| wp-mail-smtp                   | inactive | none   | 3.8.0   |
+--------------------------------+----------+--------+---------+
```

The plugins in that image end up under `/bitnami/wordpress/wp-content/plugins/`.  There is information about discovering the plugin directory at https://developer.wordpress.org/plugins/plugin-basics/determining-plugin-and-content-directories/ but possibly something like `**/wp-content/plugins/` would suffice

Looking at the (truncated) contents for one specific plugin:

```
cat /bitnami/wordpress/wp-content/plugins/wp-mail-smtp/wp_mail_smtp.php
<?php
/**
 * Plugin Name: WP Mail SMTP
 * Version: 3.8.0
 * Requires at least: 5.2
 * Requires PHP: 7.2
 * Plugin URI: https://wpmailsmtp.com/
 * Description: Reconfigures the <code>wp_mail()</code> function to use Gmail/Mailgun/SendGrid/SMTP instead of the default <code>mail()</code> and creates an options page to manage the settings.
 * Author: WP Mail SMTP
 * Author URI: https://wpmailsmtp.com/
 * Network: false
 * Text Domain: wp-mail-smtp
 * Domain Path: /assets/languages
 */

/**
 * @author    WPForms
 * @copyright WPForms, 2007-23, All Rights Reserved
 * This code is released under the GPL licence version 3 or later, available here
 * https://www.gnu.org/licenses/gpl.txt
 */

/**
 * Setting options in wp-config.php
 *
 * Specifically aimed at WP Multisite users, you can set the options for this plugin as
 * constants in wp-config.php. Copy the code below into wp-config.php and tweak settings.
 * Values from constants are NOT stripslash()'ed.
 *
 * When enabled, make sure to comment out (at the beginning of the line using //) those constants that you do not need,
 * or remove them completely, so they won't interfere with plugin settings.
 */

/*
```

disc:
Add a cataloger that detects installed Wordpress plugins by looking files in `*/wp-content/plugins/*` directories
Closes #1911 

Related MR to [Grype](https://github.com/anchore/grype)  for removing `wordpress` from `known` targets and unsuppressing of wordpress plugins vulnerabilities
https://github.com/anchore/grype/pull/1553


syft/pkg/cataloger/wordpress/parse_plugin.go
```diff
@@ -0,0 +1,94 @@
...
+type pluginData struct {
+	Licenses            []string `mapstructure:"licenses" json:"licenses,omitempty"`
+	pkg.WordpressPlugin `mapstructure:",squash" json:",inline"`
+}
+
+func parseWordpressPluginFiles(_ file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	var pkgs []pkg.Package
+	var fields = make(map[string]interface{})
+
+	bytes, err := io.ReadAll(reader)
```
wagoodman, line 42: we're really only expecting to parse the top comment of the file. I think we should try a best-effort at not searching the entire file, but instead to the first `*/` or only the first 100 lines. 

That's my initial take since this is capturing on potentially several PHP files. What do you think @disc ?

wagoodman (commented):
@disc I've caught up this branch based on the several changes made on main recently. I have a couple of small questions, otherwise I feel this is about the cross the finish line 🙌

syft/pkg/cataloger/wordpress/parse_plugin.go
disc: Hey, I agree. Will read first N kilobytes (for example, 4kb) into a buffer instead.

wagoodman (dismissed):
Thanks for adding this cataloger 🙌 !  (and thanks for your patience on this PR -- it came in the midst of an API refactor so needed to land afterwards).

wagoodman:
A couple of clarifying questions:
```json
{
  "id": "62595455b80ed7ff",
  "name": "Akismet Anti-spam: Spam Protection",
  "version": "5.3",
  "type": "wordpress-plugin",
  "foundBy": "wordpress-plugin-cataloger",
  ...  
  "metadata": {
    "plugin_name": "akismet",
    "author": "Automattic - Anti-spam Team",
    "author_uri": "https://automattic.com/wordpress-plugins/"
  }
}

```

1. Should the `pkg.Package.Name` be what is currently the `pkg.WordpressPlugin.PluginName`? (I could reason either way, curious for opinions here)
2. (Completely independent from the first question) Should `pkg.WordpressPlugin.PluginName` really be something like `pkg.WordpressPlugin.PluginInstallDirectory` (or similar) since it's derived from an installation path?

disc:
> @disc A couple of clarifying questions:
> 
> ```json
> {
>   "id": "62595455b80ed7ff",
>   "name": "Akismet Anti-spam: Spam Protection",
>   "version": "5.3",
>   "type": "wordpress-plugin",
>   "foundBy": "wordpress-plugin-cataloger",
>   ...  
>   "metadata": {
>     "plugin_name": "akismet",
>     "author": "Automattic - Anti-spam Team",
>     "author_uri": "https://automattic.com/wordpress-plugins/"
>   }
> }
> ```
> 
>     1. Should the `pkg.Package.Name` be what is currently the `pkg.WordpressPlugin.PluginName`? (I could reason either way, curious for opinions here)
> 
>     2. (Completely independent from the first question) Should `pkg.WordpressPlugin.PluginName` really be something like `pkg.WordpressPlugin.PluginInstallDirectory` (or similar) since it's derived from an installation path?
> 
> 
> (I can make the updates if there are any changes needed)

Hey @wagoodman. Looks like I need to make a few changes after a huge refactor and probably these questions won't be actual further.

wagoodman:
I can help with restoring/regenerating the JSON schema if needed. Also, is the refactor you're referring to one that happened on main (touching the cataloging API)? Or are you talking about a refactor within the new wordpress plugin cataloger you wrote?

disc:
> is the refactor you're referring to one that happened on main (touching the cataloging API)?

Yes, I meant this one. I've just completed all the needed changes - now need to look into a test run.

wagoodman:
Hey, I fixed the failing tests and updated the branch. I still have some questions left (from https://github.com/anchore/syft/pull/2218#issuecomment-1924531140)
 
1. Should the value in `pkg.Package.Name` really be what is currently in the `pkg.WordpressPlugin.PluginName` field? Said another way, should the package name be the install directory name instead of the plugin display name? 
1. Should `pkg.WordpressPlugin.PluginName` really be renamed to `pkg.WordpressPlugin.PluginInstallDirectory` (or similar) since it's derived from an installation path and not actually the `Plugin Name:` field from the underlying source code?

disc:
> Hey, I fixed the failing tests and updated the branch. I still have some questions left (from [#2218 (comment)](https://github.com/anchore/syft/pull/2218#issuecomment-1924531140))
> 
>     1. Should the value in `pkg.Package.Name` really be what is currently in the `pkg.WordpressPlugin.PluginName` field? Said another way, should the package name be the install directory name instead of the plugin display name?
> 
>     2. Should `pkg.WordpressPlugin.PluginName` really be renamed to `pkg.WordpressPlugin.PluginInstallDirectory` (or similar) since it's derived from an installation path and not actually the `Plugin Name:` field from the underlying source code?

1) I believe yes, it makes the package name short and simple, sometimes it contains long names such as `Akismet Anti-spam: Spam Protection`
2) Probably yes, it's contains almost the same as a name but uses for building CPEs

wagoodman:
It sounds like both answers point to making some small updates -- I can take a stab at it if you'd like? (or if you wanted to that's also alright)

wagoodman (approved):
I force pushed a rebase (there were a few commits from main that didn't belong). I also regenerated the JSON schema after changing the field caps to be consistent with the field convention we use (camel case).

Overall great work -- thanks for taking the time to add this cataloger 🙏 !

PR #2219, opened 2023-10-11
Remove internal string set

### LINKED ISSUE #2209: Remove internal.StringSet
Today we have `internal.StringSet` which is exposed on the `pkg.License` (the `URLs` field). This type should be swapped out with one that is accessible to others since the license object is a core object. Ideally we would replace all instances of the internal string set entirely.

wagoodman:
Removes the `internal.StringSet`, most notably from the exported `pkg.License` struct as well (for URLs, which has been converted into a `[]string`).

While debugging test failures caused by the replacement I swapped `assert.Equal` entries for non-trivial comparisons for the superior `cmp.Diff` utility to get a better sense of the cause of the failures.

Closes #2209


syft/pkg/cataloger/golang/licenses.go
```diff
@@ -31,6 +32,7 @@ type goLicenses struct {
 	opts                  GoCatalogerOpts
 	localModCacheResolver file.WritableResolver
 	progress              *monitor.CatalogerTask
+	licenseFileNames      *strset.Set
```
willmurphyscode, line 35: Can you help me understand why this is needed? There wasn't a string set here before.

syft/pkg/license.go
```diff
@@ -23,11 +24,11 @@ var _ sort.Interface = (*Licenses)(nil)
 // this is different for licenses since we're only looking for evidence
 // of where a license was declared/concluded for a given package
 type License struct {
-	Value          string             `json:"value"`
-	SPDXExpression string             `json:"spdxExpression"`
-	Type           license.Type       `json:"type"`
-	URLs           internal.StringSet `hash:"ignore"`
-	Locations      file.LocationSet   `hash:"ignore"`
+	Value          string
```
willmurphyscode, line 27: Did we intentionally lose the explicit JSON field names on this struct? Will `go` automatically change `Value` to `value` when writing this field to JSON? Otherwise this seems to be it changes the JSON schema.

syft/pkg/license.go
```diff
@@ -123,12 +128,14 @@ func NewLicenseFromFields(value, url string, location *file.Location) License {
...
-		l.URLs.Add(url)
+		l.URLs = append(l.URLs, url)
 	}
+
 	return l
 }
 
-// this is a bit of a hack to not infinitely recurse when hashing a license
+// Merge two licenses into a new license object.
+// TODO: this is a bit of a hack to not infinitely recurse when hashing a license
```
willmurphyscode, line 138: I don't understand this comment. What's hacky about this, and how does it relate to recursing on hashing a package object? Do we merge licenses somewhere?

kzantow (approved):

syft/pkg/license_test.go
```diff
@@ -97,3 +95,134 @@ func Test_Sort(t *testing.T) {
 
 	}
 }
+
+func TestLicense_Merge(t *testing.T) {
```
willmurphyscode, line 99: So merge only ever merges the URLs between two licenses objects, and requires all other fields to be equal? Just checking my understanding, because `Merge` seems like the name of a more general function than that.

syft/pkg/cataloger/golang/licenses.go
wagoodman: this is removing the string set that was in static space, which was previously assignable (thus not a good long term approach). While removing the internal string set usage I did not elect to replace it on the API in the same way, instead removing it's usage entirely. The place it was used for convenience was in this object, so that's why it was added here.

syft/pkg/license.go
wagoodman: This struct is not used in json marshaling, since there is a `syftjson/model.License` struct that covers this. I made a change to the struct that should have caused a marshaling change and was surprised when there wasn't a change in the output. While debugging I discovered this was the reason and removed the struct tags to not mislead future devs.

syft/pkg/license.go
wagoodman: I did not add this comment, but since I saw "hack" I wanted to encapsulate this with a `TODO`.

I think what this is commenting on is this exists to help with terminating the recursion of the `Hash` function used to calculate IDs, but am not entirely sure.

syft/pkg/license_test.go
wagoodman: it also merges location sets on licenses.

> Merge seems like the name of a more general function than that.

I'm not certain what you mean, are you recommending a rename of the name of the function `Merge`?

willmurphyscode: It's exported, so I don't think we should bother renaming it. I was just remarking that the current name seems surprising to me, but maybe I don't understand the semantics of the license package well. I don't think any change is needed here.

syft/pkg/license.go
willmurphyscode: This might benefit from a more specific doc comment on the function; I was surprised that "Merge" errors if the licenses aren't the same type, and only merges the location and URL fields.

wagoodman: updated!

willmurphyscode (approved):

PR #2227, opened 2023-10-13
Add additional license filenames

coheigea:
These additions help Syft to correctly identify licenses for:

- The widely deployed Spring framework (https://repo1.maven.org/maven2/org/springframework/spring-core/6.0.12/) which has a license in META-INF/license.txt
- https://repo1.maven.org/maven2/net/java/dev/jna/jna-platform/5.13.0/ which has licenses in AL2.0 and LGPL2.1 filenames.

For the first point we could instead do a case insensitive check on the filenames, but I thought that was probably overkill.

### DIFF
--- a/internal/licenses/list.go
+++ b/internal/licenses/list.go
@@ -18,6 +18,7 @@ var (
 		"LICENSE",
 		"LICENSE.md",
 		"LICENSE.markdown",
+		"license.txt",
 		"LICENSE.txt",
 		"LICENSE-2.0.txt",
 		"LICENCE-2.0.txt",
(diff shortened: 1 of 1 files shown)


wagoodman:
> For the first point we could instead do a case insensitive check on the filenames...

actually, I think ultimately we should be doing that. I can't think of a case where case should matter for a filename for the use case of "is this a common license path or not". We can get this in first though 👍

wagoodman (approved):

coheigea:
Updated PR for case insensitive matching: https://github.com/anchore/syft/pull/2235

Issue #2103, opened 2023-09-07
Support parsing license information in Maven projects via parent poms


coheigea:
I noticed that Syft does not report a license for a Maven project if the license is contained in the parent pom instead. This is a pretty common thing in a lot of open-source projects, to have a single parent pom.xml which defines commons things (including license information), and then each module in the project inherits this pom. 

For example, here are two examples of projects Syft does not detect the license, even though it's defined in both cases in the parent pom:

- https://repo1.maven.org/maven2/org/apache/httpcomponents/httpclient/4.5.13/httpclient-4.5.13.pom
- https://repo1.maven.org/maven2/org/slf4j/slf4j-api/1.7.32/slf4j-api-1.7.32.pom

Please consider this feature request, because as it stands Syft does not report any license for quite a few Maven projects due to this issue. Maybe it could be configurable via a feature flag (detect transitive licenses). To prevent infinite loops or denial of service type of attacks, it should only walk up to 2/3 parent poms.

Note that using mvn help:effective-pom -f httpclient-4.5.13.pom *does* show the correct license

coheigea:
I can try to help if someone can point me to the right direction....

coheigea:
Fixed in https://github.com/anchore/syft/releases/tag/v0.93.0

coheigea:
Closing this as both PRs applied

PR #2228, opened 2023-10-13
Read a license from a parent pom stored in Maven Central

coheigea:
Part 1 of https://github.com/anchore/syft/issues/2103

Syft does not report a license for a Maven project if the license is contained in the parent pom instead. This is a pretty common thing in a lot of open-source projects, to have a single parent pom.xml which defines commons things (including license information), and then each module in the project inherits this pom.

For example, here are two examples of projects Syft does not detect the license, even though it's defined in both cases in the parent pom:

 - https://repo1.maven.org/maven2/org/apache/httpcomponents/httpclient/4.5.13/httpclient-4.5.13.pom
 - https://repo1.maven.org/maven2/org/slf4j/slf4j-api/1.7.32/slf4j-api-1.7.32.pom

With this PR Syft correctly finds the license from Maven Central. It only checks one parent, a future PR will cater for projects that have nested parents. 

I don't know if the network call is a problem or if you want to make it configurable.


coheigea:
Static analysis fails with:
```
Error: syft/pkg/cataloger/java/archive_parser.go:310:15: G107: Potential HTTP request made with variable url (gosec)
	resp, err := http.Get(mavenCentralURL)
```
I'd be interested to hear thoughts on making this more secure. The URL starts and ends with a hard-coded String so I think it would be difficult to force it to go to an arbitrary URL.

spiffcs:
> Static analysis fails with:
> 
> ```
> Error: syft/pkg/cataloger/java/archive_parser.go:310:15: G107: Potential HTTP request made with variable url (gosec)
> 	resp, err := http.Get(mavenCentralURL)
> ```
> 
> I'd be interested to hear thoughts on making this more secure. The URL starts and ends with a hard-coded String so I think it would be difficult to force it to go to an arbitrary URL. Maybe I could check for a "@" character to make sure it can't use the trick of treating the pre-amble as a Basic Auth username/password?

👋 Let me give this a shot at getting this a bit more secure and using some of the standard library to do basic URL creation.
There are probably some usages of https://pkg.go.dev/net/url#JoinPath, https://pkg.go.dev/net/url#PathEscape, and https://pkg.go.dev/net/http#NewRequest we can use to satisfy the security fix here =)

spiffcs:
@coheigea I pushed [2702d17](https://github.com/anchore/syft/pull/2228/commits/2702d1780da573be505f5629bb60a4f889825574) - let's see if the linter has an issue here =)

coheigea:
LGTM thanks @spiffcs . My only remaining question is whether we want to make this behaviour of calling out to Maven Central controllable or not via a feature flag?

spiffcs:
> LGTM thanks @spiffcs . My only remaining question is whether we want to make this behaviour of calling out to Maven Central controllable or not via a feature flag?

Ahh great call out - we actually already have a feature flag for this kind of behavior over in grype:
https://github.com/anchore/grype/blob/6b4978f633f4c917b43e6a3f4849ba619876b5fb/cmd/grype/cli/options/datasources.go#L16

Because we also want syft to be offline by default I think we want the same kind of behavior here where the user must enable this functionality if they want syft to make requests to upstream maven.

Let me see what I can add to get us there.

spiffcs:
@wagoodman I wanted to get your opinion on adding this new config option: `cfg.Java.SearchMavenForLicenses`
```
java:
    search-maven-for-licenses: false
```

I've come to a small breaking point and don't have a clear path to proceed.

Injecting a config option into something of type `generic.Parser` has proven to be a little awkward. Rather than go full steam ahead and update `generic.Parser`'s signature I'd like to start a quick discussion here on alternatives. It also could be the case that this is solved by some work currently in flight.

At first I tried to see if `generic.Environment` would be a good place to load this value, but this is read to far into the cataloging process beyond where our different parsers and catalogers have been constructed. 

Given that `generic.Parser` is of type `func` the normal decorator pattern of doing something like `parser.WithConfig(cfg)` is a little blocked since there is no place to store the value before `Parser` is invoked.

spiffcs:
@coheigea This PR is awesome thank you so much for the work and attention that went into it. I've pushed the changes needed for the config update per some recommendations from @wagoodman 


cc @anchore/tools for a final review

spiffcs:
Here is a screenshot showing the difference between this PR and main given the example jar
Screenshot 2023-10-30 at 1 53 22 PM

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -16,7 +22,7 @@ import (
 	"github.com/anchore/syft/syft/pkg/cataloger/generic"
 )
 
-var _ generic.Parser = parseJavaArchive
+const MavenBaseURL = "https://repo1.maven.org/maven2"
```
kzantow, line 25: Should this also be configurable? E.g. if the user has a private maven repo, they could try to point at that to resolve licenses, which also could work for air-gap

wagoodman:
Just to call out some earlier conversations here: since syft is offline by default, having these feature flags off by default is important. However, in the future we will in including a top-level flag that can enable a wide range of cataloger-specific flags to do with online searches. So even if there are 10 separate configuration options across different catalogers for online search behavior the goal is to make flipping them all on as easy as `syft myimage:latest --online` (or something similar).

cmd/syft/cli/options/catalog.go
```diff
@@ -137,6 +139,8 @@ func (cfg Catalog) ToCatalogerConfig() cataloger.Config {
 		LinuxKernel: kernel.LinuxCatalogerConfig{
 			CatalogModules: cfg.LinuxKernel.CatalogModules,
 		},
+		Java: javaCataloger.NewJavaCatalogerOpts().
+			WithSearchMavenForLicenses(cfg.Java.SearchMavenForLicenses),
```
kzantow, line 143: very small nit: it looks like the other lines above and below are just directly constructing the config structs but this is introducing the ".With<thing>()" builder style. since this is just a boolean, and the only property on the struct, it would probably work fine to follow the pattern of the others and not use the builder style

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -282,6 +302,89 @@ func artifactIDMatchesFilename(artifactID, fileName string) bool {
...
+	}
+	return requestURL, err
+}
+
+func getPomFromMavenCentral(groupID, artifactID, version string) (*gopom.Project, error) {
+	requestURL, err := formatMavenPomURL(groupID, artifactID, version)
+	if err != nil {
+		return nil, err
+	}
+	log.Debugf("Trying to fetch parent pom from Maven central %s", requestURL)
```
wagoodman, line 342:
```suggestion
	log.Tracef("trying to fetch parent pom from Maven central: %s", requestURL)
```

syft/pkg/cataloger/config.go
```diff
@@ -24,13 +25,18 @@ func DefaultConfig() Config {
...
+		Java:                            java.NewJavaCatalogerOpts(),
 		ExcludeBinaryOverlapByOwnership: true,
 	}
 }
 
-func (c Config) Java() java.Config {
+// JavaConfig merges relevant config values from Config to return a java.Config struct.
+// Values like IncludeUnindexedArchives and IncludeIndexedArchives are used across catalogers
+// and are not specific to Java requiring this merge.
+func (c Config) JavaConfig() java.Config {
```
kzantow, line 36: very small nit: this rename seems unnecessary, since this is a method on the `Config` object already, it's a bit of a stutter `config.JavaConfig()` rather `config.Java()`

syft/pkg/cataloger/java/archive_parser.go
wagoodman, line 25: This could also be a java configuration to allow folks to configure additional repositories. (consider this an optional comment)

kzantow (approved):
LGTM but as noted in a comment, I do think it would be good to make the URL prefix for the maven repo configurable

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -282,6 +302,89 @@ func artifactIDMatchesFilename(artifactID, fileName string) bool {
...
+
+	mavenRequest, err := http.NewRequest(http.MethodGet, requestURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("unable to format request for Maven central: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(mavenRequest)
```
wagoodman, line 349: I don't think there is a timeout on the default client, so we should specify it explicitly here

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -272,6 +288,10 @@ func (j *archiveParser) guessMainPackageNameAndVersionFromPomInfo() (name, versi
 	if version == "" && pomProjectObject.JavaPomProject != nil {
 		version = pomProjectObject.Version
 	}
+	if !reflect.ValueOf(pomProjectObject).IsZero() && j.cfg.SearchMavevForLicenses {
```
wagoodman, line 291: nit: instead of the reflect zero value would it be possible to change the stack variables to be pointers to be able to discern optionality?

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -282,6 +302,89 @@ func artifactIDMatchesFilename(artifactID, fileName string) bool {
 	return strings.HasPrefix(artifactID, fileName) || strings.HasSuffix(fileName, artifactID)
 }
 
+func findPomLicenses(pomProjectObject *parsedPomProject) {
+	// If we don't have any licenses until now, and if we have a parent Pom, then we'll check the parent pom in maven central for licenses.
+	if pomProjectObject != nil && pomProjectObject.Parent != nil && len(pomProjectObject.Licenses) == 0 {
+		parentPom, err := getPomFromMavenCentral(pomProjectObject.Parent.GroupID, pomProjectObject.Parent.ArtifactID, pomProjectObject.Parent.Version)
+		if err != nil {
+			// We don't want to abort here as the parent pom might not exist in Maven Central, we'll just log the error
+			log.Debugf("unable to get parent pom from Maven central: %v", err)
```
wagoodman, line 311:
```suggestion
			log.Tracef("unable to get parent pom from Maven central: %v", err)
```
Since it's not clear if we should have been able to get the pom at all keeping this low in the logging calls makes sense.

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -282,6 +302,89 @@ func artifactIDMatchesFilename(artifactID, fileName string) bool {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse pom from Maven central: %w", err)
+	}
+
+	pom, err := decodePomXML(strings.NewReader(string(bytes)))
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse pom from Maven central: %w", err)
+	}
+
+	defer func() {
```
wagoodman, line 364: this defer close function should be closer to where the response object is created (in case decoding the pom fails we'll still release the resource).

syft/pkg/cataloger/java/archive_parser.go
wagoodman: never mind, we can always introduce this later

syft/pkg/cataloger/java/cataloger.go
```diff
@@ -9,17 +9,21 @@ import (
 
 // NewJavaCataloger returns a new Java archive cataloger object.
 func NewJavaCataloger(cfg Config) *generic.Cataloger {
+	gap := newGenericArchiveParserAdapter(cfg)
```
wagoodman, line 12: nit: not all of these need to be constructed based on the below conditions

syft/pkg/cataloger/java/config.go
```diff
@@ -3,4 +3,5 @@ package java
 type Config struct {
 	SearchUnindexedArchives bool
 	SearchIndexedArchives   bool
+	SearchMavevForLicenses  bool
```
wagoodman, line 6:
```suggestion
	SearchMavenForLicenses  bool
```

README.md
```diff
@@ -601,6 +601,12 @@ golang:
    # if unset this defaults to $GONOPROXY
    # SYFT_GOLANG_NOPROXY env var
    no-proxy: ""
+  
+java:
+   # when running across pom.xml files that could have more information
```
wagoodman, line 606: it's probably worth noting in this comment that this is particularly useful in situations where the parent pom has this information but it is not accessible from within the final built artifact.

syft/pkg/cataloger/config.go
spiffcs: Sorry - I think this did have to be renamed in order for the struct to be named consistently.

`Config` already had the fields `Python` and `Golang`. 

I opted to change `Java` from a `function` to the field like the other languages on that struct. This was so that `Java` could use the `CatalogerOpts` pattern like the  `Golang` field used for options that are exclusive to the specific ecosystem.

`config.JavaConfig()` doesn't seem too bad a stutter given the trade off. We still need a function that can grab fields from `Config` which are not mutually exclusive to a single cataloger to generate a `java.Config` (things like `IncludeIndexedArchives`) hence `config.JavaConfig` being born.

cmd/syft/cli/options/catalog.go
spiffcs: Oh my bad - I saw that the Golang cataloger introduced the `With` style and went with that approach.  I'm happy to change it back to a struct with two fields now (Online and URL). I do agree that the `cataloger.Config` should probably be consistent in its style

syft/pkg/cataloger/java/cataloger.go
spiffcs: moved these down into their respective configuration blocks

syft/pkg/cataloger/java/archive_parser.go
spiffcs: updated to trace

syft/pkg/cataloger/java/archive_parser.go
spiffcs: updated to a configurable value with the above as default

spiffcs: updated this to configurable with the above value as the default

syft/pkg/cataloger/java/archive_parser.go
spiffcs: Added timeout

syft/pkg/cataloger/java/archive_parser.go
spiffcs: Converted this to a pointer - I think this is why the integration tests are now failing I have to look at this.

syft/pkg/cataloger/java/archive_parser.go
spiffcs: updated to trace

syft/pkg/cataloger/java/archive_parser.go
spiffcs: defer lifted to right after error from client

syft/pkg/cataloger/java/cataloger.go
spiffcs: moved separate adapters into their respective config locations

syft/pkg/cataloger/java/config.go
spiffcs: Typo Corrected

README.md
spiffcs: updated readme

README.md
```diff
@@ -601,6 +601,15 @@ golang:
...
    no-proxy: ""
+  
+java:
+   # when running across pom.xml files that could have more information, syft will
+   # explicitly search maven for license information by querying the online pom when this is true eg:
+   # https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-starter-test/3.1.5/spring-boot-starter-test-3.1.5.pom
+   # this option is helpful for when the parent pom has this information,
+   # but it is not accessible from within the final built artifact
+   search-maven-for-licenses: false
+   maven-central-url: "https://repo1.maven.org/maven2"
```
wagoodman, line 612: maybe change this to `maven-url` and we can leave the default to point to central. There are multiple maven repos that exist, but there is only one `central`.

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -282,6 +304,92 @@ func artifactIDMatchesFilename(artifactID, fileName string) bool {
...
+	}
+	return requestURL, err
+}
+
+func getPomFromMavenCentral(groupID, artifactID, version string) (*gopom.Project, error) {
+	requestURL, err := formatMavenPomURL(groupID, artifactID, version)
+	if err != nil {
+		return nil, err
+	}
+	log.Tracef("Trying to fetch parent pom from Maven central %s", requestURL)
```
wagoodman, line 344: nit: 
```suggestion
	log.Tracef("trying to fetch parent pom from Maven central %s", requestURL)
```

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -282,6 +304,92 @@ func artifactIDMatchesFilename(artifactID, fileName string) bool {
...
+
+	mavenRequest, err := http.NewRequest(http.MethodGet, requestURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("unable to format request for Maven central: %w", err)
+	}
+
+	httpClient := &http.Client{
+		Timeout: time.Second * 2,
```
wagoodman, line 352: `2` seems too small -- maybe `10` ?

wagoodman (approved):

PR #2220, opened 2023-10-11
Account for maven bundle plugin and fix filename matching

wagoodman:
This PR makes a few changes to the java archive cataloger:
- takes into account the maven bundle plugin, which is common tooling that we can key off of patterns from the `Bundle-SymbolicName` field to get a group ID and artifact ID. Then we can extract the artifact ID as the package name.
- If a filename looks like a group ID + artifact ID, attempt to parse it as such and extract the artifact ID as the package name.
- When attempting to find if a pom.xml matches with an existing package, allow for the artifact ID to be a suffix of the filename (not just the filename as a prefix of the artifact ID as it is today)

Blocked until the following are verified:
- [x] CPE generation logic can additionally consider the new sources of information when crafting a vendor (I think this is done, but need to double check)
- [x] purl generation is using the right group ID found (again, I think this is done, but need to check)
- [x] incorporate locally into grype and run against the quality gate -- are we better or worse off on the whole? https://github.com/anchore/grype/pull/1555

Closes #2217
Fixes #2130

PR #2231, opened 2023-10-16
Syft does not handle the case of parsing a jar with multiple poms

coheigea:
The restriction to only parse one pom.xml / pom.properties in the archive fails for certain jars that shade multiple dependencies into the jar. For example:

https://repo1.maven.org/maven2/org/apache/directory/api/api-all/2.0.0/

which has:

```
./META-INF/maven/org.apache.directory.api/api-ldap-extras-codec-api/pom.xml
./META-INF/maven/org.apache.directory.api/api-all/pom.xml
./META-INF/maven/org.apache.directory.api/api-ldap-schema-converter/pom.xml
./META-INF/maven/org.apache.directory.api/api-ldap-extras-trigger/pom.xml
./META-INF/maven/org.apache.directory.api/api-ldap-extras-sp/pom.xml
./META-INF/maven/org.apache.directory.api/api-util/pom.xml
./META-INF/maven/org.apache.directory.api/api-ldap-net-mina/pom.xml
./META-INF/maven/org.apache.directory.api/api-ldap-codec-standalone/pom.xml
./META-INF/maven/org.apache.directory.api/api-ldap-extras-util/pom.xml
./META-INF/maven/org.apache.directory.api/api-ldap-extras-aci/pom.xml
```

Syft does end up finding the license for api-all anyway, because of a LICENSE file included in the archive, but otherwise wouldn't find a license as it's not proceeding to the loop to find the correct pom.

### DIFF
--- a/syft/pkg/cataloger/java/archive_parser_test.go
+++ b/syft/pkg/cataloger/java/archive_parser_test.go
@@ -1036,6 +1036,7 @@ func Test_parseJavaArchive_regressions(t *testing.T) {
 		fixtureName           string
 		expectedPkgs          []pkg.Package
 		expectedRelationships []artifact.Relationship
+		assignParent          bool
 		want                  bool
 	}{
 		{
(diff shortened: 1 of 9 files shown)


coheigea:
@wagoodman Rebased this after https://github.com/anchore/syft/pull/2220 was merged, if you could take a look please?

coheigea:
rebased again

spiffcs:
@coheigea - thanks for rebasing this! Since this code change didn't break any tests I think I'm going to take your example package and write a quick unit test that validates the change and keeps us covered going forward. Let me see if I can get that in today and then I'll 🟢 and merge this one too =)

kzantow (approved):
LGTM 🎉

syft/pkg/cataloger/java/archive_parser_test.go
```diff
@@ -1159,6 +1224,20 @@ func Test_parseJavaArchive_regressions(t *testing.T) {
 	}
 }
 
+func assignParent(expectedPackages []pkg.Package) {
```
wagoodman, line 1227: nit suggestion: a little easier to understand, even if the caller was doing something akin to `assignParent(&tt.expectedPkgs[0], tt.expectedPkgs[1:]...)`

```suggestion
func assignParent(parent *pkg.Package, childPackages ...pkg.Package) {
```

wagoodman (approved):

PR #2234, opened 2023-10-16
feat: add conaninfo.txt parser to detect conan packages in docker images

Pro:
Currently, syft only supports parsing conanfile.txt or conan.lock files.

If a conan package is installed inside a docker container (e.g., in $HOME/.conan/data), then this is not detected by syft.

With this MR the corresponding functionality is added.

I.e., every installed conan package contains a `conaninfo.txt`.

We can use this in the globbing to detect the packages and then parse the additional conan package information.

### DIFF
--- a/syft/pkg/cataloger/cpp/cataloger.go
+++ b/syft/pkg/cataloger/cpp/cataloger.go
@@ -10,5 +10,6 @@ const catalogerName = "conan-cataloger"
 func NewConanCataloger() *generic.Cataloger {
 	return generic.NewCataloger(catalogerName).
 		WithParserByGlobs(parseConanfile, "**/conanfile.txt").
-		WithParserByGlobs(parseConanlock, "**/conan.lock")
+		WithParserByGlobs(parseConanlock, "**/conan.lock").
+		WithParserByGlobs(parseConaninfo, "**/conaninfo.txt")
 }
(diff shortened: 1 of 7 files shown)


syft/pkg/cataloger/cpp/cataloger.go
```diff
@@ -10,5 +10,6 @@ const catalogerName = "conan-cataloger"
 func NewConanCataloger() *generic.Cataloger {
 	return generic.NewCataloger(catalogerName).
 		WithParserByGlobs(parseConanfile, "**/conanfile.txt").
-		WithParserByGlobs(parseConanlock, "**/conan.lock")
+		WithParserByGlobs(parseConanlock, "**/conan.lock").
+		WithParserByGlobs(parseConaninfo, "**/conaninfo.txt")
 }
```
wagoodman, line 15: it seems that the conaninfo.txt fits a different use case than the other existing parsers. The existing parsers seem best fit for detecting declarations of software to be installed where as the new conaninfo parser detects evidence of installed software. @Pro  is that interpretation right?

If so, then we should probably leave the existing cataloger as is and make a new `NewConanInfoCataloger()` (or another name) that houses your new parser. That way we could keep the existing cataloger for directory scans and the new cataloger for image scans.

Pro: @wagoodman your interpretation is fully correct!

I changed the code to have a separated ConanInfoCataloger.

Can you please check the changes?

wagoodman (approved):
nice work @Pro  🙌

PR #2235, opened 2023-10-17
Perform case insensitive matching on Java License files

coheigea:
Following on from https://github.com/anchore/syft/pull/2227, perform case insensitive matching in GlobMatch so that we can check all license files in a case insensitive manner.

### DIFF
--- a/internal/file/zip_file_manifest.go
+++ b/internal/file/zip_file_manifest.go
@@ -40,15 +40,18 @@ func (z ZipFileManifest) Add(entry string, info os.FileInfo) {
 }
 
 // GlobMatch returns the path keys that match the given value(s).
-func (z ZipFileManifest) GlobMatch(patterns ...string) []string {
+func (z ZipFileManifest) GlobMatch(caseInsensitive bool, patterns ...string) []string {
 	uniqueMatches := strset.New()
 
 	for _, pattern := range patterns {
 		for entry := range z {
 			// We want to match globs as if entries begin with a leading slash (akin to an absolute path)
 			// so that glob logic is consistent inside and outside of ZIP archives
-			normalizedEntry := normalizeZipEntryName(entry)
+			normalizedEntry := normalizeZipEntryName(caseInsensitive, entry)
 
+			if caseInsensitive {
+				pattern = strings.ToLower(pattern)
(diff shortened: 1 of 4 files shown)


internal/file/zip_file_manifest.go
```diff
@@ -48,7 +48,7 @@ func (z ZipFileManifest) GlobMatch(patterns ...string) []string {
 			// so that glob logic is consistent inside and outside of ZIP archives
 			normalizedEntry := normalizeZipEntryName(entry)
 
-			if GlobMatch(pattern, normalizedEntry) {
+			if GlobMatch(strings.ToLower(pattern), normalizedEntry) {
```
spiffcs, line 51: @coheigea just one question for this PR - this changes every pattern that uses `GlobMatch` to be case insensitive by lowercasing the inputs. 

Since the purpose of this PR is around `licenses`, what do you think about sanitizing the input of `patterns` where this function is used surrounding license detection. I can do some digging in the code to see what this looks like.

cc @wagoodman if he has any other comments about if this change has any unintended consequences

I see usages of this function under `syft/pkg/cataloger/java`:
```
discoverMainPackage
discoverPkgsFromAllMavenFiles
discoverPkgsFromZip
getLicenseFromFileInArchive
guessMainPackageNameAndVersionFromPomInfo
```

Is it ultimately these calses you want to call `strings.ToLower` on:
https://github.com/anchore/syft/blob/234ce4e1f3b995013551229105a450ed29a6112a/syft/pkg/cataloger/java/archive_parser.go#L341-L347

spiffcs (commented):
Thanks for the PR @coheigea!

Small comment about potential consequences of the change for other consumers of the `GlobMatch` function. WDYT of changing this at the caller level?

Was there another reason where you saw the `GlobMatch` needed to be case insensitive for its other uses?

internal/file/zip_file_manifest.go
coheigea: @spiffcs I changed the code to add a parameter to control whether or not case insensitive matching is done. It's false for all the use-cases apart from when I'm matching against a license file.

spiffcs (approved):

PR #2240, opened 2023-10-19
Label PRs when the json schema changes

### LINKED ISSUE #1938: Add automation to detect JSON schema breakages
Today we use chronicle to generate release notes based off of labels on PRs and issues. It would be great to add automation that can auto-label PRs with schema-specific changes. Take for instance this PR https://github.com/anchore/syft/releases/tag/v0.85.0 . 
The `breaking-change` label was manually added to the https://github.com/anchore/syft/pull/1846 but could have been automatically added by automation when a new JSON schema file was added in `schema/json` and compared the new version to the last version. Adding both the `breaking-change` and something that indicates the json schema was changed such as a new `new-json-schema` would be nice. With some changes to chronicle we could add another section that denotes the specific changes that changed the json schema.

wagoodman:
Closes #1938

This adds a utility that will label all new PRs (from forks or internal branches) based on the following criteria:
- if there is a new, deleted or modified json schema then a `json-schema` label will be added to the PR
- if the major version of the new schema version is greater than the major version of the current schema then a `breaking-change` label will be added to the PR
- if a former json schema file is modified or removed, then a sticky comment on the PR will be added warning of a possible issue. This is not necessarily a breaking change or should block the PR since there may be other reason why this change is valid. If a new commit resolves the issue (no more modified or removed schemas) then the stick comment is automatically hidden (folded). If new commits are added to the PR when there is an existing comment and the issue is not addressed no new comments are added. If a previously resolved issue becomes an issue again, then a new comment is added.

Test PR (forked) showing the intended behavior that this new workflow adds:
Screenshot 2023-10-19 at 5 35 20 PM

This new workflow necessarily uses the `pull_request_target` event trigger, which has security implications. I've left comments in the workflow to explain the context around this.

### DIFF
--- a/.github/scripts/labeler.py
+++ b/.github/scripts/labeler.py
@@ -16,7 +16,7 @@ def main(changed_files: str | None = None, merge_base_schema_files: str | None =
     comment_file_path = os.environ.get("CI_COMMENT_FILE")
 
     if not comment_file_path:
-        print("CI_COMMENT_FILE not set, skipping")
+        print("CI_COMMENT_FILE not set")
         sys.exit(1)
 
     if not pr_number:
@@ -34,11 +34,11 @@ def main(changed_files: str | None = None, merge_base_schema_files: str | None =
 
     else:
         if not is_ci():
-            print("Not in CI, skipping")
+            print("Not in CI")
             sys.exit(1)
 
         if not pr_number:
-            print("Not a PR, skipping")
+            print("Not a PR")
             sys.exit(1)
 
         pr_changed_files = get_pr_changed_files(pr_number)
--- a/.github/scripts/labeler.py
+++ b/.github/scripts/labeler.py
@@ -1,0 +1,224 @@
+from __future__ import annotations
+
+import sys
+import glob
+import subprocess
+import os
+
+
+DRY_RUN = False
+
+
+def main(changed_files: str | None = None, merge_base_schema_files: str | None = None):
+    global DRY_RUN
+
+    pr_number = os.environ.get("GITHUB_PR_NUMBER")
+    comment_file_path = os.environ.get("CI_COMMENT_FILE")
+
+    if not comment_file_path:
+        print("CI_COMMENT_FILE not set, skipping")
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -70,7 +70,7 @@ all: static-analysis test ## Run all linux-based checks (linting, license check,
 static-analysis: check-go-mod-tidy lint check-licenses check-json-schema-drift  ## Run all static analysis checks
 
 .PHONY: test
-test: unit integration validate-cyclonedx-schema benchmark cli ## Run all tests (currently unit, integration, linux compare, and cli tests)
+test: unit integration validate-cyclonedx-schema benchmark test-utils cli ## Run all tests (currently unit, integration, linux compare, and cli tests)
```
willmurphyscode, line 73: I believe this causes Python to be needed for `make test`, when it wasn't previously. Should that be added to https://github.com/anchore/syft/blob/main/DEVELOPING.md#getting-started ? (Maybe as an optional requirement since running the whole `make test` is relatively unusual.)

.github/scripts/labeler.py
```diff
@@ -0,0 +1,224 @@
...
+
+def main(changed_files: str | None = None, merge_base_schema_files: str | None = None):
+    global DRY_RUN
+
+    pr_number = os.environ.get("GITHUB_PR_NUMBER")
+    comment_file_path = os.environ.get("CI_COMMENT_FILE")
+
+    if not comment_file_path:
+        print("CI_COMMENT_FILE not set, skipping")
+        sys.exit(1)
```
willmurphyscode, line 20: We exit 1 for the case where we're "skipping" here, but also on line 58 if there are no existing JSON files. Logging that we're "skipping" and then exiting 1 seems confusing to me. Maybe this should log that it's failing?

Makefile
wagoodman: there technically was an implicit requirement on python for coverage analysis, but that just means it should have already been in the doc 😄  (I'll add it)

Issue #2255, opened 2023-10-25
.NET / nuget - invalid SBOM generated after parsing


kmoens:
**What happened**:
When using Syft to generate the SBOM for one of our Docker Images we encountered issues uploading the generated SBOM towards Dependency Track due to an invalid Package URL that was created.

Hereby the fragment of the SBOM:
```
    {
      "bom-ref": "pkg:nuget/%C3%A4b\u0001FileVersion@4.6.25512.01%20built%20by:%20dlab-DDVSOWINAGE016.%20Commit%20Hash:%20d0d5c7b49271cadb6d97de26d8e623e98abdc8db?package-id=3ddefa9d1305fed5",
      "type": "library",
      "name": "äb\u0001FileVersion",
      "version": "4.6.25512.01 built by: dlab-DDVSOWINAGE016. Commit Hash: d0d5c7b49271cadb6d97de26d8e623e98abdc8db",
      "purl": "pkg:nuget/%C3%A4b\u0001FileVersion@4.6.25512.01%20built%20by:%20dlab-DDVSOWINAGE016.%20Commit%20Hash:%20d0d5c7b49271cadb6d97de26d8e623e98abdc8db",
      "properties": [
        {
          "name": "syft:package:foundBy",
          "value": "dotnet-portable-executable-cataloger"
        },
        {
          "name": "syft:package:metadataType",
          "value": "DotnetPortableExecutableMetadata"
        },
        {
          "name": "syft:package:type",
          "value": "dotnet"
        },
        {
          "name": "syft:location:0:layerID",
          "value": "sha256:d3093f2eddfbe245a9e15bb6245e82ab28499636b9052f643b2d53a89001cadb"
        },
        {
          "name": "syft:location:0:path",
          "value": "/app/runtimes/win-x64/native/sni.dll"
        }
      ]
    },
```

This resulted in an error while parsing the SBOM as follows:
```
[2023-10-24 16:27:06.862]  malformed package url pkg:nuget/%C3%A4bFileVersion@4.6.25512.01%20built%20by:%20dlab-DDVSOWINAGE016.%20Commit%20Hash:%20d0d5c7b49271cadb6d97de26d8e623e98abdc8db
[...]

Issue #2241, opened 2023-10-20
Wrong parsing after v0.85.0 syft for some components


gandalf1990PL:
**What happened**:
Syft is parsing some compontnes on wrong way. This occurs that dependency track cant upload boms. Throwing errors like this:
Cannot insert the value NULL into column 'NAME', table 'DependencyTrackDB.dbo.COMPONENT'; column does not allow nulls. INSERT fails.

e.g when using any syft with higher version that 0.85.0

```
    <component bom-ref="pkg:nuget/%20@3.14.40721.0918?package-id=db6ac7656db915a9" type="library">
      <name> </name>
      <version>3.14.40721.0918</version>
      <cpe>cpe:2.3:a:_:_:3.14.40721.0918:*:*:*:*:*:*:*</cpe>
      <purl>pkg:nuget/%20@3.14.40721.0918</purl>
      <properties>
        <property name="syft:package:foundBy">dotnet-portable-executable-cataloger</property>
        <property name="syft:package:metadataType">DotnetPortableExecutableMetadata</property>
        <property name="syft:package:type">dotnet</property>
        <property name="syft:location:0:layerID">sha256:3112682c3f97ae1dbc6a6ad3cf230dfef4225c1a53ecc3ebd40e3f6bf36e5f50</property>
        <property name="syft:location:0:path">/app/Microsoft.IdentityModel.Clients.ActiveDirectory.Platform.dll</property>
      </properties>
    </component>
    <component bom-ref="pkg:nuget/%20@3.29.4?package-id=e6afd6dc74f9ae32" type="library">
      <name> </name>
      <version>3.29.4</version>
      <cpe>cpe:2.3:a:_:_:3.29.4:*:*:*:*:*:*:*</cpe>
      <purl>pkg:nuget/%20@3.29.4</purl>
      <properties>
        <property name="syft:package:foundBy">dotnet-portable-executable-cataloger</property>
        <property name="syft:package:metadataType">DotnetPortableExecutableMetadata</property>
        <property name="syft:package:type">dotnet</property>
        <property name="syft:location:0:layerID">sha256:3112682c3f97ae1dbc6a6ad3cf230dfef4225c1a53ecc3ebd40e3f6bf36e5f50</property>
        <property name="syft:location:0:path">/app/Microsoft.Azure.Cosmos.Direct.dll</property>
      </properties>
    </component>
```

For 0.85.0 version there is no components with empty <name> </name> section.

The result of bom files for same container images differ significantly on amount of components 

[...]

Issue #2029, opened 2023-08-14
Parser for dotnet_portable_executable using wrong attribute name. 


Roxedus:
**What happened**:

As part of v0.86.0 the release, the dotnet portable executable parser were introduced, the implementation does seem to be somewhat flawed, as it uses the dontnet projects `FileDescription` attribute as the name for the item. This does not seem like the best attribute to use, as both `ProductName` and `InternalName` fits better. 

This results in scans reporting "names" that seems to require immediate action to remove.

**What you expected to happen**:

I would expect any project that depends on [cli-prompt](https://github.com/Seldaek/cli-prompt) as PHP Composer dependency to have `Hidden Input` or `hiddeninput` as a item, rather than `Reads from stdin without leaking info to the terminal and outputs back to stdout`

**Steps to reproduce the issue**:

Scan a image with said Composer dependency:

```shell
docker run --rm \
  ghcr.io/anchore/syft:v0.86.0 \
    ghcr.io/linuxserver/grav:1.7.42.3-ls124
```
Which results in the offending result as: 

```
NAME                                                                              VERSION                 TYPE         
Reads from stdin without leaking info to the terminal and outputs back to stdout  1, 0, 0, 0              dotnet
```

**Anything else we need to know?**:

Assumed line causing issues: [parse_dotnet_portable_executable.go#L43](https://github.com/anchore/syft/blob/v0.86.0/syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go#L43)

Published SBOM from scan: [package_versions.txt](https://github.com/linuxserver/docker-grav/blob/1.7.42.3-ls124/package_versions.txt)

Tracked-down project with the description: [hidden-input](https://github.com/Seldaek/hidden-input/blob/b898988c86aa096db0edfb2bc671ab2c792562ba/src/version.rc#L18)

SuglBug2k:
I think i have a similar problem with the prometheus-net library. The published dll has close to no meta data.
In v.0.85.0 the output is:
 `musl-utils                                                  1.2.4-r1                apk      `
`prometheus-net                                              8.0.0                   dotnet `
`prometheus-net.AspNetCore                                   8.0.0                   dotnet `
`scanelf                                                     1.3.7-r1                apk     `

and in v.0.86.0 the name is empty what cases errors in my further processing:
`NAME                                                        VERSION                 TYPE   `
`                                                            8.0.0                   dotnet  `
[...]

PR #2273, opened 2023-10-30
fix: update parsing logic to remove empty space

spiffcs:
## Summary
While we wait for clarity on https://github.com/anchore/syft/issues/2029 this PR submits a small fix for the following issues:
- https://github.com/anchore/syft/issues/2241
- https://github.com/anchore/syft/issues/2255

A user submitted a valid `*.dll` where ProductName was the correct package name. This PR keeps the current behavior while falling back to `ProductName`. It also adds a check so that `name` or `version` can no longer be `" "`.


willmurphyscode (commented):
This seems like it was a subtle bug. Is it possible to get this behavior under unit test so that we don't re-introduce the bug?

kzantow (approved):
LGTM but agree with @willmurphyscode about adding a basic test for this case somehow

spiffcs:
@kzantow @willmurphyscode I redid my initial cases with the ones provided by the community in the two issues linked to this PR. I don't think it's perfect, and the full solution will come from https://github.com/anchore/syft/issues/2029, but this should get us generating valid SBOMS in more instances where these edge cases pop up around the metadata

PR #2278, opened 2023-11-01
add PURLs when scanning Gradle lock files

robbiev:
This adds PURLs when scanning Gradle lock files.

Unintuitively the correct PURL type appears to be `maven` as opposed to `gradle`. See https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst

- `gradle` for Gradle plugins
- `maven` for Maven JARs and related artifacts


spiffcs (approved):

spiffcs:
Thanks for the updates test cases and fix @robbiev!

PR #2288, opened 2023-11-03
Rename `file.Location.VirtualPath` to `AccessPath`

### LINKED ISSUE #1835: Rename Location virtualPath to accessPath
Today we have the concept of a location of a RealPath (absolute path to a file where there are no symlinks, which is canonical) and a VirtualPath (what path we requested to get to this RealPath). Given that the virtual path is all about how a path was accessed, a more accurate name would be `AccessPath`. Making this change would break most consumers that are depending on this value, and duplicating the field temporarily seems like overkill. However, we could as a hold over period allow for an application configuration where we allow for the old expression in the JSON output only.

wagoodman:
Closes #1835 

Note: must be merged after #2287


spiffcs:
#2287 merged and this refactor does not break any current tests - LGTM

PR #2294, opened 2023-11-06
include image labels in cycloneDX SBOM

### LINKED ISSUE #2267: store image annotations inside the SBOM

I would like to be able to store a [container's labels/annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md) inside a syft generated SBOM.

For example, the image `mongo:4.4` contains these labels/annotations which are set by its base image `ubuntu:focal`

```
$ docker inspect mongo:4.4 | jq ".[0].Config.Labels"
{
  "org.opencontainers.image.ref.name": "ubuntu",
  "org.opencontainers.image.version": "20.04"
}
```

which could be stored inside a CycloneDX SBOM in the properties field like

```
"properties": [
        {
            "name": "syft:image:annotation:org.opencontainers.image.ref.name", 
            "value": "ubuntu"
        },
        {
            "name": "syft:image:annotation:org.opencontainers.image.version", 
            "value": "20.04"
        }
        ...
 ]
```

**Why is this needed**:

It would be nice to see what base image a container is using, as well as other metadata. This solves https://github.com/anchore/syft/issues/1199.

noqcks:
This PR adds a `metadata.properties` field with information about image labels to cycloneDX SBOMs

resolves #2267 

image


wagoodman (dismissed):
nice add 🙌  I only added an extra test case

wagoodman (approved):
I made one more update to remove the change to the snapshot tests since they updated the golden bin that is captured in the repo. Changing this isn't a bad thing, but it seemed unnecessary since it could be captured in a lower level unit test (which I just added). Snapshot tests are great for change detection with how the component is wired together, and less about correctness of values in different circumstances.

PR #2297, opened 2023-11-07
chore: setup release task before calling go releaser

spiffcs:


Taskfile.yaml
```diff
@@ -455,6 +455,8 @@ tasks:
       - "{{ .TOOL_DIR }}/chronicle -vvv > CHANGELOG.md"
       - cmd: "cat CHANGELOG.md"
         silent: true
+      - cmd: "cat .goreleaser.yaml >> {{ .TMP_DIR }}/goreleaser.yaml"
+        silent: true
```
kzantow, line 459: I don't think we need to copy this file for releases, since we don't need to prepend an alternate target directory. Instead, we should update the [RELEASE_CMD on line 28](https://github.com/anchore/syft/pull/2297/files#diff-1552e0a6f970b2191f0e8ef34ac62e41815fc6cdbc213921e915a042e500f03aR28) to remove the non-default `--config`, e.g.:
```
RELEASE_CMD:  "{{ .TOOL_DIR }}/goreleaser release --clean --release-notes {{ .CHANGELOG }}"
```

kzantow (approved):
👍

PR #2315, opened 2023-11-13
Retrieve remote licenses using pom.properties when there is no pom.xml

coheigea:
The Java remote licensing check only works if there is a pom.xml in the archive. This PR adds support to retrieve a remote pom from maven central using the information in pom.properties. With this it correctly finds a license for the snmp4j dependency which is a sub-component of https://repo1.maven.org/maven2/org/apache/servicemix/bundles/org.apache.servicemix.bundles.snmp4j/2.6.3_1/


coheigea:
@spiffcs Reminder about a review for this please :-)

spiffcs (approved):

PR #2316, opened 2023-11-13
Add binary classifiers for MySQL and MariaDB 

duanemay:
Add binary classifiers for MySQL and MariaDB


syft/pkg/cataloger/binary/test-fixtures/classifiers/positive/mariadb-10.6.15/mariadb
wagoodman, line 1: can you instead add the smallest possible snippets of these binaries? For example: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/binary/test-fixtures/classifiers/positive/perl-5.37.8/perl

wagoodman: I added a new commit that removes these large binaries. Since we squash merge PRs there is no need to scrub history to merge this PR.

wagoodman (approved):
thanks for the additional binary classifiers 🙌

PR #2324, opened 2023-11-14
fix: capture content written to stdout outside of report

kzantow:
This PR adds handling to capture everything written to `stdout`. Syft has a guarantee (maybe an implicit one, for now) that all the content written to `stdout` is part of the report generated by the tool, such as ensuring JSON is valid JSON. Unfortunately, some libraries may write to stdout directly, [such as this](https://github.com/saferwall/pe/blob/712c8311391d9756c8c96f05004e616e658eb5c9/dotnet.go#L730). In order to guarantee stdout only includes contents for the report, this PR captures this non-report output and trace logs it instead.

NOTE: although this PR uses the `TRACE` logging level, the behavior with a TUI when set to `WARN` would look like this:
```
$ syft -o json --platform windows/amd64 registry:mcr.microsoft.com/windows/servercore:ltsc2019 >stdout
 ✔ Parsed image                                                                                                 sha256:c632661e39bb365cf34bfdd943baa4ce5ba1ebdd713d790d1af0b9da20ae0bea
 ⠸ Cataloging packages             [0 packages]  
[0135]  WARN unhandled metadata table 38 File offset 0x944426 cols 5
[0136]  WARN unhandled metadata table 38 File offset 0xafdd82 cols 5
[0137]  WARN unhandled metadata table 38 File offset 0x1e720 cols 1
[0137]  WARN unhandled metadata table 38 File offset 0x45a cols 1
```


cmd/syft/cli/cli.go
```diff
@@ -16,38 +17,39 @@ import (
...
 	return app
 }
 
 // Command returns the root command for the syft CLI application. This is useful for embedding the entire syft CLI
 // into an existing application.
 func Command(id clio.Identification) *cobra.Command {
-	_, cmd := create(id)
+	_, cmd := Create(id, os.Stdout)
 	return cmd
 }
```
kzantow, line 31: `Application` get get removed, I think. Does it make sense to just remove `Command` also, in favor of using the `Create` function?

wagoodman: only the command or the application should be used, so exposing `Create` which returns both is confusing (which is why there wasn't a single function for this to begin with). Why not extend the existing `Application` to accept a writer?

cmd/syft/internal/ui/capture.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// Capture is a utility to capture output written to an os.File, particularly suited for capturing output of
+// os.Stdout or os.Stderr. Usage is fairly straightforward: the return value is a function which stops the current
+// capture and restores the previous file. For example:
+//
+//	restore := ui.Capture(&os.Stdout, writer)
+//	// here, stdout will be captured and redirected to the provided writer
+//	restore() // block until the output has all been sent to the writer and restore the original os.Stdout
+func Capture(target **os.File, writer io.Writer) (close func()) {
```
wagoodman, line 18: nit: this is internal so grain of salt... changing this to 
```suggestion
func CaptureStdout(writer io.Writer) (close func()) {
```
would make this harder to misuse. You could still keep the internal `capture` and pivot tests against that function.

wagoodman (approved):

PR #2331, opened 2023-11-16
Parse Python licenses from LicenseFile entry in the Wheel Metadata

coheigea:
Syft only finds Python licenses if the Wheel metadata contains a "License" entry. This PR adds support to also parse the License-File entry. With this PR, Syft finds licenses for these packages:

> /PyGithub:1.58.2,LGPL-3.0-only
> /django-watson:1.6.3,BSD-3-Clause
> /setuptools:65.5.1,MIT

Further PRs will follow with enhanced support.

### DIFF
--- a/syft/pkg/cataloger/python/package.go
+++ b/syft/pkg/cataloger/python/package.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 
 	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/internal/licenses"
+	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/file"
 	"github.com/anchore/syft/syft/pkg"
 )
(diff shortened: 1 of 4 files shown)


syft/pkg/cataloger/python/package.go
```diff
@@ -55,13 +56,29 @@ func newPackageForRequirementsWithMetadata(name, version string, metadata pkg.Py
...
 }
 
-func newPackageForPackage(m parsedData, sources ...file.Location) pkg.Package {
+func newPackageForPackage(resolver file.Resolver, m parsedData, sources ...file.Location) pkg.Package {
+	var licenseSet pkg.LicenseSet
+	if m.Licenses != "" {
+		licenseSet = pkg.NewLicenseSet(pkg.NewLicensesFromLocation(m.LicenseLocation, m.Licenses)...)
+	} else if m.LicenseLocation.Path() != "" {
+		// If we have a license file then resolve and parse it
+		found, _ := resolver.FilesByPath(m.LicenseLocation.Path())
```
wagoodman, line 65: can you add `log.Trace` calls to show when there are errors here?
For instance
```golang
log.WithFields("error", err).Tracef("unable to fetch python license path %q", path)
```

Similar comment for when parsing fails a few lines below.

coheigea: Yep done

coheigea:
bump for a review @wagoodman

coheigea:
Reminder on this one please @spiffcs @wagoodman

spiffcs (approved):

PR #2338, opened 2023-11-20
Adding the ability to retrieve remote licenses for yarn.lock

coheigea:
Retrieve remote licenses from https://registry.npmjs.com for packages defined in yarn.lock.

It's very slow when enabled so maybe there is some caching mechanism that could be employed...


spiffcs:
👋 sorry @coheigea I can take a look at this today - thanks for all the support on the license gaps! I really appreciate the PR for this.

spiffcs:
This is so far a 👍 from me! I've got a few more things to wrap up for the day, but will take more time tonight to write tests for this and clean up the conflicts

spiffcs:
Merged and fixed conflicts with some of the recent plumbing changes in how catalogers work. I made the new yarn config optional for this merge commit and hard coded the online to false to just get everything square. Taking another pass now with updating tests and cataloger architecture to match the latest and greatest. Thanks again for all the contributions @coheigea!

spiffcs:
Just rewiring the cataloger a little now to make it easier to test the network call now like we did the archive parser for maven. This means adding a new baseURL for the config where users who host packages at other registries can set that up. It also allows us to do dependency injection for the internal `servemux` URL that get's setup during the test

syft/pkg/cataloger/javascript/options.go
```diff
@@ -0,0 +1,31 @@
...
+	return g
+}
+
+// NewCatalogerOpts create a NewCatalogerOpts with default options, which includes:
+// - searchRemoteLicenses is false
+func NewCatalogerOpts() CatalogerConfig {
+	g := CatalogerConfig{}
+
+	return g
+}
```
kzantow, line 19: I think this function should get removed in favor of consumers using the `DefaultCatalogerConfig()` below, which is consistent with the other cataloger configs, e.g. `javaCataloger.DefaultArchiveCatalogerConfig()` and `golangCataloger.DefaultCatalogerConfig()`

kzantow (approved):
👍

PR #2349, opened 2023-11-22
fix(java): improve identification for org.eclipse.platform artifacts

westonsteimel:


coheigea (commented):
It's not just eclipse that is an exception, but some other places I found are:
 - pkg:maven/org.eclipse.persistence.javax.persistence.osgi.Activator/javax.persistence@2.1.0 (Syft uses "persistence" instead of "javax.persistence")
 - pkg:maven/com.ibm.icu/icu@50.1.1.v201304230130 (icu instead of com.ibm.icu)

In both cases, the Bundle-SymbolicName is correct, maybe instead of having an exception for eclipse it could check if Bundle-SymbolicName is present and if it matches with the filename first?

wagoodman (approved):

PR #2365, opened 2023-11-28
Normalize cataloger configuration patterns

wagoodman:
This PR ensures the following is true amongst all catalogers:
- Constructors return `pkg.Cataloger` interfaces over instances
- Any config structs expose options to json marshallers and to anyone to set values plainly
- Configs follow similar naming convention (e.g. `*CatalogerConfig`)
- Default config constructors are present and are named similarly (e.g. `Default*CatalogerConfig`)

This also adds a common struct for any cataloger within `syft/cataloging`, which will be more heavily used in #1383 .

The breaking change made in this PR is renaming existing cataloger option structs within the `java` and `golang` packages.

(This work was broken off of PR #1383)


kzantow (approved):
LGTM except the use of "central", unless I'm misunderstanding it

PR #2369, opened 2023-11-28
Generalize UI events for cataloging tasks

wagoodman:
This PR replaces the set of events relating to cataloging down to a single event type, setting up #1383 for generalized cataloging tasks. There is no functional change to the appearance of the UI with this code change:

Screenshot 2023-11-28 at 4 24 53 PM

In the future this code reorg sets up for a set of cataloger tasks that may be related to one another in a tree-like form, expressed in the UI (see an example in https://github.com/anchore/bubbly/pull/29).

Detailed changes:
- removes the `PackageCatalogerStarted`, `FileMetadataCatalogerStarted`, and `FileDigestsCatalogerStarted` event types, handlers, and related tests. These are being replaced with the existing `CatalogerTaskStarted` event.
- Extends the `CatalogerTaskStarted` event handling with a new stateful handler. The handler creates and maintains a single `catalogerTaskState` instance. This instance is responsible for injecting new models into the bubbletea event loop.
- Adds a `simpleHandler` adapter for the existing handlers that do not raise up `tea.Cmd`s with new models (a change brought in by anchore/bubbly#29)
- Expands `monitor.GenericTask` to be able to specify a task ID, a parent ID, and limited formatting options
- Updates the bubbletea test helpers (such as `flatten`) based on the latest bubbletea lib changes
- Made a small enhancement to the UI to comma-format long numbers show as counts of objects (such as packages and digests)

This is a breaking change since some underlying event types have been removed from the API.

(This is work broken off of #1383 )


syft/file/cataloger/filemetadata/cataloger.go
```diff
@@ -20,52 +20,57 @@ func NewCataloger() *Cataloger {
...
 		locations = resolver.AllLocations()
 	} else {
 		locations = func() <-chan file.Location {
 			ch := make(chan file.Location)
 			go func() {
-				close(ch)
+				defer close(ch)
 				for _, c := range coordinates {
-					ch <- file.NewLocationFromCoordinates(c)
+					locs, err := resolver.FilesByPath(c.RealPath)
```
wagoodman, line 31: this fixes the cataloger to always reference locations from the resolver, not to create new ones based on bespoke paths.

wagoodman: I'll move this change to another PR

wagoodman: https://github.com/anchore/syft/pull/2370

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -123,7 +117,15 @@ func (c *goLicenses) getLicensesFromRemote(moduleName, moduleVersion string) ([]
 
 	proxies := remotesForModule(c.opts.Proxies, c.opts.NoProxy, moduleName)
 
-	fsys, err := getModule(c.progress, proxies, moduleName, moduleVersion)
+	prog := monitor.StartCatalogerTask(monitor.GenericTask{
+		Title: monitor.Title{
+			Default: "Downloading go mod",
+		},
+		HideOnSuccess: true,
+		ParentID:      "TODO", // TODO: setting this to non-empty will cause the progress bar to be nested, but this needs to be more specific
```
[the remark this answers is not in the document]
wagoodman: yeah, I need to remove the value, but this will be taken care of fully in #1383 . The impact is that this will be another sibling task and not nested under the "cataloging packages" task.

kzantow (approved):
👍

PR #2383, opened 2023-12-01
fix(java): improve identification for org.elasticsearch artifacts

westonsteimel:
Elasticsearch jars currently end up with incorrect groupids because syft ends up taking the groupid from the `Implementation-Title` field in the manifest, and those contain a `#` suffix on them like
`org.elasticsearch.plugin#lang-mustache;6.8.15`.  Syft strips off the component after the `;`, but this enhances the logic to also strip off the part after the `#`

An example jar that can be used to demonstrate this is https://repo1.maven.org/maven2/org/elasticsearch/elasticsearch/8.10.2/elasticsearch-8.10.2.jar

This also adds manual mapping for the major elasticsearch components which are named uniquely enough to go in the map as a fallback in any scenarios where the other logic fails


syft/pkg/cataloger/common/cpe/java_groupid_map.go
```diff
@@ -1,250 +1,320 @@
...
+	"elasticsearch-spark_2.11":          "org.elasticsearch",
+	"elasticsearch-ssl-config":          "org.elasticsearch",
+	"elasticsearch-storm":               "org.elasticsearch",
+	"elasticsearch-tdigest":             "org.elasticsearch",
+	"elasticsearch-transport-memcached": "org.elasticsearch",
+	"elasticsearch-transport-thrift":    "org.elasticsearch",
+	"elasticsearch-transport-wares":     "org.elasticsearch",
+	"elasticsearch-x-content":           "org.elasticsearch",
+	"elasticsearch-yarn":                "org.elasticsearch",
+	"elasticsearch-rest-client":         "org.elasticsearch.client", - -"elasticsearch-rest-client-sniffer": "org.elasticsearch.client", - -"elasticsearch-rest-high-level-client": "org.elasticsearch.client",
```
willmurphyscode, line 201: It looks like some diff output ended up in here or something @westonsteimel

westonsteimel: Thanks, that piece should be fixed now

Issue #1821, opened 2023-05-16
Catalog ELF metadata


wagoodman:
This topic has come up in a few different places, in the community meeting and in conversations with @nurmi and @westonsteimel ... so I wanted to capture some of those conversations here.

It is possible to [add metadata directly into ELF binaries](https://systemd.io/ELF_PACKAGE_METADATA/ ) via linker flags . This is beginning to be leveraged [by distros](https://www.fedoraproject.org/wiki/Changes/Package_information_on_ELF_objects) to encode OS and packaging level information directly into the binaries delivered. 

Adding a new cataloger that can read this metadata and capture it for all files that store data in this approach would be ideal.

For reference, there is existing tooling to read metadata from an ELF: https://github.com/systemd/package-notes/

To try this out in the go ecosystem:
```
# in the root of the syft repo
CGO_ENABLED=0 go build -o .bin/ -ldflags "-a -s -linkmode external -extld gcc -extldflags '-static -v -Xlinker --package-metadata={\"type\":\"go\",\"name\":\"syft\",\"version\":\"v0.60.0\"}'" -a ./cmd/syft
```

westonsteimel:
I think it would be useful to discuss what other keys might be useful to capture to try and drive some standards around use for non os package manager binaries

The ones I can think of would be something like:
- purl
- cpe
- sourceRepository
- commit

Also what might make sense for the `type`?  It could just follow the type from the purl spec, or could be something more generic like`static-binary` or `dynamic-binary` or something else entirely.

westonsteimel:
I think it would be useful to discuss what other keys might be useful to capture to try and drive some standards around use for non os package manager binaries

The ones I can think of would be something like:
- purl
- cpe
- sourceRepository
- commit

Also what might make sense for the `type`?  It could just follow the type from the purl spec, or could be something more generic like`static-binary` or `dynamic-binary` or something else entirely.

Also, in the case of APK type packages it would be useful to capture the upstream information.  For instance, if the APK package is something like `python3-flask` it would be useful to understand what the purl and/or cpe are for the `flask` pypi package since both should be used when trying to match to vulnerabilities based on how the alpine security fix feed is intended to be used

westonsteimel:
As an example I have updated my personal container build of curl to include this information: https://github.com/westonsteimel/container-library-curl/blob/main/stable/Dockerfile#L48

```
readelf -n /usr/local/bin/curl

Displaying notes found in: .note.package
  Owner                Data size 	Description
[...]

PR #2396, opened 2023-12-06
draft - static-binary cataloger

brian-ebarb:
In regards to Issue: https://github.com/anchore/syft/issues/1821, we are trying to come up with a solution. This isn't complete, I have some issues getting licenses.go to work properly, and in general I'm new to Syft and not sure if this is even the proper way to go about this. Also unit tests aren't currently included in this PR but they are coming.  Just looking for some help to finish this off.

### DIFF
--- a/syft/pkg/cataloger/static_binary/scan_binary.go
+++ b/syft/pkg/cataloger/static_binary/scan_binary.go
@@ -1,0 +1,228 @@
+package static_binary
+
+import (
+	"bytes"
+	"debug/elf"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/internal/unionreader"
+)
+
+type fileInfo struct {
+	Name    string `json:"name"`
+	System  string `json:"system"`
+	Version string `json:"version"`
+}
+
+func scanFile(c *staticBinaryCataloger, reader unionreader.UnionReader, filename string) ([]byte, []string, error) {
+	bi, err := io.ReadAll(reader)
+	if err != nil {
+		log.WithFields("file", filename, "error", err).Trace("unable to read binary")
+		return bi, nil, err
+	}
+	//br is our byte reader
+	br := bytes.NewReader(bi)
+	//make a new elf file we can play with
+	e, err := elf.NewFile(br)
+	if e != nil {
+
+		symbols, err := e.ImportedLibraries()
+		if err != nil {
+			log.Debugf("unable to read elf binary: %s", err)
+			symbols = nil
+		}
+
+		//notes will be the raw data pulled from the section
+		noteSection := e.Section(".note.package")
+		if noteSection != nil {
+			notes, err := noteSection.Data()
+			var metadata pkg.StaticBinaryPackageMetadata
+			newerr := json.Unmarshal(notes, &metadata)
+			if newerr != nil {
+				log.Debugf("noteSection not nill, json didn't partse error")
+			}
+			if notes == nil {
+				log.Debugf("unable to read .note.package")
+			}
+
+			return notes, symbols, err
+		} else {
+			//This file did not have a note.package section. We must infer info another way
+			ver, system := fileNameTemplateVersionMatcher(c, filename)
+
+			//System is the same thing as namespace in the metadata we send over
+			data := fileInfo{
+				Name:    filename,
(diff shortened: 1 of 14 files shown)


kzantow:
Hi @brian-ebarb thanks much for working on this addition! I see this PR is just in draft right now, so will have a look after you htink it's ready. But I noticed the DCO check is failing so I thought I'd just note the contribution guidelines (to get DCO passing / etc.): https://github.com/anchore/syft/blob/main/CONTRIBUTING.md#sign-off-your-work and also note that `make lint-fix` will often help with the static analysis failures. Cheers!

brian-ebarb:
> Hi @brian-ebarb thanks much for working on this addition! I see this PR is just in draft right now, so will have a look after you think it's ready. But I noticed the DCO check is failing so I thought I'd just note the contribution guidelines (to get DCO passing / etc.): https://github.com/anchore/syft/blob/main/CONTRIBUTING.md#sign-off-your-work and also note that `make lint-fix` will often help with the static analysis failures. Cheers!

Was mostly hoping for some early help from @wagoodman to get us over the finish line. I'm totally new to Syft and I just want to make sure the logic is sound and then I can clean things up for proper contribution.

syft/pkg/cataloger/static_binary/scan_binary.go
```diff
@@ -0,0 +1,228 @@
...
+	//First we should check currect directory
+	libPath = "./"
+	f, fileErr_ := os.Open(libPath + symbol)
+	if fileErr_ != nil {
+		//Next we try lib64
+		libPath = "/lib64/"
+		f, fileErr_ = os.Open(libPath + symbol)
+		if fileErr_ != nil {
+			//then we try ld_library_path variable
+			libPath = os.Getenv("LD_LIBRARY_PATH")
```
wagoodman, line 100: using environment variables directly within the cataloger isn't recommended, instead this should be passed in via a configuration struct, reducing side effects within catalogers

syft/pkg/cataloger/static_binary/scan_binary.go
```diff
@@ -0,0 +1,228 @@
...
+				log.Debugf("noteSection not nill, json didn't partse error")
+			}
+			if notes == nil {
+				log.Debugf("unable to read .note.package")
+			}
+
+			return notes, symbols, err
+		} else {
+			//This file did not have a note.package section. We must infer info another way
+			ver, system := fileNameTemplateVersionMatcher(c, filename)
```
wagoodman, line 58: this feels like an approach similar to our existing binary cataloger

syft/pkg/cataloger/static_binary/scan_binary.go
brian-ebarb: Makes sense, we already have a config struct that lets user pass a path, so maybe we just remove this condition? The user options path overrides all this stuff.

wagoodman:
Thanks for submitting @brian-ebarb , really looking forward to further enhancing binary analysis in syft. Here are my thoughts so far:

- capturing a "static-binary" (application) vs a "static-library" is too fine-grained for the package cataloging paths. Really we’re raising up packages which could represent either -- this is how it is for all other catalogers as well. I think the package types trying to capture these differences isn't inherently useful.
- There are two kinds of operations that can be interpreted from the original issue: “cataloging elf metadata” vs “taking elf metadata and synthesizing a package from it”. This former is akin to the original issue, the latter is akin to extending the existing binary cataloger (which synthesizes packages from patterns found in binaries)
    - In terms of “cataloging elf metadata” we could try and keep requirements in mind and work on that in parallel with https://github.com/anchore/syft/pull/2443 (about ELF security features recently opened) which is also capturing additional file information from a binary (but not necessarily a package)
    - In terms of “taking elf metadata and synthesizing a package from it” I think we should try and incorporate this into the current binary cataloger. I'm trying to understand how this wouldn't fit into the current cataloger (even if the current cataloger needs adjusting)
 - It seems like the properties from `StaticBinaryPackageMetadata` and `StaticBinaryLibraryMetadata` properties are hard coded to some assumed `.node.package` attributes we're expecting, however, the nature of that section is that anything could be there. Should we consider capturing more fields if they are present? 
 - I'm trying to understand the nature of the license cataloging component of this PR. At a surface glance it seems that all licenses found would be applied to all packages found (I might have missed a subtle detail here).

wagoodman (approved):
Looks great! We'll work on adding relationships to this next 🙌 

I'll hold off merging until we're certain on the final commit attributions -- I'll check back in tomorrow.

spiffcs:
marked as ready to review based on a sync call this morning getting the final fixtures and code pushed to the remote branch

Issue #2266, opened 2023-10-29
Java pom.xml , no version


bj1116:
Java pom.xml , no version .
Example:
pom.xml
![image](https://github.com/anchore/syft/assets/15172224/f8082c72-1667-40fe-b241-e7b7220ff48e)

syft.exe xxx/
![image](https://github.com/anchore/syft/assets/15172224/0abd3560-045e-43cf-babe-da967d20e605)

tgerla:
Hi @bj1116, can you share more information about your environment, what version of Syft you are running, and can you please share the steps and files so that we can reproduce this problem locally? Thanks!

coheigea:
Please attach the pom.xml file

coheigea:
Ah yes there is a gap here when a dependency version is declared in a parent pom. I'll take a look at it.

coheigea:
Fix submitted here: https://github.com/anchore/syft/pull/2423

It needs the configuration:

```
java:
   # when running across pom.xml files that could have more information, syft will
   # explicitly search maven for license information by querying the online pom when this is true eg:
   # https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-starter-test/3.1.5/spring-boot-starter-test-3.1.5.pom
   # this option is helpful for when the parent pom has this information,
   # but it is not accessible from within the final built artifact
   maven-url: "https://repo1.maven.org/maven2"
   max-parent-recursive-depth: 8
   use-network: true
```

tgerla:
Thanks, @coheigea. I'll close this issue. Please feel free to open a new one if anyone runs into more trouble. Thanks!

PR #2409, opened 2023-12-08
Add the option to retrieve remote licenses for projects defined in a …

coheigea:
…maven pom

Similar to work I did to remotely retrieve licenses from Maven Central for dependencies in jars, this adds the same functionality when you're scanning a pom.xml instead. I moved the maven utilities from archive_parser into a separate file to maintain a cleaner separation. 

You can reproduce with the project referenced on this issue https://github.com/anchore/syft/issues/2266
Once PR is merged I'll look at getting the versions from the parent poms as well.


spiffcs (approved):

spiffcs:
You're so awesome - thank you so much @coheigea for the improvements on this part of the code

PR #2416, opened 2023-12-11
Added OpenSSL binary matcher

LaurentGoderre:


spiffcs:
Thanks for this one as well!

cc @anchore/tools if we're ok with also moving the testing of these classifiers up to `test/` so that we don't commit any more binaries and can fetch the artifacts just in time for testing.

willmurphyscode:
> moving the testing of these classifiers up to test/ so that we don't commit any more binaries and can fetch the artifacts just in time for testing

So the suggestion is testing the binary classifiers against a test image instead of committing a binary @spiffcs? Just wanted to make sure I understood the suggestion. I agree we should avoid committing binaries and testing against a container with the target binary installed makes a lot of sense to me.

If we wanted to keep unit tests around, we could also have a byte slice literal that contains a little noise plus the bytes the regex will match on. This might also be useful for debugging the regexes and checking them for edge cases.

spiffcs (approved):

PR #2417, opened 2023-12-11
Add binary classifier for the ERLang interpreter

LaurentGoderre:


spiffcs:
👋 Thanks for the pull request @LaurentGoderre! 

I saw this comment added with respect to `nginx`
```
// [NUL]nginx version: nginx/1.25.1 - fetches '1.25.1'
// <artificial>[NUL]/usr/local/src/otp-25.3.2.7/erts/
```

I'm a little confused to how the above relates to the string we're trying to match on to identify the erlang interpreter. Can you explain how the two are linked?

Also cc @wagoodman - I know we've been discussing a bit offline about not having binary or other artifacts contributed into the syft test directories. Is this a good candidate to lift these tests into the top level `test/` directory where we can fetch the artifacts we want to classify rather than have them commited?

LaurentGoderre:
@spiffcs it isn't forgot to remogbe the line from the copy-paste. Fixed!

spiffcs (approved):

PR #2435, opened 2023-12-15
chore: remove magic files from test-fixtures for classifier unit tests

spiffcs:
## Summary
This PR aims to remove the dependence on "magic file" being committed to the repository as a function of testing the binary cataloger.

This should allow contributors to show clearly the strings/content their new classifier contributions are being tested against rather than having to commit a `magic file` with the byte contents.

### Example of a magic file
On main you can run:
```
$file -v syft/pkg/cataloger/binary/test-fixtures/classifiers/positive/go-1.14/go

file-5.41
magic file from /usr/share/file/magic
``` 

#### Change Summary
- Remove mostly all positive case magic files from test fixtures
- Update tests to assert on the contents of the string the individual classifiers were written against
    - note: this involves creating a temporary file and deleting after the test has run
 - Separate cases into `dynamic` and `content` sections of the test.

### Notes for reviewer
- I noticed that the byte diff for some of the deleted files showed a subtraction of bytes rather than the file being deleted. Is there some other motion we need here to fully delete the artifacts?


spiffcs:
@LaurentGoderre - I'm closing this PR for now. There was some internal commentary on a different rework where we move everything over to the dynamic style if we make changes here. 

In response, I'm going to just go through and get you're merged with the current pattern before we do this rework.

PR #2443, opened 2023-12-18
Detect ELF security features

### LINKED ISSUE #2434: Identify security-features-of-interest within binaries
(From a conversation with @joshbressers )

For each binary discovered in the scanned artifact (or if the artifact is a binary) denote the memory protection mechanisms being used (e.g. stack canaries, nx, relro, etc). This would be extra metadata attached to each file object (not package) that denotes the specific properties of interest.

**Why is this needed**:
This may be useful for folks that are trying to use SBOMs as a way to ensure that memory-safe practices are being followed and better figure a risk posture (without needing access to the original artifact for common cases).

**Additional context**:
- https://opensource.com/article/21/6/linux-checksec
- https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart
- https://github.com/slimm609/checksec.sh/blob/36a67b388193d26289773f4153cb3b9f5bf5c9ec/src/functions/proccheck.sh
- [CSI memory safety software guidance](https://media.defense.gov/2022/Nov/10/2003112742/-1/-1/0/CSI_SOFTWARE_MEMORY_SAFETY.PDF).

wagoodman:
This is a PoC for detecting specific ELF security features from discovered binaries (mostly ported from the bash script logic found in [checksec](http://github.com/slimm609/checksec.sh)). Specifically:
- stack canaries
- NX
- RelRO
- PIE
- SafeStack
- CFI
- Clang Fortify
- ~selfrando~ (deferred)

This extends the file attributes for that binary in the SBOM with a new `Executable` section.

## Deferred
- add macho security features https://github.com/anchore/syft/issues/2497
- add pe security features https://github.com/anchore/syft/issues/2498
- add selfrando fixture and write detection mechanism in https://github.com/anchore/syft/issues/2496

Closes #2434


syft/file/cataloger/executable/test-fixtures/checksec
wagoodman, line 1: probably should curl this in the dockerfile instead of checking it into the repo

syft/file/executable.go
```diff
@@ -0,0 +1,51 @@
...
+	PE    ExecutableFormat = "pe"
+
+	RelocationReadOnlyNone    RelocationReadOnly = "none"
+	RelocationReadOnlyPartial RelocationReadOnly = "partial"
+	RelocationReadOnlyFull    RelocationReadOnly = "full"
+)
+
+type Executable struct {
+	// Format denotes either ELF, Mach-O, or PE
+	Format ExecutableFormat `json:"kind" yaml:"kind" mapstructure:"kind"`
```
wagoodman, line 20: replace `kind` with `format`

wagoodman:
```diff
# diff schema/json/schema-13.0.0.json schema/json/schema-13.0.1.json
3c3
<   "$id": "anchore.io/schema/syft/json/13.0.0/document",
---
>   "$id": "anchore.io/schema/syft/json/13.0.1/document",
553a554,592
>       ]
>     },
>     "ELFSecurityFeatures": {
>       "properties": {
>         "symbolTableStripped": {
>           "type": "boolean"
>         },
>         "stackCanary": {
>           "type": "boolean"
>         },
>         "nx": {
>           "type": "boolean"
>         },
>         "relRO": {
>           "type": "string"
>         },
>         "pie": {
>           "type": "boolean"
>         },
>         "dso": {
>           "type": "boolean"
>         },
>         "safeStack": {
>           "type": "boolean"
>         },
>         "cfi": {
>           "type": "boolean"
>         },
>         "fortify": {
>           "type": "boolean"
>         }
>       },
>       "type": "object",
>       "required": [
>         "symbolTableStripped",
>         "nx",
>         "relRO",
>         "pie",
>         "dso"
599a639,652
>       ]
>     },
>     "Executable": {
>       "properties": {
>         "format": {
>           "type": "string"
>         },
>         "elfSecurityFeatures": {
>           "$ref": "#/$defs/ELFSecurityFeatures"
>         }
>       },
>       "type": "object",
>       "required": [
>         "format"
626a680,682
>         },
>         "executable": {
>           "$ref": "#/$defs/Executable"

```

cmd/syft/cli/options/file.go
```diff
@@ -22,6 +23,10 @@ type fileContent struct {
 	Globs              []string `yaml:"globs" json:"globs" mapstructure:"globs"`
 }
 
+type fileExecutable struct {
+	Globs []string `yaml:"globs" json:"globs" mapstructure:"globs"`
```
wagoodman, line 27: open question: does this approach make sense to folks? Right now this allows for sub-selecting by path glob, and keys off of the file.Selection to do nothing when there is no selection specified. I think it's a little awkward since:
- this cataloger implies that only binaries are scanned
- it does not yet account for the owned-files selection, which I think should be added to this PR.

internal/constants.go
```diff
@@ -3,5 +3,5 @@ package internal
 const (
 	// JSONSchemaVersion is the current schema version output by the JSON encoder
 	// This is roughly following the "SchemaVer" guidelines for versioning the JSON schema. Please see schema/json/README.md for details on how to increment.
-	JSONSchemaVersion = "14.0.0"
+	JSONSchemaVersion = "14.0.1"
```
spiffcs, line 6: Is this "MINOR when you add backwards-compatible functionalities" instead of "PATCH when you make backwards-compatible bug fixes"?

syft/file/cataloger/executable/cataloger.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to get file locations for binaries: %w", err)
+	}
+
+	locs, err = filterByGlobs(locs, i.config.Globs)
+	if err != nil {
+		return nil, err
+	}
+
+	prog := catalogingProgress(int64(len(locs)))
```
spiffcs, line 54: This might not be for this PR but for the general syft v1.0 work. I noticed that this, the fileDigest, and the fileContent, are all catalogers where we invoke or have some concept of catalogingProgress. Is this something we're trying to remove from the library code and only surface in the cmd package or is it ok that this concern is still in a handful of cataloger?

https://github.com/search?q=repo%3Aanchore%2Fsyft+catalogingProgress&type=code

spiffcs (commented):
Looks good! I agree with the security features additions and don't have comments about this going in as the first pass for ELF type with extra executable formats to follow.

I had two questions about the schema version and some presentation concerns that are in a few main library catalogers and how we want to look at those going into v1.0

internal/constants.go
wagoodman: This isn't following [semver](https://semver.org/), but instead [schemaver](https://docs.snowplow.io/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver/) https://github.com/anchore/syft/blob/main/schema/json/README.md#versioning

> Given a version number format MODEL.REVISION.ADDITION:
> - MODEL: increment when you make a breaking schema change which will prevent interaction with any historical data
> - REVISION: increment when you make a schema change which may prevent interaction with some historical data
> - ADDITION: increment when you make a schema change that is compatible with all historical data

spiffcs (approved):

syft/file/cataloger/executable/cataloger.go
wagoodman: they all have subtly different implementations so it isn't worth trying to share the implementation / export. (But yeah, it's also not an external concern even if they were the same.)

PR #2445, opened 2023-12-18
Add Redmine detection

LaurentGoderre:


kzantow (commented):
Thanks for the contribution, @LaurentGoderre! We would probably like to see a test for this classifier before merging it.

However, this particular package (redmine) seems a bit different than what most of the current binary matching is doing: rather than binary, this looks like a ruby project distributed with source code? Would this be better suited as some sort of ruby cataloger?

LaurentGoderre:
This is a pattern that I am encountering more frequently where the dependencies are detected but not the root app. It doesn't tend to happen in some ecosystem like Javascript where the `package.json` is the app manifest but ruby doesn't really have that. It only has the dependency via gem. The same applies to PHP (via Composer)

syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -249,6 +249,14 @@ var defaultClassifiers = []classifier{
 		Package: "mariadb",
 		PURL:    mustPURL("pkg:generic/mariadb@version"),
 	},
+	{
+		Class:    "redmine-binary",
+		FileGlob: "**/version.rb",
```
wagoodman, line 254: redmine ultimately is not a binary, so I don't think this belongs as a binary classifier -- is redmine a gem? ( I need to learn more about it), if so this feels like an enhancement to the ruby-base cataloger

wagoodman: ahh I just saw the comments above on this -- I still think this shouldn't be detected by the binary cataloger

LaurentGoderre: Ok, fair enough

witchcraze:
I have interested in this approach.
As binary cataloger is very powerfull, I have felt this will detect source code based products like wordpress, redmine and many CMS. And syft/grype have posibility to detect many of vuluneravility for them via CPE...

willmurphyscode:
@witchcraze and @LaurentGoderre, can you help me understand what is missing here?

I wanted to know whether syft could already find Redmine, so I did the following test:

First, make a Dockerfile:
``` Dockerfile
FROM ruby:3.3.0-slim-bookworm

RUN apt-get update && apt-get install -y build-essential

RUN gem install redmine_cli
```
Then, build it: 
``` sh
docker build -t test-redmine .
```
Then, see whether syft finds redmine:
```
$ syft -q test-redmine | grep redmine
redmine_cli                 0.7.0               gem
redmine_rest                0.8.0               gem
```

It looks like the ruby cataloger already finds redmine. I apologize if I'm missing something obvious, since I've never used Redmine, but can you all help me understand what I'm doing wrong? Is there a situation where Syft should find Redmine but doesn't?

westonsteimel:
@willmurphyscode , try one of the official docker images like https://hub.docker.com/layers/library/redmine/alpine3.18/images/sha256-ddfa98a0682593417fe87055e9a699620fdfb5fb13316a2601ab6f41cd98cd74?context=explore

redmine the server component is installed via a direct url, so syft doesn't find anything useful by itself

westonsteimel:
If you look at that image, the working directory is the redmine source directory, but that file @LaurentGoderre was setting the binary cataloger to look at is /usr/src/redmine/lib/redmine/version.rb which gives something syft can parse the module ane version from.  I guess the question would be is that standard enough that we could create a generic cataloger for ruby projects that could raise it up as a ruby package?

```
cat /usr/src/redmine/lib/redmine/version.rb
# frozen_string_literal: true

require 'redmine/scm/adapters/subversion_adapter'

module Redmine
  # @private
  module VERSION
    MAJOR = 5
    MINOR = 1
    TINY  = 1

    # Branch values:
    # * official release: nil
    # * stable branch:    stable
    # * trunk:            devel
    BRANCH = 'stable'

    # Retrieves the revision from the working copy
    def self.revision
      if File.directory?(File.join(Rails.root, '.svn'))
        begin
          path = Redmine::Scm::Adapters::AbstractAdapter.shell_quote(Rails.root.to_s)
          if `#{Redmine::Scm::Adapters::SubversionAdapter.client_command} info --xml #{path}` =~ /commit\s+revision="(\d+)"/
            return $1.to_i
          end
        rescue
          # Could not find the current revision
        end
      end
      nil
    end

    REVISION = self.revision
    ARRAY    = [MAJOR, MINOR, TINY, BRANCH, REVISION].compact
    STRING   = ARRAY.join('.')

    def self.to_a; ARRAY  end
    def self.to_s; STRING end
  end
end
```

PR #2446, opened 2023-12-19
Replace `packages` command with `scan`

### LINKED ISSUE #516: Syft command restructure

wagoodman:
One of the original ideas for syft was to expand the scope of the tooling to generally answer the question "what are all of the things that are in my artifact" even for things that might not fit into an SBOM. This has sense been focused down to items that can be expressed in an SBOM (which is why the secrets cataloger was removed for instance). For the same reason, as described in #516 , the `packages` command no longer makes sense as a name (since we are always creating SBOMs from what we scan, instead of only showing packages or <other noun here that was not implemented>). This PR addresses this by  deprecating the current `packages` command in favor of `scan` (still aliasing to the root command). 

So any current usage of the `packages` command command...
```
syft packages <input>
```

...should be migrated to `scan`:
```
syft scan <input>
```

A warning banner has been added to help users transition to the new command:

Screenshot 2023-12-19 at 11 13 00 AM

Closes #516  (generally follows the final suggestion in https://github.com/anchore/syft/issues/516#issuecomment-996999477 )

### DIFF
--- a/cmd/syft/cli/cli.go
+++ b/cmd/syft/cli/cli.go
@@ -79,7 +79,7 @@ func create(id clio.Identification, out io.Writer) (clio.Application, *cobra.Com
 	// we also need the command to have information about the `root` options because of this alias
 	scanCmd := commands.Scan(app)
 
-	// root is currently an alias for the packages command
+	// root is currently an alias for the scan command
 	rootCmd := commands.Root(app, scanCmd)
 
 	// add sub-commands
--- a/cmd/syft/cli/cli.go
+++ b/cmd/syft/cli/cli.go
@@ -77,22 +77,25 @@ func create(id clio.Identification, out io.Writer) (clio.Application, *cobra.Com
 
 	// since root is aliased as the packages cmd we need to construct this command first
 	// we also need the command to have information about the `root` options because of this alias
-	packagesCmd := commands.Packages(app)
+	scanCmd := commands.Scan(app)
 
-	// rootCmd is currently an alias for the packages command
-	rootCmd := commands.Root(app, packagesCmd)
+	// root is currently an alias for the packages command
+	rootCmd := commands.Root(app, scanCmd)
 
 	// add sub-commands
 	rootCmd.AddCommand(
-		packagesCmd,
+		scanCmd,
+		commands.Packages(app, scanCmd), // this is currently an alias for the scan command
 		commands.Attest(app),
 		commands.Convert(app),
 		clio.VersionCommand(id),
 		cranecmd.NewCmdAuthLogin(id.Name), // syft login uses the same command as crane
 	)
 
-	// explicitly set Cobra output to the real stdout to write things like errors and help
-	rootCmd.SetOut(out)
+	// note: we would direct cobra to use our writer explicitly with rootCmd.SetOut(out) , however this causes
+	// deprecation warnings to be shown to stdout via the writer instead of stderr. This is unfortunate since this
+	// does not appear to be the correct behavior on cobra's part https://github.com/spf13/cobra/issues/1708 .
+	// In the future this functionality should be restored.
 
 	return app, rootCmd
 }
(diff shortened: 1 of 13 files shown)


cmd/syft/cli/cli.go
```diff
@@ -77,22 +77,25 @@ func create(id clio.Identification, out io.Writer) (clio.Application, *cobra.Com
 
 	// since root is aliased as the packages cmd we need to construct this command first
 	// we also need the command to have information about the `root` options because of this alias
-	packagesCmd := commands.Packages(app)
+	scanCmd := commands.Scan(app)
 
-	// rootCmd is currently an alias for the packages command
-	rootCmd := commands.Root(app, packagesCmd)
+	// root is currently an alias for the packages command
```
willmurphyscode, line 82: Should this comment be updated with the new name?

test/cli/scan_cmd_test.go
```diff
@@ -281,6 +281,24 @@ func TestPackagesCmdFlags(t *testing.T) {
 				assertSuccessfulReturnCode,
 			},
 		},
+		// Testing packages alias //////////////////////////////////////////////
```
willmurphyscode, line 284: Thanks! I was going to ask for this if it wasn't here.

willmurphyscode (commented):
Also, an overall thing to double-check: Now configs change name as a result of this name? (I can't remember whether fangs/clio will generate configs that have the command name in the yaml keys or env var name.)

cmd/syft/cli/cli.go
wagoodman: updated

cmd/syft/cli/commands/scan.go
```diff
@@ -0,0 +1,255 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			restoreStdout := ui.CaptureStdoutToTraceLog()
+			defer restoreStdout()
+
+			return runScan(id, opts, args[0])
+		},
+	}, opts)
+}
+
+func validatePackagesArgs(cmd *cobra.Command, args []string) error {
```
kzantow, line 97:
```suggestion
func validateScanArgs(cmd *cobra.Command, args []string) error {
```

kzantow (approved):
LGTM

PR #2449, opened 2023-12-20
Add classifier for Drupal

LaurentGoderre:


willmurphyscode:
Hi @LaurentGoderre, thanks very much for the PR!

Syft already has PHP cataloging defined over at https://github.com/anchore/syft/tree/main/syft/pkg/cataloger/php, and it seems to find Drupal:

``` sh
syft -q drupal:latest | rg '^drupal'
drupal/core                                                          10.1.7                          php-composer
drupal/core-composer-scaffold                                        10.1.7                          php-composer
drupal/core-project-message                                          10.1.7                          php-composer
drupal/core-recommended                                              10.1.7                          php-composer
```

Do you have an example of an image or repo with drupal in it that syft doesn't detect? Or is there something else you were trying to fix?

LaurentGoderre:
@willmurphyscode I was unsure if detecting the parts was enough to encapsulate the whole application.

LaurentGoderre:
@willmurphyscode I just double check and confirmed these packages cover the whole app.

PR #2468, opened 2024-01-04
Fix diff error obfuscating binary test failures message

LaurentGoderre:
Before this change there were a lot of fatal errors

i.e.
```
Exception has occurred: panic
"MetadataType: does not exist"
```

```
Exception has occurred: panic
"cannot handle unexported field at Locations({pkg.Package}.Locations)...
```

```
Exception has occurred: panic
"cannot handle unexported field at {pkg.Package}.Licenses.set:\n\t\"g...
```

After the change:

```
Error:      	locations do not match; expected: haproxy1 got: haproxy
```

### DIFF
--- a/syft/pkg/cataloger/binary/cataloger_test.go
+++ b/syft/pkg/cataloger/binary/cataloger_test.go
@@ -953,8 +953,8 @@ func assertPackagesAreEqual(t *testing.T, expected pkg.Package, p pkg.Package) {
 				cmp.Transformer("Locations", func(l file.LocationSet) []file.Location {
 					return l.ToSlice()
 				}),
-				cmpopts.IgnoreUnexported(pkg.Package{}, file.Location{}),
-				cmpopts.IgnoreFields(pkg.Package{}, "CPEs", "FoundBy", "MetadataType", "Type"),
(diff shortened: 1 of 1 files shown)


syft/pkg/cataloger/binary/cataloger_test.go
```diff
@@ -954,7 +954,7 @@ func assertPackagesAreEqual(t *testing.T, expected pkg.Package, p pkg.Package) {
 					return l.ToSlice()
 				}),
 				cmpopts.IgnoreUnexported(pkg.Package{}, file.Location{}),
-				cmpopts.IgnoreFields(pkg.Package{}, "CPEs", "FoundBy", "MetadataType", "Type"),
+				cmpopts.IgnoreFields(pkg.Package{}, "CPEs", "FoundBy", "Metadata", "Type", "Locations", "Licenses"),
```
wagoodman, line 957: question for reviewer: does it make sense to try and reuse some of the test code from here https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/internal/pkgtest/test_generic_parser.go

wagoodman: side notes:
- metadata cannot be ignored
- locations should probably not be ignored

PR #2469, opened 2024-01-04
Add ability to extend the binaries cataloguers

LaurentGoderre:


syft/pkg/cataloger/binary/cataloger.go
```diff
@@ -12,16 +12,42 @@ import (
 
 const catalogerName = "binary-cataloger"
 
-func NewCataloger() pkg.Cataloger {
-	return &Cataloger{}
+type CatalogerConfig struct {
+	classifiers []classifier
+	overwrite   bool
```
wagoodman, line 17: from the community meeting: we can probably remove the override option here, export the default classifiers in a separate function (from https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/binary/default_classifiers.go ), and export the classifier struct https://github.com/anchore/syft/blob/4c20a74d2fc6774b7745b432995d5ff7836b90a0/syft/pkg/cataloger/binary/classifier.go#L27 and evidenceMatcher type.

LaurentGoderre:
I'm implementing the changes in multiple commit and will squash when it's ready to merge.

wagoodman:
@LaurentGoderre I pushed a few changes:
- slight renames of variables/functions
- unexport the `classifiers` attribute on the cataloger, but did export the `Classifiers` attribute on the config
- removed package attributes that can be changed on the `Classifier` to make the surface area there a little more defendable

wagoodman:
Also it seems like there are a few commits that are not signed off:
```
```

kzantow (approved):
👍

PR #2470, opened 2024-01-04
Drop go-multierror dependency

jonjohnsonjr:
Doing a little bit of dependency graph gardening and found this in our deps. We should be able to drop it since `errors.Join` is ~equivalent.

### DIFF
--- a/cmd/syft/cli/commands/scan.go
+++ b/cmd/syft/cli/commands/scan.go
@@ -1,9 +1,9 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
 
-	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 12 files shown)


jonjohnsonjr:
That's quite the conflict :(

spiffcs:
Hey @jonjohnsonjr! I'm so sorry this had such a huge conflict and we missed this PR. We're definitely looking to drop this dependency in the future as soon as the dust settles on the syft v1.0 work. Thanks for bringing this to our attention and you should see it dropped from your dependency graphs sooner rather than later. 

Thanks again for taking the time here and apologies we couldn't get this merged in a timely fashion

PR #2515, opened 2024-01-19
Swap format uses of io.ReadSeeker for io.Reader

wagoodman:
This changes the interface for format decoding to accept a more forgiving `io.Reader` instead of requiring a `io.ReadSeeker` for all calls. When provided a `io.ReadSeeker` then the execution will not change, however, when provided a reader without seeking capabilities under the hood it will decorate the reader with a bytes buffer initialized with the reader contents. This can be swapped in the future for a reader that can just-in-time reason about ad-hoc Read and Seek calls to do the same thing more efficiently, however, for the initial implementation a simpler route was taken. This saves callers from having to do the "read-content-and-wrap-with-readseeker" in cases where they do not have a read seeker handy (which is common for non-file IO).

### DIFF
--- a/syft/format/cyclonedxjson/decoder.go
+++ b/syft/format/cyclonedxjson/decoder.go
@@ -10,6 +10,7 @@ import (
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/format/common/cyclonedxhelpers"
 	"github.com/anchore/syft/syft/format/internal/cyclonedxutil"
+	"github.com/anchore/syft/syft/format/internal/stream"
 	"github.com/anchore/syft/syft/sbom"
 )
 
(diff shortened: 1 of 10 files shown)


syft/format/cyclonedxjson/decoder.go
```diff
@@ -25,10 +26,12 @@ func NewFormatDecoder() sbom.FormatDecoder {
 	}
 }
 
-func (d decoder) Decode(reader io.ReadSeeker) (*sbom.SBOM, sbom.FormatID, string, error) {
-	if reader == nil {
-		return nil, "", "", fmt.Errorf("no SBOM bytes provided")
+func (d decoder) Decode(r io.Reader) (*sbom.SBOM, sbom.FormatID, string, error) {
+	reader, err := stream.SeekableReader(r)
```
kzantow, line 30: Should wrapping the reader in something so it can be reset happen at the `syft/format`.`Decode` function instead of per-format?

syft/format/cyclonedxjson/decoder.go
```diff
@@ -68,8 +73,7 @@ func (d decoder) Identify(reader io.ReadSeeker) (sbom.FormatID, string) {
 	dec := json.NewDecoder(reader)
 
 	var doc Document
-	err := dec.Decode(&doc)
-	if err != nil {
+	if err = dec.Decode(&doc); err != nil {
 		// maybe not json? maybe not valid? doesn't matter, we won't process it.
```
willmurphyscode, line 77: Would it make sense to return an error here? It might be helpful to be able to surface to the UI so that users have an easier time fixing their document.

kzantow (approved):

syft/format/cyclonedxjson/decoder.go
wagoodman: in the decoding path that makes sense, but this is in the identification path -- there is no guarantee that what was passed in could be even JSON, so raising up errors is noisy.

wagoodman: (If we make a change here it would be an update for another PR)

PR #2517, opened 2024-01-19
Add API examples

### LINKED ISSUE #558: Top-level API should be more composable

wagoodman:
This is a starting point to showing how to use the Syft API to do common tasks such as:
- create an SBOM
- create a custom cataloger
- decode/encode an SBOM object from/to a particular format (e.g. SPDX, CycloneDX, Syft-JSON, etc)
- obtain a source object to catalog

This is partially completed at this point, I'm looking for feedback at this stage regarding two topics:
- what high-level examples do you wish were here but aren't?
- what sharp-edges in the API exist that might need refactoring?

Closes #558

### DIFF
--- a/examples/create_custom_sbom/main.go
+++ b/examples/create_custom_sbom/main.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"context"
+	"crypto"
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/anchore/syft/syft"
+	"github.com/anchore/syft/syft/cataloging"
+	"github.com/anchore/syft/syft/cataloging/filecataloging"
+	"github.com/anchore/syft/syft/cataloging/pkgcataloging"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/sbom"
+	"github.com/anchore/syft/syft/source"
+)
+
+func main() {
+	// automagically get a source.Source for arbitrary string input
+	src := getSource("alpine:3.19")
+
+	// will catalog the given source and return a SBOM keeping in mind several configurable options
+	sbom := getSBOM(src)
+
+	// show a simple package summary
+	summarize(sbom)
+
+	// show the alpine-configuration cataloger results
+	showAlpineConfiguration(sbom)
+}
+
+func getSource(input string) source.Source {
+	// refactor: source.Detect should take pointer? (to allow for nil default)
+	// refactor: keith has suggestions for refactoring the source.Detection flow
+	detection, err := source.Detect(input, source.DetectConfig{
+		// refactor: this is a magic string
+		DefaultImageSource: "docker",
+	})
+
+	if err != nil {
+		panic(err)
+	}
+
+	// refactor: take pointer and allow for nil?
+	src, err := detection.NewSource(source.DetectionSourceConfig{})
+
+	if err != nil {
+		panic(err)
+	}
+
+	return src
+}
+
+func getSBOM(src source.Source) sbom.SBOM {
+	cfg := syft.DefaultCreateSBOMConfig().
+		// run the catalogers in parallel (5 at a time concurrently max)
+		WithParallelism(5).
+		// bake a specific tool name and version into the SBOM
+		WithTool("my-tool", "v1.0").
+		// catalog all files with 3 digests
+		WithFilesConfig(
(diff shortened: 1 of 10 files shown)


examples/create_custom_sbom/main.go
```diff
@@ -0,0 +1,128 @@
...
+		// refactor: this is a magic string
+		DefaultImageSource: "docker",
+	})
+
+	if err != nil {
+		panic(err)
+	}
+
+	// refactor: take pointer and allow for nil?
+	src, err := detection.NewSource(source.DetectionSourceConfig{})
```
wagoodman, line 46: we should have used the default function https://github.com/anchore/syft/blob/c5d15d1d6c2e79f92545cf8e973317deb04b7638/syft/source/detection.go#L93

examples/source_detection/main.go
```diff
@@ -0,0 +1,54 @@
...
+    docker-archive:path/to/yourimage.tar     use a tarball from disk for archives created from "docker save"
+    oci-archive:path/to/yourimage.tar        use a tarball from disk for OCI archives (from Skopeo or otherwise)
+    oci-dir:path/to/yourimage                read directly from a path on disk for OCI layout directories (from Skopeo or otherwise)
+    singularity:path/to/yourimage.sif        read directly from a Singularity Image Format (SIF) container on disk
+    dir:path/to/yourproject                  read directly from a path on disk (any directory)
+    file:path/to/yourproject/file            read directly from a path on disk (any single file)
+
+*/
+
+const imageRef = "alpine:3.19"
```
willmurphyscode, line 28: I think this could use a different name. Maybe even read from argv, so that it's easier to play with and see what different strings result in.

syft/cataloging/filecataloging/config.go
```diff
@@ -72,7 +72,12 @@ func (cfg Config) WithSelection(selection file.Selection) Config {
 	return cfg
 }
 
-func (cfg Config) WithHashers(hashers []crypto.Hash) Config {
+func (cfg Config) WithHashers(hashers ...crypto.Hash) Config {
 	cfg.Hashers = intFile.NormalizeHashes(hashers)
 	return cfg
 }
+
+func (cfg Config) WithContent(content filecontent.Config) Config {
```
willmurphyscode, line 80: I don't understand this method. What is it for, and why is it part of this PR?

wagoodman: in an earlier commit there was a refactor comment highlighting this was missing and difficult to use in the examples.  Since then I didn't end up using it in the example, but added it since I also had the comment that the crypto.Hash list should be variadic in the same PR (which was only a few lines over from this change).

examples/source_detection/main.go
wagoodman: I want folks to be able to run them without needing to know what to pass, but I can change these to respond to argv  with a default if nothing is passed.

willmurphyscode (approved):

PR #2518, opened 2024-01-19
make AllLocations accept a context

### LINKED ISSUE #2487: LocationResolver can leak goroutines
Today the `LocationResolver` returns a channel of locations:
```go
type LocationResolver interface {
	AllLocations() <-chan Location
}
```

However, while iterating though the channel if you return early you will leak a go routine. This should probably be updated to allow for cancellation via context and follow an iterator pattern (soft suggestion):
```go
type LocationResolver interface {
	AllLocations() LocationIterator
}

type LocationIterator interface {
	Next(context.Context) (Location, bool)
}
```

Where the caller would now be able to coordinate cancellation for correct usage:
```go
ctx, cancel := context.WithCancel(ctx)
defer cancel()

iter := resolver.AllLocations()
for {
	location, ok := iter.Next(ctx)
	if !ok {
		break
	}
	// process the location...
}
```

Alternatively AllLocations can be a slice, which is simpler, but may cause memory pressures for images with many files:
```go
type LocationResolver interface {
	AllLocations() [] Location
}
```

willmurphyscode:
The previous implementation would leak a goroutine if the caller of AllLocations stopped iterating early. Now, accept a context so that the caller can cancel the AllLocations iterator rather than leak the goroutine.

Fixes #2487 .


go.mod
```diff
@@ -213,6 +213,7 @@ require (
 	go.opentelemetry.io/otel/metric v1.19.0 // indirect
 	go.opentelemetry.io/otel/trace v1.19.0 // indirect
 	go.uber.org/atomic v1.9.0 // indirect
+	go.uber.org/goleak v1.3.0 // indirect
```
willmurphyscode, line 216: TODO: Why is this listed as indirect? We directly use it ow.

willmurphyscode: Needed  a `go mod tidy` :)

syft/internal/fileresolver/unindexed_directory.go
```diff
@@ -235,10 +238,16 @@ func (u UnindexedDirectory) AllLocations() <-chan file.Location {
 				return nil
 			}
 			p = strings.TrimPrefix(p, "/")
-			out <- file.NewLocation(p)
-			return nil
+			for {
```
willmurphyscode, line 241: I wasn't certain whether this loop was necessary; I'd welcome some feedback from reviewers.

syft/internal/fileresolver/directory_test.go
```diff
@@ -1398,11 +1400,14 @@ func TestDirectoryResolver_DoNotAddVirtualPathsToTree(t *testing.T) {
 }
 
 func TestDirectoryResolver_FilesContents_errorOnDirRequest(t *testing.T) {
+	defer goleak.VerifyNone(t)
```
spiffcs, line 1403: This is very cool!

spiffcs (commented):
Looks good adding context to check for cancel. I added a comment about the select statement you had a question about.

syft/internal/fileresolver/unindexed_directory.go
wagoodman: It doesn't seem necessary. Are you seeing functional differences when removing it?

willmurphyscode: I'll remove it; I think I confused myself about the semantics of a single `select`. (Namely: a single select with no default will block until it has something to do, but once the context is cancelled it will let the goroutine return immediately.)

I'll remove it.

wagoodman (approved):

Issue #1555, opened 2023-02-08
Turn off SBOM cataloger by default


wagoodman:
Today the SBOM cataloger is on by default and has no configurability to select/deselect SBOMs in certain paths,  include/exclude items within SBOMs conditionally, or be able to enrich the package and relationships found in a flexible way. We could consider making the SBOM cataloger opt-in for now until we discover better mechanisms here. This could mean something like this to enable the SBOM cataloger:

```
syft <my-image> --include-sboms
```

There are several options forward here --interested in hearing thoughts here.

luhring:
> We could consider making the SBOM cataloger opt-in for now until we discover better mechanisms here.

Makes sense to me. 👍  I could try taking a stab at this?

kzantow:
Developer note: Since the builder PR landed, the change fairly simple, I think: [updating the SBOM cataloger line](https://github.com/anchore/syft/blob/main/internal/task/package_tasks.go#L118) to remove the `pkgcataloging.ImageTag, pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, pkgcataloging.ImageTag`, (leaving only `"sbom"`). Those tags are used to select default catalogers for image and directory scans. But now the user can just re-enable this using the selection configuration, like `syft --select-catalogers +sbom-cataloger`

PR #2519, opened 2024-01-19
fix: minor cataloger and docs nits

luhring:


internal/task/package_tasks.go
```diff
@@ -115,6 +115,6 @@ func DefaultPackageTaskFactories() PackageTaskFactories {
 			pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, pkgcataloging.InstalledTag, pkgcataloging.ImageTag, "binary"),
 		newSimplePackageTaskFactory(githubactions.NewActionUsageCataloger, pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, "github", "github-actions"),
 		newSimplePackageTaskFactory(githubactions.NewWorkflowUsageCataloger, pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, "github", "github-actions"),
-		newSimplePackageTaskFactory(sbomCataloger.NewCataloger, pkgcataloging.ImageTag, pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, pkgcataloging.ImageTag, "sbom"), // note: not evidence of installed packages
+		newSimplePackageTaskFactory(sbomCataloger.NewCataloger, pkgcataloging.ImageTag, pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, "sbom"), // note: not evidence of installed packages
```
wagoodman, line 118: this is making a functional change from how this had been working; note that the SBOM cataloger had been included in image scans in the past: https://github.com/anchore/syft/blob/a16a4ad6c993aaaa1ec502667c215b0e8c8e51ff/syft/pkg/cataloger/cataloger.go#L62

That being said this change would close out https://github.com/anchore/syft/issues/1555 . (I think this is still being debated about, but we're looking to put this to rest over the next few working days)

The impact is that we should update the description of the PR and link the issue to this PR.

luhring: I'm not sure about that! Does this need two instances of `pkgcataloging.ImageTag`? If so I think I'm not quite grasping what's happening here yet

wagoodman: ah! disregard, I didn't see that it was there twice. Thanks!

wagoodman (approved):

PR #2522, opened 2024-01-19
Deduplicate digests from user configuration

wagoodman:
Today if a user provides duplicate digests via syft application configuration:
```
# .syft.yaml
file:
  metadata:
    digests: ["sha256", "sha256"]
```
or by the api:
```
# this is one of multiple ways to provide this
	syft.DefaultCreateSBOMConfig().
		WithFilesConfig(
			filecataloging.DefaultConfig().
				WithHashers(
					[]crypto.Hash{
						crypto.SHA256,
						crypto.SHA256,
					},
				),
		)
```

This will result in twice the work and output:
```
$ go run ./cmd/syft -q alpine:latest -o json | jq '.files[0].digests'

[
  {
    "algorithm": "sha256",
    "value": "b40e37517d32d4df5d3b3ed924295f118b4965fa6a659d0db43a796a58fdafff"
  }
  {
    "algorithm": "sha256",
    "value": "b40e37517d32d4df5d3b3ed924295f118b4965fa6a659d0db43a796a58fdafff"
  }
]

```

This PR addresses this issue by:
- deduplicating when reading the application configuration, addressing config display concerns
- deduplicating when providing hashes at the top level API
- deduplicating hashes at the point of use or cataloger creation

### DIFF
--- a/cmd/syft/cli/options/source.go
+++ b/cmd/syft/cli/options/source.go
@@ -42,7 +42,7 @@ func (c *fileSource) PostLoad() error {
 	return nil
 }
 
-func (c *imageSource) PostLoad() error {
+func (c imageSource) PostLoad() error {
 	return checkDefaultSourceValues(c.DefaultPullSource)
 }
 
--- a/cmd/syft/cli/options/source.go
+++ b/cmd/syft/cli/options/source.go
@@ -2,6 +2,7 @@ package options
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/scylladb/go-set/strset"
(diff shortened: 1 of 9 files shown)


cmd/syft/cli/options/source.go
```diff
@@ -34,7 +35,14 @@ func defaultSourceConfig() sourceConfig {
...
 
-func (c imageSource) PostLoad() error {
+func (c *fileSource) PostLoad() error {
+	digests := strset.New(c.Digests...).List()
+	sort.Strings(digests)
+	c.Digests = digests
+	return nil
+}
+
+func (c *imageSource) PostLoad() error {
```
willmurphyscode, line 45: Is the change of imageSource receiver from non-pointer to pointer intentional? (I'm not sure it matters; it just looks accidental.)

willmurphyscode (approved):

cmd/syft/cli/options/source.go
wagoodman: let me back that out, it was unintentional (good catch)

PR #2530, opened 2024-01-22
Unexport types and functions cataloger packages

wagoodman:
This PR attempts to reduce the surface area of functions and types exported from the `syft/pkg/cataloger/**` packages.

This additionally adds an integration test to ensure that catalogers added are following conventions:
- Cataloger constructor functions should follow the glob `New*Cataloger`
- Config structs must be named `*Config` and must be paired with a zero parameter function called `Default*Config`

### DIFF
--- a/test/integration/package_cataloger_convention_test.go
+++ b/test/integration/package_cataloger_convention_test.go
@@ -1,0 +1,166 @@
+package integration
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/bmatcuk/doublestar/v4"
+	"github.com/scylladb/go-set/strset"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_PackageCatalogerConventions(t *testing.T) {
+	// look at each package in syft/pkg/cataloger...
+	// we want to make certain that only the following things are exported from the package:
+	// - function matching New*Cataloger (e.g. NewAptCataloger)
+	// - function matching Default*Config
+	// - struct matching *Config
+	//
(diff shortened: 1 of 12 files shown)


test/integration/package_cataloger_convention_test.go
```diff
@@ -0,0 +1,313 @@
...
+	require.NoError(t, err)
+
+	defaultConfigMatches, err := doublestar.Match("Default*Config", export.Name)
+	require.NoError(t, err)
+
+	configMatches, err := doublestar.Match("*Config", export.Name)
+	require.NoError(t, err)
+
+	switch {
+	case constructorMatches:
```
willmurphyscode, line 163: Can we double check that the constructor returns a `pkg.Cataloger` interface?

test/integration/package_cataloger_convention_test.go
```diff
@@ -0,0 +1,313 @@
...
+			if c.wantErr == nil {
+				c.wantErr = assert.NoError
+			}
+			err := validatePackageCatalogerExport(t, "test", c.export)
+			c.wantErr(t, err)
+		})
+	}
+}
+
+func Test_PackageCatalogerConventions(t *testing.T) {
```
willmurphyscode, line 115: This test looks a little bit like a custom linter. It might be worth writing a project-specific linter plugin for this type of assertion in the future.

test/integration/package_cataloger_convention_test.go
wagoodman: I like it -- just pushed an updated (and it found a case in the code I missed for this 👍 )

test/integration/package_cataloger_convention_test.go
wagoodman: I like this a lot, there are a whole class of linters for syft that I think could be helpful

willmurphyscode (approved):

PR #2539, opened 2024-01-24
README.md: add 'sudo' to recommended installation command

chkp-alexgl:


willmurphyscode:
Hey @chkp-alexgl thanks very much for the suggestion. For the reasons @spiffcs posted above, we're not comfortable putting `sudo` in a snippet of shell meant to be copied and pasted. If you'd like to add some language that would help folks less experienced with shells understand how to use the installer, please feel free.

PR #2540, opened 2024-01-24
Allow for RPM modularity to be optional

wagoodman:
Ultimately we need to distinguish between the states of is there a modularity label that can even be expressed vs an empty modularity label value in the RPM DB. This PR makes this adjustment.

The use case for this is really in grype, where there is vulnerability information from redhat that mostly tailored with modularity labels in mind. We should not attempt to use modularity logic if the SBOM which the SBOM was sourced from cannot positively express modularity information.

### DIFF
--- a/internal/constants.go
+++ b/internal/constants.go
@@ -3,5 +3,5 @@ package internal
 const (
 	// JSONSchemaVersion is the current schema version output by the JSON encoder
 	// This is roughly following the "SchemaVer" guidelines for versioning the JSON schema. Please see schema/json/README.md for details on how to increment.
-	JSONSchemaVersion = "13.0.0"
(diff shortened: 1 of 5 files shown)


wagoodman:
Schema diff for reviewers
```diff
# $ diff /Users/wagoodman/code/syft/schema/json/schema-13.0.0.json /Users/wagoodman/code/syft/schema/json/schema-14.0.0.json
3c3
<   "$id": "anchore.io/schema/syft/json/13.0.0/document",
---
>   "$id": "anchore.io/schema/syft/json/14.0.0/document",
1917d1916
<         "modularityLabel",
1974d1972
<         "modularityLabel",

```

westonsteimel (approved):

Issue #2282, opened 2023-11-02
Annotate where each CPE on a package is sourced from


wagoodman:
Today we have a list of CPEs on a package, which classically have always been generated with logic tailored for each ecosystem:
```json
"cpes": [
  "cpe:2.3:a:ansi-regex:ansi-regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi-regex:ansi_regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi_regex:ansi-regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi_regex:ansi_regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi:ansi-regex:5.0.1:*:*:*:*:*:*:*",
  "cpe:2.3:a:ansi:ansi_regex:5.0.1:*:*:*:*:*:*:*"
]
```

Since then a CPE index sourced from the NVD data [has been added](https://github.com/anchore/syft/pull/1897), which means that there is more than one way for a CPE to make it onto a package. Or, more specifically, if there is a single CPE on a package, it's unclear if that CPE was generated or if it was from the new index.

Also, there may be more places that we source this CPE information, say lifting from packaging metadata, other 3rd party indexes, etc. Ideally the SBOM should be able to represent where the information was sourced from without the consumer needing to guess this.

We could make a change to annotate CPEs with this information:
```json
"cpes": [ 
  {
   "value": "cpe:2.3:a:ansi-regex:ansi-regex:5.0.1:*:*:*:*:*:*:*",
   "source": "auto-generated"
  },
  {
   "value": "cpe:2.3:a:ansi:ansi-regex:5.0.1:*:*:*:*:*:*:*",
   "source": "nvd-cpe-index"
  }
]
```

This proposal would necessarily be a breaking change, however, open to other ways to introduce this in a non-breaking way!

willmurphyscode:
> Or, more specifically, if there is a single CPE on a package, it's unclear if that CPE was generated or if it was from the new index.

@wagoodman I don't understand why the case where there is a single CPE is special. It looks like we append, even in cases where there is an exact match in the dictionary:

https://github.com/anchore/syft/blob/b0ab75fd89246ac4ee37abebd4753d8449bfb3aa/internal/task/package_task_factory.go#L112-L117

[...]

PR #2541, opened 2024-01-24
Internalize CPE generation logic

wagoodman:
This PR moves much of the CPE generation logic to internal. ~There is one functional change made: combines the dictionary find and generation into a single `Create` function, this way consumers get the most accurate CPE as possible.~ I've decided to not make any functional changes in this PR.

Should be coordinated with https://github.com/anchore/syft/issues/2282


syft/cpe/cpe.go
```diff
@@ -123,7 +125,7 @@ func normalizeField(field string) string {
 
 	// keep dashes and forward slashes unescaped
 	if field == "*" {
-		return wfn.Any
+		return Any
```
spiffcs, line 128: Is this to decouple us from the `wfn` package in total?

spiffcs (approved):
🟢 no behavior changes and all refactoring; just a question about why we stopped referencing the wfn library's version of any and made our own `""`

syft/cpe/cpe.go
wagoodman: yeah, I didn't see a need to couple our implementation 👍

PR #2543, opened 2024-01-24
Internalize format helpers

wagoodman:
This migrates several spdx and cyclonedx format helpers to an internal package, only preserving the top-level `ToSyftModel` and `ToFormatModel` functions.


syft/format/common/spdxhelpers/to_format_model_test.go
```diff
@@ -777,3 +778,74 @@ func Test_toSPDXID(t *testing.T) {
 		})
 	}
 }
+
+func Test_otherLicenses(t *testing.T) {
```
spiffcs, line 782: Nice - thanks for adding this!

spiffcs (approved):

syft/format/common/spdxhelpers/to_format_model_test.go
wagoodman: this was a migration from elsewhere to help with package imports (the test utility is using a top-level format function, so was causing a cycle)

PR #2547, opened 2024-01-25
Implement golang Purl subpath

LaurentGoderre:
This PR implements the subpath portion of the Purl spec for golang and prevents the subpath from being reported as the package name.

i.e.

```
github.com/coreos/go-systemd/v22
```

Is now reported as `go-systemd` instead of `v22`


wagoodman (approved):
Thanks for the improvement here 🙌

PR #2552, opened 2024-01-26
Record where CPEs come from

### LINKED ISSUE #2282: Annotate where each CPE on a package is sourced from

willmurphyscode:
Syft can get CPE strings from a few places:

1. The NVD Dictionary
2. Generating them based on other fields in the package struct
3. Finding them encoded directly into another manifest or SBOM

This PR changes the cpes field on the package model from a slice of strings to a slice of objects that have a CPE string and a source.

Fixes #2282.


syft/cpe/by_source_then_specificity.go
```diff
@@ -0,0 +1,40 @@
...
+
+import "sort"
+
+type BySourceThenSpecificity []SourcedCPE
+
+func (b BySourceThenSpecificity) Len() int {
+	return len(b)
+}
+
+func (b BySourceThenSpecificity) Less(i, j int) bool {
```
willmurphyscode, line 11: I'd also like some feedback on this sorting logic. Right now, I always have: prefer a CPE from the NVD dictionary if we have one, then one declared, then one syft generated, then any other source. Does that seem reasonable? I know in some places we want the best single CPE, so putting the most specific CPE from the NVD dictionary if we have one makes sense.

syft/pkg/cataloger/binary/package.go
```diff
@@ -19,11 +19,11 @@ func newPackage(classifier Classifier, location file.Location, matchMetadata map
 
 	update := matchMetadata["update"]
 
-	var cpes []cpe.CPE
+	var cpes []cpe.SourcedCPE
 	for _, c := range classifier.CPEs {
 		c.Version = version
 		c.Update = update
-		cpes = append(cpes, c)
+		cpes = append(cpes, c.WithGeneratedSource()) // TODO: is "the binary classifier just picked" the same as "generated"
```
wagoodman, line 26: agreed, generated

cmd/syft/internal/test/integration/go_compiler_detection_test.go
```diff
@@ -12,14 +12,14 @@ func TestGolangCompilerDetection(t *testing.T) {
 		name              string
 		image             string
 		expectedCompilers []string
-		expectedCPE       []cpe.CPE
+		expectedCPE       []cpe.SourcedCPE
```
wagoodman, line 15: maybe `AnnotatedCPE` since this would not preclude folks from adding more fields that are not automatically out of scope of the single "source" field.

wagoodman: Or change `cpe.CPE` to `cpe.Attributes` and `cpe.SourcedCPE` to `cpe.CPE`

syft/cpe/by_source_then_specificity.go
```diff
@@ -0,0 +1,40 @@
...
+		// are considered ties.
+	}
+	iSource := b[i].Source
+	jSource := b[j].Source
+	rankI, rankJ := getRank(iSource), getRank(jSource)
+	if rankI != rankJ {
+		return rankI < rankJ
+	}
+
+	return weightedCountForSpecifiedFields(b[i].CPE) < weightedCountForSpecifiedFields(b[j].CPE)
```
willmurphyscode, line 33: FIXME: this comparison is backwards; more specific CPEs should be first.

willmurphyscode: Also, extract the Less from BySpecificity into a helper method and then call it here to break ties.

syft/format/syftjson/model/package.go
```diff
@@ -32,10 +32,17 @@ type PackageBasicData struct {
...
 	Licenses  licenses        `json:"licenses"`
 	Language  pkg.Language    `json:"language"`
-	CPEs      []string        `json:"cpes"`
+	CPEs      cpes            `json:"cpes"`
 	PURL      string          `json:"purl"`
 }
 
+type cpes []SourcedCPE
+
+type SourcedCPE struct {
```
willmurphyscode, line 41: @wagoodman should this be renamed too?

willmurphyscode:
The rename of cpe.CPE to cpe.Attributes might cascade to other places whose names make less sense now, like cpe.Must returns a cpe.Attributes, not a cpe.CPE, for example. There are also places in logs and comments where we talk about CPEs (meaning strings like cpe:2.3:foo:bar:*:*.... I mostly haven't changed those, since I think the string attributes in log lines or whatever won't be as much help.

Do we think this has the right amount of renaming @wagoodman?

syft/format/syftjson/model/package.go
wagoodman: it would make it easy to follow

syft/cpe/merge_cpes.go
```diff
@@ -23,3 +24,24 @@ func Merge(a, b []CPE) (result []CPE) {
 	sort.Sort(BySpecificity(result))
 	return result
 }
+
+// MergeSourcedCPEs returns unique SourcedCPEs that are found in A or B
+// Two SourcedCPEs are identical if their source and normalized string are identical
+func MergeSourcedCPEs(a, b []CPE) []CPE {
```
wagoodman, line 30: I think this can be simplified to just `Merge()` since it's already in the `cpe` package (thus `cpe.Merge(...)`)

syft/format/syftjson/model/package.go
willmurphyscode: What should the field that's currently called `CPE` be called if we rename? CPE.CPE is a little awkward. The JSON still has `cpe`.

syft/cpe/by_source_then_specificity.go
wagoodman: synced on this last week, ultimately sounds good 👍

syft/pkg/package.go
```diff
@@ -25,7 +25,7 @@ type Package struct {
 	Licenses  LicenseSet       // licenses discovered with the package metadata
 	Language  Language         `hash:"ignore" cyclonedx:"language"` // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
 	Type      Type             `cyclonedx:"type"`                   // the package type (e.g. Npm, Yarn, Python, Rpm, Deb, etc)
-	CPEs      []cpe.CPE        `hash:"ignore"`                      // all possible Common Platform Enumerators (note: this is NOT included in the definition of the ID since all fields on a CPE are derived from other fields)
+	CPEs      []cpe.CPE        `hash:"ignore"`                      // all possible Common Platform Enumerators (note: this is NOT included in the definition of the ID since all fields on a Attributes are derived from other fields)
```
wagoodman, line 28:
```suggestion
	CPEs      []cpe.CPE        `hash:"ignore"`                      // all possible Common Platform Enumerators (note: this is NOT included in the definition of the ID since all fields on a CPE are derived from other fields)
```

syft/pkg/cataloger/internal/cpegenerate/generate.go
```diff
@@ -51,7 +51,7 @@ func GetIndexedDictionary() (_ *dictionary.Indexed, err error) {
 	}
 
 	if indexedCPEDictionary == nil {
-		err = fmt.Errorf("failed to unmarshal indexed CPE dictionary")
+		err = fmt.Errorf("failed to unmarshal indexed Attributes dictionary")
```
wagoodman, line 54:
```suggestion
		err = fmt.Errorf("failed to unmarshal indexed CPE dictionary")
```

syft/pkg/cataloger/internal/cpegenerate/generate.go
```diff
@@ -61,7 +61,7 @@ func GetIndexedDictionary() (_ *dictionary.Indexed, err error) {
 func FromDictionaryFind(p pkg.Package) (cpe.CPE, bool) {
 	dict, err := GetIndexedDictionary()
 	if err != nil {
-		log.Debugf("dictionary CPE lookup not available: %+v", err)
+		log.Debugf("dictionary Attributes lookup not available: %+v", err)
```
wagoodman, line 64:
```suggestion
		log.Debugf("dictionary CPE lookup not available: %+v", err)
```

syft/pkg/cataloger/internal/cpegenerate/filter.go
```diff
@@ -19,39 +19,39 @@ var cpeFilters = []filterFn{
...
+func filter(cpes []cpe.Attributes, p pkg.Package, filters ...filterFn) (result []cpe.Attributes) {
 cpeLoop:
 	for _, c := range cpes {
 		for _, fn := range filters {
 			if fn(c, p) {
 				continue cpeLoop
 			}
 		}
-		// all filter functions passed on filtering this CPE
+		// all filter functions passed on filtering this Attributes
```
wagoodman, line 30:
```suggestion
		// all filter functions passed on filtering this CPE
```

syft/pkg/cataloger/internal/cpegenerate/filter.go
```diff
@@ -9,8 +9,8 @@ import (
 
 const jenkinsName = "jenkins"
 
-// filterFn instances should return true if the given CPE should be removed from a collection for the given package
-type filterFn func(cpe cpe.CPE, p pkg.Package) bool
+// filterFn instances should return true if the given Attributes should be removed from a collection for the given package
```
wagoodman, line 12:
```suggestion
// filterFn instances should return true if the given CPE should be removed from a collection for the given package
```

syft/pkg/cataloger/internal/cpegenerate/dictionary/index-generator/generate.go
```diff
@@ -90,7 +90,7 @@ func filterCpeList(cpeList CpeList) CpeList {
 	return processedCpeList
 }
 
-// normalizeCPE removes the version and update parts of a CPE.
+// normalizeCPE removes the version and update parts of a Attributes.
```
wagoodman, line 93:
```suggestion
// normalizeCPE removes the version and update parts of a CPE.
```

syft/pkg/cataloger/internal/cpegenerate/apk.go
```diff
@@ -23,7 +23,7 @@ type upstreamCandidate struct {
 }
 
 func upstreamCandidates(m pkg.ApkDBEntry) (candidates []upstreamCandidate) {
-	// Do not consider OriginPackage variations when generating CPE candidates for the child package
+	// Do not consider OriginPackage variations when generating Attributes candidates for the child package
```
wagoodman, line 26:
```suggestion
	// Do not consider OriginPackage variations when generating CPE candidates for the child package
```

syft/pkg/cataloger/binary/default_classifiers.go
```diff
@@ -93,7 +93,7 @@ func DefaultClassifiers() []Classifier {
 				`(?m)\x00openjdk\x00java\x00(?P<release>[0-9]+[.0-9]*)\x00(?P<version>[0-9]+[^\x00]+)\x00`),
 			Package: "java",
 			PURL:    mustPURL("pkg:generic/java@version"),
-			// TODO the updates might need to be part of the CPE, like: 1.8.0:update152
+			// TODO the updates might need to be part of the Attributes, like: 1.8.0:update152
```
wagoodman, line 96:
```suggestion
			// TODO the updates might need to be part of the CPE, like: 1.8.0:update152
```

syft/pkg/cataloger/binary/classifier.go
```diff
@@ -225,10 +225,10 @@ func getContents(resolver file.Resolver, location file.Location) ([]byte, error)
 	return contents, nil
 }
 
-// singleCPE returns a []pkg.CPE based on the cpe string or panics if the CPE is invalid
+// singleCPE returns a []cpe.Attributes based on the cpe string or panics if the Attributes is invalid
```
wagoodman, line 228:
```suggestion
// singleCPE returns a []cpe.CPE based on the cpe string or panics if the Attributes is invalid
```

wagoodman (approved):

PR #2569, opened 2024-01-31
fix(dotnet): improve forming dotnet package names

westonsteimel:


syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go
```diff
@@ -175,7 +175,7 @@ var (
 func findName(versionResources map[string]string) string {
 	// PE files found in the wild _not_ authored by Microsoft seem to use ProductName as a clear
 	// identifier of the software
-	nameFields := []string{"ProductName", "FileDescription", "InternalName", "OriginalFilename"}
+	nameFields := []string{"InternalName", "OriginalFilename", "ProductName", "FileDescription"}
```
spiffcs, line 178: Going through old PRs apologies for the ping.

Looks like we have considerations for both these field types now 🥳 
https://github.com/anchore/syft/blob/6452a19009d9ff1e06db3cb6863187b005117cb3/syft/pkg/cataloger/dotnet/package.go#L263-L269

I think we also landed some work on making the `value` here more readable with:
https://github.com/anchore/syft/blob/6452a19009d9ff1e06db3cb6863187b005117cb3/syft/pkg/cataloger/dotnet/package.go#L283-L289

I'm going to close this PR since dotnet presentation seems to be in a way better space than it was previously.

Thanks @westonsteimel for the original contribution here and I'm very sorry we did not land or accept it at the time it was created 😢

spiffcs:
Closing this PR - see above comment about latest state of dotnet and enhancements that have been made to accept both types of name fields now. Apologies for missing this when it first came in.

PR #2609, opened 2024-02-08
chore: free up build space for codeql

spiffcs:


spiffcs:
We're still seeing failures here where codeql is timing out for our go project when trying to autobuild. I'm going to close this and do more research to see if we can figure out what has changed

PR #2610, opened 2024-02-08
Refactor source creation and add `--from` flag

kzantow:
This PR makes the following high-level changes:
* The `source.Detect` function has been removed, instead use:
* the `syft.GetSource` function, which has been added as the entrypoint to use syft defined sources
* 3 source types have been moved to their respective packages:
  * `source/filesource`
  * `source/directorysource`
  * `source/stereoscopesource`
* A `--from` flag is added, allowing the user to select the source(s) to use
* An exported `source/sourceproviders` package has been added which defines the set of tagged providers Syft implements, available by calling the the `All` function

TODO:
- [x] determine if/what replacement test should be implemented for `syft/source/detection_test.go`, as there isn't the same need to test the "detection" aspect, but there probably are still some valid test cases here
- [x] revisit error handling from sources e.g. should we display each file not found with potentially more info?
- [x] is the `testutil.Chdir(t, "..")` acceptable? (at least as part of this PR to limit the changes)
- [x] correct behavior for `generateSBOMForAttestation`?


syft/source/digest_utils.go
```diff
@@ -6,6 +6,6 @@ import (
 	"github.com/anchore/syft/syft/artifact"
 )
 
-func artifactIDFromDigest(input string) artifact.ID {
+func ArtifactIDFromDigest(input string) artifact.ID {
```
wagoodman, line 9: this should probably be moved to `source/internal`

wagoodman (commented):
I think some of the `source/*` packages should be a little more consistent:
- `filesource`: good as is
- `directory` -> `directorysource`
- `stereoscope` -> `stereoscopesource`

syft/source/stereoscope/provider.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/anchore/stereoscope"
+	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/anchore/stereoscope/tagged"
+	"github.com/anchore/syft/syft/source"
+)
+
+const ImageTag = "image"
+
+type stereoscopeSourceProvider struct {
```
wagoodman, line 14:
```suggestion
type Provider struct {
```
given the suggested package name, then it reads as `stereoscopesource.Provider`

syft/source/stereoscope/provider.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+type SourceProviderConfig struct {
+	RegistryOptions *image.RegistryOptions
+	Platform        *image.Platform
+	Alias           *source.Alias
+	Exclude         source.ExcludeConfig
+}
+
+func SourceProviders(cfg SourceProviderConfig) tagged.Values[source.Provider] {
```
wagoodman, line 56:
```suggestion
func Providers(cfg SourceProviderConfig) tagged.Values[source.Provider] {
```
given the package rename

syft/source_providers.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/anchore/stereoscope/tagged"
+	"github.com/anchore/syft/syft/source"
+	"github.com/anchore/syft/syft/source/directory"
+	"github.com/anchore/syft/syft/source/filesource"
+	"github.com/anchore/syft/syft/source/stereoscope"
+)
+
+// SourceProviders returns all the configured source providers known to syft
+func SourceProviders(sourceProviderConfig ...SourceProviderConfig) tagged.Values[source.Provider] {
```
wagoodman, line 14: something doesn't quite feel right with having a stereoscope primitive (`tagged*`) bring raised up on the top-level syft api

syft/source/provider.go
```diff
@@ -0,0 +1,11 @@
+package source
+
+import (
+	"context"
+)
+
+// Provider is the signature used for detectors passed to Detect(cfg, detectors...)
+type Provider interface {
+	Name() string
+	Provide(ctx context.Context, userInput string) (Source, error)
```
wagoodman, line 10:
```suggestion
	ProvideSource(ctx context.Context, userInput string) (Source, error)
```

internal/slices.go
```diff
@@ -0,0 +1,22 @@
+package internal
```
kzantow, line 1: The functions here seem like good candidates for `go-collections`, if we like the ergonomics and can decide on appropriate names.

cmd/syft/internal/commands/attest.go
```diff
@@ -247,7 +247,7 @@ func predicateType(outputName string) string {
 }
 
 func generateSBOMForAttestation(ctx context.Context, id clio.Identification, opts *options.Catalog, userInput string) (*sbom.SBOM, error) {
-	src, err := getSource(opts, userInput, onlyContainerImages)
+	src, err := getSource(ctx, opts, userInput, stereoscopesource.ImageTag)
```
kzantow, line 250: I don't think this is the right thing to do -- aren't there only very specific sources needed here? not just that we got a container imager, but the source needs to be a registry perhaps so cosign can upload an attestation?

cmd/syft/internal/test/integration/catalog_packages_test.go
```diff
@@ -22,15 +22,11 @@ func BenchmarkImagePackageCatalogers(b *testing.B) {
...
-	detection, err := source.Detect(userInput, source.DefaultDetectConfig())
-	require.NoError(b, err)
-
-	theSource, err := detection.NewSource(source.DefaultDetectionSourceConfig())
+	theSource, err := syft.GetSource(context.Background(), tarPath, syft.DefaultGetSourceConfig().WithFromSource("docker-archive"))
 	require.NoError(b, err)
 
 	b.Cleanup(func() {
-		theSource.Close()
+		_ = theSource.Close()
```
wagoodman, line 29: nit:
```suggestion
		assert.NoError(b, theSource.Close())
```

wagoodman: (there are a few of these around the changeset, same comment applied to them)

syft/get_source.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/anchore/syft/syft/source"
+)
+
+func GetSource(ctx context.Context, userInput string, getSourceConfig ...GetSourceConfig) (source.Source, error) {
```
wagoodman, line 12: We should try and be consistent with other top-level functions, such as `CreateSBOM` which takes an optional config with `*ConfigType` instead of `...ConfigType`.

syft/get_source_config.go
```diff
@@ -0,0 +1,115 @@
...
+	c.FromSource = sources
+	return c
+}
+
+func (c GetSourceConfig) WithDefaultImageSource(defaultImageSource image.Source) GetSourceConfig {
+	c.DefaultImageSource = defaultImageSource
+	return c
+}
+
+func (c GetSourceConfig) GetProviders() ([]source.Provider, error) {
```
wagoodman, line 79:
```suggestion
func (c GetSourceConfig) getProviders() ([]source.Provider, error) {
```

Do we need to expose this?

syft/get_source_config.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (c GetSourceConfig) WithDigestAlgorithms(algorithms ...crypto.Hash) GetSourceConfig {
+	c.SourceProviderConfig.DigestAlgorithms = algorithms
+	return c
+}
+
+func (c GetSourceConfig) WithBasePath(basePath string) GetSourceConfig {
+	c.SourceProviderConfig.BasePath = basePath
+	return c
+}
```
wagoodman, line 57: these all wrap options that already exist on the source provider config, so I feel that those should in place of providing these:
```golang 
cfg.WithSourceProviderConfig(DefaultSourceProvider().WithAlias("thing"))
```

syft/pkg/cataloger/erlang/erlang_parser.go
```diff
@@ -152,7 +152,7 @@ func parseErlangNode(data []byte, i *int) (erlangNode, error) {
 		skipWhitespace(data, &offset)
 		c2 := data[offset]
 
-		// Add support for empty lists
+		// Join support for empty lists
```
wagoodman, line 155: probably an unintended change

syft/source/directorysource/directory_source_provider.go
```diff
@@ -0,0 +1,65 @@
...
+		Config{
+			Path:    location,
+			Base:    basePath(l.basePath, location),
+			Exclude: l.exclude,
+			Alias:   l.alias,
+		},
+	)
+}
+
+// FIXME why is the base always being set instead of left as empty string?
```
wagoodman, line 59: I think this is the intended behavior, but could use some clarification in a code comment for sure.

syft/testutil/chdir.go
```diff
@@ -0,0 +1,24 @@
+package testutil
+
+import (
+	"os"
+	"testing"
+)
+
+func Chdir(t *testing.T, dir string) {
```
wagoodman, line 8: this shouldn't be externalized, maybe `syft/internal/testutil`?

syft/source_provider_config.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"crypto"
+
+	"github.com/anchore/stereoscope/pkg/image"
+	"github.com/anchore/syft/syft/source"
+)
+
+// SourceProviderConfig is the uber-configuration for all Syft source providers
+type SourceProviderConfig struct {
```
wagoodman, line 11: missing `WithPlatform`

syft/source_provider_config.go
```diff
@@ -0,0 +1,52 @@
...
+	UserInput        string
```
wagoodman, line 12: I don't think that `UserInput` belongs on this struct since it is not an option on how to get the source or optional metadata to attach. Instead it is a required parameter that cannot have a safe/sane default, so should be passed into a constructor or other function. Instead of adding a constructor (since we already have a nice default constructor) this seems to really fit as a positional argument to `SourceProviders`
```golang
func SourceProviders(imageOrDir string, cfg SourceProviderConfig) []collections.TaggedValue[source.Provider] {
   ...
}
```

syft/get_source.go
```diff
@@ -0,0 +1,102 @@
...
+			if err != nil {
+				errs = append(errs, err)
+			}
+		}
+		if src != nil {
+			// if we have a non-image type and platform is specified, it's an error
+			// if _, ok := src.(*stereoscope.StereoscopeImageSource); !ok && cfg.Platform != nil {
+			//	return src, fmt.Errorf("invalid argument: --platform specified with non-image source")
+			//}
+			return src, nil
```
wagoodman, line 52: the comment doesn't match the code commented out... is this dead? or in need of refactor?

internal/slices.go
wagoodman: Could we use https://pkg.go.dev/golang.org/x/exp/slices#DeleteFunc in place of this remove function?

re: map and filter, yeah we could add that to collections, but I think there are some existing candidates to consider first https://github.com/samber/lo

syft/internal/sourcemetadata/generated.go
```diff
@@ -2,9 +2,13 @@
...
+import (
+	"github.com/anchore/syft/syft/source/directorysource"
+	"github.com/anchore/syft/syft/source/filesource"
+	"github.com/anchore/syft/syft/source/stereoscopesource"
+)
 
 // AllTypes returns a list of all source metadata types that syft supports (that are represented in the source.Description.Metadata field).
 func AllTypes() []any {
-	return []any{source.DirectorySourceMetadata{}, source.FileSourceMetadata{}, source.StereoscopeImageSourceMetadata{}}
+	return []any{directorysource.Metadata{}, filesource.Metadata{}, stereoscopesource.ImageMetadata{}}
```
wagoodman, line 13: though it feels clean to have `noun.Metadata` it ends up coupling implementation packages on the consumer side when type asserting the struct back out. Should we leave these as-is in `source` like we've done with `pkg`?

syft/get_source.go
kzantow: I have a soft preference for the varargs approach, but agree about consistency 💯

syft/get_source_config.go
kzantow: I'm pretty sure that the main reason we'd like to include the `.With***` types of methods is to make specifying the configuration more convenient and discoverable. The most likely usage is to call `syft.GetSource`, so the configuration for that call should be convenient and discoverable, which would be less so if the user had to deal with only the inner config in most cases. I _would_ use struct embedding here, but need the return type to be `GetSourceConfig`, not `SourceProviderConfig`.

internal/slices.go
kzantow: My Google-fu failed pretty miserably on this one; I definitely don't want to be reinventing the wheel!

syft/get_source.go
kzantow: comment was accurate (note the `!ok`); restored this check

wagoodman (approved):
great refactor! 💯

PR #2635, opened 2024-02-13
Enforce breaking change bump major version

wagoodman:
This changes the behavior of [chronicle](https://github.com/anchore/chronicle) when determining the next version during release to allow bumping of the major version when there are breaking changes.


wagoodman:
This should not be merged until all v1.0 changes are in

kzantow (approved):
🎉

PR #2644, opened 2024-02-14
Fix considering base path when ignoring known bad unix paths

### LINKED ISSUE #2627: unable to index filesystem for amazonlinux images
* related to https://github.com/docker/buildkit-syft-scanner/pull/94

**What happened**:

Latest syft (0.104.0) fails to index amazonlinux images when scanning root using `source.NewFromDirectory` https://github.com/docker/buildkit-syft-scanner/blob/7f1fed0d3a1494365ff612d81d1c90f777ce9f81/internal/target.go#L37-L43:

```
unable to get file resolver: unable to create directory resolver: unable to index filesystem path="/run/src/extras/sbom-base/dev/fd/2": unable to index branch="/run/src/extras/sbom-base/dev/fd/2": lstat /proc/1/fd/pipe:[33575]: no such file or directory
```

Be able to scan

Repro with this Dockerfile using BuildKit Syft Scanner: https://github.com/docker/buildkit-syft-scanner/actions/runs/7883917948/job/21511817145?pr=94#step:7:75

```dockerfile
FROM amazonlinux:2 AS base
ARG BUILDKIT_SBOM_SCAN_STAGE=true
RUN yum install -y findutils
```

```
docker buildx build --sbom=true .
```

Tried bumping to latest 0.104.0 and still repro in https://github.com/docker/buildkit-syft-scanner/pull/94.

Let me know if you need more info,, thanks.

**Environment**:
- Output of `syft version`: 0.104.0

cc @cdupuis

wagoodman:
Fixes #2627

This updates the logic when considering paths to ignore/filter out during the indexing phase of the directory file resolver. Currently we ignore known common paths that mount to filesystems that provide no packages (e.g. `/dev` and `/proc`). This currently doesn't consider cases when the filesystem being scanned isn't mounted at root (when there is a `base-path` to consider). For example, when scanning `/some/mount/path` and there is a path `/some/mount/path/dev`, we should be considering that latter path during indexing to be `/dev`. Ultimately the index value is correct, but filter functions do not have the option to observe the value of `base-path` (i.e. `/some/mount/path`) to calculate the relative path of the target.

This PR changes the behavior to allow filter functions to have access to `base-path`. We currently have some filter functions that need to consider the relative path and some that need to consider the real path, so a further-upstream-change of this function could not be made (that is, pass down the relative path into the filter functions).

### DIFF
--- a/syft/internal/fileresolver/directory_indexer.go
+++ b/syft/internal/fileresolver/directory_indexer.go
@@ -21,7 +21,7 @@ import (
 	"github.com/anchore/syft/syft/internal/windows"
 )
 
-type PathIndexVisitor func(string, os.FileInfo, error) error
+type PathIndexVisitor func(string, string, os.FileInfo, error) error
 
 type directoryIndexer struct {
 	path              string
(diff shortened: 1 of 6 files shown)


kzantow (approved):

jpalermo:
FYI, I just submitted a [PR](https://github.com/anchore/syft/pull/2668) to revert this. This causes any path starting with `/proc`, `/sys` or `/dev` to be skipped.

PR #2645, opened 2024-02-14
Survive indexing dead symlinks

wagoodman:
Related to https://github.com/anchore/syft/issues/2627 , we want to be able to survive the indexing process even when we are indexing a new branch in the directory tree that happens to be a symlink pointing to a non existent file. Ultimately this will be recorded as a known-unknown, but for now we at least want to survive when this occurs and allow for the SBOM to be created.


kzantow (approved):

mathrock:
Relatedly, symlink loops are also still fatal.  FWIW, https://github.com/anchore/stereoscope/issues/206 still fails in the latest syft v0.105.0.

PR #2654, opened 2024-02-16
fix Windows file paths in local go mod cache

### LINKED ISSUE #2615: Fundamental Windows incompatibility when accessing the file system
**What happened**:

Syft is at least not working properly on Windows analyzing a project in a directory (syft scan dir:...).
I used variations of SYFT_GOLANG... env variables to enable module remote access.
License information is not present in the Syft JSON SBOM output.

I saw the same symptoms like described in #1933.
I debugged syft . The problems go deeper and do not just affect licenses.go, so I create a dedicated bug for it.

In general, file (path) operations of syft (and I assume also stereoscope) do not support Windows properly.
Path operations for the local file system  are done with functions coming from "path" module instead of "path/filepath".
The Go documention states clear that "path" is only for Unix/Linux, but not for Windows.
Instead, "path/filepath" has to be used, because this module support both Unix-like and Windows OS.

I also recognized that syft is hanging (at the end?) of processing on Windows, if -v is NOT set.

Another funny thing I saw is that you try to access paths that are concatenated taking the local path and typical Unix/Linux locations

[0000] TRACE searching filetree by path path=D:\Projekte\Go\Hello_world\etc\os-release
[0000] TRACE searching filetree by path path=D:\Projekte\Go\Hello_world\usr\lib\os-release
[0000] TRACE searching filetree by path path=D:\Projekte\Go\Hello_world\etc\system-release-cpe
[0000] TRACE searching filetree by path path=D:\Projekte\Go\Hello_world\etc\redhat-release
[0000] TRACE searching filetree by path path=D:\Projekte\Go\Hello_world\bin\busybox

I did not investigate what would need to be considered if the scanned object is not a directory tree.

The readme file should have a clear note that there are issues on Windows until these are resolved.
Use "path/filepath" instead of "path" functions for file system accesses to make it compatible with Windows and Unixes.
Do not try to access Unix-specifc directories I saw in the TRACE output of Syft under Windows.
Syft should terminate properly at the end, even if -v is not used.
Finally #1933 is resolved also.

Execute syft on a Go project with a colorized "Hello world" (one external dependency) on WIndows.

I fixed #1933 in a local branch before seeing that problems are not located just in licenses.go ...
This is the first Go code I inspected and worked on.

**Environment**:
- Output of `syft version`:
Application: syft
Version:    0.104.0
BuildDate:  2024-02-07T20:51:36Z
GitCommit:  bd0cb916df4376a06e56ef7f3cc0da6659d7c2c6
GitDescription: v0.104.0
Platform:   windows/amd64
GoVersion:  go1.21.6
Compiler:   gc
Windows 11

willmurphyscode:
Previously, the file resolver was created from incorrect calls (path.Join instead of filepath.Join) which resulted Go license searches always missing on Windows. Use filepath.* functions when initializing the Go config, and when the unindexed file resolver is being created.

See #2615 - this fixes some of the specific problems reported in that issue, but I'm leaving that issue open to track better integration testing of syft and grype on Windows.

TODOs:

- [ ] unit tests
- [ ] audit other configs and go license paths for incorrect path.Join use where filepath is needed
- [ ] Add Windows CI runner to prevent introduction of this sort of bug in the future (probably separate PRs :) )

### DIFF
--- a/syft/internal/fileresolver/unindexed_directory.go
+++ b/syft/internal/fileresolver/unindexed_directory.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"os"
 	"path"
+	"path/filepath"
 	"slices"
 	"sort"
 	"strings"
@@ -61,11 +62,11 @@ func NewFromUnindexedDirectoryFS(fs afero.Fs, dir string, base string) file.Writ
 	}
 	wd, err := os.Getwd()
 	if err == nil {
-		if !path.IsAbs(dir) {
-			dir = path.Clean(path.Join(wd, dir))
+		if !filepath.IsAbs(dir) {
+			dir = filepath.Clean(filepath.Join(wd, dir))
 		}
-		if base != "" && !path.IsAbs(base) {
-			base = path.Clean(path.Join(wd, base))
+		if base != "" && !filepath.IsAbs(base) {
+			base = filepath.Clean(filepath.Join(wd, base))
 		}
 	}
 	return UnindexedDirectory{
--- a/syft/pkg/cataloger/golang/config.go
+++ b/syft/pkg/cataloger/golang/config.go
@@ -2,7 +2,7 @@ package golang
 
 import (
 	"os"
-	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/mitchellh/go-homedir"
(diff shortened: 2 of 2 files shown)


willmurphyscode:
Getting platform-specific testing spun up has proved pretty time consuming. I think it might be better to release this with tests only proving that it doesn't break mac/linux, and then follow up with tests.

kzantow (approved):

PR #2664, opened 2024-02-22
fix(install): actually return right error codes

hacst:
Previously exit was called with the stdout content of the main function. This did not make sense and
resulted in the script not returning the expected exit codes when main returned a non-zero code.

Strangely enough this broken code was introduced in response to #1566 but I do not see how it ever worked unless the script used to echo its intended exit code.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -694,6 +694,7 @@ main() (
 set +u
 if [ -z "${TEST_INSTALL_SH}" ]; then
   set -u
-  exit $(main "$@")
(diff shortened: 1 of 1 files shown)


kzantow (approved):
Thanks @hacst -- I can confirm this version works on `sh` in `alpine:latest` and macOS. For what it's worth: the version written as `exit $(main "$@")` did work properly on macOS, but not alpine, and this seems to be the right way to do it. 👍

PR #2668, opened 2024-02-24
Revert changes to path filtering

### LINKED ISSUE #2667: New path filtering logic excluding large number of unintended paths
**What happened**:
#2644 introduced new logic to prevent scanning of `unixSystemRuntimePrefixes` even if those paths are not at the root, for example remounted root file systems. The solution however prevents scanning any paths that include a `/proc`, `/dev` or `/sys` within them. For example:
`folder/syslog-scan.sbom.json`
`source_code/system/files`
`source_code/dev/other_files`
are all skipped after this change.

Only `/proc`, `/dev` and `/sys` that are part of fs roots should be excluded, not anything including those strings.

Attempt to scan an sbom file that starts with the string `sys`

**Environment**:
- Output of `syft version`: `0.105.0`
- OS (e.g: `cat /etc/os-release` or similar): Any

jpalermo:
Fixes #2667

#2644 caused the skipping of any files or folders that included the prefix "/proc", "/dev" and "/sys" which was not the intended change and excludes many things that should be scanned.

This reverts commit a909e3cec92da4cc68c95e72f860d233ae0531a2.


manifestori (approved):
High priority, this change is breaking catalogers.
LGTM

PR #2675, opened 2024-02-28
Consider filesystem types for mount points when ignoring system paths

wagoodman:
Implements [the suggestion](https://github.com/anchore/syft/issues/2667#issuecomment-1964744118) from @klakin-pivotal to look at the system mounts and determine if the filesystem types indicate which parts of the overall filetree to ignore when indexing for the directory file resolver.

Replaces #2668 

Fixes https://github.com/anchore/syft/issues/2667

### DIFF
--- a/syft/internal/fileresolver/directory_indexer.go
+++ b/syft/internal/fileresolver/directory_indexer.go
@@ -472,18 +472,20 @@ func keepUnixSystemMountPaths(infos []*mountinfo.Info) []string {
 		if info == nil {
 			continue
 		}
-		// on a unix system you may see:
-		// - tmpfs: a memory-based filesystem (e.g. /tmp, /dev, /sys/firmware)
-		// - devtmpfs: a virtual filesystem that provides access to devices (e.g. /dev)
-		// - shm: a shared memory filesystem (e.g. /dev/shm)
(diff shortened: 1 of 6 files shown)


syft/internal/fileresolver/directory_indexer.go
```diff
@@ -443,11 +450,50 @@ func (r *directoryIndexer) disallowRevisitingVisitor(_, path string, _ os.FileIn
...
+		// - devtmpfs: a virtual filesystem that provides access to devices (e.g. /dev)
+		// - shm: a shared memory filesystem (e.g. /dev/shm)
+		// - proc: a virtual filesystem that provides access to process information (e.g. /proc)
+		// - sysfs: a virtual filesystem that provides access to kernel information (e.g. /sys)
+
+		// on a darwin system you may see:
+		// - devfs: a virtual filesystem that provides access to devices (e.g. /dev)
+
+		switch info.FSType {
+		case "tmpfs", "devfs", "devtmpfs", "shm", "proc", "sysfs":
```
willmurphyscode, line 486: Things I'm seeing in `df -a` that aren't in here but maybe should be (running in a lima ubuntu instance if that matters):

* udev
* devpts (seems to be under `/dev` so I don't know that we need it.)

willmurphyscode (approved):

PR #2679, opened 2024-02-28
Mark duplicated rows in table output

### LINKED ISSUE #2672: Confusing number of packages
**What happened**:

I scanned a project and found 859 packages, like the following:
image

But the table below shows that there are only about 630.
This is often confusing.
I found that the package is deduplicated before printing to the table.
https://github.com/anchore/syft/blob/928511ea0f1449e057e8057e38743d258b22476b/syft/format/table/encoder.go#L64

Should there be an obvious hint?

**Environment**:
- Output of `syft version`: master

wagoodman:
The number of rows in the table output does not always match the TUI package counts when there are duplicate packages found (in terms of a `(name, version, type)` tuple). This PR adjusts the output to specifically indicate which rows are duplicated and by how much.

Closes #2672

### DIFF
--- a/syft/format/table/encoder.go
+++ b/syft/format/table/encoder.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/charmbracelet/lipgloss"
 	"github.com/olekukonko/tablewriter"
 
 	"github.com/anchore/syft/syft/sbom"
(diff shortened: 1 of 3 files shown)


kzantow (approved):

tk103331:
Should there be some summary output showing how many there are in total, how many are deduplicated, and how many are duplicates?

PR #2693, opened 2024-03-06
chore: remove deprecated os.SEEK_SET os.SEEK_CUR

testwill:


spiffcs:
Checked here and yes these are deprecated!
https://pkg.go.dev/os#pkg-constants

Thanks for the quick change

spiffcs (approved):

PR #2696, opened 2024-03-07
chore: reduce duplicate case SwiftPkg

testwill:


syft/pkg/language.go
```diff
@@ -82,7 +82,7 @@ func LanguageByName(name string) Language {
 		return Dart
 	case string(Dotnet), ".net", packageurl.TypeNuget:
 		return Dotnet
-	case packageurl.TypeCocoapods, packageurl.TypeSwift, string(CocoapodsPkg), string(SwiftPkg):
+	case packageurl.TypeCocoapods, packageurl.TypeSwift, string(CocoapodsPkg):
```
spiffcs, line 85: Thanks for the slate of PRs @testwill - I'll take some time today to validate these today. I'm trying to figure out why this was duplicated in the first place:
```
// TypeCocoapods is a pkg:cocoapods purl.
--
66 | TypeCocoapods = "cocoapods"
```
```
CocoapodsPkg          Type = "pod"]
```

Looks like these are two different strings here so it's not a duplicate case =)

Thanks for the due diligence on checking this though.

Closing for now given the different names.

If you think this could be clearer let me know and I can reopen and we can figure out what the chance should be

spiffcs (changes requested):

syft/pkg/language.go
spiffcs: 💯  - my fault I totally switched gears and looked at the wrong change - this is good to go!

spiffcs (approved):

PR #2708, opened 2024-03-13
Adding the ability to retrieve remote licenses from package-lock.json

coheigea:
Extending https://github.com/anchore/syft/pull/2338 to also be able to retrieve remote licenses from package.lock. I only tested v2 package-lock.json.


spiffcs (approved):
Thanks for the PR @coheigea - LGTM 🟢

PR #2715, opened 2024-03-14
Add relationships to ELF packages

brian-ebarb:
This PR adds `DependencyOf` relationships when ELF packages have been discovered by the binary cataloger. The discovered `file.Executable` type has a `[]ImportedLibraries` that's read from the file when discovered by syft. By mapping these imported libraries back to the package collection, syft is able to create relationships showing which packages are dependencies of other packages by just reading metadata from the ELF executable. 

### Testing
The integration test found under `cmd/syft/internal/test/integration/package_binary_elf_relationships_test.go` covers a sample image where the following graph is true:
```
"glibc": {
	"libhello_world.so",
	"syfttestfixture",
},
"libstdc++": {
	"syfttestfixture",
},
"libhello_world.so": {
	"syfttestfixture",
},
```

The above data structure can be read as `glibc IS A DEPENDENCY OF syfttestfixture`

For more information about the fixture see this document:
https://github.com/anchore/syft/blob/4fe0c98d0fe85417be596836d8798b7b10d43b1f/syft/pkg/cataloger/binary/test-fixtures/elf-test-fixtures/README.md


spiffcs:
I've updated the unit tests and SA on this PR - there are maybe 2 more cases to cover, but we're very close.

The one issue in particular I'm keen on testing is if the relationships are still created IF an RPM package shadows/overrides an ELF binary package. We need a way to pass the import paths discovered up to the RPM package so that these don't get dropped

spiffcs:
From my perspective there is one final todo on this PR:

A case needs to be covered where syft does some clean up during `Finalize`

Example: if ELF Package (a) is a part of RPM Package (b), then ELF Package (a) is deleted as a duplicate from the SBOM... this means that RPM Package (b) should all get all of ELF Package (A)s relationships.

I'm working on this now as the final addition. 

If @wagoodman could look at the whole PR and give his final thoughts on if we're missing anything else that would be very much appreciated

spiffcs:
## PR Summary

This PR adds `DependencyOf` relationships when ELF packages have been discovered by the binary cataloger. The discovered `file.Executable` type has a `[]ImportedLibraries` that's read from the file when discovered by syft. By mapping these imported libraries back to the package collection, syft is able to create relationships showing which packages are dependencies of other packages by just reading metadata from the ELF executable. 

### Testing
The integration test found under `cmd/syft/internal/test/integration/package_binary_elf_relationships_test.go` covers a sample image where the following graph is true:
```
"glibc": {
	"libhello_world.so",
	"syfttestfixture",
},
"libstdc++": {
	"syfttestfixture",
},
"libhello_world.so": {
	"syfttestfixture",
},
```

The above data structure can be read as `glibc IS A DEPENDENCY OF syfttestfixture`

For more information about the fixture see this document:
https://github.com/anchore/syft/blob/4fe0c98d0fe85417be596836d8798b7b10d43b1f/syft/pkg/cataloger/binary/test-fixtures/elf-test-fixtures/README.md

internal/relationship/binary/binary_dependencies_test.go
```diff
@@ -0,0 +1,269 @@
...
+// 1. ELF P (a) --> exec (a) ---> imports of (a) --> (b) executable (imported executable) (ELF P --> file)
+
+// Case 3 needs to be covered somewhere else given that NewDependency relationships should already have accessor with
+// de duplicatied packages
+// 3. ELF P (a) is a part of RPM P (b), thus ELF P (a) is deleted from the SBOM... this means that (b) gets all relationships of (a)
+// SEE TODO in finalize
+
+func TestNewDependencyRelationships(t *testing.T) {
+	// coordinates for the files under test
+	glibcCoordinate := file.Coordinates{RealPath: "/usr/lib64/libc.so.6"}
```
wagoodman, line 31: I'd use `file.NewCoordinates(...)` for these

internal/relationship/binary/binary_dependencies_test.go
```diff
@@ -0,0 +1,269 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	"github.com/anchore/syft/internal/sbomsync"
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/sbom"
+)
+
+// GLIBC [RPM]
```
wagoodman, line 17: dead comments now?

internal/relationship/binary/binary_dependencies_test.go
```diff
@@ -0,0 +1,269 @@
...
+		ImportedLibraries: []string{},
+	}
+
+	// dummy executable representation of glibc
+	secondGlibcExecutable := file.Executable{
+		Format:            "elf",
+		HasExports:        true,
+		HasEntrypoint:     true,
+		ImportedLibraries: []string{},
+	}
```
wagoodman, line 110: since these are the same, a single variable could be used

internal/relationship/binary/binary_dependencies_test.go
```diff
@@ -0,0 +1,269 @@
...
+			accessor := newAccesorClosure(tt.packages, tt.coordinateIndex, tt.prexistingRelationships)
+			// given a resolver that knows about the paths of the packages and executables,
+			// and given an SBOM accessor that knows about the packages and executables,
+			// we should be able to create a set of relationships between the packages and executables
+			relationships := NewDependencyRelationships(tt.resolver, accessor)
+			if d := cmp.Diff(tt.want, relationships, cmpopts.IgnoreUnexported(
+				pkg.Package{},
+				artifact.Relationship{},
+				file.LocationSet{},
+				pkg.LicenseSet{},
```
wagoodman, line 243: could we use a common test function exported from `testpkg` package?

internal/relationship/binary/relationship_index.go
```diff
@@ -0,0 +1,59 @@
...
+		exists = true
+	}
+
+	i.typesByFromTo[fromID][toID].Add(string(r.Type))
+	return exists
+}
+
+// add a new relationship to the index, returning true if the relationship is new to the index, false otherwise (thus is a duplicate).
+// nolint:unparam
+func (i *relationshipIndex) add(r artifact.Relationship) bool {
```
wagoodman, line 49: missing a test case for this function

internal/relationship/binary/shared_library_index.go
```diff
@@ -0,0 +1,140 @@
...
+func (i *sharedLibraryIndex) build(resolver file.Resolver, accessor sbomsync.Accessor) {
+	// 1. map out all locations that provide libraries (indexed by the basename)
+	i.libLocationsByBasename, i.allLibLocations = locationsThatProvideLibraries(accessor)
+
+	// 2. for each library path, find all packages that claim ownership of the library
+	i.packagesByLibraryPath = packagesWithLibraryOwnership(resolver, accessor, i.allLibLocations)
+}
+
+// TODO: we might need this
+// func (i *sharedLibraryIndex) owningLibraryLocations(libraryBasename string) file.CoordinateSet {
```
wagoodman, line 40: dead code?

internal/relationship/binary/shared_library_index.go
```diff
@@ -0,0 +1,140 @@
...
+	"path"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/internal/sbomsync"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/sbom"
+)
+
+type sharedLibraryIndex struct {
```
wagoodman, line 13: we should probably test this explicitly

spiffcs:
We're still deleting the wrong packages when going back through the overlaps. I'll look at this again this AM to get a better test written so we can get this correct

internal/sbomsync/builder.go
```diff
@@ -78,6 +80,38 @@ func (b sbomBuilder) AddPackages(p ...pkg.Package) {
...
+		if _, ok := toDelete[rel.From.ID()]; ok {
+			fromID = true
+		}
+		if _, ok := toDelete[rel.To.ID()]; ok {
+			toID = true
+		}
+		// skip (remove) relationships that reference the deleted packages
+		if fromID || toID {
+			continue
+		}
```
wagoodman, line 107:
```suggestion
		var hasID bool
		if _, ok := toDelete[rel.From.ID()]; ok {
			hasID = true
			break
		}
		if _, ok := toDelete[rel.To.ID()]; ok {
			hasID = true
			break
		}
```

internal/sbomsync/builder.go
```diff
@@ -78,6 +80,38 @@ func (b sbomBuilder) AddPackages(p ...pkg.Package) {
 	b.onWriteEvent()
 }
 
+func (b sbomBuilder) DeletePackages(ids ...artifact.ID) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	toDelete := make(map[artifact.ID]struct{})
```
wagoodman, line 87: nit/rename: I was looking at the end of the function based on the tense and realized it was done at the top already
```suggestion
	deleted := make(map[artifact.ID]struct{})
```

internal/relationship/finalize.go
```diff
@@ -1,15 +1,19 @@
...
 	"github.com/anchore/syft/syft/cataloging"
+	"github.com/anchore/syft/syft/file"
 	"github.com/anchore/syft/syft/sbom"
 )
 
-func Finalize(builder sbomsync.Builder, cfg cataloging.RelationshipsConfig, src artifact.Identifiable) {
+func Finalize(resolver file.Resolver, builder sbomsync.Builder, cfg cataloging.RelationshipsConfig, src artifact.Identifiable) {
 	accessor := builder.(sbomsync.Accessor)
 
+	// TODO (also, how should we update the TUI to reflect that we removed packages?)
```
wagoodman, line 15: dead comment now that its implemented

cmd/syft/internal/test/integration/package_binary_elf_relationships_test.go
```diff
@@ -0,0 +1,66 @@
...
+			"libhello_world.so",
+			"syfttestfixture",
+		},
+		"libstdc++": {
+			"syfttestfixture",
+		},
+		"libhello_world.so": {
+			"syfttestfixture",
+		},
+		// TODO: we need to capture devel package
```
wagoodman, line 23: is this stale? or defunct?

spiffcs: This is defunct now we need to test against this. I'l remove the old comment

wagoodman (approved):

PR #2746, opened 2024-04-03
chore: update dependency to use latest commits

spiffcs:
## Summary

Syft should consume the latest community contributions from the `anchore/packageurl-go` library


syft/pkg/cataloger/binary/classifier_cataloger_test.go
```diff
@@ -659,7 +659,7 @@ func Test_Cataloger_PositiveCases(t *testing.T) {
 				Name:      "java",
 				Version:   "19.0.1+10-21",
 				Type:      "binary",
-				PURL:      "pkg:generic/java@19.0.1+10-21",
+				PURL:      "pkg:generic/java@19.0.1%2B10-21",
```
willmurphyscode, line 662: @spiffcs the change looks good, but would you mind building a version of Grype that includes this change to Syft and testing matching behavior, and doing `syft -o json | grype` for some images with this change? I want to make sure changes like this don't interfere with Grype's ability to choose version parser and parse and compare versions. I know there's some fiddly string manipulation in there, and I don't know whether Grype knows to decode this stuff before trying to compare version strings.

The logic I'm worried about is here: https://github.com/anchore/grype/blob/556ab2b959a88a61fb70cf511637fbb3bf036e49/grype/version/version.go#L33

willmurphyscode: Thanks @spiffcs - test runs on https://github.com/anchore/grype/pull/1784 show that this still works.

willmurphyscode (approved):

PR #2748, opened 2024-04-04
Differentiate between JRE and JDK

LaurentGoderre:
This gives option to add vulnerability that apply to just the JDK. A JDK image would then report both the JRE and JDK.


spiffcs:
Thanks @LaurentGoderre this looks really good!

PR #2762, opened 2024-04-09
FIx: CPE validation correctly returns error

willmurphyscode:
Previously, this method incorrectly return an empty Attributes object and a nil error, leading to callers attempting to use the empty attributes object.

We need to be careful with this, because turning on validation that previously had no effect is always a little scary. Creating for discussion for now.


spiffcs:
@willmurphyscode - I merged this with upstream and found the NewAttributes case you mentioned in TODO and added a test. I think this is ready for review. If you 🟢 my follow up work I 🟢 the work that come before and we can merge as `paired`

spiffcs (approved):

PR #2772, opened 2024-04-12
Fix removing labels in 'Detect schema changes' job

GijsCalis:
The builds were failing on trying to remove a label that is not present in the PR. This PR adds a checks if the label is set on a PR before removing.

But also the error message is wrong (example failed [workflow](https://github.com/anchore/syft/actions/runs/8658807959/job/23743407645#step:3:16)):
```
[RUN] gh pr edit 2769 --remove-label json-schema
Unable to label PR with 'json-schema'
```
This PR also corrects the above error message and makes the error message for adding a label more clear.


spiffcs (approved):

spiffcs:
Thanks @GijsCalis! Good find and thanks for the update here.

PR #2775, opened 2024-04-12
Add support for PHP Pear

LaurentGoderre:
This adds support for the Pear package manager (quite similar in nature to PECL, but is it's predecessor) and orients all pecl/pear cataloging material to use `pear` purl types. The existing pecl cataloger is now deprecated but still reachable explicitly via the API and application configuration.


matt-phylum:
You may want to hold off on this PR pending the outcome of package-url/purl-spec#300. I think this way makes sense, but there are arguments to be made for doing it either way.

spiffcs:
Waiting on https://github.com/package-url/purl-spec/pull/300 to resolve

wagoodman:
I'm going to rebase this and try to get it in -- we should not need to wait on the purl spec for this to make it in.

wagoodman (approved):
Thanks for the upgrade of PHP pecl/pear processing. We couldn't take the PR as is since it would remove functionality (thus be a breaking change), instead I've left the existing cataloger for existing API users and users that may have referenced the specific cataloger name in their application configurations and added a new cataloger to cover the combined pecl/pear processing. Only the new cataloger is wired up for default configurations. The last change made was to ensure that the channel parameter is not required from the underlying metadata (as it would not be present for v5 metadata files).

PR #2777, opened 2024-04-15
feat: index known cpes for PHP extensions

westonsteimel:
Indexes known CPEs from `pecl.php.net` and `pear.php.net`

This will depend on https://github.com/anchore/syft/pull/2775 which in turn depends on the package-url spec discussion in https://github.com/package-url/purl-spec/pull/300 before we know how to actually make use of it


westonsteimel:
Actually, this can go in now and we can always update it later if https://github.com/anchore/syft/pull/2775 does actually end up changing the syft package type.

spiffcs (approved):

Issue #2776, opened 2024-04-15
Pom parser not resolving all dependency versions


willyw0nka:
**What happened**:
When running syft on a Java project, some variables are returted as package versions without being evaluated.
```
 ✔ Indexed file system                                                                                                                                                                                          .
 ✔ Cataloged contents                                                                                                                            cdb4ee2aea69cc6a83331bbe96dc2caa9a299d21329efb0336fc02a82e1839a8
   ├── ✔ Packages                        [17 packages]  
   └── ✔ Executables                     [0 executables]  
[0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
NAME                                 VERSION                    TYPE           
caffeine                             ${caffeine.version}        java-archive    
easy-random-core-jdk11               6.1.2                      java-archive    
easy-random-randomizers-jdk11        6.1.2                      java-archive    
kotlin-test-junit5                   1.9.20                     java-archive    
mng-authentication-interceptor       1.0.2                      java-archive    
mng-spring-boot-starter-error-log    0.0.1                      java-archive    
mockk-jvm                            1.13.8                     java-archive    
spring-boot-starter-actuator         ${project.parent.version}  java-archive    
spring-boot-starter-cache            3.2.4                      java-archive    
spring-boot-starter-data-mongodb     ${project.parent.version}  java-archive    
spring-boot-starter-security         ${project.parent.version}  java-archive    
spring-boot-starter-test             ${project.parent.version}  java-archive    
spring-boot-starter-validation       ${project.parent.version}  java-archive    
spring-boot-starter-web              ${project.parent.version}  java-archive    
spring-cloud-starter-openfeign       4.1.0                      java-archive    
springdoc-openapi-starter-webmvc-ui  2.3.0                      java-archive    
springmockk                          3.1.1                      java-archive
```

**What you expected to happen**:
All dependency versions should be resolved

**Steps to reproduce the issue**:
```bash
git clone https://github.com/willyw0nka/spring-boot-empty-project.git
syft spring-boot-empty-project
```

**Anything else we need to know?**:
The unresolved versions needed to be evaulated twice, but syft only evaluated them once. For example, in [the project I tested](https://github.com/willyw0nka/spring-boot-empty-project/blob/main/pom.xml) the `org.springframework.boot` version is `${springboot.version}` which resolves to `${project.parent.version}` which then resolves to `3.2.4`. Syft only evaluated the version once so the result is `${project.parent.version}`.

[...]

PR #2781, opened 2024-04-15
Fix: repeatedly dereference pom variables

willmurphyscode:
Previously, if there was more than one layer of variable indirection in the pom property (propert A says it has the same value as property B, property B says it has the same value as property C), then Syft would only dereference one layer. Add a loop to dereference variables until either dereferencing fails, or until the variable is completely dereferenced back to a literal.

Fixes https://github.com/anchore/syft/issues/2776

Manual testing on repro example from that issue:

``` sh
❯ diff <(syft -q dir:spring-boot-empty-project | awk '{print $1 " "  $2 " " $3}') <(/tmp/syft -q dir:spring-boot-empty-project | awk '{print $1 " " $2 " " $3}')
9c9
< spring-boot-starter-actuator ${project.parent.version} java-archive
---
> spring-boot-starter-actuator 3.2.4 java-archive
11,15c11,15
< spring-boot-starter-data-mongodb ${project.parent.version} java-archive
< spring-boot-starter-security ${project.parent.version} java-archive
< spring-boot-starter-test ${project.parent.version} java-archive
< spring-boot-starter-validation ${project.parent.version} java-archive
< spring-boot-starter-web ${project.parent.version} java-archive
---
> spring-boot-starter-data-mongodb 3.2.4 java-archive
> spring-boot-starter-security 3.2.4 java-archive
> spring-boot-starter-test 3.2.4 java-archive
> spring-boot-starter-validation 3.2.4 java-archive
> spring-boot-starter-web 3.2.4 java-archive
```

### DIFF
--- a/syft/pkg/cataloger/java/parse_pom_xml.go
+++ b/syft/pkg/cataloger/java/parse_pom_xml.go
@@ -235,59 +235,66 @@ func cleanDescription(original *string) (cleaned string) {
 func resolveProperty(pom gopom.Project, property *string, propertyName string) string {
 	propertyCase := safeString(property)
 	log.WithFields("existingPropertyValue", propertyCase, "propertyName", propertyName).Trace("resolving property")
-	return propertyMatcher.ReplaceAllStringFunc(propertyCase, func(match string) string {
-		propertyName := strings.TrimSpace(match[2 : len(match)-1]) // remove leading ${ and trailing }
-		entries := pomProperties(pom)
-		if value, ok := entries[propertyName]; ok {
-			return value
-		}
+	for propertyMatcher.MatchString(propertyCase) {
+		oldPropertyCase := propertyCase
+		propertyCase = propertyMatcher.ReplaceAllStringFunc(propertyCase, func(match string) string {
+			propertyName := strings.TrimSpace(match[2 : len(match)-1]) // remove leading ${ and trailing }
+			entries := pomProperties(pom)
+			if value, ok := entries[propertyName]; ok {
+				return value
+			}
 
-		// if we don't find anything directly in the pom properties,
-		// see if we have a project.x expression and process this based
-		// on the xml tags in gopom
-		parts := strings.Split(propertyName, ".")
-		numParts := len(parts)
-		if numParts > 1 && strings.TrimSpace(parts[0]) == "project" {
-			pomValue := reflect.ValueOf(pom)
-			pomValueType := pomValue.Type()
-			for partNum := 1; partNum < numParts; partNum++ {
-				if pomValueType.Kind() != reflect.Struct {
-					break
-				}
-				part := parts[partNum]
-				for fieldNum := 0; fieldNum < pomValueType.NumField(); fieldNum++ {
-					f := pomValueType.Field(fieldNum)
-					tag := f.Tag.Get("xml")
-					tag = strings.Split(tag, ",")[0]
-					// a segment of the property name matches the xml tag for the field,
-					// so we need to recurse down the nested structs or return a match
-					// if we're done.
-					if part == tag {
-						pomValue = pomValue.Field(fieldNum)
-						pomValueType = pomValue.Type()
-						if pomValueType.Kind() == reflect.Ptr {
-							// we were recursing down the nested structs, but one of the steps
-							// we need to take is a nil pointer, so give up and return the original match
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -235,59 +235,66 @@ func cleanDescription(original *string) (cleaned string) {
 func resolveProperty(pom gopom.Project, property *string, propertyName string) string {
 	propertyCase := safeString(property)
 	log.WithFields("existingPropertyValue", propertyCase, "propertyName", propertyName).Trace("resolving property")
-	return propertyMatcher.ReplaceAllStringFunc(propertyCase, func(match string) string {
-		propertyName := strings.TrimSpace(match[2 : len(match)-1]) // remove leading ${ and trailing }
-		entries := pomProperties(pom)
-		if value, ok := entries[propertyName]; ok {
-			return value
-		}
+	for propertyMatcher.MatchString(propertyCase) {
```
willmurphyscode, line 238: Note that this diff looks larger than it is, because the change in control flow intends the replace function one more tab, but the body of the replace function is not changed.

syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -235,59 +235,66 @@ func cleanDescription(original *string) (cleaned string) {
...
+		oldPropertyCase := propertyCase
+		propertyCase = propertyMatcher.ReplaceAllStringFunc(propertyCase, func(match string) string {
+			propertyName := strings.TrimSpace(match[2 : len(match)-1]) // remove leading ${ and trailing }
+			entries := pomProperties(pom)
+			if value, ok := entries[propertyName]; ok {
+				return value
```
kzantow, line 244: Instead of having a loop, adding a recursive call here to `resolveProperty` could also be a solution?

kzantow (approved):
LGTM

syft/pkg/cataloger/java/parse_pom_xml.go
willmurphyscode: That diff would potentially be a _lot_ easier to read. Let me give that a shot. Thanks for the suggestion!

willmurphyscode: Actually, in adding new tests for this implementation, I realized that there are two cases we need to make sure halt:

1. Non-rooted deference, where following the path of repeated de-reference ends in a variable
2. Cyclic dereference, where a variable refers to itself

Neither of these are well-formed pom.xml, we just need to defend against them hanging or overflowing the stack or something.

willmurphyscode:
@kzantow I changed to a recursive implementation and added more test cases to guard against degenerate pom files that have cyclic or unending variable definitions. Looking for a re-review when you have a moment. Thanks!

syft/pkg/cataloger/java/parse_pom_xml_test.go
```diff
@@ -514,6 +514,92 @@ func Test_resolveProperty(t *testing.T) {
...
+						"other.version":      "${cyclic.version}",
+						"springboot.version": "${other.version}",
+						"cyclic.version":     "${springboot.version}",
+					},
+				},
+			},
+			expected: "${cyclic.version}",
+		},
+		{
+			name:     "resolution  halts even if cyclic involving parent",
```
kzantow, line 587: this does not appear to have any `${project.parent.version}` references to resolve the parent version

syft/pkg/cataloger/java/parse_pom_xml_test.go
```diff
@@ -514,6 +514,92 @@ func Test_resolveProperty(t *testing.T) {
...
+				},
+			},
+			expected: "${undefined.version}",
+		},
+		{
+			name:     "resolution halts even if cyclic",
+			property: "${springboot.version}",
+			pom: gopom.Project{
+				Parent: &gopom.Parent{
+					Version: stringPointer("${springboot.version}"),
```
kzantow, line 562: this is unused, I think

kzantow (approved):

syft/pkg/cataloger/java/parse_pom_xml_test.go
willmurphyscode: Removed the redundant `&gopom.Parent`. Thanks!

Issue #2704, opened 2024-03-12
[Feature] Syft for 32 bits ARM arch


dukelaar:
A release of syft for ARM32 :)

If not possible the process for manual compilation of syft with 32 bits ARM architecture.

tgerla:
Hey @dukelaar, thanks for the request. We haven't tested this, but you can try compiling a 32-bit version of Syft by checking the source out of GitHub and running:

```GOOS=linux GOARCH=arm go build -o syft-arm32 cmd/syft/main.go
```

You should end up with a 32-bit binary called syft-arm32. Please let us know how it goes--we are curious to see how well this works on a 32-bit system.

of-many-devices:
I've just tested this (cos I needed it too).
It required updating my golang-go package (1.19-1.21)
Then it built successfully with the command above.

Then managed to run the syft-arm32 executable on an arm32 filesystem but I had to --select-catalogers "-elf-package" because without this the syft process was OOM killed:
[  561.632097] Out of memory: Killed process 2472 (syft) total-vm:787712kB, anon-rss:559840kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:626kB oom_score_adj:0

SBOM generated looked alright to my untrained eye.

The only other thing I noticed was that the syft version was reported as "not provided":
syft --version
syft [not provided]

tgerla:
That's great, @of-many-devices, thank you for the feedback. I believe the "[not provided]" version is expected since you are building from source.

tgerla:
Hey all, since we've got a way to build a 32-bit version and it seems to work, I will go ahead and close this issue. If you find any problems running in 32 bits, please let us know. We suspect the OOM kill is legitimate, due to constrained system resources but please let us know if we're wrong. Thanks!

willmurphyscode:
@of-many-devices https://github.com/anchore/syft/releases/tag/v1.3.0 has some real memory efficiency improvements; you may be able to run with the elf-binary-package-cataloger enabled now.

PR #2811, opened 2024-04-24
Spike: lazy union reader

willmurphyscode:
## TL;DR

The elf-binary-package-cataloger sometimes gets OOM-killed, and generally is responsible for a lot of allocations. This spike adds a lazy union reader which avoids some unnecessary calls to `io.ReadAll`. The code is working, at least as far as the unit tests added and some manual testing an profiling, but probably wants some clean up and more careful testing.

Rough breakdown of profiling:
1. Running on `main` with all default image catalogers on this test image allocates ~69.53 GB over the course of the run
2. Running on this branch with the same setup allocates about 64.05 GB, for a savings of ~5.5 GB allocations
3. Running on `main` with the elf-binary-package-cataloger turned off allocates ~47.65 GB, or a savings of about 22 GB.

### Questions and concerns

1. Are we interested in this approach? Do the savings (reducing total allocations by elf cataloger by ~25%) justify the risks+costs of maintaining our own little reader package?
4. Are there other investigations that should be performed?
5. I'm a bit concerned with a homemade implementation in such a central place. Specifically, there are `int` and `int64` in the interfaces around io reads, and it would be easy to have an overflow or other bug. Any suggestions for missing test cases?

## Details

This is a spike to try to reduce some of Syft's memory allocations; in particular we have had user reports of OOM when running the `elf-binary-package-cataloger`. ([report](https://github.com/anchore/syft/issues/2704#issuecomment-2040387787), [report](https://anchorecommunity.slack.com/archives/C019BUXV7R6/p1713732067638659)).

Currently, Syft uses interfaces to hide the underlying file system from the catalogers, because the filesystem might be a tar, or a representation of an OCI image, or a directory, or whatever, and the catalogers shouldn't need to know about those differences. In order to achieve this, Syft uses a type called `UnionReader`: https://github.com/anchore/syft/blob/5326efcef13db1df2f03e77ce580feef9857569d/syft/internal/unionreader/union_reader.go#L14-L19

Currently, instantiating a UnionReader, if the underlying reader doesn't implement ReadAt or Seek, results in calling `io.ReadAll` on the underlying reader, and using the resulting bytes to back the Seek and ReadAt calls: https://github.com/anchore/syft/blob/5326efcef13db1df2f03e77ce580feef9857569d/syft/internal/unionreader/union_reader.go#L47

The new `elf-binary-package-cataloger` needs to call ReadAt and Seek on a potentially large number of files, resulting in a lot of `io.ReadAll` calls if this cataloger is enabled.

This branch spikes out making a new struct called `lazyUnionReader`, which wraps an `io.ReadCloser` and presents a `UnionReader`. Basically, instead of calling `io.ReadAll` when it is created, it copies all bytes read (or `Seek`ed past) into a backing byte slice. In other words, it has similar behavior as the current UnionReader, except that the allocation is lazy; if we never read past byte 102 in a given file, the backing array will only be (about) 102 bytes long, instead of the whole length of the file.

I thought maybe another package in the go ecosystem would provide a lazy read seeker, but I haven't found one, so this branch implements one with some unit tests.

I'll post comments with some profiling.

### DIFF
--- a/syft/internal/unionreader/lazy_union_reader.go
+++ b/syft/internal/unionreader/lazy_union_reader.go
@@ -1,0 +1,142 @@
+package unionreader
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"sync"
+)
+
+const readSize int64 = 1024 * 1024
+
+// lazyUnionReader must implement UnionReader
+var _ UnionReader = (*lazyUnionReader)(nil)
+
+// lazyUnionReader wraps an io.Reader to make it into a logical ReadSeeker
+// The reader maintains a []byte, which is everything that has been read so far.
+// Otherwise, callers needing a ReadSeeker might copy the entire reader into
+// a buffer in order to have a seeker.
+type lazyUnionReader struct {
+	buf    []byte        // the bytes that have been read so far
+	cursor int64         // the current position where Read() will take place
+	done   bool          // whether we have seen EOF from rc
+	rc     io.ReadCloser // the underlying reader
+	mu     sync.Mutex    // exported methods must acquire this lock before changing any field. Unexported methods assume their caller acquired the lock
+}
+
+func (c *lazyUnionReader) Read(p []byte) (n int, err error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	needBytes := int64(len(p))
+	newOffset := c.cursor + needBytes
+	err = c.ensureReadUntil(newOffset)
+	if err != nil && !errors.Is(err, io.EOF) {
+		return 0, err
+	}
+	// stop reading either at cursor + length p, or the end of the buffer, whichever is sooner
+	end := min(c.cursor+int64(len(p)), int64(len(c.buf)))
+	copy(p, c.buf[c.cursor:end])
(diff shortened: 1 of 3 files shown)


willmurphyscode:
Here are 3 profiles:

## Main, all default image catalogers:
image

## this branch, all default image catalogers:
image

## main, `elf-binary-package-cataloger` specifically disabled:
image

willmurphyscode:
The profiles were generated by adding the following code (+ necessary imports) to `cmd/syft/main.go`:

``` go
	defer func() {
		f, err := os.Create("mem.prof")
		defer f.Close()
		if err != nil {
			panic(err)
		}
		runtime.GC() // get up to data mem info
		if err = pprof.WriteHeapProfile(f); err != nil {
			panic(err)
		}
	}()
```

And then rendered by running `go tool pprof -http=:8080 mem.prof`, selecting "Flame graph" from the view menu, and selecting "alloc_space" from the sample menu.

willmurphyscode:
@kzantow asks a good question: How much do we save of maximum heap size, rather than total allocations over the whole execution?

I'll try to find the answer, and profile some more images.

willmurphyscode:
Superseded by #2814

PR #2814, opened 2024-04-25
fix: re-use embedded union reader if possible

willmurphyscode:
Supersedes #2811, at least for now.

Previously, because file.LocationReadCloser embeds a ReadCloser that might be a UnionReader, but doesn't implement the interface itself, the type assertion would fall and Syft would fall back to io.ReadAll to enable seeking on the underlying reader, resulting in a potentially large extra allocation.

Instead, check whether the passed ReadCloser is a
file.LocationReadCloser, and if so, try to use the embedded ReadCloser as a UnionReader.

### DIFF
--- a/syft/internal/unionreader/union_reader.go
+++ b/syft/internal/unionreader/union_reader.go
@@ -3,6 +3,7 @@ package unionreader
 import (
 	"bytes"
 	"fmt"
+	"github.com/anchore/syft/syft/file"
 	"io"
 
 	macho "github.com/anchore/go-macholibre"
@@ -42,6 +43,17 @@ func GetUnionReader(readerCloser io.ReadCloser) (UnionReader, error) {
 	reader, ok := readerCloser.(UnionReader)
 	if ok {
 		return reader, nil
+	}
+
+	// file.LocationReadCloser embeds a ReadCloser, which is likely
+	// to implement UnionReader. Check whether the embedded read closer
+	// implements UnionReader, and just return that if so.
+	r, ok := readerCloser.(file.LocationReadCloser)
+	if ok {
+		ur, ok := r.ReadCloser.(UnionReader)
+		if ok {
+			return ur, nil
+		}
 	}
 
 	b, err := io.ReadAll(readerCloser)
--- a/syft/internal/unionreader/union_reader.go
+++ b/syft/internal/unionreader/union_reader.go
@@ -3,11 +3,11 @@ package unionreader
 import (
 	"bytes"
 	"fmt"
-	"github.com/anchore/syft/syft/file"
 	"io"
 
 	macho "github.com/anchore/go-macholibre"
 	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/file"
 )
 
 // UnionReader is a single interface with all reading functions needed by multi-arch binary catalogers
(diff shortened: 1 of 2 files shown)


willmurphyscode:
Here's profiling of this branch scanning a giant image (graph comparable to those attached to superseded PR #2811 

image

This fix results in only a few more lines and results in many more memory savings that the other PR.

It's worth noting that the other PR would help in the case that Syft is used as a library, and is being invoked on a source of io.Readers that isn't stereoscope.

syft/internal/unionreader/union_reader.go
```diff
@@ -44,6 +45,17 @@ func GetUnionReader(readerCloser io.ReadCloser) (UnionReader, error) {
 		return reader, nil
 	}
 
+	// file.LocationReadCloser embeds a ReadCloser, which is likely
+	// to implement UnionReader. Check whether the embedded read closer
+	// implements UnionReader, and just return that if so.
+	r, ok := readerCloser.(file.LocationReadCloser)
+	if ok {
+		ur, ok := r.ReadCloser.(UnionReader)
```
willmurphyscode, line 53: I'm not 100% certain whether this also needs to check for `nil`. Would be happy for some reviewer input here.

syft/internal/unionreader/union_reader_test.go
```diff
@@ -28,3 +30,32 @@ func Test_getUnionReader_notUnionReader(t *testing.T) {
...
+	panic("don't call this in your unit test!")
+}
+
+func (p2 *panickingUnionReader) Close() error {
+	panic("don't call this in your unit test!")
+}
+
+func Test_getUnionReader_fileLocationReadCloser(t *testing.T) {
+	// panickingUnionReader is a UnionReader
+	var _ UnionReader = (*panickingUnionReader)(nil)
```
kzantow, line 54: nit: this could be done outside the test.

syft/internal/unionreader/union_reader_test.go
```diff
@@ -28,3 +30,32 @@ func Test_getUnionReader_notUnionReader(t *testing.T) {
...
+	embedsUnionReader := file.NewLocationReadCloser(file.Location{}, &panickingUnionReader{})
+
+	// embedded union reader is returned without "ReadAll" invocation
+	ur, err := GetUnionReader(embedsUnionReader)
+	require.NoError(t, err)
+	require.NotNil(t, ur)
```
kzantow, line 60: I think this should hold a reference to the `panickingUnionReader`, e.g. `expectedUnionReader := &panickingUnionReader{}` which is passed to the `NewLocationReadCloser` call and test that's the thing it gets back from the `GetUnionReader` call, e.g. `require.Equal(t, expectedUnionReader, ur)`, WDYT?

kzantow (commented):
A couple notes about the test, but overall this looks 👍

syft/internal/unionreader/union_reader.go
kzantow: Since `file.LocationReadCloser` is a _struct_ and not a pointer (not `*file.LocationReadCloser`), it won't ever be `nil`, the check you have is right 👍 .

willmurphyscode: But the io.ReadCloser on it could be nil, right?

kzantow: Ah yes, there's definitely a possibility it could be nil. You could probably _just_ check for that, e.g. `ur, _ := r.ReadCloser...`

willmurphyscode: This behavior isn't different now - `GetUnionReader` has always been able to accept and return `nil`, so let's just leave this up to the caller.

wagoodman (approved):

kzantow (approved):

PR #2815, opened 2024-04-25
fix: improve groupid extraction for Jenkins plugins

westonsteimel:
Consider the `Group-Id` java manifest property as this is typically set for Jenkins plugins if there is no pom file


spiffcs:
@westonsteimel was there anything else you wanted to do for this PR? If you want to sync tomorrow morning or next time your free we could write down what's left and I could try and help get this across the finish line?

I didn't see anywhere if there was another fix besides the `io.jenkins.plugins` exception you wanted to include here.

spiffcs (dismissed):

westonsteimel:
@spiffcs , sorry I missed your comment and I had completely forgotten about this one.  I think all that was missing is adding some jenkins plugins into the integration test image

westonsteimel:
Oh, we also need some unit tests for the package type logic (changing the type from being `java-archive` to being `jenkins-plugin`)

spiffcs (approved):

PR #2816, opened 2024-04-25
feat: index known CPEs for go modules

westonsteimel:


syft/pkg/cataloger/internal/cpegenerate/dictionary/index-generator/generate.go
```diff
@@ -312,3 +316,16 @@ func addEntryForPHPComposerPackage(indexed *dictionary.Indexed, ref string, cpeI
 
 	updateIndex(indexed, dictionary.EcosystemPHPComposer, ref, cpeItemName)
 }
+
+func addEntryForGoModulePackage(indexed *dictionary.Indexed, ref string, cpeItemName string) {
+	// Prune off the non-package-name parts of the URL
+	ref = strings.Split(ref, "?")[0]
```
spiffcs, line 322: Is ref here always a valid URL?

If it is and we're trying to split off the Query parameters is it safer to do something like the following?
```
u, err := url.Parse(ref)
if err != nil {
    // handle error and resort to split method
}
ref = u.Scheme + "://" + u.Host + u.Path
```

spiffcs (commented):
👍 looks awesome just one ? about how the ref is parsed

spiffcs (approved):

PR #2822, opened 2024-04-29
Fill in SPDX originator for all supported package types

### LINKED ISSUE #2632: SPDX originator is not always populated
When determining originator we only consider select ecosystems:

https://github.com/anchore/syft/blob/25ae7bf55f36fc8bf786b3eb79639b8af9539fee/syft/format/internal/spdxutil/helpers/origintor.go#L17-L40

Ideally we should expand this to fill in an answer in as many ecosystems as possible. There have been suggestions to at least add Java under consideration, looking specifically at the `Specification-Vendor` followed by `Implementation-Vendor` for non-empty values to use as an originator. 

We should probably add a completion-test for this, which exhaustively covers all metadata types (use [this](https://github.com/anchore/syft/blob/25ae7bf55f36fc8bf786b3eb79639b8af9539fee/syft/testutil/package_metadata.go#L13)). This way we can ensure that as new metadata types are added we can fail if there isn't an explicit test to cover originator functionality.

wagoodman:
Today the [SPDX originator field](https://spdx.github.io/spdx-spec/v2.2.2/package-information/#76-package-originator-field) is partially filled in based on the package type. This enhances this processing in a few ways:
- adds several more types considered during processing
- adds completion testing, such that more package metadata types will result in requiring to add such originator processing here as well
- combines testing for originator and supplier fields, which are semantically linked thus should both be considered during test

Fixes #2632


spiffcs (approved):
🟢

PR #2823, opened 2024-04-29
fix: better clean up of file handles

### LINKED ISSUE #2819: New version 1.3.0 leads to "too many open files" while scanning bigger images
**What happened**:
Images that used to scan successfully under < 1.3.0 are no longer scannable by Syft 1.3.0

failed to run task: unable to catalog file digests: failed to process file "****": digests-cataloger unable to observe contents of *****: open /tmp/stereoscope-****/oci-registry-image-*****/sha256:*****.tar: **too many open files**

Scan finishes without crash

Scan any (bigger) image related to ML/AI (eg https://hub.docker.com/r/pytorch/pytorch/tags). 

**Environment**:
- Output of `syft version`: 1.3.0
- OS (e.g: `cat /etc/os-release` or similar): Linux

willmurphyscode:
Previously, the ELF binary package cataloger would leak file handles, and the generic cataloger loop would leak file handles if parsing the file panicked. Instead, ensure both situations `defer close` new file readers as soon as they are made.

The incorrect behavior had been happening for some time, but was probably exposed by #2814 which caused the ELF binary package cataloger to rely more on file handles and less on in-memory buffers.

Fixes #2819 - in my testing, syft 1.3.0 needs about 1500 concurrent file handles to parse `pytorch/pytorch:latest`; after this change, the number is closer to 800.

### DIFF
--- a/syft/pkg/cataloger/generic/cataloger.go
+++ b/syft/pkg/cataloger/generic/cataloger.go
@@ -3,6 +3,7 @@ package generic
 import (
 	"context"
 
+	"github.com/anchore/go-logger"
 	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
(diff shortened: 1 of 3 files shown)


spiffcs (approved):

syft/pkg/cataloger/generic/cataloger.go
```diff
@@ -118,20 +118,21 @@ func (c *Cataloger) Catalog(ctx context.Context, resolver file.Resolver) ([]pkg.
 	}
 
 	for _, req := range c.selectFiles(resolver) {
-		location, parser := req.Location, req.Parser
+		log.WithFields("path", req.Location.RealPath).Trace("parsing file contents")
 
-		log.WithFields("path", location.RealPath).Trace("parsing file contents")
+		invokeParser := func(location file.Location, parser Parser) ([]pkg.Package, []artifact.Relationship, error) {
```
wagoodman, line 123: could this be a static function over a closure? it would be simpler to reason about.

willmurphyscode: Agreed. It will have a lot more parameters (env and logger), but it will be easier to reason about, so that's probably the right trade off. I'll try to push up a rev.

willmurphyscode:
@wagoodman / @spiffcs  - refactored and added a unit test for not leaking file on generic cataloger panic.

syft/pkg/cataloger/generic/cataloger.go
```diff
@@ -122,17 +123,9 @@ func (c *Cataloger) Catalog(ctx context.Context, resolver file.Resolver) ([]pkg.
 
 		log.WithFields("path", location.RealPath).Trace("parsing file contents")
```
wagoodman, line 124: optional opportunity for unrelated cleanup -- this should be using `logger` not `log`

wagoodman (approved):
awesome test 🥇

Issue #2826, opened 2024-05-01
Improve linting for `defer Close` type issues


willmurphyscode:
This issue is a follow up to prevent #2819 from happening again.

Syft has a method that should be called to clean up after file IO: https://github.com/anchore/syft/blob/0774ad15e26bc9230dbfc393b3e7d6bd36b93cf7/internal/err_helper.go#L13

Ideally, we'd like to enforce via linting:

- [ ] Files are closed with with `internal.CloseAndLogError`
- [ ] The call to `internal.CloseAndLogError` must be called with `defer`
- [x] Defer must not be called in a loop - #2825 

That way, every time someone asks a file resolver for a ReadCloser, it would be guaranteed not to leak the handle.

willmurphyscode:
Here's a quick implementation that could be wired up to golangci-lint. It hard-codes the name of the function that must defer for now, but that could be passed via config.

``` go
package analyzer

import (
	"go/ast"
	"golang.org/x/tools/go/analysis"
	"golang.org/x/tools/go/analysis/passes/inspect"
	"golang.org/x/tools/go/ast/inspector"
)

func run(pass *analysis.Pass) (any, error) {
	insp := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
	nodeFilter := []ast.Node{
		(*ast.ExprStmt)(nil),
		(*ast.DeferStmt)(nil),
	}
	insp.Preorder(nodeFilter, func(node ast.Node) {
		// if we have a *ast.ExprStmt that calls internal.CloseAndLogError, report a problem.
		// (if the function is correctly called in a defer statement, the block will have
		switch t := node.(type) {
		case *ast.ExprStmt:
			if !isExprStmtAllowed(t, pass) {
				pass.Reportf(t.Pos(), "internal.CloseAndLogError must be called in defer")
			}
		}
	})
	return nil, nil
}

func isExprStmtAllowed(e *ast.ExprStmt, pass *analysis.Pass) bool {
	call, ok := e.X.(*ast.CallExpr)
	if !ok {
		return true
	}
	sel, ok := call.Fun.(*ast.SelectorExpr)
	if !ok {
		return true
	}

	obj := pass.TypesInfo.Uses[sel.Sel]
	if obj == nil {
		return true
	}
	pkg := obj.Pkg()
	if pkg == nil {
		return true
	}

	if pkg.Path() == "github.com/anchore/syft/internal" && sel.Sel.Name == "CloseAndLogError" {
		return false
	}

[...]

PR #2836, opened 2024-05-02
chore: add custom lint to check close called outside defer

willmurphyscode:
See #2826

### DIFF
--- a/test/linters/ensuredefer/analyzer.go
+++ b/test/linters/ensuredefer/analyzer.go
@@ -1,7 +1,9 @@
 package ensuredefer
 
 import (
+	"fmt"
 	"go/ast"
+	"strings"
 
 	"github.com/golangci/plugin-module-register/register"
 	"golang.org/x/tools/go/analysis"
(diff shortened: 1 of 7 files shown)


test/linters/ensuredefer/analyzer.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	analyzer.Run = makeRun(callsToCheck)
+
+	return []*analysis.Analyzer{
+		analyzer,
+	}, nil
+}
+
+func (p *analyzerPlugin) GetLoadMode() string {
+	return register.LoadModeSyntax
```
willmurphyscode, line 129: I have yet to learn what goes here; all the values I've tried work and result in approximately the same time taken to execute.

willmurphyscode:
It looks like this approach would cause the go.mod to have reference to a GPL-3.0-licensed package in it: https://github.com/golangci/plugin-module-register/tree/v0.1.1, because a call is needed into the `register` method from that package. To be on the safe side, licensing wise, I'm closing this PR, and will put up a different repo to accept the GPL'ed code.

PR #2837, opened 2024-05-02
chore: use ruleguard to test for missing defer statements

willmurphyscode:
See #2826 

Supersedes https://github.com/anchore/syft/pull/2836.

The only hang-up I've seen is that calls to `make lint` can run an old version of the rules unless `./.tool/golangci-lint cache clean` is called first, which could be annoying when PRing future rule changes.

### DIFF
--- a/syft/pkg/cataloger/debian/package.go
+++ b/syft/pkg/cataloger/debian/package.go
@@ -265,8 +265,9 @@ func fetchCopyrightContents(resolver file.Resolver, dbLocation file.Location, m 
 
 	reader, err := resolver.FileContentsByLocation(*location)
 	if err != nil {
-		log.Warnf("failed to fetch deb copyright contents (package=%s): %w", m.Package, err)
+		log.Warnf("failed to fetch deb copyright contents (package=%s): %s", m.Package, err)
 	}
+	defer internal.CloseAndLogError(reader, location.RealPath)
 
 	l := location.WithAnnotation(pkg.EvidenceAnnotationKey, pkg.SupportingEvidenceAnnotation)
 
--- a/test/rules/rules.go
+++ b/test/rules/rules.go
@@ -6,10 +6,10 @@ import "github.com/quasilyte/go-ruleguard/dsl"
 
 // nolint:unused
 func resourceCleanup(m dsl.Matcher) {
-	m.Match(`$res, $err := $resolver.FileContentsByLocation($*_); if $*_ { $*_ }; $next`).
+	m.Match(`$res, $err := $resolver.FileContentsByLocation($loc); if $*_ { $*_ }; $next`).
 		Where(m["res"].Type.Implements(`io.Closer`) &&
 			m["err"].Type.Implements(`error`) &&
 			m["res"].Type.Implements(`io.Closer`) &&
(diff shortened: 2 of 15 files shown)


willmurphyscode:
We see that this branch has the expected failures: https://github.com/anchore/syft/actions/runs/8930279182/job/24530010619?pr=2837#step:4:50

``` log
task: [lint] .tool/golangci-lint run --tests=false
Error: syft/linux/identify_release.go:67:4: ruleguard: internal.CloseAndLogError should be deferred right after the error returned from resolver.FileContentsByLocation is checked. (gocritic)
			contentReader, err := resolver.FileContentsByLocation(location)
			^
... snip many more ...
task: Failed to run task "static-analysis": exit status 1
```

Taking a look at `syft/linux/identify_release.go:67:4`, we see: https://github.com/anchore/syft/blob/9eb8c6288de56e94a7961138c86b6dd833279f46/syft/linux/identify_release.go#L66-L90 clearly has the issue

test/rules/rules.go
```diff
@@ -0,0 +1,15 @@
...
+import "github.com/quasilyte/go-ruleguard/dsl"
+
+// nolint:unused
+func resourceCleanup(m dsl.Matcher) {
+	m.Match(`$res, $err := $resolver.FileContentsByLocation($*_); if $*_ { $*_ }; $next`).
+		Where(m["res"].Type.Implements(`io.Closer`) &&
+			m["err"].Type.Implements(`error`) &&
+			m["res"].Type.Implements(`io.Closer`) &&
+			!m["next"].Text.Matches(`defer internal.CloseAndLogError`)).
+		Report(`internal.CloseAndLogError should be deferred right after the error returned from $resolver.FileContentsByLocation is checked.`)
```
kzantow, line 14: I think this approach is pretty simple to comprehend and the rule defined here makes sense for the specific case that we want to check. I think maybe the report could read something more like:
```
defer internal.CloseAndLogError($res,...) should be called immediately following the error check
```

willmurphyscode: The failing check now looks like this:

```
Error: syft/file/cataloger/executable/cataloger.go:63:3: ruleguard: please call "defer internal.CloseAndLogError(reader, loc.RealPath)" right after checking the error returned from resolver.FileContentsByLocation. (gocritic)
```

wagoodman (approved):
slick!

syft/pkg/cataloger/debian/package.go
```diff
@@ -265,8 +265,9 @@ func fetchCopyrightContents(resolver file.Resolver, dbLocation file.Location, m
 
 	reader, err := resolver.FileContentsByLocation(*location)
 	if err != nil {
-		log.Warnf("failed to fetch deb copyright contents (package=%s): %w", m.Package, err)
+		log.Warnf("failed to fetch deb copyright contents (package=%s): %s", m.Package, err)
```
spiffcs, line 268: Are we changing the error from `%w` to `%s` here?

spiffcs (approved):
Approved - Small comment on moving a `wrap` directive to `string` for an error value =)

kzantow (approved):

PR #2845, opened 2024-05-06
Add missing CPE for traefik, memcached, and postgres binaries

LaurentGoderre:


kzantow (approved):
LGTM -- I verified NVD is using all of these in at least some of the reports for these products 👍

Issue #2798, opened 2024-04-19
Golang: Search remote licenses not working in a CI pipeline when scanning Docker image


Joerki:
**What happened**:
In our CI pipeline we run syft to scan a Docker image. No Go environment is present. search-remote-licenses (in env or .syft.yaml) is configured as "true". 

But Syft does not write license information into Syft-JSON file.

It is working on my local machine where I have a Go installation.

It looks that search remote licenses function has a path dependency that it should not have when working in an environment without Go.
 
**What you expected to happen**:
If search-remote-licenses is set by config file or environment file it should work always if set.

**Steps to reproduce the issue**:
On a machine with Go installation ($HOME/go exists):

```
(base) joerg@delly:~/project$ cat ~/.syft.yaml 

quiet: false

check-for-app-update: false

parallelism: 3

# scope: "all-layers"

default-catalogers: [ "go-module-file-cataloger", "go-module-binary-cataloger"]

format:
  json:
    pretty: true

golang:
  search-local-mod-cache-licenses: false
  search-remote-licenses: true
```

Invocation:
syft scan mongo:5.0.26-focal -o syft-json=/home/joerg/projects/mongo_inspect/mongo-syft-single.json -vv

Result 1:

Go Licenses are present in Syft-JSON

$HOME/go renamed to $HOME/go_
Same invocation.

Result 2:

Go licenses are not present in Syft-JSON

**Anything else we need to know?**:
N/A

**Environment**:
- Output of `syft version`:
Application: syft
Version:    1.2.0
BuildDate:  2024-04-12T18:31:58Z
GitCommit:  dde5d349b1eef740c285255e6a9e3a8f5c9938e1
GitDescription: v1.2.0
Platform:   linux/amd64
GoVersion:  go1.21.9
Compiler:   gc

- OS (e.g: `cat /etc/os-release` or similar):
uname -a
Linux delly 6.1.0-18-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.76-1 (2024-02-01) x86_64 GNU/Linux

(Native Linux, no WSL)

kzantow:
Hi @Joerki -- I'm a bit confused what the problem is. I see `search-remote-licenses` is set to false, which is the default and I don't see this being set by an environment variable, so it's expected to be false. The default `local-mod-cache-dir` is based on your home directory, which is also expected. Could you elaborate on your configuration and the problem a bit to help me understand?

Joerki:
Hi @kzantow,
[...]

Issue #1933, opened 2023-07-12
Add licenses for Go packages


sorokinmax:
**What would you like to be added**:
The filled in "licenses" field in the SBOM report for Go packages.

**Why is this needed**:
Now it looks like a disadvantage compared to other dependencies. 
Tracking license purity is an important part of software development.

**Additional context**:
License information is easy to get at https://pkg.go.dev
For example:
![image](https://github.com/anchore/syft/assets/19753153/5741f269-6476-4009-90cb-a251e71b2e09)

kzantow:
@sorokinmax license information is not present in the `go.mod/go.sum` files or in the binary go mod data, so it needs to get retrieved from other sources. Syft already has the ability to do this for Go packages, but it is not enabled by default. There are a few places Syft can attempt to find licenses: the local go mod cache, the configured go proxy, and finally a git repo matching the Go mod. As noted in the config file documentation:
```
golang:
   # search for go package licences in the GOPATH of the system running Syft, note that this is outside the
   # container filesystem and potentially outside the root of a local directory scan
   # SYFT_GOLANG_SEARCH_LOCAL_MOD_CACHE_LICENSES env var
   search-local-mod-cache-licenses: false
   
   # specify an explicit go mod cache directory, if unset this defaults to $GOPATH/pkg/mod or $HOME/go/pkg/mod
   # SYFT_GOLANG_LOCAL_MOD_CACHE_DIR env var
   local-mod-cache-dir: ""

   # search for go package licences by retrieving the package from a network proxy
   # SYFT_GOLANG_SEARCH_REMOTE_LICENSES env var
   search-remote-licenses: false

   # remote proxy to use when retrieving go packages from the network,
   # if unset this defaults to $GOPROXY followed by https://proxy.golang.org
   # SYFT_GOLANG_PROXY env var
   proxy: ""

   # specifies packages which should not be fetched by proxy
   # if unset this defaults to $GONOPROXY
   # SYFT_GOLANG_NOPROXY env var
   no-proxy: ""
```

e.g. to enable searching both the local go mod cache and remotes, you can enable this by executing Syft like: 
```
SYFT_GOLANG_SEARCH_LOCAL_MOD_CACHE_LICENSES=true SYFT_GOLANG_SEARCH_REMOTE_LICENSES=true syft <source>
```

Does this work for you?

sorokinmax:
Thank you, @kzantow! I really overlooked these parameters. But it didn't work for me. 
[...]

PR #2852, opened 2024-05-07
fix: separate golang license caches from mod dir

kzantow:
Previously, license handling in Golang had a few quirks and may not function if the go mod directory did not exist.

* Separate local go mod cache directory from remote download cache
* Use known cache directory using XDG if not specified in config; create the directory if it does not exist
* No longer download files into a go mod directory, only the syft cache

TODO:
- [x] ensure this cache directory is excluded during directory scans (e.g. `syft /`)
- [x] ascertain if we need a `syft cache` command to view/manage the cache as part of this PR (answer: no, we are caching significantly less data than syft was previously)
- [x] add more tests
- [x] remove writable fileresolver usage and just directly use fs.FS implementations

Fixes: #2798 #1933


syft/pkg/cataloger/golang/licenses.go
```diff
@@ -23,23 +23,36 @@ import (
...
 
 func newGoLicenses(catalogerName string, opts CatalogerConfig) goLicenses {
+	var localModCacheResolver file.Resolver
+	if opts.SearchLocalModCacheLicenses {
+		localModCacheResolver = fileresolver.NewFromUnindexedDirectory(opts.LocalModCacheDir)
+	}
+
+	var remoteModCacheResolver file.WritableResolver
+	if opts.SearchRemoteLicenses {
+		remoteModCacheResolver = cache.Resolver("go-mod")
```
wagoodman, line 47: I think we'd want a way to naturally version these. Specifically, being able to know when some existing cache is compatible with the current code. I feel like all callers should specify this explicitly:
```
cache.Resolver("go-mod", "v1")
```
over
```
cache.Resolver("go-mod-v1")
```
since it is easy for callers to not know when to coordinate breaking changes.

wagoodman (commented):
This is great! I feel like this is something we can build on over time too. 

Since this PR doesn't introduce mechanisms for cache invalidation, users will start to have new cache directories that will continue to grow (this is different than grype, where new DBs get replaced). 

I feel like before merging this there should be an answer to how a user would delete the cache or how syft would automatically do this... this could be:
- like you brought up on the call, allow for TTL configuration and track in the cache dir track last write date so we can clean up periodically.
- maybe add a `cache list` and `cache delete` subcommands to syft

syft/internal/cache/error_resolver.go
```diff
@@ -0,0 +1,40 @@
+package cache
+
+import "fmt"
+
+// GetResolverCachingErrors returns a Resolver that caches errors and will return them
+// instead of continuing to call the provided resolve functions
+func GetResolverCachingErrors[T any](name string) Resolver[T] {
```
kzantow, line 7: It might be better to say: this should be handled by the resolver function, but then I foresee all the implementations duplicating something more-or-less like this, so this is just here for a convenience

cmd/syft/internal/options/cache.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/anchore/syft/syft"
+)
+
+// Cache provides configuration for the Syft caching behavior
+type Cache struct {
+	Dir string `yaml:"dir" json:"dir" mapstructure:"dir"`
+}
+
+func (c *Cache) DescribeFields(descriptions clio.FieldDescriptionSet) {
+	descriptions.Add(&c.Dir, "root directory to cache any downloaded content; defaults to use $XDG_CACHE_HOME or $HOME/.cache")
```
wagoodman, line 14: is `$HOME/.cache` isn't always true depending on the platform

syft/internal/cache/hash_type.go
```diff
@@ -0,0 +1,61 @@
...
+		v := emptyValue(t.Elem())
+		s := reflect.MakeSlice(t, 1, 1)
+		s.Index(0).Set(v)
+		return s
+	case reflect.Struct:
+		v := reflect.New(t).Elem()
+		// get all empty field values, too
+		for i := 0; i < v.NumField(); i++ {
+			f := t.Field(i)
+			if !f.IsExported() || strings.Contains(f.Tag.Get("hash"), "ignore") {
```
wagoodman, line 49: I think both `ignore` and `-` values are considered the same

syft/internal/cache/hash_type.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		ptrv := reflect.New(e)
+		ptrv.Elem().Set(v)
+		return ptrv
+	case reflect.Slice:
+		v := emptyValue(t.Elem())
+		s := reflect.MakeSlice(t, 1, 1)
+		s.Index(0).Set(v)
+		return s
+	case reflect.Struct:
```
wagoodman, line 44: why do we need to handle struct at all? It seems based off of the tests we're trying to make certain that `[]T` and `[]*T` (and similarly `T` and `*T`) resolve to the same thing. That hints that we need to be able reason about pointers and slices, but I don't see the connection to structs.

syft/internal/cache/dir.go
```diff
@@ -0,0 +1,71 @@
...
+//   - HOME/.cache
+//   - TEMP
+func findRoot() string {
+	if rootDir != "" {
+		dir, err := homedir.Expand(rootDir)
+		if err != nil {
+			log.Warnf("unable to expand cache directory %s: %v", rootDir, err)
+		} else {
+			return dir
+		}
```
wagoodman, line 36: I feel like this should be in `SetRootDir` and done once. `homedir.Expand()` can be expensive, so invoking it as few times as possible would be ideal.

syft/internal/cache/dir.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/adrg/xdg"
+	"github.com/mitchellh/go-homedir"
+
+	"github.com/anchore/syft/internal/log"
+)
+
+// SetRootDir sets the root directory to use for downloaded content caching for things like module license information
+func SetRootDir(dir string) {
+	rootDir = dir
```
wagoodman, line 18: should we always nest `syft` to whatever is passed?

syft/internal/cache/dir.go
```diff
@@ -0,0 +1,71 @@
...
+		cacheRoot, err = homedir.Dir()
+		if err != nil {
+			cacheRoot = os.TempDir()
+			log.Debugf("unable to get home directory due to: %v, defaulting cache to temp dir: %s", err, cacheRoot)
+		} else {
+			cacheRoot = filepath.Join(cacheRoot, ".cache")
+		}
+	}
+
+	return filepath.Join(cacheRoot, "syft")
```
wagoodman, line 51: I feel like this could be done in `init()` and set as the initial `rootDir` value, thus we are really always using the static `rootDir` value (either set by the user or determined at `init`).

syft/internal/cache/dir.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// getDir returns a writable directory with the given name under the root cache directory returned from findRoot,
+// the directory will be created if it does not exist
+func getDir(subDirs ...string) (string, error) {
+	dir := filepath.Join(findRoot(), filepath.Join(subDirs...))
+	stat, err := os.Stat(dir)
+	if errors.Is(err, os.ErrNotExist) {
+		err = os.MkdirAll(dir, os.ModeDir|0700)
+		stat, _ = os.Stat(dir)
```
wagoodman, line 61: I think this error should also be checked, since the value of `stat` is depended on later (and shouldn't be if this errored out)

syft/internal/fileresolver/unindexed_directory.go
```diff
@@ -33,15 +40,15 @@ type UnindexedDirectory struct {
 	fs   afero.Fs
 }
 
-func NewFromUnindexedDirectory(dir string) file.WritableResolver {
+func NewFromUnindexedDirectory(dir string) DeletingWritableResolver {
```
wagoodman, line 43: I think all of these constructors should return the concrete type `UnindexedDirectory` and let the caller determine if they want writable/deletable behavior by their variable type.

syft/internal/fileresolver/temp.go
```diff
@@ -0,0 +1,30 @@
...
+	stereoscopeFile "github.com/anchore/stereoscope/pkg/file"
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/file"
+)
+
+func UseTempResolver[T any](fn func(r file.WritableResolver) (T, error)) (T, error) {
+	var t T
+	tmpDirGen := stereoscopeFile.NewTempDirGenerator("temp-resolver")
+	if tmpDirGen == nil {
+		// TODO in-memory resolver
```
wagoodman, line 15: based off of the TODO, would it be something like:
```
NewFromUnindexedDirectoryFS(afero.NewMemMapFs(), ...)
```

syft/internal/fileresolver/temp.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+
+	stereoscopeFile "github.com/anchore/stereoscope/pkg/file"
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/file"
+)
+
+func UseTempResolver[T any](fn func(r file.WritableResolver) (T, error)) (T, error) {
+	var t T
+	tmpDirGen := stereoscopeFile.NewTempDirGenerator("temp-resolver")
```
wagoodman, line 13: nit: the name should be attributable to `syft` as a prefix

syft/internal/fileresolver/temp.go
```diff
@@ -0,0 +1,30 @@
...
+		// TODO in-memory resolver
+		return t, fmt.Errorf("unable to get temp dir")
+	}
+	defer func() {
+		err := tmpDirGen.Cleanup()
+		if err != nil {
+			log.Debugf("unable to clean up due to: %v", err)
+		}
+	}()
+	tmpDir, err := tmpDirGen.NewDirectory("root")
```
wagoodman, line 24: same as https://github.com/anchore/syft/pull/2852/files#r1631596539

syft/internal/cache/resolver.go
```diff
@@ -0,0 +1,145 @@
...
+	Resolve(key string, resolver resolverFunc[T]) (T, error)
+}
+
+// GetResolver returns a cache resolver for persistent cached data across Syft runs
+func GetResolver[T any](name string) Resolver[T] {
+	versionKey := hashType[T]()
+	return cacheResolver[T]{
+		name:         fmt.Sprintf("%s/%s", name, versionKey),
+		ttl:          7 * 24 * time.Hour,
+		fileResolver: fileResolver(name, versionKey),
```
wagoodman, line 31: [no request for change in this PR] the reason for using a `file.Resolver` earlier was because we were in the context of the catalogers, which opted to not deal with filesystem calls directly.

We can leave this for now, but it might be in the future we use `afero.FS` directly here (or `fs.FS` if we decide we don't need symlink support).

syft/internal/cache/dir.go
wagoodman: actually, I might be wrong about this -- I think `homedir` caches it's result globally.

syft/internal/cache/resolver.go
wagoodman: Well, I take this back (or it's more grey than I thought) -- we might cache entire dir trees that should be used as part of cataloging, and expose that resolver out to the cataloger. Which means catalogers only needing to reason about one way to get to files is a good thing (via the resolvers).

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -311,3 +312,57 @@ func getModuleRepository(moduleName string, moduleVersion string) (fs.FS, error)
...
+
+func toPkgLicenses(goLicenses []goLicense) []pkg.License {
+	var out []pkg.License
+	for _, l := range goLicenses {
+		out = append(out, pkg.License{
+			Value:          l.Value,
+			SPDXExpression: l.SPDXExpression,
+			Type:           l.Type,
+			URLs:           l.URLs,
+			Locations:      toPkgLocations(l.Locations),
```
wagoodman, line 334: I'm not certain if this is accurate, since these locations might be leaked from the host. Instead, for locations we got remotely should specify the URL and not contain a location https://github.com/anchore/syft/blob/254a562b4e3d5ad6c3f471342ae0107d570b5eb9/syft/pkg/license.go#L30

syft/internal/cache/test_cache.go
```diff
@@ -0,0 +1,11 @@
+package cache
+
+import "testing"
+
+func NewTestCache(t *testing.T) {
```
wagoodman, line 5: This should probably be moved into its own package, something like `cachetest` (so something like `cachetest.New()` or `cachetest.NewCache()`) otherwise the syft binary will start depending on the stdlib `testing` package. But if separated, then we'll only import `cachetest` within `_test.go` files, thus there would not be a direct dependency on `testing`.

This is not a critical comment since this is not a public package. the only hesitation is that we have public packages that depend on this package which directly uses `testing`.

syft/pkg/cataloger/golang/licenses_test.go
```diff
@@ -151,16 +153,19 @@ func Test_RemoteProxyLicenseSearch(t *testing.T) {
 		},
 	}
 
-	modDir := path.Join(t.TempDir())
+	cache.NewTestCache(t)
+	cacheDir := filepath.Join(t.TempDir(), "golang-remote-licenses") // directory does not exist and should be created
+	cache.SetRootDir(cacheDir)
```
wagoodman, line 158:
```suggestion
	cache.NewTestCache(t)
```

I might be missing something though

kzantow: We are setting this because basically just as a validation of the issue this intends to fix: when the cache directory doesn't exist. The `cache.TestCache(t)` is just a convenience to reset it.

syft/internal/cache/test_cache.go
kzantow: I've just removed this, it was no longer needed for anything

syft/internal/cache/resolver.go
kzantow: I've changed this so the cache is basically a simple key/[]byte store. if a user wanted to store a directory tree, it would be possible to use the cache directly and store, say an uncompressed .tar file. it returns a ReaderAt, so it could be efficient to traverse. we could add some cache.DirTree() utility of some sort to help this if we really needed/wanted it, but I don't really think we want this.

syft/internal/cache/hash_type.go
kzantow: The problem is: to use the hashing function, we need an object in a slice, or it's `0`, regardless of the type, so we need to iterate through the fields to make sure slices have objects

syft/internal/cache/dir.go
kzantow: This is only done "once", it's part of the configuration now.

syft/internal/cache/dir.go
kzantow: I don't believe so, if I tell syft to cache in directory like `~/syft-cache` it would be unnecessary to put a subdirectory of `syft`. I think if people are specifying this, they're managing a directory for syft caches specifically, so the nesting is unnecessary. And, of course, if they _want_ a subdirectory, they can add it themselves and use ~/syft-cache/syft`, etc..

syft/internal/cache/dir.go
kzantow: This is no longer relevant with the latest changes: the configuration sets the dir once and passes a configured manager.

syft/internal/fileresolver/unindexed_directory.go
kzantow: I removed changes to this package because of using `afero`

syft/internal/fileresolver/temp.go
kzantow: Removed this function.

internal/cache/resolver.go
```diff
@@ -0,0 +1,81 @@
...
+
+	rdr, err := r.cache.Read(key)
+	if rdr == nil || err != nil {
+		return r.resolveAndCache(key, resolver)
+	}
+	defer internal.CloseAndLogError(rdr, key)
+
+	dec := json.NewDecoder(rdr)
+	if dec == nil {
+		log.Debugf("error getting cache json decoder for %s %v: %v", r.name, key, err)
```
wagoodman, line 54: nit: probably trace level?

internal/cache/resolver.go
```diff
@@ -0,0 +1,81 @@
...
+		return r.resolveAndCache(key, resolver)
+	}
+	var t T
+	err = dec.Decode(&t)
+	if err != nil {
+		log.Debugf("error decoding cached entry for %s %v: %v", r.name, key, err)
```
wagoodman, line 60: nit: probably trace level?

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -112,49 +115,39 @@ func (c *goLicenses) getLicensesFromLocal(moduleName, moduleVersion string) ([]p
...
+		err = fs.WalkDir(fsys, ".", func(filePath string, d fs.DirEntry, _ error) error {
+			if !c.lowerLicenseFileNames.Has(strings.ToLower(d.Name())) {
+				return nil
+			}
+			rdr, err := fsys.Open(filePath)
+			if err != nil {
+				log.Debugf("error opening license file %s: %v", filePath, err)
+				return nil
+			}
+			parsed, err := licenses.Parse(rdr, file.NewLocation(path.Join(moduleDir(moduleName, moduleVersion), filePath)))
```
wagoodman, line 137: here we can capture a URL that `remoteForProxy` could return, instead of a `file.Location`

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -63,47 +80,33 @@ func remotesForModule(proxies []string, noProxy []string, module string) []strin
...
+		return toPkgLicenses(goLicenses), err
 	}
 
-	// look in the local host mod cache...
-	licenses, err = c.getLicensesFromLocal(moduleName, moduleVersion)
-	if err != nil || len(licenses) > 0 {
-		return requireCollection(licenses), err
+	// look in the local host mod directory...
+	if c.opts.SearchLocalModCacheLicenses {
+		goLicenses, err = c.getLicensesFromLocal(moduleName, moduleVersion)
```
wagoodman, line 94: should we also capture URLs here since this is not for the filesystem analyzed? Maybe something like `file:///localhost/<path>` (not ideal, but open to suggestions)

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -21,25 +21,42 @@ import (
...
+	if opts.SearchLocalModCacheLicenses {
+		localModCacheResolver = fileresolver.NewFromUnindexedDirectory(opts.LocalModCacheDir)
+	}
+
+	return goLicenseResolver{
 		catalogerName:         catalogerName,
 		opts:                  opts,
-		localModCacheResolver: modCacheResolver(opts.LocalModCacheDir),
+		localModCacheResolver: localModCacheResolver,
+		licenseCache:          cache.GetResolverCachingErrors[[]goLicense]("golang"),
```
wagoodman, line 59:
```suggestion
		licenseCache:          cache.GetResolverCachingErrors[[]goLicense]("golang-v1"),
```
so we can isolate incompatible changes

syft/pkg/cataloger/golang/licenses.go
kzantow: As noted offline, I used `$GOPATH/pkg/mod/<path>` to denote this to have a reasonably accurate representation while avoiding leaking information about the user's system

wagoodman (approved):
This cache foundation is awesome ❤️ 🚀

Joerki:
Hi @wagoodman, @kzantow 

> This is great! I feel like this is something we can build on over time too.
> 
> Since this PR doesn't introduce mechanisms for cache invalidation, users will start to have new cache directories that will continue to grow (this is different than grype, where new DBs get replaced).
> 
> I feel like before merging this there should be an answer to how a user would delete the cache or how syft would automatically do this... this could be:
> 
> * like you brought up on the call, allow for TTL configuration and track in the cache dir track last write date so we can clean up periodically.
> * maybe add a `cache list` and `cache delete` subcommands to syft

Thanks for the info,
FYI: Good to know. In my case I don't need to do an explicit clean-up step, since  my CI pipeline agent throws away all local data after doing its jobs and becomes unused for some minutes.

For other users it might be good to have a choice 
- to discard the cache (at the end of execution "I really don't want to have it anymore")
- to keep it after execution (e.g. as default, "I want to use possible cached modules and collect further on")
- or to reset/remove it (at the beginning of execution, "I want to forget cached modules up to now, but want consider cached modules in the future until my next reset")
during certain build steps.
I might prefer a command line option or config setting that set one of possible option above.
With a single command (cache delete) I see a possible, unwanted effect when 
- e.g. a pipeline stops because of a failing step
- or a final cleanup step is skipped or gets forgotton

BR,
Jörg

kzantow:
Hey @Joerki -- we are going to improve the configuration in the near future. The current configuration options allows you to achieve the different options you mentioned. There is a cache directory (`SYFT_CACHE_DIR`) configuration option (you can always see this now using `syft config`). So you can give it an explicit dir that gets cleaned up or cached between CI runs.

There is also a cache TTL, specifying the time-to-live of cache items, but this does not delete them currently. We plan to add a `syft cache purge` or similar command to do this explicitly.

And there's an obvious `CACHE_ENABLED` we should add; I'll get that done right away :)

kzantow:
@Joerki I made another PR to make a couple configuration options work in a more expected manner and be more explicit to support the cases you've mentioned: https://github.com/anchore/syft/pull/2966

PR #2853, opened 2024-05-07
Add abstraction for adding relationships from package cataloger results

wagoodman:
I made an observation that while implementing more features for #572 that there are multiple catalogers that have almost the same logic to find relationships from a collection of homogenous packages. This PR refactors those spots to use the same abstraction.

Specifically, this PR adds a new `internal/dependency` package that contains:
  - a `Specifier` function type which extracts strings that are sensitive to package resource requirements and resource provides claims from a single package, normalized as strings. A `Specification` collects all `Provides` and `Requires` claims into a single data type.
  - A `Resolver` that uses a given `Specifier` and a collection of packages to create relationships between the given packages

This abstraction has been leveraged in both the `dpkg`, `alpm`, and `apk` catalogers.


syft/pkg/cataloger/internal/dependency/resolver.go
```diff
@@ -0,0 +1,81 @@
...
+				seen.Add(pairKey)
+			}
+		}
+	}
+	return relationships
+}
+
+func deduplicate(ss []string) []string {
+	set := strset.New(ss...)
+	// note: this must be a stable function
```
willmurphyscode, line 73: I don't understand this comment. If I have a deduplicated list of strings, then it doesn't matter whether it's sorted in a stable or unstable way, since there are no duplicate keys to break ties between. Or does "stable" in the comment not mean stable sorting? What am I missing here?

wagoodman: there is no guarantee that `strset.List` will yield the same ordering on subsequent invocations. In this case we're interested in multiple invocations of syft... we want them to produce the same SBOM, thus be resistant to unstable ordering.

syft/pkg/cataloger/debian/prosumer.go
```diff
@@ -0,0 +1,76 @@
...
+func stripVersionSpecifier(s string) string {
+	// examples:
+	// libgmp10 (>= 2:6.2.1+dfsg1)         -->  libgmp10
+	// libgmp10                            -->  libgmp10
+	// foo [i386]                          -->  foo
+	// default-mta | mail-transport-agent  -->  default-mta | mail-transport-agent
+	// kernel-headers-2.2.10 [!hurd-i386]  -->  kernel-headers-2.2.10
+
+	items := internal.SplitAny(s, "[(<>=")
+	if len(items) == 0 {
```
kzantow, line 60: Same comment as above that `len(items) == 0` is not possible

syft/pkg/cataloger/internal/dependency/resolver.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/scylladb/go-set/strset"
+
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/generic"
+)
+
+// Prosumer is a producer and consumer, in this context, for packages that provide resources and require resources.
+type Prosumer interface {
```
kzantow, line 14: After reflecting on this for a bit, I would like to suggest removing the `Prosumer` interface in favor of 2 callback functions: one for getting IDs produced by a package and another getting IDs depended on. I believe that would reduce the complexity and make things more composable. The biggest draw for using the `generic.Cataloger` is that a user only has to implement a simple function, there aren't interfaces to implement. I feel that adding something requiring implementing interfaces here that is likely to be used by quite a few `generic.Cataloger` users is a bit of a step backwards in terms of usability. And the user could simply implement a `pkg.Cataloger` and there would not really be as much a need for this because they would have a list of all the packages to be able to construct dependencies on fairly simply. By keeping with the function-only usage, I think it would be more consistent with the intended users of this functionality.

I would envision it looking something like:
```
type Producer func(pkg.Package) []string

type RelationshipResolver struct {
	provides Producer
	depends Producer
}

func NewRelationshipResolver(ids Producer, dependencies Producer) generic.Processor { ...
```

Usage, is still simple, e.g.:
```
WithProcessors(dependency.Processor(idProvider, dependencyProvider)) // where idProvider and dependencyProvider are local functions to provide ids (already implemented as part of a prosumer in this PR)
```
This would make certain types of dependency mappings only need to implement one function, and potentially look like:
```
WithProcessors(dependency.Processor(dependency.NameVersionProvider, dependencyProvider))
```

And would eliminate the duplicated boilerplate code such as:
```
var _ dependency.Prosumer = (*dpkgDBEntryProsumer)(nil)

type dpkgDBEntryProsumer struct{}

func newDBProsumer() dpkgDBEntryProsumer {
	return dpkgDBEntryProsumer{}
}
```

wagoodman: Actually, I started with something that looked exactly like this (different name though), but when working with it something about it bothered me:
```
WithProcessors(dependency.Processor(idProvider, dependencyProvider))
```
vs
```
WithProcessors(dependency.Processor(dependencyProvider, idProvider))
```
would be allowed from a type point of view, but the latter is a mistake and will result in incorrect behavior. Ideally this would be as clear as possible about what is being passed and also difficult to misuse.

I then went in this direction:
```
type Producer func(pkg.Package) []string
type Consumer func(pkg.Package) []string


type RelationshipResolver struct {
	provides Producer
	depends Consumer
}
```

But then realized that this doesn't help in most cases since we're passing function literals... so the same problem would arise!

After more thought it seemed natural to lean into either one or two interfaces so what is being passed is clear from a type perspective as early as compile time, with the added benefit that the function names semantically hint at the intended behavior.

syft/pkg/cataloger/debian/parse_dpkg_db_test.go
```diff
@@ -491,36 +490,31 @@ func Test_associateRelationships(t *testing.T) {
 		},
 		{
 			name:    "relationships from dpkg example docs",
-			fixture: "test-fixtures/status/doc-examples",
+			fixture: "test-fixtures/var/lib/dpkg/status.d/doc-examples",
 			wantRelationships: map[string][]string{
-				"made-up-package-1": {"kernel-headers-2.2.10", "hurd-dev", "gnumach-dev"},
-				"made-up-package-2": {"libluajit5.1-dev", "liblua5.1-dev"},
-				"made-up-package-3": {"foo", "bar"},
+				"made-up-package-1": {"gnumach-dev", "hurd-dev", "kernel-headers-2.2.10"},
```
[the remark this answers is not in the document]
wagoodman: This goes back to this comment https://github.com/anchore/syft/pull/2853#discussion_r1598563270 ultimately the deduping implementation is depending on an unordered set, so I'm sorting to ensure I'm getting stable results.

syft/pkg/cataloger/internal/dependency/resolver.go
kzantow: If someone is using this feature, they are going to need to know what each of the required functions do whether they're interfaces or separate functions. I won't disagree that having an interface that enforces 2 separate functions to be implemented is safer, to some degree. But I definitely think it's less convenient, especially for the case where packages could be identified using a common `name@version` string or something similar where the only requirement is to provide a function for _dependencies_, not _provides_. This could be wrapped up in yet another implementation in the `dependencies` package but that's more complexity because different catalogers are doing different things. If you feel strongly this is the right way to go, then that's ok with me 👍

syft/pkg/cataloger/debian/prosumer.go
```diff
@@ -0,0 +1,76 @@
...
+
+func stripVersionSpecifier(s string) string {
+	// examples:
+	// libgmp10 (>= 2:6.2.1+dfsg1)         -->  libgmp10
+	// libgmp10                            -->  libgmp10
+	// foo [i386]                          -->  foo
+	// default-mta | mail-transport-agent  -->  default-mta | mail-transport-agent
+	// kernel-headers-2.2.10 [!hurd-i386]  -->  kernel-headers-2.2.10
+
+	items := internal.SplitAny(s, "[(<>=")
```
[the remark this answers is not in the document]
wagoodman: point taken, I'll refactor these (+similar places/comments)

syft/pkg/cataloger/internal/dependency/resolver.go
willmurphyscode: I think I see what you're saying. You're saying, "The return value of `deduplcate` must be stable across multiple runs of Syft, but `set.List()` does not return a deterministic ordering. Therefore, sort the return value."

I think I was confused because the comment `// note; this must be a stable function` is right before the invocation of a function that is _not_ stable.

This is a tiny nit, but I think it would be clearer if the comment were moved to be above the call to `strings.Sort`, and said something like, "ensure order is stable across multiple invocations of Syft," or something like that.

syft/pkg/cataloger/debian/prosumer.go
wagoodman: same remediation as https://github.com/anchore/syft/pull/2853#discussion_r1598807461

syft/pkg/cataloger/internal/dependency/resolver.go
wagoodman: from an offline conversation we've found a good alternative approach that balances many of the comments. Short answer is replace multiple function calls / an interface with a single `func(pkg.Package) dependency.Specification` (where the specification has both provides and depends data). See the code for the full final state.

syft/pkg/cataloger/debian/parse_dpkg_db_test.go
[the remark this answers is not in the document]
kzantow: No reason -- just wondering if there was a way to make the tests less brittle. Nothing needed here.

syft/pkg/cataloger/internal/dependency/resolver.go
```diff
@@ -0,0 +1,107 @@
...
+// and provider claims from any given package.
+type RelationshipResolver struct {
+	specifier Specifier
+}
+
+func NewRelationshipResolver(s Specifier) RelationshipResolver {
+	return RelationshipResolver{
+		specifier: s,
+	}
+}
```
kzantow, line 38: do these types need to be exported, or do they need to exist at all? could the Resolve function below just be something like `resolveDependencies(Specifier, pkgs)`?

kzantow (approved):
👍

syft/pkg/cataloger/internal/dependency/resolver.go
wagoodman: agreed, I'll get rid of the type altogether

PR #2864, opened 2024-05-10
Added functionality to convert major, minor, patch to version

LaurentGoderre:
In cases where the binary classifier is able to pick up individually on a major, minor, and patch version fields, but not a semantic version, then the `version` parameter used will be filled in with `major.minor.patch` in the package metadata.


syft/pkg/cataloger/binary/classifier_test.go
wagoodman, line 1: it looks like this test has been added to test the effect of `version` being filled out somewhere, but it hasn't been explicitly tested yet. Maybe this is partially completed?

LaurentGoderre: I found that the tests don't actually test whether the version is right but whether it is detected so I would to add new test for both the current approach and the new one.

LaurentGoderre: Added the tests

spiffcs (approved):

spiffcs:
Note this is usful for binaries like [haxe](https://github.com/HaxeFoundation/haxe) which incorporate the major minor patch versions as metadata in the binary and not as a specific semver string.

PR #2865, opened 2024-05-11
Fix outdated links to SPDX

tnken:


wagoodman (approved):
thanks for the doc fixes 🙌

PR #2868, opened 2024-05-13
chore: fix small tooling error for go.mod

spiffcs:
## Summary
On certain IDE and versions of neovim that develop syft an error can be encountered when running `go list -m -json all`

This is because the IDE or editor tooling tries to run this command for all `go.mod` for a given source tree.

This PR fixes a `bogus` `go.mod` so that it still fulfills the test case, but doesn't break editor tooling:

### Example of error before fix
Screenshot 2024-05-13 at 11 25 59 AM


wagoodman (approved):
❤️ oh many, this has been really annoying, thanks for fixing this

PR #2869, opened 2024-05-13
Capture dependencies when parsing SPDX SBOMs

russellhaering:
When parsing SPDX SBOMs, translate DEPENDS_ON and DEPENDENCY_OF relationships into a dependency-of relationship in the Syft model.


kzantow (approved):
Thanks @russellhaering -- this looks good to me 👍

PR #2871, opened 2024-05-14
Use values in relationship To/From fields

wagoodman:
Related to https://github.com/anchore/grype/pull/1857

This PR adjusts all locations where we are writing references into the `To` and `From` fields of the `artifact.Relationship` type such that no variable references are used in those fields. This allows for consumers to more easily type assert out specific elements without having to also check references for those types.

A new ruleguard linter rule has been introduced to help prevent from this occurring in the future.

While fixing the test for this it became apparent that lifting up `cmp.Diff` options to a common package would be useful, thus the `internal/cmptest` package was created (should only be used in testing).

### DIFF
--- a/internal/relationship/by_file_ownership.go
+++ b/internal/relationship/by_file_ownership.go
@@ -59,8 +59,8 @@ func byFileOwnershipOverlap(catalog *pkg.Collection) []artifact.Relationship {
 			child := catalog.Package(childID)   // TODO: this is potentially expensive
 
 			edges = append(edges, artifact.Relationship{
-				From: parent,
-				To:   child,
+				From: *parent,
+				To:   *child,
 				Type: artifact.OwnershipByFileOverlapRelationship,
 				Data: ownershipByFilesMetadata{
 					Files: fs,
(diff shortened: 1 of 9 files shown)


internal/relationship/by_file_ownership.go
```diff
@@ -59,8 +59,8 @@ func byFileOwnershipOverlap(catalog *pkg.Collection) []artifact.Relationship {
 			child := catalog.Package(childID)   // TODO: this is potentially expensive
 
 			edges = append(edges, artifact.Relationship{
-				From: parent,
-				To:   child,
+				From: *parent,
+				To:   *child,
```
kzantow, line 63: These could potentially be `nil`, right?

test/rules/rules.go
```diff
@@ -2,14 +2,47 @@
...
+
+	// this rule defends against using pointers as values in artifact.Relationship
+	m.Match(
+		`$x{$*_, From: $y, $*_}`,
+		`$x{$*_, To: $y, $*_}`,
+		`$x.From = $y`,
+		`$x.To = $y`,
+	).
+		Where(isRelationship(m) && hasPointerType(m)).
+		Report("pointer used as a value for From/To field in artifact.Relationship")
```
willmurphyscode, line 47: In the report, can we say what to do instead?

internal/relationship/by_file_ownership.go
```diff
@@ -59,8 +59,8 @@ func byFileOwnershipOverlap(catalog *pkg.Collection) []artifact.Relationship {
 			child := catalog.Package(childID)   // TODO: this is potentially expensive
 
 			edges = append(edges, artifact.Relationship{
-				From: parent,
-				To:   child,
+				From: *parent,
```
willmurphyscode, line 62: Do these need nil checks now?

willmurphyscode (approved):

PR #2872, opened 2024-05-14
Add support for RPM DB package relationships

wagoodman:
Related to #572 

Adds support for tracking RPM to RPM dependencies (from an RPM DB, not from RPM files) as package relationships.

Note: there are several cases where package evidence seems to support self-referential dependencies (e.g. bash depends on bash) however, the RPM/yum/dnf tooling does not support this conclusion. For this reason this cataloger will prune all self-referential RPM DB package relationships.


syft/pkg/cataloger/redhat/cataloger.go
```diff
@@ -20,7 +22,23 @@ func NewDBCataloger() pkg.Cataloger {
...
 		WithParserByGlobs(parseRpmDB, pkg.RpmDBGlob).
-		WithParserByGlobs(parseRpmManifest, pkg.RpmManifestGlob)
+		WithParserByGlobs(parseRpmManifest, pkg.RpmManifestGlob).
+		WithProcessors(dependency.Processor(dbEntryDependencySpecifier), denySelfReferences)
+}
+
+func denySelfReferences(pkgs []pkg.Package, rels []artifact.Relationship, err error) ([]pkg.Package, []artifact.Relationship, error) {
+	// it can be common for dependency evidence to be self-referential (e.g. bash depends on bash), which is not useful
+	// for the dependency graph, thus we remove these cases
+	for i := 0; i < len(rels); i++ {
```
spiffcs, line 32: I see that we're decrementing `i` to account for the filter. Are we doing the loop this way to avoid doubling the relationship allocations?

spiffcs (approved):
👍 🟢 

Just had a question about how we're doing the allocation for relationship filtering here

syft/pkg/cataloger/redhat/cataloger.go
wagoodman: exactly -- if there are no self referential package relationships then we aren't allocating another slice

PR #2878, opened 2024-05-16
Fix DecoderCollection discarding input from non-seekable Readers

russellhaering:
This fixes a bug where a `DecoderCollection` would inadvertently consume from the input stream it was given while iterating through candidate formats, unless it was passed an `io.ReadSeeker`.

This doesn't seem to currently affect the CLI.

### DIFF
--- a/syft/format/decoders_collection.go
+++ b/syft/format/decoders_collection.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/format/internal/stream"
 	"github.com/anchore/syft/syft/sbom"
 )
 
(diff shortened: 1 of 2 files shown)


syft/format/decoders_collection.go
```diff
@@ -45,10 +52,16 @@ func (c *DecoderCollection) Decode(reader io.Reader) (*sbom.SBOM, sbom.FormatID,
...
 // Identify takes a set of bytes and attempts to identify the format of the SBOM relative to the decoders in the collection.
-func (c *DecoderCollection) Identify(reader io.Reader) (sbom.FormatID, string) {
-	if reader == nil {
+func (c *DecoderCollection) Identify(r io.Reader) (sbom.FormatID, string) {
+	if r == nil {
 		return "", ""
 	}
+
+	reader, err := stream.SeekableReader(r)
+	if err != nil {
```
kzantow, line 61: this should probably get logged with `log.Debugf("unable to create seekable reader: %v", err)` or similar

russellhaering: [Addressed](https://github.com/anchore/syft/compare/445629088bc76ff4a4fd5f83e582dcde26c9b34f..c0380cecf43580fbc2de65959d826219ad5c69a3)

kzantow (approved):
LGTM, thanks @russellhaering !

PR #2882, opened 2024-05-17
chore: Reduce length of readme, moving lengthy content to the wiki

popey:
👋 

This PR reduces the length and complexity of the README.md. 

I have removed a lot of additional information and put it in the [wiki](https://github.com/anchore/syft/wiki), linking to those pages from within the README. So, no information has been lost; it's just moved out of the project's front page.

The goal was to make the information here in the README more accessible and less overwhelming. 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,24 +2,26 @@
     <img src="https://user-images.githubusercontent.com/5199289/136844524-1527b09f-c5cb-4aa9-be54-5aa92a6086c1.png" width="271" alt="Cute pink owl syft logo">
 </p>
 
-[![Validations](https://github.com/anchore/syft/actions/workflows/validations.yaml/badge.svg)](https://github.com/anchore/syft/actions/workflows/validations.yaml)
-[![Go Report Card](https://goreportcard.com/badge/github.com/anchore/syft)](https://goreportcard.com/report/github.com/anchore/syft)
-[![GitHub release](https://img.shields.io/github/release/anchore/syft.svg)](https://github.com/anchore/syft/releases/latest)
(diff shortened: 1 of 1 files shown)


kzantow:
I wonder if the _output format_ section should remain in the main readme or have some other reference in the basic usage section. The reason being things like "works seamlessly with Grype" are not true without this (the default table output will do nothing in Grype), and it's probably fairly fundamental to what people would use Syft for: if they want an SBOM, a table listing really isn't an SBOM and it definitely isn't an SBOM format that most people would want to use, so this seems like it should be very obvious up-front how to use this basic functionality. What do you think?

spiffcs:
👋 Hey @popey! This is 🟢 on my end. I think the final bit for contributing is the DCO:

To add your Signed-off-by line to every commit in this branch:

Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
In your local branch, run: `git rebase HEAD~2 --signoff`
Force push your changes to overwrite the branch: `git push --force-with-lease origin rationalise-readme`

spiffcs (approved):

betaboon:
heya, even tho this has already been merged, I'd like to chime in.

Moving these information to the wiki has the downside that they are separately versioned.
this will make it hard if not impossible to to correlate the correct version of the information to the state of the repository.

Imagine i need to see the information on "Configuration" for a release of the past.

imho a better option would have been to move those sections from the readme into a `docs/configuration.md` etc, thus keeping them versioned together with the code.

kzantow:
Hey @betaboon -- just an FYI we are planning on introducing a `config` command so you can see exactly what your current version of syft supports: https://github.com/anchore/syft/pull/2892

PR #2890, opened 2024-05-21
feat: add license field to ELF binary package metadata

brian-ebarb:
Adds a field for License information into the elf binary package metadata structure, which is set to omit if empty from sbom.


wagoodman:
I've followed up offline about this PR -- what's needed is to wire up the existing license being parsed to the package constructor

wagoodman (approved):

PR #2891, opened 2024-05-22
Go Mod Cataloger: Remove Replaced Packages

russellhaering:
When the goModCataloger processes a Replace directive it currently adds the new package to the resulting package list, but does not remove the old one unless the path is unchanged.

Based on an existing comment in the code, removing the old one seems to be the intended behavior, and results in a more expected end-result, so this does so.


willmurphyscode (approved):

willmurphyscode:
Thanks @russellhaering! Good catch, and I appreciate the added test.

PR #2914, opened 2024-05-30
Use redhat as namespace for redhat rpms

ralphbean:
The namespace value of `redhat` signifies this as an RPM package produced and distributed by Red Hat.

Using "rhel" in the namespace is not correct.


kzantow:
Hi @ralphbean -- thanks for the contribution. Do you have some supporting information why _"Using "rhel" in the namespace is not correct."_? The [RPM PURL spec types](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm) doesn't mention either, unfortunately.

wagoodman (approved):
@ralphbean I agree with your point, from the pURL type reference for `rpm`:

> The `namespace` is the vendor such as Fedora or OpenSUSE.

...thus here the vendor is RedHat. There are probably other changes to the distro-to-namespace translations --we can leave that for a future enhancement.

ralphbean:
Thanks so much @wagoodman!

ralphbean:
Adding this retroacticely to the conversation: https://github.com/RedHatProductSecurity/security-data-guidelines/blob/main/docs/purl.md :)

PR #2922, opened 2024-06-03
Close sqlite driver after testing sqlite availability

ttc0419:
For long running programs, opening a sql connection without closing it will cause go routine leaks.  Close sql driver after testing sqlite availability.


wagoodman (approved):
Great catch! 🙌

PR #2933, opened 2024-06-05
Add classifier for util-linux

LaurentGoderre:
For distro that don't install it via package manager


kzantow (approved):

kzantow:
Thanks @LaurentGoderre !

PR #2959, opened 2024-06-13
chore: use go to bootstrap downloading and running task

kzantow:


.binny.yaml
```diff
@@ -86,6 +95,10 @@ tools:
     method: github-release
     with:
       repo: wagoodman/go-bouncer
+#    method: go-install
+#    with:
+#      module: github.com/wagoodman/go-bouncer
```
kzantow, line 100: NOTE this does not work to go-install due to a step during the release which exports data from the Google license classifier: https://github.com/wagoodman/go-bouncer/blob/a28bee0a43c0c55883c7fee39c691a52d10ad0f0/Makefile#L66

PR #2965, opened 2024-06-14
fix: fix parsing for complex toml types

### LINKED ISSUE #2954: Issue scanning Poetry Project with Syft 1.6 and cataloger=python-package-cataloger
**What happened**:
We have updated Syft from Version 1.5.0 to 1.6.0 and discovered that it has issues scanning certain poetry.lock files. We try to scan a poetry project with the syft 1.6.0 with executing the following command:

```console
syft --output cyclonedx-json=reports/sbom-scan-licenses.cdx.json --source-name=blub scan .
````

and then we discovered that the output of the console has the following warning:

```console
[0001]  WARN cataloger failed cataloger=python-package-cataloger error=unable to parse poetry.lock: (0, 0): 
version = ">=1.0,<3"
markers = "platform_system == \"Windows\""
version = ">=1.6,<3"
]([]*toml.Tree) to trees location=/poetry.lock
```
and the json looks like followed:

```json
{
  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
  "serialNumber": "urn:uuid:dac4442f-fb9e-4168-b204-567579c0314e",
  "version": 1,
  "metadata": {
    "timestamp": "2024-06-13T15:31:24+02:00",
    "tools": {
      "components": [
        {
          "type": "application",
          "author": "anchore",
          "name": "syft",
          "version": "1.6.0"
        }
      ]
    },
    "component": {
      "bom-ref": "af63bd4c8601b7f1",
      "type": "file",
      "name": "blub"
    }
  }
}
```

With Syft version 1.5.0, it was possible to scan the lock file without any issues. If we remove the following part (see below) from the lock file then Syft 1.6.0 works. We can not do that because other teams are maintaining the lock files. 

```toml
[package.dependencies]
msal = ">=0.4.1,<2.0.0"
packaging = "*"
portalocker = [
    {version = ">=1.0,<3", markers = "platform_system != \"Windows\""},
    {version = ">=1.6,<3", markers = "platform_system == \"Windows\""},
]
```

I would expect that syft 1.6.0 has the same behaviour as the syft 1.5.0.

Create poetry lock file and add the following snippet:

```toml
[package.dependencies]
msal = ">=0.4.1,<2.0.0"
packaging = "*"
portalocker = [
    {version = ">=1.0,<3", markers = "platform_system != \"Windows\""},
    {version = ">=1.6,<3", markers = "platform_system == \"Windows\""},
]
```
and execute  the following command within the poetry project:

```console
syft --output cyclonedx-json=reports/sbom-scan-licenses.cdx.json --source-name=blub scan .
```

Currently nothing

**Environment**:
- Output of `syft version`: 1.6
- OS (e.g: `cat /etc/os-release` or similar): macOS 14.5

spiffcs:
### Fixes #2954
The poetry lock file allows for different types to be set as values when parsing dependencies:
```
msl = "x.x.x" <-- string
postlocker = [
    {version: "string", markers: "string"} 
    {version: "string", markers: "string"}
]
```

Because of this we needed to do two things.

We needed a new Toml library that allowed us to hook into `UnmarshalTOML` and we needed to write a custom unmarshal function for the complex type that is a poetry dependency. We do this using the new primitive toml type from the newly imported toml library.

Tests have been updated to reflect the correct state of a poetry.lock file.


syft/pkg/cataloger/python/parse_poetry_lock.go
```diff
@@ -44,39 +46,151 @@ type poetryPackageDependency struct {
...
-func (d *poetryPackageDependency) UnmarshalText(data []byte) error {
-	// attempt to parse as a map first
-	var dep map[string]interface{}
-	if err := toml.Unmarshal(data, &dep); err == nil {
-		if extras, ok := dep["extras"]; ok {
-			if extrasList, ok := extras.([]string); ok {
-				d.Extras = extrasList
-			}
+func (p *poetryPackage) UnmarshalTOML(data any) error {
+	d, _ := data.(map[string]any)
```
wagoodman, line 50: we should check for errors here

syft/pkg/cataloger/python/parse_poetry_lock.go
```diff
@@ -93,15 +60,31 @@ func parsePoetryLock(_ context.Context, _ file.Resolver, _ *generic.Environment,
...
+	for i, p := range metadata.Packages {
+		dependencies := make(map[string][]poetryPackageDependency)
+		for pkgName, du := range p.DependenciesUnmarshal {
+			var (
+				single    string
+				singleObj poetryPackageDependency
+				multiObj  []poetryPackageDependency
+			)
+
+			switch {
```
wagoodman, line 78: if we reach the end of this switch, and no action was taken (no case hit) then we dropped a field. I thikn we should trace log that

spiffcs: 👍 you're right let me add that

syft/pkg/cataloger/python/test-fixtures/poetry/dev-deps/poetry.lock
```diff
@@ -11,6 +11,12 @@ docutils = "*"
 natsort = "*"
 six = "*"
 sphinx = "*"
+packaging = "*"
+msal = ">=0.4.1,<2.0.0"
+portalocker = [
+    {version = ">=1.0,<3", markers = "platform_system != \"Windows\""},
+    {version = ">=1.6,<3", markers = "platform_system == \"Windows\""},
+]
```
wagoodman, line 19: we have a test case for a `string` value a `[]obj`, but not for just `obj` -- since the parsing code handles this case can you add one more dependency example in this fixture to make certain it's covered.

Also, it would be ideal to have a poetry.lock that is malformed in some way (e.g. `portalocker = [ [ [ { version = "1.2" }] ] ] )`) to show that we now survive this kind of parsing error

spiffcs: @wagoodman I don't we we do survive this with the parser.

The parser will return an error if we encounter a malformed poetry file.
```
failed to read poetry lock package: toml: line 10 (last key "package.dependencies"): expected a top-level item to end with a newline, comment, or EOF, but got ')' instead
```

Do you want to assert that the parser is returning this error?

wagoodman: Correction:
```
portalocker = [ [ [ { version = "1.2" } ] ] ] 
```

wagoodman (approved):
I added

PR #2978, opened 2024-06-20
Add 1.6 to supported cyclonedx spec versions.

### LINKED ISSUE #2974: Add CycloneDX 1.6 Support

Support for CycloneDX 1.6 Specification. 

**Why is this needed**:
Teams and Tools that leverage Syft as part of the SW supply chain security have a need for the additional fields and to avoid compatibility issues when working with sboms 

**Additional context**:
The underlying go library used for cyclonedx has added support for 1.6 (as of v0.9.0) and is already in use with syft as of [this](https://github.com/anchore/syft/pull/2931) PR.

ragaskar:
- Resolves #2974

I had quite the time figuring out what folks were doing for the test-fixtures involved with expanding the supported spec range. In the end I determined it's a copy/paste w/ an updated spec version and a new unique serial number (I was doing some really silly things like compiling a syft snapshot with the default set to 1.6 in order to emit a fixture file). 

There doesn't appear to be any documentation re: supported versions (outside out what is probably generated via the encoder/decoder interface itself). There are a few string mentions of spec versions, but these are input data for tests and not related to the actual supported versions ([e.g.]( https://github.com/anchore/syft/blob/e0e1c4ba0a5ec546692428efce4a3bb9b8bc250d/cmd/syft/internal/options/writer_test.go#L270))

unit/integration tests pass (adding a new version to the [list of supported versions](https://github.com/anchore/syft/blob/7392d607b6d275dfaf033d833ed3780b403a6c1b/syft/format/internal/cyclonedxutil/versions.go#L16) in versions.go causes additional checks to be run) ... Not sure if there's anything else to check here?

EDIT: Please note, I did _not_ bump the default cyclonedx spec version (so I believe by default syft will still emit v1.5 spec cdx, although it will decode 1.6 and can be configured to emit 1.6). I believe that is a change with the potential for wide user impact (👋 ) and should probably be made by core maintainers.

### DIFF
--- a/schema/cyclonedx/README.md
+++ b/schema/cyclonedx/README.md
@@ -7,3 +7,12 @@ For this reason we've included a copy of all schemas needed to validate `syft` o
 to reference local copies of dependent schemas.
 
 You can get the latest schemas from the [CycloneDX specifications repo](https://github.com/CycloneDX/specification/tree/master/schema).
+
+When the spec version is bumped an approach to determining prior modifications is to compare the 
+prior spec version (e.g. if updating to 1.7, compare the files in this directory against the 1.6 
+equivalents). 
+
+One can also update the schemas and observe the errors in order to make the necessary updates. 
+At the time of writing, the cyclonedx.xsd needed modifications to link to the local spdx.xsd, 
+and also to changes the minOccurs for a license tag to 0. (The json schema does not require 
+modification for the generated file to lint properly, but can simply be copy/pasted).   
--- a/schema/cyclonedx/cyclonedx.json
+++ b/schema/cyclonedx/cyclonedx.json
@@ -1,8 +1,8 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "http://cyclonedx.org/schema/bom-1.5.schema.json",
+  "$id": "http://cyclonedx.org/schema/bom-1.6.schema.json",
   "type": "object",
-  "title": "CycloneDX Software Bill of Materials Standard",
+  "title": "CycloneDX Bill of Materials Standard",
   "$comment" : "CycloneDX JSON schema is published under the terms of the Apache License 2.0.",
   "required": [
     "bomFormat",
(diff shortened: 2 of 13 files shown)


ragaskar:
I see it barfing on the xml schema check ... 🤔 GoLand was also not happy. Let me see if I can replicate + fix.

ragaskar:
breadcrumb for others: there's instructions [here](https://github.com/anchore/syft/blob/7392d607b6d275dfaf033d833ed3780b403a6c1b/schema/cyclonedx/README.md#L10-L9)

> modified to reference local copies of dependent schemas.

oof hope this is pretty straightforward.

ragaskar:
I was surprised the schema validation tests generated 1.6 boms by default, even with `DefaultVersion` set to "1.5"

I am seeing that by default syft is generating cdx 1.6 spec boms regardless of this DefaultVersion setting. I am going to change it to match for now and will take another peek at why this is occurring. As mentioned above, I think the impact of bumping the default spec version emitted is fairly broad. I suppose users _can lock back_ to 1.5 if desired w/ in their syft format config (maybe there's already good documentation for this, but I'm just now realizing that was an option for me as well 🤦 ). 

Perhaps if someone could comment on expected behavior here?

ragaskar:
Ok, ultimately, when an output format is passed (for a format that supports "all versions") like -o cyclonedx-json=some-file.json (WHICH! by the way, can also take a format/version pair e.g. cyclonedx-json@1.5=some-file.json), there's some logic that just takes the most recent version (see [here](https://github.com/anchore/syft/blob/e0e1c4ba0a5ec546692428efce4a3bb9b8bc250d/syft/format/encoders_collection.go#L73)). At this point I'm not sure I'm going to dig into the reasons behind also having the DefaultVersion constant; I'm just going to change it to match, since effectively 1.6 is now the default (since it's the most recent). Users who do _not_ want to generate 1.6 can either pin back to 1.5 in their config file or directly via the flag as described above.

ragaskar:
well, it does appear the DefaultVersion value _is_ used by test, and I'm having a heck of a time figuring out the right incantation to regenerate the correct fixtures. If someone could advise that would be wonderful.

spiffcs:
Thanks a million @ragaskar this looks great. Pulled it down locally and was able to generate 1.6 as the default.

I agree with the change that syft should generate 1.6 by default - I think it was just missed when we did the latest cyclone-dx library upgrades. 

Thank you so much for the contribution! 

If you ever want to talk about cyclone-dx or have other PR/Questions come check out our community meetings =)

PR #2996, opened 2024-06-25
Add detection of Erlang in Alpine linux

LaurentGoderre:


spiffcs (approved):
🥳 Thanks @LaurentGoderre this is really good!

Are there any other file globs or extensions you wanted to add here? I gave it the green given CI has run to green and the tests LGTM.

LaurentGoderre:
Hopefully that is it

PR #2998, opened 2024-06-26
feat: report unknowns in sbom

kzantow:
This PR adds information to the file model which allows surfacing "unknowns". Previously, when scanning a source, Syft runs a number of catalogers which create packages from the files found. If an error happens, packages simply would not be created, and some logging about the error would occur. With this change, many of these errors are returned and added as context to the files output in the SBOM. Examples of "unknowns" included by this PR:
* executable files which did not result in identified packages
* archives which were not scanned, or did not result in packages identified
* errors when reading information such as invalid JSON, or corrupted binary ELF data
* ...

This PR has a set of post-cataloging steps that perform the following:
* identify archives in the scan target which do not have packages reported, and label them as unknowns
* _remove_ all unknowns from files which have locations present in packages (in other words: only leave files labeled as unknowns which have no packages)

NOTE: if you would like to experiment with this, you can select the locations and unknowns from the Syft JSON like this:
```
go run ./cmd/syft maven:latest -o json | jq '.files.[]|select(.unknowns)|{location,unknowns}'
```

TODO:
- [x] configuration
- [x] assess if the unknowns in this PR are useful
- [x] add test coverage

Fixes: #518


syft/file/cataloger/executable/cataloger.go
```diff
@@ -165,16 +167,19 @@ func processExecutable(loc file.Location, reader unionreader.UnionReader) (*file
 
 	switch format {
 	case file.ELF:
-		if err := findELFFeatures(&data, reader); err != nil {
+		if err = findELFFeatures(&data, reader); err != nil {
 			log.WithFields("error", err).Tracef("unable to determine ELF features for %q", loc.RealPath)
+			err = fmt.Errorf("unable to determine ELF features: %w", err)
```
wagoodman, line 172:
```suggestion
			err = fmt.Errorf("unable to determine ELF features for %q: %w", loc.Path(), err)
```

same comment applies to surrounding cases too

wagoodman: nvm -- this is already part of the file object *facepalm*

syft/cataloging/unknowns.go
```diff
@@ -0,0 +1,15 @@
+package cataloging
+
+type UnknownsConfig struct {
+	RemoveWhenPackagesDefined         bool
+	IncludeExecutablesWithoutPackages bool
+	IncludeUnexpandedArchives         bool
```
kzantow, line 6: These 3 options are part of the public API and should be looked at perhaps with more scrutiny than other aspects of this PR

syft/pkg/cataloger/javascript/parse_package_json.go
```diff
@@ -64,11 +63,8 @@ func parsePackageJSON(_ context.Context, _ file.Resolver, _ *generic.Environment
 			return nil, nil, fmt.Errorf("failed to parse package.json file: %w", err)
 		}
 
-		if !p.hasNameAndVersionValues() {
-			log.Debugf("encountered package.json file without a name and/or version field, ignoring (path=%q)", reader.Path())
-			return nil, nil, nil
-		}
-
+		// always create a package, regardless of having a valid name and/or version,
+		// a compliance filter later will remove these packages based on compliance rules
```
kzantow, line 67: This is a small, but potentially impactful change and should be scrutinized. This will result in the cataloger returning packages with empty name and/or versions, whereas these were being entirely dropped before. There are quite a lot of these files in the sample set of image I tested on that have no name. These will be filtered out AFTER being returned by compliance rules. There will be a number of packages NOT filtered out that do not have versions, however, which will decidedly be a change that some users notice.

syft/file/cataloger/executable/elf.go
```diff
@@ -254,7 +255,7 @@ func elfHasExports(f *elf.File) bool {
 	// really anything that is not marked with 'U' (undefined) is considered an export.
 	symbols, err := f.DynamicSymbols()
 	if err != nil {
-		// TODO: known-unknowns?
+		*errs = unknown.Joinf(*errs, "unable to get ELF dynamic symbols: %w", err)
```
wagoodman, line 258: this probably could just be a trace log and not really an unknown

syft/file/cataloger/executable/elf.go
```diff
@@ -30,40 +31,40 @@ func findELFFeatures(data *file.Executable, reader unionreader.UnionReader) erro
...
 	data.HasEntrypoint = elfHasEntrypoint(f)
-	data.HasExports = elfHasExports(f)
+	data.HasExports = elfHasExports(&err, f)
 
-	return nil
+	return err
 }
 
-func findELFSecurityFeatures(f *elf.File) *file.ELFSecurityFeatures {
+func findELFSecurityFeatures(errs *error, f *elf.File) *file.ELFSecurityFeatures {
```
wagoodman, line 41: based off of the direction of the known-unknowns work it might be that most of these binary cases are really not known-unknown candidates, since if they were known would not result in any more nodes in the dependency tree.

wagoodman (approved):

Issue #2759, opened 2024-04-08
Support Swift Package Manager Package.resolved schema version 3


maxgip:
**What happened**:

When running `syft` against a project using Swift Package Manager and a version 3 Package.resolved file, an error occured (`error=unknown swift package manager version, 3.000000 location=/Package.resolved`) and the sbom output was incomplete:

```
swift-embedded-examples(main*)$ syft scan ./stm32-neopixel -o cyclonedx-json=sbom.json
 ✔ Indexed file system                                                                                                                                  stm32-neopixel
 ✔ Cataloged contents                                                                                 6ba85f929a0d558c5687b1245352e16c6ee8aa429acbdff263954992966a54e8
   ├── ✔ Packages                        [0 packages]
   └── ✔ Executables                     [0 executables]
[0000]  WARN no explicit name and version provided for directory source, deriving artifact ID from the given path (which is not ideal)
[0000]  WARN cataloger failed cataloger=swift-package-manager-cataloger error=unknown swift package manager version, 3.000000 location=/Package.resolved
swift-embedded-examples(main*)$ cat sbom.json
{"$schema":"http://cyclonedx.org/schema/bom-1.5.schema.json","bomFormat":"CycloneDX","specVersion":"1.5","serialNumber":"urn:uuid:75f909ea-e5a5-49d9-9e9e-73bb23a611a7","version":1,"metadata":{"timestamp":"2024-04-08T10:23:55-04:00","tools":{"components":[{"type":"application","author":"anchore","name":"syft","version":"1.1.1"}]},"component":{"bom-ref":"06497d32a1d71b4c","type":"file","name":"./stm32-neopixel"}}}
swift-embedded-examples(main*)$ cat sbom.json | jq
{
  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
  "bomFormat": "CycloneDX",
  "specVersion": "1.5",
  "serialNumber": "urn:uuid:75f909ea-e5a5-49d9-9e9e-73bb23a611a7",
{
  "version": 1,
  "metadata": {
    "timestamp": "2024-04-08T10:23:55-04:00",
    "tools": {
      "components": [
        {
          "type": "application",
          "author": "anchore",
          "name": "syft",
          "version": "1.1.1"
        }
      ]
    },
    "component": {
      "bom-ref": "06497d32a1d71b4c",
      "type": "file",
      "name": "./stm32-neopixel"
    }
  }
}
```

**What you expected to happen**:

[...]

PR #3001, opened 2024-06-28
Add version 3 support for swift package manager of the resolved files

4ell0:
Add version 3 support for swift package manager of the resolved files, add test coverage for v3 support as well.

This is a fix for : https://github.com/anchore/syft/issues/2759

### DIFF
--- a/syft/pkg/cataloger/binary/test-fixtures/config.yaml
+++ b/syft/pkg/cataloger/binary/test-fixtures/config.yaml
@@ -78,12 +78,12 @@ from-images:
     paths:
       - /usr/local/go/bin/go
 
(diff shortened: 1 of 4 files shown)


kzantow:
Thanks for the contribution, @4ell0! It looks like [there is a missing test file](https://github.com/anchore/syft/actions/runs/9719249578/job/26833932230?pr=3001#step:11:1717) (`PackageV3.resolved`). It looks like the V2 and V3 cases are identical (and we're not using the extra field), would it be easier to just add `3` to the case? (e.g. `case 2, 3:`)

kzantow (approved):
Thanks for the contribution, @4ell0!

PR #3054, opened 2024-07-22
Ensure accurate java main artifact name retrieval for multi-JARs and refine fallback approach

dor-hayun:
**Improving Accuracy of Package Name Retrieval in Java Archives**

This section outlines enhancements to accurately retrieve the main package artifact names and their corresponding SHA1 values. It focuses on resolving issues where incorrect package names were being retrieved, especially for JAR files with multiple internal JARs. The improvements include:

- Correctly retrieving the main package name when the main POM file is present.
- Addressing issues with incorrect package names for specific JAR files, such as those containing multiple JARs.
- Ensuring accurate results by checking prefixes and suffixes, with a fallback mechanism if no exact matches are found.

### DIFF
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -301,11 +301,17 @@ func (j *archiveParser) guessMainPackageNameAndVersionFromPomInfo(ctx context.Co
 	properties, _ := pomPropertiesByParentPath(j.archivePath, j.location, pomPropertyMatches)
 	projects, _ := pomProjectByParentPath(j.archivePath, j.location, pomMatches)
 
+	// map of all the artifacts in the pom properties, in order to chek exact match with the filename
+	artifactsMap := make(map[string]bool)
+	for _, propertiesObj := range properties {
+		artifactsMap[propertiesObj.ArtifactID] = true
+	}
+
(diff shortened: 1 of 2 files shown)


spiffcs:
Thanks for the contribution @dor-hayun!

I think for us to accept this we need to check on some of the downstream implications here and see if it changes any of our fixtures in vulnerability testing for the better.

Also, let me see about adding a test or two that can show the kind of behavior this prevents. Do you have an current cases you're running into that would serve as a good example?

dor-hayun:
HI @spiffcs ,
Please try to run syft on the following public image:
public.ecr.aws/docker/library/gradle@sha256:70da12adf27e83bcc125af9d2bc6f9432590e89c96609625aa688135b27e75fb

and then you can check what happens for 'jansi' package, it is the main artifact, you can see that there are 10 pom properties found inside the jar and my fix is to ensure Correctly retrieving the main package name when the main POM file is present.



image

kzantow:
Hey, @dor-hayun. It looks like I made a change that conflicts with this one a bit. Would you want to rebase this PR? Or I could push an update, if you don't mind.

dor-hayun:
@kzantow i'm rebasing it

spiffcs (approved):

Issue #3064, opened 2024-07-24
Python packages: name normalization


Mikcl:
The `pypi` components that `syft` generates does not normalize the python package names.

As per the python packaging [documentation](https://packaging.python.org/en/latest/specifications/name-normalization/), the following are the same:
```
    friendly-bard (normalized form)
    Friendly-Bard
    FRIENDLY-BARD
    friendly.bard
    friendly_bard
    friendly--bard
    FrIeNdLy-._.-bArD (a terrible way to write a name, but it is valid)
```

And so python packaging tools such as `pip` would understand how to treat these packages.

With normalization applied to syft sboms, it would make querying for packages less reliant on the specific display name that happens to be used. However, currently without normalization applied, consumers of syft sboms are at risk of not using the correct package names, and thus missing any checks that may happen afterward.

**What you expected to happen**:

I would expect the python package names to be normalized according to the docs/specification above.

Additionally this is mentioned in the PURL documentation: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#pypi

So i would expect this to be applied to the `purl` and the `name`

---

Is this something the syft team are will to consider adopting? 
Are there any consumer of syft that explictly require the denormalized form and would not work with the normalized form?

PR #3069, opened 2024-07-24
python-cataloger: normalize package names

Mikcl:
Fixes https://github.com/anchore/syft/issues/3064 

(provided there is agreement to normalize the names, discussion in the issue).

This PR adds a normalization function according to python packaging specification https://packaging.python.org/en/latest/specifications/name-normalization/ to each of the package types. And adds/updates existing tests.

The name and the purl are updated, the "metadata" still preserves the unnormalized name.

### DIFF
--- a/syft/pkg/cataloger/python/package.go
+++ b/syft/pkg/cataloger/python/package.go
@@ -2,6 +2,8 @@ package python
 
 import (
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/anchore/packageurl-go"
 	"github.com/anchore/syft/internal/licenses"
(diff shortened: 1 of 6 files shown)


syft/pkg/cataloger/python/package.go
```diff
@@ -10,7 +12,16 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
+func normalize(name string) string {
+	// https://packaging.python.org/en/latest/specifications/name-normalization/
+	re := regexp.MustCompile(`[-_.]+`)
```
mariocj89, line 17: For reference, the Python implementation is identical: https://github.com/pypa/packaging/blob/bb61b7bcbc429b49e6659a8b6c2966b6ebd046b6/src/packaging/utils.py#L49C13-L49C32

spiffcs:
Nice! There are no matching or downstream concerns here given that we already normalize these values for the PURL when using grype. This change should help consumers of syft SBOM going forward so 🟢

spiffcs (approved):

PR #1966, opened 2023-07-27
Guess unpinned versions in python requirements.txt

wagoodman:
Expands on #1597 by exposing a new `python.guess-unpinned-requirements` configuration, allowing loose requirements to be filled in with the lowest expressible version instead of dropping the package altogether. Note: since this is synthesizing version information this is an opt-in feature.

This replaces the v10 schema changes in https://github.com/anchore/syft/pull/1967 since they have not been released yet (a condition specifically allowed in the schema readme). The small change is to allow for some elements to be optional in the requirements metadata (everything except for name and version).

PR #3070, opened 2024-07-24
python requirements.txt cataloger: allow dots in python package names

Mikcl:
Its perfectly valid for python packages to have dots in their name
https://packaging.python.org/en/latest/specifications/name-normalization/

It seems like https://github.com/anchore/syft/pull/1966 introduced a regex match which did not include the dot. This results in silently skipping packages with dots in their requirements.txt.

This PR accepts the dot and adds a test.

Noticed this whilst doing https://github.com/anchore/syft/pull/3069 (which does the normalization of the name)

(its possible that this bug may exist in the other parsers but did not investigate that)

### DIFF
--- a/syft/pkg/cataloger/python/parse_requirements.go
+++ b/syft/pkg/cataloger/python/parse_requirements.go
@@ -26,7 +26,7 @@ const (
 	//      --hash=sha256:e9535b8c84dc9571a48999094fda7f33e63c3f1b74f3e5f3ac0105a58405bb65  # some comment
 
 	// namePattern matches: requests[security]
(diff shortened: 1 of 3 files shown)


spiffcs (approved):

spiffcs:
@Mikcl thank you so much for the two PR regarding the python cataloger

These will help a lot going into next release and we really appreciate the time and good PR description/submissions.

Mikcl:
Thanks for the quick review turn around @spiffcs 

note the test case in this pr should now be updated since the name and purl are normalised, from the other pr made. 

Thanks again

Mikcl:
put up a pr to fix it
https://github.com/anchore/syft/pull/3073

PR #3073, opened 2024-07-25
python-cataloger: fix normalization test

Mikcl:
Fixes the normalization in the test case introduced in https://github.com/anchore/syft/pull/3070


spiffcs (approved):

spiffcs:
There might be one more test case that is failing - I'll check and follow up this PR with any more required fixes

PR #3087, opened 2024-07-31
fix: update mainModuleVersion function to always prefix `v` to findings

### LINKED ISSUE #3086: Version parsing regression for Go binaries
**What happened**:

Coming from Syft v1.9.0 to v1.10.0, Syft changed how it parses versions for Go binary main modules: cases where where it used to surface `v1.2.3` (mapping exactly to the module's version) now show as `1.2.3`, which is less correct. Ideally we'd keep the correct strings in place, and avoid introducing more change than necessary to existing behavior.

Note that _non-main_ modules appear to still parse correctly (with `v` if the module itself has `v`).

No change from v1.9.0 — previous behavior was correct here. 🙏 

(using zsh or similar)

```console
$ syft -q =(curl -sL https://packages.wolfi.dev/os/aarch64/thanos-0.35.1-r3.apk | tar -Oxz usr/bin/thanos)
...
github.com/thanos-io/thanos                                                          0.35.1                                     go-module
github.com/themihai/gomemcache                                                       v0.0.0-20180902122335-24332e2d58ab         go-module
github.com/tklauser/go-sysconf                                                       v0.3.10                                    go-module
github.com/tklauser/numcpus                                                          v0.4.0                                     go-module
...
```

Here, github.com/thanos-io/thanos is the main module. And it doesn't have any versions/tags of `0.35.1` — see https://github.com/thanos-io/thanos/tags.

Some discussion in the community Slack before filing this: https://anchorecommunity.slack.com/archives/C019BUXV7R6/p1722433767717669

**Environment**:
- Output of `syft version`:

```
Application: syft
Version:    1.10.0
BuildDate:  2024-07-30T16:13:26Z
GitCommit:  a4b5dcd0df80f6a58c8610e25104647710c1da5d
GitDescription: v1.10.0
Platform:   darwin/arm64
GoVersion:  go1.22.5
Compiler:   gc
```

- OS (e.g: `cat /etc/os-release` or similar): latest macOS

spiffcs:
## Summary
Fixes #3086 

For a summary on what happened and the regression this fixes see this comment:
https://github.com/anchore/syft/issues/3086#issuecomment-2261179116


kzantow (approved):
👍

Issue #2723, opened 2024-03-18
Add time taken for each cataloger


tomerse-sg:
**What would you like to be added**:
today we can see the progress of each cataloger in the UI of syft.
I think it will be also useful to know how much each cataloger took, can be also in the logs in debug mode.
**Why is this needed**:
sometimes scans can take time, and i would like to know which cataloger took time so I can make changes in the configuration.
**Additional context**:

tomersein:
can be closed @tgerla

kzantow:
Thanks @tomersein -- yeah, this was implemented some time ago!

PR #3105, opened 2024-08-07
add log time of task

tomersein:
knowing how much a task took in syft can be useful
in this PR we add a log which specifies the time a task took.
closes - https://github.com/anchore/syft/issues/2723

### DIFF
--- a/internal/task/executor.go
+++ b/internal/task/executor.go
@@ -72,6 +72,6 @@ func runTaskSafely(ctx context.Context, t Task, resolver file.Resolver, s sbomsy
 
 	start := time.Now()
 	res := t.Execute(ctx, resolver, s)
-	log.Tracef("task name: %s time to execute: %v", t.Name(), time.Since(start))
+	log.WithFields("task", t.Name(), "elapsed", time.Since(start))
 	return res
 }
--- a/internal/task/executor.go
+++ b/internal/task/executor.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"runtime/debug"
 	"sync"
+	"time"
 
+	"github.com/anchore/syft/internal/log"
 	"github.com/hashicorp/go-multierror"
 
 	"github.com/anchore/syft/internal/sbomsync"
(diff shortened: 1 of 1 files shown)


internal/task/executor.go
```diff
@@ -68,5 +70,8 @@ func runTaskSafely(ctx context.Context, t Task, resolver file.Resolver, s sbomsy
 		}
 	}()
 
-	return t.Execute(ctx, resolver, s)
+	start := time.Now()
+	res := t.Execute(ctx, resolver, s)
+	log.Debugf("task name: %s time to execute: %v", t.Name(), time.Since(start))
```
wagoodman, line 75: this is probably better served as a `trace` log level detail

tomersein: fixed

internal/task/executor.go
```diff
@@ -68,5 +70,8 @@ func runTaskSafely(ctx context.Context, t Task, resolver file.Resolver, s sbomsy
 		}
 	}()
 
-	return t.Execute(ctx, resolver, s)
+	start := time.Now()
+	res := t.Execute(ctx, resolver, s)
+	log.Tracef("task name: %s time to execute: %v", t.Name(), time.Since(start))
```
wagoodman, line 75: I should have mentioned this on the first review we should lean into separating out data fields from the message itself:
```go
log.WithFields("task", t.Name(), "elapsed", time.Since(start))
```

tomersein: fixed

internal/task/executor.go
```diff
@@ -68,5 +70,8 @@ func runTaskSafely(ctx context.Context, t Task, resolver file.Resolver, s sbomsy
 		}
 	}()
 
-	return t.Execute(ctx, resolver, s)
+	start := time.Now()
+	res := t.Execute(ctx, resolver, s)
+	log.WithFields("task", t.Name(), "elapsed", time.Since(start))
```
kzantow, line 75: This needs to have a log directive after the `.WithFields()` call, e.g.:

```
log.WithFields("task", t.Name(), "elapsed", time.Since(start)).Info("task completed")
```

tomersein: fixed

kzantow (approved):

Issue #2977, opened 2024-06-19
Add support for java "kar" files


fernandosci:
Add support for the kar extension.

The [kar](https://karaf.apache.org/manual/latest/kar) file extension is defined and used by [Apache Karaf](https://karaf.apache.org) and karaf is usually used in JAVA OSGi-based projects.

since the kar file is essentially a jar file, the impact should be similar to #1701.

wagoodman:
Implemented by #3113

PR #3113, opened 2024-08-09
support .kar files

tomersein:
support .kar files
closes - https://github.com/anchore/syft/issues/2977


wagoodman (approved):
Thanks for the contribution! (and TIL about `kar` files 😄 )

PR #3117, opened 2024-08-12
chore: fix failing python relationship test

kzantow:
[readchar 4.2.0](https://pypi.org/project/readchar/) was released Aug 11, and seems to be picked up by the python relationship test when we have a 4.1.0 in site packages; this PR updates the test to account for the new version.


syft/pkg/cataloger/python/test-fixtures/image-multi-site-package/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN python3.9 -m pip install requests==2.25.0 certifi==2020.12.5 chardet==3.0.4
 RUN python3.8 -m pip install click==8.0.2 beautifulsoup4==4.9.2 soupsieve==2.2.0 requests==2.25.0
 RUN python3.8 -m pip install runs==1.2.2 xmod==1.8.1 # partial dependencies for inquirer in project2 (which is a red herring)
 RUN python3.8 -m pip install requests==2.25.0 certifi==2020.12.5 chardet==3.0.4 idna==2.10 urllib3==1.26.18 # total dependencies for requests
-RUN python3.8 -m pip install readchar==4.1.0
+RUN python3.8 -m pip install readchar==4.2.0
```
kzantow, line 20: Is there some way to pin this to a particular version when it's installed by dependency below?

wagoodman (approved):
This gets us unblocked, but I don't see a good path forward that forces these versions to be pinned beyond what is already being done.

I think we should merge this as is to unblock, but should continue to think about how to make this fixture better (better pinning).

PR #3118, opened 2024-08-13
RPM-invalid-cache

barnuri:
its look like the rpms in cache have packages and make the UT pass
while if there is no cache, the download RPM is packages


spiffcs:
hey @barnuri I don't think we want to comment out the workflow here for the cache - are you trying to bust the cache and then see if the UT fails? What's the expected outcome for this PR? 

I think I'm sitting on the side of closing it, but can reopen if there is a related issue or some larger problem with how our unit tests are working?

I've freshly cloned syft on my machine, busted the cache, downloaded all new RPM and the unit tests pass.

PR #3119, opened 2024-08-13
fix: add nil check to cyclonedx toBomProperties

lucasrod16:
Fixes https://github.com/anchore/grype/issues/1759

This PR ensures that `toBomProperties` returns `nil` for images that do not have labels. When an image does not have any labels, `EncodeProperties` returns a `nil` slice and `toBomProperties` then returns a pointer to that `nil` slice (`&props`), resulting in a non-compliant `null` JSON value.

SBOMs generated by syft do not appear to be affected since this code is not in the call path, so it only affects grype as a caller of this code.

### DIFF
--- a/syft/format/common/cyclonedxhelpers/to_format_model.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model.go
@@ -211,6 +211,12 @@ func toBomProperties(srcMetadata source.Description) *[]cyclonedx.Property {
 	metadata, ok := srcMetadata.Metadata.(source.ImageMetadata)
 	if ok {
 		props := helpers.EncodeProperties(metadata.Labels, "syft:image:labels")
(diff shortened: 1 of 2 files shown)


kzantow:
Thanks @lucasrod16 -- this looks like a good fix. Would you be able to add a very simple test that validates we are setting the actual nil value in the `toBomProperties` function, so we don't regress this?

lucasrod16:
> Thanks @lucasrod16 -- this looks like a good fix. Would you be able to add a very simple test that validates we are setting the actual nil value in the `toBomProperties` function, so we don't regress this?

@kzantow yes I have added a unit test for `toBomProperties`

kzantow (approved):
👍 Thanks for the contribution, @lucasrod16 !

PR #3132, opened 2024-08-17
Add configuration to disable capturing any file info in SBOM

### LINKED ISSUE #2989: Setting `file.metadata.selection` to `none` still results in files in the SBOM
**What happened**:
When I use "none" I still get "files" entry in the final json.
If I use "none" remove the "files" entry.
use this config.yaml:
```
file:

   metadata: 
      # select which files should be captured by the file-metadata cataloger and included in the SBOM. 
      # Options include:
      #  - "all": capture all files from the search space
      #  - "owned-by-package": capture only files owned by packages
      #  - "none", "": do not capture any files
      # SYFT_FILE_METADATA_SELECTION env var
      selection: "none"
```
scan an image\directory
did a little check and the issue is this function:
`func toFile(s sbom.SBOM) []model.File`
I think that in case of none it shouldn't enter this function or use skip (if all variables like metadata, digest, etc.) are empty.

**Environment**:
- Output of `syft version`: 1.8.0

tomersein:
Adds `files.enable` configuration such that, when set to `false`, no entries are populated in the  `.files` section of the SBOM.

Closes #2989

### DIFF
--- a/syft/sbom/sbom.go
+++ b/syft/sbom/sbom.go
@@ -66,9 +66,11 @@ func (s SBOM) AllCoordinates() []file.Coordinates {
 	for coordinates := range s.Artifacts.Unknowns {
 		set.Add(coordinates)
 	}
-	for _, relationship := range s.Relationships {
-		for _, coordinates := range extractCoordinates(relationship) {
-			set.Add(coordinates)
+	if len(set.ToSlice()) > 0 {
+		for _, relationship := range s.Relationships {
+			for _, coordinates := range extractCoordinates(relationship) {
+				set.Add(coordinates)
+			}
 		}
 	}
 	return set.ToSlice()
--- a/syft/sbom/sbom.go
+++ b/syft/sbom/sbom.go
@@ -66,7 +66,7 @@ func (s SBOM) AllCoordinates() []file.Coordinates {
 	for coordinates := range s.Artifacts.Unknowns {
 		set.Add(coordinates)
 	}
-	if len(set.ToSlice()) > 0 {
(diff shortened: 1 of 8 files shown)


wagoodman:
force push was to rebase / solve merge conflicts

syft/sbom/sbom.go
```diff
@@ -66,9 +66,11 @@ func (s SBOM) AllCoordinates() []file.Coordinates {
 	for coordinates := range s.Artifacts.Unknowns {
 		set.Add(coordinates)
 	}
-	for _, relationship := range s.Relationships {
-		for _, coordinates := range extractCoordinates(relationship) {
-			set.Add(coordinates)
+	if set.Size() > 0 {
```
wagoodman, line 69: the downside with this is that we are adding relationships already from pkg-to-file or file-to-file and here we're silently dropping only the nodes, but not changing the relationships. I think this part needs a little more thought -- that is, if there is an edge in the SBOM that describes the file directly, then we either need to include that file or drop the relationship.

tomersein: hi! sounds right.. however I am not sure how to solve it.
please let me know how to proceed further so this PR can be merged in the future :)

wagoodman:
I think this has been effectively replaced with #3505 (based on https://github.com/anchore/syft/issues/2989#issuecomment-2573445736), @tomersein is there any behavior in this branch you still want merged?

wagoodman:
@tomersein I'm going to close this for now since #3505 has replaced this and has been released, if this is in error please feel free to re-open and we can discuss

PR #3146, opened 2024-08-21
Detect curl binaries

KrysGor:
Detect curl binaries via the regex 'curl/[0-9]+\.[0-9]+\.[0-9]+)'.


kzantow (approved):

kzantow:
Thanks for the contribution, @KrysGor!

PR #3156, opened 2024-08-22
feat: Add support for new Copyrights and schema updates

dor-hayun:
### JSON Schema
- **Added**: New JSON schema version `16.0.16` with support for the new `Copyrights`.
- **Modified**: Updated the `JSONSchemaVersion` parameter to use the new schema.

### Package Structs
- **Added**: New `Copyrights` field to the `Package` and `PackageBasicData` structs, similar to the existing `Licenses` field.
- **Added**: New `Copyright` struct.
- **Implemented**: Sorting methods for the `Copyright` struct.

### toPackages Function
- **Changed**: Updated the `PackageCopyrightText` to use `helpers.GetCopyrights(p.Copyrights)`, which formats the copyright text and returns a string. Example output: "Copyright 2014-2014 Matt Zabriskie & Collaborators".

### toSyftPackage Function
- **Added**: `Copyrights` assignment to the `toSyftPackage` function.


kzantow:
Generally, I think this looks pretty good @dor-hayun. A couple comments:

If we're going to add this to the Syft data model and SPDX, we should also add this to CycloneDX, since it appears to [support copyright of components also](https://cyclonedx.org/docs/1.6/json/#components_items_copyright).

That said, I don't see where this is ever populated -- are there some catalogers we could add this to with the initial implementation?

dor-hayun:
@kzantow I'll add this to CycloneDX as well. The goal is to enable the addition of copyright text either as part of the existing catalogers or by injecting this data directly from other sources and formatting it. For instance, I can iterate over the `sbom.Artifacts` to manually add the copyright, and then use the formatter to display it in SPDX format too.

dor-hayun:
@kzantow added also in CycloneDX

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/anchore/syft
+module github.com/dor-hayun/syft
```
kzantow, line 1: I don't think this should get changed :)

internal/cmptest/common_options.go
```diff
@@ -61,11 +66,31 @@ func CommonOptions(licenseCmp LicenseComparer, locationCmp LocationComparer) []c
 				return true
 			},
 		),
+		cmp.Comparer(
+			func(x, y pkg.CopyrightsSet) bool {
```
kzantow, line 70: This should be split out to a separate function, following the same pattern as the other comparers

syft/format/internal/cyclonedxutil/helpers/licenses.go
```diff
@@ -5,11 +5,15 @@ import (
 	"strings"
 
 	"github.com/CycloneDX/cyclonedx-go"
-
 	"github.com/anchore/syft/internal/spdxlicense"
 	"github.com/anchore/syft/syft/pkg"
 )
 
+const (
+	noAssertion     = "NOASSERTION"
```
kzantow, line 13: CycloneDX should not use NOASSERTION, this is SPDX specific. CycloneDX should just leave this field empty if no copyright is found

syft/pkg/cataloger/internal/pkgtest/test_generic_parser.go
```diff
@@ -360,12 +362,33 @@ func AssertPackagesEqual(t *testing.T, a, b pkg.Package) {
 				return true
 			},
 		),
+		cmp.Comparer(
+			func(x, y pkg.CopyrightsSet) bool {
```
kzantow, line 366: This should follow the existing patterns with a separate function

syft/pkg/license_set.go
```diff
@@ -1,13 +1,13 @@
+//nolint:dupl
```
kzantow, line 1: why is this considered duplicated code? we should probably reuse existing functionality if needed

syft/pkg/license_set.go
```diff
@@ -1,13 +1,13 @@
...
 
 import (
 	"fmt"
 	"sort"
 
-	"github.com/mitchellh/hashstructure/v2"
-
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
+	"github.com/mitchellh/hashstructure/v2"
```
kzantow, line 10: Run `make lint-fix` to run the formatters; this will fix some of the lint issues

kzantow (commented):
I had a quick look and left a bit of feedback here that would need to be addressed.

A couple other things:
* SPDX and CycloneDX should both encode and decode copyright information -- Syft should be able to read its own files
* I do not see any catalogers which are surfacing copyrights, I don't think we would want to add to the data model without using it at all

syft/pkg/license_set.go
dor-hayun: done

syft/format/internal/cyclonedxutil/helpers/licenses.go
dor-hayun: גםמק

go.mod
dor-hayun: reverted thanks

kzantow:
Hey @dor-hayun -- sorry for the delay on this; we [discussed this on our livestream](https://anchorecommunity.discourse.group/t/february-13th-open-source-gardening-live-stream/333/1) and we think to get this moved forward, a reasonable thing to do is add functionality to [the Debian cataloger](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/debian/package.go#L32) to populate this, so we have at least one spot that this is getting filled out. That would be the next step to get this moved forward, but we definitely think it's a great idea and can help make these updates, too. Thanks very much for your contributions!

wagoodman:
Here's a sample copyright file in `debian:latest`:

<details>
  <summary>/usr/share/doc/debianutils/copyright</summary>

```
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
Comment: This is the Debian GNU/Linux package debianutils.
 It is an original Debian package.  Programs in it were maintained by
 Guy Maor <maor@debian.org>, and are now maintained by Clint Adams
 <schizo@debian.org>.

Files: *
Copyright: (see individual files)
License: GPL-2+

Files: installkernel
Copyright:
 Copyright (C) 1995 - 1998, Ian A. Murdock <imurdock@debian.org>
 Copyright (C) 1998, 1999, Guy Maor
 Copyright (C) 2002, Matthew Wilcox
 Copyright (C) 2002, 2004, 2005, 2007, 2009  Clint Adams
 Copyright (C) 2009  Manoj Srivasta
License: GPL-2+

Files: ischroot.c
Copyright:
 Copyright (C) 2011 Aurelien Jarno <aurel32@debian.org>
 Copyright (C) 2015 Andreas Henriksson <andreas@fatal.se>
License: GPL-2+

Files: run-parts.*
Copyright:
 Copyright (C) 1994 Ian Jackson.
 Copyright (C) 1996 Jeff Noxon <jeff@router.patch.net>
 Copyright (C) 1996-1999 Guy Maor <maor@debian.org>
 Copyright (C) 2002-2020 Clint Adams
License: GPL-2+

Files: update-shells
Copyright: 2021 Helmut Grohne <helmut@subdivi.de>
License: GPL-2+

License: GPL-2+
 All the package's programs except 'savelog', and 'which' may be
 redistributed under the terms of the GNU GPL, Version 2 or later,
 found on Debian systems in the file /usr/share/common-licenses/GPL-2.

Files: which
Copyright: public-domain
License: public-domain
 which is in the public domain.

Files: savelog
Copyright:
 Copyright (C)  1988 Landon Curt Noll & Ronald S. Karr
 Copyright (C)  1992 Ronald S. Karr
 Copyleft (GNU) 1988 Landon Curt Noll & Ronald S. Karr
License: SMAIL-GPL
                     SMAIL GENERAL PUBLIC LICENSE
                       (Clarified 11 Feb 1988)
 .
 Everyone is permitted to copy and distribute verbatim copies
 of this license, but changing it is not allowed.  You can also
 use this wording to make the terms for other programs.
 .
   The license agreements of most software companies keep you at the
 mercy of those companies.  By contrast, our general public license is
 intended to give everyone the right to share SMAIL.  To make sure that
 you get the rights we want you to have, we need to make restrictions
 that forbid anyone to deny you these rights or to ask you to surrender
 the rights.  Hence this license agreement.
 .
...
```
</details>

Note that there are:
- multiple copyrights applied to sub-selections of the package
- copyright(s) that are paired with select license(s)

I don't think we'd be able to describe (accurately) the relationship of debian copyrights found for a package; there are probably a few changes needed:
- include an optional subject(s) that describe what it applies to. If empty this describes the full package. If not empty it could describe sub-selections / select files from the package.
- There should probably be an object that pairs up copyright and licenses together and have a set of those on a package. If we replace license with this (which is ideal) it would be a breaking change (which is not ideal). This pairing would probably be something like "intellectualPropertyRights" or more simply "rights" (naming is hard, needs some thought).

This last point is enough to hold off trying to merge this as is. I'm going to close this for now since there would be considerable changes needed to account for this.

Issue #3172, opened 2024-08-29
CycloneDX convertion into Syft improperly handles SPDX licenses


NyanKiyoshi:
When converting a CycloneDX JSON BOM into the Syft format, SPDX licenses are dropped.

The [CycloneDX specifications](https://cyclonedx.org/docs/1.6/json/#tab-pane_components_items_licenses_oneOf_i0) define that either (one of) the following values should be set under `components[].licenses`:
- CycloneDX format: `[{"license": {"id": "...", "name": "..."}}, ...]`
- **Or** (exclusive or), SPDX format, thus only `[{"expression": "..."}, ...]`

But Syft expects to have the following value which is invalid in the specs: `{"expression": "...", "license": {...}}`.

This is caused by the lines 57 to 59 at: https://github.com/anchore/syft/blob/11d77b4a9489604c5148bba7a6ca84bd31786691/syft/format/internal/cyclonedxutil/helpers/licenses.go#L50-L73 

## What you expected to happen:

When an SPDX license is provided inside a CycloneDX component, Syft should not drop it.

## Potential Solution

We could drop the `if l.License == nil { continue }` and change the `case` conditions to handle the `nil` case properly:

```go
func decodeLicenses(c *cyclonedx.Component) []pkg.License {
	// [...]

	for _, l := range *c.Licenses {
		switch {
		case l.License != nil && l.License.ID != "":
			licenses = append(licenses, pkg.NewLicenseFromURLs(l.License.ID, l.License.URL))
		case l.License != nil && l.License.Name != "":
			licenses = append(licenses, pkg.NewLicenseFromURLs(l.License.Name, l.License.URL))
		case l.Expression != "":
			licenses = append(licenses, pkg.NewLicense(l.Expression))
		default:
		}
	}

	return licenses
}
```

Diffs:

<details>

```diff
--- a/licenses.go
+++ b/licenses.b.go
@@ -54,17 +54,13 @@ func decodeLicenses(c *cyclonedx.Component) []pkg.License {
        }
 
        for _, l := range *c.Licenses {
-               if l.License == nil {
-                       continue
-               }
-               // these fields are mutually exclusive in the spec
                switch {
-               case l.License.ID != "":
+               case l.License != nil && l.License.ID != "":
                        licenses = append(licenses, pkg.NewLicenseFromURLs(l.License.ID, l.License.URL))
-               case l.License.Name != "":
+               case l.License != nil && l.License.Name != "":
[...]

PR #3175, opened 2024-08-29
Fix improper decoding of SPDX license expressions in the CycloneDX format

NyanKiyoshi:
This fixes the issue reported at https://github.com/anchore/syft/issues/3172, where Syft would drop SPDX licenses due a logic error in the decoder.

CycloneDX specifications require `components[].licenses[].license` to be `nil` when `components[].licenses[].expression` (SPDX) is non-`nil`.


kzantow (approved):

kzantow:
Thanks for the contribution, @NyanKiyoshi !

PR #3179, opened 2024-08-30
Expose RPM signature information (for RPM DB and RPM archives)

ralphbean:
This helps with more confident identification of an rpm.

In theory, two rpms can be built that have the same purl string, and otherwise look identical in syft's output, but the PGP information would distinguish them as signed either by different keys, or signed at different times.

In practice, this usually makes no difference since rpms tend to have unique name/version/release strings. This gives increased confidence about the identity of the rpm found in the db.


ralphbean:
I still need to get some signature information into the tests before this is ready for review.

ralphbean:
(I spent some time trying to get the test suite working locally on `main` branch, but have so far failed. Just FYI, I'm not totally ignoring this.)

wagoodman:
no problem @ralphbean -- shout out if there is something we can do to help!

wagoodman:
The upstream lib doesn't seem to be working when the backing store is sqlite (where it is working when it's bdb)

Screenshot 2025-05-08 at 4 24 53 PM

Technically that doesn't affect the functionality, but it does make generating a test case a little harder. I might poke upstream to see what's going on.

wagoodman:
We don't have to wait for upstream to merge this -- we'll use the fork for now and when merged, switch to using upstream again.

wagoodman (approved):

PR #3182, opened 2024-09-01
feat: --enrich flag for data enrichment feature enablement

kzantow:
This PR introduces an `--enrich` flag, to allow Syft to enable data enrichment operations for all catalogers with a single flag. These operations often involve network requests to services like Maven Central, NPM, or GitHub. In the future there are likely to be an increasing number of remote network services that Syft is able to utilize, and this would be the first step to enabling that easily and automatically opting-in to future enrichment operations for all users who want the best quality SBOM they can get.

The easiest way to enable everything, is to use `--enrich all`, and Syft will enable all enrichment operations, which currently include Golang local and remote searches, Maven local and remote searches, and NPM lookups. This PR also supports enabling and disabling individual cataloger data data enhancement if there are any reasons to do so.


cmd/syft/internal/options/catalog.go
```diff
@@ -243,3 +247,13 @@ func (cfg *Catalog) PostLoad() error {
 
 	return nil
 }
+
+func multilevelOption(options ...*bool) bool {
```
wagoodman, line 251: I thought we had this as a util already?

wagoodman (dismissed):
(pending the fangs update)

kzantow:
A few questions came up during discussions, so I'm not going to merge this as-is; will get [some feedback first](https://anchorecommunity.discourse.group/t/enabling-syft-network-features-from-the-cli/117)

cmd/syft/internal/options/catalog.go
```diff
@@ -132,7 +133,7 @@ func (cfg Catalog) ToPackagesConfig() pkgcataloging.Config {
 		Golang: golang.DefaultCatalogerConfig().
 			WithSearchLocalModCacheLicenses(cfg.Golang.SearchLocalModCacheLicenses).
 			WithLocalModCacheDir(cfg.Golang.LocalModCacheDir).
-			WithSearchRemoteLicenses(cfg.Golang.SearchRemoteLicenses).
+			WithSearchRemoteLicenses(*multiLevelOption(false, enrichmentEnabled(cfg.Enrich, "golang", "go"), cfg.Golang.SearchRemoteLicenses)).
```
kzantow, line 136: Should these match cataloger group names somehow?

wagoodman: we could use constants up in `internal/task/package_tasks.go`?

wagoodman (approved):

wagoodman (approved):

PR #3199, opened 2024-09-05
fix: packages with empty name

idefixcert:
the problem we faced:
Some packages did not have a name.
The pkg has already the function valid, but that is not everywhere used. This commit adds the validation only for catalogers we have seen the error.


willmurphyscode:
Hi @idefixcert - I believe the issue you were facing here is fixed by #3257, so I'm closing this PR. If I've missed something, please let me know, and thanks!

PR #3190, opened 2024-09-03
Slim down docker cache size

wagoodman:
Fixes #3181 

We use docker images and other ephemeral data as a basis for test fixtures in unit/integration/cli tests...

<details>
  <summary>...but we have over `5 GB` in cache!</summary>

_There is what is listed below + a few-hundred MB centos 7 image that was removed._

```
$  find . | grep cache | grep tar | xargs du -ch | sort -hr

4.9G	total
832M	./cmd/syft/internal/test/integration/test-fixtures/cache/stereoscope-fixture-image-test-java-purls-a3aba7fe8ca58ea9586ae8752d939cc72a8a2d6fc965b8eed99ace498c867f94.tar
704M	./syft/pkg/cataloger/binary/test-fixtures/cache/stereoscope-fixture-elf-test-fixtures-72a1400c46a4f03cc0bedce1d8112feecfaddceff6f6bb665edb1a652b4757fc.tar
704M	./cmd/syft/internal/test/integration/test-fixtures/cache/stereoscope-fixture-elf-test-fixtures-72a1400c46a4f03cc0bedce1d8112feecfaddceff6f6bb665edb1a652b4757fc.tar
576M	./syft/pkg/cataloger/python/test-fixtures/cache/stereoscope-fixture-image-multi-site-package-fa8e416407542939f6a73562f19526d99a27f21d3daa5ac5c723063173cd9a1e.tar
576M	./cmd/syft/internal/test/integration/test-fixtures/cache/stereoscope-fixture-image-java-no-main-package-78384bbaa3076afc5904db77a8650aa4183c5900e1bb7cd392a0d2b47c27a438.tar
336M	./cmd/syft/internal/test/integration/test-fixtures/cache/stereoscope-fixture-image-golang-compiler-b855ba45f2a2de7d0b38fa66a62341448fdee022775ae0f05496fb90ac8ac510.tar
242M	./cmd/syft/internal/test/integration/test-fixtures/cache/stereoscope-fixture-image-large-apk-data-f05f8d9a0458190967d606a8c554d00630af9ea158aa1a1cbbcbbf25e03c2c0c.tar
192M	./syft/pkg/cataloger/redhat/test-fixtures/cache/stereoscope-fixture-image-minimal-f3f6a85ded073b94e1903b04b544746767ad151046f69e8598c0acef5846ae06.tar
[...]

PR #3200, opened 2024-09-05
Publish central test-fixture cache

wagoodman:
TODO:
- [ ] add workflow that will update the oras cache nightly from main


wagoodman:
I'm finding this difficult to coordinate separate of https://github.com/anchore/syft/pull/3190 , so I'm closing this and merging with that work

PR #3220, opened 2024-09-11
chore: make ci-check.sh an executable file

okuryu:
It is likely that this script will need to be an executable file in relation to #3218.


kzantow (approved):
Thanks! :)

PR #3224, opened 2024-09-12
Add Rakudo Star binaries cataloger

gdams:
I seem to be missing something here as I can't get it to detect the binary


wagoodman:
Taking a look at the binary you selected and the regex you have I also don't see the version embedded in the binary:
```
$strings classifiers/bin/rakudo-star/2024.08/linux-amd64/rakudo | grep -i rakudo     
ERROR: RAKUDO_HOME is invalid: %s
--rakudo-home=
RAKUDO_HOME

$ strings classifiers/bin/rakudo-star/2024.08/linux-amd64/rakudo | grep -i 2024
<no results>
```

Maybe this version is embedded not as a string? If that's the case then detecting the version might not be possible with this approach.

LaurentGoderre:
I have been looking at rakudo and it's a lot more complex than most of the binaries I have done. The binaries do not seem to include any version info and instead it is embed in the moarvm. I was able to detect moarvm though

wagoodman:
@gdams @LaurentGoderre It looks like the path forward is cataloging the moarvm binary as mentioned above -- is any help wanted with this? Or does that not quite fit what you're trying to accomplish @gdams ?

gdams:
sorry - yes I'll likely need to drop this one for now as I've been dragged into other things

PR #3228, opened 2024-09-12
Track supporting DPKG evidence

### LINKED ISSUE #3191: Identify `bash.preinst`
Syft does not appear to identify `bash.preinst`, which is present on many of the official Docker images.

**Why is this needed**:
Syft should be able to identify common executables at least as expected/known rather than unknown.

**Additional context**:
In the official docker images, as of this writing, there are 33 instances of this at `/var/lib/dpkg/info/bash.preinst`, which do not appear to be associated to any of the package-manager installed packages. From some basic web search, it looks like this may be a one-time script to associate `/bin/sh` with `/bin/bash` in some manner, but more investigation is needed where it comes from and why it doesn't get removed for proper classification.

wagoodman:
This elevates supporting locations on each package to be expressed with `evident-by` relationships --today we only do this with primary locations. This additionally adds a new 'kind' param to the relationship data, showing if this is 'primary' or 'supporting' evidence.

This helps with #3191 since now these files would be accounted for by a package (without claiming they are owned by it), thus, it will no longer show up as an unknown when #2998 lands.

Closes #3191


syft/pkg/cataloger/ocaml/parse_opam_test.go
wagoodman, line 1: ?? fun, it seems like we have some linter tooling that is go-version specific (or something)

syft/pkg/cataloger/debian/cataloger_test.go
```diff
@@ -25,15 +25,16 @@ func TestDpkgCataloger(t *testing.T) {
 					Version: "1.1.8-3.6",
 					FoundBy: "dpkg-db-cataloger",
 					Licenses: pkg.NewLicenseSet(
-						pkg.NewLicenseFromLocations("GPL-1", file.NewVirtualLocation("/usr/share/doc/libpam-runtime/copyright", "/usr/share/doc/libpam-runtime/copyright")),
-						pkg.NewLicenseFromLocations("GPL-2", file.NewVirtualLocation("/usr/share/doc/libpam-runtime/copyright", "/usr/share/doc/libpam-runtime/copyright")),
-						pkg.NewLicenseFromLocations("LGPL-2.1", file.NewVirtualLocation("/usr/share/doc/libpam-runtime/copyright", "/usr/share/doc/libpam-runtime/copyright")),
+						pkg.NewLicenseFromLocations("GPL-1", file.NewLocation("/usr/share/doc/libpam-runtime/copyright")),
```
willmurphyscode, line 28: Can you help me understand why this stopped being a virtual location and became a regular location? Is this change visible to users in the eventual SBOM?

wagoodman: This was cleanup -- since I was in this file and making new Locations, I noticed that all params to the original call were the same, which is the same as using the NewLocation constructor. So I made the remaining parts of the file that were using this function needlessly as consistent as the new code I was adding.

willmurphyscode (approved):

syft/pkg/cataloger/debian/parse_dpkg_db.go
```diff
@@ -34,12 +36,41 @@ func parseDpkgDB(_ context.Context, resolver file.Resolver, env *generic.Environ
...
+	}
+	if strings.TrimSpace(name) == "" {
+		return nil
+	}
+
+	// for typical debian-base distributions, the installed package info is at /var/lib/dpkg/status
+	// and the md5sum information is under /var/lib/dpkg/info/; however, for distroless the installed
+	// package info is across multiple files under /var/lib/dpkg/status.d/ and the md5sums are contained in
+	// the same directory
+	searchPath := filepath.Dir(dbLocation.RealPath)
```
kzantow, line 58: This should be `path`, since it's using the resolver not OS functions

kzantow (changes requested):

syft/pkg/cataloger/debian/parse_dpkg_db.go
wagoodman: good catch!

kzantow (approved):

PR #3266, opened 2024-09-23
Allow for concurrent cataloger parser calls

wagoodman:
Note: uses unmerged changes from go-sync (using go.work locally)


wagoodman:
Closing in favor of #3636

PR #2918, opened 2024-05-31
fix: only skip tmpfs mounts for some paths

willmurphyscode:
See https://github.com/anchore/syft/issues/2894, https://github.com/anchore/syft/issues/2847, https://github.com/anchore/grype/issues/1822

# TODO:

- [ ] Test on real systems with different mount configurations
- [ ] Actually add unit tests

PR #3269, opened 2024-09-24
Improve subpath to mount matching

cdupuis:
👋🏽 we are seeing issues when upgrading the `syft-sbom-indexer` to the latest version of Syft running on latest GitHub Action Runner images (ubuntu-22.04). Effectively the recently introduced mount path checking prevents the mounted filesystem to be analysed. We see the following log message in the GHA logs when enabling debug logging:

```
DEBUG ignoring path based on mountpoint filesystem type condition=/run fs=tmpfs mountpoint=/proc/latency_stats path=/run/src/core/sbom
```

We tracked this down to https://github.com/anchore/syft/pull/2918. 

This PR adds a new test mimicking the file mounts as seen on the GitHub Action Runner image for ubuntu-22.04 and changes the code to let the requested path `/run/src/core/sbom` not get blocked. 

Please take a look at this change and let us know what you think? Thank you.

### DIFF
--- a/syft/internal/fileresolver/path_skipper.go
+++ b/syft/internal/fileresolver/path_skipper.go
@@ -93,12 +93,13 @@ func (ps pathSkipper) pathIndexVisitor(_ string, givenPath string, _ os.FileInfo
 	for _, mi := range ps.mounts {
 		conditionalPaths, ignorable := ps.ignorableMountTypes[mi.FSType]
 
-		if len(conditionalPaths) == 0 {
-			// Rule 1: ignore any path within a mount point that is of the given filesystem type unconditionally
-			if !containsPath(givenPath, mi.Mountpoint) {
-				continue
-			}
+		// Rule 0: Make sure the given path is within the mount point; if not let the scan continue
+		if !containsPath(givenPath, mi.Mountpoint) {
+			continue
+		}
 
+		// Rule 1: ignore any path within a mount point that is of the given filesystem type unconditionally
+		if len(conditionalPaths) == 0 {
 			if !ignorable {
 				// we've matched on the most specific path at this point, which means we should stop searching
 				// mount points for this path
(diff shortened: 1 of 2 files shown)


LaurentGoderre:
This is showing the bug as well as the fix in action: https://github.com/LaurentGoderre/build-debug/actions/runs/11016581191

Source: https://github.com/LaurentGoderre/build-debug/blob/main/.github/workflows/build.yml#L18-L26

LaurentGoderre (approved):

kzantow:
I think `make lint-fix` should correct the static analysis failure

cdupuis:
> I think `make lint-fix` should correct the static analysis failure

sorry @kzantow, I pushed an update. Hopefully this is fixed now.

syft/internal/fileresolver/path_skipper.go
```diff
@@ -93,12 +93,12 @@ func (ps pathSkipper) pathIndexVisitor(_ string, givenPath string, _ os.FileInfo
 	for _, mi := range ps.mounts {
 		conditionalPaths, ignorable := ps.ignorableMountTypes[mi.FSType]
 
-		if len(conditionalPaths) == 0 {
-			// Rule 1: ignore any path within a mount point that is of the given filesystem type unconditionally
-			if !containsPath(givenPath, mi.Mountpoint) {
-				continue
-			}
+		if !containsPath(givenPath, mi.Mountpoint) {
```
kzantow, line 96: A lot of the path handling is not entirely obvious. Could we add a comment indicating what this is doing? If I'm reading this correctly, this is something like:
```
// Rule 0: do not exclude mount points where the scan root is a subdirectory
```
... is this accurate?

cdupuis: I've added a comment to clarify the behaviour.

kzantow (approved):
Sorry it took me a bit to understand some of the nuance in this behavior. This definitely looks like a correct change I see now, thanks much @cdupuis!

LaurentGoderre:
@kzantow thank you!!

PR #3273, opened 2024-09-24
fix: add dependency relationships for Java archives and Maven sources

kzantow:
This PR makes 2 changes: 
* refactor the Maven resolver to an internal package in order to better separate callable methods from internal recursive calls
* add dependency information for Maven projects where found and dependency information for found nested Java archives


kzantow:
This was superseded by #3363

PR #3290, opened 2024-10-01
fix: don't use licensecheck's builtin scanner

govrin:
The built-in scanner is never cleaned, so it leaks memory


spiffcs (approved):

spiffcs:
Thanks a million @govrin - this was a great find and will help improve the next release

Issue #1227, opened 2022-09-23
Get licenses for NuGet packages


fg-j:
**What would you like to be added**:
#726 brought initial support for generating SBOMs for NuGet packages 🎉  . One significant gap in the metadata in those SBOMs is license information. It'd be awesome if licenses for NuGet packages were included.

**Why is this needed**:
License information is a key value proposition for compliance-minded users who are building .NET apps with NuGet dependencies.

**Additional context**:

The file that `project.assets.json` that Syft scans for SBOM info doesn't include license information, but I wonder if it's possible to get it from somewhere else.

Here's a snippet of a SPDX SBOM generated with syft for a NuGet package:
```
...
  {
   "SPDXID": "SPDXRef-99c6488b6206ecc1",
   "downloadLocation": "NOASSERTION",
   "externalRefs": [
    {
     "referenceCategory": "SECURITY",
     "referenceLocator": "cpe:2.3:a:Humanizer:Humanizer:2.14.1:*:*:*:*:*:*:*",
     "referenceType": "cpe23Type"
    },
    {
     "referenceCategory": "PACKAGE_MANAGER",
     "referenceLocator": "pkg:dotnet/Humanizer@2.14.1",
     "referenceType": "purl"
    }
   ],
   "filesAnalyzed": false,
   "licenseConcluded": "NONE",
   "licenseDeclared": "NONE",
   "name": "Humanizer",
   "sourceInfo": "acquired package info from dotnet project assets file: NugetBenchmarking.deps.json",
   "versionInfo": "2.14.1"
  },
...
```
the corresponding NuGet package, [Humanizer.Core](https://www.nuget.org/packages/Humanizer.Core) uses the MIT license.

spiffcs:
Thanks so much for the issue @fg-j!

I also added the good first issue label since the write-up you did was so good. 

If we have the bandwidth in the coming week/s we'll try and get this in, but also anyone who comes across this issue consider this fair game to attempt as a contribution. 

Feel free to tag me in a draft PR if you do and I can always help with testing or cleanup!

vanthome:
I have the exact same issue but with Rust Crates. Here is an example output of the Toml files:

```
[package]
edition = "2018"
name = "actix-server"
version = "2.1.1"
[...]

Issue #1562, opened 2023-02-09
Invoke known tools to gather build-time dependency information


kzantow:
**What would you like to be added**:
Add the ability to shell-out to known tools such as `go` and `mvn` in order to capture more accurate build-time dependency information.

**Why is this needed**:
To improve the build-time dependency support in Syft.

**Additional context**:
From a working document:

Creating higher quality SBOMs in Syft at build-time

At build time, static analysis of dependencies implemented today is limited. Improving static analysis metrics can be done by simulating what build systems do. This is subject to drift and additional maintenance to keep up with behaviors of the build systems.

One approach to resolving this issue is to call out to build systems to get that information instead. This introduces additional (optional) dependencies.

**Syft as a build-time SBOM generator tool**

Syft can be seen as a “build-time” SBOM generator tool, and can start thinking about utilizing build tooling. Calling out to build tools can be such as the following, we will use 2 examples.

**Golang**

Instead of reading and trying to parse and resolve the go.mod file, the `go mod graph` command can be used to get a fully resolved dependency tree.

**Java**

Maven has the `mvn dependency:tree` command which shows the fully-resolved dependency graph.

**Considerations**

* When using external tooling, version and parameter information should be captured
* Warnings on quality of what is being used to generate it can be made visible, as well as suggestions on how to obtain better SBOMs (e.g. dependency pinning)

kzantow:
cc: @lumjjb

wagoodman:
This is most likely needed on order to achieve https://github.com/anchore/syft/issues/1674 and https://github.com/anchore/syft/issues/572 in a meaningful way.

This functionality should be opt-in, that is, by default syft should remain a static analysis tool. Executing other commands on the system should still be not allowed by default (again, unless the user opts in).

Considerations:

- should these "external querying capabilities" be encapsulated into their own separate catalogers? For example go-mod-file-cataloger stays as it is today and allow for a new go-tooling-cataloger. In this way opting in would be adding a cataloger (or enabling a flag which would automatically swap out one cataloger for another)... or should we go in the direction of keeping the existing catalogers today that behave differently based on configuration? (one assumption Im making by going down this path is that the impl for the go.mod cataloging today is mutually exclusive to using the build tooling)

- we probably don't want to find duplicate packages by doing a static analysis and tooling query, there should be an obvious mechanism for enforcing mutual exclusivity for existing (static) analysis and tooling analysis.

[...]

Issue #17, opened 2020-05-13
Idea: add additional catalogers via plugin


wagoodman:
If a plugin system is adopted then it would be ideal to be able to support extra catalogers via a user-provided plugin.

wagoodman:
Presently we don't have a strong motivation for completing this work --adding new catalogers via PR seems to be working alright and there is little signal to add the complexity of a plugin system. Additionally it is nice to have the flexibility of changing the cataloger interface with little consequence (which would not be possible with a plugin system).

PR #3329, opened 2024-10-14
Issue #1227 - Get licenses for NuGet packages

HeyeOpenSource:
# Description

- Fixes [Get licenses for NuGet packages #1227](https://github.com/anchore/syft/issues/1227) 

## Type of change

# Checklist:


HeyeOpenSource:
Just for the record:
I oriented myself at the golang cataloger.

The configuration can also be influenced by the following three environment variables:

- NUGET_SEARCH_LOCAL_LICENSES
  - 'true' | 'false'
  - Search for NuGet packages in all known local cache directories.
- NUGET_SEARCH_REMOTE_LICENSES
  - 'true' | 'false'
  - Search for NuGet packages in all **_enabled_** known remote NuGet repositories. (unless the NuGet package profiders setting is overridden)
- NUGET_PACKAGE_PROVIDERS
  - Expects a comma-separated lists of the URLs of remote NuGet repositories to use when searching for remote NuGet packages / licenses.
  - Defaults to all **_enabled_** known remote NuGet repositories if undefined.

HeyeOpenSource:
FYI:
The feature for defining credentials should mostly be applicable to commercial CI/CD environments,
where access restriction is required by appropriate data security requirements.

[Gitea](https://about.gitea.com/) for example allows to create such code- and NuGet package repositories.

HeyeOpenSource:
Switched back to draft due to commit by @wagoodman
[Create single license scanner for all catalogers (#3348)](https://github.com/anchore/syft/commit/e4e985b9b05b9570648399adf292ab7f659b4cef)

HeyeOpenSource:
Since ```make test``` did not throw any errors, I feel free to return this pull request back to ```Ready for review```. 😄

HeyeOpenSource:
The Validations workflow finally throws no more errors at me. 👍 
cf. [Validations #17](https://github.com/HeyeOpenSource/syft/actions/runs/11480604479)

cmd/syft/internal/options/dotnet.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/anchore/clio"
+	"github.com/anchore/syft/syft/pkg/cataloger/dotnet"
+)
+
+type dotnetConfig struct {
+	SearchLocalLicenses  bool   `yaml:"search-local-licenses" json:"search-local-licenses" mapstructure:"search-local-licenses"`
+	SearchRemoteLicenses bool   `yaml:"search-remote-licenses" json:"search-remote-licenses" mapstructure:"search-remote-licenses"`
+	Providers            string `yaml:"package-providers,omitempty" json:"package-providers,omitempty" mapstructure:"package-providers"`
+	ProviderCredentials  string `yaml:"package-provider-credentials,omitempty" json:"package-provider-credentials,omitempty" mapstructure:"package-provider-credentials"`
```
kzantow, line 14: Should this be a structure with fields split out instead of parsing a string? For example, [RegistryCredentials has split out Username and Password](https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/registry.go#L10); we should probably use the same thing for consistency here, and we should use the `secret` type instead of `string` in order to appropriately redact these values everywhere throughout the app. E.g.:

```
type dotnetProviderCredential struct {
	Username secret `yaml:"username" json:"username" mapstructure:"username"`
	Password secret `yaml:"password" json:"password" mapstructure:"password"`
}
type dotnetConfig struct {
	SearchLocalLicenses  bool   `yaml:"search-local-licenses" json:"search-local-licenses" mapstructure:"search-local-licenses"`
	SearchRemoteLicenses bool   `yaml:"search-remote-licenses" json:"search-remote-licenses" mapstructure:"search-remote-licenses"`
	Providers            []string `yaml:"package-providers,omitempty" json:"package-providers,omitempty" mapstructure:"package-providers"`
	ProviderCredentials  []dotnetProviderCredential `yaml:"package-provider-credentials,omitempty" json:"package-provider-credentials,omitempty" mapstructure:"package-provider-credentials"`
}
```

... or maybe the `Providers` makes sense to split into a struct with URL, Username, Password?

Related to the comment about not needing to deal with environment variables later, the `mapstructure` tags in these options is what drives the automatic environment mapping that you will see with `syft config`, however these are a bit funky when it comes to _lists_ such as what I've suggested above. We do, unfortunately end up with some code for that, in a `PostLoad` hook, e.g.: https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/registry.go#L35

syft/pkg/cataloger/dotnet/config.go
```diff
@@ -0,0 +1,255 @@
...
+				Username: credentialParts[0],
+				Password: credentialParts[1],
+			})
+		}
+	}
+	return g
+}
+
+type sourceAPIResource struct {
+	ID      string `json:"@id"`
```
kzantow, line 112: is this JSON-LD format?

cmd/syft/internal/options/catalog.go
```diff
@@ -165,6 +168,11 @@ func (cfg Catalog) ToPackagesConfig() pkgcataloging.Config {
 					WithFromBuildSettings(cfg.Golang.MainModuleVersion.FromBuildSettings).
 					WithFromLDFlags(cfg.Golang.MainModuleVersion.FromLDFlags),
 			),
+		DotNet: dotnet.DefaultCatalogerConfig().
+			WithSearchLocalLicenses(cfg.DotNet.SearchLocalLicenses).
+			WithSearchRemoteLicenses(cfg.DotNet.SearchRemoteLicenses).
```
kzantow, line 173: We definitely will want to tie these options in to the `enrich` functionality, e.g.:
```
			WithSearchLocalLicenses(*multiLevelOption(false, enrichmentEnabled(cfg.Enrich, task.Dotnet), cfg.DotNet.SearchLocalLicenses)).
			WithSearchRemoteLicenses(*multiLevelOption(false, enrichmentEnabled(cfg.Enrich, task.Dotnet), cfg.DotNet.SearchRemoteLicenses)).
```
... what this does is enable these features with `--enrich all` or a more specific `--enrich dotnet` from the Syft CLI

syft/pkg/cataloger/dotnet/config.go
```diff
@@ -0,0 +1,255 @@
...
+	Comment string `json:"comment,omitempty"`
+}
+
+type sourceAPI struct {
+	Version   string              `json:"version"`
+	Resources []sourceAPIResource `json:"resources"`
+	Context   *sourceAPIContext   `json:"@context,omitempty"`
+}
+
+func determineDotnetExecutablePath() (string, error) {
```
kzantow, line 128: Today, Syft does not shell out to any tools (with one very specific exception of `cosign`). And while we have talked about using tools if they are found, we haven't pulled the trigger on this anywhere yet and _very much want to avoid executing tools_, if we can.

Are there some specific files that this configuration is present we could use instead of shelling out to `dotnet`?

syft/pkg/cataloger/dotnet/config.go
```diff
@@ -0,0 +1,255 @@
...
+	defaultProvidersMutex.Lock()
+	defer defaultProvidersMutex.Unlock()
+
+	if defaultProviders != "" {
+		return defaultProviders
+	}
+
+	defaultProviders = defaultProvider
+
+	packageSources := getPackageSourcesFromSDK(false)
```
kzantow, line 216: If the user doesn't have dotnet installed on the host, it looks like this will fail because it doesn't have a default URL to access for remote data -- is there one or more standard URLs that could just be hardcoded here, at least in the case dotnet doesn't exist?

I ask this because there are a lot of cases where people are scanning something with Syft and they won't have dotnet available on the host. For example: if scanning containers in a small cloud runner; it would be great to be able to resolve licenses in all cases for these if the user enables the feature

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,609 @@
...
+			}
+		}
+	}
+
+	return paths
+}
+
+func getLocalNugetFolderResolvers(assetDefinitions []projectAssets) []file.Resolver {
+	nugetPackagePaths := []string{}
+	if injectedCachePath := os.Getenv("TEST_PARSE_DOTNET_DEPS_INJECT_CACHE_LOCATION"); injectedCachePath != "" {
```
kzantow, line 594: We probably should have this functionality entirely configuration-based, if tests need to do something special they could use some configuration. Ideally, tests don't need to do something special.

syft/pkg/cataloger/dotnet/parse_dotnet_deps.go
```diff
@@ -94,10 +101,16 @@ func parseDotnetDeps(_ context.Context, _ file.Resolver, _ *generic.Environment,
 			reader.Location.WithAnnotation(pkg.EvidenceAnnotationKey, pkg.PrimaryEvidenceAnnotation),
 		)
 
-		if dotnetPkg != nil {
```
kzantow, line 97: it looks like this `nil` check got removed, maybe add `if dotnetPkg == nil { continue }`? it's unclear to me if this will cause a panic

syft/pkg/cataloger/dotnet/parse_dotnet_deps_test.go
```diff
@@ -362,5 +399,14 @@ func TestParseDotnetDeps(t *testing.T) {
 		},
 	}
 
-	pkgtest.TestFileParser(t, fixture, parseDotnetDeps, expectedPkgs, expectedRelationships)
+	t.Run(fixture, func(t *testing.T) {
+		os.Setenv("TEST_PARSE_DOTNET_DEPS_INJECT_CACHE_LOCATION", "test-fixtures/NuGetCache")
+		defer os.Setenv("TEST_PARSE_DOTNET_DEPS_INJECT_CACHE_LOCATION", "")
```
kzantow, line 404: Noted before: this should not rely on hidden environment lookups

syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable_test.go
```diff
@@ -48,17 +49,18 @@ func TestParseDotnetPortableExecutable(t *testing.T) {
 			name: "dotnet package with malformed field and extended version",
 			versionResources: map[string]string{
 				"CompanyName":      "Microsoft Corporation",
-				"FileDescription":  "äbFile\xa0\xa1Versi on",
+				"FileDescription":  "äb\x01File\xa0\xa1Versi on",
```
kzantow, line 52: I'm not entirely sure, but these values seem like they were intentional, maybe add a second test with different ones?

kzantow (commented):
Hey @HeyeOpenSource, thanks very much for this PR! Generally speaking, it looks like you've followed the same patterns used elsewhere and this change looks really good. There are a couple specific things I'd like to note:
* instead of adding a bunch of binary files for the tests (including unneeded images and such), could we build these to a test docker image? for example [Java runs some builds for test artifacts](https://github.com/anchore/syft/tree/main/syft/pkg/cataloger/java/test-fixtures/java-builds)
* it looks as though there are a number of things reliant on the dotnet executable; can we get this information without it? if we could, this will definitely help move this PR forward more quickly
* environment variable usage should be cleaned up as noted in more specific comments

syft/pkg/cataloger/dotnet/config.go
```diff
@@ -0,0 +1,255 @@
...
+// - setting the default remote package providers if none are provided
+func DefaultCatalogerConfig() CatalogerConfig {
+	g := CatalogerConfig{
+		SearchLocalLicenses:  false,
+		SearchRemoteLicenses: false,
+		Providers:            []string{},
+		ProviderCredentials:  []nugetProviderCredential{},
+	}
+
+	nuget := os.Getenv("NUGET_SEARCH_LOCAL_LICENSES")
```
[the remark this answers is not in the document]
HeyeOpenSource: I guess you are right.
This will make the related Unit Test obsolete as well.
Adapted in commit fda1188d64ff7e23bfc54ceacbb93e2f6ed3afd4.

cmd/syft/internal/options/dotnet.go
HeyeOpenSource: My initial objective was to make the use of the parameters as easy as possible, hence the orginal decision to use strings, but I do see your point. (Especially the use of the ```secret``` type.) 😉

syft/pkg/cataloger/dotnet/config.go
HeyeOpenSource: I referenced the JSON tags against the output of [https://api.nuget.org/v3/index.json](https://api.nuget.org/v3/index.json)...

```
{
    version: "3.0.0",
    resources: [
    {
        @id: "https://azuresearch-usnc.nuget.org/query",
        @type: "SearchQueryService",
        comment: "Query endpoint of NuGet Search service (primary)"
    },
    [...]
}
```

cmd/syft/internal/options/catalog.go
HeyeOpenSource: Guess I'll have to look into the ```enrich``` functionality first... 🤷‍♂️

syft/pkg/cataloger/dotnet/config.go
HeyeOpenSource: I don't like the necessity either, but so far I have unfortunately not found a way around it. ☹️

syft/pkg/cataloger/dotnet/config.go
HeyeOpenSource: The default provider is defined in line 17. 😉 

```
const (
	defaultProvider = "https://api.nuget.org/v3-flatcontainer/"
)
```

syft/pkg/cataloger/dotnet/licenses.go
HeyeOpenSource: Fixed in a7452235035f63215ac8373c410544582bc05d0e .

HeyeOpenSource:
After tackling all review conversations, the ```Validations``` action is all green once more:
[Validations #27](https://github.com/HeyeOpenSource/syft/actions/runs/11498536974)

> * instead of adding a bunch of binary files for the tests (including unneeded images and such), could we build these to a test docker image? for example [Java runs some builds for test artifacts](https://github.com/anchore/syft/tree/main/syft/pkg/cataloger/java/test-fixtures/java-builds)

I have removed any artifacts, which are not strictly neccessary.
As you correctly stated it does not really make sense to have obsolete images and binaries clogging up the repo.

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,608 @@
...
+	"github.com/anchore/syft/internal/licenses"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/internal/fileresolver"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/scylladb/go-set/strset"
+)
+
+type nugetLicenseResolver struct {
+	opts                     CatalogerConfig
+	localNuGetCacheResolvers []file.Resolver
```
kzantow, line 29: When I initially read this, I thought there was some incorrect usage of file resolvers and wrote up some stuff about the differences between scan root and the host executing Syft. Then I realized you'd used these correctly with unindexed resolvers against the local filesystem, etc.. I don't think this is a blocker or necessarily needs to change at this point, but I did want to bring up that we've gone back and forth about this -- initially the golang licenses lookup code did the same thing; that's why the unindexed resolver was created, etc. but then we walked that decision back and [replaced the usage with standard `fs` package](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/golang/licenses.go#L44). Our latest discussions about it resulted in a decision to keep the `file.Resolver` _only_ for source content, and always use `fs` or `afero` for things we access on the host. So, if you think it would improve things to transition these to `fs.FS` or similar, and feel like doing the work, that would be great! But also, I think what you've done is fine here, too, but wanted to mention this since I'm sure you were not privy to any conversations about this.

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+}
+
+func extractLicensesFromNuGetContentFile(ctx context.Context, scanner licenses.Scanner, filePath string, nugetArchive *zip.Reader) []pkg.License {
+	out := []pkg.License{}
+
+	if nugetArchive != nil {
+		if licenseFile, err := nugetArchive.Open(filePath); err == nil {
+			licenseFileData, err := io.ReadAll(licenseFile)
+			licenseFile.Close()
```
kzantow, line 255: there is an `internal.CloseAndLogError(closer, someLocationString)` ([here](https://github.com/anchore/syft/blob/df3998b4f1c47f8d9fe8d08e244ed674eb38e9ef/internal/err_helper.go#L13)) that can be used which will automatically log on error; this could be useful in the event of read errors anywhere we are closing things -- most places would be something like this to be the most consistent:
```
defer internal.CloseAndLogError(licenseFile, filePath)
```

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+
+	return asset, nil
+}
+
+func getNuGetCachesFromProjectAssets(assets []projectAssets) []string {
+	paths := []string{}
+
+	// Try to determine NuGet package folders from project assets
+	if len(assets) > 0 {
```
kzantow, line 525: unnecessary check -- `range assets` will handle this and nil

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,608 @@
...
+
+func (c *nugetLicenseResolver) getLocalNugetFolderResolvers(assetDefinitions []projectAssets) []file.Resolver {
+	nugetPackagePaths := []string{}
+	if len(c.opts.LocalCachePaths) > 0 {
+		nugetPackagePaths = append(nugetPackagePaths, c.opts.LocalCachePaths...)
+	} else {
+		nugetPackagePaths = append(nugetPackagePaths, getNuGetCachesFromProjectAssets(assetDefinitions)...)
+
+		nugetPackagePaths = append(nugetPackagePaths, getNuGetCachesFromSDK()...)
+	}
```
kzantow, line 599: Not knowing much about NuGet -- are these mutually exclusive? should we just always append all the paths we can find?

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,608 @@
...
+
+	resolvers := []file.Resolver{}
+	if len(nugetPackagePaths) > 0 {
```
kzantow, line 602: unnecessary check

syft/pkg/cataloger/dotnet/parse_dotnet_deps.go
```diff
@@ -80,6 +85,8 @@ func parseDotnetDeps(_ context.Context, _ file.Resolver, _ *generic.Environment,
 	// sort the names so that the order of the packages is deterministic
 	sort.Strings(names)
 
+	c.licenses.assetDefinitions, _ = getProjectAssets(resolver)
```
kzantow, line 88: we should at least debug log any errors returned

syft/pkg/cataloger/dotnet/parse_dotnet_deps.go
kzantow: the answer is: `newDotnetDepsPackage` never returns nil 👍

syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go
```diff
@@ -50,6 +53,14 @@ func parseDotnetPortableExecutable(_ context.Context, _ file.Resolver, _ *generi
 		return nil, nil, err
 	}
 
+	c.licenses.assetDefinitions, _ = getProjectAssets(resolver)
```
kzantow, line 56: should log errors

kzantow (commented):
Thanks for the latest updates; I found a few other things after a finer tooth comb. Other than what's mentioned here, it's looking really good -- very appreciated!

syft/pkg/cataloger/dotnet/licenses.go
HeyeOpenSource: Thanks for the clarification about the decisions taken.
At the moment I am not quite sure, how I'd go about performing the switch towards the ```fs``` package...

Then again:
In a sense, the localNuGetCacheResolvers **are** source content:
These are the folders ('sources') from where the external NuGet dependencies are linked... 🤷‍♂️

@kzantow : What do you say?

HeyeOpenSource (commented):
In a sense, the localNuGetCacheResolvers are source content:
These are the folders ('sources') from where the external NuGet dependencies are linked...

Hence, I would currently refrain from changing from the ```file.Resolver``` implementation to an ```fs``` package implementation.

@kzantow : What do you say?

willmurphyscode:
This pr is blocked on https://github.com/anchore/syft/issues/1562 - we need to clarify the UX and policies around execing tools on the host system running Syft before we can move forward here. Please follow #1562 for more updates there.

naoriel:
I'd say this pull request may be closed if #3946 is merged. 🤷🏻‍♂️

HeyeOpenSource:
Closing this now in favor of #3946.

PR #3333, opened 2024-10-15
Performance: Use file indexer when scanning with file source

adammcclenaghan:
# Description
This PR alters `file_source.go` to use a new **file indexer,** rather than the existing **directory indexer**.

Currently, when scanning a non-archive file, file_source.go applies a filter function to the directory indexer such that all files other than the file being scanned and its parent directory are ignored by the directory indexer. See [here](https://github.com/anchore/syft/blob/main/syft/source/filesource/file_source.go#L176-L200).

This approach becomes problematic when the scanned file is inside a directory with a large number of files, for two reasons:

1. Go’s filepath.Walk provides lexical ordering guarantees by reading the entire directory into memory. For sufficiently large containing directories, **scanning a single file takes many GB’s of memory**.
2. The total time to scan the single file also increases wrt the number of files in the containing directory due to the directory walk and the time taken to perform memory allocation etc.

This Pprof shows heap allocation when scanning a file within a directory containing a large number of files, I’m including it here as proof of my root cause analysis
Screenshot 2024-10-15 at 11 45 59

Walking all of the files in the containing directory is redundant when using a file source, since as mentioned above the filter function will ignore everything other than the scanned file and its parent dir.

In this change, I have added a new **file indexer** which should match the existing behaviour of the directory indexer for a single file source. However, instead of walking the file system, it simply makes an attempt to index the containing directory and the file target.

I have also added `file.go` to satisfy the resolver interface when using the file indexer. Much of the functionality matches that of `directory.go` and I would appreciate it if there are any suggestions for improvement here, as I appreciate there's a bit of duplicated code.

The existing `directory.go` has many unit tests to verify behaviour in the event that the directory being walked contains symlinks etc. I have attempted to simplify the unit tests for `file.go` as it does not have to handle all of the complexity that `directory.go` does, but I would really appreciate extra review attention in this area as I may not be aware of all the ways a target for file_source may be defined.

I haven’t got a pprof diagram for the new approach, but memstat profiling has shown O(1) heap use wrt the number of files in the containing directory when using file source as expected.

Additionally, creating a resolver via a file_source is also happening in O(1) time wrt the number of files in the containing directory too. 

## Type of change

# Checklist:

### DIFF
--- a/syft/internal/fileresolver/file.go
+++ b/syft/internal/fileresolver/file.go
@@ -14,50 +14,43 @@ import (
 	"github.com/anchore/syft/syft/internal/windows"
 )
 
-var ErrSkipPath = errors.New("skip path")
+// Compile time assurance that we meet the Resolver interface.
+var _ file.Resolver = (*File)(nil)
 
-var _ file.Resolver = (*Directory)(nil)
-
-// Directory implements path and content access for the directory data source.
-type Directory struct {
+// File implements path and content access for the file data source.
+type File struct {
 	path          string
 	chroot        ChrootContext
 	tree          filetree.Reader
 	index         filetree.IndexReader
 	searchContext filetree.Searcher
-	indexer       *directoryIndexer
+	indexer       *fileIndexer
 }
 
-func NewFromDirectory(root string, base string, pathFilters ...PathIndexVisitor) (*Directory, error) {
-	r, err := newFromDirectoryWithoutIndex(root, base, pathFilters...)
-	if err != nil {
-		return nil, err
-	}
-
-	return r, r.buildIndex()
-}
-
-func newFromDirectoryWithoutIndex(root string, base string, pathFilters ...PathIndexVisitor) (*Directory, error) {
-	chroot, err := NewChrootContextFromCWD(root, base)
+// parent should be the symlink free absolute path to the parent directory
+// path is the filepath of the file we're creating content access for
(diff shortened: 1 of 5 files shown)


syft/internal/fileresolver/file.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	r.tree = tree
+	r.index = index
+	r.searchContext = filetree.NewSearchContext(tree, index)
+
+	return nil
+}
+
+// TODO: These are Copy-pasted from Directory.go - should we consider splitting them out into a shared place?
```
wagoodman, line 67: it might be a good time to make a `filetree` resolver (unexported, for internal use only) that is based around just taking an index and dealing with the access (the duplicated parts). This way the file and directory resolver would use this type as an embedding, but the tests and functionality live in one spot (this new filetree resolver) while the file and directory resolvers get to leverage them.

adammcclenaghan: Good idea - have gone ahead and done this, thanks!

adammcclenaghan:
Hey @wagoodman 👋 

I wanted to check in to see if there's anything else you think we need for this change, or if you're otherwise happy with the latest set of changes I've made wrt your last review

wagoodman (approved):
awesome work -- thanks for taking the time profiling and contributing the fix 🙌

PR #3339, opened 2024-10-16
fix: improve mariadb binary classifer to detect older versions

westonsteimel:
# Description

With older versions of mariadb the binary name was `mysql`, so this adjusts the binary classifier to additionally search for the expected version pattern in `mysql` binaries.

### DIFF
--- a/syft/pkg/cataloger/binary/classifiers.go
+++ b/syft/pkg/cataloger/binary/classifiers.go
@@ -351,7 +351,7 @@ func DefaultClassifiers() []Classifier {
 		},
 		{
 			Class:    "mariadb-binary",
-			FileGlob: "**/mariadb",
+			FileGlob: "**/{mariadb,mysql}",
 			EvidenceMatcher: FileContentsVersionMatcher(
 				// 10.6.15-MariaDB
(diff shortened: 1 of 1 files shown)


westonsteimel:
An example image that illustrates this is `--select-catalogers binary registry:mariadb:10.1 | grep mariadb`

syft/pkg/cataloger/binary/classifiers.go
```diff
@@ -351,14 +351,24 @@ func DefaultClassifiers() []Classifier {
...
-			FileGlob: "**/mariadb",
+			FileGlob: "**/{mariadb,mysql}",
 			EvidenceMatcher: FileContentsVersionMatcher(
 				// 10.6.15-MariaDB
 				`(?m)(?P<version>[0-9]+(\.[0-9]+)?(\.[0-9]+)?(alpha[0-9]|beta[0-9]|rc[0-9])?)-MariaDB`),
 			Package: "mariadb",
 			PURL:    mustPURL("pkg:generic/mariadb@version"),
 			CPEs:    singleCPE("cpe:2.3:a:mariadb:mariadb:*:*:*:*:*:*:*:*", cpe.NVDDictionaryLookupSource),
 		},
+		// {
```
kzantow, line 362: should this just get deleted instead of being commented out?

kzantow (approved):
LGTM

Issue #3065, opened 2024-07-24
Support Bitnami embedded SBOMs


willmurphyscode:
**What would you like to be added**:

As part of anchore/grype#1609, Syft should pick up on sboms in containers located at `/opt/bitnami` because this is how Bitnami records what's in an image.

The SBOM cataloger would probably do this already, but is off by default.

There are a few open questions here:

1. How should packages discovered by other catalogers interact with these SBOMs? For example, the binary cataloger might find Redis or MariaDB executables.
2. What if someone is building something `FROM` a Bitnami image? How do we know we can trust the SBOM?
3. If we are special-casing Bitnami images, e.g. turning the SBOM cataloger on by default only for certain images or certain paths, how do we detect this situation and what configuration options are available?

**Why is this needed**:

This is primarily needed so that running `grype` on a Bitnami image (see anchore/grype#1609) is as accurate as possible.

**Additional context**:

There are a few [open requests for more accurate Bitnami classification](https://github.com/anchore/syft/issues?q=is%3Aissue+is%3Aopen+bitnami). Ideally this work might also fix those.

kzantow:
Is there another way to scan these artifacts? Are these container images in some differing format from OCI? If the only way to identify what is installed is by scanning an SBOM, there could probably just be a Bitnami cataloger that looks for specific SBOMs in these known bitnami locations, instead of enabling the SBOM cataloger itself. It's pretty easy to just pass a reader to the SBOM decoder. And then we'd probably want to have a way to prevent SBOMs from getting scanned twice if a user _does_ enable the SBOM cataloger.

willmurphyscode:
Two questions for investigation:

1. If we add a bitnami cataloger, and turn both it and the SBOM cataloger on, do we get duplicates?
2. Do we and should we surface all artifacts and relationships from the bitnami SPDX in the Syft output SBOM?

willmurphyscode:
> 1. If we add a bitnami cataloger, and turn both it and the SBOM cataloger on, do we get duplicates?

I did an experiment to answer this.

1. Copy the [SBOM cataloger](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/sbom/cataloger.go) to make a new bitnami cataloger, but change the glob list to be only `"/opt/bitnami/**/*.spdx"`
2. Wire the new cataloger up here: https://github.com/anchore/syft/blob/main/internal/task/package_tasks.go#L151
[...]

PR #3341, opened 2024-10-16
feat: add support for Bitnami cataloguer

juan131:
# Description

This PR adds supports for a new Bitnami cataloguer so Syft is able to recognize Bitnami SBOMs and properly detect the packages available on Bitnami images.

- Fixes https://github.com/anchore/syft/issues/3065

## Type of change

# Checklist:


juan131:
cc @willmurphyscode

willmurphyscode:
Hi all! I just wanted to check in and see if there's anything I can do to help here.

My understanding is that the current state of this PR is that @juan131 is going to implement the changes discussed in https://github.com/anchore/syft/issues/3065#issuecomment-2492433009. Is that everyone else's understanding? Is there anything you all need from me? Thanks!

juan131:
Yes @willmurphyscode! That's the current status

Sorry for the long delay. I was on marriage leave almost all December and then Christmas came, so I've been out for a while. Let me resume the work where I left it

willmurphyscode:
Hi @juan131! I just testing this, but I'm seeing something surprising: 

``` sh
❯ SYFT_PACKAGE_EXCLUDE_BINARY_OVERLAP_BY_OWNERSHIP=true go run ./cmd/syft -q bitnami/postgresql:14 | rg -e NAME -e postgres
NAME                    VERSION                  TYPE           
postgresql              14.15                    binary          
postgresql              14.15.0-9                bitnami         


❯ SYFT_PACKAGE_EXCLUDE_BINARY_OVERLAP_BY_OWNERSHIP=false go run ./cmd/syft -q bitnami/postgresql:14 | rg -e NAME -e postgres
NAME                    VERSION                  TYPE           
postgresql              14.15                    binary          
postgresql              14.15.0-9                bitnami         
```

I expected those two outputs to be different, namely that in the first output, only the bitnami postgresql package would appear.

Would you mind taking a look at this? Does your understanding of what's meant to happen match mine? (Also while you're pushing changes, there are a couple conflicts with `main` and the JSON schema moved again.)

juan131:
@willmurphyscode as I reported at https://github.com/anchore/syft/issues/3065#issuecomment-2595509889, the problem seems to be that binary packages doen't implement the OwnedFiles interface.

willmurphyscode:
Hi @juan131! Thanks for that comment.

We've pushed a partial fix that needs some modification before it can be merged. More on that below.

@wagoodman and I did some digging on this branch to understand why `SYFT_PACKAGE_EXCLUDE_BINARY_OVERLAP_BY_OWNERSHIP=true` wasn't causing deduplication. It turns out that binaryPkg does not need to implement `FileOwner`, but what does need to happen is that the bitnamiPkg.OwnedFiles needs to return the full path to the binary.

At https://github.com/anchore/syft/blob/main/internal/relationship/by_file_ownership.go#L87-L131, we are adding file overlap relationships in the situation where the parent (which must be a file owner)  and the child (which does _not_ need to be a file owner) if the parent owns the path that is evidence of the child. (Even non-file-owner packages have a location that Syft uses as the evidence of the package.

The reason why the deduplication wasn't occurring is that on the bitnami postgres package, only `/opt/bitnami/postgresql`, was returned, but file ownership expects files, not directories, so `/opt/bitnami/postgresql/bin/postgresql` need to be returned in order for the deduplication to occur.

On this branch, we added a commit that simply makes the packages from the Bitnami SPDX JSON own everything under, for example `/opt/bitnami/postgresql`. This works, but it makes every package in the SDPX JSON own every path, so we need some way of subsetting down which things own which paths, or maybe we just make the "main one" (in this case postgres) own all the paths, and the others don't own any. Does your SPDX SBOM give us a good way to assign file ownership more correctly than what we're doing? The best idea we've had so far is to discover a "main" package by looking at the `documentDescribes` field on the SDPX SBOM, and make that main package own all the files, and make none of the other packages own any files. How does that sounds to you all?

I'm happy to sync on this next week, since I know there's been some back and forth and we're all eager to get this in.

juan131:
@willmurphyscode thanks so much for digging on the issue and pushing the partial fix! 😊 

> ... but file ownership expects files, not directories, so `/opt/bitnami/postgresql/bin/postgresql` need to be returned in order for the deduplication to occur.
> On this branch, we added a commit that simply makes the packages from the Bitnami SPDX JSON own everything under, for example `/opt/bitnami/postgresql`

I can see that SPDX files are also reported as files belonging to the Bitnami package. I guess, we'll have to improve the glob expression to avoid that, right? I mean, it makes little sense to consider the SPDX (e.g. `/opt/bitnami/postgresql/.spdx-postgresql.spdx`) part of the package.

> we need some way of subsetting down which things own which paths, or maybe we just make the "main one" (in this case postgres) own all the paths, and the others don't own any. Does your SPDX SBOM give us a good way to assign file ownership more correctly than what we're doing?

I see what you mean but, unfortunately, our SPDX files do not help on identifying what files under the directory belongs to a specific package. For instance, taking as an example the `.spdx-apache.spdx` file included on this diff, we have packages like the one below (`modsecurity2) and it's not possible to identify what files under `/opt/bitnami/apache` are part of it:

```json
        {
            "SPDXID": "SPDXRef-modsecurity2",
            "name": "modsecurity2",
            "versionInfo": "2.9.7",
            "downloadLocation": "https://github.com/SpiderLabs/ModSecurity/releases/download/v2.9.7/modsecurity-2.9.7.tar.gz",
            "licenseConcluded": "Apache-2.0",
            "licenseDeclared": "Apache-2.0",
            "filesAnalyzed": false,
            "externalRefs": [
                {
                    "referenceType": "cpe23Type",
                    "referenceLocator": "cpe:2.3:*:trustwave:modsecurity:2.9.7:*:*:*:*:*:*:*"
                },
                {
                    "referenceType": "purl",
                    "referenceLocator": "pkg:bitnami/modsecurity2@2.9.7?arch=arm64&distro=debian-12"
                }
            ],
            "copyrightText": "NOASSERTION"
        }
```

> The best idea we've had so far is to discover a "main" package by looking at the documentDescribes field on the SDPX SBOM, and make that main package own all the files, and make none of the other packages own any files. How does that sounds to you all?

I don't think we have other alternatives. My concern here is: would that imply that we're unable to detect duplicates on these other secondary packages?

juan131:
@willmurphyscode your approach makes sense to me, I'll implement it!!

By the way, do you agree on adapting  the glob expression to avoid SPDX files themselves be listed as part of the "main" package?

> I'm not sure what the intermediate package is telling me. The PURL for the overall `pljava` package is `pkg:bitnami/pljava@1.6.8?arch=arm64&distro=debian-12` and the PURL for the intermediate one is `pkg:maven/org.postgresql/pljava@1.6.8`

Let me figure out why we're including that "intermediate" package you mentioned with some colleague.

juan131:
@willmurphyscode the approach you suggested was implemented at https://github.com/anchore/syft/pull/3341/commits/928e49769e05f31a35ce55bb00b4ac63704babf3

As you can see below, there are no duplicates:

```console
$ go run ./cmd/syft bitnami/postgresql -q | grep -e NAME -e postgres
NAME                    VERSION                  TYPE
postgresql              17.2.0-8                 bitnami
```

And... `pljava` reports the JAR files:

```console
$ go run ./cmd/syft bitnami/postgresql -q -o syft-json | jq '.artifacts[] | select(.name == "pljava")'
{
  "id": "81ad52c1ce3ceb60",
  "name": "pljava",
  "version": "1.6.8",
  "type": "bitnami",
  "foundBy": "bitnami-cataloger",
  "locations": [
    {
      "path": "/opt/bitnami/postgresql/.spdx-postgresql.spdx",
      "layerID": "sha256:a579369b3f5df4a2ebf32d61602c7cb8fe6138fb07120f17069e7dde9e6e7a4f",
      "accessPath": "/opt/bitnami/postgresql/.spdx-postgresql.spdx",
      "annotations": {
        "evidence": "primary"
      }
    }
  ],
  "licenses": [
    {
      "value": "BSD-3-Clause",
      "spdxExpression": "BSD-3-Clause",
      "type": "concluded",
      "urls": [],
      "locations": []
    },
    {
      "value": "BSD-3-Clause",
      "spdxExpression": "BSD-3-Clause",
      "type": "declared",
      "urls": [],
      "locations": []
    }
  ],
  "language": "",
  "cpes": [
    {
      "cpe": "cpe:2.3:*:pl\\/java_project:pl\\/java:1.6.8:*:*:*:*:*:*:*",
      "source": "declared"
    }
  ],
  "purl": "pkg:bitnami/pljava@1.6.8?arch=arm64&distro=debian-12",
  "metadata": {
    "name": "pljava",
    "arch": "arm64",
    "distro": "debian-12",
    "revision": "",
    "version": "1.6.8",
    "path": "/opt/bitnami/postgresql",
    "files": [
      "/opt/bitnami/postgresql/share/pljava/pljava-api-1.6.8.jar",
      "/opt/bitnami/postgresql/share/pljava/pljava-1.6.8.jar",
      "/opt/bitnami/postgresql/share/pljava/pljava-examples-1.6.8.jar"
    ]
  }
}
```

juan131:
@willmurphyscode did you see my last commit?

syft/format/internal/spdxutil/helpers/originator_supplier.go
```diff
@@ -43,6 +43,10 @@ func Originator(p pkg.Package) (typ string, author string) { //nolint: funlen
 	case pkg.ApkDBEntry:
 		author = metadata.Maintainer
 
+	case pkg.BitnamiEntry:
```
wagoodman, line 46: this should probably be called `BitnamiSBOMEntry` (I'll push an update)

wagoodman (approved):
amazing work! thanks for all your effort on this 🙌

juan131:
Integration tests were failing after adding `GraalVMNativeImagePkg` type, I just added a commit addressing that

PR #3343, opened 2024-10-17
fix: use official CPE for linux kernel

westonsteimel:
# Description

Uses the official cpe `cpe:2.3:o:linux:linux_kernel:*:*:*:*:*:*:*:*` for the `linux-kernel-cataloger` so that downstream vulnerability matching will function for those building a grype database with operating system type CPEs included

### DIFF
--- a/syft/pkg/cataloger/kernel/package.go
+++ b/syft/pkg/cataloger/kernel/package.go
@@ -4,11 +4,25 @@ import (
 	"strings"
 
 	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/syft/cpe"
 	"github.com/anchore/syft/syft/file"
 	"github.com/anchore/syft/syft/pkg"
 )
 
 const linuxKernelPackageName = "linux-kernel"
+
+func createLinuxKernelCPEs(version string) []cpe.CPE {
+	c := cpe.NewWithAny()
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/kernel/package.go
```diff
@@ -18,6 +19,7 @@ func newLinuxKernelPackage(metadata pkg.LinuxKernel, archiveLocation file.Locati
 		PURL:      packageURL(linuxKernelPackageName, metadata.Version),
 		Type:      pkg.LinuxKernelPkg,
 		Metadata:  metadata,
+		CPEs:      []cpe.CPE{cpe.Must("cpe:2.3:o:linux:linux_kernel:*:*:*:*:*:*:*:*", cpe.NVDDictionaryLookupSource)},
```
willmurphyscode, line 22: Is it worth trying to put a version in the CPE? I don't know much about how CPEs are used downstream or whether we usually put version info in them, or maybe some other part of syft will. But we're setting a version on the package, so it would make sense to me to set a version on the CPE. But maybe I'm missing something.

westonsteimel: Yes, it needs to set the version, thanks!

willmurphyscode (approved):

Issue #3145, opened 2024-08-21
Support scanning filesystems without building an index


ariel-miculas:
**What would you like to be added**:
My use case is to scan my host filesystem and get the Linux distro information alongside with the host packages and their versions. Unfortunately, indexing the entire filesystem takes too much time (~7 minutes to call `directorySource.FileResolver` from `directory_source.go`).
I would like to avoid the indexing step and directly read the necessary files required by the `os` cataloger. I noticed there is an `unindexed_directory.go` file in the `fileresolver` package, but that's an internal package and the functions there cannot be used.
Is there a plan to expose the functionality from `unindexed_directory.go` as a public API? Or is there another way to speed up the scanning process / avoid indexing the entire filesystem beforehand?

**Why is this needed**:
This would bring a significant performance improvement to the filesystem scanning, especially when the filesystem is very large and thus building the entire index would take too much time.

**Additional context**:

wagoodman:
It ends up that using the unindexed approach is much slower on large filesystems relative to the time it takes to index the filesystem. This is because using globs like `**/package.json` (and similar) means we need to have total knowledge of the paths on the system, or pay the penalty of making several syscalls to find this out (...and that cost for each glob searched for, which for syft is a lot!). 

That being said, there are probably ways forward here, here are some high-level thoughts:

1. Build partial indexes. The main idea is to be able to reference things outside of what is being cataloged, but not pay the cost of indexing the entire search space. This seems like it would be helpful with #2574 . The cons are that there are a lot of unknowns on how this would work
2. Extend the index on-the-fly as-needed. This sounds good at first glance, but actually isn't really an option since the first `**` glob that is used would require that the full search space is indexed.
3. Speed up indexing with concurrency. @kzantow is working on that! It's not ready, so we'll keep exploring options here.
4. **[most promising]** Allow the user to opt into non-`**` globs only and use the unindexed resolver. We'd want to be able to be loud in terms of logging/UI when a `**` glob is used when in an "unindexed" mode. This has a few variants, such as:
  a. for select catalogers (primarily OS ones) have hard coded absolute paths to common DB locations (inflexible but simple)
  b. expose glob-level configuration for all catalogers, allowing the user to extend or override the globs that are searched for (flexible but potentially dangerous) 

Assuming that one day we'll land 3 in the long-term and 4 in the near-term, I think 4 has the most promise and is really down to: how can we provide that functionality while keeping it simple, flexible and safe.

ariel-miculas:
[...]

PR #3360, opened 2024-10-21
Expose option to scan without indexing 

adammcclenaghan:
# Description

There was some discussion recently regarding supporting scanning without indexing in this [issue](https://github.com/anchore/syft/issues/3145)

In that issue, the main complaint is related to the **time** taken to scan.

The most promising solution presented was:
```
[most promising] Allow the user to opt into non-** globs only and use the unindexed resolver. We'd want to be able to be loud in terms of logging/UI when a ** glob is used when in an "unindexed" mode. This has a few variants, such as:
a. for select catalogers (primarily OS ones) have hard coded absolute paths to common DB locations (inflexible but simple)
b. expose glob-level configuration for all catalogers, allowing the user to extend or override the globs that are searched for (flexible but potentially dangerous)
```

However, on memory constrained systems with large disks, file system indexing can use considerable memory. 

I am raising this PR as I'd like to propose exposing the unindexed resolver, as a form of preferring slower scan times in favour of lower memory usage. 

As a result, I would like to propose getting this change in without the added complexity of `Allow the user to opt into non-** globs`. Perhaps adding that is a reasonable next step as part of a follow up PR?

This does not necessarily fix https://github.com/anchore/syft/issues/3145 as it will not improve their scan times, however it will allow for reduced memory imapct on memory constrained systems.

As part of this change, I've changed the behaviour of `RelativeFileByPath` and added a unit test, the existing implementation didn't seem to work. When `package.go::fetchMd5Contents` ran with the unindexed resolver it did not resolve correctly because the `location` field passed into the function points at the DPKG `status` file, so the existing code which tried to do `path.Clean(path.Join(l.RealPath, p))` wound up trying to join something like `/var/lib/dpkg/status` with `/info` to search `/var/ib/dpkg/status/info` which is incorrect as we want `/var/lib/dpkg/info`.

One other thing I've noticed while testing this change is that I get more 'location' files for DPKG than the indexed resolver, however I'm not sure this is a bug, perhaps there is a bug with the existing indexed resolver?

With the indexed resolver, I get these locations for one of my DPKG artifacts:
```
      "locations": [
        {
          "path": "var/lib/dpkg/info/acpid.conffiles",
          "accessPath": "var/lib/dpkg/info/acpid.conffiles",
          "annotations": {
            "evidence": "supporting"
          }
        },
        {
          "path": "var/lib/dpkg/info/acpid.md5sums",
          "accessPath": "var/lib/dpkg/info/acpid.md5sums",
          "annotations": {
            "evidence": "supporting"
          }
        },
        {
          "path": "var/lib/dpkg/status",
          "accessPath": "var/lib/dpkg/status",
          "annotations": {
            "evidence": "primary"
          }
        }
      ],
```

With the unindexed resolver, I get these locations for the same DPKG artifact:
```
      "locations": [
        {
          "path": "var/lib/dpkg/info/acpid.conffiles",
          "accessPath": "var/lib/dpkg/info/acpid.conffiles",
          "annotations": {
            "evidence": "supporting"
          }
        },
        {
          "path": "var/lib/dpkg/info/acpid.list",
          "accessPath": "var/lib/dpkg/info/acpid.list"
        },
        {
          "path": "var/lib/dpkg/info/acpid.md5sums",
          "accessPath": "var/lib/dpkg/info/acpid.md5sums",
          "annotations": {
            "evidence": "supporting"
          }
        },
        {
          "path": "var/lib/dpkg/info/acpid.postinst",
          "accessPath": "var/lib/dpkg/info/acpid.postinst"
        },
        {
          "path": "var/lib/dpkg/info/acpid.postrm",
          "accessPath": "var/lib/dpkg/info/acpid.postrm"
        },
        {
          "path": "var/lib/dpkg/info/acpid.preinst",
          "accessPath": "var/lib/dpkg/info/acpid.preinst"
        },
        {
          "path": "var/lib/dpkg/info/acpid.prerm",
          "accessPath": "var/lib/dpkg/info/acpid.prerm"
        },
        {
          "path": "var/lib/dpkg/status",
          "accessPath": "var/lib/dpkg/status",
          "annotations": {
            "evidence": "primary"
          }
        }
      ],
```

The difference in behaviour relates to `parse_dpkg_db.go::findDpkgInfoFiles` and this call in particular:
```go
        // look for /var/lib/dpkg/info/NAME.*
	locations, err := resolver.FilesByGlob(path.Join(searchPath, name+".*"))
```

Here we pass something like:
- searchPath = `/var/lib/dpkg/info/`
- name + .* = `acpid.*`

Giving us a `FilesByGlob` path of `var/lib/dpkg/info/acpid.*`

On my system, I believe we should be matching all of these:
```
$ find ./var/lib/dpkg/info -name 'acpid.*'
./var/lib/dpkg/info/acpid.postinst
./var/lib/dpkg/info/acpid.md5sums
./var/lib/dpkg/info/acpid.postrm
./var/lib/dpkg/info/acpid.prerm
./var/lib/dpkg/info/acpid.preinst
./var/lib/dpkg/info/acpid.conffiles
./var/lib/dpkg/info/acpid.list
```

However, with the indexed directory resolver the pattern `var/lib/dpkg/info/acpid.*` gets converted in the `r.chroot.ToNativeGlob(pattern)` call to `/var/lib/dpkg/info/acpid./*` which returns no matches. This explains why the indexed resolver doesn't find locations like `acpid.prerm` in the output. The `md5sums` and `conffiles` are found by `package.go::fetchMd5Contents` and `package.go::fetchConffileContents` respectively and they use `RelativeFileByPath` which explains why they still work.

Maybe I'm unclear on what `FilesByGlob` _should_ be returning here, and perhaps the issue is with the existing unindexed_directory impl, but I'd like to get some clarification on what we expect the behaviour to be here.

## Type of change

# Checklist:

### DIFF
--- a/cmd/syft/internal/options/catalog.go
+++ b/cmd/syft/internal/options/catalog.go
@@ -53,6 +53,7 @@ type Catalog struct {
 	Platform   string         `yaml:"platform" json:"platform" mapstructure:"platform"`
 	Source     sourceConfig   `yaml:"source" json:"source" mapstructure:"source"`
 	Exclusions []string       `yaml:"exclude" json:"exclude" mapstructure:"exclude"`
+	Unindexed  bool           `yaml:"unindexed" json:"unindexed" mapstructure:"unindexed"`
 
 	// configuration for inclusion of unknown information within elements
 	Unknowns unknownsConfig `yaml:"unknowns" mapstructure:"unknowns"`
@@ -228,6 +229,9 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
 
 	flags.StringVarP(&cfg.Source.BasePath, "base-path", "",
 		"base directory for scanning, no links will be followed above this directory, and all paths will be reported relative to this directory")
+
+	flags.BoolVarP(&cfg.Unindexed, "unindexed", "",
+		"whether to index the file system or not, indexing can improve scan times but incurs a memory overhead, default false.")
 }
 
 func (cfg *Catalog) DescribeFields(descriptions fangs.FieldDescriptionSet) {
--- a/syft/source/directorysource/directory_source.go
+++ b/syft/source/directorysource/directory_source.go
@@ -21,16 +21,17 @@ import (
 var _ source.Source = (*directorySource)(nil)
 
 type Config struct {
-	Path    string
-	Base    string
-	Exclude source.ExcludeConfig
-	Alias   source.Alias
+	Path      string
+	Base      string
+	Exclude   source.ExcludeConfig
+	Alias     source.Alias
+	Unindexed bool
 }
 
 type directorySource struct {
 	id       artifact.ID
 	config   Config
-	resolver *fileresolver.Directory
+	resolver file.Resolver
 	mutex    *sync.Mutex
 }
 
@@ -145,12 +146,15 @@ func (s *directorySource) FileResolver(_ source.Scope) (file.Resolver, error) {
 		// this should be the only file resolver that might have overlap with where files are cached
 		exclusionFunctions = append(exclusionFunctions, excludeCachePathVisitors()...)
 
-		res, err := fileresolver.NewFromDirectory(s.config.Path, s.config.Base, exclusionFunctions...)
-		if err != nil {
-			return nil, fmt.Errorf("unable to create directory resolver: %w", err)
+		if s.config.Unindexed {
+			s.resolver = fileresolver.NewFromRootedUnindexedDirectory(s.config.Path, s.config.Base)
+		} else {
+			res, err := fileresolver.NewFromDirectory(s.config.Path, s.config.Base, exclusionFunctions...)
+			if err != nil {
+				return nil, fmt.Errorf("unable to create directory resolver: %w", err)
+			}
+			s.resolver = res
 		}
-
-		s.resolver = res
 	}
 
 	return s.resolver, nil
--- a/syft/source/directorysource/directory_source_provider.go
+++ b/syft/source/directorysource/directory_source_provider.go
@@ -10,20 +10,22 @@ import (
 	"github.com/anchore/syft/syft/source"
 )
 
-func NewSourceProvider(path string, exclude source.ExcludeConfig, alias source.Alias, basePath string) source.Provider {
+func NewSourceProvider(path string, exclude source.ExcludeConfig, alias source.Alias, basePath string, unindexed bool) source.Provider {
 	return &directorySourceProvider{
-		path:     path,
-		basePath: basePath,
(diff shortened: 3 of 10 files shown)


cmd/syft/internal/options/catalog.go
```diff
@@ -228,6 +229,9 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
 
 	flags.StringVarP(&cfg.Source.BasePath, "base-path", "",
 		"base directory for scanning, no links will be followed above this directory, and all paths will be reported relative to this directory")
+
+	flags.BoolVarP(&cfg.Unindexed, "unindexed", "",
+		"whether to index the file system or not, indexing can improve scan times but incurs a memory overhead, default false.")
```
wagoodman, line 234: Should we expose a CLI option for this? I feel like this is more of a low-level option, which hints at leaving it only in the config (and overridable via env vars too).

syft/source/directorysource/directory_source.go
```diff
@@ -145,12 +146,15 @@ func (s *directorySource) FileResolver(_ source.Scope) (file.Resolver, error) {
 		// this should be the only file resolver that might have overlap with where files are cached
 		exclusionFunctions = append(exclusionFunctions, excludeCachePathVisitors()...)
 
-		res, err := fileresolver.NewFromDirectory(s.config.Path, s.config.Base, exclusionFunctions...)
-		if err != nil {
-			return nil, fmt.Errorf("unable to create directory resolver: %w", err)
+		if s.config.Unindexed {
+			s.resolver = fileresolver.NewFromRootedUnindexedDirectory(s.config.Path, s.config.Base)
```
wagoodman, line 150: I think we'll need to modify this resolver to accept exclusions -- I think it would be surprising to users to have different exclusion behavior based off of changing only an index related configuration. If this is too difficult or infeasible then we should at least warn the user with `log.Warn()`.

syft/source/directorysource/directory_source_provider.go
```diff
@@ -10,20 +10,22 @@ import (
 	"github.com/anchore/syft/syft/source"
 )
 
-func NewSourceProvider(path string, exclude source.ExcludeConfig, alias source.Alias, basePath string) source.Provider {
+func NewSourceProvider(path string, exclude source.ExcludeConfig, alias source.Alias, basePath string, unindexed bool) source.Provider {
```
spiffcs, line 13: Since we're in v1.0 would this be considered a breaking change for user's who are using this method for syft as a library? 

@anchore/tools before we go to merge this sans other comments is this something we want to put more 👀 on before merge, release, and then potentially break some people's usage?

PR #3363, opened 2024-10-21
feat: Java dependency graph information

### LINKED ISSUE #3189: Java dependency graph
One of the ecosystems that often has large dependency graphs is Java, especially in the Maven ecosystem. Syft should be able to represent a full dependency graph for Maven projects (pom.xml). These dependencies should have full transitive graph information conveyed in the Syft output. 

We need to add [dependency relationship](https://github.com/anchore/syft/blob/7c96a10cbea82e94c843112c8394abac7672b0dc/syft/artifact/relationship.go#L20) types to the [Java catalogers](https://github.com/anchore/syft/tree/7c96a10cbea82e94c843112c8394abac7672b0dc/syft/pkg/cataloger/java).

This issue is tracking Java dependencies of [the larger epic](https://github.com/anchore/syft/issues/572).

kzantow:
# Description

This PR implements dependency graph information for Java packages. This applies primarily to these scenarios:
* source scan of Java Maven-based projects, which now includes appropriate top-level and _optionally, transitive dependency information_
* nested archive with embedded packages

For example: a user scanning a `.war` file with multiple embedded `.jar` files, the topmost package representing the `.war` file will have dependency relationships from the specific packages surfaced from the included `.jar` files.

Additionally, this PR includes a refactoring the internal Maven resolver functionality to an internal package to provide clearer boundaries of usage vs. internal methods.

- Fixes #3189

## Type of change

# Checklist:


syft/pkg/cataloger/java/archive_parser_test.go
```diff
@@ -1401,10 +1415,7 @@ func Test_deterministicMatchingPomProperties(t *testing.T) {
 func assignParent(parent *pkg.Package, childPackages ...pkg.Package) {
 	for i, jp := range childPackages {
 		if v, ok := jp.Metadata.(pkg.JavaArchive); ok {
-			parent := *parent
-			// PURL are not calculated after the fact for parent
-			parent.PURL = ""
```
kzantow, line 1406: this was a workaround to a timing issue that is now solved

syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -53,59 +47,172 @@ func (p pomXMLCataloger) Catalog(ctx context.Context, fileResolver file.Resolver
...
+			Name:        r.ResolveProperty(ctx, pom, pom.Name),
+			Description: r.ResolveProperty(ctx, pom, pom.Description),
+			URL:         r.ResolveProperty(ctx, pom, pom.URL),
+		},
+	}
+
+	p := &pkg.Package{
+		Name:    id.ArtifactID,
+		Version: id.Version,
+		Locations: file.NewLocationSet(
```
kzantow, line 137: Is this a bug? When packages are entirely from remote online sources and not directly based on a file, should the source that caused the package to appear get set or should this be empty?

syft/pkg/cataloger/java/internal/maven/resolver.go
```diff
@@ -293,7 +335,7 @@ func (r *mavenResolver) findPom(ctx context.Context, groupID, artifactID, versio
 
 	// resolve via network maven repository
 	if pom == nil && r.cfg.UseNetwork {
```
wagoodman, line 337: I think this was always a problem -- but this is checking the local var `pom` which [isn't being overridden in  by the local repo search](https://github.com/anchore/syft/pull/3363/files#diff-005fd8c5b8db8fa24b7ca155ab6e0b9f103f6e7f88bdb1bf459623dd8900cea1R328), it's being shadowed. So the [first check](https://github.com/anchore/syft/pull/3363/files#diff-005fd8c5b8db8fa24b7ca155ab6e0b9f103f6e7f88bdb1bf459623dd8900cea1R320) will always take precedence.

kzantow: I see that `pom` is always going to be nil here, but the logic is always supposed to run in this order to minimize network usage: check in-memory cache, check local repository, check remote. Is there anything to do other than just remove the `pom == nil` condition?

syft/pkg/cataloger/java/config.go
```diff
@@ -11,16 +14,19 @@ type ArchiveCatalogerConfig struct {
...
-		MavenLocalRepositoryDir: defaultMavenLocalRepoDir(),
-		MavenBaseURL:            mavenBaseURL,
-		MaxParentRecursiveDepth: 0, // unlimited
+		ArchiveSearchConfig:           cataloging.DefaultArchiveSearchConfig(),
+		UseNetwork:                    mavenCfg.UseNetwork,
+		UseMavenLocalRepository:       mavenCfg.UseLocalRepository,
+		MavenLocalRepositoryDir:       mavenCfg.LocalRepositoryDir,
+		MavenBaseURL:                  strings.Join(mavenCfg.Repositories, ","),
+		MaxParentRecursiveDepth:       mavenCfg.MaxParentRecursiveDepth,
+		ResolveTransitiveDependencies: false,
```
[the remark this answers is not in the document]
wagoodman: since the default for use network and use local repo are false this shouldn't be an impactful fast follow -- generally agreed 👍

syft/pkg/cataloger/java/parse_pom_xml.go
wagoodman: this is questionable for sure, but the rule of thumb here is to track the files you needed to open and read to detect the packages existence. In this case it's a little misleading since the pom might be for an entirely different package, however, it still is defensible since without that location there really would be no package to find.

I think in the future we want to be able to have locations that represent remote URLs and not exclusively file paths -- that way we could express more accurate info here on the package as to where it was discovered from.

syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -115,36 +222,45 @@ func newPomProject(ctx context.Context, r *mavenResolver, path string, pom *gopo
...
+			Description: r.ResolveProperty(ctx, pom, pom.Description),
+			URL:         r.ResolveProperty(ctx, pom, pom.URL),
 		}
 	}
 
+	m := pkg.JavaArchive{
+		PomProperties: &pkg.JavaPomProperties{
+			GroupID:    id.GroupID,
+			ArtifactID: id.ArtifactID,
+			Scope:      r.ResolveProperty(ctx, pom, dep.Scope),
```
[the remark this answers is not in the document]
wagoodman: agreed -- we should make a syft 2.0 for this and mark the field as deprecated. There is a discussion in #572  on how to mark dev/test deps on edges, but for now we've elected to not track these.

syft/pkg/cataloger/java/internal/maven/resolver.go
wagoodman: I think removing the condition is safe since there is the early return for a hit when searching m2 anyway.

wagoodman (approved):
🚀

PR #3366, opened 2024-10-22
feat: update licenses to including custom content when SPDX expressions are unable to be determined

### LINKED ISSUE #3090: feat: dpkg license improvement for non SPDX licenses
**What happened**:
Sometimes syft can encounter a dpkg license where the regular expression used to match on contents cannot correctly identify the license.

In the following example we should find things like:
```
NVIDIA Software License Agreement and CUDA Supplement to Software License Agreement
```

#### Sends contents for parsing
https://github.com/anchore/syft/blob/ca945d16e0949a41aa8786f55d21908242b224c8/syft/pkg/cataloger/debian/package.go#L101-L106

#### Searches for license clause
https://github.com/anchore/syft/blob/48f1e975f05183390d7c01718865f5f66e3f9012/syft/pkg/cataloger/debian/parse_copyright.go#L22-L41

Given a copyright file is found SOME license information should be created for a given package. No licenses is a bug.

```
syft -o json nvidia/cuda:12.5.1-cudnn-runtime-ubuntu20.04 | grant list -o json | jq -r '.results[]
 | [.license.license_id, .license.name] | @csv' | sed 's/"//g'
```

- Output of `syft version`: devel (tip of main)
- OS (e.g: `cat /etc/os-release` or similar): OSX

### LINKED ISSUE #3412: License files which do not match an SPDX expression are erroneously handled as 'unlicensed'.
**What happened:**
While working with non-OpenSource Licenses in Golang modules it was discovered, that license files which cannot be matched to an SPDX expression are erroneously handled as 'unlicensed'. (cf. #3090)

This is not the expected behaviour for an existing license file and should hence be considered a bug.

Originally posted by @HeyeOpenSource in https://github.com/anchore/syft/pull/3366#discussion_r1816179361

HeyeOpenSource:
# Description

Implements support for custom licenses.
Should probably be reviewed by @wagoodman and / or @spiffcs.

- Fixes #3412 
- Potentially fixes #3090 

## Type of change

# Checklist:

- [x] I have run ```make test``` locally without receiving any failures
- [x] 'Validations' workflow returns no errors


syft/pkg/license.go
```diff
@@ -29,6 +29,7 @@ type License struct {
 	Type           license.Type
 	URLs           []string         `hash:"ignore"`
 	Locations      file.LocationSet `hash:"ignore"`
+	Contents       string           `hash:"ignore"` // The optional binary contents of the license file
```
wagoodman, line 32: If we add support for this we should make this configurable. It could be a simple on/off (collect or do not collect contents) but there is some argument for more options here. Specifically, when we can get the ID from contents and we have a high degree of confidence that it matches then including the contents is redundant / not necessary (one of multiple possible middle of the road options). 

I think the default for this option should either be "off" or the middle of the road option.

HeyeOpenSource:
The Validations workflow finally throws no more errors at me. 👍
cf. [Validations #22](https://github.com/HeyeOpenSource/syft/actions/runs/11482588231)

syft/pkg/license.go
HeyeOpenSource: My implementation **only** ever fills the ```Contents``` field if a **non-empty** file identified as potential license file (by name) exists and **no SPDX-expression** can be determined for it with the required confidence.
Otherwise the status quo implementation is retained.
Hence I would assume that the option as you describe it is currently "off" by default. (Unless of course in the event of a non-identifiable license.) 😉

I would advocate for a later implementation of the configurability.

HeyeOpenSource: I would advocate for a later implementation of the configurability, as the current implementation mitigates the fact, that packages with an identified (but non-SPDX-compatible) license file will be erroneously handled as **unlicensed**.

HeyeOpenSource:
The implementation of this pull request mitigates the fact, that packages with an identified (but non-SPDX-compatible) license file will be erroneously handled as unlicensed.

spiffcs:
rebased and pull in upstream main - I'll work through the CI, add my review, and add @wagoodman's requested functionality about contents configuration.

Thanks for the contribution @HeyeOpenSource! Sorry this one has taken a little while.

spiffcs:
PR looks great! Adding some new tests around the format model changes to double check behavior and will merge 👍

spiffcs (approved):

HeyeOpenSource:
> rebased and pull in upstream main - I'll work through the CI, add my review, and add @wagoodman's requested functionality about contents configuration.
> 
> Thanks for the contribution @HeyeOpenSource! Sorry this one has taken a little while.

No worries.
I'm glad my contribution was accepted.

PR #3370, opened 2024-10-22
Make pre-release integration PRs

wagoodman:
This updates the existing stereoscope workflow into a generic "update anchore deps" workflow, leveraging the work done in https://github.com/anchore/workflows/pull/11 . Another repo is meant to invoke this workflow, so no taskfile updates will be added.

See an example run: https://github.com/anchore/syft/pull/3372


.github/actions/bootstrap/action.yaml
```diff
@@ -18,6 +18,9 @@ inputs:
     description: "Download test fixture cache from OCI and github actions"
     required: true
     default: "false"
+  tools:
+    description: "weather to install tools"
```
kzantow, line 22: nit: whether

Taskfile.yaml
```diff
@@ -132,6 +133,17 @@ tasks:
...
+  ## Dependency tasks #################################
+
+  update-anchore-dependencies:
+    desc: Update Anchore dependencies
+    deps: [tools]
+    requires:
+      vars:
+        - FROM
+    cmd: |
+      {{ .GH }} workflow run update-anchore-dependencies.yml -f repos="github.com/anchore/stereoscope@{{ .FROM }}"
```
kzantow, line 145: Would this task be used for anything? Generally, I think the dependencies would be updated by dependabot once an upstream project has a new version released. If I'm making a change in syft, here, I would be most interested in testing _this specific change_ across the downstream projects, namely: Grype.

.github/workflows/update-anchore-dependencies.yml
```diff
@@ -0,0 +1,49 @@
+name: PR to update Anchore dependencies
```
kzantow, line 1: I think this workflow just goes away, and there's a different workflow to run `grype/validations.yml` against a syft change in a syft PR.

.github/actions/bootstrap/action.yaml
wagoodman: 🌤️

Taskfile.yaml
wagoodman: when there are releases, yes. But this is also used to control integration branches too (update to commits without releases), something that dependabot can't do. This mechanism would happen to be used for both integration and main though during release trains.

.github/workflows/update-anchore-dependencies.yml
wagoodman: The goal is to open these pre-release PRs as integration branches in all repos during a release day, starting with pointing at integration branches for all dependencies with the PR in a draft state, then once releases start (after all pre-release PRs are green) we bump all dependencies to released versions and promote the PR from draft to mergable.

.github/workflows/update-anchore-dependencies.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Bootstrap environment
+        uses: ./.github/actions/bootstrap
+        with:
+          tools: false
+          bootstrap-apt-packages: ""
+
+      - name: Update dependencies
+        id: update
+        uses: anchore/workflows/.github/actions/update-go-dependency@add-dep-update
```
willmurphyscode, line 28: I assume this will change to `main` or a pinned tag when https://github.com/anchore/workflows/pull/11 is merged?

.github/workflows/update-anchore-dependencies.yml
```diff
@@ -0,0 +1,49 @@
+name: PR to update Anchore dependencies
+on:
+  workflow_dispatch:
+    inputs:
```
willmurphyscode, line 4: Does this need an input to control `FROM`?

.github/workflows/update-anchore-dependencies.yml
wagoodman: correct, I'd like to merge that PR first and update here

willmurphyscode (approved):

Issue #3130, opened 2024-08-15
Support to exclude scope for Maven Projects like compile, runtime, provided or test in Syft


arkajnag23:
**What happened**: Not able to find any support to decide or exclude what scope to be included for scanning.

**What you expected to happen**:
Support like how we have in Maven arguments -DincludeCompileScope=true -DincludeProvidedScope=false -DincludeRuntimeScope=true -DincludeTestScope=false

**Steps to reproduce the issue**:

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: release tag='v1.11.0' version='1.11.0' os='linux' arch='amd64'
- OS (e.g: `cat /etc/os-release` or similar):

kzantow:
I think this is a great idea. I don't think we would want to include the particular CLI option described here, at least to start but there are a couple paths forward.

One path is more complicated: capture java relationships for these dependencies (we are going to do this, anyway) but capture additional information including the dependency scope, and post-process relationships to exclude "development" scoped dependencies. This has the benefit that NPM has a similar issue where `devDependencies` often want to be excluded and could share some general logic to exclude these development-only dependencies. But there are plenty of questions with this approach, and I don't think it's going to be done in this manner especially soon.

Another path is fairly simple: [we have some maven configuration](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/config.go#L9-L13) ([and the CLI configuration](https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/java.go#L9-L13)), we could simply add a `MavenExcludeScopes []string` property, which would map to environment variable `SYFT_JAVA_MAVEN_EXCLUDE_SCOPES` and [when syft reads the dependencies](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/parse_pom_xml.go#L79), check the scope and exclude them and you could accomplish this more-or-less like:
```
SYFT_JAVA_MAVEN_EXCLUDE_SCOPES=test,runtime syft ...
```
Does this sound like something that would work for you?

arkajnag23:
Yes @kzantow this sounds good with the second approach... It's not only scope exclusion, but providing support to exclude groups or artifacts IDs , is a great benefit to Maven Users.

albert0815:
[...]

PR #3371, opened 2024-10-22
Exclude devDependencies from package-lock.json parsing

### LINKED ISSUE #2348: Syft outputs devDependencies for package-lock.json files
**What happened**:
When scanning a directory with the following files:
`package.json`

```
{
    "name": "first_app",
    "version": "1.0.0",
    "description": "a first app",
    "main": "index.js",
    "scripts": {
      "test": "testing"
    },
    "author": "",
    "license": "ISC",
    "dependencies": {
    },
    "devDependencies": {
      "async": "^3.2.4"
    }
  }
```

package-lock.json
```
{
    "name": "first_app",
    "version": "1.0.0",
    "lockfileVersion": 3,
    "requires": true,
    "packages": {
      "": {
        "name": "first_app",
        "version": "1.0.0",
        "license": "ISC",
        "devDependencies": {
          "async": "^3.2.4"
        }
      },
      "node_modules/async": {
        "version": "3.2.4",
        "resolved": "https://registry.npmjs.org/async/-/async-3.2.4.tgz",
        "integrity": "sha512-iAB+JbDEGXhyIUavoDl9WP/Jj106Kz9DEn1DPgYw5ruDn0e3Wgi3sKFm55sASdGBNOQB8F59d9qQ7deqrHA8wQ==",
        "dev": true
      }
    }
  }
```

It outputs

```
 ✔ Indexed file system                                                                                                                                                                                            tests/data/js_dependencies
 ✔ Cataloged packages              [2 packages]
NAME       VERSION  TYPE
async      3.2.4    npm
first_app  1.0.0    npm
```

with the `async` devDependency.

Syft do not output dev-dependencies as it's done when scanning a `Pipfile.lock`.

Run syft on a directory containing the above file.

**Environment**:
- Output of `syft version`:

````
$ syft version
Application: syft
Version:    0.97.1
BuildDate:  2023-11-17T20:53:01Z
GitCommit:  Homebrew
Platform:   darwin/arm64
GoVersion:  go1.21.4
Compiler:   gc
````

````
$ sw_vers
ProductName:		macOS
ProductVersion:		14.1.1
BuildVersion:		23B81
````

njv299:
# Description

This updates the Javascript cataloger so that packages that are only present as dev dependencies found in `package-lock.json` files are ignored and not included in the resulting SBOM. This behavior more closely aligns with that of other similar catalogers, such as the Python Pipfile.lock parser.

This change supports all current `package-lock.json` schema versions (`1`, `2`, and `3`) in accordance with the spec found [here](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json).

- Fixes #2348 

## Type of change

# Checklist:

### DIFF
--- a/cmd/syft/internal/options/catalog.go
+++ b/cmd/syft/internal/options/catalog.go
@@ -166,6 +166,7 @@ func (cfg Catalog) ToPackagesConfig() pkgcataloging.Config {
 					WithFromLDFlags(cfg.Golang.MainModuleVersion.FromLDFlags),
 			),
 		JavaScript: javascript.DefaultCatalogerConfig().
+			WithIncludeDevDependencies(*multiLevelOption(false, cfg.JavaScript.IncludeDevDependencies)).
 			WithSearchRemoteLicenses(*multiLevelOption(false, enrichmentEnabled(cfg.Enrich, task.JavaScript, task.Node, task.NPM), cfg.JavaScript.SearchRemoteLicenses)).
 			WithNpmBaseURL(cfg.JavaScript.NpmBaseURL),
 		LinuxKernel: kernel.LinuxKernelCatalogerConfig{
--- a/cmd/syft/internal/options/javascript.go
+++ b/cmd/syft/internal/options/javascript.go
@@ -3,8 +3,9 @@ package options
 import "github.com/anchore/clio"
 
 type javaScriptConfig struct {
-	SearchRemoteLicenses *bool  `json:"search-remote-licenses" yaml:"search-remote-licenses" mapstructure:"search-remote-licenses"`
-	NpmBaseURL           string `json:"npm-base-url" yaml:"npm-base-url" mapstructure:"npm-base-url"`
+	SearchRemoteLicenses   *bool  `json:"search-remote-licenses" yaml:"search-remote-licenses" mapstructure:"search-remote-licenses"`
+	NpmBaseURL             string `json:"npm-base-url" yaml:"npm-base-url" mapstructure:"npm-base-url"`
+	IncludeDevDependencies *bool  `json:"include-dev-dependencies" yaml:"include-dev-dependencies" mapstructure:"include-dev-dependencies"`
 }
 
 var _ interface {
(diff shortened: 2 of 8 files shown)


wagoodman (approved):

njv299:
@joaquinricci I would certainly be in support of a `--include-dev-deps` configuration option, but I would argue that that should be an option that is consistently applied across all catalogers (or made available as a configurable option for all relevant catalogers). As of right now, there is different behavior across catalogers with respect to dev dependencies: `Pipfile.lock` file parsing does not include dev deps, `package.json` also doesn't include them, while `package-lock.json` parsing does. 

It's obviously not my decision to make, but in the near-term my opinion is that this PR makes the behavior more consistent as-is and there should be follow-on issues and PRs to holistically add a consistent, comprehensive set of `--include-dev-deps` functionality.

kzantow:
For what it's worth, I agree with all of the comments here: @joaquinricci and @njv299 that this probably should be configurable as part of this PR, the configuration should be applied consistently across pertinent catalogers, and _also_ agree that the default behavior should be to _exclude_ development-only time dependencies.

This is absolutely something that other ecosystems would benefit from: [Java, for example, should probably exclude `test`-scoped dependencies by default](https://github.com/anchore/syft/issues/3130). I think it would be worth spending a little time thinking about how configuration should work. It may be useful to move this discussion to Discourse (there is [a related topic](https://anchorecommunity.discourse.group/t/syft-own-format-how-to-distinguish-dependency-vs-devdependency-for-npm/151)).

njv299:
To clarify my stance on this: In my opinion this PR should be merged as-is. Based on the discussion in the relevant [issue](https://github.com/anchore/syft/issues/2348), the user-feedback that devDependencies are _generally_ not preferred to be included, and my own N=1 experience, it seems very valuable to change the default behavior of Syft today to exclude devDependencies. This solves what appears to be a fairly pressing need for a not-insignificant number of people (myself included), and makes the behavior of Syft more consistent across all catalogers.

I am also fully in support of subsequently making this behavior configurable as discussed above. Personally, though, I think this should be done via follow-on PRs so that the advantages mentioned above can be realized as soon as possible. If time permits I will be happy to help with the implementation of the configuration options, but getting this PR merged and released soon will fix some significant issues in my personal use of Syft.

kzantow:
Hey @njv299 -- I should note, my comments above were not necessarily blockers by any means. However, since there was an ask for making this configurable directly on the PR, would you mind if I pushed a small change to just add the appropriate config/environment variable for the flag so everyone would probably be happy moving this forward?

njv299:
@kzantow Absolutely, that sounds great to me! Thank you!

wagoodman (approved):

kzantow:
Thanks very much for the contribution, @njv299!

PR #3399, opened 2024-10-30
Add support for dependencies and purl for Native Image SBOMs

rudsberg:
# Description
This PR adds support for the "dependencies" and "purl" fields in SBOMs embedded in Native Images. These fields will be included in SBOMs starting from GraalVM for JDK 24.

Other fixes:
- Resolves empty package ID warnings by setting the package ID. For example, warnings like the one below were previously emitted for valid SBOM components.
```
[0000]  WARN found package with empty ID while adding to the collection: Pkg(name="commons-validator" version="1.9.0" type="graalvm-native-image" id="")
```
- Added support for including the component listed under `metadata/component` in "components". 
Note: While such a component ideally belong under `metadata/component`, the Cataloger interface (`[]pkg.Package, []artifact.Relationship, error`) does not allow specifying that.

## Type of change

# Checklist:
- [x] Added unit tests for new behavior 
- [x] Tested common scenarios to prevent regressions
- [x] Added code comments for complex sections

### DIFF
--- a/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
+++ b/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
@@ -8,7 +8,6 @@ import (
 	"debug/macho"
 	"debug/pe"
 	"encoding/binary"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -135,58 +145,71 @@ func getPackage(component nativeImageComponent) pkg.Package {
...
-		return nil, fmt.Errorf("could not read the java native-image SBOM: %w", err)
+		return nil, nil, fmt.Errorf("could not read the java native-image SBOM: %w", err)
 	}
 
-	var sbomContent nativeImageCycloneDX
-	err = json.Unmarshal(output, &sbomContent)
-	if err != nil {
-		return nil, fmt.Errorf("could not unmarshal the java native-image SBOM: %w", err)
+	var sbom cyclonedx.BOM
+	if err := json.Unmarshal(output, &sbom); err != nil {
```
kzantow, line 195: It might be best to use a [cyclondex.BOMDecoder](https://pkg.go.dev/github.com/CycloneDX/cyclonedx-go#BOMDecoder) or even [format.Decode](https://github.com/anchore/syft/blob/main/syft/format/decoders.go#L36C6-L36C75) -- if we used the `format.Decode`, it _absolutely_ should handle the dependency mapping to relationships for CycloneDX so we wouldn't have duplicated code here or even have to know anything about cyclonedx specifics

kzantow (commented):
Thanks for the contribution, @rudsberg -- overall this looks great, I'm debating whether it's important enough to centralize the CycloneDX -> Syft data model using the `format` package. I think that would be the only hold up, though

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
rudsberg: Thanks for your feedback. I updated the code to call
```
sbom, _, _, err := cyclonedxjson.NewFormatDecoder().Decode(gzreader)
```
and simply return the packages and relationships. This effectively reuses the existing infrastructure ✅ 

This should resolve your concerns. Please have a look.

kzantow: > This reuses the existing infrastructure in an effective way

Perfect! Thanks much; this looks great!

kzantow (approved):
🎉

PR #3402, opened 2024-10-31
Add package dependency completness field

### LINKED ISSUE #3010: Supply "depth" information when including relationships
Relationship _depth_ information, when Syft is unable to provide a full transitive dependency graph.

**Why is this needed**:
One of the data elements mentioned in the [NTIA minimum requirements](https://www.ntia.doc.gov/files/ntia/publications/sbom_minimum_elements_report.pdf) is the _depth_ of relationships. If Syft is able to build an accurate SBOM with a full transitive-dependency graph, that would be ideal, but different scenarios prevent this information from being included or accurately depicting the transitive graph. Some examples are Python `requirements.txt` and Go binary mod information, which only provide a flat list of dependencies. Or binaries which are only directly identified without dependent component information.

One solution is to provide an "unknown" indicator that Syft was unable to determine a full transitive dependency graph, or Syft stopped after 5-levels deep resolving online parent references. These can be returned as "unknowns" from catalogers where appropriate to be associated with the file(s) where package graph information originated.

**Additional context**:
This is likely to be dependent [the PR for known unknowns](https://github.com/anchore/syft/pull/2998) getting merged.

This is a part of #632

wagoodman:
# Description

Adds a field that describes the completeness of a package's direct dependencies:

- `complete`: the package has all of its direct dependencies resolved and related to this package.
- `incomplete`: the package does not have all of its direct dependencies resolved.
- `mixed`: a superset of `complete` -- indicates that the package has all of its direct dependencies resolved as well as some or all of indirect dependencies.  What is notable about this is that direct and indirect dependencies are linked directly to this package and are not separable (you cannot distinguish between a direct and indirect dependency from the perspective of this package).
- `unknown`: indicates when dependency resolution mechanism for this package is not well understood.

In addition to adding this new field, all catalogers were updated to raise up accurate values for this field:

| Cataloger                            | Completeness | Comment |
|--------------------------------------|--------------|---------|
| alpm-db-cataloger                    | complete     |         |
| apk-db-cataloger                     | complete     |         |
| binary-classifier-cataloger          | unknown      | classifiers are limited to identifying package identities, but have no information about dependencies |
| cargo-auditable-binary-cataloger     | incomplete     |         |
| cocoapods-cataloger                  | incomplete     |         |
| conan-cataloger                      | conanfile.txt: incomplete, conan.lock: complete   |         |
| conan-info-cataloger                 | complete        |         |
| dart-pubspec-lock-cataloger          | incomplete     |         |
| dotnet-deps-cataloger                | complete     |         |
| dotnet-portable-executable-cataloger | incomplete      |         |
| dpkg-db-cataloger                    | complete     |         |
| elf-binary-package-cataloger         | unknown      | though we can look for shared libs, we cannot see static dependencies nor dynamic dependencies using dlopen. This means that, even in cases where the dep info is actually complete, we can't programmatically determine that. |
| elixir-mix-lock-cataloger            | incomplete     |         |
| erlang-otp-application-cataloger     | incomplete     |         |
| erlang-rebar-lock-cataloger          | incomplete     |         |
| github-action-workflow-usage-cataloger | complete   |  the only known exception to this is shared workflows |
| github-actions-usage-cataloger       | incomplete      |         |
| go-module-binary-cataloger           | mixed        |         |
| go-module-file-cataloger             | mixed     |         |
| graalvm-native-image-cataloger       | incomplete      |         |
| haskell-cataloger                    | incomplete        |         |
| java-archive-cataloger               | complete/incomplete      | all packages are assumed to be incomplete unless searching for transitive dependencies configuration option has been enabled |
| java-gradle-lockfile-cataloger       | incomplete     |         |
| java-jvm-cataloger                   | incomplete      |         |
| java-pom-cataloger                   | incomplete     |         |
| javascript-lock-cataloger            | incomplete     |         |
| javascript-package-cataloger         | incomplete     | we find all dependency nodes but are not crafting relationships |
| linux-kernel-cataloger               | incomplete      |         |
| lua-rock-cataloger                   | incomplete        |         |
| nix-store-cataloger                  | incomplete     |         |
| opam-cataloger                       | incomplete   |         |
| php-composer-installed-cataloger     | incomplete   |         |
| php-composer-lock-cataloger          | incomplete   |         |
| php-pecl-serialized-cataloger        | incomplete   |         |
| portage-cataloger                    | incomplete   |         |
| python-installed-package-cataloger   | incomplete   |         |
| python-package-cataloger             | requirements.txt: incomplete, setup.py: incomplete, poetry.lock: complete, Pipfile.lock: incomplete   |         |
| r-package-cataloger                  | incomplete   |         |
| rpm-archive-cataloger                | incomplete   |         |
| rpm-db-cataloger                     | complete   |         |
| ruby-gemfile-cataloger               | incomplete   |         |
| ruby-gemspec-cataloger               | incomplete   |         |
| ruby-installed-gemspec-cataloger     | incomplete   |         |
| rust-cargo-lock-cataloger            | incomplete   |         |
| sbom-cataloger                       | unknown   | this may need some discussion |
| swift-package-manager-cataloger      | incomplete   |         |
| swipl-pack-cataloger                 | incomplete   |         |
| wordpress-plugins-cataloger          | incomplete   |         |

- Closes #3010

## Type of change

# Checklist:


willmurphyscode:
@wagoodman I wanted to make sure I understand why the `rust-cargo-lock-cataloger` is labeled as "complete" not as "mixed".

It seems like as we make a set of dependency relationships more complete, it passes from complete, to mixed, and back to complete, which is surprising:

1. It finds all the direct dependencies and nothing else -> "complete"
2. It finds all the direct and some or all indirect dependencies, but you can't tell how direct the dependencies are -> "mixed"
3. It finds all direct and indirect dependencies, and you _can_ tell which ones are direct -> "complete"

It seems weird to me that a cataloger can effectively get promoted from complete to mixed, and then from mixed to complete, by adding relationship data.

When I first read the descriptions above, I thought, "rust isn't complete, it's mixed, because it includes transitive dependencies."

Am I understanding correctly here?

wagoodman:
you're correct, cargo.lock should be `mixed` 👍

syft/pkg/cataloger/golang/parse_go_mod.go
wagoodman, line 1: this is not labeling indirect deps with `incomplete`

willmurphyscode:
Cross posting https://github.com/anchore/syft/issues/3010#issuecomment-2485475311 so that it gets seen here.

syft/pkg/cataloger/golang/parse_go_mod.go
wagoodman: fixed

syft/format/syftjson/model/package.go
```diff
@@ -24,16 +24,17 @@ type Package struct {
...
+	Name         string                     `json:"name"`
+	Version      string                     `json:"version"`
+	Type         pkg.Type                   `json:"type"`
+	FoundBy      string                     `json:"foundBy"`
+	Locations    []file.Location            `json:"locations"`
+	Licenses     licenses                   `json:"licenses"`
+	Language     pkg.Language               `json:"language"`
+	CPEs         cpes                       `json:"cpes"`
+	PURL         string                     `json:"purl"`
+	Dependencies pkg.DependencyCompleteness `json:"dependencies"`
```
willmurphyscode, line 37: Is this the right name for this field? `licenses` is the actual licenses, `cpes` is the actual CPEs, `locations` is the actual locations, but `dependencies` is the completeness and directness of the dependencies. What about `dependencyCompleteness` or something? I hate to make it that much longer, but I also hate to have `dependencies` be the only plural noun on here that's no a slice of noun.

syft/format/syftjson/to_syft_model.go
```diff
@@ -338,16 +338,17 @@ func toSyftPackage(p model.Package, idAliases map[string]string) pkg.Package {
...
-		Version:   p.Version,
-		FoundBy:   p.FoundBy,
-		Locations: file.NewLocationSet(p.Locations...),
-		Licenses:  pkg.NewLicenseSet(toSyftLicenses(p.Licenses)...),
-		Language:  p.Language,
-		Type:      p.Type,
-		CPEs:      cpes,
-		PURL:      p.PURL,
-		Metadata:  p.Metadata,
+		Name:         p.Name,
```
willmurphyscode, line 341: I was expecting to see `syft/format/...` for other formats. Do SPDX and CDX have anywhere to put dependency completeness?

syft/pkg/cataloger/binary/elf_package.go
```diff
@@ -16,7 +16,10 @@ func newELFPackage(metadata elfBinaryPackageNotes, locations file.LocationSet) p
 		PURL:      packageURL(metadata),
 		Type:      pkgType(metadata.Type),
 		Locations: locations,
-		Metadata:  metadata.ELFBinaryPackageNoteJSONPayload,
+		// though we can look for shared libs, we cannot see static dependencies nor dynamic dependencies using dlopen.
+		// this means that, even in cases where the dep info is actually complete, we can't programmatically determine that.
```
willmurphyscode, line 20: I can't figure out whether this means it should be `Incomplete` or `Unknown`. Maybe add a sentence to the comment about why it's not `Unknown`?

syft/pkg/cataloger/cpp/package.go
```diff
@@ -67,35 +67,43 @@ func splitConanRef(ref string) *conanRef {
 }
 
 func newConanfilePackage(m pkg.ConanfileEntry, locations ...file.Location) *pkg.Package {
-	return newConanPackage(m.Ref, m, locations...)
+	// though a conanfile is a listing of direct dependencies, we are not capturing these today, so are forced to answer incomplete
```
willmurphyscode, line 70: https://github.com/anchore/syft/issues/572 and this file reverse conaninfo.txt and conanfile.txt. I'm not sure which one is right.

syft/pkg/cataloger/dart/parse_pubspec_lock_test.go
```diff
@@ -14,85 +14,92 @@ func TestParsePubspecLock(t *testing.T) {
...
-			Locations: fixtureLocationSet,
-			Language:  pkg.Dart,
-			Type:      pkg.DartPubPkg,
+			Name:         "ale",
+			Version:      "3.3.0",
+			PURL:         "pkg:pub/ale@3.3.0?hosted_url=pub.hosted.org",
+			Locations:    fixtureLocationSet,
+			Language:     pkg.Dart,
+			Type:         pkg.DartPubPkg,
+			Dependencies: pkg.IncompleteDependencies,
```
willmurphyscode, line 23: As an aside, it's really nice to have this field in tests, because now as we work on #572, we can see "oh yeah, this is complete now," and fix it.

syft/pkg/cataloger/elixir/package.go
```diff
@@ -14,7 +14,9 @@ func newPackage(d pkg.ElixirMixLockEntry, locations ...file.Location) pkg.Packag
 		Locations: file.NewLocationSet(locations...),
 		PURL:      packageURL(d),
 		Type:      pkg.HexPkg,
-		Metadata:  d,
+		// we do not attempt to parse dependencies from the mix.lock file
```
willmurphyscode, line 17: Because we know that we didn't try to parse the dependencies, it's incomplete. Right?

syft/pkg/cataloger/githubactions/parse_composite_action_test.go
```diff
@@ -15,18 +15,20 @@ func Test_parseCompositeActionForActionUsage(t *testing.T) {
 
 	expected := []pkg.Package{
 		{
-			Name:      "actions/setup-go",
-			Version:   "v4",
-			Type:      pkg.GithubActionPkg,
-			Locations: fixtureLocationSet,
-			PURL:      "pkg:github/actions/setup-go@v4",
+			Name:         "actions/setup-go",
```
willmurphyscode, line 18: Do we need a test case that exercises the complete dependencies path?

willmurphyscode: Nevermind, posted too soon.

syft/pkg/cataloger/java/parse_pom_xml.go
```diff
@@ -196,16 +219,25 @@ func collectDependencies(ctx context.Context, r *maven.Resolver, resolved map[ma
...
-		pkgs = append(pkgs, *depPkg)
+		pkgs = append(pkgs, depPkg)
 		if parentPkg != nil {
-			relationships = append(relationships, artifact.Relationship{
-				From: *depPkg,
-				To:   *parentPkg,
+			relationships = append(relationships, artifact.Relationship{ //nolint:gocritic // we intentionally want to use the reference to the package which will still be mutated and finalized to a value later
+				// both the to and from packages may be mutated based on the resolved dependencies, so we need references to these values
+				// to ensure the nodes used are consistent with the final state of the packages.
+				// note: it is VITAL that these references are replaced with values by the caller of this function before using
```
willmurphyscode, line 227: These kind of comments always make me nervous. Is there something we can do to make sure this happens? What if we return something that represents instructions for building an `[]artifact.Relationships` in the caller, so that the caller can't forget and return a `[]artifact.Relationships` with unexpected types in it?

syft/pkg/cataloger/python/parse_wheel_egg.go
```diff
@@ -40,6 +40,7 @@ func parseWheelOrEgg(ctx context.Context, resolver file.Resolver, _ *generic.Env
 		newPackageForPackage(
 			*pd,
 			findLicenses(ctx, licenseScanner, resolver, *pd),
+			pkg.CompleteDependencies,
```
willmurphyscode, line 43: I'm missing something. This says "complete dependencies" but always returns a nil relationships array.

syft/pkg/cataloger/redhat/parse_rpm_manifest.go
```diff
@@ -44,7 +44,10 @@ func parseRpmManifest(_ context.Context, _ file.Resolver, _ *generic.Environment
 			continue
 		}
 
-		p := newDBPackage(reader.Location, *metadata, nil, nil)
+		// Create a new package representing the RPM manifest entry
+		// note: we don't know the quality of the manifest dependency information, so even though there are multiple
```
willmurphyscode, line 48: Can you help me understand why this is "Incomplete" and not "Unknown"? It sounds like we don't know the completeness state, which sounds "unknown" to me.

How is this different than the elf metadata cataloger, where we say, "someone could link extra stuff in at runtime and we'll never know, so unknown"?

syft/pkg/cataloger/rust/package.go
```diff
@@ -46,6 +47,8 @@ func newPackageFromAudit(dep *rustaudit.Package, locations ...file.Location) pkg
 		Language:  pkg.Rust,
 		Type:      pkg.RustPkg,
 		Locations: file.NewLocationSet(locations...),
+		// no attempt is made by the parser function to resolve dependencies
+		Dependencies: pkg.IncompleteDependencies,
```
willmurphyscode, line 51: The rust auditable binary cataloger is a lot like the Go binary cataloger. I think this cataloger can be changed to emit a "main" package and CompleteWithIndirect. Or do we not trust the completeness of this metadata if it's present?

syft/pkg/package.go
```diff
@@ -17,17 +17,18 @@ import (
...
+	FoundBy      string                 `hash:"ignore" cyclonedx:"foundBy"` // the specific cataloger that discovered this package
+	Locations    file.LocationSet       // the locations that lead to the discovery of this package (note: this is not necessarily the locations that make up this package)
+	Licenses     LicenseSet             // licenses discovered with the package metadata
+	Language     Language               `hash:"ignore" cyclonedx:"language"` // the language ecosystem this package belongs to (e.g. JavaScript, Python, etc)
+	Type         Type                   `cyclonedx:"type"`                   // the package type (e.g. Npm, Yarn, Python, Rpm, Deb, etc)
+	CPEs         []cpe.CPE              `hash:"ignore"`                      // all possible Common Platform Enumerators (note: this is NOT included in the definition of the ID since all fields on a CPE are derived from other fields)
+	PURL         string                 `hash:"ignore"`                      // the Package URL (see https://github.com/package-url/purl-spec)
+	Dependencies DependencyCompleteness `hash:"ignore"`                      // the completeness of the dependency information for this package
```
willmurphyscode, line 30: I think I said this above, but I don't think `Dependencies` is the right name for this field. `Licenses` is of type `LicenseSet`, which is a set of licenses. `CPEs` is a slice of `cpe.CPE`. `Locations` is a `LocationSet`. `Dependencies` is an enum about how complete the dependencies are. It seems to break an established pattern.

syft/pkg/cataloger/binary/elf_package.go
wagoodman: it should be unknown -- the table in the PR description is right, the code is wrong. I'll adjust the code.

syft/pkg/cataloger/cpp/package.go
wagoodman: 572 is wrong, the code and table in PR description is right. I've updated the PR description on 572 to reflect the correct state + PR references

syft/pkg/cataloger/dotnet/parse_dotnet_portable_executable.go
```diff
@@ -76,12 +76,14 @@ func buildDotNetPackage(versionResources map[string]string, f file.LocationReadC
...
 		Name:      name,
-		Version:   version,
+		Version:   ver,
 		Locations: file.NewLocationSet(f.Location.WithAnnotation(pkg.EvidenceAnnotationKey, pkg.PrimaryEvidenceAnnotation)),
 		Type:      pkg.DotnetPkg,
 		Language:  pkg.Dotnet,
-		PURL:      portableExecutablePackageURL(name, version),
-		Metadata:  metadata,
+		PURL:      portableExecutablePackageURL(name, ver),
+		// by nature PE metadata does not have any dependency information, thus we are forced to claim incomplete
```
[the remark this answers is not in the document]
wagoodman: basic definitions:

> incomplete: the package does not have all of its direct dependencies resolved.
> unknown: indicates when dependency resolution mechanism for this package is not well understood.

I've struggled through most of development on a good firm definition for unknown vs incomplete. 

> Is "PE file does not include dependency information, so this package's dependencies are known to be incomplete." a good summary of the comment here then?

It seems like one, but I'm stuck on "dependencies are known to be incomplete", which isn't quite true.

Do you think we should extend `unknown` to really mean "no assertion" (for any reason), in which case may things listed as incomplete in the interim would switch to unknown.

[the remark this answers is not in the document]
wagoodman: I've adjusted the definitions based off of our offline discussions:

incomplete: indicates that the package is known to not have all of its direct dependencies listed.

unknown: indicates that the completeness of the dependencies cannot be  considered positively complete or incomplete. This should be used when the dependency resolution mechanism is not well understood, the set of dependencies is unknowable, or no attempt has been made to resolve dependencies (no assertion).

That also means that I don't think we have a case yet where we would list 'incomplete'.

syft/pkg/cataloger/java/parse_pom_xml.go
wagoodman: yes that's what the `finalizeRelationships` call at the final return of the cataloger (a chokepoint) https://github.com/anchore/syft/blob/5b7ec60f8d13f7a211aa4a799bcbd84edd1c35af/syft/pkg/cataloger/java/parse_pom_xml.go#L84-L97

syft/pkg/cataloger/python/parse_wheel_egg.go
wagoodman: In cases where a cataloger parser function only has visibility about a single package, but the parser function is called numerous times in a filetree, then there needs to be a post-cataloger function that "ties" the discovered packages together outside of the parser function -- python egg/wheel metadata works like this. Here's the post-processor... invoked within the cataloger after all parser function calls, but before it returns it's results: https://github.com/anchore/syft/blob/05e8ba948d94de9a3df86c79bd0255eabefea1b8/syft/pkg/cataloger/python/cataloger.go#L43 .

syft/pkg/cataloger/redhat/parse_rpm_manifest.go
wagoodman: This was changed to "unknown" per https://github.com/anchore/syft/pull/3402#discussion_r1871423812

syft/format/syftjson/to_syft_model.go
willmurphyscode: Did this change get missed? I don't see a diff in the cyclonedx and cyclonedx json formatters, and I expected to. (I think it's also ok to do this as a follow up PR, since the current PR is large.)

syft/pkg/cataloger/java/parse_pom_xml.go
willmurphyscode: I see that. I think I'd rather the type system remember that `finalizeRelationships` needs to be called. For example, you could have a local struct `unfinalizedRelationship` and return a slice of those, and then finalize relationship can take a slice of `unfinalizedRelationship` and return a slice of a `artifact.Relationship`. That way if someone tries to use this function without finalizing the relationships, the compiler won't let them.

syft/pkg/cataloger/rust/cataloger_test.go
```diff
@@ -12,27 +12,29 @@ func TestNewAuditBinaryCataloger(t *testing.T) {
...
-			Language:  pkg.Rust,
-			Type:      pkg.RustPkg,
+			Name:         "auditable",
+			Version:      "0.1.0",
+			PURL:         "pkg:cargo/auditable@0.1.0",
+			FoundBy:      "cargo-auditable-binary-cataloger",
+			Locations:    file.NewLocationSet(file.NewVirtualLocation("/hello-auditable", "/hello-auditable")),
+			Language:     pkg.Rust,
+			Type:         pkg.RustPkg,
+			Dependencies: pkg.UnknownDependencyCompleteness,
```
willmurphyscode, line 22: Does this need to change now that #3500 is merged?

syft/pkg/cataloger/rust/package.go
```diff
@@ -46,6 +47,8 @@ func newPackageFromAudit(dep *rustaudit.Package, locations ...file.Location) pkg
 		Language:  pkg.Rust,
 		Type:      pkg.RustPkg,
 		Locations: file.NewLocationSet(locations...),
+		// no attempt is made by the parser function to resolve dependencies
```
willmurphyscode, line 50: I think the attempt is made now that #3500 is merged.

syft/pkg/cataloger/swipl/parse_pack_test.go
```diff
@@ -9,17 +9,18 @@ import (
 	"github.com/anchore/syft/syft/pkg/cataloger/internal/pkgtest"
 )
 
-func TestParsePackPackage(t *testing.T) {
+func xTestParsePackPackage(t *testing.T) {
```
willmurphyscode, line 12: Is committing this `x` intentional?

wagoodman: great catch, that's a typo

syft/pkg/cataloger/rust/package.go
wagoodman: I'll handle this on the rebase

henrysachs:
@wagoodman is there still an ongoing effort on this or are there other plans for dependency completeness? I would love this feature/field in syft!

wagoodman:
We want to do this, however, how this should be done is way out of date. [This PR](https://github.com/anchore/syft/pull/4317) added a new capabilities subcommand, which is driven by test observations and some parsing of golang code. This nice thing about this is that the source of truth is the code and the tests that prove the code works. Part of this output is a description of which catalogers have what dependency capabilities (direct vs transitive dependencies, topology, kinds of nodes). This is exactly what this PR is attempting to convey, but embedded into the SBOM instead of a generic cataloger description in a separate command.

This should be rewritten with this new dataset in mind. The only open question is how should this be conveyed in the output (a single value, an object describing nodes and edges separately, something else).

PR #3410, opened 2024-11-01
chore: update executor back to default of not bubbling errors to main flow

spiffcs:
# Description

The latest releases of syft saw a change where `err` was returned by the `executor` causing the main control flow to exit on task failure. This PR reverts this change and instead writes the errors to a multi err value which sets prog warnings in the TUI

- Fixes: NO BUG FILED

# Checklist:

Behavior on latest release:
```
 ✔ Loaded image                                                                                                                                                                                                                        
 ✔ Parsed image                                                                                                                                                                                 
 ⠧ Cataloging contents             ━━━━━━━━━━━━━━━━━━━━                                                                                                                                                
   └── ⠧ Packages                        [302 packages]
failed to run tasks: 1 error occurred:
        * failed to run task: java-binary-openjdk: failed to resolve files by glob (**/java): unable to search by glob="**/java": unable to get ref for path="/etc/alternatives/java": cycle during symlink resolution
java-binary-ibm: failed to resolve files by glob (**/java): unable to search by glob="**/java": unable to get ref for path="/etc/alternatives/java": cycle during symlink resolution
java-binary-oracle: failed to resolve files by glob (**/java): unable to search by glob="**/java": unable to get ref for path="/etc/alternatives/java": cycle during symlink resolution
java-binary-graalvm: failed to resolve files by glob (**/java): unable to search by glob="**/java": unable to get ref for path="/etc/alternatives/java": cycle during symlink resolution
```

Behavior on this PR showing warnings in cataloged contents, but still produces an SBOM
```
 ✔ Loaded image                                                                                                                                                                                                                        
 ✔ Parsed image                                                                                                                                                                                 
 ✘ Cataloged contents                                                                                                                                                                                  
   ├── ✔ Packages                        [296 packages]
   ├── ✔ File digests                    [3,829 files]
   ├── ✔ File metadata                   [3,829 locations]
   └── ✔ Executables                     [998 executables]
NAME                                VERSION                            TYPE
adduser                             3.118ubuntu2                       deb
...
```

### DIFF
--- a/internal/unknown/path_error.go
+++ b/internal/unknown/path_error.go
@@ -1,0 +1,33 @@
+package unknown
+
+import (
+	"regexp"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/file"
+)
+
+var pathErrorRegex = regexp.MustCompile(`.*path="([^"]+)".*`)
+
+// ProcessPathErrors replaces "path" errors returned from the file.Resolver into unknowns,
+// and warn logs non-unknown errors, returning only the unknown errors
+func ProcessPathErrors(err error) error {
+	if err == nil {
+		return nil
+	}
+	errText := err.Error()
+	if pathErrorRegex.MatchString(errText) {
+		foundPath := pathErrorRegex.ReplaceAllString(err.Error(), "$1")
+		if foundPath != "" {
+			return New(file.NewLocation(foundPath), err)
+		}
+	}
+	unknowns, remainingErrors := ExtractCoordinateErrors(err)
+	log.Warn(remainingErrors)
+
+	var out []error
+	for _, u := range unknowns {
+		out = append(out, &u)
+	}
+	return Join(out...)
+}
--- a/syft/pkg/cataloger/binary/classifier_cataloger_test.go
+++ b/syft/pkg/cataloger/binary/classifier_cataloger_test.go
@@ -1667,8 +1667,7 @@ func Test_Cataloger_ResilientToErrors(t *testing.T) {
 	c := NewClassifierCataloger(DefaultClassifierCatalogerConfig())
 
 	resolver := &panicyResolver{}
-	_, _, err := c.Catalog(context.Background(), resolver)
-	assert.Error(t, err)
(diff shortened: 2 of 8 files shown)


spiffcs:
I think we want to keep the executor behavior and instead change the `task` behavior that is responsible for surfacing these errors that should not be causing syft to exit. Let me go a bit deeper in the calls so we don't undo all executor err reporting and find the specific change made in #2998

kzantow (approved):

internal/unknown/path_error.go
```diff
@@ -0,0 +1,33 @@
+package unknown
+
+import (
+	"regexp"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/file"
+)
+
+var pathErrorRegex = regexp.MustCompile(`.*path="([^"]+)".*`)
```
wagoodman, line 10: Can we look at the error type instead of the text like this?

wagoodman: We syned on this offline, we're going to follow up with another PR that has more broad impact/fixes

syft/pkg/cataloger/binary/classifier_cataloger_test.go
```diff
@@ -1667,8 +1667,7 @@ func Test_Cataloger_ResilientToErrors(t *testing.T) {
 	c := NewClassifierCataloger(DefaultClassifierCatalogerConfig())
 
 	resolver := &panicyResolver{}
-	_, _, err := c.Catalog(context.Background(), resolver)
-	assert.Error(t, err)
+	_, _, _ = c.Catalog(context.Background(), resolver)
```
wagoodman, line 1670: I think instead of deleting the error nil check we should check that it's either no nil, nil, or a specific value. It's true that we survived a panic but not clear what the impact to results is under this test.

wagoodman (approved):
I have the one comment about the test that should be addressed, but otherwise 💯  -- we'll do a pseudo-fast follow with a broader fix. Nice job @spiffcs @kzantow 🙌

PR #3443, opened 2024-11-14
feat: emit dependency relationships found in Cargo.lock

### LINKED ISSUE #2353: Surface Rust dependency relationships
**What happened**:
Syft not created "dependencies" in cyclonedx report for Go, Rust, Java and etc.
Cargo.lock had "dependencies". View like

```
[[package]]
name = "yaserde"
version = "0.8.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "4bf52af554a50b866aaad63d7eabd6fca298db3dfe49afd50b7ba5a33dfa0582"
dependencies = [
 "log",
 "xml-rs",
 "yaserde_derive",
]
```
But output syft doesnt reflect this
Want full cyclonedx report with "dependencies" and "depensOn" included

syft -o cyclone-dx-json@1.5 Cargo.lock > report.json

**Environment**:
Syft 0.94

willmurphyscode:
Include updating test dependency to have dependencies on multiple versions of the same crate.

# Description

Partially implements #572 by adding support for emitting dependency relationships between crates as enumerated in Cargo.lock files.

Fixes #2353.

## Type of change

This is a new feature.

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/rust/parse_cargo_lock.go
+++ b/syft/pkg/cataloger/rust/parse_cargo_lock.go
@@ -48,11 +48,15 @@ func parseCargoLock(_ context.Context, _ file.Resolver, _ *generic.Environment, 
 			newPkg,
 		)
 		newIx := len(pkgs) - 1
+		// Cargo.lock dependencies are strings that are the name of a package, if that
+		// is unambiguous, or a string like "name version" if the name alone is not
+		// ambiguous. Set both keys in the map, since we don't know which key is
+		// going to be used until we're trying to resolve dependencies. If the
+		// first key is overwritten, that means the package name was an ambiguous dependency
+		// and "name version" will be used as the key anyway.
 		keys := []string{
 			newPkg.Name,
 			fmt.Sprintf("%s %s", newPkg.Name, newPkg.Version),
-			fmt.Sprintf("%s %s", newPkg.Name, newPkg.Metadata.(pkg.RustCargoLockEntry).Source),
-			fmt.Sprintf("%s %s %s", newPkg.Name, newPkg.Version, newPkg.Metadata.(pkg.RustCargoLockEntry).Source),
 		}
 		for _, k := range keys {
 			pkgIndex[k] = newIx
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/rust/parse_cargo_lock.go
```diff
@@ -33,19 +33,42 @@ func parseCargoLock(_ context.Context, _ file.Resolver, _ *generic.Environment,
...
 		pkgs = append(
 			pkgs,
-			newPackageFromCargoMetadata(
-				p,
-				reader.Location.WithAnnotation(pkg.EvidenceAnnotationKey, pkg.PrimaryEvidenceAnnotation),
-			),
+			newPkg,
 		)
+		newIx := len(pkgs) - 1
+		keys := []string{
```
willmurphyscode, line 51: Note: right now only "name" and "name version" type keys are tested. I believe the others are possible and are included for completeness, but I haven't found a good way to tell `cargo` to make a Cargo.lock that exhibits them.

wagoodman (approved):
nice!

Just an fyi, there is a [`dependency.Specifier`](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/internal/dependency/resolver.go#L14-L35) and [processor](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/internal/dependency/resolver.go#L38) for these kinds of things (see the [dpkg dependency](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/debian/dependency.go#L12-L46) and its usage as a [cataloger post-processor](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/debian/cataloger.go#L18). 

The one difference with that specific approach and here is that we'd want to still keep dependencies from the single cargo.lock related (and not related to seemingly overlapping dependencies in other cargo.lock files), which means it's use as a post-processor across all packages found by the cataloger wouldn't be correct, but it could be used within the cataloger instead as a post-processor, thus reusing the abstraction.

No need to change this code on the PR, but could be helpful for future PRs like this one.

syft/pkg/cataloger/rust/parse_cargo_lock.go
```diff
@@ -33,19 +33,46 @@ func parseCargoLock(_ context.Context, _ file.Resolver, _ *generic.Environment,
...
+			pkgIndex[k] = newIx
+		}
+	}
+	var relationships []artifact.Relationship
+	for _, p := range pkgs {
+		meta := p.Metadata.(pkg.RustCargoLockEntry)
+		for _, d := range meta.Dependencies {
+			relationships = append(relationships, artifact.Relationship{
+				From: p,
+				To:   pkgs[pkgIndex[d]],
```
willmurphyscode, line 71: It's probably worth defending against a panic when there's a malformed Cargo.lock. Will update.

Issue #2724, opened 2024-03-19
Extract full license text


mmarseu:
**What would you like to be added**:
SBOM formats such as CycloneDX and SPDX support including the full text of a license with a component. It would be great if syft could extract this information when scanning for licenses.

**Why is this needed**:
OSS license compliance is one important use case for SBOMs, especially in large enterprises. SBOMs produced by syft today include components with licenses identified by name (not SPDX ID) which is mostly useless without the accompanying text.

Comment https://github.com/anchore/syft/issues/2002#issuecomment-1674193547 has also asked for such a feature to be implemented, however, I believe it was eventually overlooked when the corresponding issue was closed.

**Additional context**:
Example for curl produced by dpkg cataloger in CycloneDX (modify for conciseness):
```json
{
    "type": "library",
    "name": "curl",
    "version": "7.81.0-1ubuntu1.15",
    "licenses": [
        {
            "license": {
                "name": "other"
            }
        },
        {
            "license": {
                "name": "public-domain"
            }
        }
    ],
    "purl": "pkg:deb/ubuntu/curl@7.81.0-1ubuntu1.15?arch=amd64&distro=ubuntu-22.04",
    "properties": [
        {
            "name": "syft:location:0:path",
            "value": "usr/share/doc/curl/copyright"
        },
        {
            "name": "syft:location:1:path",
            "value": "var/lib/dpkg/info/curl.md5sums"
        },
        {
            "name": "syft:location:2:path",
            "value": "var/lib/dpkg/status"
        }
    ]
},
```

italvi:
Maybe a good source for licenses like `public-domain` could be the [ScanCode LicenseDB](https://scancode-licensedb.aboutcode.org/), as unfortunately SPDX [will not](https://wiki.spdx.org/view/Legal_Team/Decisions/Dealing_with_Public_Domain_within_SPDX_Files) add such an ID to their list.

tgerla:
Hi @mmarseu, thanks for the suggestion. We think it makes sense to include full license text or license snippets where available, as an opt-in configuration. We've got some more design work to do but we'll put this issue in the backlog for implementation at some point. If you're interested in working on this, let us know and we can collaborate. Thanks!

wagoodman:
dev note: we could start adding full license text, when filename/contents are detected to be licenses, or partial license text within a file. These could be persisted on ** file ** object in the SBOM, not the ** package ** object.

mmarseu:
[...]

PR #3450, opened 2024-11-17
Detect when full license text has been provided and preserve as separate field

### LINKED ISSUE #3088: Detect whether full license text or a license name has been provided
**What happened**:
I scanned an image of python, and one of the packages has a full description of the license.
here is an example:
```
        {
          "path": "/usr/local/lib/python3.9/site-packages/numpy-1.26.4.dist-info/RECORD",
          "layerID": "sha256:6baba55ee976c6469c4b3b5a3c4585320bc0da3e3c9638e2169195f7243b8d03",
          "accessPath": "/usr/local/lib/python3.9/site-packages/numpy-1.26.4.dist-info/RECORD",
          "annotations": {
            "evidence": "supporting"
          }
        }
      ],
      "licenses": [
        {
          "value": "Copyright (c) 2005-2023, NumPy Developers.\n All rights reserved.\n \n Redistribution and use in source and binary forms, with or without\n modification, are permitted provided that the following conditions are\n met:\n \n * Redistributions of source code must retain the above copyright\n notice, this list of conditions and the following disclaimer.\n \n * Redistributions in binary form must reproduce the above\n copyright notice, this list of conditions and the following\n disclaimer in the documentation and/or other materials provided\n with the distribution.\n \n * Neither the name of the NumPy Developers nor the names of any\n contributors may be used to endorse or promote products derived\n from this software without specific prior written permission.\n \n THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS\n \"AS IS\" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT\n LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR\n A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT\n OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,\n SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT\n LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,\n DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY\n THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT\n (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
```
In case license has \n inside of it, I think syft should trim it so it will not make the SBOM look weird.
Here is the Dockerfile to build a sample of an image:
```
# Use the official Python 3.9 image from Docker Hub
FROM python:3.9

# Set the working directory in the container
WORKDIR /app

# Install the specific version of NumPy
RUN pip install numpy==1.26.4

# Specify the command to run on container start
CMD ["python"]
```

**Environment**:
- Output of `syft version`: 1.9.0

spiffcs:
This PR updates the syft `License` model to include a new `FullText` field without any breaking changes to the current license behavior. We select candidates for this new field based on if the metadata being analyzed contains any new line characters. Because we still want `Value` to be populated as it is a required field I've included a default string that will be added here when `FullText` is the selected outcome for a newly constructed license.

- Fixes #3088, Partial #2724 (next step here is adding Scan Code DB to match full text against)

#### Verification

Use the following `Dockerfile` and build a test image
`docker build -t syft-3088:latest .`
```
# Use the official Python 3.9 image from Docker Hub
FROM python:3.9

# Set the working directory in the container
WORKDIR /app

# Install the specific version of NumPy
RUN pip install numpy==1.26.4

# Specify the command to run on container start
CMD ["python"]
```

Run the latest syft against this image using this branch:
`go run cmd/syft/main.go -o json syft-3088 | jq '.artifacts[] | select(.name=="numpy") | { name: .name, licenses: .licenses }'`

The large license value extracted from the package should now be listed under the field `fullText` with `value` being set to `FullText` to keep the field required and not incur any breaking changes.

## Type of change

# Checklist:


syft/pkg/license.go
```diff
@@ -12,6 +13,11 @@ import (
 	"github.com/anchore/syft/syft/license"
 )
 
+// FullTextValue is a placeholder value when license metadata is discovered to be fullText
+// A license with "FullText" in its value refers the consumer to look at the FullText field
+// This is so that we can keep Value as a required field up until syft 2.0
+var FullTextValue = "FullText"
```
wagoodman, line 19: instead of a placeholder field that is attempting to have consumers reference the full text, should we just populate this with an empty value? That is, the json schema requires the field to be present but does not state that it must not be empty. I feel that if we don't know the ID we should leave this blank, but allow the new full text field to convey any discovered information. That way we can stay compliant to syft v1 json output but still add full text.

spiffcs: When we do things like `otherLicenses` in SPDX what would we use for the license ID?

spiffcs: @wagoodman - it looks like in syft we have a couple of places where we assume value should not be `""`

This includes the constructor for creating a license set:
https://github.com/anchore/syft/blob/02f9350fa53fdc8ccb3ad85bf77f103e70e1fb23/syft/pkg/license_set.go#L48-L64

If we were to keep this `const` as a placeholder some alternatives that @willmurphyscode and I discussed could be as follows:

#### Other Options for Value

- SeeFullText
- “<First_50_Characters of file> … [TRUNCATED]”

Do we want a better name for this `const` or do we want to go back and break up the license set/spdx formatter assumptions about `value` being present?

wagoodman: I think this section will need to be updated already to account for filtering out "FullText" values. In that same way, can't we do the same for an empty value and the full text is specified in the other field?

henrysachs:
I just hit this case and maybe this could get picked up again as the new field would help me a lot :)

spiffcs:
> I just hit this case and maybe this could get picked up again as the new field would help me a lot :)

Yep! This is in the queue of a few other PR I have to clean up and get merged for the next up coming release. Apologies for the delay on this one.

spiffcs:
Updating this in relation to https://github.com/anchore/syft/pull/3366

Funsho-Agboola:
Hi team, great work on this. I have a primary issue #3780 which is about returning the full license strings in SPDX outputs instead of the LicenseRef-‹hash> identifiers as in [here](https://github.com/anchore/syft/blob/main/syft/format/internal/spdxutil/helpers/license.go). Can you consider extending this implementation to address this as well?

syft/pkg/license.go
```diff
@@ -68,15 +74,31 @@ func NewLicense(value string) License {
...
-	var spdxExpression string
-	if value != "" {
+	var (
+		spdxExpression string
+		fullText       string
+	)
+	// Check parsed value for newline character to see if it's the full license text
+	// License: <HERE IS THE FULL TEXT> <Expressions>
+	// DO we want to also submit file name when determining fulltext
+	if strings.Contains(value, "\n") {
```
wagoodman, line 84: may want to do a `value = strings.TrimSpace(value)` before this

syft/pkg/license_set.go
```diff
@@ -50,15 +50,16 @@ func (s *LicenseSet) Add(licenses ...License) {
...
-				// doesn't exist, add it
-				s.set[id] = l
-			} else if err != nil {
-				log.Trace("license set failed to add license %#v: %+v", l, err)
-			}
+		if id, merged, err := s.addToExisting(l); err == nil && !merged {
+			// doesn't exist, add it
+			s.set[id] = l
+		} else if err != nil {
+			log.Trace("license set failed to add license %#v: %+v", l, err)
```
wagoodman, line 62:
```suggestion
			log.WithFields("error", err, "license", l).Trace("failed to add license to license set")
```

syft/pkg/license.go
```diff
@@ -25,9 +25,15 @@ var _ sort.Interface = (*Licenses)(nil)
...
+// If a license is given as it's full text in the metadata rather than it's value or SPDX expression
+
+// The FullText field is used to represent this data
+// A Concluded License type is the license the SBOM creator believes governs the package (human crafted or altered SBOM)
+// The Declared License is what the authors of a project believe govern the package. This is the default type syft declares.
 type License struct {
-	Value          string
 	SPDXExpression string
+	Value          string
+	FullText       string
```
wagoodman, line 36: nit on naming: I don't think we should claim that the value is the "full text", but instead maybe just the "contents"

wagoodman (approved):

syft/pkg/license.go
spiffcs: These are good suggestions for 2.0 when we revisit the model again

PR #3452, opened 2024-11-18
Go source cataloger

### LINKED ISSUE #3451: cataloger: golang source cataloger
A cataloger that is invoked on a directory scan for the Golang Ecosystem that discovers all `**/go.mod` files. The cataloger will use the `main` module at the top of the modfile and then load all the root packages via the Golang tools utilities found here `golang.org/x/tools/go/packages`.

These root packages will be created as syft packages for the SBOM. The cataloger will also visit all of the packages in the import graph whose roots are in the inital root package list. This will allow the cataloger to surface all direct and transitive dependencies for a given `main` module. Vendored dependencies also have the additional bonus of allowing for license discovery during this method since the modules `dirPath` is surfaced during its discovery by the `x/tools/go/packages` utility

**Why is this needed**:

More accurate Golang source cataloging. Currently we only parse the flat list from `go.mod` or dependency information from discovered go binaries. This new feature will give a more complete overview of all the modules imported by a go project. Some projects can have more than one `go.mod` file which could lead to "duplicate" dependencies being discovered during this process. It's important to include in the main package discovered the root parent so that it's obvious to consumers of the SBOM that these "duplicates" belong to different `main` modules.

spiffcs:
# Description

This PR adds a new directory source cataloger that goes beyond reading declarations from the `go.mod` file.

The `go-module-source-file-cataloger` uses the `golang.org/x/tools/go/packages` library to read the "main" module from detected `go.mod` files. It then uses this main module information to load all the `root` packages for a given main module. The tool chain then visits all the packages in the import graph whose roots are packages in the discovered `root` packages.

This PR is currently WIP and needs further discussion around the following sections:
- The first draft of what relationships we want to surface from this cataloger should be included as none are implemented at the moment.
- License inclusion needs to be added. There is a good example in the `go-licenses` open source program at how to do this.

- Fixes #3451 

## Type of change

# Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -92,6 +92,7 @@ require (
 	github.com/adrg/xdg v0.5.3
 	github.com/magiconair/properties v1.8.7
 	golang.org/x/exp v0.0.0-20240719175910-8a7402abbf56
+	golang.org/x/tools v0.23.0
 )
 
 require (
--- a/internal/task/package_tasks.go
+++ b/internal/task/package_tasks.go
@@ -84,6 +84,11 @@ func DefaultPackageTaskFactories() PackageTaskFactories {
 			},
 			pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, pkgcataloging.LanguageTag, Go, Golang, "gomod",
 		),
+		newPackageTaskFactory(
+			func(cfg CatalogingFactoryConfig) pkg.Cataloger {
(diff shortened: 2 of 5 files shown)


spiffcs:
This was closed by mistake because of the vendor dirs PR. Reopened since we still have feature work on this PR

Issue #3428, opened 2024-11-10
support configuration of layer size in Syft


TimBrown1611:
**What would you like to be added**:
hi, according to this issue - https://github.com/anchore/stereoscope/issues/311
can syft now support this feature in steroscope (new tag was already released)?
**Why is this needed**:
be able to scan large layers of images
**Additional context**:

PR #3464, opened 2024-11-19
add set layer limit

tomersein:
# Description

- Fixes https://github.com/anchore/syft/issues/3428

## Type of change

# Checklist:

### DIFF
--- a/cmd/syft/internal/options/source.go
+++ b/cmd/syft/internal/options/source.go
@@ -22,13 +22,14 @@ type sourceConfig struct {
 }
 
 type fileSource struct {
-	Digests      []string `json:"digests" yaml:"digests" mapstructure:"digests"`
-	MaxLayerSize string   `json:"max-layer-size" yaml:"max-layer-size" mapstructure:"max-layer-size"`
+	Digests []string `json:"digests" yaml:"digests" mapstructure:"digests"`
 }
 
 var _ interface {
 	clio.FieldDescriber
 } = (*sourceConfig)(nil)
+
+var _ clio.PostLoader = (*imageSource)(nil)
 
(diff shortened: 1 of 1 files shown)


cmd/syft/internal/options/source.go
```diff
@@ -20,7 +22,8 @@ type sourceConfig struct {
 }
 
 type fileSource struct {
-	Digests []string `json:"digests" yaml:"digests" mapstructure:"digests"`
+	Digests      []string `json:"digests" yaml:"digests" mapstructure:"digests"`
+	MaxLayerSize string   `json:"max-layer-size" yaml:"max-layer-size" mapstructure:"max-layer-size"`
```
kzantow, line 26: Hey, @tomersein I think we should add this configuration to the `imageSource` ([see here](https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/source.go#L36C6-L38)), instead of the `fileSource`. Other than that, I think this PR is probably good. Thanks 👍

tomersein: hi @kzantow the image-souce doesn't work when i put it inside, only filesource.
does it make sense?

kzantow: You need to add a `PostLoad` hook for the `imageSource`; I think it would look like this:
```
func (c *imageSource) PostLoad() error {
	if c.MaxLayerSize != "" {
		perFileReadLimit, err := humanize.ParseBytes(c.MaxLayerSize)
		if err != nil {
			return err
		}
		stereoscopeFile.SetPerFileReadLimit(int64(perFileReadLimit))
	}
	return nil
}
```
You could ensure it's the right signature with:
```
var _ clio.PostLoader = (*imageSource)(nil)
```

kzantow (approved):
LGTM

PR #3468, opened 2024-11-20
disable package scan

### LINKED ISSUE #3128: warn instead of error if zero package catalogers are select - user might still run file metadata cataloger, for example
I am trying to scan but don't use the packages tasks. is it possible to disable it and use only files tasks?
**Why is this needed**:
want to separate tasks
**Additional context**:

tomersein:
# Description
this PR enables the way to scan only using files cataloger without package cataloger
related to an older request I've submitted 3128 (link below)

- Fixes #3128

## Type of change

# Checklist:

### DIFF
--- a/syft/create_sbom_config.go
+++ b/syft/create_sbom_config.go
@@ -7,6 +7,7 @@ import (
 	"runtime/debug"
 	"strings"
 
+	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/internal/task"
 	"github.com/anchore/syft/syft/cataloging"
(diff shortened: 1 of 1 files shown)


spiffcs:
Linked PR from @wagoodman. This is the direction we probably want to go where we elevate the file catalogers to be selected in the same way package catalogers are being used.

https://github.com/anchore/syft/pull/3505

TLDR we don't want to hide this and have it happen implicitly when `finalTasks` is  0.

@tomersein do you mind giving your feedback on 3505?

willmurphyscode (approved):

PR #3484, opened 2024-11-27
feat: add cataloger for NuGet packages

### LINKED ISSUE #373: Support cataloging NuGet packages

Kemosabert:
# Description

This PR adds a cataloger for NuGet lockfiles in the form of packages.lock.json files.

- Fixes #373 

## Type of change

# Checklist:


syft/pkg/cataloger/dotnet/parse_dotnet_packages_lock.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/generic"
+)
+
+var _ generic.Parser = parseDotnetPackagesLock
+
+type dotnetPackagesLock struct {
+	Version      int                                         `json:"version"`
+	Dependencies map[string]map[string]dotnetPackagesLockDep `json:"dependencies"`
```
spiffcs, line 22: Comment for other reviewers - the reason this looks this way is that the dependencies have this nested map quality where there are sub fields. @Kemosabert do you have an example where we can see how a sibling to the top level package would behave? In the bottom example it would be a sibling package to `netcore2.0` - I pulled this from [here](https://github.com/NuGet/Home/wiki/Enable-repeatable-package-restore-using-lock-file)


```
  "dependencies": {	
    "netcore2.0": {	
      "Contoso.Base": {	
        "type": "direct",	
        "requested": "3.0.0",	
        "resolved": "3.0.0",
        "contentHash":"fVXsnMP2Wq84VA533zj0a/Et+QoLoeNpVXsnMP2Wq84l+hsUxfwunkbqoIHIvpOqwQ/+HIvprVKs+QOihnkbqod=="
        "dependencies": {
             "Contoso.Core": "1.2.3",
             "Fabrikam.Utilities": "[3.1.0]"
         }		
      }	
      "Contoso.Core": {
        "type": "transitive",	
        "requested": "1.2.3",	
        "resolved": "1.2.3",
        "contentHash":"xScnMP2Wq84VA533zj0a/Et+QoLoeNpVXsnMP2Wq84l+hsUxfwunkbqoIHIvpOqwQ/+HIvprVKs+QOihnkbmoq=="
        "dependencies": {
           ...
           ...
         }
```

syft/pkg/cataloger/dotnet/parse_dotnet_packages_lock.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// collect all deps here
+	allDependencies := make(map[string]dotnetPackagesLockDep)
+
+	var names []string
+	for _, dependencies := range lockFile.Dependencies {
+		for name, dep := range dependencies {
+			names = append(names, name)
+			allDependencies[name] = dep
```
spiffcs, line 53: What if the map has multiple entries here for the same name key? Are we overriding the `allDependencies` entry?

syft/pkg/cataloger/dotnet/parse_dotnet_packages_lock.go
```diff
@@ -0,0 +1,130 @@
...
+			pkgs = append(pkgs, *dotnetPkg)
+			pkgMap[name] = *dotnetPkg
+		}
+	}
+
+	// fill up relationships
+	for name, dep := range allDependencies {
+		parentPkg, ok := pkgMap[name]
+		if !ok {
+			log.Debugf("unable to find package in map: %s", name)
```
spiffcs, line 74: I think this might need a better abstraction for this debug message to be helpful. Which map? What does the pkgMap represent? Are allDependencies only inclusive of `dotNetPkgs` or can they be other types?

Apologies for all the question on this review. Is there a good reference or specification document I can look at for the lockfile so it's easier to maybe build a mental model of what we're cataloging here? That might aid in the review and help me understand if the relationships being created are correct.

From what I am reading the lockfile is:
```
dependencies[topPackages]map[subPackages]metadata

where topPackages and subPackages can be > 1
```

Is pkg map just inclusive of `topPackages`?

syft/pkg/cataloger/dotnet/test-fixtures/packages.lock.json
```diff
@@ -0,0 +1,55 @@
+{
+    "version": 1,
+    "dependencies": {
+        "net8.0": {
```
spiffcs, line 4: I think this is where my understanding goes sideways. 
Is the entry at the top of this map (net8.0) always singular, and where (if anywhere) does the metadata come from that describes `net8.0`

spiffcs (commented):
Thanks for the PR @Kemosabert! I hope you don't mind I added some questions to get a better handle on the package lock specification.

syft/pkg/cataloger/dotnet/test-fixtures/packages.lock.json
Kemosabert: As I understand it, `net8.0` is the runtime target version for the packages listed as its siblings. The packages.lock.json file just lists all of these and does not say which one to use like the .deps.json file does.

syft/pkg/cataloger/dotnet/parse_dotnet_packages_lock.go
Kemosabert: As mentioned below, `.netcore2.0` is the runtime target version for the deps. I am not quite sure how to let Syft describe this in the SBOM since it's not really a package, its just to tell nuget which packages to install for the user's runtime of .NET.

My first thought would be to let the SBOM include all packages from all runtime versions, and show the runtime versions where they appear in either in the locations or in the metadata. What do you think?

syft/pkg/cataloger/dotnet/parse_dotnet_packages_lock.go
Kemosabert: I changed it so it tracks by name and version of package. Overwrites in that case should be rare, but fine

syft/pkg/cataloger/dotnet/parse_dotnet_packages_lock.go
Kemosabert: I went mainly off off this document here: https://github.com/NuGet/Home/wiki/Enable-repeatable-package-restore-using-lock-file and looking at lockfiles that I could generate myself with examples.

I updated the debug message and the logic as well. The test packages.lock.json file is one I generated using dotnet commands and includes a few examples of how it handles conflicting dependencies and versions.

Kemosabert:
@spiffcs would you mind to have another look at this when you have some time, happy to hear more suggestions/feedback 👍

spiffcs:
Thanks @Kemosabert - I'm traveling this afternoon, but put this back at the top to add some commits for the SA/Unit tests that are failing. The changes look good and I appreciate the link to documentation showing what you followed for implementation!

wagoodman (approved):

PR #3500, opened 2024-12-05
Add relationships for rust audit binary packages

wagoodman:
# Description

Adds relationships for packages found by the rust audit binary cataloger. This also replaces the existing test fixture with one that is reproducible locally and more realistic.

- Partially addresses #572

## Type of change

# Checklist:


syft/pkg/cataloger/rust/parse_audit_binary.go
```diff
@@ -61,3 +65,64 @@ func parseAuditBinaryEntry(reader unionreader.UnionReader, filename string) ([]r
...
 }
+
+// auditPkgPair is a helper struct to track the original index of the package in the original audit report + the syft package created for it
+type auditPkgPair struct {
+	pkg     *pkg.Package
+	rustPkg rustaudit.Package
+	index   int
+}
+
+func processAuditVersionInfo(location file.Location, versionInfo rustaudit.VersionInfo) ([]pkg.Package, []artifact.Relationship) {
```
wagoodman, line 76: I would have loved to have used the `dependency.Resolve` but this requires material that is not persisted to a package, thus, cannot be used.

syft/pkg/cataloger/rust/test-fixtures/image-audit/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "hello_world"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+argh = "0.1"
```
willmurphyscode, line 7: When I saw this name in test fixtures, I thought it was made up, but it's real: https://crates.io/crates/argh

What a great name for an arg parsing lib 😄

syft/pkg/cataloger/rust/parse_audit_binary.go
willmurphyscode: I didn't quite understand why this is true. Can you explain what information is missing, and maybe add this as a comment in the file so future people working in here don't try to factor into dependency.Resolve?

willmurphyscode (commented):
The only change I think we need is a comment in code about why dependency.Resolve couldn't be used.

syft/pkg/cataloger/rust/test-fixtures/image-audit/Cargo.toml
wagoodman: right?!

syft/pkg/cataloger/rust/parse_audit_binary.go
wagoodman: dependency.Resolve requires a Specifier function, which takes only a package. We tend to take information from the pkg.Metadata, however, in this case, the `rustaudit.VersionInfo` data structure is used to determine what the dependencies are which is not in the package metadata. I could add dependency indexes to the metadata, but that would be confusing since it is only useful if you have the original `[]rustaudit.Package` list (since they are indexes and not names). I probably shouldn't try and craft the name/name-version/name-version-source strings to do this lookup since that's not how they are referenced in the underlying material.

willmurphyscode (approved):

PR #3502, opened 2024-12-06
fix order of rust dependencies and support git sources in Cargo.lock dependencies

willmurphyscode:
# Description

This PR has two related effects:

1. It fixes a bug where the dependency-of relationship in the Cargo.lock cataloger was previously reversed
2. It implements resolving Cargo.lock dependencies when crates of the same name and version but from different sources (e.g. from git vs from crates.io) are present in the Cargo.lock.

## Type of change

I'm considering this a bug fix, not a feature, because the previous Cargo.lock relationships PR intended to be correct and complete, and just missed these two things.

# Checklist:


wagoodman (approved):
💯

Issue #2093, opened 2023-09-04
SPDX FileName should not be absolute


vargenau:
**What happened**:

Syft generates the following SPDX (tag:value):

```
FileName: /usr/share/ca-certificates/mozilla/Actalis_Authentication_Root_CA.crt
SPDXID: SPDXRef-File-...Actalis-Authentication-Root-CA.crt-b2e28e6876228bbb
FileType: TEXT
FileChecksum: SHA1: 511ca95607022a99ed8e68bd63f136c4854cefcb
LicenseConcluded: NOASSERTION
FileComment: layerID: sha256:3f946b95045046b182ad195bfdb24fe56dd6ea12d34e35a0995218d22c05102a
```

[tools-python](https://github.com/spdx/tools-python) complains that it is invalid SPDX.

```
file name must not be an absolute path starting with "/", but is: /usr/share/ca-certificates/mozilla/Actalis_Authentication_Root_CA.crt
```

The SPDX spec says: "A relative filename".

**What you expected to happen**:

Have a relative path in FileName.

**Steps to reproduce the issue**:

```
syft docker:bitnami/mongodb -o spdx-tag-value > bitnami-mongodb.spdx
```

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`: syft 0.89.0
- OS (e.g: `cat /etc/os-release` or similar): macos 13.5.1

m-dhana:
I am also failing similar error, when I run the sbom json file through Ntia-checker. 

As per the spdx specs, [https://spdx.github.io/spdx-spec/v2.3/file-information/#81-file-name-field](https://spdx.github.io/spdx-spec/v2.3/file-information/#81-file-name-field), the filename field - "In general, every filename is preceded with a ./"

I am using the syft 0.87.1.

kzantow:
Just a note: regardless if there are changes to be made in Syft, it looks like both of these tools are incorrectly validating this field. The spec says:

> Identify the full path and filename that corresponds to the file information in this section. The metadata for the file name field is shown in Table 36.
>
> Table 36 — Metadata for the file name field
>
> ```Attribute	Value
> Required	Yes
> Cardinality	1..1
> Format		A relative filename with the root of the package archive or directory.
>			In general, every filename is preceded with a ./, see http://www.ietf.org/rfc/rfc3986.txt for syntax.
> ```

Nowhere in this text does it say that the filename is _required_ to be a relative path starting with `./`.

vargenau:
[...]

PR #3509, opened 2024-12-09
feat: convert spdx absolute to relative

spiffcs:
# Description

@popey I branched this off of my other SPDX work going into the plug fest submissions.

This is ONLY the change for `absolute --> relative` paths for the spdx format model. Some of the other format validations and NTIA minimum work is still in progress.

## Type of change

Bug Fix - SPDX documents were being flagged as invalid:
```
file name must not be an absolute path starting with "/", but is: /.github/actions/bootstrap/action.yaml
file name must not be an absolute path starting with "/", but is: /.github/workflows/benchmark-testing.yaml
file name must not be an absolute path starting with "/", but is: /.github/workflows/dependabot-automation.yaml
file name must not be an absolute path starting with "/", but is: /.github/workflows/oss-project-board-add.yaml
```

This change updates file paths in the format model logic to be relative to the `/` directory.

# Checklist:


vargenau:
Fixes https://github.com/anchore/syft/issues/2093

Issue #3441, opened 2024-11-13
SPDX expressions are lost from CycloneDX if they contain extra parenthesis


pasieronen:
**What happened**:

Given a very minimal CycloneDX SBOM as input:

```
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.6",
  "components": [
    {
      "type": "library",
      "name": "one",
      "licenses": [{"expression": "BSD-3-Clause OR MIT"}]
    },
    {
      "type": "library",
      "name": "two",
      "licenses": [{"expression": "(BSD-3-Clause OR MIT)"}]
    }
  ]
}
```

Note that in the input, component "two" has extra parenthesis around the SPDX expression (which are allowed by the SPDX spec, as far as I can tell).

Running syft SBOM cataloger and outputting to CycloneDX:

`syft scan file:./test.cdx.json --output=cyclonedx-json --select-catalogers "+sbom-cataloger"`

**What you expected to happen**:

I'd expect both components to have licenses  in the output. But what happens is that component "one" has the expected license, but component "two" does not have a license at all.

Interestingly enough, if I use `--output=json`, it looks like both components have licenses....

**Environment**:

- Output of `syft version`:

```
Application: syft
Version:    1.16.0
BuildDate:  2024-11-04T22:29:33Z
GitCommit:  8a41d772509d37267a65e0b425808e883e4b9dce
GitDescription: v1.16.0
Platform:   darwin/arm64
GoVersion:  go1.22.8
Compiler:   gc
```

- OS (e.g: `cat /etc/os-release` or similar): MacOS 14.7.1

spiffcs:
Thanks @pasieronen! I've reproduced this on my local and have picked this bug up to fix ASAP.

willmurphyscode:
Fixed by https://github.com/anchore/syft/pull/3517

PR #3517, opened 2024-12-10
fix: stop omitting redundantly parenthesized licenses in CDX formatter

willmurphyscode:
Previously, a bug in the formatter would cause SPDX expressions that were surrounded in redundant parentheses to be dropped instead of normalized.

# Description

- Fixes # https://github.com/anchore/syft/issues/3441

## Type of change

# Checklist:

### DIFF
--- a/syft/license/license_test.go
+++ b/syft/license/license_test.go
@@ -21,6 +21,11 @@ func TestParseExpression(t *testing.T) {
 			want:       "MIT OR Apache-2.0",
 		},
 		{
+			name:       "accept redundant parentheses",
+			expression: "(MIT OR Apache-2.0)",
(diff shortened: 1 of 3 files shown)


syft/license/license_test.go
```diff
@@ -20,6 +20,11 @@ func TestParseExpression(t *testing.T) {
 			expression: "MIT OR Apache-2.0",
 			want:       "MIT OR Apache-2.0",
 		},
+		{
+			name:       "accept redundant parentheses",
```
willmurphyscode, line 24: This test case was never failing on main, but it was the first place I looked, so adding it seems worthwhile to me.

PR #3522, opened 2024-12-12
feat(golang): add license parsing from vendor dirs

dschmidt:
# Description

This PR adds options to parse license information from local vendor dirs as they have a slightly different structure than mod cache dirs.

## Type of change

# Checklist:


spiffcs:
@dschmidt I kicked off the build and CI tooling for this PR.

We have a related PR where we parse the entire go source tree and will be able to pull license data from this. I think this would precludes the need for the vendor directory, but am interested in how these two might overlap.

https://github.com/anchore/syft/pull/3452

dschmidt:
Thanks! 

What's the state of your PR? Should I be able to test it? 
If it solves my use case I'm happy to wait for that one. It's not superurgent, I just need a solution in the long run. 

Forgot to check in the test fixtures... will fix tomorrow.

dschmidt:
Added the missing fixtures, can you approve the workflows again?

dschmidt:
> We have a related PR where we parse the entire go source tree and will be able to pull license data from this. I think this would precludes the need for the vendor directory, but am interested in how these two might overlap.
> 
> #3452

I'm pretty new to syft (and trying to make it work in a project I'm working on), so I'm not so sure how everything is supposed to work:
Can I use the new source file cataloger when analyzing a docker image? (with the source code on the host, not inside the image)
How would I invoke syft to do that?

Without a docker image I've tried
`syft scan --select-catalogers "+go-module-source-file-cataloger" . -ospdx-json | jq`
on a smaller project I'm working on and don't see license information from the vendor dir.

in github.com/owncloud/ocis the same command crashes with
```
 ✔ Indexed file system                                                                                                                                                                                             .
 ⠸ Cataloging contents             ━━━━━━━━━━━━━━━━━━━━                                                                                             cdb4ee2aea69cc6a83331bbe96dc2caa9a299d21329efb0336fc02a82e1839a8
   └── ⠸ Packages                        [2,531 packages]  
         └── ⠙ Go module source file         cataloger                       
[0046]  WARN found package with empty ID while adding to the collection: Pkg(name="prometheus" version="v0.4.2" type="go-module" id="")
runtime: goroutine stack exceeds 1000000000-byte limit
                                                      runtime: sp=0xc0575aa598 stack=[0xc0575aa000, 0xc0775aa000]
                                                                                                                 fatal error: stack overflow

                                                                                                                                            runtime stack:
                                                                                                                                                          runtime.throw({0x276e858?, 0x7fab857f9d50?})
                                                                                                                                                                                                        /usr/lib64/go/1.23/src/runtime/panic.go:1067 +0x48 fp=0x7fab857f9d10 sp=0x7fab857f9ce0 pc=0xdda568
                                                                                    runtime.newstack()
                                                                                                        /usr/lib64/go/1.23/src/runtime/stack.go:1117 +0x5bd fp=0x7fab857f9e50 sp=0x7fab857f9d10 pc=0xdbc19d
[…]
```

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -105,23 +120,21 @@ func (c *goLicenseResolver) getLicenses(ctx context.Context, scanner licenses.Sc
 	return toPkgLicenses(goLicenses), err
 }
 
-func (c *goLicenseResolver) getLicensesFromLocal(ctx context.Context, scanner licenses.Scanner, moduleName, moduleVersion string) ([]goLicense, error) {
-	if c.localModCacheDir == nil {
+func (c *goLicenseResolver) getLicensesFromLocal(ctx context.Context, scanner licenses.Scanner, moduleDir fs.FS, moduleSubdir string) ([]goLicense, error) {
```
kzantow, line 123: Instead of overloading this function to support local go mod directories and vendor directories, I think there should be a different function to provide a different file URL, since this is NOT searching `$GOPATH/pkg/mod`. Since it's required to specify a vendor dir, we could use that. Or at least something indicating it came from a vendor directory; something like this:
```go
func (c *goLicenseResolver) getLicensesFromLocalVendor(ctx context.Context, scanner licenses.Scanner, moduleName string) ([]goLicense, error) {
	if c.localVendorDir == nil {
		return nil, nil
	}
	dir, err := fs.Sub(c. localVendorDir, moduleDirVendor(moduleName))
	if err != nil {
		return nil, err
	}

	// if we're running against a directory on the filesystem, it may not include the
	// user's homedir or other sensitive information, so we don't include the base directory
	return c.findLicensesInFS(ctx, scanner, "file://$GO_VENDOR_DIR/"+subdir+"/", dir)
}
```

kzantow (commented):
This PR looks very close -- there is one small change I'd like to see and it needs to be rebased; I could take care of this if you like.

dschmidt:
> This PR looks very close -- there is one small change I'd like to see and it needs to be rebased; I could take care of this if you like.

I'd be happy if you could take over!
Feel free to push to my fork/this PR, if you're allowed to or send a successor PR otherwise.

kzantow:
Hi @dschmidt, sorry I seem to have missed the question you posted above 🤦 :
> Can I use the new source file cataloger when analyzing a docker image? (with the source code on the host, not inside the image)
> How would I invoke syft to do that?
>
> Without a docker image I've tried
> `syft scan --select-catalogers "+go-module-source-file-cataloger" . -ospdx-json | jq`

Given this PR is doing what the question asks, I think you already know the answer. TL;DR: syft does not scan anything outside of what you tell it to scan by default and has to get specialized handling for each task as is done in this PR.

I made a few tweaks to resolve conflicts. I'm impressed by how well you adhered to the existing code patterns! Thanks for the contribution 👍

kzantow (approved):

PR #3539, opened 2024-12-18
feat: syft 3435 - add file components to cyclonedx bom output when file metadata is available

### LINKED ISSUE #3435: cyclone-dx presenter drops files, includes only packages
**What happened**:
export SYFT_FILE_METADATA_SELECTION="all"

syft scan "${REPO_PATH}" -o cyclonedx-json > sbom.json
syft scan "${REPO_PATH}" -o spdx-json > sbom.json

for -o cyclonedx-json I expect that all files of the directory are in the sbom

syft scan "${REPO_PATH}" -o cyclonedx-json > sbom.json
syft scan "${REPO_PATH}" -o spdx-json > sbom.json

with -o spdx-json the sbom filled up

cat sbom.json
{"$schema":"http://cyclonedx.org/schema/bom-1.6.schema.json","bomFormat":"CycloneDX","specVersion":"1.6","serialNumber":"urn:uuid:d9f32702-f7d9-44a4-bd21-7b02f4c2ff67","version":1,"metadata":{"timestamp":"2024-11-11T19:10:42+01:00","tools":{"components":[{"type":"application","author":"anchore","name":"syft","version":"1.16.0"}]},"component":{"bom-ref":"c89118b3fe999aab","type":"file","name":"/***********************************"}}}

**Environment**:
- Output of `syft version`:  1.16.0
- OS (e.g: `cat /etc/os-release` or similar): Linux and macos

spiffcs:
# Description
This change updates syft to add cyclone-dx components with the type `file` when the underlying SBOM has coordinates that are mapped to their respective file metadata.

- Fixes #3435 

### Example
```
export SYFT_FILE_METADATA_SELECTION="all"
go run cmd/syft/main.go -o cyclonedx-json dir:../stereoscope/ > sbom.json
```

Slice of new file types from output
```
    {
      "bom-ref": "7a9779339fe3868f",
      "type": "file",
      "name": "/Users/hal/development/stereoscope/.DS_Store",
      "hashes": [
        {
          "alg": "SHA-1",
          "content": "c918669c6ac5631bee00ceaef5f2e8a023945697"
        },
        {
          "alg": "SHA-256",
          "content": "3617159d3c3333bda8497d3c60fdda46f5ce0ab72bf3ba772dbb9128e0aca2d5"
        }
      ]
    },
    {
      "bom-ref": "ec6f846aad26adea",
      "type": "file",
      "name": "/Users/hal/development/stereoscope/.binny.yaml",
      "hashes": [
        {
          "alg": "SHA-1",
          "content": "41928b763c5478f3bc3ddde1135b0ece1536ff86"
        },
        {
          "alg": "SHA-256",
          "content": "fb6962a7f181dd5bb3e74ba637b83e91800197863b7d56cb53a6a6b96cb78668"
        }
      ]
    },
```

## Type of change

# Checklist:


syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -28,12 +29,42 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
...
+	artifacts := s.Artifacts
+	coordinates := s.AllCoordinates()
+	fileComponents := make([]cyclonedx.Component, len(coordinates))
+	for i, coordinate := range coordinates {
+		var metadata *file.Metadata
+		// File Info
+		fileMetadata, exists := artifacts.FileMetadata[coordinate]
+		// no file metadata then don't include in SBOM
+		if !exists {
+			continue
```
willmurphyscode, line 50: On line 43 above, we make a slice with the same list as `s.AllCoordinates()`, but here we sometimes skip coordinates. I think this may leave unitialized file components in the output. When do we expect `exists` to be false?

spiffcs: That's a good question - let me update this code and investigate the config state where metadata selection is none.

My understanding is there would still be coordinates, but no associated metadata, but that might not be the case and `coordinates := s.AllCoordinates()` would just be nothing.
```
file:
  metadata:
    # select which files should be captured by the file-metadata cataloger and included in the SBOM.
    # Options include:
    #  - "all": capture all files from the search space
    #  - "owned-by-package": capture only files owned by packages
    #  - "none", "": do not capture any files (env: SYFT_FILE_METADATA_SELECTION)
    selection: 'none'
```

syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -28,12 +29,42 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
...
 	components = append(components, toOSComponent(s.Artifacts.LinuxDistribution)...)
+
+	// Files
+	artifacts := s.Artifacts
+	coordinates := s.AllCoordinates()
+	fileComponents := make([]cyclonedx.Component, len(coordinates))
+	for i, coordinate := range coordinates {
+		var metadata *file.Metadata
+		// File Info
+		fileMetadata, exists := artifacts.FileMetadata[coordinate]
```
willmurphyscode, line 47: Do we want directories in the output? I don't know the answer. I can't think of why we would. In the current implementation the show up like this:

``` json
    {
      "bom-ref": "ab68ef0e1832e438",
      "type": "file",
      "name": "/"
    }
```

That doesn't tell anyone a lot; it just looks like a file with no digests. In contrast, a regular file will have content digests:

``` json
    {
      "bom-ref": "6185b4b8a7b64f56",
      "type": "file",
      "name": "/bin/[",
      "hashes": [
        {
          "alg": "SHA-1",
          "content": "912faeca732392cd21175ae53ae49624da034f1c"
        },
        {
          "alg": "SHA-256",
          "content": "25015cc97808781979490c4843c4a483019ec5efc0ecfae648c7fd4f36d18096"
        }
      ]
    },
```

Syft JSON has both files and directories, but the metadata says it's a directory, whereas here it just looks like an incomplete file. The spec has a type `file` but no type directory: https://cyclonedx.org/docs/1.6/json/#components_items_type

I'm not sure.

syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -44,6 +75,33 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 	return cdxBOM
 }
 
+func digestsToHashes(digests []file.Digest) *[]cyclonedx.Hash {
+	hashes := make([]cyclonedx.Hash, len(digests))
+	for i, digest := range digests {
+		cdxAlgo := toCycloneDXAlgorithm(digest.Algorithm)
```
willmurphyscode, line 81: Do we need to handle a map miss here? What should happen if there's a digest for an algorithm outside CDX spec? For example, what if someone crafts an SBOM with xx64 hashes of files? (I'm not sure any current spec allows this, but I think right now we would end up with a blank algorithm field, which is not allowed: https://cyclonedx.org/docs/1.6/json/#components_items_hashes_items_alg

spiffcs: Yep! That's a good point. I was only thinking about syft generated SBOMs who's digest algorithm are a subset of the current cdx specification.

syft/format/common/cyclonedxhelpers/to_format_model.go
spiffcs: I think I'll air on the side of dropping these for now since it's a little ambiguous. I'll try to look for examples of others who might be including/dropping for comparison.

spiffcs: Dev note: we should probably be dropping directories from other SBOM formats if we go about not including them in this change as well. It should be one way or the other for all formats.

syft/format/common/cyclonedxhelpers/to_format_model_test.go
```diff
@@ -143,6 +144,52 @@ func Test_relationships(t *testing.T) {
...
+func Test_fileComponents(t *testing.T) {
+	tests := []struct {
+		name string
+		sbom sbom.SBOM
+		want []cyclonedx.Component
+	}{
+		{
+			name: "sbom coordinates with file metadata are serialized to cdx",
+			sbom: sbom.SBOM{
+				Artifacts: sbom.Artifacts{
```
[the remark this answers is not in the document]
spiffcs: I think I've addressed all the above cases. Let me know if there is anything else you'd like to see 😄

syft/format/common/cyclonedxhelpers/to_format_model.go
spiffcs: Directories are now dropped.

syft/format/common/cyclonedxhelpers/to_format_model.go
spiffcs: I've update this to just be a regular append to account for where we have different amounts of metadata vs `AllCoordinates`

syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -28,12 +42,51 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
...
+		}
+		metadata = &fileMetadata
+
+		// Digests
+		var digests []file.Digest
+		if digestsForLocation, exists := artifacts.FileDigests[coordinate]; exists {
+			digests = digestsForLocation
+		}
+
+		// if cdx doesn't have an algorithm for the SBOM we need to drop the components
```
willmurphyscode, line 80: I think this comment is outdated.

syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -44,6 +97,29 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
...
+		}
+		hashes = append(hashes, cyclonedx.Hash{
+			Algorithm: cdxAlgo,
+			Value:     digest.Value,
+		})
+	}
+	return hashes
+}
+
+// Not pulling it's weight, let's just do strings to lower on the map
```
willmurphyscode, line 116: I think this comment wasn't meant to make the final cut :)

willmurphyscode (approved):

PR #3558, opened 2025-01-02
Use reader when scanning for package versions over reading entire binary into memory

wagoodman:
# Description

Today the golang binary cataloger as well as the binary classifier cataloger both read the entire target binary into memory in order to search for the application version. This is mostly because of limitations with the stdlib not accepting a reader for regexes. This PR adds new reader-based implementation to search for N many regex matches against overlapping chunks of the file. The default chunk size is 1MB and overlaps 500KB from the previous chunk when searching; this does mean that we're searching across more bytes that before since we search overlapping areas twice, but I could not find a measurable impact with the samples used. I presume that there are pessimistic cases where the target version string is at the end of the file being searched and would take more time to find when compared to todays implementation -- this seems like the right trade off.

Take this container image for example:
```dockerfile
FROM traefik:1.7.34@sha256:d7bacaf9ca8d59b0e7c304920629d98bb98a545127ca4b10e16c8b252b9351b9 as traefik-base

FROM alpine:3.18
COPY --from=traefik-base /traefik /app/traefik

# make several useless copies...
RUN for i in $(seq -w 1 5); do \
      mkdir -p /app/$i && \
      cp /app/traefik /app/$i/traefik; \
    done
```

Syft today requires ~875 MB of memory to scan this image:
Screenshot 2025-01-02 at 1 20 56 PM

With the adjustments in this PR syft only needs ~4 MB (in fact is not even captured in the pprof samples):
Screenshot 2025-01-02 at 2 25 19 PM

While working on this change (and when there were smaller chunk sizes in use at the time) I noticed that the redis-server patterns were not always detected in all samples. I made an attempt here to specify different patterns that did not require matching across such large boundaries.

- Fixes

## Type of change

# Checklist:


internal/regex_helpers.go
```diff
@@ -32,6 +38,99 @@ func MatchNamedCaptureGroups(regEx *regexp.Regexp, content string) map[string]st
...
+	halfChunkSize := chunkSize / 2
+	prevChunk := make([]byte, 0, halfChunkSize)
+	currentChunk := make([]byte, chunkSize)
+
+	for {
+		// read the next chunk
+		n, err := buf.Read(currentChunk)
+		if n > 0 {
+			// combine the last half of the previous chunk with the current chunk
+			combined := make([]byte, len(prevChunk)+n)
```
kzantow, line 100: Since this PR is about reducing allocations, I would suggest that there is a ring-buffer-like method that could be used to avoid all the secondary allocations could be avoided and this would only require one copy per iteration. Something sorta like this (it's untested and probably wrong):
```
	chunkSize := 10
	half := chunkSize / 2
	bufSize := chunkSize + half
	buf := make([]byte, bufSize)

	lastRead := 0
	for {
		offset := half
		if lastRead < half {
			offset = lastRead
		}
		start := half - offset
		if lastRead > 0 {
			copy(buf[start:], buf[half+offset:half+lastRead])
		}
		n, err := rdr.Read(buf[half:])
		if err != nil {
			break
		}
		// read buf[start:half+n] 
		lastRead = n
	}
```

kzantow (approved):
I don't see anything necessarily wrong aside from the obvious "this won't work if the regex needs to read more than 1.5x the buffer size" problem. But we really shouldn't need to make regexes that require 1 MB -- we're probably looking at the wrong bytes if so! I left a suggestion for an approach I think would be more efficient by reducing allocations and managing instead the indexes into a single existing array.

I don't know if `go` does anything similar by default, but we could/should also look at using buffer pools for this stuff.

Is it worth investigating if an existing streaming regex library can help here? e.g. https://github.com/tonymet/regexpscanner

internal/regex_helpers.go
wagoodman: much better! I'm incorporating now 👍

wagoodman:
> we could/should also look at using buffer pools for this stuff

yeah, that'll be the next iteration of performance gains here, this is good enough for now.

> worth investigating if an existing streaming regex library can help here?

I found a couple but the main problem is that we're searching for multiple regexes at a time and the implementations I found only support one regex at a time.

PR #3572, opened 2025-01-09
fix: fetch Dart package versions from sdk entries

### LINKED ISSUE #3158: Dart: Syft incorrectly generates SBOM with version 0.0.0 for SDK dependencies
**What happened**:

Trivy tool had the same issue which was [raised here](https://github.com/aquasecurity/trivy/discussions/5984).

We are using syft to generate SBOMs for our Dart/Flutter project.
Syft incorrectly generates a dependency for flutter@0.0.0, ignoring the sdk setting.

Corresponding part of the pubspec.lock:

```yaml
…
flutter:
    dependency: "direct main"
    description: flutter
    source: sdk
    version: "0.0.0"
…
```

As you can see it specifies source as sdk.

Further down in the pubspec.lock file there is more information on the sdk version:

```yaml
sdks:
  dart: ">=2.15.1 <3.0.0"
  flutter: ">=2.8.0"

```

The generated SBOM should not contain the incorrect (but unfortunately valid) version 0.0.0.

Our project code is private, but the problem exists in any Flutter project.

You can reproduce the problem easily by running `syft scan dir:./ -o spdx-json=sbom.spdx.json` on this project: 

https://github.com/KhoaSuperman/findseat/

One more thing, **dart** ain't been added to the SBOM either. So it is not only flutter that fails due to the actuall version been in the `sdks` zone but also dart is not being captured.

**Environment**:
- Output of `syft version`: 1.11.1
- OS (e.g: `cat /etc/os-release` or similar): Sonoma 14.1

sgreg:
# Description

Packages that are provided by an SDK, mainly Flutter, will have their version set to 0.0.0 in Dart's pubspec.lock file. Their actual version is linked to that SDK, which is defined either as a version range or a minimum supported version, rather than an explicit, single version.

The pubspec.lock file has a dedicated section to define those SDK version range constraints, which is already stored internally when parsing the file itself. The solution now is to look up such a package's SDK name, retrieve the defined version range / lower version boundary, and set the minimum supported version as the package's new version.

- Fixes #3158 

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/dart/parse_pubspec_lock.go
+++ b/syft/pkg/cataloger/dart/parse_pubspec_lock.go
@@ -7,9 +7,9 @@ import (
 	"regexp"
 	"sort"
 
+	"github.com/Masterminds/semver"
 	"gopkg.in/yaml.v3"
 
-	"github.com/Masterminds/semver"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/internal/unknown"
 	"github.com/anchore/syft/syft/artifact"
@@ -85,12 +85,9 @@ func parsePubspecLock(_ context.Context, _ file.Resolver, _ *generic.Environment
 			if err != nil {
 				log.Tracef("failed to resolve %s SDK version for package %s: %v", sdkName, name, err)
 				continue
-
-			} else {
-				log.Tracef("resolved %s SDK version for package %s to %s", sdkName, name, sdkVersion)
-				pkg.Version = sdkVersion
-				p.Packages[name] = pkg
 			}
+			pkg.Version = sdkVersion
+			p.Packages[name] = pkg
 		}
 
 		names = append(names, name)
@@ -168,7 +165,7 @@ func parseMinimumSdkVersion(constraint string) (string, error) {
 	}
 
 	// Read "version" subexpression (see 2. above) into version variable
-	version := []byte{}
+	var version []byte
 	matchIndex := re.FindStringSubmatchIndex(constraint)
 	version = re.ExpandString(version, "$version", constraint, matchIndex)
 
--- a/syft/pkg/cataloger/dart/parse_pubspec_lock.go
+++ b/syft/pkg/cataloger/dart/parse_pubspec_lock.go
@@ -71,8 +71,6 @@ func parsePubspecLock(_ context.Context, _ file.Resolver, _ *generic.Environment
 
 	var names []string
 	for name, pkg := range p.Packages {
-		names = append(names, name)
-
 		if pkg.Source == "sdk" && pkg.Version == "0.0.0" {
 			// Packages that are delivered as part of an SDK (e.g. Flutter) have their
 			// version set to "0.0.0" in the package definition. The actual version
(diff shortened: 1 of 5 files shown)


syft/pkg/cataloger/dart/parse_pubspec_lock.go
```diff
@@ -68,8 +70,28 @@ func parsePubspecLock(_ context.Context, _ file.Resolver, _ *generic.Environment
...
+			// should refer to the SDK version, which is defined in a dedicated section
+			// in the pubspec.lock file and uses a version range constraint.
+			//
+			// If such a package is detected, look up the version range constraint of
+			// its matching SDK, and set the minimum supported version as its new version.
+			sdkName := pkg.Description.Name
+			sdkVersion, err := p.getSdkVersion(sdkName)
+
+			if err != nil {
+				log.Infof("Failed to resolve %s SDK version for package %s: %v", sdkName, name, err)
```
wagoodman, line 88:
```suggestion
				log.Tracef("failed to resolve %s SDK version for package %s: %v", sdkName, name, err)
```

syft/pkg/cataloger/dart/parse_pubspec_lock.go
```diff
@@ -68,8 +70,28 @@ func parsePubspecLock(_ context.Context, _ file.Resolver, _ *generic.Environment
...
+			} else {
+				log.Debugf("Resolved %s SDK version for package %s to %s", sdkName, name, sdkVersion)
```
wagoodman, line 90:
```suggestion
				log.Tracef("resolved %s SDK version for package %s to %s", sdkName, name, sdkVersion)
```

syft/pkg/cataloger/dart/parse_pubspec_lock.go
```diff
@@ -68,8 +70,28 @@ func parsePubspecLock(_ context.Context, _ file.Resolver, _ *generic.Environment
...
+			// version set to "0.0.0" in the package definition. The actual version
+			// should refer to the SDK version, which is defined in a dedicated section
+			// in the pubspec.lock file and uses a version range constraint.
+			//
+			// If such a package is detected, look up the version range constraint of
+			// its matching SDK, and set the minimum supported version as its new version.
+			sdkName := pkg.Description.Name
+			sdkVersion, err := p.getSdkVersion(sdkName)
+
+			if err != nil {
```
wagoodman, line 87: if we run into an error, I think we should clear the version from the package entirely (since 0.0.0 is not correct). This would allow the package name/version compliance policy to be correctly applied.

syft/pkg/cataloger/dart/parse_pubspec_lock.go
```diff
@@ -89,6 +111,68 @@ func parsePubspecLock(_ context.Context, _ file.Resolver, _ *generic.Environment
...
+//	getSdkVersion("flutter") -> "3.24.5"
+//	getSdkVersion("undefined") -> error
+func (psl *pubspecLock) getSdkVersion(sdk string) (string, error) {
+	constraint, found := psl.Sdks[sdk]
+
+	if !found {
+		return "", fmt.Errorf("cannot find %s SDK", sdk)
+	}
+
+	return parseMinimumSdkVersion(constraint)
```
wagoodman, line 137: the python cataloger does similar behavior for requirements.txt parsing (when a range is found). In that case we have a configuration item for the cataloger. Just mentioning from a consistency perspective, but ultimately I think that a config item probably is not needed in this case.

syft/pkg/cataloger/dart/parse_pubspec_lock.go
sgreg: Hope you understood you correct with this: if resolving the version fails, the package itself is now skipped and not added to the package list.

syft/pkg/cataloger/dart/parse_pubspec_lock.go
sgreg: I think the case is a bit different with Dart here.

As far as I can tell, `pubspec.lock` always lists exact versions for the packages (`pubspec.yaml` would be a different situation), only the SDK section in it uses ranges / minimum version. Making the version guessing a config option would potentially mean that anything related to the Flutter SDK, including Flutter itself, might be omitted from the SBOM. Not sure if that's a good idea :)

derbauer97:
Hello,

We are very interested in the fix you have proposed as it will benefit our users. Could you please let us know when you plan to release it? We already tested it against one of our applications and the flutter version is now recognized correctly :)

dariozachow:
any updates on this pr?

sgreg:
@wagoodman just checking, is there anything you'd still need me to do here? Or is this simply waiting for your internal review process? Just making sure I didn't miss something and block my own PR here :)

PR #3615, opened 2025-01-23
Fix namespace value for OpenSUSE distros

### LINKED ISSUE #3534: RPM-based PURLs sometimes have incorrect namespace (specifically OpenSUSE)
**What happened**:

Packages created based on RPM database information (such as those from OpenSUSE) sometimes contain incorrect PURL namespace values. 

For example, running Syft against the `opensuse/leap:15.6` official Docker image produces PURLs similar to the following:

```
pkg:rpm/opensuse-leap/bash-sh@4.4-150400.25.22?arch=x86_64&distro=opensuse-leap-15.6&upstream=bash-4.4-150400.25.22.src.rpm
pkg:rpm/opensuse-leap/boost-license1_66_0@1.66.0-12.3.1?arch=noarch&distro=opensuse-leap-15.6&upstream=boost-base-1.66.0-12.3.1.src.rpm
pkg:rpm/opensuse-leap/libpsl5@0.20.1-150000.3.3.1?arch=x86_64&distro=opensuse-leap-15.6&upstream=libpsl-0.20.1-150000.3.3.1.src.rpm
```

While there aren't any 'official' documents detailing what the PURL `namespace` field should be, it appears that the expected value for OpenSUSE Leap is just `opensuse`, and 'Leap 15.6' should be encoded into the `distro` qualifier (as it correctly is in the above-mentioned cases).

The supporting evidence I have found for the 'correct' namespace simply being `opensuse` is:
- The one opensuse example PURL on the [official PURL spec github page](https://github.com/package-url/purl-spec):
```pkg:rpm/opensuse/curl@7.56.1-1.1.?arch=i386&distro=opensuse-tumbleweed```
- The OpenSUSE PURL entries that currently exist on osv.dev such as [this one](https://osv.dev/vulnerability/SUSE-SU-2024:4333-1):
```pkg:rpm/opensuse/libaom-devel-doc&distro=openSUSE%20Leap%2015.5```

It appears that the Syft [logic to generate PURLs for such packages](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/redhat/package.go#L90) simply uses the value of `Distro.ID` verbatim, with the exception of translating `rhel` to `redhat`. I believe that the most straightforward fix would be to add an additional check, something along the lines of "if the `Distro.ID` value starts with `opensuse`, set the namespace to be `opensuse`".

This same issue likely applies to other RPM-based systems (it all depends on how the `Distro.ID` values they use correspond to the expected PURL `namespace` values), but I haven't done any additional research yet into other distros.

The PURL `namespace` value for all OpenSUSE packages should be `opensuse`.

Other RPM-based distros should be evaluated to determine if similar issues apply to them.

Run Syft against the official `opensuse/leap:15.6` docker image.

mprpic:
# Description

Instead of namespacing them to the specific distro version, such as Leap or Tumbleweed, the namespace value is set to the vendor itself: "opensuse".

This change is similar to that made in https://github.com/anchore/syft/pull/2914.

- Fixes #3534

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/redhat/package.go
+++ b/syft/pkg/cataloger/redhat/package.go
@@ -92,6 +92,9 @@ func packageURL(name, arch string, epoch *int, srpm string, version, release str
 	if namespace == "rhel" {
 		namespace = "redhat"
 	}
+	if strings.HasPrefix(namespace, "opensuse") {
+		namespace = "opensuse"
+	}
 
 	qualifiers := map[string]string{}
 


spiffcs (approved):

spiffcs:
Thanks @mprpic - change LGTM. I'll follow up with #3534 to double check this behavior and see if there are any other edge cases we might want to cover in a follow up PR

PR #3631, opened 2025-01-29
feat: 3626 add option enable license content; disable by default

### LINKED ISSUE #3626: feat: disable license contents from being included in SBOM
We need a config value and/or flag that users can invoke to reduce the size of SBOMs when licenses are unable to be identified by the license scanner. 

This should focus on the change found here:
https://github.com/anchore/syft/pull/3366

**Why is this needed**:
Some users do not care about passing along the license contents in the final results of a scan. A license being unrecognized by the scanner should have the option of dropping the found bytes.

spiffcs:
# Description

With https://github.com/anchore/syft/pull/3366 landing the other day we need a sensible way to toggle the new content option before the next syft release.

This PR adds a new top level syft configuration for licenses
```yaml
license:
  # include the content of a license in the SBOM when syft
  # cannot determine a valid SPDX ID for the given license (env: SYFT_LICENSE_INCLUDE_UNKNOWN_LICENSE_CONTENT)
  include-unknown-license-content: false

  # adjust the percent as a fraction of the total text, in normalized words, that
  # matches any valid license for the given inputs, expressed as a percentage across all of the licenses matched. (env: SYFT_LICENSE_LICENSE_COVERAGE)
  license-coverage: 75
```

The new syft configuration keeps the current behavior where license content is disabled by default.
Content for unknown licenses to be included in the SBOM can be enabled on request

- Fixes #3626

When this lands I plan on updating the wiki under configuration with examples:
https://github.com/anchore/syft/wiki/configuration

### How to Validate
I've added a new CLI image that has a jar with an unknown license.

Reviewers can scan this jar using syft on this branch to see how contents can be optionally toggled on:

##### Content Enabled
```
SYFT_LICENSE_INCLUDE_UNKNOWN_LICENSE_CONTENT=true go run cmd/syft/main.go test/cli/test-fixtures/image-unknown-licenses/WowLibrary-1.0.0.jar -o json
 ✔ Indexed file system                                                     test/cli/test-fixtures/image-unknown-licenses/WowLibrary-1.0.0.jar
 ✔ Cataloged contents                                                        1de53b527bd3c2749dbbdb4fa6134adb531ca55cd83af14d4095c019063af569
   ├── ✔ Packages                        [1 packages]
   ├── ✔ File digests                    [1 files]
   ├── ✔ File metadata                   [1 locations]
   └── ✔ Executables                     [0 executables]
{
 "artifacts": [
  {
   "id": "c6849507c5393b1f",
   "name": "WowLibrary",
   "version": "1.0.0",
   "type": "java-archive",
   "foundBy": "java-archive-cataloger",
   "locations": [
    {
     "path": "/WowLibrary-1.0.0.jar",
     "accessPath": "/WowLibrary-1.0.0.jar",
     "annotations": {
      "evidence": "primary"
     }
    }
   ],
   "licenses": [
    {
     "value": "UNKNOWN",
     "spdxExpression": "UNKNOWN_53dd0aabc81f0bb155c021240a512bb1bf94f95a613c2d55566231ddbfef5dd3",
     "type": "declared",
     "urls": [],
     "locations": [
      {
       "path": "/WowLibrary-1.0.0.jar",
       "accessPath": "/WowLibrary-1.0.0.jar",
       "annotations": {
        "evidence": "primary"
       }
      }
     ],
     "contents": "Wow License v1.0\n\nThis is a custom license for the Wow Library.\nUsage is restricted as per the terms outlined in this file.\n"
    }
   ],
```

## Type of change

# Checklist:


syft/pkg/cataloger/golang/licenses_test.go
```diff
@@ -71,7 +71,11 @@ func Test_LicenseSearch(t *testing.T) {
 
 	localVendorDir := filepath.Join(wd, "test-fixtures", "licenses-vendor")
 
-	licenseScanner := licenses.NewScanner(licensecheck.Scan, float64(75))
+	sc := &licenses.ScannerConfig{
+		CoverageThreshold: 75,
+		Scanner:           licensecheck.Scan,
+	}
+	licenseScanner := licenses.NewScanner(sc)
```
kzantow, line 78: What if the scanner (and ScannerConfig) held the configuration for whether to include licenses or not? [IdentifyLicenseIDs](https://github.com/anchore/syft/blob/main/internal/licenses/scanner.go#L45C85-L45C91) could just return nil if configured to exclude contents. This would avoid storing this configuration in the context, and the `Search` function just only includes contents if returned. I don't know if returning no license and nil contents would be a problem.

spiffcs: I don't think I understand the question?

`Scanner` is an interface. There is no `ScannerConfig`. 

The two functions where we can make a decision on content are `Search` and `IdentifyLicenseIDs`

> [IdentifyLicenseIDs](https://github.com/anchore/syft/blob/main/internal/licenses/scanner.go#L45C85-L45C91) could just return nil if configured to exclude contents

> I don't know if returning no license and nil contents would be a problem.

In short it is in the current form.

There was an initial attempt to include the config on the `scanner` struct and not use context, but we need to return the content to `Search` from `IdentifyLicenseIDs` no matter what. This is because we're still creating a license for unknown licenses with locations and an ID that takes the sum of the content, just not including the entire text.

kzantow: The only hold up for me is using context to pass configuration in this way, since we're not using this same pattern elsewhere; everything else is 👍.

Outside of tests, `IdentifyLicenseIDs` is only used by `license.Search`, [as far as I can tell](https://github.com/search?q=repo%3Aanchore%2Fsyft%20IdentifyLicenseIDs&type=code), and since both of these are `internal` we can modify the functions/interfaces as much as we want. I don't think we need to change the `Search` signature at all (though I'm confused why it takes Context and Scanner; I was under the impression it was supposed to get the scanner out of the context or use a default, maybe the scanner is that default, but it's referenced without being checked for nil).

Why don't we just change the `Scanner` interface to instead of `IdentifyLicenseIDs` has `IdentifyLicenses` which returns the `([]license.License, error)`, matching `Search`, and [this code](https://github.com/anchore/syft/blob/main/internal/licenses/search.go#L33-L57) just gets moved to the scanner, this allows the scanner to have all the appropriate configuration for what gets returned and do whatever the right thing is there.

spiffcs:
@kzantow  I've updated this PR so we're now passing configuration to the scanner constructor at the initialization of the SBOM process and not mixing its config in context here.

The only ctx value we're using at the moment is to access the scanner.
This is currently done since it's shared across the different catalogers.

The Cover functionality is thread safe:
https://github.com/google/licensecheck/issues/16

I'm open to maybe using another PR in the future  to do a refactor that moves licenses.Scanner out of a context lookup and onto the config object of each cataloger that invokes it, but that would require a larger refactor of:
https://github.com/anchore/syft/blob/main/internal/task/package_tasks.go

The problem being that we only want one Scanner because of the cost of initialization for all of the default licenses it knows about:
https://github.com/anchore/syft/blob/684b6e3f9809a95ba08cb83c8d0661ab22eed8cc/internal/licenses/scanner.go#L26-L36

It's currently very difficult with the way the task package injects configuration into the different catalogers.

Take the go module catalog constructor as an example: https://github.com/anchore/syft/blob/684b6e3f9809a95ba08cb83c8d0661ab22eed8cc/internal/task/package_tasks.go#L82C3-L87

We have access to CatalogingFactoryConfig which is built by DefaultCatalogingFactoryConfig , but I'm not a fan of initializing the scanner on this struct and passing it around the catalogers since it's currently part of the public syft API. It's also called in multiple places and currently results in multiple scanners being initalized.

internal/licenses/scanner.go
```diff
@@ -2,66 +2,84 @@ package licenses
...
-const coverageThreshold = 75 // determined by experimentation
+const (
+	DefaultCoverageThreshold     = 75 // determined by experimentation
+	DefaultIncludeLicenseContent = false
+)
 
 type Scanner interface {
 	IdentifyLicenseIDs(context.Context, io.Reader) ([]string, []byte, error)
+	FileSearch(context.Context, file.LocationReadCloser) ([]file.License, error)
+	PkgSearch(context.Context, file.LocationReadCloser) ([]pkg.License, error)
```
kzantow, line 23: TIL we have both `file.License` and `pkg.License`

kzantow (approved):
LGTM 🎉

internal/licenses/search_test.go
```diff
@@ -56,8 +108,26 @@ func TestSearch(t *testing.T) {
 			},
 		},
 		{
-			name: "custom license",
+			name: "custom license no content by default",
 			in:   "test-fixtures/nvidia-software-and-cuda-supplement",
+			expected: expectation{
+				yieldError: false,
```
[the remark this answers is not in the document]
spiffcs: Ah Thanks! - I'll update this and then merge

PR #3633, opened 2025-01-29
chore: replace all shorthand tags of mapstruct -> mapstructure

spiffcs:
# Description

When doing https://github.com/anchore/syft/pull/3631 I noticed that syft had some incorrect tags for map-structure that might have been auto completed incorrectly in prior PR. This PR updates the struct tags to use the correct `mapstructure` 

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_package_json.go
+++ b/syft/pkg/cataloger/javascript/parse_package_json.go
@@ -36,9 +36,9 @@ type packageJSON struct {
 }
 
 type author struct {
-	Name  string `json:"name" mapstruct:"name"`
-	Email string `json:"email" mapstruct:"email"`
-	URL   string `json:"url" mapstruct:"url"`
+	Name  string `json:"name" mapstructure:"name"`
+	Email string `json:"email" mapstructure:"email"`
+	URL   string `json:"url" mapstructure:"url"`
 }
 
 type repository struct {
--- a/syft/pkg/python.go
+++ b/syft/pkg/python.go
@@ -13,18 +13,18 @@ var _ FileOwner = (*PythonPackage)(nil)
 // Historically these were defined in PEPs 345, 314, and 241, but have been superseded by PEP 566. This means that this
 // struct can (partially) express at least versions 1.0, 1.1, 1.2, 2.1, 2.2, and 2.3 of the metadata format.
 type PythonPackage struct {
-	Name                 string                     `json:"name" mapstruct:"Name"`
-	Version              string                     `json:"version" mapstruct:"Version"`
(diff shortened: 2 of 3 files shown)


kzantow (commented):
These files under the `syft/` package are all part of the syft library API. These do not use mapstructure for anything, which is probably why no one has reported issues with these tags before. The only structs that use mapstructure are the config structs, which all reside next to the command-specific stuff under the top-level `cmd/` package, such as [the command options structs](https://github.com/anchore/syft/blob/main/cmd/syft/internal/commands/scan.go#L67-L73) or the [options](https://github.com/anchore/syft/tree/main/cmd/syft/internal/options).

I think having these tags on these structs in the `syft/` package at all is confusing and we should remove all the mapstructure and mapstruct tags on these structs so contributors don't keep adding them for no reason and potentially confuse _these structs_ with CLI configuration structs.

Other than that, it doesn't really _hurt_ anything to have these tags, but I would update them to match the json tags or the fields exactly -- some of them have changes to the case of letters (e.g. .`AuthorEmail` vs `Authoremail` -- the former would have been used in the absence of an appropriate `mapstructure` tag, so fixing the tag changes the behavior.)

spiffcs:
> These files under the `syft/` package are all part of the syft library API. These do not use mapstructure for anything, which is probably why no one has reported issues with these tags before. The only structs that use mapstructure are the config structs, which all reside next to the command-specific stuff under the top-level `cmd/` package, such as [the command options structs](https://github.com/anchore/syft/blob/main/cmd/syft/internal/commands/scan.go#L67-L73) or the [options](https://github.com/anchore/syft/tree/main/cmd/syft/internal/options).
> 
> I think having these tags on these structs in the `syft/` package at all is confusing and we should remove all the mapstructure and mapstruct tags on these structs so contributors don't keep adding them for no reason and potentially confuse _these structs_ with CLI configuration structs.
> 
> Other than that, it doesn't really _hurt_ anything to have these tags, but I would update them to match the json tags or the fields exactly -- some of them have changes to the case of letters (e.g. .`AuthorEmail` vs `Authoremail` -- the former would have been used in the absence of an appropriate `mapstructure` tag, so fixing the tag changes the behavior.)

I've updated them to match their fields with the correct casing.

I'm not sure about removing them from `syft/*`:
Here we're using them to help with the advanced decoding from a `package.json`
https://github.com/anchore/syft/blob/a16e374a5053e88660aa848c6ae635cfcbfe6ccc/syft/pkg/cataloger/javascript/parse_package_json.go#L83-L102

We can potentially remove them for the python/swipl case as those structs are not used by any downstream `mapstructure.Decode` calls.

kzantow (approved):

PR #3645, opened 2025-02-05
fix: reduce warn levels to debug for non-actionable errors

### LINKED ISSUE #3470: syft should not warn on known bad package.json
syft should ignore explicitly malformed test file package.json

e.g. https://github.com/browserify/resolve/blob/main/test/resolver/malformed_package_json/package.json

**Why is this needed**:
Reduce false positive warnings in runs.

Other tooling already excludes the file: https://github.com/search?q=%22malformed_package_json%22&type=code

**Additional context**:

Syft runs presently throw these warnings for it:
```
#29 0.052 time="2024-11-21T05:23:08Z" level=info msg="starting syft scanner for buildkit v1.4.0"
#29 7.466 [0007]  WARN cataloger failed cataloger=javascript-package-cataloger error=failed to parse package.json file: unexpected EOF location=/app/node_modules/resolve/test/resolver/malformed_package_json/package.json
#29 37.18 [0037]  WARN cataloger failed cataloger=javascript-package-cataloger error=failed to parse package.json file: unexpected EOF location=/app/node_modules/eslint-plugin-react/node_modules/resolve/test/resolver/malformed_package_json/package.json
#29 37.27 [0037]  WARN unable to extract licenses from javascript package.json: unmarshal failed
#29 37.29 [0037]  WARN cataloger failed cataloger=javascript-package-cataloger error=failed to parse package.json file: unexpected EOF location=/app/node_modules/resolve/test/resolver/malformed_package_json/package.json
```

kzantow:
# Description

This PR just updates the majority of WARN log calls to DEBUG, these _should_ all be messages that contain information that really isn't _actionable_ by the user: for example, a user wouldn't be expected to modify the material that they are scanning to fix package.json errors, these would likely be in a prebuilt container or other source not controlled by the user; and a user would not be expected to modify syft code due to a bug.

I've tried to leave as warnings things that _are_ actionable, things like invalid configuration: misconfigured cache directories, invalid cataloger selection, etc.. I've also updated any uses of `log.WithFields` that were _also_ using a formatted message to just include the data as fields, e.g. `log.WithFields(..).Debugf(...)` -> `log.WithFields(..., <data used in Debugf>).Debug(...)`.

- Fixes #3470

## Type of change

# Checklist:

### DIFF
--- a/cmd/syft/cli/ui/handle_attestation.go
+++ b/cmd/syft/cli/ui/handle_attestation.go
@@ -61,7 +61,7 @@ type backgroundLineReader struct {
 func (m *Handler) handleAttestationStarted(e partybus.Event) []tea.Model {
 	reader, prog, taskInfo, err := syftEventParsers.ParseAttestationStartedEvent(e)
 	if err != nil {
-		log.WithFields("error", err).Warn("unable to parse event")
+		log.WithFields("error", err).Debug("unable to parse event")
 		return nil
 	}
 
--- a/cmd/syft/cli/ui/handle_cataloger_task.go
+++ b/cmd/syft/cli/ui/handle_cataloger_task.go
@@ -89,7 +89,7 @@ func (cts catalogerTaskModel) View() string {
 func (m *Handler) handleCatalogerTaskStarted(e partybus.Event) ([]tea.Model, tea.Cmd) {
 	mon, info, err := syftEventParsers.ParseCatalogerTaskStarted(e)
 	if err != nil {
-		log.WithFields("error", err).Warn("unable to parse event")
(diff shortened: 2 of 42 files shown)


spiffcs (approved):
I checked each file.

If we're sticking with the philosophy that `Warn` should only be something that the user can take an action on then this PR get's us closer to that goal. 

I'm a little worried about throwing some of these over the wall and making the `Debug` output more crowded with these messages, but maybe the next step after this PR is to start pruning those messages down based on feedback of what we find useful. 

Also, more information is always better when debugging so these might be overblown concerns.

PR #3647, opened 2025-02-06
Support extracting symbols in .dynsym section for GraalVM Native Images

rudsberg:
# Description

This PR adds support for extracting SBOMs embedded in native images where the SBOM symbols are saved in the `.dynsym` section of ELF files. Newer versions of GraalVM Native Image saves the SBOM symbols in the `.dynsym` section while older versions saves them in the `.symtab` section. This PR ensures SBOM extraction works in both cases. 

## Type of change

# Checklist:

- [x] I have added unit tests that cover changed behavior*

*No additional tests were added. The new `getSymbols` function is exercised via [TestParseNativeImage that calls fetchPkgs](https://github.com/anchore/syft/blob/e584c9f416028051fbdc80fffda4e624aaed528d/syft/pkg/cataloger/java/graalvm_native_image_cataloger_test.go#L49). This is a negative test testing that an error is thrown when SBOM extraction is attempted on a native image that wasn't created with an embedded SBOM. 

### DIFF
--- a/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
+++ b/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
@@ -251,11 +251,11 @@ func (ni nativeImageElf) fetchPkgs() (pkgs []pkg.Package, relationships []artifa
 	var sbomLength elf.Symbol
 	var svmVersion elf.Symbol
 
-	si, err := bi.Symbols()
+	si, err := ni.getSymbols()
 	if err != nil {
-		return nil, nil, fmt.Errorf("no symbols found in binary: %w", err)
+		return nil, nil, err
 	}
-	if si == nil {
+	if len(si) == 0 {
 		return nil, nil, errors.New(nativeImageMissingSymbolsError)
 	}
 	for _, s := range si {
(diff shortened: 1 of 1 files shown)


wagoodman:
How many symbols would be added to the SBOM for these packages (say on average)?

martijndwars:
> How many symbols would be added to the SBOM for these packages (say on average)?

Just to make sure I understand where the question comes from; are you concerned that Syft would spend too much time/memory on reading the (dynamic) symbol table from the ELF? Note that symbols are added to the ELF, not the SBOM. The SBOM is stored in the symbol named `sbom`.

rudsberg:
@wagoodman, as @martijndwars pointed out: the relevant symbols (`sbom`, `sbom_length`, and `__svm_version_info`) are part of the Native Image ELF file and not the final SBOM. An [error is thrown](https://github.com/anchore/syft/pull/3647/files#diff-e01e6ad41bdcf288e86087471aa8ec9c8052916e50a96683fdad7b4879d37c14R272) if any of those symbols are not present. `TestParseNativeImage` tests that error path. There is currently no integration test that: creates a Native Image with an embedded SBOM, invokes syft, and makes assertions on the extracted SBOM. Such a test would indirectly assert that the three symbols are part of the ELF file. Note that the SBOM feature is only available in Oracle GraalVM (not the community version) which is released under GFTC license and that can be pulled from [Oracle's container registry](https://container-registry.oracle.com/ords/f?p=113:4:103988352861753:::RP,4:P4_REPOSITORY,AI_REPOSITORY,P4_REPOSITORY_NAME,AI_REPOSITORY_NAME:2204,2204,Oracle%20GraalVM%20Native%20Image%20released%20under%20GFTC,Oracle%20GraalVM%20Native%20Image%20released%20under%20GFTC&cs=3Rz2dR5LiY7xkv1YZOLPEBddj5ROcymauMxzlFQHP7BHUHTlydgFB89LxDik9Z0orF-p4AfPVObBs4ihNdzfjng).

wagoodman:
@martijndwars @rudsberg thanks for the corrections -- I assumed that there was a need to extract symbols for the intention of having more detail in the SBOM, but now I see that this is really a response to a change upstream and the cataloging behavior is still the same. 

Regarding the test update, I agree, that improvement can always come in another PR (I see how that would be more difficult to test here!).

wagoodman (approved):
Thanks for the contribution!

PR #3657, opened 2025-02-11
chore: update runners to ubuntu-24.04

kzantow:


popey (approved):
LGTM!

PR #3672, opened 2025-02-20
fix: Version parse caused by line breaks on different platforms

idhyt:
# Description

Under windows platform, `conanfile.txt` has a line break of `\r\n`, so the parsed version number will have '\r', this problem does not exist under unix platform.

- Fixes

# Checklist:


kzantow (approved):
LGTM, thanks @idhyt!

PR #3680, opened 2025-02-22
fix: typed nils in cyclonedx

kzantow:
# Description

This PR fixes an issue where "typed nil" values are causing invalid cyclonedx to be generated with components containing `"hashes": null`, which [does not pass JSON validation](https://github.com/anchore/grype/actions/runs/13465429155/job/37630101321#step:4:159).

## Type of change

# Checklist:

### DIFF
--- a/syft/format/common/cyclonedxhelpers/to_format_model.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model.go
@@ -70,27 +70,16 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 			continue
 		}
 		metadata = &fileMetadata
-
-		// Digests
-		var digests []file.Digest
-		if digestsForLocation, exists := artifacts.FileDigests[coordinate]; exists {
-			digests = digestsForLocation
-		}
-
-		cdxHashes := digestsToHashes(digests)
 		components = append(components, cyclonedx.Component{
 			BOMRef: string(coordinate.ID()),
 			Type:   cyclonedx.ComponentTypeFile,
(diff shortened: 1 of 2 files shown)


syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -70,27 +70,16 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
...
-			digests = digestsForLocation
-		}
-
-		cdxHashes := digestsToHashes(digests)
 		components = append(components, cyclonedx.Component{
 			BOMRef: string(coordinate.ID()),
 			Type:   cyclonedx.ComponentTypeFile,
 			Name:   metadata.Path,
-			Hashes: &cdxHashes,
+			Hashes: slicePtr(digestsToHashes(artifacts.FileDigests[coordinate])),
```
kzantow, line 77: I've tried to add a test showing this fix working, but for some reason it doesn't have the same behavior as Grype. We are [effectively running the same test in both, and in Syft it works](https://github.com/anchore/syft/blob/main/schema/cyclonedx/Makefile#L5-L11) as expected while [in Grype](https://github.com/anchore/grype/blob/main/schema/cyclonedx/Makefile#L12-L14) it [has a bug where a null value is output for `"hashes"`](https://github.com/anchore/grype/actions/runs/13511227636/job/37751705949?pr=2473#step:4:159). I've verified this specific fix corrects the issue in the Grype test, but I've very confused why there is different behavior.

willmurphyscode (approved):

PR #3697, opened 2025-02-28
Add downloadLocation URI validation

### LINKED ISSUE #3696: Download location is not a valid URI
**What happened**:

Syft created an SBOM with packages containing an invalid URI for downloadLocation. This happens with npm packages using spdx spec `spdx-json`

Example incompliant package:
```json
{
      "name": "@isaacs/cliui",
      "SPDXID": "SPDXRef-Package-npm--isaacs-cliui-7026ea92955de2ad",
      "versionInfo": "8.0.2",
      "supplier": "Person: Ben Coe (ben@npmjs.com)",
      "originator": "Person: Ben Coe (ben@npmjs.com)",
      "downloadLocation": "yargs/cliui",
      "filesAnalyzed": false,
      "sourceInfo": "acquired package info from installed node module manifest file: /usr/local/lib/node_modules/npm/node_modules/@isaacs/cliui/package.json",
      "licenseDeclared": "ISC",
      "description": "easily create complex multi-column command-line-interfaces",
      "externalRefs": [
        {
          "referenceType": "cpe23Type",
          "referenceLocator": "cpe:2.3:a:\\@isaacs\\/cliui:\\@isaacs\\/cliui:8.0.2:*:*:*:*:*:*:*"
        },
        {
          "referenceType": "purl",
          "referenceLocator": "pkg:npm/%40isaacs/cliui@8.0.2"
        }
      ]
    }
```

A package with a downloadLocation compliant to the spdx spec.

```
syft scan redis/redisinsight:2.60.0 -o spdx-json=scan.json
```

**Environment**:
- Output of `syft version`: v1.20.0
- OS (e.g: `cat /etc/os-release` or similar): Ubuntu 22.04.5 LTS

stgrace:
# Description

- Fixes #3696

## Type of change

# Checklist:


stgrace:
Attempted same command with changes from issue #3696, which sets it to NOASSERTION because of the validation issue

```json
{
            "name": "@isaacs/cliui",
            "SPDXID": "SPDXRef-Package-npm--isaacs-cliui-7026ea92955de2ad",
            "versionInfo": "8.0.2",
            "supplier": "Person: Ben Coe (ben@npmjs.com)",
            "originator": "Person: Ben Coe (ben@npmjs.com)",
            "filesAnalyzed": false,
            "sourceInfo": "acquired package info from installed node module manifest file: /usr/local/lib/node_modules/npm/node_modules/@isaacs/cliui/package.json",
            "licenseDeclared": "ISC",
            "description": "easily create complex multi-column command-line-interfaces",
            "externalRefs": [
                {
                    "referenceType": "cpe23Type",
                    "referenceLocator": "cpe:2.3:a:\\@isaacs\\/cliui:\\@isaacs\\/cliui:8.0.2:*:*:*:*:*:*:*"
                },
                {
                    "referenceType": "purl",
                    "referenceLocator": "pkg:npm/%40isaacs/cliui@8.0.2"
                }
            ]
        }
```

syft/format/internal/spdxutil/helpers/download_location.go
```diff
@@ -17,17 +36,17 @@ func DownloadLocation(p pkg.Package) string {
 	if hasMetadata(p) {
 		switch metadata := p.Metadata.(type) {
 		case pkg.ApkDBEntry:
-			return NoneIfEmpty(metadata.URL)
+			return checkUri(metadata.URL)
```
kzantow, line 39: Instead of updating every spot here to call `checkUri`, could we centralize this call? maybe add a var `location := ""`, set this variable in each of these case statements instead of returning, and `return UriValue(location)`

syft/format/internal/spdxutil/helpers/download_location.go
```diff
@@ -1,10 +1,29 @@
...
 
 const NONE = "NONE"
 const NOASSERTION = "NOASSERTION"
 
+func isUriValid(uri string) bool {
+	_, err := urilib.NewURIRef(uri)
+	return err == nil
+}
+
+func checkUri(uri string) string {
```
kzantow, line 16: nit: I think this could be renamed to be slightly more representative of what it's doing, maybe `UriValue`, since it's not "checking" but rather returning the value we should be using.

syft/format/internal/spdxutil/helpers/download_location.go
stgrace: I've updated it, but it does lead to some change in one of the testcases, where providing no package result in NONE instead of NOASSERTION, which makes more sense in my opinion anyway.

syft/format/internal/spdxutil/helpers/download_location.go
stgrace: Updated 👍

kzantow (approved):
Looks great, thanks very much @stgrace !

syft/format/internal/spdxutil/helpers/download_location.go
kzantow: Agree -- NONE seems like the right value

kzantow:
Argh, I spoke too soon; it looks like we need some snapshots and other tests updated, but I think the core change here is the right thing. I can help with this, also fine if you'd like to do it.

`make lint-fix` should help the static analysis issue(s), but the naming it probably won't help with.

syft/format/internal/spdxutil/helpers/download_location.go
```diff
@@ -14,21 +17,38 @@ func DownloadLocation(p pkg.Package) string {
...
+	return UriValue(location)
+}
+
+func isUriValid(uri string) bool {
+	_, err := urilib.NewURIRef(uri)
+	return err == nil
+}
+
+func UriValue(uri string) string {
+	if NoneIfEmpty(uri) != NONE {
```
kzantow, line 46: I think using `NOASSERTION` here actually ends up being the right thing to do. The reason is, we're not claiming that there is no download location, we're only claiming that we cannot provide one. As such, I think we should update this to: `if isURIValid(uri) { return uri } return NOASSERTION`, or instead get rid of this function altogether and just update the last bit of `DownloadLocation` to have this, and ensure `isURIValid` returns false for empty string. I'm happy to make these updates.

stgrace: So if I understand it correctly, NONE should not be included here if it's an empty string. It is however a valid option according to the spec. Should there be a clause to set the download location to NONE if it is set as such by the creator of the package?

From the spec:

> NONE if there is no download location whatsoever.

I'll make the changes to set to NOASSERTION, unless a package explicitly sets the download location to NONE. Let me know what you think :)

stgrace: Added some changes suggested by `make lint-fix` as well.

kzantow (approved):
I think this makes a lot of sense, thanks much @stgrace!

PR #3704, opened 2025-03-03
feat: add Debian archive (.deb) file cataloger

### LINKED ISSUE #3315: Catalog deb archives directly
(1) real deb files in folder "/test/aaa*.deb ; /test/bbb*.deb".  
(2) run cmd "syft scan dir:/test -o cyclonedx-xml=aaa.xml"   
(3) result: no deb name listed in the result file aaa.xml  
(4) expected result: syft scan dir could list all deb file 's name as result  in the aaa.xml .
(5) rpm packages in redhat env is ok. 
(6) but in debian os system, after user login the os system ,in real filesystem env,run this command , no respected result show.

popey:
Add a cataloger that parses Debian package (.deb) archive files directly, allowing Syft to discover packages from .deb files without requiring them to be installed on the system. This implements issue #3315.

Key features:
- Parse .deb AR archives to extract package metadata
- Support for gzip, xz, and zstd compressed control files
- Extract package metadata from control files
- Process file information from md5sums files
- Mark configuration files from conffiles entries
- Handle trailing slashes in archive member names

# Description

- Fixes #3315

## Type of change

# Checklist:


syft/pkg/cataloger/debian/parse_deb_archive.go
```diff
@@ -0,0 +1,268 @@
...
+)
+
+// parseDebArchive parses a Debian package archive (.deb) file and returns the packages it contains.
+// A .deb file is an ar archive containing three main files:
+// - debian-binary: Version of the .deb format (usually "2.0")
+// - control.tar.gz/xz/zst: Contains package metadata (control file, md5sums, conffiles)
+// - data.tar.gz/xz/zst: Contains the actual files to be installed (not processed by this cataloger)
+//
+// This function extracts and processes the control information to create package metadata.
+func parseDebArchive(_ context.Context, _ file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
```
wagoodman, line 31: this should be decomposed into smaller functions to help with readability (linting is primarily failing for that reason)

syft/pkg/cataloger/debian/parse_deb_archive.go
```diff
@@ -0,0 +1,268 @@
...
+				return nil, nil, fmt.Errorf("failed to read control.tar: %w", err)
+			}
+			
+			// Determine compression type (.tar.gz, .tar.xz, .tar.zst)
+			compressionType := detectCompression(header.Name)
+			if compressionType == "" {
+				return nil, nil, fmt.Errorf("unsupported control file compression: %s", header.Name)
+			}
+			
+			log.Debugf("decompressing control.tar using %s compression", compressionType)
```
wagoodman, line 66: there are a few log statements that probably aren't needed (this one, the "found debian package" one, and the "found archive entry")

syft/pkg/cataloger/debian/parse_deb_archive.go
```diff
@@ -0,0 +1,268 @@
...
+		return "xz"
+	case strings.HasSuffix(cleanName, ".zst"):
+		return "zstd"
+	default:
+		return ""
+	}
+}
+
+// decompress handles decompression of the control.tar file based on compression type
+func decompress(data []byte, compressionType string) ([]byte, error) {
```
wagoodman, line 139: no comment for a code change, but maybe a code comment. We've been using https://github.com/anchore/archiver/tree/v3 for unarchiving arbitrary archives which handles several use cases. The downside is that the method in question takes filepaths and not readers. We've also wanted to see what it would be like to replace this lib with https://github.com/mholt/archives .

popey: I don't understand if you're saying moving to anchore/archiver, or mholt/archives is a requirement or nice to have?

wagoodman: this is more of an fyi / comment for a future PR down the road sometime -- no need to change the code for now since we're in a half way state for that particular dependency.

syft/pkg/cataloger/debian/parse_deb_archive_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/generic"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseDebArchive(t *testing.T) {
+	// Use the toilet package as a test fixture
+	fixture := "test-fixtures/toilet_0.3-1.4build1_amd64.deb"
```
wagoodman, line 19: ideally we should try and not checkin binary test fixtures if we can help it. Can you change this to pull from a remote source or extract from a container image?

wagoodman (approved):
I made a few changes, but overall looks great!

changes made:
- added a specific DPKG archive metadata type; though it's the same as the DB entry we want to have the flexibility to be able to change one or the other independent of one another. This required a JSON schema update.
- Added licenses extraction from the embedded data.tar.* file
- removed the binary test fixture and used the common container image + pkgtest test harness
- removed io.ReadAll() uses and instead plumped through readers to  prevent large chunks of memory being allocated

PR #3714, opened 2025-03-07
chore: upload individual binaries as artifacts

kzantow:
# Description

- Fixes

## Type of change

# Checklist:


.github/workflows/validations.yaml
```diff
@@ -89,6 +89,61 @@ jobs:
...
+          key: snapshot-build-${{ github.run_id }}
+
+      # workaround for https://github.com/actions/cache/issues/1265
+      - name: (cache-miss) Snapshot build missing
+        if: steps.snapshot-cache.outputs.cache-hit != 'true'
+        run: echo "unable to download snapshots from previous job" && false
+
+      - run: npm install @actions/artifact@2.2.2
+
+      - uses: actions/github-script@v7
```
wagoodman, line 122: why not use https://github.com/actions/upload-artifact directly as an action?

kzantow: @wagoodman the problem using upload-artifact is that you are only able to upload a single artifact -- if you give it 4 files, for example, it combines them into a single zip and uploads that. I would rather have these available for download individually so an interested party doesn't have to download _all_ builds just to get the windows build, for example

wagoodman (approved):

PR #3720, opened 2025-03-11
Location order on packages should consider evidence annotations when sorting

wagoodman:
# Description

Today when sorting locations on packages we sort based on the real and access path on these objects. This is almost correct: we should additionally consider first any package annotations that exist (i.e. is this primary evidence or supporting evidence).

This PR adjusts the behavior such that all primary evidence is listed before supporting evidence.

While working on tests for this PR and a soon-to-follow on PR, it made sense to improve some of the `cmptest` options we use around location sets and license sets (which have locations on them and affect comparisons). No functional change was made here.

## Type of change

# Checklist:

### DIFF
--- a/internal/cmptest/set.go
+++ b/internal/cmptest/set.go
@@ -1,0 +1,24 @@
+package cmptest
+
+type slicer[T any] interface {
+	ToSlice() []T
+}
+
+func buildSetComparer[T any, S slicer[T]](l func(x, y T) bool) func(x, y S) bool {
+	return func(x, y S) bool {
+		xs := x.ToSlice()
+		ys := y.ToSlice()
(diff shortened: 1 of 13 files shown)


internal/cmptest/set.go
```diff
@@ -0,0 +1,24 @@
+package cmptest
+
+type slicer[T any] interface {
+	ToSlice() []T
+}
+
+func buildSetComparer[T any, S slicer[T]](l func(x, y T) bool) func(x, y S) bool {
```
kzantow, line 7: `cmp` never ceases to amaze me how well it works.

kzantow (approved):

Issue #3705, opened 2025-03-04
Allow cancelling tasks that generate SBOM with syft.CreateSBOM()


danielpacak:
**What would you like to be added**:

It seems that `syft.CreateSBOM(context.Context, source.Source, *CreateSBOMConfig)` accepts the Go context as the first argument, but when I pass a cancellable context it's ignored and cataloguers' tasks continue running nevertheless.

``` go
ctx := context.Background()
scanCtx, cancelScan := context.WithCancelCause(ctx)
defer cancelScan(nil)

go func() {
	// periodically check for memory pressure against soft memory limit ...

	// if memory is above the watermark
	cancelScan(memory.ErrMemoryPressure)
}()

syftJson, err := syft.CreateSBOM(scanCtx, ...)
```

**Why is this needed**:

I'm facing some high memory usage issues with Syft running in a container as a Kubernetes Pod. To bound somehow the upper memory limit before I get OOMKill errors I just want to cancel SBOM generation in idiomatic Go way, i.e. by cancelling the context passed to Syft library method.

**Additional context**:

* It seems that the cancellation request is respected by functions that pull down sources from remote registry, but once cataloguer tasks start I cannot stop Syft from running
* https://go.dev/doc/database/cancel-operations

kzantow:
Hi @danielpacak, this sounds like something that we would be happy to get added, generally speaking. The challenge is getting all the functions to respond to the cancel request at appropriate times. Most catalogers search for a set of files and then parse the file(s) one at a time, so a reasonably easy spot to add a hook to stop processing is during a file cataloging loop. However, there are a lot of catalogers and this may require a lot of different updates to really halt processing when asked to. There is a [generic cataloger](https://github.com/anchore/syft/blob/7571f8dfbaa7147030f4286571cc791c09be3607/syft/pkg/cataloger/generic/cataloger.go#L164) that handles selecting files and calling a parse function in a loop for a large number of catalogers, so updating this one cataloger would probably get over 60% cataloger coverage. But adding cancel behavior _to catalogers_ is only a part of the problem -- there are other processes such as filesystem indexing, online enrichment, or other processes that could result in a lot of memory used, too.

I would also just note that I'm looking to [parallelize a number of the catalogers similarly](https://github.com/anchore/syft/pull/3636), which could be a good time to think about adding some cancel behavior.

[...]

PR #3721, opened 2025-03-11
refactor: move license scanner out from internal package

danielpacak:
The goal is to reuse the default license scanner instance, which is set as Context value, between different scans. This is the main use case for Syft running as a daemon that periodically invokes CreateSBOM() method.

Towards: #3705

### DIFF
--- a/examples/create_simple_sbom/main.go
+++ b/examples/create_simple_sbom/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/anchore/syft/internal/licenses"
 	"github.com/anchore/syft/syft"
 	"github.com/anchore/syft/syft/format"
 	"github.com/anchore/syft/syft/format/syftjson"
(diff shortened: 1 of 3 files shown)


examples/create_simple_sbom/main.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/anchore/syft/internal/licenses"
```
wagoodman, line 8: this doesn't work as is since the example is suggesting for external API users to import from `syft/internal` which isn't possible

wagoodman: I think the easiest thing to do would be to export the existing `setupContext`

kzantow:
Hey @danielpacak thanks for the contribution, I think someone else's PR ended up getting merged for this: https://github.com/anchore/syft/pull/3796, so I'm going to close this PR. Please shout if the current changes on `main` are not sufficient for your needs!

PR #3776, opened 2025-04-02
Parse GitHub actions comments

wagoodman:
This updates the github actions cataloger to consider cases when there are explicit version annotations added to pinned action references. So without this change:

```
jobs:
  static-analysis:
    runs-on: ubuntu-24.04
    steps:
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
```

... syft would pick up on this package as `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683`, however with this PR now it will be raised up as `actions/checkout@v4.2.2` with additional metadata exposing the original values:

```golang
Metadata: pkg.GitHubActionsUseStatement{
	Value:   "actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683",
	Comment: "v4.2.2",
},
```

Partially addresses https://github.com/anchore/grype/issues/2530

## Type of change

# Checklist:


syft/pkg/cataloger/githubactions/parse_workflow.go
```diff
@@ -97,3 +118,101 @@ func parseWorkflowForActionUsage(_ context.Context, _ file.Resolver, _ *generic.
...
+	comment := node.LineComment
+	if comment == "" {
+		comment = node.HeadComment
+	}
+	if comment == "" {
+		comment = node.FootComment
+	}
+
+	if comment != "" {
+		versionRegex := regexp.MustCompile(`v\d+\.\d+\.\d+`)
```
kzantow, line 197: does it make sense to make this a bit more lenient -- `v?\d+(\.\d+)*` or so?

syft/pkg/github.go
```diff
@@ -0,0 +1,6 @@
+package pkg
+
+type GitHubActionsUseStatement struct {
```
kzantow, line 3: This seems to border "debug info" a little bit -- is it necessary? If Syft hasn't parsed a version or a sha, it's unlikely this is of use to anyone else, is it? I wonder if there would be utility in allowing different "levels" of detail in the SBOM. I know we have other catalogers doing similar stuff, so this isn't a blocker, just an opinion.

kzantow (approved):

syft/pkg/github.go
wagoodman: It felt necessary since we're parsing a comment and treating it as the source of truth for a package version. The original version is lost without having a metadata type that exposes this.

> If Syft hasn't parsed a version or a sha, it's unlikely this is of use to anyone else, is it?

this resonates in the sense that if there is no comment to parse this is a useless metadata object.

> I wonder if there would be utility in allowing different "levels" of detail in the SBOM.

Do you have an example or idea how this would work?

kzantow: I don't really have an example but I would think something like logging levels but for SBOM detail. Maybe we determine level=1 is NTIA minimum elements, and level=2, 3, etc.. add more details up to including parsed data like full MANFIEST.MF and Portable Executable properties, etc.. where the lowest level excludes these details. There probably isn't a reason to do this, just an idle thought. This all was me thinking about SBOM size that some people may at some point be concerned with this.

PR #3783, opened 2025-04-07
fix:Resolve ancestral symlinks correctly

### LINKED ISSUE #3614: go-module-file-cataloger fails if symlinks in path
**What happened**:

Scanning a directory containing a Go project results in the `go-module-file-cataloger` failing to properly resolve the path. As a result it fails to report the h1 digests of the Go dependencies and reports the warning `[0000]  WARN unable to build link resolution index for filetree search context error=unable to get node for path="xyz": cycle during symlink resolution`.

Scan reporting same results no matter if symlinks are in the path or not.

Scan the same Go project with and without symlinks in the path and compare results.

If I have diagnosed it right, the issue comes from the [`go-module-file-cataloger` implementation of the method `RelativeFileByPath`](https://github.com/anchore/syft/blob/222e6548a96f8c80015c1d24f01dea3052a04893/syft/pkg/cataloger/golang/parse_go_mod.go#L121).

**Environment**:
- Output of `syft version`: `syft 1.19.0`

VictorHuu:
The [directory resolver](https://github.com/anchore/syft/blob/f11377fe30f53aec0c8e6ce38b3a480e1ffa1889/syft/internal/fileresolver/directory_indexer.go#L376-L386) does not consider cases when the [```Rel```](https://pkg.go.dev/path/filepath#Rel) method in the standard library may return ```..```, since ```targetPath``` may be the ancestor of the ```basePath```. 

- Fixes #3614 

## Type of change

# Checklist:


wagoodman (approved):
🚀

PR #3795, opened 2025-04-12
(feat): support skipping archive extraction with file source

adammcclenaghan:
# Description

Currently when an archive file is scanned with a file source, the archive is extracted before scanning

This is different from the default behavior when an archive file is encountered via directory source, in that case archives are only extracted when deemed necessary by catalogers (ie the java archive cataloguer)

As a Syft library user, when scanning with file source I would like the option to skip extracting archives and instead allow cataloguers decide whether to extract them. 

I've added `SkipExtractArchive` to the filesource Config here, and by default it will be `false` so all existing users will be unaffected.

## Type of change

# Checklist:

### DIFF
--- a/syft/source/filesource/file_source.go
+++ b/syft/source/filesource/file_source.go
@@ -26,10 +26,11 @@ import (
 var _ source.Source = (*fileSource)(nil)
 
 type Config struct {
-	Path             string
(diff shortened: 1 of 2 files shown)


wagoodman (approved):
🙌

PR #3796, opened 2025-04-12
Perf: skip license scanner injection

adammcclenaghan:
# Description
Creating the default license scanner is a little computationally expensive.

I use Syft as a library and would like the option to set the license scanner myself on the context.

This change skips injecting the license scanner within CreateSbom if it is already set on the context.

The licensectx package has been created to expose `GetContextKey` and `NewDefaultLicenseScanner` to Syft library users.

## Type of change

# Checklist:


adammcclenaghan:
I've just seen https://github.com/anchore/syft/pull/3721

Which is a similar effort - sorry I missed that. The main difference with this PR is that it goes slightly further and allows library users to set the value on the context & access `NewDefaultScanner` 

Be good to get your thoughts too @danielpacack

syft/licensectx/context.go
```diff
@@ -0,0 +1,13 @@
+package licensectx
```
kzantow, line 1: We have some exported license-related things here: https://github.com/anchore/syft/blob/main/syft/cataloging/license.go, maybe instead of making a new package this would be a better spot to include the exported context key and creation function? These seem like configuration items, even though they are using context to provide them.

adammcclenaghan: Thanks for the suggestion -  that's a good place for these! Have moved them :)

syft/cataloging/license.go
```diff
@@ -13,3 +13,15 @@ func DefaultLicenseConfig() LicenseConfig {
...
+
+// GetContextKey allows callers to access the key used by
+// syft to store the license scanner in the context
+func GetContextKey() licenses.LicenseScannerKey {
+	return licenses.CtxKey
+}
+
+// NewDefaultLicenseScanner creates a default license scanner and exists
+// to export NewDefaultScanner to those who use Syft as a library.
+func NewDefaultLicenseScanner(o ...licenses.Option) (licenses.Scanner, error) {
```
kzantow, line 25: A user would not be able to use this due to the internal options. Also, it would not be ideal to return the internal `licenses.Scanner`. However, I would like to avoid exporting anything we don't need to, as it will become part of the Syft 1.x API henceforth.

syft/create_sbom.go
```diff
@@ -100,6 +100,11 @@ func setupContext(ctx context.Context, cfg *CreateSBOMConfig) (context.Context,
 }
 
 func setContextLicenseScanner(ctx context.Context, cfg *CreateSBOMConfig) (context.Context, error) {
+	// skip injecting a license scanner if one already set on context
+	if _, ok := ctx.Value(licenses.CtxKey).(licenses.Scanner); ok {
+		return ctx, nil
+	}
```
kzantow, line 106: What if we simply exported this function and changed the configuration from `*CreateSBOMConfig` to `cataloging.LicensesConfig`, which it looks like the only thing that is needed, below. You would be able to call this once and wouldn't need any further exported context management keys / etc.

kzantow (commented):
I think there may be an easier solution to export fewer things without the need to provide context keys and such

syft/create_sbom.go
adammcclenaghan: @kzantow Thank you - I like this idea too.

I've made the necessary changes, hopefully its in line with what you envisioned :)

spiffcs (approved):

spiffcs:
TY @adammcclenaghan - this change makes sense!

PR #3797, opened 2025-04-12
(fix): delete collection name/type key entries when empty

adammcclenaghan:
# Description
The collection used to store packages does not delete map keys from the `byName` or `byType` maps when the key's respective slice is emptied.

This is already done for the `byPath` map - this change fixes the `idsByName` and `idsByType` maps

## Type of change

# Checklist:


wagoodman (approved):

wagoodman (approved):
🙌

PR #3805, opened 2025-04-16
fix: Make Native Image contains no embedded SBOM Error Discoverable

### LINKED ISSUE #3731: Make it discoverable if Native Image contains no embedded SBOM
It should be easy for users to detect if `syft` encounters a Native Image without an embedded SBOM. This is important for security critical environments where it's required to enforce that Native Images contain embedded SBOMs to ensure vulnerability scanning remains functional. If a Native Image is created without an embedded SBOM, any vulnerable components will go undetected in vulnerability scanning.

Currently, the only possible way to get an _indication_ of the problematic situation of a Native Image without an embedded SBOM is by running syft in debug mode `-vv` and looking for [the Trace message](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go#L534) 
```
unable to extract SBOM from possible java native-image
```
This is not sufficient for two reasons:
* It's hidden in the debug logs.
* It doesn't differentiate between a Native Image without an embedded SBOM and any other executable file. The message will be the same for any other non-Native Image ELF, Mach-O, and PE file. 

**Why is this needed**:
To clearly signal to `syft` users of a potentially dangerous configuration of their Native Images. 

**Additional context**:
No comments.

sathiya06:
# Description

This PR improves Syft's visibility into native image binaries by surfacing cases where an expected embedded SBOM cannot be extracted. Previously, this failure was only shown in verbose debug logs and was indistinguishable from other non-native binaries. With this change, users will now receive a structured warning under the unknowns section of the SBOM output when a Java Native Image is found but no embedded SBOM could be retrieved.

- Fixes #3731 

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
+++ b/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
@@ -16,6 +16,7 @@ import (
 	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/internal/mimetype"
+	"github.com/anchore/syft/internal/unknown"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/file"
 	"github.com/anchore/syft/syft/format/cyclonedxjson"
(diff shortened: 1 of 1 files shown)


wagoodman:
shout out if this is ready for review or there is functionality you are still trying to incorporate

sathiya06:
@wagoodman This is ready for review.

wagoodman (approved):

PR #3821, opened 2025-04-23
PE cataloger should consider compile target paths from deps.json

wagoodman:
Today the PE cataloger considers the `runtime`, `resources`, and `native` target types within a deps.json file, however, the `compile` type is not considered:

>  If a given dependency is only listed for compilation, then its runtime, resources and native properties is omitted. Similarly if the dependency is only listed for runtime, then its compile property is omitted.

Per the [.NET CLI documentation](https://github.com/dotnet/cli/blob/v3.1.426/Documentation/specs/runtime-configuration-file.md).

This mirrors the same logic applied to the runtime and resources type to the compile type, allowing us to pair up deps.json entries that look like this with the DLLs they reference:
```json
... 
"DotNetNuke.Core/9.9.1": {
  "dependencies": {
    "DotNetNuke.DependencyInjection": "9.9.1"
  },
  "compile": {
    "lib/net45/DotNetNuke.dll": {},
    "lib/net45/Microsoft.ApplicationBlocks.Data.dll": {}
  }
},
...
```

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/dotnet/test-fixtures/image-net8-compile-target/Dockerfile
+++ b/syft/pkg/cataloger/dotnet/test-fixtures/image-net8-compile-target/Dockerfile
@@ -9,7 +9,8 @@ COPY src/*.cs .
 
 RUN dotnet publish -c Release --no-restore -o /app
 
-# note: we're not using a runtime here to make testing easier... but this functionally will not work if this was a produciton app
+# note: we're not using a runtime here to make testing easier... so you cannot run this container and expect it to work
+# we do this to keep the test assertions small since the don't want to include the several other runtime packages
 FROM busybox:latest
 
 WORKDIR /app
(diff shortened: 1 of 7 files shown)


syft/pkg/cataloger/dotnet/test-fixtures/image-net8-compile-target/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+WORKDIR /src
+
+COPY src/helloworld.csproj .
+RUN dotnet restore -r $RUNTIME
+
+COPY src/*.cs .
+
+RUN dotnet publish -c Release --no-restore -o /app
+
+# note: we're not using a runtime here to make testing easier... but this functionally will not work if this was a produciton app
```
willmurphyscode, line 12: I don't understand what this comment is trying to tell me.

willmurphyscode: I see. "this functionality will not work if this was a production app" sounds like the _cataloger_ wouldn't work on a production app, which really confused me.

You could say something like, "copy compiled app into empty container to prevent extraneous packages in the test case (the app itself wouldn't work, but the artifacts can still be cataloged)" or something like that?

wagoodman: adjusted

PR #3825, opened 2025-04-24
Add support for detecting javascript assets in .NET projects using libman

wagoodman:
When building an ASP .NET project it's fairly common to reference and bundle javascript assets. In the past these assets had been included by referencing nuget packages that were thin wrappers for the upstream javascript packages. However, more modern applications use [libman](https://learn.microsoft.com/en-us/aspnet/core/client-side/libman/libman-cli?view=aspnetcore-9.0) to describe bundled javascript dependencies. 

This PR adds support for reading these `libman.json` files within published application directories for cdn, unpkg, and other non-filesystem providers. The packages discovered are raised up as `NPM` package types with no additional metadata with a relationship between that package and the root package described in the `deps.json` sibling file.

To be clear, libman must be used in the .NET project in order for this kind of cataloging to work; the older [nuget-wrapped-javascript package references](https://www.nuget.org/packages/jquery/) will not be detected this way.

## Type of change

## Checklist

## PR stack

- #3822


syft/pkg/cataloger/dotnet/libman_json.go
```diff
@@ -0,0 +1,107 @@
...
+	"path"
+	"strings"
+
+	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/internal"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+type libmanJSON struct {
```
spiffcs, line 14: Found this as the current reference updated as recently as 5 hours ago. Might be a good inclusion as a comment in case we see any breakage in the future or if the version upgrades.
https://github.com/aspnet/LibraryManager/wiki/libman.json-reference

spiffcs (approved):
👍  - I added a comment that has where the libman.json reference docs is kept. Feel free to include it if you think we should track that in comments.

PR #3828, opened 2025-04-24
Resolve owned file paths when searching for overlaps

wagoodman:
Today we add relationships between packages that overlap in file ownership, say when the python cataloger finds a package at path `X` and the debian DB cataloger also finds the same package at the same file path `X` --in this case we leave both packages and add a `file-ownership-by-overlap` relationship between them. In the case where the same path conflict occurs and this is between a binary file and a debian package then we remove the binary package altogether.

The problem is that package ownership locations are unresolved strings (copied from the RPM DB or the Dpkg DB, etc) and the package locations are resolved real paths. This can cause issues int he case of https://github.com/anchore/grype/issues/2527 where the debian package writes through the `/bin` symlink to the realpath of `/usr/bin` to drop in the `gzip` binary.

This PR fixes this behavior by just-in-time attempting to resolve any paths from the `FileOwner`

- Fixes related issue https://github.com/anchore/grype/issues/2527

## Type of change

## Checklist

## TODO
- [x] Check the performance implications of putting path resolutions in the relationship creation path

### DIFF
--- a/internal/relationship/by_file_ownership.go
+++ b/internal/relationship/by_file_ownership.go
@@ -151,7 +151,6 @@ func allPaths(ownedFilePath string, resolver file.Resolver) []string {
 	for _, loc := range locs {
 		ownedFilePaths.Add(loc.RealPath)
 	}
-	ownedFilePaths.Remove("")
 	return ownedFilePaths.List()
 }
 
--- a/internal/relationship/by_file_ownership.go
+++ b/internal/relationship/by_file_ownership.go
@@ -9,6 +9,7 @@ import (
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/internal/sbomsync"
 	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/file"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/sbom"
 )
@@ -32,11 +33,11 @@ type ownershipByFilesMetadata struct {
 	Files []string `json:"files"`
 }
 
-func ByFileOwnershipOverlapWorker(accessor sbomsync.Accessor) {
+func ByFileOwnershipOverlapWorker(resolver file.Resolver, accessor sbomsync.Accessor) {
 	var relationships []artifact.Relationship
 
 	accessor.ReadFromSBOM(func(s *sbom.SBOM) {
-		relationships = byFileOwnershipOverlap(s.Artifacts.Packages)
+		relationships = byFileOwnershipOverlap(resolver, s.Artifacts.Packages)
 	})
 
 	accessor.WriteToSBOM(func(s *sbom.SBOM) {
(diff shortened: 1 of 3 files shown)


internal/relationship/by_file_ownership.go
```diff
@@ -32,11 +33,11 @@ type ownershipByFilesMetadata struct {
 	Files []string `json:"files"`
 }
 
-func ByFileOwnershipOverlapWorker(accessor sbomsync.Accessor) {
+func ByFileOwnershipOverlapWorker(resolver file.Resolver, accessor sbomsync.Accessor) {
```
spiffcs, line 36: Internal change - nonbreaking for consumers

internal/relationship/by_file_ownership.go
```diff
@@ -102,34 +103,58 @@ func findOwnershipByFilesRelationships(catalog *pkg.Collection) map[artifact.ID]
...
-			// look for package(s) in the catalog that may be owned by this package and mark the relationship
-			for _, subPackage := range catalog.PackagesByPath(ownedFilePath) {
-				subID := subPackage.ID()
-				if subID == id {
+		for _, ownedFilePath := range pkgFileOwner.OwnedFiles() {
+			// find the first path that results in a hit
+			for _, ownedPath := range allPaths(ownedFilePath, resolver) {
+				if matchesAny(ownedPath, globsForbiddenFromBeingOwned) {
+					// we skip over known exceptions to file ownership, such as the RPM package owning
+					// the RPM DB path, otherwise the RPM package would "own" all RPMs, which is not intended
```
spiffcs, line 112: 👍

internal/relationship/by_file_ownership.go
```diff
@@ -102,34 +103,58 @@ func findOwnershipByFilesRelationships(catalog *pkg.Collection) map[artifact.ID]
...
+			log.WithFields("error", err, "path", ownedFilePath).Trace("unable to find path for owned file")
+			locs = nil
+		}
+	}
+
+	ownedFilePaths := strset.New(ownedFilePath)
+	for _, loc := range locs {
+		ownedFilePaths.Add(loc.RealPath)
+	}
+	ownedFilePaths.Remove("")
```
spiffcs, line 154: small nit: maybe a comment here about why `""` is a potential `RealPath` that can be added here

internal/relationship/by_file_ownership.go
```diff
@@ -102,34 +103,58 @@ func findOwnershipByFilesRelationships(catalog *pkg.Collection) map[artifact.ID]
...
+	var err error
+	if resolver != nil {
+		locs, err = resolver.FilesByPath(ownedFilePath)
+		if err != nil {
+			log.WithFields("error", err, "path", ownedFilePath).Trace("unable to find path for owned file")
+			locs = nil
+		}
+	}
+
+	ownedFilePaths := strset.New(ownedFilePath)
```
spiffcs, line 150: if the resolver returns an error for `FilesByPath` do we still want `ownedFilePaths` to be created with `ownedFilePath` as an entry?

internal/relationship/by_file_ownership.go
wagoodman: actually, I should remove this since there is no risk of this

internal/relationship/by_file_ownership.go
wagoodman: indeed, that's the unresolved path that we look for today, which makes sense to do if we can't do anything else.

spiffcs (approved):

Issue #3575, opened 2025-01-09
syft convert cycloneDx Metadata is lost after reconvert


henrysachs:
**What happened**:

regarding #3574 after converting from cycloneDx to syft-json to cycloneDx the metadata object is empty. 

**What you expected to happen**:

I would love to keep the metadata somehow

**Steps to reproduce the issue**:

convert the following sbom https://github.com/CycloneDX/bom-examples/blob/master/SBOM/keycloak-10.0.2/bom.json from cdx to syft to cdx and inspect the metadata section

```bash
syft convert bom.json -o syft-json > bom-syft.json
syft convert bom-syft.json -o cyclonedx-json
```

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`:

```
Application: syft
Version:    1.18.1
BuildDate:  2024-12-13T18:41:10Z
GitCommit:  5e16e5031a13f8a11057feb8544decebfc43b4ed
GitDescription: v1.18.1
Platform:   darwin/arm64
GoVersion:  go1.23.4
Compiler:   gc
```

- OS (e.g: `cat /etc/os-release` or similar):

VictorHuu:
Hi @henrysachs , this issue is the very reason that causes #3574. So it's duplicated

henrysachs:
> Hi @henrysachs , this issue is the very reason that causes #3574. So it's duplicated

Thanks a lot for keeping an eye on this. Lets just close both when your PR has landed?

PR #3833, opened 2025-04-27
fix: [WIP]introduce Component Type Unknown in Cyclone DX for better conversion

### LINKED ISSUE #3574: Syft convert from cdx.json -> syft.json -> cdx.json fails
**What happened**:

I tried to convert a cyclone dx json to syft-json and back to cyclone dx, after that i received an error from syft 

```
[0000] ERROR failed to decode SBOM: unable to decode syft-json document: unable to find source metadata type=""
```

I receive a file in the CycloneDX JSON format

taken the example cyclonedx json from: https://github.com/CycloneDX/bom-examples/blob/master/SBOM/keycloak-10.0.2/bom.json

```
syft convert bom.json -o syft-json > bom-syft.json
syft convert bom-syft.json -o cyclonedx-json
```

I already dug a bit deeper into this and its because the whole source area is empty as soon as you write "directory" into type for example everything works

**Environment**:
- Output of `syft version`:

```
Application: syft
Version:    1.18.1
BuildDate:  2024-12-13T18:41:10Z
GitCommit:  5e16e5031a13f8a11057feb8544decebfc43b4ed
GitDescription: v1.18.1
Platform:   darwin/arm64
GoVersion:  go1.23.4
Compiler:   gc
```

MacOs 15.2

VictorHuu:
# Description

1. In case of unexpected CycloneDX ```ComponentType```, ```Unknown``` is used to prevent crash during conversion between Cyclone DX and other formats.
2. Though Syft only support some of the ```ComponentType``` like container and file, [other types](https://ecma-international.org/wp-content/uploads/ECMA-424_1st_edition_june_2024.pdf#page=46.53) defined in Cyclone DX Spec should also be included. Since most of these types share almost the same structure, a 'Unknown' type is enough to handle these types that Syft doesn't support now.

I'm not sure whether the solution is good or not,but I think a bit lossy information like 'application/library' to 'unknown' is acceptable.Otherwise,if we add types for every Component Type, quite a lot of even overwhelming test cases have to be updated.

Follow up: It seems that there are lots of issues related to the conversion, so this PR attempts to make the minimum changes in order not to introduce catastrophic results. 

- Fixes #3574 and #3575

## Type of change

# Checklist:


syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -40,6 +40,7 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 	// https://github.com/CycloneDX/specification/blob/master/schema/bom-1.3-strict.schema.json#L36
 	// "pattern": "^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
 	cdxBOM.SerialNumber = uuid.New().URN()
+	log.Warn("descriptor name:" + s.Descriptor.Name + ";source:" + s.Source.ID)
```
VictorHuu, line 43: This is simply due to my mistake, which will be removed soon

VictorHuu:
Opps! The latest error in Unit Test is weird,because the line number is wrong ,the testcase name doesn't exist ,but the test works well on my local PR: https://github.com/VictorHuu/syft/actions/runs/14735428241/job/41360196223.

VictorHuu:
Troubleshoot: The so-called error line [193](https://github.com/anchore/syft/actions/runs/14734760829/job/41357952382?pr=3833#step:4:2231) and the test case ```TestSource_UnmarshalJSON/empty``` don't exist in the original [file](https://github.com/anchore/syft/blob/bd5aa2da30fd5606f2d736d3b08809f5470432f1/syft/format/syftjson/model/source_test.go#L193). And the CI/CD for same commit works well in my [local PR](https://github.com/VictorHuu/syft/actions/runs/14735428241/job/41360196223).

Can someone help me troubleshoot the weird error if available?

syft/source/unknown_metadata.go
```diff
@@ -0,0 +1,16 @@
...
+// UnknownMetadata represents the CycloneComponentType that Syft doesn't support at present
+type UnknownMetadata struct {
+	UserInput   string                             `json:"name" yaml:"name"`
+	ID          string                             `json:"bom-ref" yaml:"bom-ref"`
+	Version     string                             `json:"version" yaml:"version"`
+	Group       string                             `json:"group,omitempty" yaml:"group,omitempty"`
+	Description string                             `json:"description,omitempty" yaml:"description,omitempty"`
+	PackageURL  string                             `json:"purl" yaml:"purl"`
+	Licenses    *cyclonedx.Licenses                `json:"licenses,omitempty" xml:"licenses,omitempty"`
+	ExternalRef *[]cyclonedx.ExternalReference     `json:"externalRef,omitempty" yaml:"externalRef,omitempty"`
```
kzantow, line 14: We won't be able to use CycloneDX data structures directly in public Syft APIs.

VictorHuu: Ok, I'll take it carefully. So these data structures needs to be tailored to store as much important infos as it can.

What about the following scheme?
```go
type UnknownMetadata struct {
......
	Licenses    *[]LicenseChoice         `json:"licenses,omitempty" xml:"licenses,omitempty"`
	ExternalRef *[]ExternalReference     `json:"externalRef,omitempty" yaml:"externalRef,omitempty"`
	Authors     *[]OrganizationalContact `json:"authors,omitempty" xml:"authors>author,omitempty"`
}

type OrganizationalContact struct {
	Name string `json:"name,omitempty" xml:"name,omitempty"`
}

type ExternalReference struct {
	URL    string  `json:"url" xml:"url"`
	Hashes *[]Hash `json:"hashes,omitempty" xml:"hashes>hash,omitempty"`
	Type   string  `json:"type" xml:"type,attr"`
}

type Hash struct {
	Algorithm string `json:"alg" xml:"alg,attr"`
	Value     string `json:"content" xml:",chardata"`
}

type LicenseChoice struct {
	License    *License `json:"license,omitempty" xml:"-"`
}

type License struct {
	ID   string `json:"id,omitempty" xml:"id,omitempty"`
}
```

henrysachs: @VictorHuu @kzantow sorry to push this but could you take a look into this as it would help us with the syft convert issue a lot!

henrysachs: I just took some time and had a look and the initial concern that the cyclonedx types would have been used directly should be fixed. @kzantow could you take a look at this again? we currently use some workarounds to fix the linked issues and I would love to get rid of them 😅

PR #3849, opened 2025-05-04
fix: indicate upstream packages for sbom cataloger

### LINKED ISSUE #3662: sbom cataloger returning upstream package
**What happened**:
When scanning `docker:cgr.dev/chainguard/curl:latest` with the sbom cataloger we are seeing the upstream package from an SBOM showing up "flat" in the SBOM. For example curl `8.12.0` and `8.12.0-r1` are both in the SPDX SBOM with 8.12.0-r1 being "Generated_from" 8.12.0. 

I would expect the "Generated_from" pacakge (8.12.0) to no be in the returned b/c the SPDX/SBOM isn't saying its in the image but that its generated from the package. 

Another option (maybe better id) would be to preserve that relationship and show it in the results so its clear that 8.12.0 isnt in the SBOM but that 8.12.0-r1 was generated from it. 

config.yaml
```
output: "json"
quiet: true
check-for-app-update: false
default-catalogers:
  - sbom-cataloger
  - apk-db-cataloger
```
```
syft -c config.yaml scan docker:cgr.dev/chainguard/curl:latest
```
**Environment**:
```
$ syft version
Application: syft
Version:    1.19.0
BuildDate:  2025-01-22T19:57:08Z
GitCommit:  222e6548a96f8c80015c1d24f01dea3052a04893
GitDescription: v1.19.0
Platform:   linux/amd64
GoVersion:  go1.23.4
Compiler:   gc
```

VictorHuu:
# Description

To include the GENERATED_FROM relationships is a good way to resolve the issue, (but there might be better solutions to it.)

- Fixes #3662

## Type of change

# Checklist:

### DIFF
--- a/syft/artifact/relationship.go
+++ b/syft/artifact/relationship.go
@@ -21,6 +21,9 @@ const (
 
 	// DescribedByRelationship is a proxy for the SPDX 2.2.2 DESCRIBED_BY relationship.
 	DescribedByRelationship RelationshipType = "described-by"
+
+	// GeneratedFromRelationship is a proxy for the SPDX 2.2.2 GENERATED_FROM relationship.
+	GeneratedFromRelationship RelationshipType = "generated-from"
 )
 
 func AllRelationshipTypes() []RelationshipType {
@@ -29,6 +32,7 @@ func AllRelationshipTypes() []RelationshipType {
 		ContainsRelationship,
 		DependencyOfRelationship,
 		DescribedByRelationship,
+		GeneratedFromRelationship,
 	}
 }
 
--- a/syft/format/common/spdxhelpers/to_syft_model.go
+++ b/syft/format/common/spdxhelpers/to_syft_model.go
@@ -391,6 +391,10 @@ func collectDocRelationships(spdxIDMap map[string]any, doc *spdx.Document) (out 
 			case helpers.ContainsRelationship:
 				typ = artifact.ContainsRelationship
 				to = toPackage
+			case helpers.GeneratedFromRelationship:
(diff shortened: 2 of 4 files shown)


kzantow:
Hey @VictorHuu -- thanks for working on this! We had a brief discussion on the team about whether or not we wanted to include things in our own data model that seem to be predominantly SPDX-only (although perhaps [CycloneDX formulation](https://cyclonedx.org/docs/1.6/json/#formulation) is similar?), and we had pros and cons for both as it relates to this particular situation.

One _con_ is that today, every package is supposed to be something Syft believes to be installed in the scanned target. By adding packages such as this, it breaks that assumption and makes downstream consumers more difficult, as they need to process relationships to determine if a package has some sort of exclusionary relation such as this. As you've astutely pointed out: "it's up to Grype not to make the upstream package false positive"

One _pro_ is that we are able to more accurately convert from SPDX to SPDX, but there are other ways to accomplish this and there are a number of other parts of SPDX that we probably will never include native support for in the Syft model, so this seems to be less important than the former point.

Would you be able to help me understand why you've opted for this change rather than filtering out these packages when reading SPDX or in the SBOM cataloger (which admittedly would need the relationship)? Are there any other changes that might work for your purposes?

kzantow:
Hi @VictorHuu -- we discussed this as a team and have path forward for this change. I made another PR which temporarily ignores the packages during import that we'll have to revert as part of this change, as well as make a separate location in the Syft SBOM for packages that describe things which are not considered to be present in the scanned system. More details are in the issue: https://github.com/anchore/syft/issues/3983

PR #3853, opened 2025-05-05
feat: purl-list format

kzantow:
# Description

This PR adds a `purl-list` format (with aliases: `purl` and `purls`). This allows a user to export and by association convert to and from a list of PURLs.

Additionally, after reading an SBOM, formats are able to "enhance" Syft package data from information in the PURL if the data isn't present in the package. For example: a `maven` PURL includes the GroupID, something like: `pkg:maven/org.apache/something`, but this GroupID was not being mapped to the internal data location for GroupID which could result in false positives in Grype. This happens for PURL input as well as other formats including SPDX and CycloneDX in order to improve support for SBOMs created with tools other than Syft.

- Fixes

## Type of change

# Checklist:

### DIFF
--- a/syft/format/internal/backfill.go
+++ b/syft/format/internal/backfill.go
@@ -1,0 +1,95 @@
+package internal
+
+import (
+	"fmt"
+
+	"github.com/anchore/packageurl-go"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+// Backfill takes all information present in the package and attempts to fill in any missing information
+// from any available sources, such as the Metadata and PURL.
+//
+// Backfill does not call p.SetID(), but this needs to be called later to ensure it's up to date
+func Backfill(p *pkg.Package) error {
+	if p.PURL == "" {
+		return nil
+	}
+	var errs error
+	pu, err := packageurl.FromString(p.PURL)
+	if err != nil {
+		errs = fmt.Errorf("unable to parse purl: %s: %w", p.PURL, err)
+	} else {
+		if p.Type == "" {
+			setTypeFromPurl(p)
+		}
+		if p.Language == "" {
+			setLanguageFromPurl(p)
+		}
+		if p.Name == "" {
+			setNameFromPurl(p, pu)
+		}
+		if p.Version == "" {
+			setVersionFromPurl(p, pu)
+		}
+		if p.Language == pkg.Java {
+			setJavaMetadataFromPurl(p, pu)
+		}
+	}
+	return errs
+}
+
+func setTypeFromPurl(p *pkg.Package) {
+	if p.Type == "" {
+		p.Type = pkg.TypeFromPURL(p.PURL)
+	}
+}
+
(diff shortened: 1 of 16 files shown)


syft/format/internal/backfill.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+}
+
+func setVersionFromPurl(p *pkg.Package, purl packageurl.PackageURL, epoch string) {
+	if p.Version == "" {
+		p.Version = purl.Version
+	}
+
+	if epoch != "" && p.Type == pkg.RpmPkg && !epochPrefix.MatchString(p.Version) {
```
wagoodman, line 120: missing a test case for RPM epoch

syft/format/internal/stream/seekable_reader.go
```diff
@@ -22,3 +22,47 @@ func SeekableReader(reader io.Reader) (io.ReadSeeker, error) {
...
+		if err != nil {
+			return 0, fmt.Errorf("cannot seek current: %w", err)
+		}
+		if currentOffset-o.offset+offset < 0 {
+			return 0, fmt.Errorf("cannot seek < 0")
+		}
+		newOffset, err := o.rdr.Seek(offset, io.SeekCurrent)
+		return newOffset - o.offset, err
+	}
+	return 0, fmt.Errorf("only SeekStart and SeekCurrent supported")
```
wagoodman, line 65: missing a test case for when whence is not seekstart or seekcurrent

syft/format/internal/stream/seekable_reader.go
```diff
@@ -12,7 +12,7 @@ func SeekableReader(reader io.Reader) (io.ReadSeeker, error) {
 	}
 
 	if r, ok := reader.(io.ReadSeeker); ok {
-		return r, nil
+		return newOffsetReadSeeker(r)
```
wagoodman, line 15: I think a comment here on what case this is trying to protect would be helpful

syft/format/purls/decoder.go
```diff
@@ -0,0 +1,80 @@
...
+	if r == nil {
+		return "", ""
+	}
+
+	buf := [4]byte{}
+	bufs := buf[:]
+	_, _ = r.Read(bufs)
+	if string(bufs) == "pkg:" {
+		return ID, version
+	}
```
wagoodman, line 43: [optional, could be implemented later] I think reading until we find the first non-whitespace character then checking the next 4 characters would be more resilient to cases where there are newlines or spaces at the front.

syft/pkg/url.go
```diff
@@ -10,6 +10,7 @@ import (
 
 const (
 	PURLQualifierArch   = "arch"
+	PURLQualifierCPES   = "cpes"
```
wagoodman, line 13: could use a test around CPEs as a qualifier (as horrifying as using it sounds 😳 )

wagoodman (approved):
Added a few comments, but crosses the threshold of mergable for me 👍

PR #3854, opened 2025-05-05
feat:parse golang test binaries and resolve deps

### LINKED ISSUE #3629: Support extracting go dependencies from test binaries

Test binaries, generated using `go test -c path/to/package`, seem to contain dependency information. I would like to be able to analyze these

**Why is this needed**:

I'm working on Syft-based SBOM tooling for a go monorepo. One of the things I'm attempting to achieve is to seperate out SBOMs per service, and additionally produce an SBOM for the tests of each service. The idea is to help devs with understanding if a given package update is going to affect a service, a test, or both

**Additional context**:

Running `go tool nm ./package.test` I see output like:
```
... lots of similar lines
metadata.test:  14c90e0 T type:.eq.google.golang.org/protobuf/encoding/protojson.MarshalOptions
... lots of similar lines
```

So to me, it _seems_ like this is doable, though it seems to not be supported today

VictorHuu:
# Description

I've come up with an elegant solution to the issue: use Go module cache or ```vendor/``` dir to resolve the dependencies appeared in the test binary. 

Notice: 
1. if the test binary is built with ```-ldflags="-s -w"``` which strips the ```symtab.``` section, this PR fails to detect most of the dependencies. 

- Fixes #3629 

## Type of change

# Checklist:


VictorHuu:
Limitations:
1. Can't handle the patterns which can be divided by '/' into 1 or multiple(>3,except those ending with v+num) parts

The core reason is that it's hard to determine the boundary between a module and its subdirectory names followed by the function name when the above cases occur. Though I've attempted to use path traversal to match directly, it's too costly.

Therefore, this PR is a best-effort PR at least by now

VictorHuu:
Does anyone have better solutions? I'm not sure if I'm on the right track. And I think it might be of low priority, because it's just about the local SBOM lookup of a test binary.

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -89,6 +116,277 @@ func getCryptoSettingsFromVersion(v version.Version) []string {
...
+			Path:    fmt.Sprintf("%s/%s", urlDir, nameWithoutVersion),
+			Version: Version,
+			Sum:     checksum,
+			Replace: nil,
+		}
+	}
+	return uniqueModules
+}
+
+func uncapitalize(name string) (newName string) {
```
kzantow, line 289: There is another function already doing this we should reuse: https://github.com/anchore/syft/blob/f9d0fa81ab6c2232df7b6b2d4b966797483f2ba5/syft/pkg/cataloger/golang/licenses.go#L293

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -89,6 +116,277 @@ func getCryptoSettingsFromVersion(v version.Version) []string {
...
+		urlDir, nameWithoutVersion := trimmedAsURL(sym.Name)
+		// the sym is not a mark of third-party function
+		if len(urlDir) == 0 {
+			continue
+		}
+		modPair := fmt.Sprintf("%s/%s", urlDir, nameWithoutVersion)
+		if _, exists := uniqueModules[modPair]; exists {
+			continue
+		}
+		var Version string
```
kzantow, line 333: nit: variables should be lowerCamelCase

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -2,46 +2,73 @@ package golang
...
+	"unicode"
 
-	"github.com/kastenhq/goversion/version"
-
+	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/internal/unknown"
 	"github.com/anchore/syft/syft/file"
 	"github.com/anchore/syft/syft/internal/unionreader"
+	"github.com/kastenhq/goversion/version"
```
kzantow, line 19: this should be causing a lint error; please run `make lint-fix`

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -2,46 +2,73 @@ package golang
...
+		var sym []elf.Symbol
+		// a test binary has no dependencies in "buildinfo", so we resort to the .symtab section (i.e. Symbols() )
+		if bi.Deps == nil {
+			sym, err = getSymbolsInfo(r)
+			if err != nil {
+				log.WithFields("file", location.RealPath, "error", err).Trace("unable to read golang symtab")
+				continue
+			}
+			if sym != nil {
+				btyp = testBinaryType
```
kzantow, line 69: It looks like the only 2 values returned are `devBinaryType` and `testBinaryType`. It seems these correspond to "dev" and "test" binaries, but what about "normal release" binaries? Release binaries are the primary thing the cataloger is intended to scan, there is probably just a naming thing we should consider changing here

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -89,6 +116,277 @@ func getCryptoSettingsFromVersion(v version.Version) []string {
...
+	if lastIndex != 0 {
+		urlDir = strings.Join(parts[:lastIndex], "/")
+	}
+	urlName = parts[lastIndex]
+	return
+}
+
+func findVersionInCache(basePath fs.FS, baseName string) (string, error) {
+	var res string
+	err := fs.WalkDir(basePath, ".", func(filePath string, _ fs.DirEntry, err error) error {
```
kzantow, line 172: instead of using WalkDir should this just iterate the direct subdirectories?

(same comment for other WalkDir calls)

kzantow (commented):
Hey @VictorHuu, I didn't quite grok the entirety of this PR, but I left a few specific comments and wanted to give you some general feedback. Generally, we want to use anything available to Syft in order to get the best possible SBOM, so from that perspective using the go mod cache / vendor dir / etc. is ok. However, the mod cache dir could have multiple versions, so we should probably take this into account and only surfacing version information if we aren't guessing between multiple.

It also looks like, if scanning a test binary, we are creating test dependency relationships. But many of these are not test dependencies, rather normal dependencies that are also used in tests, we probably don't want to label everything as a test dependency here, though I'm not sure how to differentiate. (Are the dependencies of a test binary technically not test dependencies but only test dependencies of the main package?)

If we're using the vendor dir, it's a case the user is not using go modules, but if they are using go modules we could probably get the version information from go.mod, rather than needing to look in the mod cache directory.

Additionally, by default Syft _only uses the source_ to scan and you must opt-in to enable scanning the host filesystem (e.g. go mod cache) using `--enrich` or other options, and if we're going to use the mod cache or a vendor dir outside of the scan target, this should similarly be opt-in (though if we're doing a directory scan with a vendor dir, that's obviously part of the scan target).

VictorHuu:
Hello,@kzantow, thanks for your reviews which I'll address soon.

And here's some of my opinions:
> If we're using the vendor dir, it's a case the user is not using go modules, but if they are using go modules we could probably get the version information from go.mod, rather than needing to look in the mod cache directory.

Maybe a test binary can be moved around even everywhere, and because the locations of go mod cache or ```vendor/``` are fixed, so this way is preferred. Anyway, to determine whether a test binary is generated where it should be is way too hard.

As for the discrimination between test dependencies and normal ones, we can use the following commands,(and there are corresponding go packages to execute it)
```
go list -json ./util | jq '.Imports, .TestImports, .XTestImports'
```
But as is implied previously, we must ensure the test binary is generated where it should be this way.

And for the accessibility to the host filesystem, the privilege to search within these dirs are the same as license resolver does, so separation is required.

VictorHuu:
Hi @spiffcs, I want to know whether this PR is pending or discarded? If it's discarded, is there any alternative to the solution? 
CC @kzantow

PR #3857, opened 2025-05-05
fix: use "contents" field and remove "fullText" license field 

spiffcs:
Before we do a release of syft we want to make sure the license fields are not duplicated. This removes `fullText` which was added https://github.com/anchore/syft/pull/3450. It routes the changes of #3450 through `contents` which was added previously here:  https://github.com/anchore/syft/pull/3366

Note to Reviewers: We previously included a struct tag for the contents field that ignored it as part of our `LicenseSet` identification process. This tag has been removed. Given that licenses can now be some combination of `UNKNOWN` -> `CONTENTS` where the difference in licenses is the "content" of their `contents` we could not longer continue ignoring this field. The resulting license sets would only allow for a single `UNKNOWN` license.

I have added a test that illustrates this behavior.

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/license_set_test.go
+++ b/syft/pkg/license_set_test.go
@@ -1,6 +1,7 @@
 package pkg
 
 import (
+	"os"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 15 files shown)


syft/pkg/license_set_test.go
```diff
@@ -160,3 +178,11 @@ func defaultLicenseComparer(x, y License) bool {
 		},
 	))
 }
+
+func readFileAsString(filepath string) string {
+	data, err := os.ReadFile(filepath) // Alternatively, ioutil.ReadFile in older Go versions
```
wagoodman, line 183:
```suggestion
	data, err := os.ReadFile(filepath)
```

wagoodman (approved):

PR #3858, opened 2025-05-05
Order locations by container layer order

### LINKED ISSUE #435: Provide a way to get the LayerID the package was first found in
A flag that would include the layerID the package first showed up in.

When tracking down a package (maybe b/c it has vulnerabilities or Im not sure why its in my SBOM) it would be helpful to know what layer it first showed up in so I can look at the commands run to generate that layer. 

Currently it looks like the layerID returned under locations is the last layerID the path was touched in. So for example if I did something like:
```
# layer1
FROM alpine
# layer2
RUN apk add --update nodejs-current=15.10.0-r0
```
The package “busybox” version "1.32.1-r6" has a layerID of layer2. I know there is the “–scope all-layers” option but that would also return any packages that were removed from the final image.

wagoodman:
This adjusts the output of any locations field in the JSON output to always order locations within the same evidence group to have the same layer ordering as seen in the container image.

So for instance, say we have a package with the following locations:
- `/b @ layer 1 [primary evidence]`
- `/b @ layer 2 [primary evidence]`
- `/c @ layer 1 [supporting evidence]`
- `/c @ layer 2 [supporting evidence]`
- `/a @ layer 1 [no evidence]`

... this would also be the new sorting order. 

This is a follow up to #3138  and #3855 so that consumers can now take the locations within each package location listing and be able to reason that earlier primary locations with the same path are overridden with later instances in the list with the same path (users can derive overriding precedence). 

This change applies to:
- package locations (`.artifacts[].locations`)
- package license locations (`.artifacts.licenses[].locations`)
- files section (`.files`)

Closes #435

## Type of change

## Checklist

### DIFF
--- a/syft/file/locations_test.go
+++ b/syft/file/locations_test.go
@@ -137,9 +137,9 @@ func TestLocationAndCoordinatesSorters(t *testing.T) {
 				},
 			},
 			wantLocs: []string{
-				"/c (/c) @ fsid-1 map[" + evidence.AnnotationKey + ":" + evidence.PrimaryAnnotation + "]",
-				"/b (/b) @ fsid-1 map[" + evidence.AnnotationKey + ":" + evidence.SupportingAnnotation + "]",
-				"/a (/a) @ fsid-1 map[" + evidence.AnnotationKey + ":]",
+				"/c (/c) @ fsid-1 map[evidence:primary]",
+				"/b (/b) @ fsid-1 map[evidence:supporting]",
+				"/a (/a) @ fsid-1 map[evidence:]",
 			},
 			wantCoords: []string{
 				"/a @ fsid-1",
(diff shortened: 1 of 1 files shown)


syft/format/syftjson/to_format_model.go
```diff
@@ -47,6 +49,16 @@ func ToFormatModel(s sbom.SBOM, cfg EncoderConfig) model.Document {
 	}
 }
 
+func getLayerOrderMap(s sbom.SBOM) map[string]int {
```
kzantow, line 52: I think we want to apply this same logic to SPDX and CycloneDX. It seems like having some of this logic only for syft json will mean we need to refactor it to apply to other formats.

syft/file/locations.go
```diff
@@ -33,3 +37,52 @@ func (l Locations) Less(i, j int) bool {
 func (l Locations) Swap(i, j int) {
 	l[i], l[j] = l[j], l[i]
 }
+
+type locationsByContainerOrder struct {
+	locations          []Location
+	layerOrderByDigest map[string]int
+}
+
+func LocationsByContainerOrder(locations []Location, layerOrderByDigest map[string]int) sort.Interface {
```
kzantow, line 46: Could this be internal? The `map[string]int` seems a bit brittle for a public API here.

syft/format/syftjson/to_format_model.go
```diff
@@ -278,6 +293,20 @@ func toPackageModel(p pkg.Package, cfg EncoderConfig) model.Package {
 	}
 }
 
+func toLocationsModel(locations file.LocationSet, layerOrderByDigest map[string]int) []file.Location {
```
kzantow, line 296: What if instead of dealing directly with the `layerOrderByDigest` this took a type-safe sort func, e.g.: `func(a, b file.Location) int`. This could allow us to construct a layer sorter and just pass the sort func reference, e.g. something along these lines:
```
func LayerSorter(layers []source.LayerMetadata) func(a, b file.Location) int
```
... and users call `slices.Sort` with it

syft/format/syftjson/to_format_model.go
wagoodman: I was hoping to kick that can, but I can take another stab at that

kzantow (approved):
LGTM 🎉

PR #3860, opened 2025-05-06
chore: update license sort to be stable with new only content

spiffcs:
# Description

https://github.com/anchore/syft/pull/3857 introduced an unstable sort case in the test where license that are composed of ONLY contents could be in different positions when calling `sets` to `slices`. This PR adds guarantees around comparing the first 10 non blank characters of a content field and sorting there as a last resort if other primary fields are left blank.

Fixes unstable sort seen here in automated PR update: https://github.com/anchore/syft/pull/3859

## Type of change

# Checklist:


syft/pkg/license.go
```diff
@@ -49,13 +49,18 @@ func (l Licenses) Less(i, j int) bool {
...
-				// While URLs and location are not exclusive fields
-				// returning true here reduces the number of swaps
-				// while keeping a consistent sort order of
-				// the order that they appear in the list initially
-				// If users in the future have preference to sorting based
-				// on the slice representation of either field we can update this code
-				return true
+				leftContents := firstTenNonEmpty(l[i].Contents)
+				rightContents := firstTenNonEmpty(l[j].Contents)
+				if leftContents == rightContents {
```
wagoodman, line 54: why look at the first 10 characters? we could compare full licenses since a `string < string` comparison would bail early on the first byte difference found.

spiffcs: I thought if we cut it off to first 10 non blank characters we could prevent any kind of performance issues where contents are extremely long and extremely similar.

I could compare the lengths of the strings which is pretty cheap and potentially prevent that kind of issue. 

Also happy to remove all of this and just do `string < string` if you feel it's a non issue here 😄

wagoodman: we might run into issues, but most license texts are relatively short (fit within a memory page). But since there could be license texts where the first 10 chars are the same but there is a difference later in the text, so it would still be unstable.

PR #3865, opened 2025-05-08
Persist artifact ID as supplemental package data

wagoodman:
Today if you use `syft convert`, the syft API (say via grype) with an SBOM, or use the SBOM cataloger, all packages raised up from the underlying SBOM get new IDs derived from the data discovered as opposed to using the ID found within the artifact. There are pros and cons with each approach, however, this PR is changing syft's opinion on this to prefer the ID of artifacts from the SBOM discovered.

This is done by adding a new `SupplementalData []any` field to `pkg.Package` which can be populated with any type. This is intended to introduce new data onto the package without encoding the field directly into the JSON format, without affecting the package ID hash, nor guaranteeing any type expectations around this field.

Here's an example of before and after of a grype run with these changes integrated; now the artifact IDs in the grype JSON are the native SBOM ID:

Screenshot 2025-05-07 at 10 13 04 PM

Supersedes #1872 and addresses https://github.com/anchore/grype/issues/1265 

## Type of change

## Checklist

### DIFF
--- a/syft/format/common/spdxhelpers/to_syft_model.go
+++ b/syft/format/common/spdxhelpers/to_syft_model.go
@@ -499,14 +499,15 @@ func extractPkgInfo(p *spdx.Package) pkgInfo {
 func toSyftPackage(p *spdx.Package) pkg.Package {
 	info := extractPkgInfo(p)
 	sP := &pkg.Package{
-		Type:     info.typ,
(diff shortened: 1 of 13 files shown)


wagoodman:
I've decided to incorporate the original approach proposed in https://github.com/anchore/syft/pull/1872

PR #3876, opened 2025-05-09
Detect license ID from full text when incidentally provided as a value

### LINKED ISSUE #3088: Detect whether full license text or a license name has been provided

spiffcs:
- detects when license values are really full text, attempts to identify the license ID, and migrate the contents to the `pkg.License.Content` field and fills in the ID to `pkg.License.Value`
- adds a new `license.content` configuration, deprecating the existing `license.include-unknown-license-content` configuration. The configuration allows for the following values:
  - `unknown` (default): only include contents for licenses that cannot be identified
  - `none`: never include license contents
  - `all`: include license contents wherever possible 
- replaces the `license.license-coverage` with `license.coverage`

```
# .syft.yaml configuration summary
license:
  # include the content of licenses in the SBOM for a given syft scan; valid values are: [all unknown none] (env: SYFT_LICENSE_CONTENT)
  content: 'none'

  # deprecated: please use 'license-content' instead (env: SYFT_LICENSE_INCLUDE_UNKNOWN_LICENSE_CONTENT)
  include-unknown-license-content:

  # adjust the percent as a fraction of the total text, in normalized words, that
  # matches any valid license for the given inputs, expressed as a percentage across all of the licenses matched. (env: SYFT_LICENSE_COVERAGE)
  coverage: 75

  # deprecated: please use 'coverage' instead (env: SYFT_LICENSE_LICENSE_COVERAGE)
  license-coverage:

```

## Development details

The current license constructors have been deprecated. The new constructors are copies of the old ones but `WithContext` and accept `context.Context` as their new initial argument.

By refactoring these constructors we can now access the `License Scanner` during license construction. This allows all catalogers the have a `file.ReadCloser` to query the license scanner during license construction. 

This enables #3088 to be solved. 

The new builder (under the ctx license constructors) has better logic surrounding how to detect if a metadata value is a license ID, some custom license title string, or actually the full contents itself. By detecting this at construction and running the scanner against metadata values that are likely license contents we can prevent `license.Value` from being populated with the full contents of the license text.

All catalogers now return the full content of the licenses discovered. These contents are dropped from the licenses in a post processing step.

The user is given the option to enable if they want the contents of licenses to appear in their SBOM.
```
None
Unknown
All
```

#### SPDX OtherLicenses
Given some of the changes made regarding how the values and IDs are set during license construction in this PR, the SPDX OtherLicenses format code has been moved to be done with SPDX package creation in syft's format modules. 

When SPDX packages are having their `concluded` and `declared` fields decorated, the spdx format code is now using the available license information to maintain a set of `spdx.OtherLicense` that will be returned with the packages.

This is different from our previous approach where `spdx.OtherLicense` would be recomputed after packages had already been assembled. This new approach is easier to test given that the licenses under packages are compared with `spdx.OtherLicense` when external validator tools are run against spdx formatted SBOM.

- Fixes #3088 

## Type of change

## Checklist


internal/licenses/context_test.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func TestSetContextLicenseScanner(t *testing.T) {
-	scanner := testScanner(true)
+	scanner := testScanner()
```
spiffcs, line 11: Scanner no longer maintains config about whether is should return content. It will always return evidence for a scanned license file.

cmd/syft/internal/options/license.go
```diff
@@ -5,7 +5,9 @@ import (
 )
 
 type licenseConfig struct {
+	// Deprecated: please use include-license-content instead
 	IncludeUnknownLicenseContent bool    `yaml:"include-unknown-license-content" json:"include-unknown-license-content" mapstructure:"include-unknown-license-content"`
```
wagoodman, line 9: we should add a postload check to see if both the legacy and new fields are set, it so error out with a helpful message to the user to migrate to the new option.

cmd/syft/internal/options/license.go
```diff
@@ -14,15 +16,17 @@ var _ interface {
 } = (*licenseConfig)(nil)
 
 func (o *licenseConfig) DescribeFields(descriptions clio.FieldDescriptionSet) {
-	descriptions.Add(&o.IncludeUnknownLicenseContent, `include the content of a license in the SBOM when syft
+	descriptions.Add(&o.IncludeUnknownLicenseContent, `deprecated: include the content of a license in the SBOM when syft
```
wagoodman, line 19: ideally we could remove the description for this so it doesn't show up in the config command output, but I don't think we have a way (yet) to tell clio/fangs to hide this field. It might be that we do with the right combination of `yaml` and `mapstructure` tag values (ignore yaml specify mapstructure), but I haven't tried this.

internal/licenses/scanner.go
```diff
@@ -8,33 +8,37 @@ import (
...
+	Start int               // Start offset of match in text; match is at text[Start:End].
+	End   int               // End offset of match in text.
+	IsURL bool              // Whether match is a URL.
+}
+
 type Scanner interface {
-	IdentifyLicenseIDs(context.Context, io.Reader) ([]string, []byte, error)
-	FileSearch(context.Context, file.LocationReadCloser) ([]file.License, error)
-	PkgSearch(context.Context, file.LocationReadCloser) ([]pkg.License, error)
+	FindEvidence(context.Context, io.Reader) ([]Evidence, []byte, error)
```
wagoodman, line 29: Why return the full contents?
```suggestion
	FindEvidence(context.Context, io.Reader) ([]Evidence, error)
```

This way you could change evidence to:
```
type Evidence struct {
	ID    string            // License identifier. (See licenses/README.md.)
	Type  licensecheck.Type // The type of the license: BSD, MIT, etc.
	Start int               // Start offset of match in text; match is at text[Start:End].
	End   int               // End offset of match in text.
    Contents []byte
	IsURL bool              // Whether match is a URL.
}
```

Worth asking and chatting about it before refactoring with this suggestion.

syft/cataloging/license.go
```diff
@@ -1,15 +1,46 @@
...
-import "github.com/anchore/syft/internal/licenses"
+import (
+	"github.com/anchore/syft/internal/licenses"
+)
+
+// IncludeLicenseContent controls whether license content should be included.
+type IncludeLicenseContent string
+
+const (
+	IncludeLicenseContentAll     IncludeLicenseContent = "all"
```
wagoodman, line 11: 3088 is really about trying to add license content for cases where it was found in the value and moved to the content field and using the scanner to fill in `value` with an ID where possible. I haven't looked over the full PR yet, but do we want to introduce `all` now in this PR?

syft/format/common/spdxhelpers/to_format_model_test.go
```diff
@@ -938,22 +941,25 @@ func Test_otherLicenses(t *testing.T) {
...
-			name:     "big licenses get hashed",
+			name:     "big licenses get hashed and space is trimmed",
 			packages: []pkg.Package{pkg3},
 			expected: []*spdx.OtherLicense{
 				{
-					LicenseIdentifier: "LicenseRef-e9a1e42833d3e456f147052f4d312101bd171a0798893169fe596ca6b55c049e",
-					ExtractedText:     bigText,
+					LicenseIdentifier: "LicenseRef-3f17782eef51ae86f18fdd6832f5918e2b40f688b52c9adc07ba6ec1024ef408",
+					// Carries through the syft-json license value when we shasum large texts
+					LicenseName:   "LicenseRef-sha256:3f17782eef51ae86f18fdd6832f5918e2b40f688b52c9adc07ba6ec1024ef408",
```
wagoodman, line 961: is the added `sha256:` intentional? are we detecting this or creating this? its odd that the identifier and name are different.

wagoodman: or is this an spdx limitation?

cmd/syft/internal/options/license.go
spiffcs: Yea +1 - I thought the file above would take care of a graceful upgrade of the value, but the new field will never be "". It will always default to `none` which means if someone sets the old value and just takes the default it will swallow their config silently. I'll add the postload.

cmd/syft/internal/options/license.go
spiffcs: I'll take a look to see what this could look like

syft/cataloging/license.go
spiffcs: I can remove this - It was part of my notes from the design session. Since it was the default case post package cataloging tasks(not modifying the contents) it felt easy to add, but I can remove it to reduce scope.

syft/format/common/spdxhelpers/to_format_model_test.go
spiffcs: This `LicenseRef-sha256:` is what would be in `Value` of a `syft-json` license if we found the content, could not identify the license, and needed to shasum it's contents for a value. We can totally drop it from the spdx model when translating a syft license -> spdx.

internal/licenses/scanner.go
wagoodman: from an offline chat: there is a need when there are no Evidence objects returned to reference the entire contents (as a sha256)

syft/format/internal/spdxutil/helpers/license.go
```diff
@@ -74,17 +87,22 @@ func ParseLicenses(raw []pkg.License) (concluded, declared []SPDXLicense) {
...
-	if l.SPDXExpression == "" {
-		candidate.Value = l.Value
+	var candidate SPDXLicense
+	candidate.ID = generateLicenseID(l)
+	// we have consumed a LicenseRef from the metadata of some package; we should encode all the info we have
+	if strings.Contains(candidate.ID, "LicenseRef-") {
+		candidate.LicenseName = l.Value
+		// we need to populate this field in the spdx document if we have a license ref
+		// 0..1 (Mandatory, one) if there is a License Identifier assigned (LicenseRef).
+		candidate.FullText = "UNKNOWN"
```
wagoodman, line 101: I think this should be NOASSERTION

syft/pkg/cataloger/generic/cataloger.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 // Processor is a function that can filter or augment existing packages and relationships based on existing material.
-type Processor func([]pkg.Package, []artifact.Relationship, error) ([]pkg.Package, []artifact.Relationship, error)
+type Processor func(context.Context, []pkg.Package, []artifact.Relationship, error) ([]pkg.Package, []artifact.Relationship, error)
```
wagoodman, line 19: this is a breaking change

syft/pkg/cataloger/golang/licenses.go
```diff
@@ -73,7 +63,7 @@ func newGoLicenseResolver(catalogerName string, opts CatalogerConfig) goLicenseR
 		opts:                  opts,
 		localModCacheDir:      localModCacheDir,
 		localVendorDir:        localVendorDir,
-		licenseCache:          cache.GetResolverCachingErrors[[]goLicense]("golang", "v1"),
+		licenseCache:          cache.GetResolverCachingErrors[[]pkg.License]("golang", "v1"),
```
wagoodman, line 66:
```suggestion
		licenseCache:          cache.GetResolverCachingErrors[[]pkg.License]("golang", "v2"),
```

Since you've changed the types we need to change the version here so that cache with the wrong data shape is not used.

syft/pkg/cataloger/golang/licenses_test.go
```diff
@@ -25,6 +25,14 @@ import (
 )
 
 func Test_LicenseSearch(t *testing.T) {
+	sc := &licenses.ScannerConfig{
+		CoverageThreshold: 75,
+		Scanner:           licensecheck.Scan,
+	}
+	scanner, err := licenses.NewScanner(sc)
+	require.NoError(t, err)
+	ctx := licenses.SetContextLicenseScanner(context.Background(), scanner)
```
wagoodman, line 34: we could probably add a helper function to `pkgtest`

cmd/syft/internal/options/license.go
spiffcs: I could not get this to work with a couple different combinations. I think it's something we might have to add to the clio library

cmd/syft/internal/options/license.go
spiffcs: Added a warning in postload to let the user know we are ignoring the deprecated value and to change over to the latest enum

syft/pkg/cataloger/generic/cataloger.go
spiffcs: @wagoodman this has been reverted

syft/pkg/cataloger/golang/licenses_test.go
spiffcs: Looks like we already had one. I've moved all instances of creating the scanner to consuming this already existing context function

syft/pkg/cataloger/golang/licenses_test.go
```diff
@@ -358,35 +366,31 @@ func Test_noLocalGoModDir(t *testing.T) {
...
-			URLs:           nil,
-			Locations:      file.NewLocationSet(file.NewLocation("LICENSE2")),
-			Contents:       "NVIDIA Software License Agreement and CUDA Supplement to Software License Agreement",
-		},
+func mustContentsFromLocation(t *testing.T, loc file.Location, offset ...int) string {
+	t.Helper()
+	contentsPath := "test-fixtures/licenses/pkg/mod/" + loc.RealPath
+	contents, err := os.ReadFile(contentsPath)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
spiffcs: Updated to use `require...`

syft/cataloging/license.go
spiffcs: Sync review we decided this was a free addition to we're keeping it in this PR

cmd/syft/internal/options/license.go
```diff
@@ -14,15 +17,25 @@ var _ interface {
...
 cannot determine a valid SPDX ID for the given license`)
+	descriptions.Add(&o.IncludeLicenseContent, `include the content of licenses in the SBOM for a given syft scan; valid values are: all, unknown, or none; default is none`)
 	descriptions.Add(&o.LicenseCoverage, `adjust the percent as a fraction of the total text, in normalized words, that
 matches any valid license for the given inputs, expressed as a percentage across all of the licenses matched.`)
 }
 
+func (o *licenseConfig) PostLoad() error {
+	if o.IncludeUnknownLicenseContent && (o.IncludeLicenseContent == "none" || o.IncludeLicenseContent == "all") {
+		log.Warnf("ignoring deprecated option: include-unknown-license-content=%t infavor of include-license-content=%s", o.IncludeUnknownLicenseContent, o.IncludeLicenseContent)
+		log.Warnf("please update your configuration to use include-license-content option")
```
wagoodman, line 30: this shouldn't be logging, but instead raise an error. If someone has conflicting configuration we should stop and not produce a result until the configuration is sorted.

wagoodman (approved):

PR #3900, opened 2025-05-15
fix: include default license config for sbom builder

spiffcs:
# Description

With the latest syft release of v1.24.0 some users were seeing contents of licenses in their default outputs. This change fixes the link between the lost config values so that contents can be dropped in a post processing step. It also updates the CLI test fixture to use an image that covers this case correctly

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### Demo showing new change
On branch
```json
      "licenses": [
        {
          "value": "LicenseRef-sha256:9e3a4384b6d8d2358d44103f62bcd948328b3f8a63a1a6baa66abeb43302d581",
          "spdxExpression": "",
          "type": "concluded",
          "urls": [],
          "locations": [
            {
              "path": "/usr/share/doc/libcom-err2/copyright",
              "layerID": "sha256:a8c68591d421fc2d4bdda704f67a796edf5ff880c59358d75107eb5261821650",
              "accessPath": "/usr/share/doc/libcom-err2/copyright",
              "annotations": {
                "evidence": "supporting"
              }
            }
          ]
        }
      ],

```

Syft v1.24.0
```json
      "licenses": [
        {
          "value": "LicenseRef-sha256:9e3a4384b6d8d2358d44103f62bcd948328b3f8a63a1a6baa66abeb43302d581",
          "spdxExpression": "",
          "type": "concluded",
          "urls": [],
          "locations": [
            {
              "path": "/usr/share/doc/libcom-err2/copyright",
              "layerID": "sha256:a8c68591d421fc2d4bdda704f67a796edf5ff880c59358d75107eb5261821650",
              "accessPath": "/usr/share/doc/libcom-err2/copyright",
              "annotations": {
                "evidence": "supporting"
              }
            }
          ],
[...]

PR #3903, opened 2025-05-15
fix: update applyLicenseContentRules to factor default case

spiffcs:
# Description
This PR updates the license config to account for a default case where users could potentially have misused the API and gotten the undesired result of contents being included for all licenses on all packages.

This PR also removes old fields that were previously not wired up correctly in the syft config. See: https://github.com/anchore/syft/pull/3900

## Type of change

# Checklist:

### DIFF
--- a/internal/task/package_task_factory.go
+++ b/internal/task/package_task_factory.go
@@ -277,17 +277,17 @@ func applyLicenseContentRules(p *pkg.Package, cfg cataloging.LicenseConfig) {
 	for i := range licenses {
 		l := &licenses[i]
 		switch cfg.IncludeContent {
-		case cataloging.LicenseContentIncludeUnknown:
-			// we don't have an SPDX expression, which means we didn't find an SPDX license
-			// include the unknown licenses content in the final SBOM
-			if l.SPDXExpression != "" {
-				licenses[i].Contents = ""
-			}
 		case cataloging.LicenseContentExcludeAll:
 			// clear it all out
 			licenses[i].Contents = ""
 		case cataloging.LicenseContentIncludeAll:
 			// always include the content
+		default:
+			// we don't have an SPDX expression, which means we didn't find an SPDX license
+			// include the unknown licenses content in the final SBOM
+			if l.SPDXExpression != "" {
+				licenses[i].Contents = ""
+			}
 		}
 	}
 
--- a/syft/cataloging/license.go
+++ b/syft/cataloging/license.go
@@ -14,10 +14,6 @@ const (
 )
 
 type LicenseConfig struct {
-	// IncludeUnknownLicenseContent controls whether the content of a license should be included in the SBOM when the license ID cannot be determined.
-	// Deprecated: use IncludeContent instead
-	IncludeUnknownLicenseContent bool `json:"-" yaml:"-" mapstructure:"-"`
-
 	// IncludeContent controls whether license copy discovered should be included in the SBOM.
 	IncludeContent LicenseContent `json:"include-content" yaml:"include-content" mapstructure:"include-content"`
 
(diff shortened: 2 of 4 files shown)


syft/cataloging/license.go
```diff
@@ -14,10 +14,6 @@ const (
 )
 
 type LicenseConfig struct {
-	// IncludeUnknownLicenseContent controls whether the content of a license should be included in the SBOM when the license ID cannot be determined.
-	// Deprecated: use IncludeContent instead
-	IncludeUnknownLicenseContent bool `json:"-" yaml:"-" mapstructure:"-"`
```
kzantow, line 19: Removing this is technically a breaking change, should we leave this deprecated? It looks like this option has been [in the codebase for a little while](https://github.com/anchore/syft/blob/e584c9f416028051fbdc80fffda4e624aaed528d/syft/cataloging/license.go)

syft/cataloging/license.go
```diff
@@ -27,7 +23,7 @@ type LicenseConfig struct {
 
 func DefaultLicenseConfig() LicenseConfig {
 	return LicenseConfig{
-		IncludeContent: LicenseContentExcludeAll,
+		IncludeContent: LicenseContentIncludeUnknown,
```
kzantow, line 26: I thought from the discussion we said the default should be to never include content but instead the licenses show up as `UNKNOWN`, otherwise users will potentially start getting very large SBOMs.

internal/task/package_task_factory.go
```diff
@@ -277,17 +277,17 @@ func applyLicenseContentRules(p *pkg.Package, cfg cataloging.LicenseConfig) {
...
 		switch cfg.IncludeContent {
-		case cataloging.LicenseContentIncludeUnknown:
-			// we don't have an SPDX expression, which means we didn't find an SPDX license
-			// include the unknown licenses content in the final SBOM
-			if l.SPDXExpression != "" {
-				licenses[i].Contents = ""
-			}
 		case cataloging.LicenseContentExcludeAll:
 			// clear it all out
 			licenses[i].Contents = ""
```
kzantow, line 282: Should we flag licenses that have contents but no SPDX expression as `UNKNOWN`/`NOASSERTION`/`cataloging.UnknownStubValue`/ or something here? (Maybe that's already happening and I just didn't find it)

syft/cataloging/license.go
wagoodman: we discovered it wasn't wired up, thus it wasn't actually usable. It is possible that someone was blindly using the struct, thus changing it would be a compile error, but seeing that doing so would have no effect. 

Given this, this is a pretty good time to remove it.

spiffcs: @wagoodman mentioned this in an offline sync. Because it was NOT wired up I think we're ok to remove this since it will not break anyone's behavior

syft/cataloging/license.go
wagoodman: I encouraged this change, however after thinking about it more, I agree with you and chris on this: we should leave this as ExcludeAll. The revised thinking: we should treat #3088 as a bug... meaning we should have excluded contents to begin with. Preserving the preexisting behavior isn't necessary.

internal/task/package_task_factory.go
wagoodman: @kzantow there are two places where this question applies: syft json and SPDX formats. For syft json there is no precedent for labeling licenses as "unknown", so I think that doesn't apply here. In terms of SPDX, this only applies for Other Licenses, which we area already claiming NOASSERTION when there are no contents (since we always include a LicenseRef- identifier).

kzantow: This was really more of a question than a request for a change, I'm ok with whatever behavior everyone else wants here, and agree maybe this is a SPDX concern, specifically.

PR #3923, opened 2025-05-20
fix(relationship): favor real paths over symlinks for ownership by file

### LINKED ISSUE #3847: Syft incorrectly reports multiple APKs as parents of symlinked files
**What happened**:

We have images which contain two APK packages: one ships (e.g., in the case I've been investigating) `/usr/bin/docker-machine-driver-harvester` and the other ships a symlink pointing to it at `/opt/drivers/management-state/bin/docker-machine-driver-harvester`.

`syft` has started incorrectly reporting both packages as the parent of `/usr/bin/docker-machine-driver-harvester`: this causes downstream issues because we have advisories for CVEs in `docker-machine-driver-harvester` issued for the binary-shipping package but (of course) not for the symlink-shipping package, meaning the image still appears vulnerable.

I've bisected and determined that https://github.com/anchore/syft/commit/03fa142de99e1c5694571ab535f65fed14aedcfa is the commit which caused this behaviour to start happening.

Syft should have continued reporting only the APK package which actually ships the binary as the parent.

The images in question are private, so I don't have a readily runnable reproducer to share.

To at least illustrate how to reproduce a little: while bisecting, I've been iterating with `go run cmd/syft/main.go -o json <affected image> > syft.json && python3 ~/scratch/rancher/reproducer.py syft.json` where `reproducer.py` is:

```py
import json
import sys

j = json.load(open(sys.argv[1]))

# Taken from Grype output
AID = '32b3882325e32538'

print("== ARTIFACTS ==")
for artifact in j["artifacts"]:
    if artifact["id"] == AID:
        print(artifact)

print()
print("== RELATIONS ==")
parents = []
for r in j['artifactRelationships']:
    if r['child'] == AID and r['type'] == 'ownership-by-file-overlap':
        print(r)
        parents.append(r['parent'])

print()
print("== PARENTS ==")
for artifact in j["artifacts"]:
    if artifact["id"] in parents:
        print(artifact['name'], artifact['type'])

if len(parents) != 1:
    sys.exit(1)
```

**Environment**:
- Output of `syft version`: observed on 1.23.1 but see above for bisection results
- OS (e.g: `cat /etc/os-release` or similar): Ubuntu, but seen in other environs as well

luhring:
# Description

This adjusts the way "ownership by file" relationships are formed, by preventing the scenario where a symlinked path can cause two packages to "own" the same package (one via the real path, and the other via the symlink). Now, when we're about to form a relationship via a symlink, we first check to see if there's a non-symlink ownership that exists, which should prevent the symlink-based relationship from being created.

- Fixes #3847 

## Type of change

# Checklist:


internal/relationship/by_file_ownership.go
```diff
@@ -135,6 +162,43 @@ func findOwnershipByFilesRelationships(resolver file.Resolver, catalog *pkg.Coll
 	return relationships
 }
 
+func directOwnersByPath(catalog *pkg.Collection) map[string]map[string][]artifact.ID {
+	directOwnership := make(map[string]map[string][]artifact.ID)
```
kzantow, line 166: What would you think about simplifying this to: `map[string][]pkg.Type`? I think it could make both the population and usage above simpler, I think.

kzantow: Answer from team discussion: it seems more-or-less the safest to check within the same type, as implemented here 👍

luhring: Okay cool! Sorry for the delay, been at an in-person summit thing this week. Yeah I could go either way too, but it seemed more semantically correct in my head (for hand-waivy reasons) for the interpretation of symlinks for a given installation DB to be scoped to that same DB

kzantow: @luhring do you mind if I push a couple tweaks to this for performance/clarity?

luhring: Not at all! Thanks! 🙏

kzantow (approved):
LGTM, thanks @luhring! 👍

PR #3929, opened 2025-05-21
Add ability to scan snaps (as a source)

### LINKED ISSUE #1088: add native support for flatpak and snap images
**What would you like to be added**: support for pulling flatpak and snap images from repos for scanning

**Why is this needed**:  scan packages from snap and flatpak without needing to install them first

**Additional context**:

It is possible to scan flatpak and snap images with some level of success once they have been installed using the dir: specifier.  It would be neat if syft (and grype) knew how to fetch these packages directly (from original and alternative sources) and generate an sbom without needing to 'install' them first.

wagoodman:
This adds support for cataloging single snap files, cataloging all contents. This can scan a local snap file:

```
syft /Users/wagoodman/OrbStack/ubuntu/home/wagoodman/snaps/etcd.snap
```

or grab the snap [from snapcraft](https://snapcraft.io/etcd):
```
# assumes latest stable with amd64 arch
syft --from snap etcd

# specify a channel and arch
syft --from snap etcd@3.1/stable --platform i386
```

Screenshot 2025-05-23 at 3 20 15 PM

Screenshot 2025-05-23 at 3 21 06 PM

Note: if `--platform` is given an OS then the source will error out and not continue (as this is not a valid form).

- Fixes #1088

## Type of change

## Checklist

## TODO
- [x] Add support for remote snaps
- [x] Add LZO support (pending https://github.com/diskfs/go-diskfs/pull/298)
- [x] Capture `snap/manifest.yaml` and surrounding data onto `source.Metadata`
- [x] Confirm that only the `install` cataloger set is being used for this new source
- [x] Rebase after merging https://github.com/anchore/syft/pull/3953


popey:
I'm seeing this a lot:

```
[0002] ERROR could not determine source: an error occurred attempting to resolve 'adguard-home': snap: unable to create snap file resolver: failed to walk squashfs file="/tmp/syft-snap-653810807/UXZIkJfJT2SPCGejjnSjOBqJ71yHk8bw_8117.snap": path="/" already exists but is NOT a regular file
```

Yet:

```
ls -l /tmp/syft-snap-653810807/UXZIkJfJT2SPCGejjnSjOBqJ71yHk8bw_8117.snap
-rw-rw-r-- 1 alan alan 9334784 May 27 15:31 /tmp/syft-snap-653810807/UXZIkJfJT2SPCGejjnSjOBqJ71yHk8bw_8117.snap
```

```
file /tmp/syft-snap-653810807/UXZIkJfJT2SPCGejjnSjOBqJ71yHk8bw_8117.snap
/tmp/syft-snap-653810807/UXZIkJfJT2SPCGejjnSjOBqJ71yHk8bw_8117.snap: Squashfs filesystem, little endian, version 4.0, xz compressed, 9332165 bytes, 8 inodes, blocksize: 131072 bytes, created: Tue May 27 12:20:03 2025
```

wagoodman:
@popey pointed out that there are region locked snaps that fail with obtuse errors:
```
[0000] ERROR could not determine source: an error occurred attempting to resolve 'jp-ledger': snap: unable to open snap manifest file: failed to resolve request: failed to get download URL for snap "jp-ledger": API request failed with status code 404
```

We should probably use the [search API](https://api.snapcraft.io/docs/search.html#snaps_find) to resolve an exact name match then extract the download link from there (search with `q=NAME` and only take results that locally match the given name exactly... or something similar). Then we could get an error message more like:

```
found snap 'jp-ledgrer' (id=jyDlMmifyQhSWGPM9fnKc1HSD7E6c47e) but it is unavailable for download
```

internal/file/getter.go
wagoodman, line 1: this was ported from grype -- but we should consider breaking this out into its own module somewhere since it wires up progress concerns nicely with events for bubbletea

syft/format/common/spdxhelpers/to_format_model.go
```diff
@@ -30,13 +30,15 @@ import (
 const (
 	noAssertion = "NOASSERTION"
 
+	spdxPrimaryPurposeArchive   = "ARCHIVE"
 	spdxPrimaryPurposeContainer = "CONTAINER"
```
wagoodman: I could be convinced that a snap should be considered a container in the context of a primary purpose (as it's more accurate than just "archive"). Any thoughts here?

kzantow: "container" description is: the Element is a container image which can be used by a container runtime application. is there some sort of snap runtime? if not, archive does seem more accurate to me, or maybe even application, despite it being an archive that contains an application? a quick naïve search of the snap store for "mail" looks like each snap is essentially an application, or is that not really true?

popey: I'd lean toward "CONTAINER" for snaps in the SPDX context. The key distinction is that snaps aren't just bundled files - they're executable environments with runtime characteristics. When building SBOMs, we're likely more concerned with the software composition and security boundaries than the storage format.
The fact that snaps provide isolation, have their own security model, and encapsulate both the application and its runtime dependencies makes them functionally closer to containers than traditional archives.

wagoodman: agreed on all fronts -- updating now

cmd/syft/internal/commands/scan.go
```diff
@@ -190,6 +190,10 @@ func runScan(ctx context.Context, id clio.Identification, opts *scanOptions, use
 		return err
 	}
 
+	if src == nil {
+		return fmt.Errorf("unable to determine source for %q", userInput)
```
kzantow: I'm a little surprised `syft.GetSource` doesn't return an error

syft/get_source_test.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+
+		err := validateSourcePlatform(nil, cfg)
+		if err != nil {
+			t.Errorf("Expected no error for nil source, got: %v", err)
+		}
+	})
+
+	t.Run("nil platform config", func(t *testing.T) {
+		tests := []struct {
```
kzantow: nit: I find it harder to find the code responsible for the testing logic when there are nested `t.Run` and within those nested tables. To me, at least, it would be nice if table tests were only in top-level test functions. In other words -- it makes things easier to find in the future to just look at the end of a function to see the testing logic rather than needing to scan through a bunch of nested testing functions, missing the end, getting confused. For example, I was confused at first why there weren't some obvious positive tests in this function. no change really requested, just perhaps something to think about (I've noticed AI generating things like this quite often).

Unless there is a need like having a common setup, I would - generally speaking - prefer to avoid nested test functions in favor of top-level test functions.

syft/internal/unionreader/union_reader.go
```diff
@@ -75,3 +81,49 @@ func GetUnionReader(readerCloser io.ReadCloser) (UnionReader, error) {
...
+		ReadSeekCloser: rs,
+		mu:             &sync.Mutex{},
+	}
+}
+
+func (r *readerAtAdapter) ReadAt(p []byte, off int64) (n int, err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	currentPos, err := r.Seek(0, io.SeekCurrent)
```
kzantow: This should probably be tracked when the reader is created -- I don't think getting current position here will always be the right thing to do. I think this existing seekable reader behavior is close, albeit it's not ReaderAt but ReadSeeker: https://github.com/anchore/syft/blob/c36c69779aa5e2e5154ef74eadc6407643d8169b/syft/format/internal/stream/seekable_reader.go#L14

syft/pkg/cataloger/python/parse_wheel_egg_record.go
```diff
@@ -21,12 +21,19 @@ func parseWheelOrEggRecord(reader io.Reader) []pkg.PythonFileRecord {
...
+				break
+			}
+			var parseErr *csv.ParseError
+			if errors.As(err, &parseErr) {
+				log.WithFields("error", parseErr).Debug("unable to read python record entry (skipping entry)")
+				continue
+			}
+
+			// probably an I/O error, so we log it and stop processing
+			log.WithFields("error", err).Debug("unable to read python record file")
```
kzantow: Should these result in an unknown error?

syft/source/snap_metadata.go
```diff
@@ -0,0 +1,12 @@
...
+
+import "github.com/anchore/syft/syft/file"
+
+type SnapMetadata struct {
+	Summary       string        `yaml:"summary" json:"summary,omitempty"`
+	Base          string        `yaml:"base" json:"base,omitempty"`
+	Grade         string        `yaml:"grade" json:"grade,omitempty"`
+	Confinement   string        `yaml:"confinement" json:"confinement,omitempty"`
+	Architectures []string      `yaml:"architectures" json:"architectures,omitempty"`
+	Digests       []file.Digest `yaml:"digests" json:"digests,omitempty"`
```
kzantow: Since this is fresh in your mind, it might be good to add some comments about what these fields are, at least Base, Grade, and Confinement

syft/source/snapsource/snap.go
```diff
@@ -0,0 +1,260 @@
...
+	return newSnapFileFromRemote(ctx, fs, cfg, getter, info)
+}
+
+func newSnapFileFromRemote(ctx context.Context, fs afero.Fs, cfg Config, getter intFile.Getter, info *remoteSnap) (*snapFile, error) {
+	t, err := afero.TempDir(fs, "", "syft-snap-")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp directory: %w", err)
+	}
+
+	snapFilePath := filepath.Join(t, filepath.Base(info.URL))
```
kzantow: `filepath.Base` -> `path.Base`

syft/source/snapsource/snap.go
```diff
@@ -0,0 +1,260 @@
...
+	var architecture string
+	if cfg.Platform != nil {
+		architecture = cfg.Platform.Architecture
+	}
+
+	if architecture != "" {
+		return nil, fmt.Errorf("architecture cannot be specified for local snap files: %q", cfg.Request)
+	}
+
+	absPath, err := filepath.Abs(cfg.Request)
```
kzantow, line 111: `fileExists` uses `homedir`, but I don't see it being used here

syft/source/snapsource/snap_source.go
```diff
@@ -0,0 +1,316 @@
...
+	manifest         snapManifest
+	digests          []file.Digest
+	fs               filesystem.FileSystem
+	squashfsPath     string
+	squashFileCloser func() error
+	closer           func() error
+}
+
+func New(cfg Config) (source.Source, error) {
+	client := intFile.NewGetter(cfg.ID, cleanhttp.DefaultClient())
```
kzantow, line 103: I think we have some client configuration and I saw some testing about it, but it looks like we're not passing any of that in? maybe it's in the wrong part of the config to be applicable to these http requests?

syft/source/snapsource/snap_source.go
```diff
@@ -0,0 +1,316 @@
...
+	f, err := getSnapFile(context.Background(), afero.NewOsFs(), client, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	s := &snapSource{
+		id:           deriveID(cfg.Request, cfg.Alias.Name, cfg.Alias.Version),
```
kzantow: It looks like this calculates a digest but we already did associated with `f`, could we consolidate this or skip digest for the ID? Isn't there a name? not sure what the ID is used for

syft/source/snapsource/snap_source.go
```diff
@@ -0,0 +1,316 @@
...
+
+func deriveID(path, name, version string) artifact.ID {
+	info := fmt.Sprintf("%s:%s@%s", digestOfFileContents(path), name, version)
+	return internal.ArtifactIDFromDigest(digest.SHA256.FromString(info).String())
+}
+
+func digestOfFileContents(path string) string {
+	file, err := os.Open(path)
+	if err != nil {
+		return digest.SHA256.FromString(path).String()
```
kzantow: is this for naming uniqueness or reproducibility; if the former consider xxh?

syft/source/snapsource/snap_source.go
```diff
@@ -0,0 +1,316 @@
...
+	if mimeType == "application/vnd.squashfs" || mimeType == "application/x-squashfs" {
+		return true
+	}
+
+	ext := filepath.Ext(path)
+	return ext == ".snap" || ext == ".squashfs"
+}
+
+func deriveID(path, name, version string) artifact.ID {
+	info := fmt.Sprintf("%s:%s@%s", digestOfFileContents(path), name, version)
```
kzantow: seems like `path.Base(path)` might be a useful default here somewhere? and usually config aliases aren't set so name and version will be empty. e.g. `<name | filename>@<version | hash>` or something? this may not make a bit of difference, I didn't quite understand what the `id` is used for

syft/source/sourceproviders/source_providers.go
```diff
@@ -24,11 +26,18 @@ func All(userInput string, cfg *Config) []collections.TaggedValue[source.Provide
 	stereoscopeProviders := stereoscopeSourceProviders(userInput, cfg)
 
 	return collections.TaggedValueSet[source.Provider]{}.
+		// try all specific and local sources first...
 		// --from file, dir, oci-archive, etc.
 		Join(stereoscopeProviders.Select(FileTag, DirTag)...).
+
+		// --from snap (remote and local)
+		Join(tagProvider(snapsource.NewSourceProvider(userInput, cfg.Exclude, cfg.DigestAlgorithms, cfg.Alias), SnapTag)).
```
kzantow: I think users scanning a file or directory will be more common than expecting to scan a remote snap. we should probably test all local files / directories and pull sources before making a network request for a snap, unless I've missed something and it only happens with certain uncommon name patterns.

syft/source/snapsource/snap.go
wagoodman: I'll do the homedir / cleaning earlier in processing

syft/source/snapsource/snap_source.go
wagoodman: the goal is to get a stable ID for the source when making relationships between other elements (packages / files) and the source. 

Ideally this is content addressable -- regardless of the name of the file or any aliases used, we will get the same ID for the same file every time. However, we've found that for other kinds of sources that if there is a different logical name or version then we should have separate IDs. 

This is mimicking the ID derivation from the file source.

syft/source/sourceproviders/source_providers.go
wagoodman: however, the generic file cataloger will always succeed in cases where this is a local snap and `--from`/scheme was left out, so the (more specific) snap provider has to be attempted before at least the file provider.

kzantow: Right, I guess what I was asking / thinking: we should separate this into local, which happens before the other file sources, and remote, which happens after the pull sources. As an example: what happens when I scan `mysql`? Before it would look for `library/mysql:latest`, and now it will give me `mysql` snap results, right? or am I missing something?

syft/internal/unionreader/union_reader.go
wagoodman: This is attempting to adapt the seeking ability (stateful) with a reader-at like experience (stateless), but doing so affects the pointer into the file, which would affect subsequent reads and seeks, so the first step is to keep the original position in order to restore it later in this call.

I think there is a problem in the sense that we are only locking for ReadAt, which in hindsight doesn't make sense because we're not locking for other operations.

syft/source/snapsource/snap_source.go
wagoodman: do you mean something at the top-level syft config? or something locally?

kzantow: There are a bunch of tests added in this getter_tests.go: https://github.com/anchore/syft/pull/3929/files#diff-47eefc3f0d47502fae4265d4167e1472ddf407138623e146b526b453957e1e88R21 that look like they are dealing with various configurations: https://github.com/anchore/syft/pull/3929/files#diff-47eefc3f0d47502fae4265d4167e1472ddf407138623e146b526b453957e1e88R45, but it doesn't look like we are providing the ability to set any of the same options? I didn't quite follow the flow here, so feel free to ignore any comments if I've misunderstood.

Also, we are setting similar options for registry access:
```
registry:
  insecure-skip-tls-verify: false
  insecure-use-http: false
  auth: []
  ca-cert: ''
```
... which admittedly are registry options, but seem like some could be pertinent to other remote URLs.

syft/internal/unionreader/union_reader.go
kzantow: Let's say I have this reader, and I call `.Read(bytes)` and then I call `.ReadAt(0,...)`, it will not have the correct starting position of the reader. I think, if we track the starting position when this custom reader is created, it should be reasonably assumed to be the "start of the stream". Happy to chat on this further, I was just pointing out it's a similar problem to the other reader I mentioned above

syft/source/sourceproviders/source_providers.go
wagoodman: ah that makes sense, I'll split this up -- good catch

syft/internal/unionreader/union_reader.go
wagoodman: The definition of ReaderAt is that the offset is absolute, not relative to any other read calls --we're only preserving the pointer state for other read/seek calls. That is, it could be that the file pointer is specifically there to subset a selection of the file, in which case the underlying reader should be wrapped with [io.SectionReader](https://pkg.go.dev/io#SectionReader) or similar, where subsetting is explicit (and you can get the underlying non-section reader with `.Outer()`). However, us creating this wrapper around a io.ReadSeeker should not implicitly mean that we are intending to act like a io.SectionReader.

syft/source/snapsource/snap_source.go
wagoodman: it sounds like we could be using a unified configuration for proxy/tls/etc details for any http/https based request. I don't think I want to try and introduce that here though.

syft/internal/unionreader/union_reader.go
wagoodman: tangent, I did update this to account for locking for other read/seek calls

PR #3936, opened 2025-05-27
Harden container image by adding non-root user and dropping root priv…

MikeTheCyberGuy:
# Description

### Summary

This PR hardens the `syft` container by introducing a non-root user and setting it as the default runtime user. This aligns with container security best practices and reduces the potential impact of a container compromise.

### Details

- Introduced an intermediate Alpine stage (`security_provider`) to create a `nonroot` user and group.
- Copied `/etc/passwd` from the Alpine stage to enable non-root user resolution in the final `distroless` image.
- Set `USER nonroot` in the final stage to enforce least-privilege execution at runtime.
- Added `syft/syft` to `.gitignore` to avoid committing the local build artifact.

### Security Impact

Running containers as root can allow for privilege escalation and container escape if the application is compromised, particularly in multi-tenant Kubernetes clusters or untrusted workloads. This change follows container security best practices by enforcing a non-root runtime, and aligns with recommendations in the CIS Docker Benchmark and common container hardening guidelines.

No functional behavior has changed; this is a defense-in-depth measure to reduce risk in production deployments.

## Type of change

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,5 @@ cosign.pub
 __pycache__/
 *.py[cod]
 *$py.class
+
+syft/syft
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,19 @@
 FROM gcr.io/distroless/static-debian12:latest AS build
 
-FROM scratch
-# needed for version check HTTPS request
+FROM alpine:latest AS security_provider
+RUN addgroup -S nonroot && adduser -S nonroot -G nonroot
+
+FROM gcr.io/distroless/base-debian12
+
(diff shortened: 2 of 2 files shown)


MikeTheCyberGuy:
This PR addresses and resolves the hardening changes discussed under the private GitHub Security Advisory [GHSA-m5jj-hv49-xgrh]. As confirmed in our previous discussion, this issue is not considered a vulnerability but is accepted as a defense-in-depth improvement.

Once this PR is approved and merged, I will close out the security advisory.

Let me know if anything else is needed to move this forward.

Thanks!

PR #3941, opened 2025-05-28
Harden Container Runtime with Non-Root User

MikeTheCyberGuy:
- uses `gcr.io/distroless/static-debian12` as the base image
- ensures the built containers are non-root users

In the process of working on this additional docker manifest updates were made:
- `image_templates` section was missing from the `debug` variant
- the top-level images should be the manifests and the tagged images should always be architecture specific (according to best practices)

## Changes

## Checklist

### DIFF
--- a/.github/workflows/validations.yaml
+++ b/.github/workflows/validations.yaml
@@ -63,18 +63,18 @@ jobs:
 
 
   Build-Snapshot-Artifacts:
-    name: "Build snapshot artifacts"
     runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
-
+      - uses: actions/checkout@v4
+      - name: Enable QEMU for multi-arch
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Buildx (needed by GoReleaser)
+        uses: docker/setup-buildx-action@v3
       - name: Bootstrap environment
         uses: ./.github/actions/bootstrap
-        with:
-          bootstrap-apt-packages: ""
-
       - name: Build snapshot artifacts
         run: make snapshot
+
 
       # why not use actions/upload-artifact? It is very slow (3 minutes to upload ~600MB of data, vs 10 seconds with this approach).
       # see https://github.com/actions/upload-artifact/issues/199 for more info
--- a/.github/workflows/validations.yaml
+++ b/.github/workflows/validations.yaml
@@ -63,18 +63,28 @@ jobs:
 
 
   Build-Snapshot-Artifacts:
+    name: "Build snapshot artifacts"
     runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@v4
-      - name: Enable QEMU for multi-arch
-        uses: docker/setup-qemu-action@v3
-      - name: Set up Buildx (needed by GoReleaser)
-        uses: docker/setup-buildx-action@v3
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #v4.2.2
+
       - name: Bootstrap environment
         uses: ./.github/actions/bootstrap
+        with:
+          bootstrap-apt-packages: ""
+
+      - name: Enable QEMU for multi-arch
+        uses: docker/setup-qemu-action@976512e99f06adc0ee8522552d4506bb9e4e2cbe   # v3 pinned
+        with:
+          platforms: all
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@0fb43db2cebcdcac6112ee20be2c60dfe4e86193 # v3 pinned
+        with:
+          builder: multiarch
+
       - name: Build snapshot artifacts
         run: make snapshot
-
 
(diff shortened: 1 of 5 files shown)


spiffcs:
Running the checks on this now -- thank you for taking the time to improve the security posture of our images!
@anchore/tools do we want to offer these as new separate images or is worth the cost of breaking user pipelines here?

We might have to ensure file permissions are compatible here and set the user before the copy:
`RUN chown -R nonroot:nonroot /tmp /syft /other-paths` checking this now.

.github/workflows/validations.yaml
```diff
@@ -73,6 +73,16 @@ jobs:
...
+      - name: Enable QEMU for multi-arch
+        uses: docker/setup-qemu-action@v3
+        with:
+          platforms: all
+      
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          builder: multiarch
+
```
wagoodman: This used to not be necessary since we were taking the cross-built binaries, baking separate images, and crafting the manifest to point to each tag. However, since this changes to a distroless build I can see this has changed some. I initially didn't see why we needed qemu in this branch since we really don't need emulation to copy in pre-built binaries ... but I think the crux is the `addgroup` command, right? Since we're executing a binary and that binary is platform dependent qemu is needed.

If true, then there is an opportunity here to simplify the goreleaser file a lot as well as stop publishing separate tags for these platforms. Something like this:
```yaml

dockers:
  # debug images
  - image_templates:
      - anchore/syft:debug
      - anchore/syft:{{.Tag}}-debug
      - ghcr.io/anchore/syft:debug
      - ghcr.io/anchore/syft:{{.Tag}}-debug
    dockerfile: Dockerfile.debug
    use: buildx
    build_flag_templates:
      - "--platform=linux/amd64,linux/arm64,linux/ppc64le,linux/s390x"
      - "--build-arg=BUILD_DATE={{.Date}}"
      - "--build-arg=BUILD_VERSION={{.Version}}"
      - "--build-arg=VCS_REF={{.FullCommit}}"
      - "--build-arg=VCS_URL={{.GitURL}}"

  # production images
  - image_templates:
      - anchore/syft:latest
      - anchore/syft:{{.Tag}}
      - ghcr.io/anchore/syft:latest
      - ghcr.io/anchore/syft:{{.Tag}}
    dockerfile: Dockerfile
    use: buildx
    build_flag_templates:
      - "--platform=linux/amd64,linux/arm64,linux/ppc64le,linux/s390x"
      - "--build-arg=BUILD_DATE={{.Date}}"
      - "--build-arg=BUILD_VERSION={{.Version}}"
      - "--build-arg=VCS_REF={{.FullCommit}}"
      - "--build-arg=VCS_URL={{.GitURL}}"
```

If this is true then we'll additionally need this same qemu setup step in the `release.yaml` workflow.

MikeTheCyberGuy: Yes, you're right. The introduction of addgroup in the distroless context means Syft is now executing platform-specific binaries during the build, which is why QEMU is required. I almost have the new goreleaser file finished I'm just waiting on this test to build. Should be able to push it here shortly.

PR: Goreleaser update #3954

wagoodman: I think keeping these changes together in this PR will help with the review, since the refactor of the goreleaser file is motivated by the changes in this PR. The new PR seems to also be pulling in the dockerfile changes too, which might have been unintentional.

wagoodman: However, looking further into my suggestion, I don't think what I'm saying is possible with goreleaser. When trying this out I'm seeing:
```
ERROR: docker exporter does not currently support exporting manifest lists
```

wagoodman: It looks like another path forward would be to use the nonroot tags for distroless:
```
$ docker run --rm -it gcr.io/distroless/static-debian12:debug-nonroot 
~ $ id
uid=65532(nonroot) gid=65532(nonroot) groups=65532(nonroot)

```

There are both `gcr.io/distroless/static-debian12:debug-nonroot` and `gcr.io/distroless/static-debian12:nonroot` 🎉

Dockerfile
```diff
@@ -1,14 +1,21 @@
 FROM gcr.io/distroless/static-debian12:latest AS build
 
-FROM scratch
-# needed for version check HTTPS request
+ARG TARGETPLATFORM
+FROM --platform=$TARGETPLATFORM alpine:3.18 AS security_provider
+RUN addgroup -S nonroot && adduser -S nonroot -G nonroot
+
+FROM gcr.io/distroless/base-debian12
```
wagoodman: I don't think there is a need for multiple stages anymore, we can use the static variant directly.

wagoodman:
I refactored this to use the `gcr.io/distroless/static-debian12` images directly, so QEMU is not needed (which is much simpler). The difference is that we're using the `nonroot` tags which should deal with the core issue more simply than needing our own security context stage.

wagoodman (approved):

kzantow:
Hi @MikeTheCyberGuy 👋  -- due to a number of users having permission issues that were not obvious to workaround, we've reverted the nonroot-by-default (in `latest` tags), so the latest images are back to using root by default, however we've introduced a "nonroot" tag for users who want to run the `nonroot` variant; that is to say you will probably want to use `anchore/syft:nonroot` (and similar for Grype) for your purposes.

PR #3946, opened 2025-06-01
Feature: DotNet Licenses

### LINKED ISSUE #1227: Get licenses for NuGet packages

naoriel:
# Description

This is an alternative pull request to #3329 which does not require shelling out to the dotnet cli.

- Fixes #1227

## Type of change

# Checklist:


naoriel:
@spiffcs , @kzantow :
May I ask you to perform the review?
The code should mostly look familiar, as I created the #3329 pull request for my employer, when I was still allowed to put some work in. 😉

spiffcs:
@naoriel ty so much for all the effort! I'll try and take a look today

naoriel:
Bugger for the Unit Tests.
I'll have to revisit those. 🤦‍♂️

Unfortunately I have been unable to compile syft under Windows proper.
I'll re-check with my Linux machines a.s.a.p.

Sorry.

naoriel:
@spiffcs 
I've finally found the time to fix the Unit Tests.

I do not have a runner for the Unit Tests on GitHub, but they throw no more errors locally.
As for all the other parts of the Validations workflow: they run through smoothly.

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,557 @@
...
+					out = append(out, c.extractLicensesFromArchive(ctx, entries, zr)...)
+				}
+			}
+		}
+	}
+
+	return out
+}
+
+func (c *nugetLicenseResolver) getResponseForRemotePackage(providerURL, moduleName, moduleVersion string) (*http.Response, error) {
```
martincostello, line 485: Just curious, have you considered getting the license directly from the package's catalog entry, rather than downloading the package and unpacking it?

For example once per NuGet registry:

1. Get the service index (e.g. `https://api.nuget.org/v3/index.json`)
2. Get the URL (`@id`) of the `@type=SearchQueryService` resource from the response (e.g. `https://azuresearch-usnc.nuget.org/query`)

Then for each package:

1. Query that for the package (e.g. `https://azuresearch-usnc.nuget.org/query?prerelease=true&q=PackageId%3APolly&semVerLevel=2.0.0&take=1`)

Then for each version of a package:

1. Find the desired version in the response and get `@id` (e.g. `https://api.nuget.org/v3/registration5-gz-semver2/polly/8.6.0.json`)
2. Get the catalog entry from that URL and get the registration URL `registration` (e.g. `https://api.nuget.org/v3/catalog0/data/2025.06.10.15.45.52/polly.8.6.0.json`).
3. Get the registration from that URL and use the `licenseExpression` and/or `licenseUrl` properties as appropriate, which in this example is `BSD-3-Clause`.

martincostello:
Does .NET need adding to the enrichment options?

```
--enrich stringArray                        enable package data enrichment from local and online sources (options: all, golang, java, javascript)
```

https://github.com/anchore/syft/blob/181e180284ea0ed2458e92f06279cd4184ce2053/cmd/syft/internal/options/catalog.go#L315-L320

syft/pkg/cataloger/dotnet/config.go
```diff
@@ -36,11 +56,71 @@ func (c CatalogerConfig) WithPropagateDLLClaimsToParents(propagate bool) Catalog
...
+
+	return c
+}
+
+func getDefaultLocalNuGetCachePath() string {
+	if runtime.GOOS == "windows" {
+		return path.Clean(path.Join(os.Getenv("USERPROFILE"), ".nuget", "packages"))
+	}
+	return "~/.nuget/packages"
+}
```
martincostello: This function should prefer the [`NUGET_PACKAGES` environment variable](https://learn.microsoft.com/nuget/reference/cli-reference/cli-ref-environment-variables
) if it is set.

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,557 @@
...
+			c.localNuGetCacheResolvers = c.getLocalNugetFolderResolvers(c.assetDefinitions)
+		}
+
+		// if we're running against a directory on the filesystem, it may not include the
+		// user's homedir, so we defer to using the localModCacheResolvers
+		for _, resolver := range c.localNuGetCacheResolvers {
+			if lics, err := c.findLocalLicenses(ctx, resolver, moduleSearchGlob(moduleName, moduleVersion)); err == nil {
+				licenses = appendNewLicenses(licenses, lics...)
+				return licenses, nil
+			}
```
martincostello: If `lics` is empty when no NuGet packages were found on disk, then the function returns instead of searching the remote sources.

I changed it locally do this:

```go
for _, resolver := range c.localNuGetCacheResolvers {
	if lics, err := c.findLocalLicenses(ctx, resolver, moduleSearchGlob(moduleName, moduleVersion)); err == nil {
		if len(lics) != 0 {
			licenses = appendNewLicenses(licenses, lics...)
			return licenses, nil
		}
	}
}
```

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,557 @@
...
+	}
+	return licenses, nil
+}
+
+func (c *nugetLicenseResolver) findLocalLicenses(ctx context.Context, resolver file.Resolver, globMatch string) ([]pkg.License, error) {
+	if resolver == nil {
+		return nil, nil
+	}
+
+	locations, err := resolver.FilesByGlob(globMatch)
```
martincostello: Running this code locally on my laptop, this function is incredibly slow to return (minutes per package). It might be because this is my development machine so I have a relatively large NuGet package cache (shown below), but the runtime performance is terrible.

Maybe the glob needs to be more specific? Debugging locally it looks something like `"**/polly/8.6.1-alpha.0.14/*"`. Looking at how the cache is laid out on my machine, I think the leading `**/` needs to be dropped.

```go
func moduleSearchGlob(moduleName, moduleVersion string) string {
	return fmt.Sprintf("%s/*", moduleDir(moduleName, moduleVersion))
}
```

image

image

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,557 @@
...
+					response.Body.Close()
+				}
+			}
+		}
+	}
+
+	return response, err
+}
+
+func (c *nugetLicenseResolver) getLicensesFromRemotePackage(ctx context.Context, providerURL, moduleName, moduleVersion string) ([]pkg.License, bool) {
```
martincostello, line 517: Running this locally, it's **very** slow as repeated references to the same package cause the search-download-extract process to be repeated many times. I sped this up significantly by caching the lookups:

```go
type licenseCacheEntry struct {
	Licenses []pkg.License
	Found    bool
}

var licenseCache = make(map[string]licenseCacheEntry)

func (c *nugetLicenseResolver) getLicensesFromRemotePackage(ctx context.Context, providerURL, moduleName, moduleVersion string) ([]pkg.License, bool) {

	cacheKey := providerURL + "|" + moduleName + "|" + moduleVersion
	if entry, found := licenseCache[cacheKey]; found {
		return entry.Licenses, entry.Found
	}

	out := []pkg.License{}
	foundPackage := false

	response, err := c.getResponseForRemotePackage(providerURL, moduleName, moduleVersion)
	if err == nil && response.StatusCode == http.StatusOK {
		foundPackage = true
		moduleData, err := io.ReadAll(response.Body)
		response.Body.Close()

		if err == nil {
			out = c.extractLicensesFromBinaryNuGetPackage(ctx, moduleData)
		}
	}

	licenseCache[cacheKey] = licenseCacheEntry{
		Licenses: out,
		Found:    foundPackage,
	}

	return out, foundPackage
}
```

I'm not a Go developer and that was mostly written by Copilot, I'm sure there's better/more-idiomatic ways to do the same thing.

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,557 @@
...
+		}
+	}
+
+	if len(licenses) == 0 {
+		return licenses, errors.New("no licenses found")
+	}
+	return licenses, nil
+}
+
+func (c *nugetLicenseResolver) findLocalLicenses(ctx context.Context, resolver file.Resolver, globMatch string) ([]pkg.License, error) {
```
martincostello: This could maybe similarly be sped up by using a cache. I made a change locally to try it, but it only seemed to save me about 10 seconds.

With the two cache changes it takes about 1m48s to run `.\syft.exe "{clone of https://github.com/App-vNext/Polly}" --enrich "dotnet" --output syft-json`

martincostello:
With the changes mentioned in my review made, I was able to run this change against my local clone of [Polly](https://github.com/App-vNext/Polly), and then extract all the unique SPDX expressions from the generated SPDX JSON file to see what licences all our dependencies use 😎:

```pwsh
❯ Get-Content .\output.json | ConvertFrom-Json | `
∙     Select-Object -ExpandProperty artifacts | `
∙     Select-Object -ExpandProperty licenses | `
∙     Select-Object -ExpandProperty spdxExpression | `
∙     Where-Object { $_ } | Sort-Object -Unique
Apache-2.0
BSD-3-Clause
MIT
```

syft/pkg/cataloger/dotnet/config.go
kzantow: All environment lookups should be done at the CLI configuration level and final resolved values passed in config to the catalogers.

syft/pkg/cataloger/dotnet/licenses.go
kzantow: There is a file-based caching mechanism you should use instead of a global map, look at: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/golang/licenses.go#L174

syft/pkg/cataloger/dotnet/licenses.go
kzantow: There is a file-based caching mechanism you should use instead of a global map, look at: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/golang/licenses.go#L174

syft/pkg/cataloger/dotnet/licenses.go
kzantow: Glob searching against the filesystem is incredibly slow, this is why we use indexes for the source filesystems (which have their own performance issues, but are orders of magnitude faster). Do you need to use globs or can you just do a walk of the directory tree?

kzantow: If you know a specific subdirectory, and can avoid the `**`, it is definitely going to be a lot faster using globs or walking the files directly on a `fs.FS` instance or similar.

syft/pkg/cataloger/dotnet/licenses.go
naoriel: In my previous (failed) attempt I had implemented shelling out to the dotnet CLI tool in order to get the available service indices, which I then queried for the appropriate registry URLs.
Unfortunately the use of the CLI tool blocked the merge, due to understandable dependency concerns.

syft/pkg/cataloger/dotnet/licenses.go
naoriel: Good point.

syft/pkg/cataloger/dotnet/licenses.go
martincostello: I don't follow how that's relevant to my question. You're already connecting to the NuGet API to download the packages remotely - this is just a change to the methodology of what you do with that remote registry connection to avoid the IO overhead of downloading the (potentially large) NuGet package file(s) and unpacking them to get the `.nuspec` to read the SPDX expression when you could just get the same information directly from the registry.

syft/pkg/cataloger/dotnet/config.go
naoriel: Should be fixed by 7d0f87f9e1e93f985dd6c7be6dea4eff2fbed0c3

syft/pkg/cataloger/dotnet/licenses.go
naoriel: Fixed in 17eb377c422f45e89637ccec0e1b25182dabe3bd

syft/pkg/cataloger/dotnet/licenses.go
naoriel: Should be fixed by commits
40f00b47fb99e5356e2ee3687ccee463da938f64
to
19106b870a8e676a0e99b2c1120c10337c319a6a

syft/pkg/cataloger/dotnet/licenses.go
naoriel: Fixed in f4d9073b0178c57e5f763c8d6907f1e78a0d62ff

syft/pkg/cataloger/dotnet/licenses.go
naoriel: Fixed in f4d9073b0178c57e5f763c8d6907f1e78a0d62ff

syft/pkg/cataloger/dotnet/licenses.go
naoriel: Guess I didn't quite catch your meaning... 🤦🏻‍♂️ 

I have to admit, that I did not consider only downloading the ```.nuspec``` file, as I am not **that** fluent in the NuGet registry functionalities.
I also figured, that it would make sense to have the ```.nupkg``` file locally to be able to verify the retrieved data more easily. (And to possibly have them for the next build straight away, eg. if it is just a cloned repository you are analysing.)

I hope this answer is a bit more useful... 🤷🏻‍♂️

martincostello: I just figured that with the caching, this will speed things up (downloading a small XML file) vs. downloading the whole package and all the data it contains that might be megabytes.

martincostello:
Just running the latest code on my laptop now with my manual changes reverted, and it is still pretty slow. With my local changes the other day it took about 2 minutes to scan Polly. ~~It's been running for over 5 minutes so far...~~ Updated: cancelled after 17 minutes. My hunch is the local package glob is still over-broad.

This is what I'm doing if you want to try it yourself:

First clone and build Polly:

```pwsh
git clone https://github.com/App-vNext/Polly.git
cd Polly
.\build.ps1
```

Then build and run Syft from this branch:

```
.\syft.exe "<Polly clone directory>" --enrich "all" --output syft-json
```

WesHaze:
@naoriel Do you plan to continue work on this pr?

naoriel:
@WesHaze 
As a matter of fact, I do plan to continue on this, but time is hard to come by these days...

WesHaze:
@naoriel nice, no one is in a hurry i think. I just wanted to check, thanks for your work.

NathPKP:
Hello @martincostello , @kzantow sorry to bother but is the review planned for soon? Or should I search for another solution?

martincostello:
I'm not a maintainer, but from what I've seen the performance issues from my local testing here haven't been addressed: https://github.com/anchore/syft/pull/3946#issuecomment-2981031596

syft/pkg/cataloger/dotnet/config.go
```diff
@@ -36,11 +53,61 @@ func (c CatalogerConfig) WithPropagateDLLClaimsToParents(propagate bool) Catalog
...
+
+func (c CatalogerConfig) WithCredentials(input []credential.SimpleCredential) CatalogerConfig {
+	if len(input) == 0 {
+		return c
+	}
+
+	c.ProviderCredentials = []credential.SimpleCredential{}
+
+	for _, _credential := range input {
```
kzantow: using underscores in variable names is quite uncommon

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,657 @@
...
+	var licenses []pkg.License
+
+	if c.cfg.SearchLocalLicenses {
+		if c.localNuGetCacheResolvers == nil {
+			// Try to determine NuGet package folder resolvers
+			c.localNuGetCacheResolvers = c.getLocalNugetFolderResolvers(c.assetDefinitions)
+		}
+
+		// if we're running against a directory on the filesystem, it may not include the
+		// user's homedir, so we defer to using the localModCacheResolvers
```
kzantow: I think this comment has leftovers from golang

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,657 @@
...
+		}
+	}
+	return licenses
+}
+
+func (c *nugetLicenseResolver) getLicenses(ctx context.Context, moduleName, moduleVersion string) ([]pkg.License, error) {
+	var licenses []pkg.License
+
+	if c.cfg.SearchLocalLicenses {
+		if c.localNuGetCacheResolvers == nil {
```
kzantow: `len() == 0` in case it was set to non-nil empty slice

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,657 @@
...
+}
+
+func getProjectAssets(resolver file.Resolver) ([]projectAssets, error) {
+	assets := []projectAssets{}
+	var err error
+
+	// Try to determine NuGet package assets from temporary object files
+	// (usually located in the /obj folder)
+	var assetFiles []file.Location
+	if assetFiles, err = resolver.FilesByGlob("**/*.json"); err == nil && len(assetFiles) > 0 {
```
kzantow: depending on context, this will return MANY files, and might be related to the performance issues mentioned, it looks like this is only looking for `project.assets.json`, why not just glob search for this directly? Refer to the other comment about host filesystem vs source

cmd/syft/internal/options/catalog.go
```diff
@@ -168,13 +168,19 @@ func (cfg Catalog) ToPackagesConfig() pkgcataloging.Config {
 		IncludeIndexedArchives:   cfg.Package.SearchIndexedArchives,
 		IncludeUnindexedArchives: cfg.Package.SearchUnindexedArchives,
 	}
+	cfg.Dotnet.AddDefaultLocalNuGetCachePathIfEmpty()
```
kzantow: This is an atypical configuration function, if we want to have default values, either return it from `dotnet.DefaultCatalogerConfig()`, or detect something unset in the cataloger itself and use default values there.

syft/credential/simple_credential.go
```diff
@@ -0,0 +1,10 @@
+package credential
+
+type SimpleCredential struct {
```
kzantow, line 3: This seems like a useful shared datatype, but it seems this is only used by the Dotnet configuration so probably should be there until we have more examples to understand some sort of common pattern to implement.

syft/pkg/cataloger/dotnet/licenses_test.go
```diff
@@ -0,0 +1,154 @@
...
+			SPDXExpression: "MIT",
+			Type:           "declared",
+			URLs:           []string{"https://licenses.nuget.org/MIT"},
+			Locations:      file.LocationSet{},
+		},
+	}
+
+	licenseParser := newNugetLicenseResolver(CatalogerConfig{})
+
+	t.Run(fixture, func(t *testing.T) {
```
kzantow: there isn't a reason to `t.Run` here unless we're using table tests in a for loop or similar

syft/pkg/cataloger/dotnet/licenses_test.go
```diff
@@ -0,0 +1,154 @@
...
+		result, _ := licenseParser.getLicensesFromRemotePackage(context.Background(), defaultNuGetProvider, "newtonsoft.json", "13.0.1")
+		assert.Equal(t, expected, result)
```
kzantow: `assert.` -> `require.` (same comment everywhere assert is used, we should probably use require)

syft/pkg/cataloger/dotnet/licenses.go
```diff
@@ -0,0 +1,657 @@
...
+	nugetPackagePaths := []string{}
+	if len(c.cfg.LocalCachePaths) > 0 {
+		nugetPackagePaths = append(nugetPackagePaths, c.cfg.LocalCachePaths...)
+	} else {
+		nugetPackagePaths = append(nugetPackagePaths, getNuGetCachesFromProjectAssets(assetDefinitions)...)
+	}
+
+	resolvers := []file.Resolver{}
+	for _, nugetPackagePath := range nugetPackagePaths {
+		resolvers = append(resolvers, fileresolver.NewFromUnindexedDirectory(nugetPackagePath))
```
kzantow, line 677: Generally, file resolvers should be used to represent the source we are scanning: image, filesystem, etc., but if we're reading files from the host machine, we should just use a normal `fs.FS`. If these paths are already indexed in the scanned source, we don't need new resolvers, just more targeted glob searches

syft/pkg/cataloger/dotnet/config.go
```diff
@@ -40,11 +57,61 @@ func (c CatalogerConfig) WithPropagateDLLClaimsToParents(propagate bool) Catalog
...
 func DefaultCatalogerConfig() CatalogerConfig {
 	return CatalogerConfig{
 		DepPackagesMustHaveDLL:             false,
 		DepPackagesMustClaimDLL:            true,
 		PropagateDLLClaimsToParents:        true,
 		RelaxDLLClaimsWhenBundlingDetected: true,
+		SearchLocalLicenses:                true,
+		LocalCachePaths:                    []string{},
+		SearchRemoteLicenses:               false,
+		Providers:                          []string{defaultNuGetProvider},
```
kzantow: Should this be called "BaseURLs"? "Provider" seems a bit of a misleading name. And, if there are credentials should these be set together?

naoriel:
> Just running the latest code on my laptop now with my manual changes reverted, and it is still pretty slow. With my local changes the other day it took about 2 minutes to scan Polly. ~It's been running for over 5 minutes so far...~ Updated: cancelled after 17 minutes. My hunch is the local package glob is still over-broad.
> 
> This is what I'm doing if you want to try it yourself:
> 
> First clone and build Polly:
> 
> ```powershell
> git clone https://github.com/App-vNext/Polly.git
> cd Polly
> .\build.ps1
> ```
> 
> Then build and run Syft from this branch:
> 
> ```
> .\syft.exe "<Polly clone directory>" --enrich "all" --output syft-json
> ```

I've tried to follow your short build example, but have yet been unable to build it successfully.
Which framework version should I be using? net6.0? net8.0? another?

I'll see whether I can perform a proper timing analysis for a few other projects for the moment being in order to determine the actual timing percentage additionally required for the determination of the licenses.
However free time is somewhat sparse on my side.

martincostello:
Polly uses the .NET 9 SDK: [global.json](https://github.com/App-vNext/Polly/blob/50b13895856d53bf84e82cc7643982ed21ab308c/global.json#L3)

kzantow (commented):
I left some targeted comments, but generally I think the term "provider" should probably be replaced with something more appropriate throughout, and we should probably not use a file resolver to search for license files on the host, but rather a `fs.FS` and doublestar directly, or just iterating the filesystem, unless it's accessing files in the scanned source, in which case there already is a file resolver; this may be related to the performance issues noted in the comments.

syft/pkg/cataloger/dotnet/config.go
naoriel: Good point.

syft/pkg/cataloger/dotnet/licenses.go
naoriel: 🤦🏻‍♂️

syft/pkg/cataloger/dotnet/licenses.go
naoriel: You're right of course.

syft/pkg/cataloger/dotnet/licenses.go
naoriel: I'll see whether that doesn't trigger any tests to fail...
I think to recall, that it did...

cmd/syft/internal/options/catalog.go
naoriel: It's probably a good idea to shift the related code right into the license resolver implementation, where it's being used anyway.
👍🏻

syft/credential/simple_credential.go
naoriel: Fair enough.

syft/pkg/cataloger/dotnet/licenses_test.go
naoriel: Guess I'll just add more test cases then. 😉

syft/pkg/cataloger/dotnet/licenses_test.go
naoriel: I'm getting onto it.

syft/pkg/cataloger/dotnet/config.go
naoriel: I'd say "NuGetRepositoryURLs" would be a more fitting name in that case.

syft/credential/simple_credential.go
naoriel: Unfortunately the PackageCatalogerConventions integration test forbade me to shove it into the dotnet package...

syft/pkg/cataloger/dotnet/licenses.go
naoriel: The file resolvers created here represent the local NuGet cache directories, so they will most likely not yet have been scanned and are almost certainly not part of the source.

syft/credential/simple_credential.go
naoriel: > Unfortunately the PackageCatalogerConventions integration test forbade me to shove it into the dotnet package...

It just doesn't like non-cataloger structs in the cataloger package...

naoriel: @kzantow 
Any suggestions?

syft/pkg/cataloger/dotnet/licenses.go
naoriel: @kzantow 
What do you suggest?

floric:
Hey :) any way we could support this MR?

naoriel:
> Hey :) any way we could support this MR?

Sure, if you have an idea on how to improve the timing that @martincostello complained about... 🤷‍♂️
My first guess was a myriad of dependencies, but I haven't gotten around to actually check that due to VERY limited time available. (At least I got around to merging the main branch again. 🙈)
Maybe a parameter to limit the hierarchical depth of the dependencies to consider might help?

Issue #3957, opened 2025-06-04
opentelemetry CPE vendor & product are wrong


jayvdb:
**What happened**:

Here are examples of a NVD entry with a CPE which would not match what syft produces

https://nvd.nist.gov/vuln/detail/CVE-2023-39951  uses `cpe:2.3:a:linuxfoundation:opentelemetry_instrumentation_for_java:*:*:*:*:*:*:*:*`  

This is in https://github.com/open-telemetry/opentelemetry-java-instrumentation

https://nvd.nist.gov/vuln/detail/CVE-2023-25151  uses `cpe:2.3:a:linuxfoundation:opentelemetry-go_contrib:0.38.0:*:*:*:*:opentelemetry-go:*:*`

This is in https://github.com/open-telemetry/opentelemetry-go-contrib

However, https://nvd.nist.gov/vuln/detail/CVE-2023-43810 uses `cpe:2.3:a:opentelemetry:opentelemetry:*:*:*:*:*:*:*:*`

This is https://github.com/open-telemetry/opentelemetry-python-contrib

I assume the vendor should be `linuxfoundation` instead of `opentelemetry` for all of these, and syft should be emitting that.

Note I have not seen SBOMs from syft with those CPEs, but in my Rust project I see CPE `cpe:2.3:a:opentelemetry:opentelemetry:0.23.0:*:*:*:*:*:*:*`

which happens to match https://nvd.nist.gov/vuln/detail/CVE-2023-43810 , and that led me down this garden path.

**What you expected to happen**:

If I understand correctly, entries are needed in https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/internal/cpegenerate/dictionary/data/cpe-index.json to fix the vendor & product for the various opentelemetry implementations in each language.

I expect that syft provides CPEs that match those used by NVD for CVE-2023-25151 & CVE-2023-39951 and similar.

I expect that syft would generate a CPE that matches CVE-2023-43810 , but also another CPE which uses the appropriate vendor `linuxfoundation`

I expect for Rust opentelemetry crates, the vendor is `linuxfoundation` and the product is the crate name.

**Steps to reproduce the issue**:

**Anything else we need to know?**:

**Environment**:
- Output of `syft version`:
- OS (e.g: `cat /etc/os-release` or similar):

jayvdb:
[...]

Issue #3956, opened 2025-06-04
CPE target_sw not being set consistency for Rust crates


jayvdb:
**What happened**:

When I use syft on binaries built with cargo-auditable, the CPEs always have a `language` = `*`.

When I use syft on the Cargo.lock or binaries built with cargo-auditable, the `language` is set to `rust` for these 25 CPEs in my project
```xml
            <cpe>cpe:2.3:a:actix:actix-codec:0.5.2:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:actix:actix-http:3.10.0:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:actix:actix-service:2.0.3:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:actix:actix-utils:3.0.1:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:actix:actix-web:4.10.2:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:bzip2_project:bzip2:0.5.2:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:crossbeam-channel_project:crossbeam-channel:0.5.15:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:rust-lang:futures-task:0.3.31:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:generic-array_project:generic-array:0.14.7:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:getrandom_project:getrandom:0.2.15:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:getrandom_project:getrandom:0.3.2:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:heapless_project:heapless:0.7.17:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:lock_api_project:lock_api:0.4.12:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:dimforge:nalgebra:0.32.6:*:*:*:*:rust:*:*</cpe>
            <cpe>cpe:2.3:a:nix_project:nix:0.23.2:*:*:*:*:rust:*:*</cpe>
[...]

PR #3962, opened 2025-06-04
fix: Remove two Rust crate false positive CPE matches

jayvdb:
# Description

Rust crates opentelemetry and redis are being given CPEs that match CVEs such as CVE-2023-45142 and CVE-2022-24735 respectively. The vendor overrides added here prevent that.

Relates to https://github.com/anchore/syft/issues/3956 & https://github.com/anchore/syft/issues/3957 , both of which are broader, while this fixes the immediate false positives. 

- Fixes

## Type of change

# Checklist:


kzantow:
Hey @jayvdb -- this file is generated automatically from a small program here: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/internal/cpegenerate/dictionary/index-generator/main.go so any changes will get overridden when that runs again. That said: this is supposed to be all CPEs that are known from NVD so adding other CPEs are just guesses (probably correct guesses), but we can't be guaranteed NVD will use the same thing. I think it might be more useful to look in this package, where we have a lot of handling specific to ecosystems including overrides for CPEs we know are wrong and need to be corrected, such as: https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/internal/cpegenerate/candidate_by_package_type.go#L110

jayvdb:
@kzantow thanks for the guidance.

I have removed the manual edits to `cpe-index.json` , and now using fixes in `candidate_by_package_type.go`.  

This removes the CPEs from these entries in my SBOM, which is acceptable to me, as I already know they have no entries in NVD, but it feels like it isnt a great solution, as the generated SBOM is permanent, but the NVD data grows and could include CVEs for these crates in the future.

If I understood correctly, you do not want me to be using `candidateAddition(..)` because I could only be adding guesses at the NVD vendor or product name, and it sounds like you dont want syft to be guessing those in advance of NVD adding them to their data.

spiffcs (approved):

Issue #3985, opened 2025-06-09
Provider Cleanup Image Resource - Memory Leak Fix


houdini91:
**What would you like to be added**:
Introduce an automatic “cleanup” step that runs after every Syft/Stereoscope scan, deleting any image layers or temp data the providers created—so disk space never accumulates between runs.

**Why is this needed**:
Running syft scan or syft attest against images pulled from Docker or Containerd leaves those images—and their layer files—inside the local daemon. On CI runners this causes disk usage in /var/lib/docker/overlay2 (or the Containerd content store) to grow continuously, eventually filling the filesystem. Previous discussions (e.g. [stereoscope #161](https://github.com/anchore/stereoscope/pull/161), [syft #416](https://github.com/anchore/syft/issues/416)) covered temporary directories, but did not address the persistent layers created by daemon-based providers.

**Additional context**:
Suggested Patches:

Patches are intended as a proof of concept: feel free to ignore them, cherry-pick pieces, or push changes directly onto the branches if there’s a cleaner way to fold this into the projects’ architecture. I’m mainly after feedback on the overall approach and any edge-cases I might have missed (e.g., multi-arch images, custom content-stores, Windows daemon paths).

**What i did**

<html><head></head><body><h3>What I did – high-level per patch</h3>
<hr>
<h4><strong>Syft patch</strong></h4>

Area | High-level change
-- | --
Lifecycle support | Introduced a new Cleanup(ctx) method to the source.Provider interface.
Central helper | Added syft.CleanupSource() to iterate over all candidate providers and call their Cleanup.
CLI wiring | Updated getSource() to return (src, cleanupFunc, err); syft scan and syft attest now defer both src.Close() and cleanupFunc().
Config / options | Propagated registry, platform, and hashing options unchanged; only the return signature changed.
Fixtures | Removed stray merge-conflict lines from a Python RECORD test file.

<p>Result: Stereoscope can actively remove any layers it imported into Docker or Containerd, keeping the host’s LayerFS from ballooning during repeated scans.</p>
<hr>
<p>Together, these two patches establish a complete <strong>create → use → cleanup</strong> lifecycle across both projects.</p></body></html>

houdini91:
Hope this makes sense ..

houdini91:
I am concerned about the cleanup operation potentially removing images that were already present and not explicitly pulled by Syft. If you want to continue with this suggestion we may consider an approach to retain this context, ensuring cleanup targets only images pulled during Syft operations.

[...]

PR #3986, opened 2025-06-09
Suggest Image cleanup flow

houdini91:
See https://github.com/anchore/syft/issues/3985


kzantow:
I don't think this is needed -- `image.Cleanup()` should already be called: https://github.com/anchore/syft/blob/4eb8ba457570f7ed6659d03577564ee2ad0aae6e/syft/source/stereoscopesource/image_source.go#L128

PR #3987, opened 2025-06-09
Undeprecate binary classifier type

### LINKED ISSUE #3972: Expose binutils externally

Hi, PR https://github.com/anchore/syft/pull/2585 moved some of the binutils to an internal package and deprecated the existing one. We are relying on some of this to maintain our own subset of binary detection and we will lose this ability in the future. Would it be possible to have that package be available externally?

**Why is this needed**:

To maintain binary detection extensions

LaurentGoderre:
# Description

Expose binutils types used for extending the binary classifier

- Fixes #3972

## Type of change

# Checklist:


syft/pkg/cataloger/binary/deprecated.go
```diff
@@ -4,12 +4,6 @@ import "github.com/anchore/syft/syft/pkg/cataloger/internal/binutils"
 
 // Note: all generic utilities for catalogers have been moved to the internal/binutils package.
 
-// Deprecated: This package is deprecated and will be removed in syft v2
-type Classifier = binutils.Classifier
```
kzantow, line 8: Hey @LaurentGoderre 👋 I think I may have misunderstood the issue in my response... I didn't notice that these types are already exported with aliases as I was suggesting. 🤦 I see the `FileContentsVersionMatcher` also is exported, similarly. Which were the specific things that were not properly re-exported?

LaurentGoderre: Since these are deprecated, they are going to be removed in the future, no? Is the entire binary package going to be deprecated?

kzantow: I guess what you're really looking for is to move the `internal/binutils` out of an `internal` package, right?

LaurentGoderre: Yeah, ideally

kzantow: If there was another way to specify a classifier, say, by providing yaml configuration with the same matching type functionality that we're using today, would that end up being easier for your needs? do you have a need to write custom evidence matchers or essentially just add more of what exists today?

I ask because I've had an idea for quite some time to make this data driven (probably with a yaml configuration) and I wonder if pursuing this might end up being a better approach than exporting these fairly bespoke functions. I made [a prototype](https://github.com/anchore/syft/compare/main...kzantow-anchore:syft:feat/regex-cataloger?expand=1) of this idea, and the prototype had a couple examples like (one of the trickier ones you worked on that I think we didn't accept due to it not being "binary"):
```
Class:    'redmine'
FileGlob: '**/redmine/version.rb'
Extractor:
  # taken from: https://github.com/anchore/syft/pull/2445
  Regex: '(?m)module Redmine\s*# @private\s*module VERSION\s*MAJOR\s*=\s*(?P<major>[0-9]+)\s*MINOR\s*=\s*(?P<minor>[0-9]+)\s*TINY\s*=\s*(?P<patch>[0-9]+)'
Package:
  Name: 'redmine'
  Version: '{{.major}}.{{.minor}}.{{.patch}}'
  PURL: 'pkg:generic/redmine@{{.version}}'
  CPEs:
  - 'cpe:2.3:a:redmine:redmine:{{.version}}:*:*:*:*:*:*:*'
```

... maybe exporting the binutils is the way to go, or a combination of both, just curious to understand what would end up making your situation better. I'm not convinced this prototype is the best idea, either. But being data driven would enable some future state to be able to provide updates to users without needing to update code, of course. (Also there's an easy way to `//go:embed` files into the binary)

What do you think?

LaurentGoderre: I think that would work for out needs.

PR #3998, opened 2025-06-11
fix: provide separate nonroot image

kzantow:
# Description

This PR reverts a change to use `nonroot` images by default, and provides an alternate `nonroot` set of tagged images. Using `nonroot` causes a number of issues causing users friction such as accessing and modifying the `docker.sock` and mount volumes.

## Type of change

# Checklist:

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -126,6 +126,59 @@ dockers:
       - "--build-arg=VCS_REF={{.FullCommit}}"
       - "--build-arg=VCS_URL={{.GitURL}}"
 
+  # nonroot images...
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -180,49 +233,70 @@ dockers:
       - "--build-arg=VCS_URL={{.GitURL}}"
 
 docker_manifests:
-  # anchore/syft manifests...
```
kzantow, line 183: I reorganized this by tag first similar to the organization of the top section.

wagoodman (approved):

PR #4029, opened 2025-06-26
chore: migrate syft to use mholt/archives instead of anchore fork

Rupikz:
# Description

Rewritten deprecated fork [github.com/anchore/archiver](https://github.com/anchore/archiver) to [github.com/mholt/archives](https://github.com/mholt/archives)

## Type of change

# Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -309,14 +309,14 @@ require (
 	github.com/clipperhouse/stringish v0.1.1 // indirect
 	github.com/clipperhouse/uax29/v2 v2.2.0 // indirect
 	github.com/hashicorp/aws-sdk-go-base/v2 v2.0.0-beta.65 // indirect
-	github.com/nwaples/rardecode/v2 v2.2.0 // indirect
-	github.com/olekukonko/errors v1.1.0 // indirect
-	github.com/olekukonko/ll v0.1.2 // indirect
 	github.com/henvic/httpretty v0.1.4 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
+	github.com/nwaples/rardecode/v2 v2.2.0 // indirect
 	github.com/olekukonko/cat v0.0.0-20250911104152-50322a0618f6 // indirect
+	github.com/olekukonko/errors v1.1.0 // indirect
+	github.com/olekukonko/ll v0.1.2 // indirect
 	github.com/smallnest/ringbuffer v0.0.0-20241116012123-461381446e3d // indirect
 	gonum.org/v1/gonum v0.15.1 // indirect
 )


syft/source/filesource/file_source.go
```diff
@@ -256,15 +250,57 @@ func digestOfFileContents(path string) string {
...
 	tempDir, err := os.MkdirTemp("", "syft-archive-contents-")
 	if err != nil {
-		return "", func() error { return nil }, fmt.Errorf("unable to create tempdir for archive processing: %w", err)
+		return "", cleanupFn, fmt.Errorf("unable to create tempdir for archive processing: %w", err)
 	}
 
-	cleanupFn := func() error {
-		return os.RemoveAll(tempDir)
+	visitor := func(_ context.Context, file archives.FileInfo) error {
+		destPath, err := intFile.SafeJoin(tempDir, file.NameInArchive)
```
spiffcs, line 256: I updated this to use the same kind of SafeJoin functionality so we don't escape the tmpDir like we do in other parts of syft. If the archives library already handles this internally apologies I just couldn't find it on a quick inspection.

spiffcs:
@wagoodman I've looked at this one and added a small protection.

The visitor now uses a path aware directory join and cannot write outside the temp directory. 
Tests also added.
Can you give a second pair of 👀 on this since I added a commit and would like a second review before we 🟢.

Also, I don't know what Static Analysis is on about here. I've checked out this branch even on a different machine and it's told me the go.mod and go.sum are tidy locally ☹️. I ended up doing a manual edit to fix the go.sum, but that seems super wrong.

syft/source/filesource/file_source.go
```diff
@@ -246,15 +239,69 @@ func digestOfFileContents(path string) string {
...
 
-	return tempDir, cleanupFn, unarchiver.Unarchive(path, tempDir)
+	return tempDir, func() error {
+		return os.RemoveAll(tempDir)
+	}, unarchiver.Extract(context.Background(), archive, visitor)
+}
+
+// safeJoinPath ensures that any destinations do not resolve to a path above the prefix path.
+// This protects against directory traversal attacks (zip slip).
+func safeJoinPath(prefix string, dest ...string) (string, error) {
```
wagoodman: this is already implemented in `zip_file_traversal.go` -- can we reuse this instead of reimplementing?

spiffcs: reused the internal `SafeJoin` in `filesource` now

internal/file/tar_file_traversal.go
```diff
@@ -1,27 +1,52 @@
...
 
+// TraverseFilesInTar enumerates all paths stored within a tar archive using the visitor pattern.
+func TraverseFilesInTar(ctx context.Context, archivePath string, visitor archives.FileHandler) error {
+	tarReader, err := os.Open(archivePath)
+	if err != nil {
+		return fmt.Errorf("unable to open tar archive (%s): %w", archivePath, err)
+	}
+	defer func() {
+		if err := tarReader.Close(); err != nil {
+			log.Errorf("unable to close tar archive (%s): %+v", archivePath, err)
```
kzantow: This should not be an Error log, probably Debug. also there's an `internal.CloseAndLogError`

internal/file/zip_file_traversal_test.go
```diff
@@ -307,9 +311,255 @@ func TestSafeJoin(t *testing.T) {
...
+	require.NoError(t, err)
+	defer archive.Close()
+
+	extractDir := filepath.Join(tempDir, "extract")
+	require.NoError(t, os.MkdirAll(extractDir, 0755))
+
+	extractedFiles := make(map[string]bool)
+
+	// Extract with SafeJoin protection
+	visitor := func(_ context.Context, file archives.FileInfo) error {
```
kzantow: Should this call `UnzipToDir` instead of implementing another visitor?

internal/file/zip_file_traversal_test.go
```diff
@@ -307,9 +311,255 @@ func TestSafeJoin(t *testing.T) {
...
+	// Open the archive for extraction
+	archive, err := os.Open(maliciousArchive)
+	require.NoError(t, err)
+	defer archive.Close()
+
+	extractDir := filepath.Join(tempDir, "extract")
+	require.NoError(t, os.MkdirAll(extractDir, 0755))
+
+	// Attempt to extract with SafeJoin protection
+	visitor := func(_ context.Context, file archives.FileInfo) error {
```
kzantow: Why does this test have a custom visitor, would it be better to call `UnzipToDir`?

internal/file/zip_file_traversal.go
```diff
@@ -92,83 +89,83 @@ func ExtractFromZipToUniqueTempFile(archivePath, dir string, paths ...string) (m
...
 		if err != nil {
-			return fmt.Errorf("unable to read file=%q from zip=%q: %w", file.Name, archivePath, err)
+			return fmt.Errorf("unable to read file=%q from zip=%q: %w", file.NameInArchive, archivePath, err)
 		}
 		defer func() {
-			err := zippedFile.Close()
-			if err != nil {
-				log.Errorf("unable to close source file=%q from zip=%q: %+v", file.Name, archivePath, err)
+			if err := zippedFile.Close(); err != nil {
+				log.Errorf("unable to close source file=%q from zip=%q: %+v", file.NameInArchive, archivePath, err)
```
kzantow, line 96: This also should not be a log.Error, the same is true for all of the log messages, I think. in the case of an error reading, we'll be returning an error rather than logging it, and don't need to error log close failures

internal/file/zip_file_traversal_test.go
```diff
@@ -307,9 +311,255 @@ func TestSafeJoin(t *testing.T) {
...
 			test.errAssertion(t, err)
 			assert.Equal(t, test.expected, actual)
 		})
 	}
 }
+
+// TestSymlinkProtection demonstrates that SafeJoin protects against symlink-based
+// directory traversal attacks by validating that archive entry paths cannot escape
+// the extraction directory.
+func TestSymlinkProtection(t *testing.T) {
```
wagoodman:
```suggestion
func TestSafeJoin_SymlinkProtection(t *testing.T) {
```

PR #4041, opened 2025-06-30
feat: add go source cataloger for SBOM generation from go source directories

spiffcs:
### Summary

This PR introduces a new *golang* cataloger to Syft. This new golang cataloger allows for the detection and extraction of go packages and their dependencies/licenses when the syft target is a directory that contains source code for a go program (syft dir:./my-go-project). 

This cataloger is different from the current binary and module catalogers. The source cataloger leverages the https://pkg.go.dev/golang.org/x/tools/go/packages library to load/traverse the package graph for the top most module of the directory source.

This method allows for a few package metadata enhancements that the mod or binary cataloger currently don't have access to. For example, since the https://pkg.go.dev/golang.org/x/tools/go/packages library points to the module cache for dependencies it's possible to look up `License` files for each dependency and run the google license classifier against that file for potential matches. This means that users with a populated local go mod have the ability to match licenses without using any of the current online options.

This cataloger is also useful for source directories with multiple binaries. 

The default import-path search for a given scan is `./...`. This search is overwritten if the user provides paths as seen below. The cataloger can use the new config option `config.golang.go-source.import-paths` to provide search patterns for separate entry points.

#### Example
```
golang:
  go-source:
    import-paths:
      - "./cmd/bin1/..."
      - "./cmd/bin2/..."
```

In the above example, given a `syft dir:.` scan for the source directory, the cataloger will search for modules at the two above paths and build the root modules for both. It will then find all dependencies/transitive dependencies for the two entrypoints and build the correct relationships to complete the graph.

The relationships are modeled by `Relationship{From: dep, To: parent, Type: DependencyOfRelationship}` 

### ✅ Fixes

- Adds support for module-based Go source cataloging [#3451](https://github.com/anchore/syft/issues/3451)

---

## Type of change

- [x] New feature (adds functionality without breaking existing behavior)

---

## Checklist

---


spiffcs:
closed in favor of the merged approach with mod/source found here: https://github.com/anchore/syft/pull/4127

PR #4042, opened 2025-07-01
chore: update tests to read from latest test-fixture-cache and fix cache publish

spiffcs:
# Description

Our test-fixture-cache publish action has been failing recently because a few of our fixtures have drifted forward for non pinned packages. This PR get's all tests to pass on fresh fixtures with the updates expectations.

Rebuilt test-fixture-cache can be found here:
https://github.com/anchore/syft/actions/runs/15990642133

### Changes
- Taskfile whitespace stripped via my nvim plugin
- `cmd/syft/internal/test/integration/package_deduplication_test.go` expectations changed for a more recent httpd
- `syft/pkg/cataloger/php/interpreter_cataloger_test.go` changed for a more recent php interpreter
- `test/cli/dir_root_scan_regression_test.go` extralines and whitespace stripped by nvim plugin
- `test/cli/utils_test.go` updated to account for arm64 architecture development

# Checklist:

### DIFF
--- a/test/cli/utils_test.go
+++ b/test/cli/utils_test.go
@@ -83,19 +83,15 @@ func pullDockerImage(t testing.TB, image string) {
 func runCycloneDXInDocker(t testing.TB, env map[string]string, image string, f *os.File, args ...string) (*exec.Cmd, string, string) {
 	t.Helper()
 
-	// Base Docker args
-	allArgs := []string{
-		"run",
-		"-t",
-		"-v", fmt.Sprintf("%s:/sbom", f.Name()),
+	allArgs := []string{"run", "-t"}
+
+	if runtime.GOARCH == "arm64" {
+		t.Logf("Detected %s/%s — adding --platform=linux/amd64 for emulation", runtime.GOOS, runtime.GOARCH)
+		allArgs = append(allArgs, "--platform=linux/amd64")
 	}
 
-	if runtime.GOOS == "linux" && runtime.GOARCH == "arm64" {
-		t.Log("Detected linux/arm64 — adding --platform=linux/amd64 for qemu emulation")
-		allArgs = append([]string{"--platform=linux/amd64"}, allArgs...)
-	}
+	allArgs = append(allArgs, "-v", fmt.Sprintf("%s:/sbom", f.Name()))
 
-	// Append the image and CycloneDX args
 	allArgs = append(allArgs, image)
 	allArgs = append(allArgs, args...)
 


syft/pkg/cataloger/php/interpreter_cataloger_test.go
```diff
@@ -18,54 +18,54 @@ func Test_InterpreterCataloger(t *testing.T) {
 			fixture: "image-extensions",
 			expectedPkgs: []string{
 				// interpreters
-				"php-cli @ 8.3.21 (/usr/local/bin/php)",
-				"php-fpm @ 8.3.21 (/usr/local/sbin/php-fpm)",
+				"php-cli @ 8.3.22 (/usr/local/bin/php)",
```
kzantow, line 21: It looks like there are a [couple](https://github.com/anchore/syft/blob/8941b4fac009022425340b60e687e71567f8e8c7/syft/pkg/cataloger/php/test-fixtures/image-extensions/Dockerfile#L2C33-L2C50) [spots](https://github.com/anchore/syft/blob/8941b4fac009022425340b60e687e71567f8e8c7/syft/pkg/cataloger/php/test-fixtures/image-extensions/Dockerfile#L92) we could pin images to hashes, but we're [getting these packages](https://github.com/anchore/syft/blob/8941b4fac009022425340b60e687e71567f8e8c7/syft/pkg/cataloger/php/test-fixtures/image-extensions/Dockerfile#L7) via [package manager installs](https://github.com/anchore/syft/blob/8941b4fac009022425340b60e687e71567f8e8c7/cmd/syft/internal/test/integration/test-fixtures/image-vertical-package-dups/Dockerfile#L10), which is drifting separately. Should we include versions with the package manager installs, too? Maybe it's `apt install <package>=<version>` and `dnf install <package>-<version>` for the package_deduplication_test above?

test/cli/utils_test.go
```diff
@@ -81,18 +81,27 @@ func pullDockerImage(t testing.TB, image string) {
...
+	// Base Docker args
+	allArgs := []string{
+		"run",
+		"-t",
+		"-v", fmt.Sprintf("%s:/sbom", f.Name()),
+	}
+
+	if runtime.GOOS == "linux" && runtime.GOARCH == "arm64" {
+		t.Log("Detected linux/arm64 — adding --platform=linux/amd64 for qemu emulation")
+		allArgs = append([]string{"--platform=linux/amd64"}, allArgs...)
```
kzantow: We probably want this on `darwin/arm64`, too -- should this `linux/amd64` platform just always be added if there's no `--platform` flag in the provided `args`?

kzantow (approved):

syft/pkg/cataloger/php/interpreter_cataloger_test.go
spiffcs: I can take a look at these separately from this PR to see about pinning for the future. The outer loop on getting these updated along with the trial and error associated on tests failing on the cache rebuild is a little long at the moment so this might be a good weekend activity.

test/cli/utils_test.go
spiffcs: Yep! Updated this one to just look for `.GOARCH == arm64` and then added the flag.

This was from the local error:
```
latest: Pulling from cyclonedx/cyclonedx-cli
docker: no matching manifest for linux/arm64/v8 in the manifest list entries
```

I matched with the manifest, but I forgot docker manifest paths don't follow the GOOS constants.

Either way we won't find a published image for cdx using arm64 so we will use platform in this case for the emulation.

PR #4046, opened 2025-07-02
fix: align binary java detection with jvm cataloger + support IBM

### LINKED ISSUE #3893: Azul JDK classified as Oracle JRE
**What happened**:
When scanning an image that contained an [Azul JDK](https://www.azul.com/downloads/?package=jdk#zulu), it is misidentified as `oracle:jdk`.

This java binary is classified as `oracle:openjdk`, which Azul is based on and appears to have compatible numbering with.

```
syft --override-default-catalogers binary azul/zulu-openjdk:8
```

kzantow:
# Description

This PR makes some changes to binary detection of Java packages:
* a sequential/branching evidence matcher is introduced, where the first result stops further tests
* java binary is read once and kept in memory instead of repeatedly read by multiple evidence matchers
* the package names, PURLs, and CPEs have been updated to match what the JVM cataloger returns

- Fixes #3893 

## Type of change

# Checklist:


syft/pkg/cataloger/binary/classifiers.go
```diff
@@ -113,73 +113,10 @@ func DefaultClassifiers() []binutils.Classifier {
 				cpe.Must("cpe:2.3:a:redis:redis:*:*:*:*:*:*:*:*", cpe.NVDDictionaryLookupSource),
 			},
 		},
-		{
```
kzantow, line 116: All the java classifiers have been moved to a separate file 👇

syft/pkg/cataloger/internal/binutils/branching_matcher.go
```diff
@@ -0,0 +1,49 @@
...
+// try classifier EvidenceMatcher until the first set of packages is found
+func BranchingEvidenceMatcher(classifiers ...Classifier) EvidenceMatcher {
+	return func(_ Classifier, context MatcherContext) ([]pkg.Package, error) {
+		// we are scanning the same contents multiple times, so read it into memory for a reader that can reset
+		rdr, err := getReader(context)
+		if err != nil {
+			return nil, err
+		}
+		defer internal.CloseAndLogError(rdr, context.Location.RealPath)
+		contents, err := io.ReadAll(rdr)
```
wagoodman: I see that we're trying to allow for multiple readers for each classifier, but this could be handled by making a single unionreader, which automatically detects if the underlying reader can be reused to seek/reset, or if the contents need to be fully read if the reader does not support seeking (the stereoscope readers do support seeking so should be pretty efficient here). If these were small-ish files then this would be less of an issue, but the readers could be representing a binary that is tens or hundreds of MB in size.

kzantow: Turns out I was already getting a unionreader 🤦

wagoodman (approved):

PR #4081, opened 2025-07-24
Vcpkg Cataloger

### LINKED ISSUE #2110: Support for `vcpkg`
`syft` should be able to use `vcpkg` to generate SBOMs for C++ projects

**Why is this needed**:
At @reMarkable, we have a giant C++ codebase. We'd love to use `syft` to generate SBOMs for this project, but we're reliant on using `vcpkg` as a package manager, not `conan`

**Additional context**:

gabetrau:
# Support C++ package manager Vcpkg

- Fixes [issue link](https://github.com/anchore/syft/issues/2110)

## Questions I have
- vcpkg has "[Port versions](https://learn.microsoft.com/en-us/vcpkg/users/versioning.concepts)" these get added to the end of the version if they are 1 or above. They are separated by a hashtag ex. 1.0.2#1. I add the combined version to the PURL. Is it okay to have  '#' in a PURL? Also this [issue](https://github.com/package-url/purl-spec/pull/245) is relevant to PURL 
- If there is no VCPKG_ROOT env variable set, I'm currently looking at the vcpkg cache to see if the info needed is in the git directory stored there. To do this I use the cache manager and then relative to that, I retrieve the folder for the vcpkg cache. Is this allowed?
- Some of the fields in VcpkgManifest are already in the Pkg struct, for example name and version. Is this allowed? or do I need to make sure it's only included once.

## Other Comments I have
- Vcpkg uses git internal git tree objects in order to fetch dependencies, so to hunt down information I relied heavily on go-git. I wanted to make sure I was getting as much relevant data as possible, and the files in the vcpkg build directory  on their own didn't seem to work.
- This is my first large (at least to me it's large) open source PR, so I'd love feedback on how I can improve.

## Type of change

# Checklist:


syft/pkg/cataloger/cpp/test-fixtures/vcpkg/helloworld/build/HelloWorld
wagoodman, line 1: looks like a binary was committed -- is this necessary?

syft/pkg/cataloger/cpp/test-fixtures/vcpkg/helloworld/build/CMakeFiles/3.28.3/CMakeDetermineCompilerABI_C.bin
wagoodman, line 1: bin committed, but probably not needed

syft/pkg/cataloger/cpp/test-fixtures/vcpkg/helloworld/build/CMakeFiles/3.28.3/CMakeDetermineCompilerABI_CXX.bin
wagoodman, line 1: bin committed, but probably not needed

syft/pkg/cataloger/cpp/test-fixtures/vcpkg/helloworld/build/CMakeFiles/HelloWorld.dir/helloworld.cpp.o
wagoodman, line 1: bin committed, but probably not needed

syft/pkg/cataloger/cpp/test-fixtures/vcpkg/helloworld/build/vcpkg_installed/x64-linux/debug/lib/libfmtd.a
wagoodman, line 1: large binary -- should not be committed

syft/pkg/cataloger/cpp/test-fixtures/vcpkg/helloworld/build/vcpkg_installed/x64-linux/lib/libfmt.a
wagoodman, line 1: bin committed, but probably not needed

syft/pkg/cataloger/cpp/test-fixtures/vcpkg/helloworld/build/.ninja_deps
wagoodman, line 1: bin committed, but probably not needed

wagoodman:
@gabetrau from our conversation in office hours, feel free to open a PR to https://github.com/anchore/vcpkg-test-fixture for any of the test fixtures you need. If you get a PR going that has at least the test fixtures in place , we can then get a ghcr.io image in that repo created that encapsulates any build processes if we need it. Even if there are no build processes we might still capture this as an image so we can test against the container image, which is a [common pattern in testing](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/redhat/cataloger_test.go#L19-L192) (the pattern may need tweaking to account for the external image).

gabetrau:
> @gabetrau from our conversation in office hours, feel free to open a PR to https://github.com/anchore/vcpkg-test-fixture for any of the test fixtures you need. If you get a PR going that has at least the test fixtures in place , we can then get a ghcr.io image in that repo created that encapsulates any build processes if we need it. Even if there are no build processes we might still capture this as an image so we can test against the container image, which is a [common pattern in testing](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/redhat/cataloger_test.go#L19-L192) (the pattern may need tweaking to account for the external image).

@wagoodman It might be simpler if we make the vcpkg-test-fixture repo a [custom vcpkg registry](https://learn.microsoft.com/en-us/vcpkg/maintainers/registries). Then I can just enable git clone for the test and remove all of the unneeded files from the helloworld example. [vcpkg-test-fixture](https://github.com/gabetrau/vcpkg-test-fixture). I updated this PR to use the custom one. Would this route be okay, or would you still prefer the ghcr.io image route?

Also, I'm having trouble making a PR to https://github.com/anchore/vcpkg-test-fixture since it is empty

spiffcs:
I've kicked off the checks on this one but since @wagoodman already has some context here no need to add more cooks to the kitchen.

houdini91:
Thanks for doing this i was looking in to vcpkg cataloger my self recently thankfully you one step ahead

syft/pkg/vcpkg.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// used for metadata. Includes summary of data found in manifest (vcpkg.json file) relevant to just that vcpkg
+type VcpkgManifest struct {
+	Description   []string `json:"description,omitempty"`
+	Documentation string   `json:"documentation,omitempty"`
+	FullVersion   string   `json:"full-version"`
+	Version       string   `json:"version"`
+	PortVersion   int      `json:"port-version"`
+	License       string   `json:"license,omitempty"`
```
wagoodman: I'll make one more adjustment; license shouldn't really be here, it should be on pkg.Package thus omitted from here.

wagoodman (approved):

PR #4106, opened 2025-07-31
support multiple letters in openssl patch version

honigbot:
# Description

Some openssl versions have two letter versions. This is not supported by the current binary-classifier regex:
https://github.com/anchore/syft/blob/801b21bb3780abf50a0e790ad1375e07c7342e94/syft/pkg/cataloger/binary/classifiers.go#L486

These two-letter versions are referenced for example here:
* https://openssl-library.org/news/secadv/20241016.txt
* https://openssl-library.org/news/secadv/20250120.txt

Furthermore, grype db lists two-letter versions, but without syft detecting multi-letter versions it generate false positives:
```sh
$ grype sbom-1.1.1zb.json 
NAME        INSTALLED  FIXED IN                                ...
openssl     1.1.1z     1.0.2zk, 1.1.1za, 3.0.15, 3.1.7, 3.2.3, ...
openssl     1.1.1z     1.0.2zl, 1.1.1zb, 3.0.16, 3.1.8, 3.2.4, ...
```

Therefore this patch fixes the pattern to allow multiple letters after the patch level:

- fix: accept '[a-z]+' letters in openssl regex

```
        NAME      VERSION  TYPE      
BEFORE  openssl   1.1.1z   binary  
 AFTER  openssl   1.1.1zb  binary
```

## Type of change

# Checklist:


kzantow (approved):
LGTM, thanks for the contribution @honigbot

PR #4118, opened 2025-08-06
fix: nondeterministic Java archive cataloging and improve groupID

### LINKED ISSUE #3521: PURL is not deterministic in java-archive cataloger
**What happened**:
When i scan a file (.war) I get different results each scan. A jar with multiple `pom.xml` can result in, for example:
```
pkg:maven/org.glassfish.jaxb/jaxb-core@2.2.11
pkg:maven/com.sun.xml.bind/jaxb-core@2.2.11
```

same result each time

Repeatedly scan `webgoat/webgoat` [container](https://hub.docker.com/r/webgoat/webgoat) or JAR [releases](https://github.com/WebGoat/WebGoat/releases)

it impacts the number of results I get from syft.

**Environment**:
- Output of `syft version`: 1.17.0

kzantow:
# Description

This PR fixes an issue where the Java archive cataloger could select the groupID in a nondeterministic fashion, resulting in different SBOMs on the same source material. Additionally, the logic to select the groupID and artifactID combination for Maven artifacts did not take into account the potential inclusion of groupID in the filename. As noted in the comment, this PR adjusts the behavior to be what I think is more correct when determining which pom the artifact represents: it will now favor POMs where both groupID and artifactID match, followed by exact match of the filename, and so on.

- Fixes #3521

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -1,15 +1,18 @@
 package java
 
 import (
+	"cmp"
 	"context"
 	"crypto"
 	"fmt"
 	"io"
+	"iter"
 	"os"
 	"path"
 	"slices"
 	"strings"
 
+	"github.com/scylladb/go-set/strset"
 	"golang.org/x/exp/maps"
 
 	"github.com/anchore/syft/internal"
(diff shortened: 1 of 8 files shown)


syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -391,20 +394,32 @@ func (j *archiveParser) discoverMainPackageFromPomInfo(ctx context.Context) (gro
 	projects, _ := pomProjectByParentPath(j.archivePath, j.location, j.fileManifest.GlobMatch(false, pomXMLGlob))
 
 	// map of all the artifacts in the pom properties, in order to chek exact match with the filename
-	artifactsMap := make(map[string]bool)
+	artifactsMap := strset.New()
```
kzantow, line 397: This is just for memory optimization.

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -784,3 +800,15 @@ func updateParentPackage(p pkg.Package, parentPkg *pkg.Package) {
 		parentPkg.Metadata = parentMetadata
 	}
 }
+
```
ariel-miculas, line 803: I think a [BTreeMap](https://pkg.go.dev/github.com/tidwall/btree#Map) would remove the need for this sortedIter.

kzantow: This is a good suggestion, but would require adding yet another library and reworking all the parts here using maps, which I'd like to avoid just for this small bugfix.

spiffcs (approved):

PR #4127, opened 2025-08-11
feat: combine go module file and go source discovery into single cataloger

### LINKED ISSUE #432: Catalog entire build list for Go projects, not just packages listed in go.mod
### Value

Today, we parse `go.mod` files and catalog the packages listed therein. However, this is not a complete list of software dependencies had by a given project, it's just a list of modules required by the project's "main module".

If users are to understand a project's complete set of project dependencies, we should consider surfacing the ["build list"](https://golang.org/ref/mod#glos-build-list), which is an exhaustive set of dependencies that a Go project has (i.e., the entire dependency graph). This list can be viewed by running the command `go list -m all` from a Go project's root directory.

### Implementation considerations

1. Go computes a set of required modules (i.e. generating the build list) using an approach called ["MVS"](https://research.swtch.com/vgo-mvs). It's important to note that one implication of this approach is that not all dependencies are identifiable using just a static analysis. Instead, the Go tooling issues multiple queries across a network (including to hosts like: GitHub (or other git providers), the [checksum database](https://golang.org/ref/mod#checksum-database), etc.) in order to discover indirect dependencies, and to select a single version of each Go module, even when multiple versions of a given Go module are cited throughout the dependency graph. Currently, Syft's analysis is entirely static. **Should Syft start incorporating analysis that depends on these network queries?** Always? Never? Sometimes (i.e. config value)?

spiffcs:
# Description

This PR enhances the GoModuleFileCataloger with the following improvements:

- Added golang.org/x/tools/go/packages integration - The cataloger now uses packages.Load with the "all" pattern to load complete dependency information when the Go toolchain is available for all `go.mod` files found for a given `dir` target scan
- modules are then used in combination with packages.Visit to build the full import graph to get a complete picture of the project's dependencies. It allows loading of go module license files from the mod cache into the license classifier for better license accuracy (see below notes).
- We also are able to extract the main module for given `go.mod` findings which is an improvement over the default process which just reads the required block (see below notes) 
- If the Go toolchain is not available or fails to load packages, the cataloger falls back to the original go.mod parsing behavior.

### Limitations
- The implementation first tries to use the Go toolchain, then fills in any missing packages from the go.mod file to ensure no dependencies are missed. This accounts for dependencies that might be locked behind build flags, architecture, os, or build tags that are not compatible with the current system, and other environment specific things that could affect what modules the go tooling chooses to reveal on dependency tree inspection using `go list`

The key changes were made to `syft/syft/pkg/cataloger/golang/parse_go_mod.go`

The enhancement provides more detailed dependency information when possible while maintaining backward compatibility with systems that don't have access to `go list` commands.

- Fixes #432 

### Notes

This PR was tested against multiple repositories to illustrate the improvements made to the cataloger:
```
caddy/
client/
consul/
cue/
flux2/
go-containerregistry/
go-struct-converter/
golangci-lint/
kind/
minio/
prometheus/
stereoscope/
```
#### Improved Package Discovery:
For things like `go-containerregistry` new modules are now discovered because of the source analysis:
```
syft -o json --override-default-catalogers go-module-file-cataloger dir:../go-source-test/repos/go-containerregistry > a.json
go run cmd/syft/main.go -o json --override-default-catalogers go-module-file-cataloger dir:../go-source-test/repos/go-containerregistry/ > b.json

jq -s '
 (.[0].artifacts | map(.name) // []) as $A |
 (.[1].artifacts | map(.name) // []) as $B |
 {
   only_syft:   (($A - $B) | unique | sort),
   only_go_run: (($B - $A) | unique | sort)
 }
' a.json b.json
```
```
{
  "only_syft": [],
  "only_go_run": [
    "github.com/google/go-containerregistry/cmd/krane",
    "github.com/google/go-containerregistry/pkg/authn/k8schain"
  ]
}
```

This above command can be run for all the test repositories to show new modules being included in the results.

#### Improved License Discovery
For a project like minio syft currently has issues discovering any of the licenses when given a `go.mod` file as part of the cataloging process
```
syft -o json --override-default-catalogers go-module-file-cataloger dir:../go-source-test/repos/minio | jq '.artifacts | [ .[] | select(.licenses == []) ] | length'
 ✔ Indexed file system                                                                                                                                                                 ../go-source-test/repos/minio
 ✔ Cataloged contents                                                                                                                               e1b7a8e4e8802913e7811ccc122df8d58a539e901b87e2b1f170a2acfd60b9dd
   ├── ✔ Packages                        [295 packages]
   ├── ✔ Executables                     [0 executables]
   ├── ✔ File metadata                   [5 locations]
   └── ✔ File digests                    [5 files]
295
```

With this PR we can see this number is down to 29
```
go run cmd/syft/main.go -o json --override-default-catalogers go-module-file-cataloger dir:../go-source-test/repos/minio | jq '.artifacts | [ .[] | select(.licenses == []) ] | length'
 ✔ Indexed file system                                                                                                                                                                 ../go-source-test/repos/minio
 ✔ Cataloged contents                                                                                                                               e1b7a8e4e8802913e7811ccc122df8d58a539e901b87e2b1f170a2acfd60b9dd
   ├── ✔ Packages                        [301 packages]
   ├── ✔ Executables                     [0 executables]
   ├── ✔ File metadata                   [6 locations]
   └── ✔ File digests                    [6 files]
29
```

Similar improvements in license detection and package discovery can be seen for the listed repositories.

Most of the packages that still do not have licenses can be attributed to the packages that are still necessarily discoverable only via the `go.mod` file. A small percentage also have licenses that do not return a result from the classifier.

The packages that still need to be discovered via the `go.mod` file can be because of things like build flags, build tags (including os, arch, or more complex header declarations), or other go build settings that are not set for a given environment when the underlying tools run `go list`
```
go run cmd/syft/main.go -o json --override-default-catalogers go-module-file-cataloger dir:../go-source-test/repos/minio |
jq '{
 go_source: ([.artifacts[] | select(.type=="go-source")] | length),
 go_module: ([.artifacts[] | select(.type=="go-module")] | length)
}'

 ✔ Indexed file system                                                                                                                                                                 ../go-source-test/repos/minio
 ✔ Cataloged contents                                                                                                                               e1b7a8e4e8802913e7811ccc122df8d58a539e901b87e2b1f170a2acfd60b9dd
   ├── ✔ Packages                        [301 packages]
   ├── ✔ Executables                     [0 executables]
   ├── ✔ File metadata                   [6 locations]
   └── ✔ File digests                    [6 files]
{
  "go_source": 281,
  "go_module": 20
}
```

### Potential Breaking Consideration
The go source package type was added here partially to show a difference of the origin for the packages given the new cataloging process. I'm not sure if this constitutes a breaking change or not. If this PR were to merge as is then users WOULD see a different output for identical scan targets between latest syft and the next release. This is because packages previously cataloged as `go-module` would be 'upgraded' to `go-source` given this new behavior.

I left the metadata as the same between the two and was able to merge the digest information from `go-sum` with the `go-source` information. If we want we can just remove the `go_source` type and leave these all as `go_modules` that might be the best way forward.

I've removed the configuration options for now for `test` and `searchPattern`

The `true` and `all` respective values for the above are more inline with syft's philosophy on cataloging ALL the things.
## Feature

# Checklist:
- Documentation needs to be updated
- [x] I have added unit tests that cover changed behavior <--- 


spiffcs:
Also fixes: https://github.com/anchore/syft/issues/3451

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
 
 	return out, nil
 }
+
+// loadPackages uses golang.org/x/tools/go/packages to get dependency information.
+func (c *goModCataloger) loadPackages(modDir string) (pkgs map[string][]pkgInfo, modules map[string]*packages.Module, dependencies map[string][]string) {
+	cfg := &packages.Config{
+		Mode:  packages.NeedModule | packages.NeedName | packages.NeedFiles | packages.NeedDeps | packages.NeedImports,
```
wagoodman, line 96: needs comments on why these options are selected and the performance impact of expanding this selection

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+		Dir:   modDir,
+		Tests: true,
+	}
+
+	// Load all packages for the given mod file
+	rootPkgs, err := packages.Load(cfg, "all")
```
wagoodman, line 109: need docs for `all` (seems important)

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+	// Check for any errors in loading
+	for _, p := range rootPkgs {
+		if len(p.Errors) > 0 {
+			// Log errors but continue processing
+			for _, e := range p.Errors {
+				log.Debugf("package load error for %s: %v", p.PkgPath, e)
+			}
+		}
+	}
+
```
wagoodman: not returning is the right call, but you also want to capture `unknowns` here since there were packages we expected to find but did not due to some reason. This should make it to the SBOM, but we should also keep processing partial results and return those.

wagoodman: side effect: that means that the top-level go.mod file that started this would have these `unknown` reasons hanging off of it (no matter where if this is for another module or a package within another module).

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+	case len(p.OtherFiles) > 0:
+		return filepath.Dir(p.OtherFiles[0])
+	default:
+		return ""
+	}
+}
+
+func shouldSkipVisit(p *packages.Package) bool {
+	// skip packages with errors
+	if len(p.Errors) > 0 {
```
wagoodman: this should raise up more `unknown` errors in these cases, which should be associated with the top-level go.mod that started this search

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+	// handle replace directives
+	// Note, we specifically want to replace version field.
+	// Haven't confirmed, but we may also need to override the
+	// entire struct when using replace directive with local folders.
+	tmp := *mod
+	if tmp.Replace != nil {
+		tmp = *tmp.Replace
+	}
+
+	// The +incompatible suffix does not affect module version.
```
wagoodman: this should really be handled in grype in the [golang matcher](https://github.com/anchore/grype/blob/main/grype/matcher/golang/matcher.go)

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+		if pkgDir == "" {
+			return true
+		}
+
+		module := newModule(p.Module)
+		if module.Dir == "" {
+			// A known cause is that the module is vendored, so some information is lost.
+			isVendored := strings.Contains(pkgDir, "/vendor/")
+			if !isVendored {
+				log.Warnf("module %s does not have dir and it's not vendored", module.Path)
```
wagoodman: nit: should be `log.debug...` (since the user cannot take any action to correct this)

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+
+		pkgDir := resolvePkgDir(p)
+		if pkgDir == "" {
+			return true
+		}
+
+		module := newModule(p.Module)
+		if module.Dir == "" {
+			// A known cause is that the module is vendored, so some information is lost.
+			isVendored := strings.Contains(pkgDir, "/vendor/")
```
wagoodman: this is a fun case -- there should probably be more comments here about the oddities around it.

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+// isStdLib returns true if this package is part of the Go standard library.
+func isStdLib(pkg *packages.Package) bool {
+	if pkg.Name == "unsafe" {
+		// Special case unsafe stdlib, because it does not contain go files.
+		return true
+	}
+	if len(pkg.GoFiles) == 0 {
+		return false
+	}
+	prefix := build.Default.GOROOT
```
wagoodman, line 440: this probably warrants a TODO -- are there other locations that would not be in the default goroot? (not asking for a code change)

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+	}
+
+	if !strings.HasPrefix(dir, rootDir) {
+		return nil, fmt.Errorf("licenses.Find: rootDir %s should contain dir %s", rootDir, dir)
+	}
+
+	return findAllUpwards(dir, licenseRegexp, rootDir)
+}
+
+func findAllUpwards(dir string, r *regexp.Regexp, stopAt string) ([]string, error) {
```
wagoodman:
```suggestion
func findAllLicenseCandidatesUpwards(dir string, r *regexp.Regexp, stopAt string) ([]string, error) {
```

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+			}
+
+			if r.MatchString(f.Name()) {
+				path := filepath.Join(dir, f.Name())
+				licenseCandidates = append(licenseCandidates, path)
+			}
+		}
+
+		parent := filepath.Dir(dir)
+		if parent == dir {
```
wagoodman: I think there is some risk of not breaking here, and we loop forever, but I haven't drawn up a specific case yet. I think the bothersome thing is that the input stopAt might not trigger if there is a symlink that gets resolved somehow within the loop and the pathing changes (don't see how that's possible yet). But I do think this needs one more pass at defensiveness.

wagoodman: consider using afero and removing the os calls in this function and setting up some memory mapped FSs during testing for this function.

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +182,353 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+		goModulePkg.SetID()
+
+		moduleToPackage[m.Path] = goModulePkg
+		syftPackages = append(syftPackages, goModulePkg)
+	}
+
+	return syftPackages, moduleToPackage
+}
+
+// resolveModuleLicenses finds and parses license files for Go modules.
```
wagoodman: probably needs more comments here as to why a bubble-up search per-package is done instead of a walk-down per-module is done. We talked through several pros, but it took a while to see them all (so worth capturing).

syft/pkg/type.go
```diff
@@ -24,6 +24,7 @@ const (
 	GithubActionPkg         Type = "github-action"
 	GithubActionWorkflowPkg Type = "github-action-workflow"
 	GoModulePkg             Type = "go-module"
+	GoSourcePkg             Type = "go-source"
```
wagoodman: we should use the existing `go-module`

wagoodman (commented):
we still need to find a way to nicely distinct which packages come from which sub-cataloger functions. The best option for this might be to express more metadata, thus make a new metadata type (but needs a specific recommendation)

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -32,8 +41,20 @@ func newGoModCataloger(opts CatalogerConfig) *goModCataloger {
 //
 //nolint:funlen
 func (c *goModCataloger) parseGoModFile(ctx context.Context, resolver file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
```
wagoodman, line 42: this method is mixing lower level concerns with higher level concerns -- could use a one-over for decomposing into smaller functions with smaller concerns.

syft/pkg/cataloger/golang/parse_go_mod_test.go
wagoodman, line 1: need to update existing tests: if load.packages fails, we still can gracefully degrade and do simple go.mod cataloging. This functionality is important from two perspectives:
1. the fallback is happening
2. the fallback is not fatal
3. the fallback still yields good results

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: Comments have been added here describing the options selected

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: Added documentation for selection of "all"

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: updated visit and root load to now consider unknown errors for (p.Errors)
unknown errors are now returned as part of parseGoModFile

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: We no longer consider errors as part of the skip visit function. It's been lifted into `packages.Visit`, cataloged as a part of the unknown errors, and then the package is skipped.

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: removed this from the cataloger. I have made a note to check this over in the golang matcher.

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: Added comments describing this case

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: transformed this to specific license candidates and injected FS with afero

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: added afero and additional testing showing how new fail safes can break us out of the loop and make sure we don't infinitely resolve here:
- added visited map to track directories; break in the case of retrace
- added safety check where we ensure parent is actually a parent; break if we are outside of parent

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: bubble-up strategy outlined. Added consideration for when we might want to implement Tip to stem. Specifically added how bubble-up prevents licenses from being associated to incorrect 'parent module'.

syft/pkg/type.go
spiffcs: Removed and updated schema with new metadata

syft/pkg/cataloger/golang/parse_go_mod_test.go
spiffcs: Note - the original tests `TestParseGoMod` have not been altered. The expected packages contain the metadata `GolangModuleEntry` - This metadata is associated with the fallback go.mod cataloging. These tests are expressing a call to `packages.Load` which is failing and then falling back to the original catalog method.

If these tests were to fail it would show that the fallback is not working correctly (fatal). If the expectations needed to be changed in anyway it would show that these new changes may have altered the results in an undesirable way.

This is because the target for the tests is `FromFile`, the cataloger is not considering any other inputs and thus falls back to the default `go.mod` parsing.

spiffcs:
> we still need to find a way to nicely distinct which packages come from which sub-cataloger functions. The best option for this might be to express more metadata, thus make a new metadata type (but needs a specific recommendation)

I added a new metadata type and updated the schema for this: `GolangSourceEntry` 

It keeps the H1Digest while also added information about the Go environment that the package was found in
```
		H1Digest:   h1Digest,
		GOROOT:     build.Default.GOROOT,
		GOPATH:     build.Default.GOPATH,
		GOOS:       build.Default.GOOS,
		GOARCH:     build.Default.GOARCH,
		Compiler:   build.Default.Compiler,
		BuildTags:  strings.Join(build.Default.BuildTags, ","),
		CgoEnabled: build.Default.CgoEnabled,
```

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: I'm not sure why github is not showing my changes here on the PR, but if you pull you'll see the new comments and docs/refactoring that was asked for in this file 😢 

Edit: It loaded for me when I switched between the old/new experience views

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: This is now debug

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: decomposed this to make it more readable, but there are still some overlaps in concerns. I'm going to take another stab at this in a second.

spiffcs:
I've updated the branch so we're building a new schema since main moved forward before this was merged

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -93,25 +129,35 @@ func (c *goModCataloger) parseGoModFile(ctx context.Context, resolver file.Resol
...
-	for _, p := range packages {
+	for _, p := range catalogedPkgs {
 		p.SetID()
-		pkgsSlice[idx] = p
-		idx++
+		pkgsSlice = append(pkgsSlice, p)
+	}
+
+	for _, p := range goModPackages {
+		p.SetID()
```
wagoodman: we should be setting ID when we craft the package.

wagoodman: try to follow package constructor conventions in both file organization and constructor functions , e.g. https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/alpine/package.go

wagoodman: current bug: these IDs are finalized after the relationships are created

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -151,3 +197,445 @@ func parseGoSumFile(resolver file.Resolver, reader file.LocationReadCloser) (map
...
+	// Stop once we go out of the stopAt dir.
+	licenseCandidates := make([]string, 0)
+	visited := make(map[string]bool) // Track visited directories to prevent infinite loops
+
+	for strings.HasPrefix(dir, stopAt) {
+		// Resolve any symlinks to get the actual path
+		// Note: For in-memory filesystems, EvalSymlinks may not work as expected,
+		// but this provides a fallback for real filesystems
+		resolvedDir := dir
+		if _, ok := fs.(*afero.OsFs); ok {
```
wagoodman: take a look at https://github.com/spf13/afero/blob/2fb20ad3b2f7a6be83bd142888ea702c3388d3cc/symlink.go#L20-L50 to prevent the need to type assert to the concrete implementation and use the stdlib EvalSymlinks

spiffcs: I've updated this code to use the afero.LinkReader interface and added new test cases

syft/pkg/cataloger/golang/parse_go_mod.go
spiffcs: @wagoodman moved ID creation to right after package generation

wagoodman (approved):
🎉

PR #4132, opened 2025-08-12
feat: update cyclonedx encoder to use url as Name field for empty case

### LINKED ISSUE #1964: Can not have license ID
Can not have license ID
Exec Command : 
```
syft packages sonatype/nexus3 -o cyclonedx-json
```
But can not have license id  :
```json
      "licenses": [
        {
          "license": {
            "name": "http://www.apache.org/licenses/LICENSE-2.0.txt"
          }
        }
      ],
```
Good json：
```json
      "licenses": [
        {
          "license": {
            "id": "Apache-2.0"
          }
        }
      ],
```

spiffcs:
# Description
This updates the cyclonedx encoder so that if we encode a license that ONLY has a URL we also include it as the name to maintain a valid cyclonedx format.

- Fixes #1964

## Type of change

# Checklist:

### DIFF
--- a/syft/format/internal/cyclonedxutil/helpers/licenses.go
+++ b/syft/format/internal/cyclonedxutil/helpers/licenses.go
@@ -163,10 +163,16 @@ func processCustomLicense(l pkg.License) cyclonedx.Licenses {
 func processLicenseURLs(l pkg.License, spdxID string, populate *cyclonedx.Licenses) {
 	for _, url := range l.URLs {
 		if spdxID == "" {
+			// CycloneDX requires either an id or name to be present for a license
+			// If l.Value is empty, use the URL as the name to ensure schema compliance
+			name := l.Value
+			if name == "" {
+				name = url
+			}
 			*populate = append(*populate, cyclonedx.LicenseChoice{
 				License: &cyclonedx.License{
 					URL:  url,
-					Name: l.Value,
+					Name: name,
 				},
 			})
 		} else {
--- a/syft/format/internal/cyclonedxutil/helpers/licenses_test.go
+++ b/syft/format/internal/cyclonedxutil/helpers/licenses_test.go
@@ -122,6 +122,22 @@ func Test_encodeLicense(t *testing.T) {
 			},
 		},
 		{
+			name: "with URL only and no name or SPDX ID",
+			input: pkg.Package{
+				Licenses: pkg.NewLicenseSet(
+					pkg.NewLicenseFromURLsWithContext(ctx, "", "http://jaxen.codehaus.org/license.html"),
+				),
+			},
+			expected: &cyclonedx.Licenses{
+				{
+					License: &cyclonedx.License{
+						Name: "http://jaxen.codehaus.org/license.html",
+						URL:  "http://jaxen.codehaus.org/license.html",
+					},
+				},
+			},
+		},
+		{
 			name: "with multiple values licenses are deduplicated",
 			input: pkg.Package{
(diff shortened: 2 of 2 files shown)


internal/spdxlicense/license.go
```diff
@@ -35,3 +35,22 @@ func cleanLicenseID(id string) string {
...
+}
+
+// LicenseByURL returns the license ID and name for a given URL from the SPDX license list
+// The URL should match one of the URLs in the seeAlso field of an SPDX license
+func LicenseByURL(url string) (LicenseInfo, bool) {
+	url = strings.TrimSpace(url)
+	if info, exists := urlToLicense[url]; exists {
+		return LicenseInfo{
+			ID:   info.ID,
+			Name: info.Name,
```
kzantow: For SPDX, this should output either ID or Name, I believe. We're only replacing known URLs with SPDX IDs, right? including the name here seems like it isn't very useful. (This is a follow-on to the comment above that I think we could just track URL -> ID)

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -31,6 +31,16 @@ var licenseIDs = map[string]string{
...
 {{- end }}
 }
+
+// urlToLicense maps license URLs from the seeAlso field to license IDs and names
+var urlToLicense = map[string]struct {
+	ID   string
+	Name string
+}{
+{{- range $url, $info := .URLToLicense }}
+	{{ printf "%q" $url }}: {ID: {{ printf "%q" $info.ID }}, Name: {{ printf "%q" $info.Name }}},
```
kzantow: Does this need to have both ID and Name? It would seem to me that all of these have IDs, and we're only concerned with URLs that map to known IDs  in order to output the SPDX ID across formats, and there's no need for the name. Are there licenses here that don't have SPDX IDs? If not, could we make this a simpler map of just URL -> ID ?

internal/spdxlicense/license.go
spiffcs: Yea I can remove name here. I was a little on the fence including it, but thought more information better. I agree though that you can get everything you want from just the ID

internal/spdxlicense/generate/generate_license_list.go
spiffcs: Can and will remove name. It was a fence sitter for me too and I just erred on including it if we wanted it in the future.

internal/spdxlicense/license.go
spiffcs: These have been removed

internal/spdxlicense/generate/generate_license_list.go
spiffcs: These have been removed

internal/spdxlicense/generate/generate_license_list.go
```diff
@@ -31,6 +31,15 @@ var licenseIDs = map[string]string{
 	{{ printf "%q" $k }}: {{ printf "%q" $v }},
 {{- end }}
 }
+
+// urlToLicense maps license URLs from the seeAlso field to license IDs and names
+var urlToLicense = map[string]struct {
```
kzantow: Since this is just ID now, I think this could be `map[string]string`, which would simplify a little more stuff throughout this PR

kzantow (approved):
Optional suggestion, otherwise LGTM 👍

internal/spdxlicense/generate/generate_license_list.go
spiffcs: Sending that up now

Issue #3734, opened 2025-03-14
Add arbitrary name-value pairs to SBOM


kzantow:
**What would you like to be added**:
A request from @henrysachs [on Discourse](https://anchorecommunity.discourse.group/t/what-are-the-plans-for-metadata-in-the-syft-format/318): add the ability to specify arbitrary name-value pairs that are added to the SBOM.

**Why is this needed**:
Users have various needs -- particularly at the SBOM-level -- to add information that is useful for downstream consumers.

**Additional context**:
After a discussion during weekly gardening, it was determined that we could probably move forward with this, and start with the simplest solution of adding a name-value `map[string]string` somewhere. We need to work out the UX for this, but it would probably be something like adding a multiple-allowed flag `--sbom-property <name>=<value>`.

I believe the consensus is that this is probably best suited to be added to [the SBOM itself](https://github.com/anchore/syft/blob/main/syft/sbom/sbom.go#L15), rather than being associated with the _source_, as the `--source-name` and `--source-version` are.

Additional concerns are: these need to be output in a reproducible order in all serialization formats. JSON output of map[string] in go will do this automatically, but if we need to iterate a map, for example to add CycloneDX properties, this needs to be done in an ordered way.

henrysachs:
Hey @kzantow thanks for adding this issue and that you're open for implementations. I think this nails the idea that I also had in mind and I would like to help implementing this.

PR #4142, opened 2025-08-17
feat: add custom SBOM tags support

henrysachs:
Add support for arbitrary key-value pairs in SBOMs through new --sbom-property CLI flags. This enables users to add custom metadata to SBOMs for better organization and compliance tracking.

Closes: #3734

# Description

- Fixes

## Type of change

# Checklist:


henrysachs:
@spiffcs @kzantow maybe you could talk about that PR in the livestream today?

henrysachs:
I rebased the branch to match the latest syft changes and regenerated the schema

cmd/syft/internal/options/catalog.go
```diff
@@ -43,6 +43,9 @@ type Catalog struct {
 	Compliance        complianceConfig    `yaml:"compliance" json:"compliance" mapstructure:"compliance"`
 	Enrich            []string            `yaml:"enrich" json:"enrich" mapstructure:"enrich"`
 
+	// custom SBOM properties
+	SBOMProperties []string `yaml:"sbom-properties" json:"sbom-properties" mapstructure:"sbom-properties"`
```
kzantow, line 47: Should is be a `map[string]string`? I _think_ this is properly supported with all the configuration tooling, but we'd need a different field for the CLI flags of the type `[]string`, probably

syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -301,19 +301,41 @@ func toDependencies(relationships []artifact.Relationship) []cyclonedx.Dependenc
...
-	return nil
+
+	// Add SBOM tags with guaranteed alphabetical order
+	if len(tags) > 0 {
+		// Get sorted keys for reproducible output
+		keys := make([]string, 0, len(tags))
+		for k := range tags {
+			keys = append(keys, k)
+		}
+		slices.Sort(keys)
```
kzantow: I think you can just use `slices.Sorted(maps.Keys(theMap))`

syft/format/common/spdxhelpers/to_format_model.go
```diff
@@ -159,6 +159,28 @@ func ToFormatModel(s sbom.SBOM) *spdx.Document {
...
+	keys := make([]string, 0, len(tags))
+	for k := range tags {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+
+	var parts []string
+	parts = append(parts, "SBOM Properties:")
+	for _, key := range keys {
+		parts = append(parts, fmt.Sprintf("  %s: %s", key, tags[key]))
```
kzantow: Is there a reason for whitespace here? I think newlines _shouldn't_ cause a problem, but maybe it's safer to use a different separator? Have you tried this in Tag-Value format? maybe we just output this like `<key>=<value>;<key>=<value>`, I think the `=` at least, is more consistent with the configuration here?

syft/format/syftjson/model/document.go
```diff
@@ -7,13 +7,14 @@ import (
...
+	Artifacts             []Package         `json:"artifacts"` // Artifacts is the list of packages discovered and placed into the catalog
+	ArtifactRelationships []Relationship    `json:"artifactRelationships"`
+	Files                 []File            `json:"files,omitempty"` // note: must have omitempty
+	Source                Source            `json:"source"`          // Source represents the original object that was cataloged
+	Distro                LinuxRelease      `json:"distro"`          // Distro represents the Linux distribution that was detected from the source
+	Descriptor            Descriptor        `json:"descriptor"`      // Descriptor is a block containing self-describing information about syft
+	Schema                Schema            `json:"schema"`          // Schema is a block reserved for defining the version for the shape of this JSON document and where to find the schema document to validate the shape
+	Tags                  map[string]string `json:"tags"`            // Tags contains custom properties added to the SBOM
```
kzantow: `Tags` seems like it might be misleading, confused with image tags maybe? Is there another term we could use here to be more clear what this is used for? CDX `Properties` seems a bit better, to me, maybe we adopt `Properties`?

syft/format/syftjson/to_format_model.go
```diff
@@ -85,6 +86,18 @@ func toDescriptor(d sbom.Descriptor) model.Descriptor {
 	}
 }
 
+func toTags(tags map[string]string) map[string]string {
+	if tags == nil {
+		return make(map[string]string)
+	}
+	// Create a copy to avoid modifying the original map
```
kzantow: Why would this be modified?

henrysachs:
@kzantow all recommendations should be implemented now. I just wrestle a bit with github actions 😅

I personally liked tags more as my background is from cloud providers, but i understand the "conflict" with image tags and switched to properties by now.

syft/format/common/spdxhelpers/to_format_model.go
henrysachs: implemented

cmd/syft/internal/options/catalog.go
henrysachs: i didnt quite follow through this comment but it should work

syft/format/syftjson/to_format_model.go
henrysachs: i removed this completly as I didnt even quite knew myself when this is modyfied 😅

cmd/syft/internal/options/catalog.go
kzantow: Hey @henrysachs, sorry for the delay. As I noted in the [comment to your colleague](https://github.com/anchore/syft/pull/4378/files#r2586292128), we've had some time off affecting the team's availability to get some answers for this set of PRs, that I think are directly related. I tried to leave information pertaining to both PRs there, so it makes sense where the end goal is, essentially, from where this PR is:
* `SBOMProperties` is moved to a top-level options struct: `sbomConfig.Properties` in a syft-config-friendly way
* there is nothing added to the _cataloging config_, but these are directly set via CLI and formats on the SBOM
* CycloneDX should symmetrically encode and decode the top-level properties (this should fix the test failures)

This PR is otherwise very close 👍

PR #4150, opened 2025-08-21
chore: redhat cataloger error when sqlite not regsitered

### LINKED ISSUE #3234: Warn loudly if SQLite driver is not present when needed
The SBOM generated for the MySQL:8.x container does not include the RPM packages. It seems that important package information is missing

the code for generating the SBOM:

```
func analyzeImageUsingSyft(l *logger.Logger, imageId string) (source.Source, *sbom.SBOM, error) {

	img, err := stereoscope.GetImageFromSource(context.Background(), imageId, oci.Registry, stereoscope.WithPlatform("linux/amd64"))
	if err != nil {
		l.Error("Could not create image source object. err: %v", err)
		return nil, nil, err
	}

	imageSource := stereoscopesource.New(img, stereoscopesource.ImageConfig{Reference: imageId})
	if err != nil {
		l.Error("Could not pull image: %s. err: %v", imageId, err)
		return nil, nil, err
	}

	s, err := getSBOM(imageSource, true)
	if err != nil {
		l.Error("Could get image SBOM. image: %s. err: %v", imageId, err)
		return nil, nil, err
	}
	return imageSource, &s, nil
}

func getSBOM(src source.Source, saveToFile bool) (sbom.SBOM, error) {
	s, err := syft.CreateSBOM(context.Background(), src, cfg)
	if err != nil {
		return sbom.SBOM{}, err
	}
	return *s, nil
}
```

Other images with RPMs are getting scanned correctly, but this issue seems specific to MySQL:8.x.

**Environment**:
- syft version: 1.11.1
- OS: tested on Windows,Linux and macOS

kzantow:
# Description

This PR causes a syft scan to return an error when there is no SQLite driver registered. This may occur when using Syft as a library and not including the import `_ "modernc.org/sqlite"`.

- Fixes #3234

## Type of change

# Checklist:

### DIFF
--- a/examples/create_simple_sbom/main.go
+++ b/examples/create_simple_sbom/main.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 
+	_ "modernc.org/sqlite"
+
 	"github.com/anchore/syft/syft"
 	"github.com/anchore/syft/syft/format"
 	"github.com/anchore/syft/syft/format/syftjson"
--- a/examples/create_simple_sbom/main.go
+++ b/examples/create_simple_sbom/main.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"os"
 
-	_ "modernc.org/sqlite"
+	_ "modernc.org/sqlite" // required for rpmdb and other features
 
 	"github.com/anchore/syft/syft"
 	"github.com/anchore/syft/syft/format"
--- a/syft/pkg/cataloger/redhat/cataloger.go
+++ b/syft/pkg/cataloger/redhat/cataloger.go
@@ -5,8 +5,8 @@ package redhat
 
 import (
 	"database/sql"
+	"fmt"
 
-	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/pkg/cataloger/generic"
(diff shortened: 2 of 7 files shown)


examples/create_simple_sbom/main.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 
+	_ "modernc.org/sqlite"
```
wagoodman: nit: might want to leave a comment as to why this is needed

syft/pkg/cataloger/redhat/cataloger.go
```diff
@@ -47,11 +43,11 @@ func NewArchiveCataloger() pkg.Cataloger {
 		WithParserByGlobs(parseRpmArchive, "**/*.rpm")
 }
 
-func isSqliteDriverAvailable() bool {
+func ensureSqliteDriverAvailable() error {
 	db, err := sql.Open("sqlite", ":memory:")
 	if err != nil {
-		return false
+		return fmt.Errorf("sqlite driver is required for cataloging newer RPM databases, none registered: %v", err)
```
wagoodman, line 49: how does this get expressed? as a log.warn or a log.error?

kzantow: This causes the scan to fail, which I believe is the right behavior: if they are using this cataloger and have not registered a SQLite driver. If they do not want to register a sqlite driver, they can skip this cataloger explicitly.

Issue #3718, opened 2025-03-11
Extend integration tests for GraalVM Native Image Cataloger


rudsberg:
**What would you like to be added**:
The integration tests for the GraalVM Native Image Cataloger should be extended to ensure `syft` is able to extract embedded SBOMs from Native Images. Such end-to-end integration tests would create a Native Image with an embedded SBOM, invoke `syft`, and make assertions on the extracted SBOM. 

There are currently two tests for the GraalVM Native Image Cataloger. [`TestParseNativeImage`](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/graalvm_native_image_cataloger_test.go#L22) is a negative test, testing that the SBOM extraction fails for a Native Image _without_ an embedded SBOM. [`TestParseNativeImageSbom`](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/graalvm_native_image_cataloger_test.go#L65) mocks a compressed SBOM in a Native Image and tests the [decompressSbom](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go#L96) function, a subset of the Cataloger.

For comprehensive test coverage, the implementation should generate Native Image executables in all three major executable formats: ELF, Mach-O, and PE. This would ensure that all platform-specific code paths in the Cataloger are exercised.

**Why is this needed**:
This addresses a gap in the testing of `syft` and will help to ensure that the `syft` integration with GraalVM Native Image remains functional across all supported platforms.

**Additional context**:
* To create a Native Image with an embedded SBOM, Oracle GraalVM (not the community version) must be used. It can be pulled from [Oracle's container registry](https://container-registry.oracle.com/ords/f?p=113:4:116479161087584:::RP,4:P4_REPOSITORY,AI_REPOSITORY,P4_REPOSITORY_NAME,AI_REPOSITORY_NAME:2204,2204,Oracle%20GraalVM%20Native%20Image%20released%20under%20GFTC,Oracle%20GraalVM%20Native%20Image%20released%20under%20GFTC&cs=371p79AR9JXwAlKNzhG3sISpwJq6BS8jF-e-_7Yeflqrk4eZYHLm179fZVHt-vJBT0ctAMZUg8UH7UIdCDfZUgA).
[...]

PR #4186, opened 2025-09-03
Native Image SBOM: Add Support for Locations Data

rudsberg:
# Description
This PR adds support to the `graalvm-native-image-cataloger` to associate location data with the packages extracted from native images. Packages extracted are now associated with `file.Location` metadata, which refers to the native image. 

This enables users to easily trace packages back to the native image executable, simplifying vulnerability identification.

## Example
Consider this scenario: `syft` scans a Docker container containing a native image named `app` that has a dependency on `org.json`. The `org.json` artifact currently looks like this in the `syft-json` format:
```
{
    "id": "pkg:maven/org.json/json@20250517",
    "name": "json",
    "version": "20250517",
    "type": "java-archive",
    "foundBy": "graalvm-native-image-cataloger",
    "locations": [], <-- locations is empty
    ...
}
```
This PR ensures the locations data is added:
```
{
    "id": "pkg:maven/org.json/json@20250517",
    "name": "json",
    "version": "20250517",
    "type": "java-archive",
    "foundBy": "graalvm-native-image-cataloger",
    "locations": [
        {
            "path": "/app",
            "layerID": "sha256:f1cf2e7664a1a9e45052d20a7061d9e46eb93091e092b4aa4166197cee5a3604",
            "accessPath": "/app"
        }
    ],
    ...
}
```
## Type of change

# Checklist:
  _No new unit tests were added since the logic is trivial and there is currently no straightforward way to test it. Issue [#3718](https://github.com/anchore/syft/issues/3718) would enable end-to-end tests using native images with embedded SBOMs._


douglasclarke:
@kzantow

syft/pkg/cataloger/java/graalvm_native_image_cataloger.go
```diff
@@ -534,6 +535,10 @@ func fetchPkgs(reader unionreader.UnionReader, filename string) ([]pkg.Package,
 				log.Tracef("unable to extract SBOM from possible java native-image %s: %v", filename, err)
 				continue
 			}
+			// Associate extracted packages with the native image location
+			for i := range newPkgs {
+				newPkgs[i].Locations.Add(location)
```
spiffcs, line 540: I think this requires a quick review of the different implementations of `fetchPkgs` depending on the imageFormats being considered here. 

It's my understanding that after `fetchPkgs` is called here the ID is already set on the package. The location set IS considered during the ID generation so we don't want to mutate AFTER the ID has been set:
https://github.com/anchore/syft/blob/2531bfd8cb9c08c8b956e5ca53c4a72aa8be949e/syft/pkg/package.go#L19-L45

I'll take some time to look at how the interface is implemented or the different formats to confirm this when I have a few more cycles.

I'll slot this PR into our gardening this PM to get a quick design discussion on how we might want to inject this location(s) (i think it's only one since it's going from filename -> location but have to check that possibility as well) into the package construction and not do a post assemble mutation  😄

spiffcs: Ok - all roads lead to decompressSbom: https://github.com/anchore/syft/blob/2531bfd8cb9c08c8b956e5ca53c4a72aa8be949e/syft/pkg/cataloger/java/graalvm_native_image_cataloger.go#L96-L133

If we modify the LocationSet after doing a decode from the discovered SBOM do we need to alter the IDs based on the new package information we created? Do we get unexpected behavior in the library if we modify this field and then the ID on the package is technically different given we didn't hash this field on decode. Should we generate a new value in this case and update the relationships or since `ID()` returns the already set `ID` from decode we just know that this is the value being preserved going forward?

spiffcs: We discussed this during our internal standup. Since we're calling `OverrideID` (line 70 below) https://github.com/anchore/syft/blob/c6cd66357a8c738a13193415a2438e450a0c8f87/syft/format/internal/cyclonedxutil/helpers/decoder.go#L61-L86 on decode here we're already preserving the CycloneDX ID from the document so the modification of packages here doesn't impact the recompute downstream. The ref from the document is the canonical ID for this package and what's used in the relationships so mutating the package has no affect since the `SetID` was not called for the package. Note: there is a code path where id there is no ref THEN we are using the syft id which would make this mutation conflict with what `SetID` came up with vs what eventually ended up on the package struct.

There are some considerations for syft 2.0 on ID preservation that we'll cover in another issue. 

Does mutating the location data change the package in a significant enough way that we need an identifier that syft knows about, while also preserving the ID we got from reading another SBOM. Since that's a world where there are two types of id (internal syft/external document) it's v2.0 land and has no consideration for the 👍  or 👎  on the 🟢 for this contribution.

PR #4207, opened 2025-09-10
feat: CPEs format decoder

chovanecadam:
# Description

Closes: 

Adds a CPE decoder to aid Grype with scanning a file of CPEs. I have a WIP commit that implements that here: https://github.com/chovanecadam/grype/commit/53bd626ae444ea1d13615dd948e0f1592c696dbb

This is my first bigger contribution to Syft, so please double-check if I didn't make some mistake.

I didn't implement an encoder. Syft generates many possible CPEs, I don't see a use-case for this output format.

- Fixes https://github.com/anchore/syft/issues/3890

## Type of change

# Checklist:

I have adopted the unit tests from [grype/grype/pkg/cpe_provider_test.go](https://github.com/chovanecadam/grype/blob/main/grype/pkg/cpe_provider_test.go) as this basically implements the same functionality.

### DIFF
--- a/syft/format/internal/backfill.go
+++ b/syft/format/internal/backfill.go
@@ -10,6 +10,7 @@ import (
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/cpe"
 	"github.com/anchore/syft/syft/pkg"
+	cataloger "github.com/anchore/syft/syft/pkg/cataloger/common/cpe"
 )
 
 // Backfill takes all information present in the package and attempts to fill in any missing information
@@ -29,7 +30,7 @@ func backfillFromCPE(p *pkg.Package) {
 	c := p.CPEs[0]
 
 	if p.Type == "" {
-		p.Type = CPETargetSoftwareToPackageType(c.Attributes.TargetSW)
+		p.Type = cataloger.TargetSoftwareToPackageType(c.Attributes.TargetSW)
 	}
 }
 
--- a/syft/pkg/cataloger/common/cpe/target_software_to_pkg_type.go
+++ b/syft/pkg/cataloger/common/cpe/target_software_to_pkg_type.go
@@ -1,4 +1,4 @@
-package internal
+package cpe
 
 import (
 	"strings"
@@ -6,9 +6,9 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
-// CPETargetSoftwareToPackageType is derived from looking at target_software attributes in the NVD dataset
+// TargetSoftwareToPackageType is derived from looking at target_software attributes in the NVD dataset
 // TODO: ideally this would be driven from the store, where we can resolve ecosystem aliases directly
-func CPETargetSoftwareToPackageType(tsw string) pkg.Type {
+func TargetSoftwareToPackageType(tsw string) pkg.Type {
 	tsw = strings.NewReplacer("-", "_", " ", "_").Replace(strings.ToLower(tsw))
 	switch tsw {
 	case "alpine", "apk":


chovanecadam:
Most of the code is adopted from decoder of package urls in syft and the tests are taken from Grype. I don't want to make it seem like I claim authorship of that.

chovanecadam:
Is there anything I can do to move this forward?

kzantow:
Thanks for the contribution @chovanecadam, sorry it got lost in the noise 🤦

For setting the package type from the CPE target_sw, I think we could move that to the [Backfill process](https://github.com/anchore/syft/blob/af4d19f81dd26de095214c33a9c43ac3dbf3323f/syft/format/internal/backfill.go#L51) as the very last step if we could not determine package type otherwise. This would help to improve support for SBOMs with only CPEs and no other way to determine package type, too.

~~Also, would it be worthwhile adding a CPE output format? I'm inclined to say no, since it is such a lossy identifier compared to PURLs, which we _do_ output.~~ Leaving this input-only for now makes the most sense to me.

syft/format/internal/backfill.go
```diff
@@ -13,10 +13,32 @@ import (
...
-// from any available sources, such as the Metadata and PURL.
+// from any available sources, such as the Metadata, PURL, or CPEs.
 //
 // Backfill does not call p.SetID(), but this needs to be called later to ensure it's up to date
 func Backfill(p *pkg.Package) {
+	if p.PURL != "" {
+		backfillFromPurl(p)
+	}
+
+	if len(p.CPEs) != 0 {
```
kzantow: nit: is there a reason to have this check, since the `backfillFromCPE` is doing the same thing?

syft/pkg/cataloger/common/cpe/cpe_target_software_to_pkg_type.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"strings"
+
+	"github.com/anchore/syft/syft/pkg"
+)
+
+// CPETargetSoftwareToPackageType is derived from looking at target_software attributes in the NVD dataset
+// TODO: ideally this would be driven from the store, where we can resolve ecosystem aliases directly
+func CPETargetSoftwareToPackageType(tsw string) pkg.Type {
```
kzantow: We should add some basic test for this function, probably in the decoder_test above -- I don't see any CPEs with targetSW set that result in execution of this logic, did I miss this or could you add a couple tests?

syft/format/cpes/decoder_test.go
```diff
@@ -0,0 +1,101 @@
...
+			sbom: &sbom.SBOM{
+				Artifacts: sbom.Artifacts{
+					Packages: pkg.NewCollection(),
+				},
+			},
+		},
+	}
+
+	syftPkgOpts := []cmp.Option{
+		cmpopts.IgnoreFields(pkg.Package{}, "id", "Type", "Language"),
```
kzantow: We probably want to actually test the `Type` here and add at least one test exercising multiple CPEs

syft/format/internal/backfill.go
chovanecadam: good nit, I removed superfluous tests in b3da15b5541e

syft/pkg/cataloger/common/cpe/cpe_target_software_to_pkg_type.go
chovanecadam: I added a few test cases to the cpe decoder_test as well as to the backfill_test

syft/format/cpes/decoder_test.go
chovanecadam: Good point.

chovanecadam:
@kzantow I added test cases and rebased on top of main

syft/format/cpes/decoder_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/require"
+
+	"github.com/anchore/syft/syft/cpe"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/sbom"
+)
+
+func Test_CPEProvider(t *testing.T) {
```
wagoodman, line 17: we should also test a list of cpes as well

syft/pkg/cataloger/common/cpe/target_software_to_pkg_type.go
wagoodman, line 1: This was copied from grype/internal/cpe_target_software_to_pkg_type.go , which is ok, however, now we have two sources of truth for this kind of functionality. We should probably consider moving this to syft/pkg/cataloger/common/cpe/* expose this on the public API for grype to use.

syft/format/cpes/decoder_test.go
chovanecadam: thanks for the review, I added the tests

syft/pkg/cataloger/common/cpe/target_software_to_pkg_type.go
chovanecadam: Implemented. The function is used in Grype in two places. If this mergerd we should open issue in Grype to make use of this function instead its own.

chovanecadam:
Once this merges, I will create MR in Grype to make use of [syft/pkg/cataloger/common/cpe/target_software_to_pkg_type.go](https://github.com/anchore/syft/pull/4207/files/7b9df152d8e6de4c87040857d7b84635f4c134d6#diff-817d6b5ecc347485ffcb663a1d8bdfa3ceeeb2a47a9c03cbd975eaad71049dd7) instead its own and a separate MR that enables Grype to scan a list of CPEs.

Issue #2709, opened 2024-03-13
pdm support


RichardDally:
**What would you like to be added**:
pdm.lock compatibility to scan requirements.

**Why is this needed**:
pdm is pretty popular according to number of stars (~6K).

tgerla:
Thank you for the suggestion, @RichardDally, I presume you are talking about https://pdm-project.org/latest/ -- the Python package manager, but please correct me if I'm wrong.

RichardDally:
> Thank you for the suggestion, @RichardDally, I presume you are talking about https://pdm-project.org/latest/ -- the Python package manager, but please correct me if I'm wrong.

Yes 😊

tgerla:
Thanks! We'll put this in our backlog. Is this something you're interested in working on? We would be happy to help get you pointed in the right direction. Just let us know.

Dev note: an example pbm lock file: https://github.com/pdm-project/pdm/blob/main/pdm.lock

RichardDally:
Not necessary anymore.

paulslaby:
Hi folks, I need PDM support and I am willing to cooperate.

kzantow:
Hi @paulslaby, catalogers are relatively small, self-contained and simple to implement. There are [lots of examples](https://github.com/anchore/syft/tree/main/syft/pkg/cataloger). We would love to have support!

I'm going to reopen this issue just to make it clear we would accept a contribution for support.

PR #4234, opened 2025-09-23
Add PDM parser

paulslaby:
# Description

Adds parser for [PDM dependency manager](https://pdm-project.org/en/latest/) lockfile.

- Fixes https://github.com/anchore/syft/issues/2709

## Type of change

# Checklist:

I'll be honest, neither Go or Python is my primary language, but I am open to helping you. Please help me polish this PR so that it can eventually get merged into syft codebase.


syft/pkg/cataloger/python/cataloger.go
```diff
@@ -28,7 +28,8 @@ func NewPackageCataloger(cfg CatalogerConfig) pkg.Cataloger {
 		WithParserByGlobs(parsePoetryLock, "**/poetry.lock").
 		WithParserByGlobs(parsePipfileLock, "**/Pipfile.lock").
 		WithParserByGlobs(parseSetup, "**/setup.py").
-		WithParserByGlobs(parseUvLock, "**/uv.lock")
+		WithParserByGlobs(parseUvLock, "**/uv.lock").
+		WithParserByGlobs(parsePipfileLock, "**/pdm.lock").
```
kzantow: This looks like the wrong parser, it should be `parsePdmLock`

kzantow: Also, generally speaking, I'm not sure that these should all be _the same_ cataloger. This prevents disabling of a single one or enabling just one directly. I don't want to ask you to refactor this, but I think PDM should be a separate cataloger rather than getting added to this existing one. You will then give it `python` tags in the task definition, in addition to `pdm`, etc.

paulslaby:
I see locally some failed unit tests, but it looks irrelevant to my changes. 
 One example: 
 ```
 #0 building with "orbstack" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 321B done
#1 DONE 0.0s

#2 [internal] load metadata for docker.io/library/ibmjava:8-jre@sha256:3588cd1cc9b8646fe03b3b15210e69b1b520f1321f8518b69c0e7013d702fd23
#2 ERROR: no match for platform in manifest: not found
------
 > [internal] load metadata for docker.io/library/ibmjava:8-jre@sha256:3588cd1cc9b8646fe03b3b15210e69b1b520f1321f8518b69c0e7013d702fd23:
------
Dockerfile:1
--------------------
   1 | >>> FROM ibmjava:8-jre@sha256:3588cd1cc9b8646fe03b3b15210e69b1b520f1321f8518b69c0e7013d702fd23 AS builder
   2 |     
   3 |     FROM scratch
--------------------
ERROR: failed to solve: ibmjava:8-jre@sha256:3588cd1cc9b8646fe03b3b15210e69b1b520f1321f8518b69c0e7013d702fd23: failed to resolve source metadata for docker.io/library/ibmjava:8-jre@sha256:3588cd1cc9b8646fe03b3b15210e69b1b520f1321f8518b69c0e7013d702fd23: no match for platform in manifest: not found
--- FAIL: Test_JavaBinaryImage (2.15s)
    --- FAIL: Test_JavaBinaryImage/image-java-zulu-8 (0.68s)
        image_fixtures.go:241: Build docker image: name="stereoscope-fixture-image-java-zulu-8" tag="39fda6a633d8f50e1f4c3a145eb5305e5abca841d6d1763fd28c1e551d709a20"
        image_fixtures.go:250: 
                Error Trace:    /Users/pavel/go/pkg/mod/github.com/anchore/stereoscope@v0.1.10/pkg/imagetest/image_fixtures.go:250
                                                        /Users/pavel/go/pkg/mod/github.com/anchore/stereoscope@v0.1.10/pkg/imagetest/image_fixtures.go:185
                                                        /Users/pavel/go/pkg/mod/github.com/anchore/stereoscope@v0.1.10/pkg/imagetest/image_fixtures.go:210
                                                        /Users/pavel/go/pkg/mod/github.com/anchore/stereoscope@v0.1.10/pkg/imagetest/image_fixtures.go:35
                                                        /Users/pavel/go/pkg/mod/github.com/anchore/stereoscope@v0.1.10/pkg/imagetest/image_fixtures.go:64
                                                        /Users/pavel/Work/syft/syft/pkg/cataloger/internal/pkgtest/test_generic_parser.go:191
                                                        /Users/pavel/Work/syft/syft/pkg/cataloger/binary/classifiers_java_test.go:61
                Error:          Received unexpected error:
                                exit status 1
                Test:           Test_JavaBinaryImage/image-java-zulu-8
                Messages:       could not build docker image (shell out)
```

It's hard to notice what is relevant :-/

**Comment by kzantow** [comment]
@paulslaby you can safely ignore that test failure, as you noted I really don't think it is related to your changes, it seems like something I've probably caused and I don't think will be happening in CI. I'll try to get it sorted, but it shouldn't be a blocker for this PR.

`TestAllNames` is a legitimate failure, though, I think that's the only test issue in CI (and static analysis)

**Comment by paulslaby** [comment]
@kzantow I see one more issue  i think it should be handled by me: 
```
--- FAIL: Test_OriginatorSupplier (0.00s)
    originator_supplier_test.go:13: metadata type PythonPdmLockEntry is not covered by a test
FAIL
coverage: 92.8% of statements
FAIL    github.com/anchore/syft/syft/format/internal/spdxutil/helpers   0.333s
```
but what coverage does it expect?

kzantow:
@paulslaby have a look at this test, you just need to add your newly created metadata: https://github.com/anchore/syft/blob/e1483e02857f9536d7cb8715e5be6f84647b57e7/syft/format/internal/spdxutil/helpers/originator_supplier_test.go#L12

kzantow:
@paulslaby it looks like there are some failing tests that are not related to your changes. I will have a look at this later today. Do you mind if I push to your branch?

paulslaby:
> @paulslaby it looks like there are some failing tests that are not related to your changes. I will have a look at this later today. Do you mind if I push to your branch?

It's up to you :) wanna me rebase or something?

kzantow:
Hey @paulslaby I think I fixed up the test issue and the static analysis problem (schema file wasn't matching up). I had a couple notes:

First, what is the purpose of the `Index` field? It is being set to a constant and I don't see anything in the sample file that looks like it would populate this, should it get removed? It is always easier to _add things_ later rather than remove them.

Second, it looks like files on the filesystem are referenced in the PDM file. There is a [FileOwner](https://github.com/anchore/syft/blob/605a275dd3e0645edc9e21659032d8bec42b1f29/syft/pkg/file_owner.go#L7) interface that metadata can implement to return these files and indicate that the particular package "owns" them, in other words, there's a claim that the package manager is provided them. Various things happen with this information, but one is that vulnerability results can be better if, for example, one was a binary that was also surfaced by a different cataloger... the PDM package "owning" it can convey this. I think it might be better to include file references with hash values instead of just the hashes, this way it could also capture the FileOwner information and the hashes would be more useful.

Last, it looks like the PDM files have _dependency information_ in them. The catalogers return relationships with this information, users generally want this information to be surfaced, would you want to add this? Here's [an example using utility functions](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/alpine/cataloger.go#L16).

paulslaby:
> First, what is the purpose of the `Index` field? It is being set to a constant and I don't see anything in the sample file that looks like it would populate this, should it get removed?

In the beginning I was thinking that this might be the way how syft manages [URL dependencies](https://pdm-project.org/latest/usage/dependency/#url-dependencies). Now I am not sure, if we should parse package name from URL. Do you have a pattern for that already established? 

I will have a look iat the rest. Thank you for your feedback.

paulslaby:
@kzantow I implemented relationship parsing. but those can be conditional [based on python version (or platform)](https://github.com/anchore/syft/blob/58f278cc3bf90108fe0c813c18fdebf348c2c70e/syft/pkg/cataloger/python/test-fixtures/pdm-lock/pdm.lock#L86). You probably know, python lockfiles are not for one python version, but for [constrained python version](https://github.com/anchore/syft/blob/58f278cc3bf90108fe0c813c18fdebf348c2c70e/syft/pkg/cataloger/python/test-fixtures/pdm-lock/pdm.lock#L11). Those conditions are currently ignored. Do you have a different opinion?

kzantow:
@paulslaby generally speaking, the philosophy of syft is to surface as much information as we find. In the case of conditional things, if there is a possibility that a dependency could be included, we generally would include this by default. For example: NPM currently includes devDependencies and Java currently includes test dependencies. I think the correct default behavior is to include the dependencies, regardless of python version constraints. I don't think this necessarily needs to be configurable as part of this PR, but we could an option to exclude things at a later time. Am I understanding the question correctly?

paulslaby:
@kzantow Yes, you understand it correctly. That's the way it working now.

kzantow:
Hey @paulslaby, I pushed a few changes to:
* populate the metadata dependencies
* correct the direction of the dependency relationship
* avoid duplication of name & version in metadata
* other misc. cleanup

Would you be able to give this a try to make sure it's still giving you the results you are expecting?

paulslaby:
@kzantow  Yes, that works. thank you for finishing it 🙇

Issue #4179, opened 2025-09-01
Add Windows ARM64 releases


compnerd:
**What would you like to be added**:

It would be really nice to have Windows ARM64 binaries in the releases so that this tool can be natively used on that architecture.

**Why is this needed**:

Windows ARM64 is gaining adoption finally, it would be useful to be able to build natively and generate a SBoM using this tool on that platform.

**Additional context**:

popey:
Hi @compnerd, what an interesting idea! 
I don't know if any of us have ARM64 machines to test this on.
Have you tried building and running syft from source on such a device? 

If you have some time, the following should work:

1. Install go as per their [docs](https://go.dev/doc/install) - probably this: https://go.dev/dl/go1.25.0.windows-arm64.msi
2. `git clone https://github.com/anchore/syft`
3. `cd syft`
4. `make bootstrap`

The above step may fail - as I don't know if we have all the dependencies built for windows/arm64.

5. `make build`

Let us know if it builds and works.

compnerd:
Testing would be easier, but building might be more of a challenge. Most of the work that I'm doing on ARM64 currently is more on a consumer grade rather than developer grade machine. Note that Windows ARM64 runners are now available in GitHub. If I manage to find some time, I can try to build it as well, but I expect that I wouldn't be as effective at this as my Go skills aren't as strong as C/C++.

kzantow:
This should be relatively easy to add -- we just need to update the goreleaser configuration here: https://github.com/anchore/syft/blob/main/.goreleaser.yaml#L45

compnerd:
The GNU make dependency (and its linuxisms) make this difficult to build locally.

popey:
Just as a quick shake, I tried to build and run syft, and.. well, it runs.

* Installed UTM on my Mac: https://mac.getutm.app/
* Installed Windows 11 in it: https://docs.getutm.app/guides/windows/
* Installed "Guest additions" for UTM: 
* Installed Chocolatey: https://chocolatey.org/install
* `choco install make`
* `choco install git`
* Install bash via git-bash: https://git-scm.com/downloads/win
* Open git-bash
* `git clone https://github.com/anchore/syft`
* `cd syft`
* `go build ./cmd/syft`

It works 🤷 

<img width="1915" height="1018" alt="Image" src="https://github.com/user-attachments/assets/275ad486-60f3-4664-b247-8caac8537b0b" />

compnerd:
[...]

PR #4237, opened 2025-09-24
goreleaser: add ARM64 GOARCH for Windows

compnerd:
Update the architectures to support Windows AMD64 and ARM64 releases.

Fixes: #4179
Signed-off-by: Saleem Abdulrasool <compnerd@compnerd.org>

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -42,7 +42,7 @@ builds:
     dir: ./cmd/syft
     binary: syft
     goos: [windows]
-    goarch: [amd64]
+    goarch: [amd64, arm64]
     mod_timestamp: *build-timestamp
     ldflags: *build-ldflags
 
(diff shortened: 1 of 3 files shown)


kzantow:
I should have responded here about the test failures, I think this file needs to be updated to account for the new windows assets: https://github.com/anchore/syft/blob/323fd3e34cd3febc61e5a38b2f83187f18fdab5f/test/install/1_download_snapshot_asset_test.sh#L105-L106

And possibly here: https://github.com/anchore/syft/blob/323fd3e34cd3febc61e5a38b2f83187f18fdab5f/test/install/3_install_asset_test.sh#L101

Apologies, I have not dealt with these tests before, so there may be more pieces to update

compnerd:
Thanks! Looking forward to the next release!

PR #4252, opened 2025-10-02
Improve struct and field comments and incorporate into json schema

wagoodman:
This adds several more field and struct comments for the `pkg` and `file` structs represented in the JSON output. Additionally, this wires up the json schema generation code to add these comments to the as descriptions for each field and type definition.

The existing `16.0.40` schema has been regenerated with these enhancements.

## Type of change

# Checklist:

### DIFF
--- a/syft/internal/jsonschema/comments.go
+++ b/syft/internal/jsonschema/comments.go
@@ -1,0 +1,159 @@
+package main
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/invopop/jsonschema"
+)
+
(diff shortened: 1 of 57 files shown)


syft/internal/jsonschema/comments.go
```diff
@@ -0,0 +1,159 @@
...
+			// store the alias mapping: aliasName -> sourceName
+			aliases[typeSpec.Name.Name] = ident.Name
+			return true
+		})
+
+		return nil
+	})
+
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "warning: failed to find type aliases: %v\n", err)
```
spiffcs: not even nit worthy, but if I'm printing to stderr and then a panic the following might be better as`:%s/warning/error`

PR #4255, opened 2025-10-04
feat: enhance setup.py parser to handle unquoted dependencies

HalaAli198:
# Description
Enhance `syft/pkg/cataloger/python/parse_setup.go` to correctly parse unquoted Python dependencies defined in `setup.py` files, addressing a common pattern where dependencies are listed in multiline strings followed by .split().

## Why is this needed?
Many Python projects use the following pattern in their `setup.py` files:
```install_requires = 
"""
django==4.2.23
CairoSVG==2.7.1
Pillow==11.0.0
""".split()
```
Currently, Syft’s parser only recognizes dependencies wrapped in quotes (e.g., "django==4.2.23").
As a result, all dependencies written in this unquoted .split() form are missed, leading to incomplete SBOMs.

## Fix
- Added a secondary regex pattern `^\s*(\w+)\s*==\s*([\w\.\-]+)` to detect unquoted dependencies.
- Ensured backward compatibility with existing quoted parsing.
- Prevented duplicate detection.
- Preserved all existing functionality.

## Testing
 To verify:
1. Download mayan-edms (which uses this pattern): `pip download --no-deps --no-binary :all: mayan-edms==4.9.2`
2. Extract the source and inspect setup.py — it defines dependencies via """.split().
3. Run Syft before and after applying the fix: `syft dir:"/path/to/project" --select-catalogers python --output cyclonedx-json=output.json`

- Before: SBOM is empty.
- After: SBOM includes all pinned dependencies from install_requires.

# Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/python/parse_setup.go
+++ b/syft/pkg/cataloger/python/parse_setup.go
@@ -33,66 +33,87 @@ func parseSetup(_ context.Context, _ file.Resolver, _ *generic.Environment, read
 		line := scanner.Text()
 		line = strings.TrimRight(line, "\n")
 
-		for _, match := range pinnedDependency.FindAllString(line, -1) {
-			parts := strings.Split(match, "==")
-			if len(parts) != 2 {
-				continue
-			}
-			name := strings.Trim(parts[0], "'\"")
-			name = strings.TrimSpace(name)
-			name = strings.Trim(name, "'\"")
-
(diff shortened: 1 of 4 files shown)


spiffcs:
👋 Thanks for the PR @HalaAli198 - I can add some tests here and get it over the line

HalaAli198:
@spiffcs Thank you! Appreciate the help with this!

Issue #3927, opened 2025-05-21
PNPM latest lockfile (version 9.0)


spiffcs:
**What would you like to be added**:
The PNPM folks are pretty prolific and we'd like to support more versions of decoding dependencies from their lockfiles. The latest version v9.0 looks like this:
```
lockfileVersion: '9.0'
settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false
importers:
  .:
    dependencies:
      react:
        specifier: ^19.1.0
        version: 19.1.0
      react-dom:
        specifier: ^19.1.0
        version: 19.1.0(react@19.1.0)
    devDependencies:
      jest:
        specifier: ^29.7.0
        version: 29.7.0(@types/node@22.15.21)
packages:
  '@ampproject/remapping@2.3.0':
    resolution: {integrity: sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==}
    engines: {node: '>=6.0.0'}
  '@babel/code-frame@7.27.1':
    resolution: {integrity: sha512-cjQ7ZlQ0Mv3b47hABuTevyTuYN4i+loJKGeV9flcCgIK37cCXRh+L1bd3iBHlynerhQ7BhCkn2BPbQUL+rGqFg==}
    engines: {node: '>=6.9.0'}
  '@babel/compat-data@7.27.2':
    resolution: {integrity: sha512-TUtMJYRPyUb/9aU8f3K0mjmjf6M9N5Woshn2CS6nqJSeJtTtQcpLUXjGt9vbF8ZGff0El99sWkLgzwW3VXnxZQ==}
    engines: {node: '>=6.9.0'}
  '@babel/core@7.27.1':
    resolution: {integrity: sha512-IaaGWsQqfsQWVLqMn9OB92MNN7zukfVA4s7KKAI0KfrrDsZ0yhi5uV4baBuLuN7n3vsZpwP8asPPcVwApxvjBQ==}
    engines: {node: '>=6.9.0'}
  '@babel/generator@7.27.1':
    resolution: {integrity: sha512-UnJfnIpc/+JO0/+KRVQNGU+y5taA5vCbwN8+azkX6beii/ZF+enZJSOKo11ZSzGJjlNfJHfQtmQT8H+9TXPG2w==}
    engines: {node: '>=6.9.0'}
```

[...]

PR #4256, opened 2025-10-04
Parse pnpm v9 lockfiles

bernardoamc:
# Description

Allow pnpm v9 lockfiles to be parsed by Syft. This version has a different structure compared to previous lockfiles, nesting fields like `dependencies` and `devDependencies` within the concept of an [importer](https://github.com/orgs/pnpm/discussions/6214). In order to keep backwards compatibility the code has been refactored to:

1. Create a shared interface for parsing pnpm files while attempting to keep most of the functionality consolidated across common functions
2. We have also optimize our logic to check for duplicated packages by utilizing a map rather than repeatedly iterating through a slice.

Fixes: https://github.com/anchore/syft/issues/3927

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


bernardoamc:
We are only parsing the `packages` key while leaving the option to parse the remaining fields. I think this covers the current functionality since `packages` should contain every package listed within every other field.

bernardoamc:
Adjusted the code so `newPnpmLockfileParser` doesn't get exported.

spiffcs (approved):

bernardoamc:
For some reason my local environment did no catch these exported warnings. I've also made the interface private it.

bernardoamc:
Thanks for the review @spiffcs, I appreciate it! Hopefully this build will be a green one. 😄

bernardoamc:
Glad the export-check was the only blocker. 😄

PR #4269, opened 2025-10-09
feat: SPDX 3

### LINKED ISSUE #4250: SPDX 3 Support
With the recent announcement [Anchore Enterprise is now SPDX 3 Ready](https://anchore.com/blog/anchore-enterprise-is-now-spdx-3-ready/), support for [SPDX 3.0.1](https://spdx.github.io/spdx-spec/v3.0.1/) would be beneficial to have in syft.

**Why is this needed**:
With SPDX 3.0.1 generally available now, it would be nice to have support through syft as well.
The [German guidance for Cyber Resilience](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TR03183/BSI-TR-03183-2_v2_1_0.pdf) also requires SPDX 3.0.1 or CycloneDX 1.6: 

**Additional context**:

https://github.com/anchore/syft/issues/1683#issuecomment-2701385993

kzantow:
# Description

SPDX 3 support:
* reading SPDX 3 uses a custom decoder, implemented here, as the tools-golang does not have a backwards conversion to 2.3 at this time
* writing SPDX 3 just uses the tools-golang conversion from our existing 2.3 populated model

- Fixes #4250

## Type of change

# Checklist:


syft/format/common/spdxhelpers/to_syft_model_v3.go
```diff
@@ -0,0 +1,736 @@
...
+
+func ptrTo(k any) (reflect.Value, error) {
+	rv := reflect.ValueOf(k)
+	if rv.Kind() != reflect.Pointer {
+		return rv, fmt.Errorf("value is not a pointer: %#v", k)
+	}
+	return rv, nil
+}
+
+func panicIfErr(e error) {
```
wagoodman: why should we be panicing in this code over propagating errors?

syft/format/common/spdxhelpers/to_syft_model_v3.go
```diff
@@ -0,0 +1,736 @@
...
+
+	return nil
+}
+
+func v3collectSyftPackages(s *sbom.SBOM, spdxMap ptrMap[any], relationships ptrMap[[]spdx.AnyRelationship], doc *spdx.Document) {
+	skipIDs := v3packageIDsToSkip(doc)
+	found := ptrMap[struct{}]{}
+	for _, elementList := range []spdx.ElementList{doc.Elements, doc.RootElements} {
+		for _, p := range elementList.Packages() {
+			if p == nil || skipIDs.Has(p.GetID()) || found.Has(p.GetID()) {
```
wagoodman: I think this might not be working as expected; `skipIDs.Has` can find the poitned in memory, however, we're passing a string from `p.GetID()` so it's going to always silently fail.

```golang
	m := ptrMap[struct{}]{}

	p := &spdx.Package{
		ID:   "test-id",
		Name: "test-pkg",
	}

	require.NoError(t, m.Set(p, struct{}{})) // this is what v3packageIDsToSkip does
	assert.True(t, m.Has(p.GetID()), "woops") // fails!
```

syft/format/common/spdxhelpers/to_syft_model_v3.go
```diff
@@ -0,0 +1,736 @@
...
+				continue
+			}
+			panicIfErr(found.Set(p, struct{}{}))
```
wagoodman: I think something similar is happening here, but I didn't trace it exactly

syft/format/common/spdxhelpers/to_syft_model_v3.go
```diff
@@ -0,0 +1,736 @@
...
+			UserInput:      container,
+			ID:             p.GetID(),
+			Layers:         nil, // TODO handle formats with nested layer packages like Tern and K8s BOM tool
+			ManifestDigest: digest,
+		},
+	}
+}
+
+func v3fileSource(p spdx.AnyPackage) source.Description {
+	typeRegex := regexp.MustCompile("DocumentRoot-([^-]+)-.*$")
```
wagoodman, line 173:
```suggestion
	typeRegex := regexp.MustCompile("^DocumentRoot-([^-]+)-.*$")
```
I think without this you'll still match on stuff like `https://ns/DocumentRoot-Directory-myapp` (I realize a little contrived, but still possible)

syft/format/common/spdxhelpers/to_syft_model_v3.go
```diff
@@ -0,0 +1,736 @@
...
+
+	digest := ""
+	if len(p.GetVerifiedUsing().Hashes()) > 0 {
+		h := p.GetVerifiedUsing().Hashes()[0]
+		if h != nil {
+			digest = fmt.Sprintf("%s:%s", v3fromChecksumAlgorithm(h.GetAlgorithm()), h.GetValue())
+		}
+	}
+
+	return source.Description{
```
wagoodman, line 158: I think this is missing assigning supplier like what was done https://github.com/anchore/syft/pull/4269/changes#diff-1e5052f452c5389250d68459a0427ad595986d89a65f0c188e2458f2d7e1893fR188-R197 (I might be missing something)

syft/format/spdxjson/spdx3_decoder.go
```diff
@@ -0,0 +1,58 @@
...
+	spdxVersion := ""
+
+	switch {
+	case doc.Context == "":
+	case spdxContextRegex(spdxutil.V3_0).MatchString(doc.Context):
+		spdxVersion = spdxutil.V3_0
+	default:
+	}
+
+	return spdxutil.JSONFormatID, spdxVersion
```
wagoodman: I think we're still returning a claim that the formatID is JSON even if we pass no content into Context

syft/format/spdxjson/spdx3_decoder.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func spdxContextRegex(minorVersion string) *regexp.Regexp {
+	return regexp.MustCompile(regexp.QuoteMeta("https://spdx.org/rdf/") + minorVersion + `\.\d+` + regexp.QuoteMeta("/spdx-context.jsonld"))
```
wagoodman: this MustCompile call (and others) should probably be global vars instead of being invoked on each call

cmd/syft/internal/options/format_spdx_json.go
```diff
@@ -17,8 +17,9 @@ func (o FormatSPDXJSON) config(v string) spdxjson.EncoderConfig {
...
 		pretty = *o.Pretty
 	}
-	return spdxjson.EncoderConfig{
-		Version: v,
-		Pretty:  pretty,
-	}
+	c := spdxjson.DefaultEncoderConfig()
+	c.Version = v
+	c.Pretty = pretty
+	c.DefaultVersion = "2.3"
```
wagoodman:
```suggestion
	c.DefaultVersion = spdxutil.DefaultVersion
```

syft/format/common/spdxhelpers/to_syft_model_v3.go
```diff
@@ -0,0 +1,736 @@
...
+			}
+		}
+	}
+	return out
+}
+
+func v3toSyftCoordinates(f spdx.AnyFile) file.Coordinates {
+	const layerIDPrefix = "layerID: "
+	var fileSystemID string
+	if strings.Index(f.GetComment(), layerIDPrefix) == 0 {
```
wagoodman: There are a lot of `strings.Index(...) == 0` calls that would be more readalbe if it was `strings.HasPrefix()`

cmd/syft/internal/options/format_spdx_json.go
kzantow: This is not exported to the cmd package; it looks like this isn't necessary to be set here, anyway.

syft/format/common/spdxhelpers/to_syft_model_v3.go
kzantow: This was definitely a bug, unsure how the tests were passing 🤷

syft/format/common/spdxhelpers/to_syft_model_v3.go
kzantow: I've updated this to be more clear. There was a problem that was fixed, you noted. Getting non-pointers is a programming problem. To keep this simple, I've left the panic but in one spot with a comment about why

syft/format/common/spdxhelpers/to_syft_model_v3.go
kzantow: This actually should match within a larger string, document IDs are prefixed with something, e.g. `SPDXRef:` or `https://some-url/`

syft/format/common/spdxhelpers/to_syft_model_v3.go
kzantow: Odd, seems like this was the only spot I missed; good catch

wagoodman (approved):

wagoodman (approved):

gvillenave:
Thanks for implementing this! Is it ready to merge?

wagoodman (approved):

willmurphyscode (approved):

PR #4278, opened 2025-10-14
support universal (fat) mach-o binary files

JoeyShapiro:
# Description

This change adds support for Mach-o Universal (Fat) binaries. This was a `TODO` in the code that I ran across while using the package. I have made a few changes that handle fat files differently and I am not sure if this is the best way, so I wanted someone to check and see if they wanted a different way.
- Using the reader, I check the magic of the file to determine if it is fat or not
- I create a unique list of all libraries used over all of the arches
  - I could also add text to the library string to say which arch it is from.
-  I set the flag to true if any of the arches have `HasExports` or `HasEntrypoint`.
  - Rather than making sure all of them have it. There is not a standard way of handling this for fat binaries

## Type of change

### DIFF
--- a/syft/file/cataloger/executable/macho.go
+++ b/syft/file/cataloger/executable/macho.go
@@ -2,8 +2,6 @@ package executable
 
 import (
 	"debug/macho"
-	"encoding/binary"
-	"errors"
 
 	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/syft/file"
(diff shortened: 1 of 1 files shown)


JoeyShapiro:
the static analysis says to rewrite my if-else to a switch. but I can't see how it can be rewritten as a switch statement.

kzantow:
@JoeyShapiro you can make a switch statement like:
```
switch {
case <any boolean condition>:
case <another boolean condition>:
}
```

JoeyShapiro:
ah I see. how is this? I didn't know a switch statement could do that.

kzantow:
Golang `switch` specialty!

We have some handling of universal binaries already, the [UnionReader](https://github.com/anchore/syft/blob/07029ead8a2c6ba727757e494b516ea740b210e2/syft/internal/unionreader/union_reader.go#L28), for example. Maybe it would be simpler to use that?

JoeyShapiro:
ok, how does this look. it seems quite a bit cleaner. but I'm wondering if it isn't as obvious that it is working on universal binaries. If that even matters

kzantow:
It might be good to have some indication about the file being a universal binary, but it seems to me that returning information about all the libraries used, existence of entrypoint, etc. is more important, where we may have been missing some of this information before. We can always add the universal binary indicator later.

kzantow (approved):
LGTM, thanks @JoeyShapiro 👍

JoeyShapiro:
ok cool. thanks for the help

PR #4281, opened 2025-10-14
Add docs to API-level configs

wagoodman:
Primarily, this PR makes the following changes:
- adds missing API-level docs to cataloger configurations
- adds `# app-config: <field>` to API-level config fields to map which fields affect application-level configuration. In the near future we can use these in documentation to hint at how the catalogers can be configured.

This PR is also setting up for future 
- migrates cataloger globs and names to be literals passed to generic cataloger constructors. This will help with detecting capabilities in upcoming PRs.
- When making the above change it was found that the `pkgtest` helpers were making assumptions about the `FoundBy` field that were incorrect. When comparing both packages and relationships, you cannot depend on comparing this field if the parser function leans on the generic cataloger capabilities to fill this value in (as they should). For this reason the FoundBy field on a package is now ignored when comparing relationships (but not packages).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,8 @@ bin/
 /snapshot
 /.tool
 /.task
+/generate
+/specs
 
 # changelog generation
 CHANGELOG.md
--- a/internal/packagemetadata/generate/main.go
+++ b/internal/packagemetadata/generate/main.go
@@ -5,8 +5,9 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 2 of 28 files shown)


wagoodman:
I do agree with your points about the awkwardness here, but I currently don't have another better spot to put this information without similar caveats (after chatting through the details offline). I think I'll continue forth with the PR as-is since there is no impact on the movability of this information later (we can always change this later since there is no impact on the API stability or anything like that).

PR #4283, opened 2025-10-14
Extract zip archive with multiple entries

Rupikz:
# Description

Overwrite existing files during unpacking zip archive

## Type of change

# Checklist:

### DIFF
--- a/syft/source/filesource/file_source.go
+++ b/syft/source/filesource/file_source.go
@@ -234,10 +234,11 @@ func fileAnalysisPath(path string, skipExtractArchive bool) (string, func() erro
 		// when tar/zip files are extracted, if there are multiple entries at the same
 		// location, the last entry wins
 		// NOTE: this currently does not display any messages if an overwrite happens
-		if tar, tarOk := unarchiver.(*archiver.Tar); tarOk {
-			tar.OverwriteExisting = true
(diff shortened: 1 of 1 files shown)


syft/source/filesource/file_source.go
```diff
@@ -231,11 +231,13 @@ func fileAnalysisPath(path string, skipExtractArchive bool) (string, func() erro
...
 	envelopedUnarchiver, err := archiver.ByExtension(path)
 	if unarchiver, ok := envelopedUnarchiver.(archiver.Unarchiver); err == nil && ok {
-		if tar, ok := unarchiver.(*archiver.Tar); ok {
-			// when tar files are extracted, if there are multiple entries at the same
-			// location, the last entry wins
-			// NOTE: this currently does not display any messages if an overwrite happens
+		// when tar/zip files are extracted, if there are multiple entries at the same
+		// location, the last entry wins
+		// NOTE: this currently does not display any messages if an overwrite happens
+		if tar, tarOk := unarchiver.(*archiver.Tar); tarOk {
```
wagoodman: nit: now that there are two cases this would be better as a type assertion switch case (non-blocking)

Rupikz: Done

Issue #4271, opened 2025-10-10
syft can't extract go module information from executables on Windows


flichtenheld:
**What happened**:

Trying to scan executables on Windows that were built with go but it shows no information. Scanning the same file on Linux works fine.

**What you expected to happen**:

Output on Linux and Windows should be identical.

**Steps to reproduce the issue**:

Have syft scan itself:

Linux:

```
syft file:syft.exe
 ✔ Indexed file system                                                                    syft.exe 
 ✔ Cataloged contents              3212796a93710a736a66bacf8b6fe45a0590ff669dc0ecbeb546ad0d64929560
   ├── ✔ Packages                        [264 packages]  
   ├── ✔ Executables                     [1 executables]  
   ├── ✔ File digests                    [1 files]  
   └── ✔ File metadata                   [1 locations]  
NAME                                                                         VERSION                                TYPE         
cloud.google.com/go                                                          v0.116.0                               go-module    
cloud.google.com/go/auth                                                     v0.9.9                                 go-module    
cloud.google.com/go/auth/oauth2adapt                                         v0.2.4                                 go-module    
cloud.google.com/go/compute/metadata                                         v0.7.0                                 go-module    
cloud.google.com/go/iam                                                      v1.2.2                                 go-module    
[...]
```

Windows:

```
syft.exe file:syft.exe
 ✔ Indexed file system                                                                    syft.exe   
 ✔ Cataloged contents              3212796a93710a736a66bacf8b6fe45a0590ff669dc0ecbeb546ad0d64929560  
   ├── ✔ Packages                        [1 packages]
   ├── ✔ Executables                     [0 executables]
   ├── ✔ File metadata                   [0 locations]
   └── ✔ File digests                    [0 files]
NAME   VERSION  TYPE
\syft  UNKNOWN  binary
```

**Anything else we need to know?**:

Tried on different Windows and Linux systems, but behavior was identical on all of them.

**Environment**:
- Output of `syft version`:

```
Application:   syft
Version:       1.33.0
BuildDate:     2025-09-15T20:38:16Z
GitCommit:     b87b9191497c2d410b42c05591347d428129fd2a
GitDescription: v1.33.0
Platform:      windows/amd64
GoVersion:     go1.24.7
Compiler:      gc
SchemaVersion: 16.0.39
```

- OS (e.g: `cat /etc/os-release` or similar):

Windows 11, Windows Server 2022

kzantow:
[...]

PR #4285, opened 2025-10-15
convert posix path back to windows

JoeyShapiro:
# Description

This is a fix to allow windows to properly open and get the MIME type of a file during an FS scan. This was a PR to [stereoscope](https://github.com/anchore/stereoscope/pull/464), but it was decided to move it here. It also moves in a function and test from there that is only used for syft.

This allows this function to properly use the `windows` package to check and convert back to a windows path, so the file can be opened.

The function moved in still calls to stereoscope, because it requires structs and functions that are heavily integrated into that package.
- Fixes https://github.com/anchore/stereoscope/pull/464

## Type of change

### DIFF
--- a/syft/internal/fileresolver/metadata.go
+++ b/syft/internal/fileresolver/metadata.go
@@ -4,7 +4,7 @@ import (
 	"os"
 
 	"github.com/anchore/stereoscope/pkg/file"
-	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/syft/internal/windows"
 )
 
@@ -26,11 +26,7 @@ func NewMetadataFromPath(path string, info os.FileInfo) file.Metadata {
 			// TODO: it may be that the file is inaccessible, however, this is not an error or a warning. In the future we need to track these as known-unknowns
 			f = nil
 		} else {
-			defer func() {
-				if err := f.Close(); err != nil {
-					log.Warnf("unable to close file while obtaining metadata: %s", path)
-				}
-			}()
+			defer internal.CloseAndLogError(f, usablePath)
 		}
 
 		mimeType = file.MIMEType(f)


kzantow:
Thanks for the update here, @JoeyShapiro. It looks like your commits just need to have the DCO `Sign-Off`

JoeyShapiro:
ok. I did a sign off. and I also made get_xid cross platform

syft/internal/fileresolver/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+		if windows.HostRunningOnWindows() {
+			usablePath = windows.FromPosix(usablePath)
+		}
+
+		f, err := os.Open(usablePath)
+		if err != nil {
+			// TODO: it may be that the file is inaccessible, however, this is not an error or a warning. In the future we need to track these as known-unknowns
+			f = nil
+		} else {
+			defer func() {
```
kzantow: There's [a handy utility](https://github.com/anchore/syft/blob/065ac13ab726f00e2f32a243c23a598d820ae25b/internal/err_helper.go#L14):  that would make this `defer internal.CloseAndLogError(f, path)`

syft/internal/fileresolver/metadata.go
```diff
@@ -0,0 +1,48 @@
...
+	if ty == file.TypeRegular {
+		usablePath := path
+		// denormalize the path back to windows so we can open the file
+		if windows.HostRunningOnWindows() {
+			usablePath = windows.FromPosix(usablePath)
+		}
+
+		f, err := os.Open(usablePath)
+		if err != nil {
+			// TODO: it may be that the file is inaccessible, however, this is not an error or a warning. In the future we need to track these as known-unknowns
```
kzantow, line 26: Actually, this would be good to do as part of this PR, since we're introducing a function that we have full control over, e.g. return this error and have this returned as an "unknown" error, which can be reported in the SBOM. I can have a look at making this change, if you like

JoeyShapiro:
I did the close and log part, but im not sure how to make an unknown error. I am fine doing it myself if you give me a hint of some kind

kzantow:
I haven't quite worked out all the details, but we will want to collect all the errors, for example in [the](https://github.com/anchore/syft/blob/e923db2a94343ff8ed86b296dbd1e98bea7a955b/syft/internal/fileresolver/directory_indexer.go#L27) [errPaths](https://github.com/anchore/syft/blob/e923db2a94343ff8ed86b296dbd1e98bea7a955b/syft/internal/fileresolver/file_indexer.go#L21) and we will want to get all these path errors added to the [Unknowns](https://github.com/anchore/syft/blob/e923db2a94343ff8ed86b296dbd1e98bea7a955b/syft/sbom/sbom.go#L29) section of the SBOM. This is something many people probably don't know about, but we capture many errors and surface these with the file that the error originated from. Maybe we could add an option and get a reference to these path errors [around here](https://github.com/anchore/syft/blob/main/internal/task/unknowns_tasks.go#L51), it would probably require type asserting the file resolver to a new interface that has a `PathErrs() <something>`, because we shouldn't change the public API in a breaking way. I'm happy for you to make these changes, but it might take a little experimentation & digging to find the right spots for it, which is why I offered to help out.

JoeyShapiro:
I can try it out. I have had some experience with the unknowns. That is how I found this problem in the first place. It might have to wait until later today though, as I am currently at work. Especially if it would require some more work to add in the interfaces

flichtenheld:
I can confirm that this fixes #4271

kzantow:
We can always get this merged as-is and worry about the unknowns later, I didn't mean to add a large ask here. I think I'm going to merge this, and if you do feel like contributing updates to handle unknowns, that would be great but can come as a new PR 👍

kzantow (approved):

JoeyShapiro:
yeah, I didn't get a chance to look at it yesterday. I can always make another PR, or someone else can if they beat me to it. thanks for helping get this pulled in

PR #4286, opened 2025-10-15
feat: Add the ability to fetch remote licenses for pnpm-lock.yaml files

timols:
# Description

Extending https://github.com/anchore/syft/pull/2708 to also be able to retrieve remote licenses from pnpm-lock.yaml.

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/javascript/package.go
+++ b/syft/pkg/cataloger/javascript/package.go
@@ -173,7 +173,6 @@ func newPnpmPackage(ctx context.Context, cfg CatalogerConfig, resolver file.Reso
 		if err != nil {
 			log.Debugf("unable to extract licenses from javascript pnpm-lock.yaml for package %s:%s: %+v", name, version, err)
 		}
-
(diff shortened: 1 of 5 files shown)


kzantow:
This looks great but I don't see that the remote license functionality is well tested. Would it make sense to add a test like [the yarn cataloging](https://github.com/anchore/syft/blob/e923db2a94343ff8ed86b296dbd1e98bea7a955b/syft/pkg/cataloger/javascript/parse_yarn_lock_test.go#L238-L277)? Maybe I've missed some tests, but it would be great to have a little more confidence there, to me.

P.S. it looks like `static-analysis` failures, you could try to use `make lint-fix` to fix them

timols:
Certainly - happy to do it. Will turn this around shortly.

timols:
@kzantow test added!

timols:
@kzantow added this for Python too - https://github.com/anchore/syft/pull/4295 😄

PR #4289, opened 2025-10-15
Remove duplicate image source providers

Rupikz:
# Description

Remove duplicate image source providers if the DefaultImagePullSource parameter is set.

Details:
1) Move getProviders() logic from source config builder
2) Fixes combining providers `base.Join(def...).Join(pull...)` previously "pull" continued to contain "def"

## Type of change

# Checklist:

### DIFF
--- a/syft/get_source_config.go
+++ b/syft/get_source_config.go
@@ -2,7 +2,9 @@ package syft
 
 import (
 	"crypto"
+	"fmt"
 
+	"github.com/anchore/go-collections"
 	"github.com/anchore/stereoscope/pkg/image"
 	"github.com/anchore/syft/syft/source"
 	"github.com/anchore/syft/syft/source/sourceproviders"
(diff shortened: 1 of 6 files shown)


syft/get_source_config.go
```diff
@@ -61,29 +59,6 @@ func (c *GetSourceConfig) WithDefaultImagePullSource(defaultImagePullSource stri
...
-
-	// if the "default image pull source" is set, we move this as the first pull source
-	if c.DefaultImagePullSource != "" {
-		base := providers.Remove(sourceproviders.PullTag)
-		pull := providers.Select(sourceproviders.PullTag)
-		def := pull.Select(c.DefaultImagePullSource)
-		if len(def) == 0 {
-			return nil, fmt.Errorf("invalid DefaultImagePullSource: %s; available values are: %v", c.DefaultImagePullSource, pull.Tags())
-		}
-		providers = base.Join(def...).Join(pull...)
```
kzantow, line 75: This set is supposed to remove duplicates, but I see through some debugging that some of the sources are not comparable, so they are not being removed. An easier change would just be to add this line:
`pull = pull.Remove(c.DefaultImagePullSource)`

kzantow:
Hey @Rupikz, thanks for the contribution. However, `base.Join(def...).Join(pull...)` should prevent duplicates. I made a PR to fix that issue: https://github.com/anchore/go-collections/pull/4

Rupikz:
Hey @kzantow,

Awesome, your solution is better.
Should we close this PR now, or would it be okay to push a small refactor? (move getProviders from source config builder)?

kzantow:
Hey @Rupikz adding tests here is great, I'd love to get those contributed. Does the function need to be refactored? It seems to me there isn't much benefit to moving it. Maybe just revert that part of the change, bump the dependency, and update the tests to use the method instead of refactored function?

PR #4294, opened 2025-10-16
fix: return PanicErrors

kzantow:
# Description

This PR updates the `go-sync` library to return panics as errors during concurrent calls, and adds an option to loudly log, but not fail on errors. This option will allow users to opt in to behavior in their CI environment, for example, that errors including panics will not cause the process to return nonzero, but instead only be _error_ logged.

## Type of change

- [x] Feature

# Checklist:

TODO:
- [ ] configuration to continue in the event of a non-unknown error bubbling up to the top with `--continue-on-error` or similar

### DIFF
--- a/syft/pkg/cataloger/generic/cataloger_test.go
+++ b/syft/pkg/cataloger/generic/cataloger_test.go
@@ -183,9 +183,9 @@ func TestClosesFileOnParserPanic(t *testing.T) {
 		upstreamCataloger: "unit-test-cataloger",
 	}
 
-	assert.PanicsWithValue(t, "panic!", func() {
-		_, _, _ = c.Catalog(ctx, resolver)
-	})
+	_, _, err := c.Catalog(ctx, resolver)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "panic!")
 	require.True(t, spy.closed)
 }
 


internal/unknown/coordinate_error.go
```diff
@@ -184,18 +185,25 @@ func visitErrors(err error, fn func(error) error) error {
 		// errors.Join omits nil errors and will return nil if all passed errors are nil
 		return errors.Join(out...)
 	}
+	// handle PanicErrors directly, since these are wrapped errors
+	if e, ok := err.(sync.PanicError); ok {
```
wagoodman, line 189: nit: why not `errors.As`?

internal/unknown/coordinate_error.go
```diff
@@ -184,18 +185,25 @@ func visitErrors(err error, fn func(error) error) error {
...
 		wrapped := e.Unwrap()
-		got := visitErrors(wrapped, fn)
-		if got == nil {
-			return nil
-		}
-		if wrapped.Error() != got.Error() {
-			prefix := strings.TrimSuffix(err.Error(), wrapped.Error())
-			return fmt.Errorf("%s%w", prefix, got)
+		// if there is not a wrapped error, fall through to visit the error directly
+		if wrapped != nil {
```
wagoodman, line 196: nit / non-blocking: this would be a little more readable with a guard clause instead of wrapping  (e.g. instead do `if wrapped == nil { return nil}` ahead of the remaining logic.)

wagoodman (approved):

internal/unknown/coordinate_error.go
kzantow: This function recursively unwraps errors, it doesn't need to let errors.As also do this

internal/unknown/coordinate_error.go
kzantow: It's a little different behavior than this -- if wrapped is nil, it continues to visit the container

PR #4284, opened 2025-10-14
feat: Add dependency parsing to javascript package locks

timols:
# Description

This PR adds dependency parsing for javascript/npm package-lock.json files (v2 and v3). If this approach suits, then I'll add PRs for yarn, pnpm etc. This tackles part of https://github.com/anchore/syft/issues/572 and https://github.com/anchore/syft/issues/3109 (somewhat).

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] I have added unit tests that cover changed behavior
- [x] I have tested my code in common scenarios and confirmed there are no regressions
- [x] I have added comments to my code, particularly in hard-to-understand sections

PR #4295, opened 2025-10-17
feat: Add license enrichment from pypi to python packages

timols:
# Description

This PR introduces license enrichment from PyPI for python-related packages (excluding source based packages defined by a wheel or egg since license info should typically be present on the file system)

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/python/license.go
+++ b/syft/pkg/cataloger/python/license.go
@@ -90,14 +90,15 @@ func getLicenseFromPypiRegistry(baseURL, packageName, version string) (string, e
 	if err != nil {
 		return "", fmt.Errorf("unable to get package from pypi registry: %w", err)
 	}
-	if resp.StatusCode != 200 {
-		return "", fmt.Errorf("unable to get package from pypi registry")
-	}
 	defer func() {
 		if err := resp.Body.Close(); err != nil {
 			log.Errorf("unable to close body: %+v", err)
 		}
 	}()
+
+	if resp.StatusCode != 200 {
+		return "", fmt.Errorf("unable to get package from pypi registry")
+	}
 
 	bytes, err := io.ReadAll(resp.Body)
 	if err != nil {
--- a/syft/pkg/cataloger/python/license_test.go
+++ b/syft/pkg/cataloger/python/license_test.go
@@ -51,7 +51,11 @@ func TestFormatPyPiRegistryURL(t *testing.T) {
 			got, err := formatPypiRegistryURL(cfg.PypiBaseURL, test.name, test.version)
 
 			require.Equal(t, test.expected, got)
-			require.Equal(t, test.expectedError, err)
+			if test.expectedError != nil {
+				require.ErrorContains(t, err, test.expectedError.Error())
+			} else {
+				require.NoError(t, err)
+			}
 		})
 	}
 
@@ -112,15 +116,13 @@ func TestGetLicenseFromPypiRegistry(t *testing.T) {
 			}
 			got, err := getLicenseFromPypiRegistry(url, tc.name, tc.version)
 			require.Equal(t, tc.expected, got)
-			if err != nil {
-				require.Equal(t, tc.expectedError.Error(), err.Error())
+			if tc.expectedError != nil {
+				require.ErrorContains(t, err, tc.expectedError.Error())
 			} else {
-				require.Equal(t, tc.expectedError, err)
+				require.NoError(t, err)
 			}
-
 		})
 	}
-
 }
 
 type handlerPath struct {


syft/pkg/cataloger/python/package.go
```diff
@@ -123,3 +136,85 @@ func vcsURLQualifierForPackage(p *pkg.PythonDirectURLOriginInfo) packageurl.Qual
...
+	requestURL, err = url.JoinPath(baseURL, urlPath...)
+	if err != nil {
+		return requestURL, fmt.Errorf("unable to format pypi request for pkg:version %s%s; %w", packageName, version, err)
+	}
+	return requestURL, nil
+}
+
+func getLicenseFromPypiRegistry(baseURL, packageName, version string) (string, error) {
+	// "https://pypi.org/pypi/%s/%s/json", packageName, version
+	requestURL, err := formatPypiRegistryURL(baseURL, packageName, version)
```
kzantow: One thing I missed with the prior PR, since I think it was using an existing function, is that we have a simple mechanism to cache results we should use for remote fetches like this, for example [golang license lookups use this caching](https://github.com/anchore/syft/blob/07029ead8a2c6ba727757e494b516ea740b210e2/syft/pkg/cataloger/golang/licenses.go#L174). We should probably use this functionality for any new remote fetches. (And of course not as part of this PR, but retrofit any existing fetching.) Aside from this detail, everything else looks great to me here -- especially you've made sure not to break the public API, which is very important

timols: @kzantow I think I addressed this with the most recent commit. Happy to add the license caching to the JS stuff after this PR and https://github.com/anchore/syft/pull/4284 gets merged. Appreciate the reviews!

syft/pkg/cataloger/python/license.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	resp, err := httpClient.Do(pypiRequest)
+	if err != nil {
+		return "", fmt.Errorf("unable to get package from pypi registry: %w", err)
+	}
+	if resp.StatusCode != 200 {
+		return "", fmt.Errorf("unable to get package from pypi registry")
+	}
+	defer func() {
```
kzantow, line 93: Should this precede the status code check? there can be response bodies that need to be closed with other statuses, I think

syft/pkg/cataloger/python/license_test.go
```diff
@@ -0,0 +1,175 @@
...
+			// set up the mock server
+			for _, handler := range tc.requestHandlers {
+				mux.HandleFunc(handler.path, handler.handler)
+			}
+			got, err := getLicenseFromPypiRegistry(url, tc.name, tc.version)
+			require.Equal(t, tc.expected, got)
+			if err != nil {
+				require.Equal(t, tc.expectedError.Error(), err.Error())
+			} else {
+				require.Equal(t, tc.expectedError, err)
```
kzantow: tc.expectedError could be nil here. there are some require helpers for errors like `require.NoError` and `require.ErrorContains`. I think this could be a bit better written like:
```
if tc.expectedError != nil {
  require.ErrorContains(t, err, tc.expectedError.Error())
} else {
  require.NoError(t, err)
```
or similar

syft/pkg/cataloger/python/license.go
```diff
@@ -0,0 +1,130 @@
...
+		return "", fmt.Errorf("unable to format pypi request for pkg:version %s%s; %w", packageName, version, err)
+	}
+	log.WithFields("url", requestURL).Info("downloading python package from pypi")
+
+	pypiRequest, err := http.NewRequest(http.MethodGet, requestURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("unable to format remote request: %w", err)
+	}
+
+	httpClient := &http.Client{
```
kzantow, line 85: We should have some common configuration for timeout options, retry behavior possibly, rate limiting, etc.. This is becoming more important as we add more online resolution. We also might think about adding these features in a way that could be used in parallel and to enhance existing SBOMs rather than only at creation time. I don't think that has to happen as part of this PR, but I think we're getting to the point that we need to start being conscious of some problems people will start running into more frequently like the Maven rate limiting.

syft/pkg/cataloger/python/license.go
timols: You're right.

syft/pkg/cataloger/python/license.go
timols: Yep, that makes sense.

syft/pkg/cataloger/python/license_test.go
timols: In reality we control what the expected error is and this would blow up in a very obvious way when running unit tests. That said, it's a small change to have an even better error message. 😄

timols:
Hi @kzantow - let me know if there is anything remaining to get this across the line

timols:
Hi @kzantow anything holding this back from being approved and merged?

spiffcs:
LGTM TY @timols these contributions have been excellent and I really appreciate all the thought and effort you put into both PR

timols:
Thanks @spiffcs - happy to contribute! 😄

PR #4298, opened 2025-10-18
Refactor fileresolver to not require base path

Rupikz:
# Description

1. Close source in tests and examples
2. Make `fileresolver.NewFromFile` itself-resolve absolute paths
3. ~Move `GetDirectoryExclusionFunctions` to a common package~
4. Defer `fileAnalysisPath` calculation to preserve the cleanupFn

## Type of change

- [X] Documentation (updates the documentation)
- [X] Chore (improve the developer experience, fix a test flake, etc, without changing the visible behavior of Syft)

# Checklist:


syft/internal/fileresolver/file.go
```diff
@@ -36,12 +50,10 @@ func NewFromFile(parent, path string, pathFilters ...PathIndexVisitor) (*File, e
 			Opener: nativeOSFileOpener,
 		},
 		indexer: newFileIndexer(path, cleanBase, pathFilters...),
-	}
-
-	return file, file.buildIndex()
+	}, nil
```
kzantow, line 53: this was calling `file.buildIndex()`

syft/internal/fileresolver/directory.go
```diff
@@ -10,25 +10,25 @@ import (
 
 var ErrSkipPath = errors.New("skip path")
 
-var _ file.Resolver = (*Directory)(nil)
+var _ file.Resolver = (*DirectoryResolver)(nil)
```
kzantow: I think the reason we left `Resolver` off the name is due to it being in the package name already, e.g. `fileresolver.Directory` vs `fileresolver.DirectoryResolver` is unnecessarily duplicated. I don't have a strong opinion to keep it that way.

syft/internal/fileresolver/file.go
```diff
@@ -8,26 +8,40 @@ import (
...
 	path    string
 	indexer *fileIndexer
 }
 
-// parent should be the symlink free absolute path to the parent directory
+// NewFromFile parent should be the symlink free absolute path to the parent directory
 // path is the filepath of the file we're creating content access for
-func NewFromFile(parent, path string, pathFilters ...PathIndexVisitor) (*File, error) {
-	chroot, err := NewChrootContextFromCWD(parent, parent)
+func NewFromFile(path string, pathFilters ...PathIndexVisitor) (*FileResolver, error) {
```
kzantow: It looks like this is removing the parent here, I'm pretty sure there are some use cases where this is necessary to be a directory different than the parent of the provided path

syft/source/directorysource/directory_source.go
```diff
@@ -141,83 +91,78 @@ func (s *directorySource) FileResolver(_ source.Scope) (file.Resolver, error) {
...
 
 func (s *directorySource) Close() error {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
+
 	s.resolver = nil
 	return nil
 }
 
-func GetDirectoryExclusionFunctions(root string, exclusions []string) ([]fileresolver.PathIndexVisitor, error) {
```
kzantow, line 171: I see this is moved between packages. We shouldn't change any of the public API in breaking ways. If a refactoring makes sense, we could of course _add_ something and call it from an existing spot, deprecating the old function, etc.. Although, this is returning an internal interface, maybe it should just be made internal?

syft/source/filesource/file_source.go
```diff
@@ -100,27 +90,11 @@ func New(cfg Config) (source.Source, error) {
...
-	return internal.ArtifactIDFromDigest(digest.SHA256.FromString(info).String()), d
-}
-
-func (s fileSource) ID() artifact.ID {
+func (s *fileSource) ID() artifact.ID {
 	return s.id
 }
 
-func (s fileSource) Describe() source.Description {
+func (s *fileSource) Describe() source.Description {
```
kzantow: This is also technically a breaking change, since the function will no longer work with non-pointer instances (same comment for other similar changes).

syft/internal/fileresolver/directory.go
```diff
@@ -10,25 +10,25 @@ import (
...
 
-// Directory implements path and content access for the directory data source.
-type Directory struct {
+// DirectoryResolver implements path and content access for the directory data source.
+type DirectoryResolver struct {
```
wagoodman: nit: the reason why we have this named `Directory` is because it's used in conjunction with the package, which already mentions "resolver": `fileresolver.Directory` thus the suggested rename is a little more redundant (so the original name is probably better)

syft/internal/fileresolver/file.go
```diff
@@ -8,26 +8,40 @@ import (
 )
 
 // Compile time assurance that we meet the Resolver interface.
-var _ file.Resolver = (*File)(nil)
+var _ file.Resolver = (*FileResolver)(nil)
 
-// File implements path and content access for the file data source.
-type File struct {
+// FileResolver implements path and content access for the file data source.
+type FileResolver struct {
```
wagoodman: [same comment](https://github.com/anchore/syft/pull/4298/files#r2445149366)

syft/source/directorysource/directory_source.go
wagoodman, line 171: we can't move this function since it's part of the public API and syft is v1 (we can move this in v2)

syft/source/exclude.go
```diff
@@ -1,5 +1,69 @@
...
+
+	"github.com/anchore/syft/syft/internal/fileresolver"
+	"github.com/bmatcuk/doublestar/v4"
+)
+
 type ExcludeConfig struct {
 	Paths []string
 }
+
+func GetDirectoryExclusionFunctions(root string, exclusions []string) ([]fileresolver.PathIndexVisitor, error) {
```
kzantow: This is returning an internal package. I don't think this function should be a part of this package, probably is better to be internal (or left alone)

syft/internal/fileresolver/directory.go
Rupikz: Thanks for the explanation, fixed

syft/source/directorysource/directory_source.go
Rupikz: You're absolutely right. Fixed

syft/internal/fileresolver/file.go
Rupikz: This is called from NewFromFile. I've extracted this into a separate newFromFileWithoutIndex function because I wanted to make the implementations of fileresolver.File/Directory a bit more similar

syft/internal/fileresolver/file.go
Rupikz: Looking at the tests and fileSource.FileResolver(), I didn't see a need for this. But on a broader note, is it even correct to index a directory that's different from the file's absolute path?

syft/source/filesource/file_source.go
Rupikz: Lesson learned! I won't be touching public api anymore, even for what seem like minor changes)

Rupikz:
Thanks for review and the detailed explanations. I'll be more careful to avoid making breaking changes like this in the future

wagoodman (approved):

PR #4317, opened 2025-10-29
Add experimental cataloger capabilities command

### LINKED ISSUE #4155: Command output to give more information on what catalogers look for and what they can find
A command to output more information on what a given cataloger can find, how it finds it, and what the output type (in the sbom) of a cataloger will be so I can easily answer "can Syft find thing X or not". Specifically, the biggest challenge is answering that question for binaries since some can be identified but it isn't generic and there isn't a clear list published anywhere.

Something like this even as simply as a comment/doc string for unstructured output:
```
syft cataloger list --details
| Cataloger        |  Findings Type            |  Comments                                                         |
-------------------------------------------------------------------------------------
| binary              | binary                          | Can identify: curl, nginx, java, ruby,... binaries|
------------------------------------------------------------------------------------|
| rpm                  | rpm                              | Uses rpm-db metadata to identify installed rpms and dependency information|
........
```

There could be more structured formats for this kind of information or query options, but I think even just a mechanism where I can grep for the name of the binary I'm asking about. `syft cataloger list --details | grep curl` would be reasonable.

**Why is this needed**:
When investigating a "why isn't thing X found", it would be nice to use the tool itself to answer that question rather than looking at code. Documentation would be helpful too, but since there is a cataloger query interface in the tool already, it seems useful to use that to get more information on the entries already there.

**Additional context**:
Open to other UX options/opinions for how to present this and what can be presented as long as I can have a binary name and find out if there is support for finding that binary.

wagoodman:
This adds the ability, when `SYFT_EXP_CAPABILITIES=true`, to use an internal `syft cataloger caps` command to describe cataloger capabilities such as:
- what catalogers exist
- what globs / evidence are searched for
- if a cataloger finds licenses
- if a cataloger detects particular package manager claims (listing of files, digests, package integrity hash)
- what dependencies (if any) can be detected (depth of nodes, topology of the edges, and kinds of dependencies included)
- what API and app-level configurations exist for each cataloger

This is available via an ascii table and JSON output.

> [!CAUTION]
> This is an experimental feature and can change without warning and could be removed entirely. Do not depend on this command in production.

The way the capabilities are tracked is described in depth in the [internal tooling's readme](https://github.com/anchore/syft/blob/ast-parse-cataloger-capabilities/internal/capabilities/generate/README.md).

A quick summary is that we use the source code and test observations as a basis for what catalogers exist, how they are configured, and what they output. These things are then used to cross-validate a pseudo-generated `packages/*.yaml` (some auto generated items, some manually filled in items) with a set of completion tests (tests that ensure the full universe of things are defined and self-consistent) and then used to drive the `cataloger caps` command.

Fixes #4155

### DIFF
--- a/.github/actions/bootstrap/action.yaml
+++ b/.github/actions/bootstrap/action.yaml
@@ -45,7 +45,10 @@ runs:
     - name: Install project tools
       shell: bash
       if: inputs.tools == 'true'
-      run: make tools
+      run: |
+        make tools
+        .tool/binny list
+        .tool/binny check
 
     - name: Install go dependencies
       if: inputs.go-dependencies == 'true'
--- a/.github/workflows/test-fixture-cache-publish.yaml
+++ b/.github/workflows/test-fixture-cache-publish.yaml
@@ -30,7 +30,9 @@ jobs:
           download-test-fixture-cache: false
 
       - name: Run all tests
-        run: make test
+        run: |
+          make list-tasks-targets
+          make --debug=verbose test
         env:
           # we want to rebuild the cache with no previous state
           DOWNLOAD_TEST_FIXTURE_CACHE: "false"
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ make-default: $(TASK)
 	@$(TASK)
 
 # for those of us that can't seem to kick the habit of typing `make ...` lets wrap the superior `task` tool
-TASKS := $(shell bash -c "test -f $(TASK) && $(TASK) -l | grep '^\* ' | cut -d' ' -f2 | tr -d ':' | tr '\n' ' '" ) $(shell bash -c "test -f $(TASK) && $(TASK) -l | grep 'aliases:' | cut -d ':' -f 3 | tr '\n' ' ' | tr -d ','")
+TASKS := $(shell bash -c "test -f $(TASK) && NO_COLOR=1 $(TASK) -l | grep '^\* ' | cut -d' ' -f2 | tr -d ':' | tr '\n' ' '" ) $(shell bash -c "test -f $(TASK) && NO_COLOR=1 $(TASK) -l | grep 'aliases:' | cut -d ':' -f 3 | tr '\n' ' ' | tr -d ','")
 
 .PHONY: $(TASKS)
 $(TASKS): $(TASK)
--- a/cmd/syft/internal/test/integration/package_deduplication_test.go
+++ b/cmd/syft/internal/test/integration/package_deduplication_test.go
@@ -1,7 +1,6 @@
 package integration
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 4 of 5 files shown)


internal/capabilities/packages.yaml
```diff
@@ -0,0 +1,4236 @@
...
+        package_types: # AUTO-GENERATED
+          - lua-rocks
+        json_schema_types: # AUTO-GENERATED
+          - LuarocksPackage
+        capabilities: # MANUAL - config-driven capability definitions
+          - name: license
+            default: true
+          - name: dependency.depth
+            default:
+              - direct
```
willmurphyscode: I don't see relationships being returned from https://github.com/anchore/syft/blob/153f2321ce6c076d4cfa843ac944402ee9169c38/syft/pkg/cataloger/lua/parse_rockspec.go#L82 so I'm not sure this is correct.

wagoodman: you're right, I think the fact that the parser tracked dependencies when parsing threw me. Technically for the dependencies depth description we only need to look at the list of packages, there don't necessarily need to be relationships... so the fact that this returns a single package is the clue.

cmd/syft/internal/commands/cataloger_caps.go
```diff
@@ -0,0 +1,589 @@
...
+			DefaultCatalogers: []string{"all"},
+		},
+	}
+}
+
+func CatalogerCaps(app clio.Application) *cobra.Command {
+	opts := defaultCatalogerCapsOptions()
+
+	return app.SetupCommand(&cobra.Command{
+		Use:     "caps [OPTIONS] [CATALOGER_NAMES...]",
```
kzantow: My vote would be for `details`, or `capabilities`... `caps` isn't a common contraction like `info`, using the full term would be consistent with the `SYFT_EXP_CAPABILITES` environment var requirement.

.github/workflows/validations.yaml
```diff
@@ -23,6 +23,11 @@ jobs:
 
       - name: Bootstrap environment
         uses: ./.github/actions/bootstrap
+        with:
+          # the self-consistency tests for the output of the capabilities code generation depends on unit test
+          # output from  ./syft/pkg/... packages. Therefore we need to download the test fixture cache here
+          # so that running the few unit tests as part of static analysis works correctly.
+          download-test-fixture-cache: true
```
kzantow, line 34: what if cache gets purged? What if cache gets downloaded from the PR after pushing? This seems brittle/wrong. is this something we could download from `main` directly or generate from `main` to compare?

cmd/syft/internal/commands/cataloger.go
```diff
@@ -16,5 +18,14 @@ func Cataloger(app clio.Application) *cobra.Command {
...
+	// only add cataloger info command if experimental capabilities feature is enabled
+	if isCapabilitiesExperimentEnabled() {
+		cmd.AddCommand(CatalogerCaps(app))
+	}
+
 	return cmd
 }
+
+func isCapabilitiesExperimentEnabled() bool {
+	return os.Getenv("SYFT_EXP_CAPABILITIES") == "true"
```
kzantow, line 30: ... maybe have a generic "experimental" flag for any experimental features: `SYFT_EXPERIMENTAL`, and just use this for any other things that fall into this same category?

internal/capabilities/capabilities.go
```diff
@@ -0,0 +1,125 @@
...
+//go:embed appconfig.yaml
+var appconfigYAML []byte
+
+//go:embed packages/*.yaml
+var catalogerFiles embed.FS
+
+// LoadDocument loads and returns the complete document including configs and app-configs
+func LoadDocument() (*Document, error) {
+	// parse application config
+	var appDoc struct {
```
kzantow, line 35: TIL

kzantow (approved):
I don't want to hold this up and overall I think this is fine, especially since there are no changes to the external API, so that's great. In terms of the actual configuration files, I think they are ok -- I'd like them to be split up more (one per parser or cataloger entry), and when a new one is generated it would be nice to have empty defaults for all the fields that need to be manually filled out, but this can always be refined later. Two main thoughts:

I still think we should put capabilities files in the same directory as the cataloger, though it would probably take a little weirdness of a top-level embed (e.g. `//go:embed syft/pkg/cataloger/*/*.capability.yaml` or similar), which calls back to set the results on the internal package, but it could be done without any external API. Having these in separate locations I'm pretty sure is a large part of the confusion for editing them to me -- if they were in the same directory when people edited the catalogers there would be a significantly higher chance that they are updated properly when making changes that affects these files, especially those that aren't detected by the analysis.

The other holdup I have level of integration when there is still a the dichotomy of detected capability configuration vs. required manual configuration. Including this in a test hook for all PRs for what ends up being a moderately small amount of configuration. In the example cataloger, [this is the generated amount of configuration](https://github.com/anchore/syft/pull/4455/changes#diff-56ebe0522a9b27baada3022efe13763fff7b0424d24343d59e149d25fc0041cfR3-R15), less than half of what I think is needed to be complete. (And in this particular case, it's wrong because of an oddity that required writing tests differently than expected, but that's probably a bug that should be fixed). By integrating this into the PR process, it gives us the maintenance burden of all the generator code -- almost 10,000 fairly complicated (AST parsing, etc.) LoC to generate on the order of 2,500 LoC -- that gives me pause from a maintenance perspective. I wouldn't want to get rid of this, but I wonder if it was a _hint_ rather than an _error_, would that reduce the required maintenance for something that is likely to change significantly with 2.0? For example: a user adds a new cataloger without any capabilities, the generator detects what it can and gives the user a new file to edit, it doesn't really matter if it's wrong because the user just manually edits it afterwards. I acknowledge the desire to prevent drift, though. But I foresee this causing some friction for reasons I can't put my finger on -- probably not a large amount of friction, though -- moreso some new cataloger needs to do something unique and generation doesn't work quite right, but because it's all internal we can always change it later.

All that said, this is so very useful for many people, and we can iterate on the details later.

cmd/syft/internal/commands/cataloger_caps.go
wagoodman: I'll go back to `info` then (it's shorter and simpler than the alternatives)

cmd/syft/internal/commands/cataloger.go
wagoodman: I was going to add a options struct but I wanted this flag to affect the appearance of the --help output (not show the new command if the switch is not one). This is why I did a one-off env var check.

Also, the narrower the feature flag the better in my experience. We could put more things under a single flag, but it ends up cornering us in a way (but it is simpler).

.github/workflows/validations.yaml
wagoodman: the cache busts on a per-fixture level. These are generated from main every week. So when a new test comes in and introduces a fixture or there is a fixture modification, the cache busts (just for that fixture) and is rebuilt within testing. This is purely a time savings, but there is no incorrect results from this.

If the cache is purged (which is not likely since it's hosted in OCI under `latest`) then all test fixtures will be built from scratch (which takes nearly an hour). On a cache hit (or only a fixture or two to rebuild) this takes only seconds.

(I think I'm missing some specifics about what the clarification or request for change is)

kzantow: I think this a misunderstanding on my part, or forgetting it runs the tests to get the intermediary files to generate things.

cmd/syft/internal/commands/cataloger.go
kzantow: This is not at all a blocking comment, just a suggestion that made sense no action needed 👍

wagoodman:
I took another shot at moving the cataloger.yaml files, I couldn't easily get it working earlier, but did find a way in the end. It did require introducing an `init` function to register the capabilities into an internal package, which is not ideal, but it doesn't not introduce a breaking change and does not affect the public API.

PR #4279, opened 2025-10-14
feat: 4184 gguf parser (ai artifact cataloger) part 1

spiffcs:
# Description

Add new AI artifact cataloger for Syft that detects and parses GGUF (GPT-Generated Unified Format) model files. This includes a proper package type, parser, and metadata structure so Syft can identify and describe .gguf models like any other package in the `artifacts` section of the syft sbom.

### What’s New

- ModelPkg type added to syft/pkg/type.go `ModelPkg                Type = "model"`
- GGUFFileMetadata struct introduced (syft/pkg/gguf.go) with fields for model format, architecture, quantization, tensor counts, etc. A general holding of key value pairs is also included for values seen not as frequently.
- Header parser (cataloger/aiartifact/parse_gguf.go) that reads the GGUF magic number, extracts key metadata (name, license, architecture, etc.), and applies safety limits for corrupted files.
- Cataloger (cataloger/aiartifact/cataloger.go) that locates .gguf files, reads just the header (up to 10 MB), and creates type=model packages tagged with ai-artifact, model, gguf, and ml.

Happy to renamespace the cataloger from `aiartifact` -> anything that comes up in this review

Integration hooks added in package_tasks.go and packagemetadata/names.go. Cataloger will work across both directory and image scans.

### Output Support
Syft JSON: includes full gguf-file-metadata with all parsed fields.
CycloneDX 1.6 ML-BOM: emits machine-learning-model components with GGUF metadata encoded as properties.
SPDX: Separate PR

### Demo
TODO

### Fast Follow
- OCI Support for Docker Model Registry
- Hugging Face Source support

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation (updates the documentation)

# Checklist:

- [ ] I have added unit tests that cover changed behavior
- [ ] I have tested my code in common scenarios and confirmed there are no regressions
- [ ] I have added comments to my code, particularly in hard-to-understand sections

PR #4335, opened 2025-11-05
feat: Add support for scanning GGUF models from OCI registries

spiffcs:
## Summary
This PR follows up on #4279 by adding support for a new source type.

With this change users can do the following: `syft -o json docker.io/ai/qwen3-vl | jq .`:

They'll get an SBOM with a single package showing the gguf model and details for the model pulled from `https://hub.docker.com/u/ai`

### Example of metadata extracted
```
 "metadata": {
        "modelFormat": "gguf",
        "modelName": "Qwen3-Vl-8B-Instruct",
        "modelVersion": "unknown",
        "hash": "321c13d3e93151b5",
        "license": "apache-2.0",
        "ggufVersion": 3,
        "architecture": "qwen3vl",
        "quantization": "Q4_K_M",
        "parameters": 8190735360,
        "tensorCount": 399,
        "header": {
          "general.base_model.0.name": "Qwen3 VL 8B Instruct",
          "general.base_model.0.organization": "Qwen",
          "general.base_model.0.repo_url": "https://huggingface.co/Qwen/Qwen3-VL-8B-Instruct",
          "general.base_model.count": 1,
          "general.basename": "Qwen3-Vl-8B-Instruct",
          "general.file_type": 15,
          "general.finetune": "Instruct",
          "general.quantization_version": 2,
          "general.quantized_by": "Unsloth",
          "general.repo_url": "https://huggingface.co/unsloth",
          "general.size_label": "8B",
          "general.tags": {
            "type": 8,
            "len": 2,
            "startOffset": 741,
            "size": 41
          },
          "general.type": "model",
          "quantize.imatrix.chunks_count": 694,
          "quantize.imatrix.dataset": "unsloth_calibration_Qwen3-VL-8B-Instruct.txt",
          "quantize.imatrix.entries_count": 252,
          "quantize.imatrix.file": "Qwen3-VL-8B-Instruct-GGUF/imatrix_unsloth.gguf",
          "qwen3vl.attention.head_count": 32,
          "qwen3vl.attention.head_count_kv": 8,
          "qwen3vl.attention.key_length": 128,
          "qwen3vl.attention.layer_norm_rms_epsilon": 0.000001,
          "qwen3vl.attention.value_length": 128,
          "qwen3vl.block_count": 36,
          "qwen3vl.context_length": 262144,
          "qwen3vl.embedding_length": 4096,
          "qwen3vl.feed_forward_length": 12288,
          "qwen3vl.n_deepstack_layers": 3,
          "qwen3vl.rope.dimension_sections": {
            "type": 5,
            "len": 4,
            "startOffset": 1268,
            "size": 16
          },
          "qwen3vl.rope.freq_base": 5000000,
         "tokenizer.ggml.add_bos_token": false,
          "tokenizer.ggml.bos_token_id": 151643,
          "tokenizer.ggml.eos_token_id": 151645,
          "tokenizer.ggml.merges": {
            "type": 8,
            "len": 151387,
            "startOffset": 3197544,
            "size": 2731548
          },
```
 This PR adds support for scanning OCI model artifacts (specifically GGUF-based AI models) stored in OCI registries. The
  implementation enables Syft to extract SBOM information from AI models without downloading the entire model file—only
  the GGUF headers are fetched using a lazy reader that closes after we've read all the header bytes.

  Key changes:

  - New ocimodelsource package implementing a source provider for OCI model artifacts
  - Registry client that fetches only GGUF headers (up to 8MB) instead of full model files
  - New OCIMediaTypeResolver interface and ContainerImageModel file resolver for media type-based file discovery
  - Extended generic cataloger with WithParserByMediaType() to support cataloging by OCI media type
  - Post-processor for GGUF cataloger to merge metadata found from multi-layer model artifacts

  OCI Model Source

  This new source type enables scanning AI models stored as OCI artifacts in container registries (like Docker Hub or other OCI registries).

  How it works:

  1. Detection: Identifies model artifacts by checking if the manifest's config media type starts with
  application/vnd.docker.ai.model.config.*
  2. Selective fetching: Instead of downloading entire GGUF layers (often 4-70GB), it uses HTTP range requests to fetch
  only the first 8MB containing the GGUF header metadata. The download is terminated early by closing the reader after
  reading the header bytes.
  3. Temporary storage: Headers are written to temp files, keyed by the layer digest, and cleaned up when the source is
  closed.
  4. File resolver: A new ContainerImageModel resolver implements both file.Resolver and file.OCIMediaTypeResolver,
  allowing catalogers to discover layers by OCI media type (application/vnd.docker.ai.gguf.v3) rather than filesystem
  paths. 
  5. Provider integration: Registered as the oci-model provider, positioned after the regular pull providers in the source
   provider chain—so normal container images are handled first, and this kicks in only for artifacts with model-specific
  media types. It also makes it so that for syft's current MOST common cases (non ai containers) we're not adding extra work.

Keys Goals:
  - Scanning a 70GB LLaMA model takes seconds instead of minutes
  - No disk space needed for full model storage
  - Works with Docker's AI model packaging spec and Ollama registries


syft/source/ocimodelsource/metadata.go
```diff
@@ -0,0 +1,39 @@
...
+	OS             string                 `json:"os"`
+	Labels         map[string]string      `json:"labels,omitempty"`
+
+	// OCI-specific metadata
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Model-specific metadata
+	ModelFormat string          `json:"modelFormat,omitempty"` // e.g., "gguf"
+	GGUFLayers  []GGUFLayerInfo `json:"ggufLayers,omitempty"`
+}
```
wagoodman:
```suggestion
type OCIModelMetadata{
   ImageMetadata
   Annotations map[string]string
}
```

or


```suggestion
type OCIModelMetadata ImageMetadata
```

wagoodman: I'm not saying this must be done, but given that the gguf/model format properties are redundant, then it is essentially the same. The annotations are new, however, still relevant to OCI images.

wagoodman: we should probably have `id:-` to ensure this doesn't affect the ID of existing sources. Probably worth a test to ensure this isn't changing

syft/pkg/cataloger/ai/cataloger.go
```diff
@@ -5,12 +5,165 @@ including support for GGUF (GPT-Generated Unified Format) model files.
...
+
+	// Fall back to generic glob-based discovery
+	log.Debug("using glob-based discovery for GGUF models")
+	return c.genericCataloger.Catalog(ctx, resolver)
+}
+
+// catalogFromOCILayers discovers GGUF models by querying OCI layers by media type.
+func (c *ggufCataloger) catalogFromOCILayers(ctx context.Context, resolver ocimodelsource.OCIResolver) ([]pkg.Package, []artifact.Relationship, error) {
+	// Find all GGUF layers by media type
+	digests, err := resolver.LayerDigestsByMediaType(ocimodelsource.GGUFLayerMediaType)
```
wagoodman: we shouldn't import from a package meant for a specific implementation of a source

syft/source/ocimodelsource/registry_client.go
```diff
@@ -0,0 +1,224 @@
...
+		return opts
+	}
+
+	// Build authenticator
+	authenticator := buildAuthenticator(registryOpts)
+	opts = append(opts, remote.WithAuth(authenticator))
+
+	// Handle TLS settings
+	if registryOpts.InsecureSkipTLSVerify {
+		transport := remote.DefaultTransport.(*http.Transport).Clone()
```
wagoodman: beware of possible panic here

syft/source/ocimodelsource/registry_client.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// Fetch descriptor
+	desc, err := remote.Get(ref, c.options...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch descriptor: %w", err)
+	}
+
+	// Parse manifest
+	manifest := &v1.Manifest{}
+	if err := json.Unmarshal(desc.Manifest, manifest); err != nil {
```
wagoodman, line 130: this could be a v1.Index maybe? worth checking out https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/puller.go#L93 and if its connected or not to https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/index.go  and https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/descriptor.go#L44-L45. We should probably check if this is a manifest or index and return a helpful error if its an index.

syft/source/ocimodelsource/registry_client.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// Get the compressed reader
+	reader, err := layer.Compressed()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get layer reader: %w", err)
+	}
+	defer reader.Close()
+
+	// Read up to maxBytes
+	data := make([]byte, maxBytes)
```
wagoodman, line 209: a comment that we're capped at 10 MB, not reading a full GGUF file into memory

syft/source/ocimodelsource/oci_model_source_provider.go
```diff
@@ -0,0 +1,76 @@
...
+	if !isModel {
+		log.WithFields("reference", p.reference).Debug("reference is not a model artifact")
+		return nil, fmt.Errorf("not an OCI model artifact")
+	}
+
+	log.WithFields("reference", p.reference).Info("detected OCI model artifact, fetching headers")
+
+	// Fetch the full model artifact with metadata
+	artifact, err := client.FetchModelArtifact(ctx, p.reference)
+	if err != nil {
```
wagoodman: `client.IsModelArtifactReference` shuold probably return the `remote.Descriptor` so we don't call `remote.Get` again when you are using `client.FetchModelArtifact`.

syft/source/sourceproviders/source_providers.go
```diff
@@ -40,6 +42,9 @@ func All(userInput string, cfg *Config) []collections.TaggedValue[source.Provide
 
 		// 3. try remote sources after everything else...
 
+		// --from oci-model (model artifacts with header-only fetching)
+		Join(tagProvider(ocimodelsource.NewSourceProvider(userInput, cfg.RegistryOptions, cfg.Alias), OCIModelTag)).
```
wagoodman: This is tech debt for later: we're starting to do some work across independent providers that is wasted work and could consider some caching or another solution to share work across providers (e.g. this calls `remote.Get` as well as the `registry` provider... and this is called before the registry/docker cases (our most common case).

We should look into options where we can put this line below the pull tag approach.

wagoodman: Say for instance, we start updating the providers down in stereoscope to ignore the image entirely when there is a gguf mediatype layer found ( https://github.com/anchore/stereoscope/blob/8c8ea254b2613dae35f475343354c15ca04e16b0/pkg/image/containerd/daemon_provider.go#L230 and https://github.com/anchore/stereoscope/blob/8c8ea254b2613dae35f475343354c15ca04e16b0/pkg/image/docker/daemon_provider.go#L272-L290 and https://github.com/anchore/stereoscope/blob/8c8ea254b2613dae35f475343354c15ca04e16b0/pkg/image/oci/registry_provider.go#L73-L78 ... etc)

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -0,0 +1,257 @@
...
+// ociModelSource implements the source.Source interface for OCI model artifacts.
+type ociModelSource struct {
+	id       artifact.ID
+	config   Config
+	resolver *ociModelResolver
+	mutex    *sync.Mutex
+}
+
+// NewFromArtifact creates a new OCI model source from a fetched model artifact.
+func NewFromArtifact(artifact *ModelArtifact, client *RegistryClient, alias source.Alias) (source.Source, error) {
```
wagoodman: takes as input a type that is really an internal concern for this package. This should really pass in a description of what the user is asking (declarative) and let the constructor do the remote checks / work to figure out if it's valid. I think much of the work that is in the provider should really be in the constructor here, and the constructor should just be `New(Config)`

syft/source/ocimodelsource/resolver.go
```diff
@@ -0,0 +1,256 @@
...
+
+// ociModelResolver is a file.Resolver implementation that provides access to
+// GGUF header data fetched from OCI model artifacts via range-GET requests.
+// It also implements OCIResolver for layer-aware access patterns.
+type ociModelResolver struct {
+	tempFiles map[string]string // maps virtual path -> temporary file path
+	locations []file.Location
+
+	// OCI layer-aware fields
+	client   *RegistryClient // registry client for fetching layers
```
wagoodman: client should not be here (that's what the temp location is for)

syft/internal/fileresolver/container_image_model.go
wagoodman, line 1: there are some implementations for file.Resolver that should really simply return nil, nil (but they currently have other behavior)

spiffcs:
First big change to make here is how the Provider and Source interact. Provider should be thin and call Source, Source should do most of the work, provider should return the source.Source answer.

syft/source/ocimodelsource/registry_client.go
spiffcs: We checked this in sync review and found this would always be a Manifest

syft/source/ocimodelsource/metadata.go
```diff
@@ -3,37 +3,7 @@ package ocimodelsource
...
-
-// GGUFLayerInfo represents metadata about a GGUF layer in the OCI artifact.
-type GGUFLayerInfo struct {
-	Digest       string            `json:"digest"`
-	Size         int64             `json:"size"`      // Full blob size in registry
-	MediaType    string            `json:"mediaType"` // Should be "application/vnd.docker.ai.gguf.v3"
-	Annotations  map[string]string `json:"annotations,omitempty"`
-	FetchedBytes int64             `json:"fetchedBytes"` // How many bytes we actually fetched via range-GET
+	source.ImageMetadata
+	Annotations map[string]string `json:"annotations,omitempty" id:"-"`
```
wagoodman: we should add a comment as to why id is `-` (otherwise this could get removed and cause subtle bugs down the line)

syft/source/ocimodelsource/metadata.go
```diff
@@ -3,37 +3,7 @@ package ocimodelsource
...
-}
-
-// GGUFLayerInfo represents metadata about a GGUF layer in the OCI artifact.
-type GGUFLayerInfo struct {
-	Digest       string            `json:"digest"`
-	Size         int64             `json:"size"`      // Full blob size in registry
-	MediaType    string            `json:"mediaType"` // Should be "application/vnd.docker.ai.gguf.v3"
-	Annotations  map[string]string `json:"annotations,omitempty"`
-	FetchedBytes int64             `json:"fetchedBytes"` // How many bytes we actually fetched via range-GET
+	source.ImageMetadata
```
wagoodman: missing json inline struct tag

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -18,95 +21,119 @@ import (
...
-	Alias     source.Alias
-	Client    *RegistryClient
-	Metadata  *OCIModelMetadata
-	TempFiles map[string]string // Virtual path -> temp file path
-
-	// OCI layer-aware fields for the resolver
-	Ref      name.Reference // parsed OCI reference
-	Manifest *v1.Manifest   // manifest containing layer information
+	Reference    string
+	RegistryOpts *image.RegistryOptions
```
wagoodman:
```suggestion
RegistryOptions *image.RegistryOptions
```
we should be consistent with the ImageConfig

syft/source/ocimodelsource/registry_client.go
wagoodman: unexport this type

wagoodman: unexport this constructor

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -247,8 +265,10 @@ func (s *ociModelSource) Close() error {
 	defer s.mutex.Unlock()
 
 	if s.resolver != nil {
-		if err := s.resolver.cleanup(); err != nil {
-			return err
+		if r, ok := s.resolver.(*ociModelResolver); ok {
+			if err := r.cleanup(); err != nil {
```
wagoodman: the resolver shouldn't really be involved with cleanup since the source creates/owns this temp dir

syft/source/ocimodelsource/registry_client.go
wagoodman: these constants should not be used by consumers, so if used here should be unexported

wagoodman: this should be unexported

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -18,95 +21,119 @@ import (
...
+	}
 
-		// Create temp file
-		tempPath, err := createTempFileFromData(headerData, virtualPath)
+	layerFiles := make(map[string]layerInfo)
+	for _, layer := range artifact.GGUFLayers {
+		li, err := fetchSingleGGUFHeader(ctx, client, artifact.Reference, layer, tempDir)
 		if err != nil {
-			return nil, fmt.Errorf("failed to create temp file: %w", err)
+			os.RemoveAll(tempDir)
```
wagoodman: should not be ignoring an error (we should at least log)

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -18,95 +21,119 @@ import (
...
+	}
+
+	return artifact, nil
+}
 
-		// Extract virtual path from annotations
-		virtualPath := extractVirtualPath(idx)
+// fetchAndStoreGGUFHeaders fetches GGUF layer headers and stores them in temp files.
+func fetchAndStoreGGUFHeaders(ctx context.Context, client *RegistryClient, artifact *ModelArtifact) (string, map[string]layerInfo, error) {
+	tempDir, err := os.MkdirTemp("", "oci-gguf")
```
wagoodman:
```suggestion
tempDir, err := os.MkdirTemp("", "syft-oci-gguf")
```
makes it easier to understand un crashes what resources are ours

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -116,55 +143,48 @@ func buildMetadata(artifact *ModelArtifact) *OCIModelMetadata {
...
-		return make(map[string]string)
+		ImageMetadata: source.ImageMetadata{
+			UserInput:      artifact.Reference.String(),
+			ID:             artifact.ManifestDigest,
+			ManifestDigest: artifact.ManifestDigest,
+			MediaType:      string(artifact.Manifest.MediaType),
+			Tags:           tags,
+			Size:           calculateTotalSize(layers),
+			Layers:         layers,
+			RawManifest:    artifact.RawManifest,
```
wagoodman: can you confirm that the manifest and config are base64 encoded?

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -18,95 +21,119 @@ import (
...
 	metadata := buildMetadata(artifact)
+	tempDir, layerFiles, err := fetchAndStoreGGUFHeaders(ctx, client, artifact)
+	if err != nil {
+		return nil, err
+	}
 
-	// Fetch GGUF layer headers via range-GET
-	tempFiles := make(map[string]string)
-	ggufLayers := make([]GGUFLayerInfo, 0, len(artifact.GGUFLayers))
+	id := deriveID(cfg.Reference, cfg.Alias, metadata.ManifestDigest)
```
wagoodman: we should have a test that ensures the ID generated for an artifact here vs an stereoscope image is the same

syft/source/ocimodelsource/oci_model_source.go
spiffcs: confirmed on sync call these are encoded

syft/source/ocimodelsource/oci_model_source_provider.go
spiffcs: removed the double Get here to reduce network calls

syft/source/ocimodelsource/oci_model_source.go
wagoodman: well we should really try to move the deriveIDFromStereoscopeImage function in the stereoscopesource package to source/internal package and share the invocation. We should be careful to not change existing behavior. Could be worth capturing an ID for each case we care about before making refactors, adding a test for that (if it does not already exist), then doing the refactor and ensure they dont change for stereoscope images.

syft/source/ocimodelsource/registry_client.go
wagoodman: true! we can always use v1.ParseManifest(bytes.NewReader(dest.Manifest)) though, which would be nice

syft/source/ocimodelsource/registry_client.go
```diff
@@ -172,26 +179,23 @@ func extractGGUFLayers(manifest *v1.Manifest) []v1.Descriptor {
...
 	repo := ref.Context()
 
-	// Fetch the layer (blob) using remote.Layer
 	layer, err := remote.Layer(repo.Digest(digest.String()), c.options...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch layer: %w", err)
 	}
 
-	// Get the compressed reader
 	reader, err := layer.Compressed()
```
wagoodman, line 192: can you verify this is truly an incremental pull of the layer?

syft/internal/fileresolver/container_image_model.go
wagoodman, line 1: lets try to stick with existing conventions and house this under syft/internal/fileresolver

syft/source/ocimodelsource/registry_client.go
spiffcs: 👍 just validated that this is done via partial downloads which closes when the reader closes

syft/source/stereoscopesource/image_source.go
```diff
@@ -163,61 +162,6 @@ func imageMetadataFromStereoscopeImage(img *image.Image, reference string) sourc
...
 }
 
-// deriveIDFromStereoscopeImage derives an artifact ID from the given image metadata. The order of data precedence is:
-//  1. prefer a digest of the raw container image manifest
-//  2. if no manifest digest is available, calculate a chain ID from the image layer metadata
-//  3. if no layer metadata is available, use the user input string
-//
-// in all cases, if an alias is provided, it is additionally considered in the ID calculation. This allows for the
-// same image to be scanned multiple times with different aliases and be considered logically different.
-func deriveIDFromStereoscopeImage(alias source.Alias, metadata source.ImageMetadata) artifact.ID {
```
spiffcs, line 173: Reviewer note: This has been moved to the `internal` package so it can be shared between the new OCI source and the stereoscope source.

spiffcs: A test has been added the captures ID from the previous implementation. These snapshots ensure that the behavior has not changed AND will not change moving forward

syft/source/internal/image_id.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+// DeriveImageID derives an artifact ID from the given image metadata. The order of data precedence is:
+//  1. prefer a digest of the raw container image manifest
+//  2. if no manifest digest is available, calculate a chain ID from the image layer metadata
+//  3. if no layer metadata is available, use the user input string
+//
+// in all cases, if an alias is provided, it is additionally considered in the ID calculation. This allows for the
+// same image to be scanned multiple times with different aliases and be considered logically different.
+func DeriveImageID(alias source.Alias, metadata source.ImageMetadata) artifact.ID {
```
spiffcs, line 19: Moved from `syft/source/stereoscopesource/image_source.go`

syft/source/ocimodelsource/metadata.go
wagoodman, line 1: this should be in the source package

syft/source/ocimodelsource/metadata.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/anchore/syft/syft/source"
+
+// OCIModelMetadata represents all static metadata that defines what an OCI model artifact is.
+// The struct tag for Annotations with id:"-" is used so that we do not generate different source ID based
+// on changes to the annotations of a source image. This does not constitute a different identifiable.
+type OCIModelMetadata struct {
+	source.ImageMetadata `json:",inline"`
+	Annotations          map[string]string `json:"annotations,omitempty" id:"-"`
+}
```
wagoodman: is there anything really distinct between an OCi model metadata and image metadata? What if we want to add annotations (which is an OCI inherent property) to ImageMetadata?

syft/source/ocimodelsource/oci_model_source.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil, err
+	}
+
+	metadata := buildMetadata(artifact)
+	tempDir, resolver, err := fetchAndStoreGGUFHeaders(ctx, client, artifact)
+	if err != nil {
+		return nil, err
+	}
+
+	id := internal.DeriveImageID(cfg.Alias, metadata.ImageMetadata)
```
wagoodman: since this method passes in the image metadata and not the whole metadata the new tests doesn't do anything (and there is no need to add one)

cmd/syft/internal/commands/scan.go
```diff
@@ -59,7 +59,9 @@ const (
 	nonImageSchemeHelp = `    {{.appName}} {{.command}} dir:path/to/yourproject                  read directly from a path on disk (any directory)
     {{.appName}} {{.command}} file:path/to/yourproject/file            read directly from a path on disk (any single file)
 `
-	scanSchemeHelp = "\n  " + schemeHelpHeader + "\n" + imageSchemeHelp + nonImageSchemeHelp
+	modelSchemeHelp = `    {{.appName}} {{.command}} oci-model-registry:ai/llama3.2           scan an OCI model artifact from a registry (e.g. Docker Hub AI models)
```
kzantow: I assume `oci-model-registry` is something already settled on; it makes sense, but is there any value to have non-model things use the same partial pull?

PR #4345, opened 2025-11-07
Add Indent Option to Output Formats

JoeyShapiro:
# Description

It would be nice if the syft package allowed for pretty printing SBOMs with custom indenting. The default for most formats is only 1 space, but I personally use a tab.

This change is in the works, but I wanted to bring attention to it because I have a few issues.
- the cyclone formats use a separate encoder package, that do not allow setting the indent
  - I can make a PR and do the changes to that package, but it would require a bit more coordination.
- Some formats do not have a `Pretty` option.
  - I can add an Indent option, but I feel like I should add a Pretty option while I'm at it. Unless there is a reason they do not have a Pretty option
  - But adding an Indent option without a Pretty option feels wrong for some reason

## Type of change

# Checklist:

### DIFF
--- a/syft/format/spdxjson/encoder.go
+++ b/syft/format/spdxjson/encoder.go
@@ -41,7 +41,7 @@ func DefaultEncoderConfig() EncoderConfig {
 	return EncoderConfig{
 		Version: spdxutil.DefaultVersion,
 		Pretty:  false,
-		Indent:  " ",
+		Indent:  "  ",
 	}
 }
 


syft/format/spdxjson/encoder.go
```diff
@@ -40,6 +41,7 @@ func DefaultEncoderConfig() EncoderConfig {
 	return EncoderConfig{
 		Version: spdxutil.DefaultVersion,
 		Pretty:  false,
+		Indent:  " ",
```
kzantow: we should default this to either 2 or 4 spaces, probably.

JoeyShapiro: that's fine by me. I was just copying the default from what it was before. I think the rest of the code uses 2, so I will do that

PR #4351, opened 2025-11-07
Migrate CI to runs-on

wagoodman:
Migrated core workflow files to use runs-on EC2 runners for better resource allocation and cost optimization (this is cheaper and faster on average). This leverages tmpfs mounts for most runners for testing and building since we don't need persistence. This shaves off ~5 minutes from CI runs on a PR.

Before this change:
Screenshot 2025-12-31 at 3 03 26 PM

After this change:
Screenshot 2025-12-31 at 3 03 11 PM

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -39,9 +39,9 @@ vars:
   # e.g. when installing snapshot debs from a local path, ./ forces the deb to be installed in the current working directory instead of referencing a package name
   SNAPSHOT_DIR: ./snapshot
   SNAPSHOT_BIN: "{{ .PROJECT_ROOT }}/{{ .SNAPSHOT_DIR }}/{{ .OS }}-build_{{ .OS }}_{{ .ARCH }}/{{ .PROJECT }}"
-  SNAPSHOT_CMD: "{{ .TOOL_DIR }}/goreleaser release --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --skip=publish --skip=sign --parallelism=12"
-  BUILD_CMD:    "{{ .TOOL_DIR }}/goreleaser build   --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --single-target --parallelism=12"
-  RELEASE_CMD:  "{{ .TOOL_DIR }}/goreleaser release --clean --release-notes {{ .CHANGELOG }} --parallelism=12"
+  SNAPSHOT_CMD: "{{ .TOOL_DIR }}/goreleaser release --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --skip=publish --skip=sign"
+  BUILD_CMD:    "{{ .TOOL_DIR }}/goreleaser build   --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --single-target"
+  RELEASE_CMD:  "{{ .TOOL_DIR }}/goreleaser release --clean --release-notes {{ .CHANGELOG }}"
   VERSION:
     sh: git describe --dirty --always --tags
 
@@ -225,14 +225,14 @@ tasks:
       COVERAGE_THRESHOLD: 62
     cmds:
       - task: clean-test-observations
-      - "go test -p=4 -parallel=8 -coverprofile {{ .TMP_DIR }}/unit-coverage-details.txt {{ .TEST_PKGS }}"
+      - "go test -coverprofile {{ .TMP_DIR }}/unit-coverage-details.txt {{ .TEST_PKGS }}"
       - cmd: ".github/scripts/coverage.py {{ .COVERAGE_THRESHOLD }} {{ .TMP_DIR }}/unit-coverage-details.txt"
         silent: true
 
   integration:
     desc: Run integration tests
     cmds:
-      - "go test -v -p=4 -parallel=8 ./cmd/syft/internal/test/integration"
+      - "go test -v ./cmd/syft/internal/test/integration"
       # exercise most of the CLI with the data race detector
       # we use a larger image to ensure we're using multiple catalogers at a time
       - "go run -race cmd/syft/main.go anchore/test_images:grype-quality-dotnet-69f15d2"


.github/actions/bootstrap/action.yaml
```diff
@@ -13,7 +13,7 @@ inputs:
   cache-key-prefix:
     description: "Prefix all cache keys with this value"
     required: true
-    default: "181053ac82"
+    default: "53ac821810"
```
wagoodman, line 16: (since we have some fundamental changes, lets start with fresh caches)

.github/workflows/test-fixture-cache-publish.yaml
```diff
@@ -13,8 +13,12 @@ jobs:
...
   Publish:
     name: "Publish test fixture image cache"
-    # we use this runner to get enough storage space for docker images and fixture cache
-    runs-on: ubuntu-22.04-4core-16gb
+    # runs-on.com: memory-optimized instances for building test fixture cache
+    #   spot disabled: can run up to an hour, avoid interruptions
+    #   disk=large: need more storage for docker images and fixture cache
+    #   s3-cache: faster actions cache
+    #   tmpfs: faster io-intensive workflows
+    runs-on: runs-on=${{ github.run_id }}/cpu=4+8/ram=64+128/family=r5+r6+r7+r8/spot=false/disk=large/extras=s3-cache+tmpfs
```
willmurphyscode: Is it correct to have `disk=large` and `+tmpfs`? If I understand correctly, tmpfs mounts everything in RAM for faster I/O operations. Then what good does the large disk do? https://runs-on.com/caching/tmpfs/

.github/workflows/validations.yaml
```diff
@@ -12,14 +12,17 @@ on:
...
-
   Static-Analysis:
     # Note: changing this job name requires making the same update in the .github/workflows/release.yaml pipeline
     name: "Static analysis"
-    runs-on: ubuntu-24.04
+    # runs-on.com: memory & general purpose instances for testing
+    #   spot enabled: ok to interrupt non-production workloads
+    #   s3-cache: faster actions cache
+    #   tmpfs: faster io-intensive workflows
+    runs-on: &test-runner "runs-on=${{ github.run_id }}/cpu=4+8/ram=32+128/family=r5+r6+r7+r8+m4+m5+m6+m7+m8/spot=price-capacity-optimized/extras=s3-cache+tmpfs"
```
willmurphyscode, line 27: The `ram=32+128` syntax is confusing. This means accept any amount of RAM from 32GB up to 128GB I think. If so, that's fine. It's just confusing to have + separate the ends of a range. (Not really feedback on this PR so much as feedback on runs-on's yaml choices.)

I had to look this syntax up to convince myself it was right, so here's the link: https://runs-on.com/guides/best-practices/#how-should-i-specify-ram-and-cpu-requirements

.github/workflows/validations.yaml
```diff
@@ -55,11 +56,10 @@ jobs:
       - name: Run unit tests
         run: make unit
 
-
   Integration-Test:
     # Note: changing this job name requires making the same update in the .github/workflows/release.yaml pipeline
     name: "Integration tests"
-    runs-on: ubuntu-24.04
+    runs-on: *test-runner
```
willmurphyscode, line 64: I think I like having just the "this is a test runner". It's defined locally in the file, so it's obvious what's affected by any change, but I don't have to type it every time. 👍

.github/workflows/validations.yaml
```diff
@@ -76,11 +76,18 @@ jobs:
...
-    runs-on: ubuntu-24.04
+    # runs-on.com: compute instances for parallel builds
+    #   spot disabled: reliability for build workflows (used for releases too)
+    #   goreleaser uses parallelism of 12, so we need more CPUs
+    #   s3-cache: faster actions cache
+    #   tmpfs: faster io-intensive workflows
+    runs-on: "runs-on=${{ github.run_id }}/cpu=16+32/ram=32+128/family=c5+c6+c7+c8/spot=false/extras=s3-cache+tmpfs"
     steps:
+      # required for magic-cache from runs-on to function with artifact upload/download (see https://runs-on.com/caching/magic-cache/#actionsupload-artifact-compatibility)
+      - uses: runs-on/action@v2
```
willmurphyscode: Should this be pinned?

.github/workflows/validations.yaml
```diff
@@ -232,34 +172,30 @@ jobs:
...
-
   Cli-Linux:
     # Note: changing this job name requires making the same update in the .github/workflows/release.yaml pipeline
     name: "CLI tests (Linux)"
     needs: [Build-Snapshot-Artifacts]
-    runs-on: ubuntu-24.04
+    runs-on: *test-runner
     steps:
+      # required for magic-cache from runs-on to function with artifact upload/download (see https://runs-on.com/caching/magic-cache/#actionsupload-artifact-compatibility)
+      - uses: runs-on/action@v2
```
willmurphyscode: Should this be pinned?

Taskfile.yaml
```diff
@@ -39,9 +39,9 @@ vars:
   # e.g. when installing snapshot debs from a local path, ./ forces the deb to be installed in the current working directory instead of referencing a package name
   SNAPSHOT_DIR: ./snapshot
   SNAPSHOT_BIN: "{{ .PROJECT_ROOT }}/{{ .SNAPSHOT_DIR }}/{{ .OS }}-build_{{ .OS }}_{{ .ARCH }}/{{ .PROJECT }}"
-  SNAPSHOT_CMD: "{{ .TOOL_DIR }}/goreleaser release --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --skip=publish --skip=sign"
-  BUILD_CMD:    "{{ .TOOL_DIR }}/goreleaser build   --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --single-target"
-  RELEASE_CMD:  "{{ .TOOL_DIR }}/goreleaser release --clean --release-notes {{ .CHANGELOG }}"
+  SNAPSHOT_CMD: "{{ .TOOL_DIR }}/goreleaser release --config {{ .TMP_DIR }}/goreleaser.yaml --clean --snapshot --skip=publish --skip=sign --parallelism=12"
```
willmurphyscode: Is there a syntax for specifying `parellism=auto` so that we saturate the runner if we're assigned a large one?

Taskfile.yaml
```diff
@@ -225,14 +225,14 @@ tasks:
       COVERAGE_THRESHOLD: 62
     cmds:
       - task: clean-test-observations
-      - "go test -coverprofile {{ .TMP_DIR }}/unit-coverage-details.txt {{ .TEST_PKGS }}"
+      - "go test -p=4 -parallel=8 -coverprofile {{ .TMP_DIR }}/unit-coverage-details.txt {{ .TEST_PKGS }}"
```
willmurphyscode: Same question: is there a way to as the tool to use the best parallelism to saturate the machine it's assigned?

.github/workflows/test-fixture-cache-publish.yaml
wagoodman: correct, I accidentally left that from a previous iteration -- it should be removed

.github/workflows/validations.yaml
wagoodman: > This means accept any amount of RAM from 32GB up to 128GB

yup 👍

wagoodman: > The ram=32+128 syntax is confusing

yup 👍

.github/workflows/validations.yaml
wagoodman: yeah, this is the way to go I feel. The config files and shared locations have too many caveats.

.github/workflows/validations.yaml
wagoodman: yeah, I'll pin it

Taskfile.yaml
wagoodman: in theory I could not set the flag and it would expand to use the available CPUs https://goreleaser.com/customization/builds/go/?h=parallelism#gomaxprocs-for-goreleaser ... I should try that

Taskfile.yaml
wagoodman: `-p` controls how many test packages can be run in parallel and `-parallel` controls how many test functions within a package (when opted in with `t.Parallel()` within the test) can be run simultaneously. It looks like GOMAXPROC is the default for `-parallel` and `-p`.

wagoodman:
I'm leaving all parallelism options to their defaults on this latest run (it should perform similarly since the defaults are sane) https://github.com/anchore/syft/actions/runs/20718016707

PR #4352, opened 2025-11-07
Support `extras` statements in Python PDM cataloger

wagoodman:
PDM lock files represent the same logical package as multiple `[[package]]` entries when extras are used in the dependency tree. This differs from Poetry, which uses a single entry with conditional markers.

For example, when a project depends on `coverage[toml]`, PDM generates:

```toml
[[package]]
name = "coverage"
version = "7.4.1"
# base package

[[package]]
name = "coverage"
version = "7.4.1"
extras = ["toml"]
dependencies = [
    "coverage==7.4.1",
    "tomli; python_full_version <= \"3.11.0a6\"",
]
```

Similarly, if different packages require different extras combinations (e.g., `package[redis]` and `package[postgres]`), PDM creates separate entries for each.

The fix in this PR merges `pkg.Package` instances for the same high-level/logical package together. Variants are captured in the metadata.

### Changes
- removed custom buildPdmRelationships() function and used the standard dependency specifier pattern used in other catalogers
- added support for detecting and processing extras (added PythonPdmLockExtraVariant struct to represent extras combinations)
- preserves full dependency strings including environment markers (e.g., `; python_version < "3"`).... these were being stripped
- filters self-references in extras dependencies (e.g., `coverage[toml]` depending on `coverage==7.4.1`)

## Type of change

## Checklist

### DIFF
--- a/syft/pkg/cataloger/python/parse_pdm_lock.go
+++ b/syft/pkg/cataloger/python/parse_pdm_lock.go
@@ -90,10 +90,13 @@ func mergePdmLockPackages(packages []pdmLockPackage) pkg.PythonPdmLockEntry {
 	var baseFiles []pkg.PythonPdmFileEntry
 
 	// Separate base package from extras variants
+	// note: this logic processes packages in order and assumes the base package (no extras) appears
+	// before extras variants in the PDM lock file, which is PDM's current behavior
 	for _, p := range packages {
 		// Convert files format
 		var files []pkg.PythonPdmFileEntry
 		for _, f := range p.Files {
+			// skip files with invalid hash format (missing colon separator between algorithm and value)
 			if colonIndex := strings.Index(f.Hash, ":"); colonIndex != -1 {
 				algorithm := f.Hash[:colonIndex]
 				value := f.Hash[colonIndex+1:]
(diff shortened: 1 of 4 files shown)


Kelleretoro (approved):

Kelleretoro (approved):

syft/pkg/cataloger/python/parse_pdm_lock.go
```diff
@@ -51,102 +53,149 @@ func newPdmLockParser(cfg CatalogerConfig) pdmLockParser {
...
-		var deps []string
-		for _, dep := range p.Dependencies {
-			// remove environment markers (after semicolon)
-			dep = strings.Split(dep, ";")[0]
-			dep = strings.TrimSpace(dep)
-			if dep != "" {
-				deps = append(deps, dep)
+		// Base package (no extras field or empty extras)
+		if len(p.Extras) == 0 {
+			entry.Summary = p.Summary
```
spiffcs, line 116: `p` here is `pdmLockPackage` and `entry` is `pkg.PythonPdmLockEntry`
Should we also be preserving `Marker` from `pdmLockPackage` on this `entry`?

I might have missed if it was preserved elsewhere but I saw in the PR description:
```
preserves full dependency strings including environment markers 
(e.g., ; python_version < "3").... these were being stripped
```

I did not see entry.Marker getting the value brought over, but again might have been an oversight when reading ☹️

wagoodman: good catch! I'll incorporate a fix here

PR #4375, opened 2025-11-17
Detect embedded deps.json in .NET binaries

### LINKED ISSUE #4344: syft isn’t picking up embedded deps.json metadata in single-file .NET exec

Modern .net app, you can publish your app as a single self-contained file using PublishSingleFile, it packs everything inside including deps.json.

**Why is this needed**:

syft doesn't pick up the embedded deps.json in single-file executables, so it misses dependency info for .net apps built this way.

**Additional context**:

t deps.json file is baked into the singlefile bundle as a bin resource, so to get it out you gotta read the executables resource section and parse the data, adding this would let syft report deps more accurately for modern .net apps

build with

```
  --self-contained true \
  -p:PublishSingleFile=true 
```

rezmoss:
# Description

detect embedded deps.json metadata in single-file .NET exec

- Fixes #4344

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/internal/pe/bundle.go
+++ b/syft/pkg/cataloger/internal/pe/bundle.go
@@ -10,18 +10,12 @@ import (
 )
 
 // dotNetBundleSignature is the SHA-256 hash of ".net core bundle" used to identify single-file bundles.
-// this marker is located at the last 32 bytes of a .NET single-file bundle.
 var dotNetBundleSignature = []byte{
 	0x8b, 0x12, 0x02, 0xb9, 0x6a, 0x61, 0x20, 0x38,
 	0x72, 0x7b, 0x93, 0x02, 0x14, 0xd7, 0xa0, 0x32,
 	0x13, 0xf5, 0xb9, 0xe6, 0xef, 0xae, 0x33, 0x18,
 	0xee, 0x3b, 0x2d, 0xce, 0x24, 0xb3, 0x6a, 0xae,
 }
-
-const (
-	// bundleMarkerSize is the size of the bundle marker at the end of a .NET single-file bundle (8 byte offset + 32 byte signature)
-	bundleMarkerSize = 40
-)
 
 // dotNetBundleHeader represents the fixed portion of the bundle header (version 1+)
 type dotNetBundleHeader struct {
--- a/syft/pkg/cataloger/internal/pe/bundle_test.go
+++ b/syft/pkg/cataloger/internal/pe/bundle_test.go
@@ -1,0 +1,58 @@
+package pe
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_extractDepsJSONFromBundle_Versions(t *testing.T) {
+	tests := []struct {
+		name            string
+		fixture         string
+		path            string
+		wantDepsJSON    bool   // true if deps.json should be found
+		wantJSONContain string // string that should be in the JSON (varies by .NET version)
+	}{
+		{
+			name:            "V1 bundle (.NET Core 3.1)",
+			fixture:         "image-dotnet31-single-file",
+			path:            "/app/hello.exe",
+			wantDepsJSON:    true,
+			wantJSONContain: "runtimeOptions", // .NET Core 3.1 uses runtimeOptions
+		},
+		{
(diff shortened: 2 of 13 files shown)


syft/pkg/cataloger/dotnet/test-fixtures/embedded-deps/dotnet-single-file-embed-deps-json.exe
wagoodman, line 1: we shouldn't be checking in binaries unless absolutely necessary. We can follow the same pattern we're doing for other .net test fixtures (build and extract)

syft/pkg/cataloger/dotnet/deps_binary_cataloger.go
```diff
@@ -473,8 +479,23 @@ func readPEFile(resolver file.Resolver, loc file.Location) (*logicalPE, error) {
 	}
 	defer internal.CloseAndLogError(reader, loc.RealPath)
 
-	ldpe, err := readLogicalPE(file.NewLocationReadCloser(loc, reader))
+	data, err := io.ReadAll(reader)
 	if err != nil {
+		return nil, unknown.New(loc, fmt.Errorf("unable to read file data: %w", err))
+	}
+
+	embeddedJSON := extractEmbeddedDepsJSONFromBytes(data)
```
wagoodman: more ideally this would be in the internal pe utils that are referenced in this cataloger

syft/pkg/cataloger/dotnet/deps_binary_embed_deb_test.go
```diff
@@ -0,0 +1,21 @@
+package dotnet
+
+import (
+	"testing"
+
+	"github.com/anchore/syft/syft/pkg/cataloger/internal/pkgtest"
+)
+
+func Test_EmbeddedDepsExtraction(t *testing.T) {
```
wagoodman: the checked in binary is not complete (not a valid PE file) yet we still pick up on deps.json resources. This is probably ok, but not ideal.

syft/pkg/cataloger/dotnet/pe.go
```diff
@@ -12,6 +15,8 @@ type logicalPE struct {
 	// TargetPath is the path is the deps.json target entry. This is not present in the PE file
 	// but instead is used in downstream processing to track associations between the PE file and the deps.json file.
 	TargetPath string
+
+	EmbeddedDepsJSON string
```
wagoodman: this could get moved to pe.File

syft/pkg/cataloger/dotnet/pe.go
```diff
@@ -28,3 +33,37 @@ func readLogicalPE(reader file.LocationReadCloser) (*logicalPE, error) {
 		File: *peFile,
 	}, nil
 }
+
+func extractEmbeddedDepsJSONFromBytes(data []byte) string {
```
wagoodman: This is a good start, however, we might want to find the bundle header/manifest to search for where in the binary the deps.json offset is. I need to do a little more research on the details here.

syft/pkg/cataloger/dotnet/deps_binary_embed_deb_test.go
wagoodman: I think we should leverage the existing single-file test fixture instead of adding a new one. Infact, I'm surprised that a test update was not needed with this additional functionality. There might be a bug somewhere (or I'm missing something).

wagoodman (commented):
This is a great start, but there is more to do. The approach is a good PoC, but I think we should be subsetting what we are searching within the binary for a more fool-proof search (for instance, what if there is a dep.json-like content in other bundled files?). 

I'm going to push up adjustments here to this branch in a little bit -- I'll shout out on the progress in a bit.

syft/pkg/cataloger/dotnet/pe.go
rezmoss: found some info https://github.com/0xced/SingleFileAppDependencyContext and https://github.com/dotnet/runtime/blob/main/src/installer/managed/Microsoft.NET.HostModel/Bundle/Manifest.cs and https://github.com/dotnet/designs/blob/main/accepted/2020/single-file/bundler.md 
 maybe i can work on that tmr using the same approach

wagoodman: really nice find on the 0xced material!

PR #4378, opened 2025-11-18
feat: add SBOM authors flag

### LINKED ISSUE #2876: Option in parameter or configuration to set value in metadata > authors in SBOM (CycloneDX)
Parameter or configuration to add and set metadata > authors > name / email in SBOM in CycloneDX format.
I was actually looking for this ability in the parameter and configuration options but did not find any.  Please let me know if I am wrong.

**Why is this needed**:
To automatically include required metadata in CycloneDX format when generating SBOM.

**Additional context**:

Based on CycloneDX schema, it should look like this:

"metadata": {
	"authors":{
		"name": "Name"
		"email": "email@email.com"
	}
}

derbauer97:
# Description

This PR implements the `--authors` flag to allow users to specify custom SBOM authors for all output formats (SPDX, CycloneDX, and Syft-JSON). This addresses the need for organizations to identify themselves and their tools in generated SBOMs, which is important for SBOM provenance and tooling attribution.

- Fixes #2876

## Implementation Details

- **CLI Flag**: Added `--authors` flag that accepts multiple author entries in the format `type:name:email`
  - Valid types: `Person`, `Organization`, `Tool`
  - Example: `--authors "Person:John Doe:john@example.com"`

### Default behavior (SPDX only has default authors):
```bash
# SPDX format - has defaults
$ syft alpine:latest -o spdx-json -q | jq '.creationInfo.creators'
[
  "Organization: Anchore, Inc",
  "Tool: syft-[version]"
]

# CycloneDX format - no defaults
$ syft alpine:latest -o cyclonedx-json -q | jq '.metadata.authors'
null

# Syft-JSON format - no defaults
$ syft alpine:latest -o syft-json -q | jq '.source.authors'
null
```

### With custom authors:
```bash
# SPDX format
$ syft alpine:latest \
  --authors "Person:John Doe:john@example.com" \
  --authors "Organization:ACME Corp:info@acme.com" \
  -o spdx-json -q | jq '.creationInfo.creators'
[
  "Person: John Doe (john@example.com)",
  "Organization: ACME Corp (info@acme.com)"
]

# CycloneDX format
$ syft alpine:latest \
  --authors "Person:Jane Smith:jane@example.com" \
  -o cyclonedx-json -q | jq '.metadata.authors'
[
  {
    "name": "Jane Smith",
    "email": "jane@example.com"
  }
]

# Syft-JSON format
$ syft alpine:latest \
  --authors "Tool:Custom Scanner:v2.0" \
  -o syft-json -q | jq '.source.authors'
[
  {
    "Name": "Custom Scanner",
    "Email": "v2.0",
    "Type": "Tool"
  }
]
```

## Question

I noticed that only the SPDX format has default authors (`Organization: Anchore, Inc` and `Tool: syft-[version]`), while CycloneDX and Syft-JSON formats have no default authors. Is this intentional behavior? Should all formats have consistent defaults, or is there a specification reason for this difference?

Also only SPDX expects to add an type to the creators info: https://spdx.github.io/spdx-spec/v2.3/document-creation-information/ should we also cover this?

## Type of change

# Checklist:


cmd/syft/internal/options/catalog.go
```diff
@@ -263,6 +263,9 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
 
 	flags.StringVarP(&cfg.Source.Supplier, "source-supplier", "",
 		"the organization that supplied the component, which often may be the manufacturer, distributor, or repackager")
+
+	flags.StringArrayVarP(&cfg.Source.Authors, "authors", "",
```
kzantow: Sorry for the delay getting back to you with some feedback, due to holidays and time off it took a while to work out some of the details with the team.

I, personally, think this is part of a larger feature: specifying arbitrary SBOM data that Syft doesn't produce. But we've decided that we should probably kick that can down the road until Syft 2.0 if we decide to make changes to the way any of the current `--source-name` and these additional flags work, but for now adding flags for these purposeful new 2 features makes the most sense.

As for _this_ PR, there are a few things I think we should change, assuming I understand what's being implemented correctly. Am I correct in saying the _author_ is intended to specify the _author of the SBOM_, rather than _author of the source_ (the source being "thing that was scanned" -- container, directory, etc.)? I'm assuming so, in which case I think Henry's PR is closer to the end result and we should get them aligned by adding the `Authors` to the `SBOM`, rather than the _source_, and similarly use an appropriate flag: `--sbom-author`.

For both of these PRs I think we should add exported properties directly on the [SBOM](https://github.com/anchore/syft/blob/main/syft/sbom/sbom.go#L15) for `Authors` and the arbitrary `Properties`. I don't see any other spot where it makes sense to put them. And because these aren't really involved in cataloging, I don't think it's useful to add _cataloging configuration_ to set these -- it's just unnecessary and possibly confusing indirection. Instead, they should be set directly on the SBOM during decoding by the format decoders, or by the CLI scan command. If we determine it's important to have this in the configuration that gets passed to CreateSBOM, it can be added later, but we can't _remove it_ after we add it.

As far as the CLI goes, instead of piggybacking on the source config, let's introduce a new top-level `sbomConfig` (similar to the [sourceConfig](https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/source.go)) where the authors and properties will be set to more accurately reflect how this data is used, and it should live [alongside the cataloging config here](https://github.com/anchore/syft/blob/main/cmd/syft/internal/commands/scan.go#L70C61-L71C61) (with the json/yaml/mapstructure key `sbom`, so syft config makes sense under `sbom:`) like:
```
type actor struct {
  Type string `json:"type" yaml:"type" mapstructure:"type"`
  Name string `json:"name" yaml:"name" mapstructure:"name"`
  Email string `json:"email" yaml:"email" mapstructure:"email"`
}

type sbomConfig struct {
  Authors []actor `json:"authors" yaml:"authors" mapstructure:"authors"`
  authors []string // used for CLI input
  Properties map[string]string `json:"properties" yaml:"properties" mapstructure:"properties"`
  properties []string // used for CLI input
}

func (c *sbomConfig) PostLoad() error {
  // to support env vars e.g. SYFT_SBOM_AUTHOR_NAME, we need custom env var lookups see: https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/registry.go#L35
  // or maybe SYFT_SBOM_AUTHORS using a flattened name=value list as described below
  // do the custom CLI parsing of strings here, we could accept JSON in addition to any more bespoke format
}
```

By having this configuration structure, the yaml configuration will work as expected, so the last remaining bit is exactly how this gets specified for the flags. I've given a suggestion above, keeping the bespoke `name=value` that we've adopted elsewhere and for the properties. I think we should be careful about introducing bespoke parsing, like the `<type>:<name>:<value>` -- the only other spot that I'm aware we use `:` is in a URI-like spot, where the preceding value is the _scheme_, whereas this is just a way to split values. Maybe think about adopting a `name=value` like we use in other spots and the properties also should probably use, e.g. `--sbom-author type=person,name=TheName,email=e@mail` (though using a comma is another pattern we have that is equivalent to multiple flags, e.g. `--from docker,registry` == `--from docker --from registry`, and using the [Flatten](https://github.com/anchore/syft/blob/b0c74d4104c76dafa25c3494910ed6b743fb54b1/cmd/syft/internal/options/catalog.go#L285-L290) function, it could also make properties easier e.g. `--sbom-property name1=value1,name2=value2`, so consider a different separator character. Sorry I don't have a great answer here, but we can finalize this detail fairly quickly with the whole team towards the end; I would lean towards `&`, but it needs to be escaped in shells, so maybe `+`?, definitely open to suggestions here).

Again, apologies for the delay and I hope this is understandable enough!

derbauer97:
Hi @kzantow,

Sorry for the delay - our priorities shifted at the end of the year. Thanks for the detailed implementation plan and feedback, it was extremely helpful!

We've implemented the feature following your suggestions:

- Created `SBOMConfig` struct with `Authors` field (alongside cataloging config with yaml key `sbom`)
- Used `--sbom-author` flag (not `--authors` on source)
- Added type validation (person, organization, tool only)

The implementation handles CycloneDX and SPDX differently based on author type:
- **CycloneDX**: `person` → `metadata.authors`, `organization` → `metadata.supplier`, `tool` → `metadata.tools.components`
- **SPDX**: All authors → `creationInfo.creators` with type prefixes (per SPDX 2.3 spec)

For the separator, we went with `&` as you suggested. Since names with spaces require quotes anyway, the shell escaping isn't an additional burden and `&` provides good readability.

**Example usage:**

```bash
# Single author
syft scan alpine:latest -o spdx-json \
  --sbom-author "type=person&name=Max Mustermann&email=max@example.com" \
  --sbom-author "type=organization&name=Example Company" \
  --sbom-author "type=tool&name=syft"

# Multiple authors with different types
syft scan alpine:latest -o cyclonedx-json \
  --sbom-author "type=person&name=Max Mustermann&email=max@example.com" \
  --sbom-author "type=organization&name=Example Company" \
  --sbom-author "type=tool&name=syft"
```

**Expected CycloneDX output:**
```json
{
  "metadata": {
    "authors": [
      {"name": "Max Mustermann", "email": "max@example.com"}
    ],
    "supplier": {
      "name": "Example Company"
    },
    "tools": {
      "components": [
        {"type": "application", "name": "syft"}
      ]
    }
  }
}
```

**Expected SPDX output:**
```json
{
  "creationInfo": {
    "creators": [
      "Person: Max Mustermann (max@example.com)",
      "Organization: Example Company",
      "Tool: syft"
    ]
  }
}
```

We haven't added environment variable support yet (`SYFT_SBOM_AUTHORS`) because `PostLoad` is called multiple times in the configuration chain, which causes duplication. We can add it with deduplication logic if that's important.

We focused on implementing the authors feature first. If this approach looks good, we'd be happy to add the properties feature in the other PR.

Let us know if you'd like any adjustments!

derbauer97:
Hey @spiffcs @kzantow,

any thoughts on this? We would like to push this forward. I am happy to add improvements based on your feedback :)

PR #4383, opened 2025-11-18
chore: drop cpe from gguf

spiffcs:
# Description
This drops the CPE generation from `pkg.Model` type

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/internal/cpegenerate/generate.go
+++ b/syft/pkg/cataloger/internal/cpegenerate/generate.go
@@ -98,17 +98,17 @@ func FromDictionaryFind(p pkg.Package) ([]cpe.CPE, bool) {
 	case pkg.WordpressPluginPkg:
 		metadata, valid := p.Metadata.(pkg.WordpressPluginEntry)
 		if !valid {
-			return []cpe.CPE{}, false
+			return nil, false
 		}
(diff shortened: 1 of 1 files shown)


syft/pkg/cataloger/internal/cpegenerate/generate.go
```diff
@@ -62,16 +62,13 @@ func GetIndexedDictionary() (_ *dictionary.Indexed, err error) {
 
 func FromDictionaryFind(p pkg.Package) ([]cpe.CPE, bool) {
 	dict, err := GetIndexedDictionary()
-	parsedCPEs := []cpe.CPE{}
 	if err != nil {
 		log.Debugf("CPE dictionary lookup not available: %+v", err)
-		return parsedCPEs, false
+		return []cpe.CPE{}, false
```
wagoodman, line 67: nit: this can be `nil`, I see that it was allocated to begin with, but since your changing much of this do we still need to allocate? (we probably never did, but I haven't checked)

spiffcs: It was allocated to begin with -- I was just trying to shave some lines for the linter. I can check the callers and see if we can just get away with this being `nil`.

Issue #4394, opened 2025-11-20
Syft should not define subpaths by default in PURLs


JArndt-WI:
**What happened**:

I process a go.mod with a module statement like this:

`module dev.azure.com/ab-cde/ABC/_git/repo-name.git/abcdef`

A `Syft-JSON` SBOM is generated, the PURL is

`pkg:golang/dev.azure.com/ab-cde/ABC#_git/repo-name.git/abcdef`

The general part of the purl-spec defines subpath as '_Subpath within a package, relative to the package root_', '_The subpath must be interpreted as relative to the root of the package_'. 

**What you expected to happen**:

The PURL should be constructed without subpath ('#'). 

Note from purl-spec: _The current definition predates Go modules and has several practical problems, and in particular it is impossible to determine what is a module and what is a package short of having full access to the source code or making an API call to the Go module proxy._

**Steps to reproduce the issue**:

Create a go.mod with an Azure Repo compatible module statement and a fictive module having the same leading path + "/other_module" as given above an scan it.

**Anything else we need to know?**:
N/A

**Environment**:
- Output of `syft version`:
Windows (Win 11)
Application:   syft
Version:       1.38.0
BuildDate:     2025-11-17T17:42:49Z
GitCommit:     a033ae525f6c7ef937c6f49513e3403f07a1d6c0
GitDescription: v1.38.0
Platform:      windows/amd64
GoVersion:     go1.25.4
Compiler:      gc
SchemaVersion: 16.1.0

Linux (Debian wsl2)
Application:   syft
Version:       1.38.0
BuildDate:     2025-11-17T17:42:49Z
GitCommit:     a033ae525f6c7ef937c6f49513e3403f07a1d6c0
GitDescription: v1.38.0
Platform:      linux/amd64
GoVersion:     go1.25.4
Compiler:      gc
SchemaVersion: 16.1.0

willmurphyscode:
Hi @JArndt-WI - this seems very similar to https://github.com/anchore/syft/issues/4316, but that issue is scoped only to non-zero major versions, instead of being scoped to all subpath generation, so they're not really duplicates. (It also looks like the [pr meant to address that issue](https://github.com/anchore/syft/pull/4395) will not fix this issue.)

> Create a go.mod with an Azure Repo compatible module statement and a fictive module having the same leading path + "/other_module" as given above an scan it.

[...]

PR #4395, opened 2025-11-20
fix: golang PURL should include full module

### LINKED ISSUE #4316: Incorrect Go module PURL generation when module path contains /vN (e.g. /v5)
**Description**:

When generating a BOM for Go projects, Syft incorrectly parses module paths that include a version suffix like /v5 as part of the semantic version instead of the package path.

**Example:**
We use the Go module:
`github.com/golang-jwt/jwt/v5 v5.2.1
`
In the generated bom.json, Syft produces:
```
"purl": "pkg:golang/github.com/golang-jwt/jwt@v5.2.1#v5",
"bom-ref": "pkg:golang/github.com/golang-jwt/jwt@v5.2.1#v5"
```
The correct PURL should be:
`"purl": "pkg:golang/github.com/golang-jwt/jwt/v5@v5.2.1"
`

**Impact:**

- Dependency Track tools misidentify the dependency : 

Image

- The used version is not recognized correctly.
- Associated vulnerabilities and metadata are missing.

**Environment**

- Syft version:  v1.33.0
- Language: Go
- Example package: github.com/golang-jwt/jwt/v5@v5.2.1

**Expected Behavior**
Syft should preserve the module path (/v5) and produce:
`pkg:golang/github.com/golang-jwt/jwt/v5@v5.2.1
`

**Actual Behavior**
Syft outputs:
`pkg:golang/github.com/golang-jwt/jwt@v5.2.1#v5
`

rezmoss:
# Description

fixed wrong go module purl when module paths had version suffixes like /v5

- Fixes #4316 

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/golang/package.go
+++ b/syft/pkg/cataloger/golang/package.go
@@ -64,27 +64,23 @@ func newBinaryMetadata(dep *debug.Module, mainModule, goVersion, architecture st
 func packageURL(moduleName, moduleVersion string) string {
 	// source: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang
 	// note: "The version is often empty when a commit is not specified and should be the commit in most cases when available."
-
-	fields := strings.Split(moduleName, "/")
-	if len(fields) == 0 {
+	if moduleName == "" {
 		return ""
 	}
 
 	namespace := ""
-	name := ""
-	// The subpath is used to point to a subpath inside a package (e.g. pkg:golang/google.golang.org/genproto#googleapis/api/annotations)
-	subpath := ""
+	name := moduleName
 
-	switch len(fields) {
-	case 1:
-		name = fields[0]
-	case 2:
-		name = fields[1]
-		namespace = fields[0]
-	default:
-		name = fields[2]
-		namespace = strings.Join(fields[0:2], "/")
-		subpath = strings.Join(fields[3:], "/")
+	// golang PURLs from _modules_ are constructed as pkg:golang/<module>@version, where
+	// the full module name often includes multiple segments including `/v#`-type versions, for example:
+	//  pkg:golang/github.com/cli/cli/v2@2.63.0
+	// see: https://github.com/package-url/purl-spec/issues/63
+	// and: https://github.com/package-url/purl-spec/blob/main/types-doc/golang-definition.md#subpath-definition
+	// by setting the namespace this way, it does not escape the slashes:
+	lastSlash := strings.LastIndex(moduleName, "/")
+	if lastSlash > 0 && lastSlash < len(moduleName)-1 {
+		name = moduleName[lastSlash+1:]
+		namespace = moduleName[0:lastSlash]
 	}
 
 	return packageurl.NewPackageURL(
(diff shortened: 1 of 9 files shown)


syft/pkg/cataloger/golang/package.go
```diff
@@ -82,9 +82,28 @@ func packageURL(moduleName, moduleVersion string) string {
...
 		namespace = fields[0]
 	default:
-		name = fields[2]
-		namespace = strings.Join(fields[0:2], "/")
-		subpath = strings.Join(fields[3:], "/")
+		lastField := fields[len(fields)-1]
+		isVersionSuffix := strings.HasPrefix(lastField, "v") && len(lastField) > 1
+		if isVersionSuffix {
+			for _, r := range lastField[1:] {
+				if r < '0' || r > '9' {
```
kzantow: Would this be slightly more clear with `unicode.IsDigit(r)`?

rezmoss: unicode.IsDigit() matches all kinds of digits like arabic/etc, not just 0-9, for ver suffixes we only want ascii digits, right?

kzantow (dismissed):
I did some digging into this, are you seeing this issue when scanning source code with `go.mod` or Go binaries?

I think the `go.mod` case should have a different fix, and I think the go binary case could also have a slightly different fix. I assume we are, in fact, using the packages library now to determine the `go.mod` case, which should be returning the correct modules, that match the `go.mod`, and similarly should just use those directly rather than being split to subpaths at all.

For `go.mod`, I think we should be using the full value we found in the `go.mod` or `go.sum`, not processing it to have any subpaths -- these _are modules_ already. The subpath from the PURL spec is "used to point to a subpath inside a **package**" -- _package_, not module.

~~For the binary case, we could think about extending the [online license lookup](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/golang/licenses.go#L162-L173) to include identification of the module paths (not asking for this as part of your PR here), but default to a slightly different behavior that looks for a `/v#` _anywhere within the path_ before falling back to `<host>/<segment1>/<segment2>`, though a package with `/v#` doesn't necessarily mean the _module_ has that, we need the `go.mod` to be able to tell for sure, though I suspect most modules are good citizens and update when using this versioning scheme.~~ EDIT: I was conflating this with something else we're working on, but today with binaries, we also get _module names_, not packages, and these should similarly be used directly. At such time as we surface _packages_, we will have some additional metadata that may be output in a PURL subpath somewhere, but for all current processing we should use module names directly.

rezmoss:
@kzantow t main issue the user reported in https://github.com/anchore/syft/issues/4316 was about go.mod, not the go binary

t purl returned something like "pkg:golang/github.com/golang-jwt/jwt@v5.3.0#v5" but it should be "pkg:golang/github.com/golang-jwt/jwt/v5@v5.2.1"

and this is the output from my implementation

Before

b2


After

b3


to reproduce, just create a go.mod file

```
module tst_gomod

go 1.25.3

require github.com/golang-jwt/jwt/v5 v5.3.0 // indirect
```

and scan

kzantow:
@rezmoss right, read [the comment I left on the issue](https://github.com/anchore/syft/issues/4394#issuecomment-3607574838) -- I think when reading `go.mod` that we should be using the module names directly from that file, not looking for `v#` or any special processing, really. If we did that, it would solve the problem you mentioned here as well as the other, related issue.

rezmoss:
@kzantow added new func to handle modules straight from go.mod, no processing needed, kept the original packageurl() for binary scan support, jwt/v5 and aws sdk edge cases are now handled

kzantow:
I believe this PR will also partially fix the following Grype issues:
* https://github.com/anchore/grype/issues/3095
* https://github.com/anchore/grype/issues/2838

PR #4408, opened 2025-11-24
fix: normalize python package names from dependency lists

### LINKED ISSUE #4401: Ssyft: The dependency graph does not include all Requires-Dist relationships defined in the package’s METADATA file
**What happened**:
When generating a CycloneDX SBOM for Python packages using Syft (1.38.0), the dependency graph does not include all Requires-Dist relationships from the package’s METADATA file.
Example: For Flask 1.1.2, Syft only reports two dependencies (click, itsdangerous) but omits others (Jinja2, Werkzeug), even though all four are declared in the METADATA and installed in the environment.
Snippet from the SBOM:

```

dependencies": [
  {
    "ref": "pkg:pypi/flask@1.1.2?package-id=7c9599e2d82779f5",
    "dependsOn": [
      "pkg:pypi/click@8.3.1?package-id=f5afd6ba55e9b798",
      "pkg:pypi/itsdangerous@2.2.0?package-id=1b2e20a8cd0be3ca"
    ]
  },
  {
    "ref": "pkg:pypi/importlib-metadata@8.0.0?package-id=587e5d030197d8be"
  }
]
```
Only 2 dependencies are listed for Flask instead of 4.

Syft should include all dependencies declared in the METADATA file for Flask:
```
Requires-Dist: Werkzeug (>=0.15)
Requires-Dist: Jinja2 (>=2.10.1)
Requires-Dist: itsdangerous (>=0.24)
Requires-Dist: click (>=5.1)
```

```
# Start a Python 3.11 container
docker run -it --name py-env python:3.11 bash

# Install Syft
curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin

# Create an app directory and requirements file
mkdir app && cd app
echo -e "Flask==1.1.2\nJinja2==2.11.3" > requirements.txt

# Create virtual environment and install dependencies
python3.11 -m venv .venv
source .venv/bin/activate
pip install -r requirements.txt

# Generate SBOM
syft dir:.venv -o cyclonedx-json > sbom.json
```

Motivation: Missing dependencies lead to false negatives in vulnerability detection.
Example: Grype reports three vulnerabilities for Jinja2, including:
GHSA-h75v-3vvj-5mfj (HTML attribute injection)
CVE-2024-34064 (XSS risk)

If Syft omits Jinja2 from Flask’s dependency graph, these vulnerabilities will not be associated with Flask in SBOM-based scans.

Impact: SBOM consumers relying on Syft for Python dependency graphs may miss critical transitive dependencies, affecting vulnerability scanning and compliance checks.
Although this version is old, it's necessary to capture this relation, this will help researchers and tools that use aim to reproduce CVEs and vulnerabilities. 

Even open-rouce dependency projects such as deps.dev can capture this dependency, (though the version is wrong)

Image

[sbom_venv_new_syft.json](https://github.com/user-attachments/files/23719689/sbom_venv_new_syft.json)
[vulns.json](https://github.com/user-attachments/files/23719688/vulns.json)

willmurphyscode:
Because package names in METADATA files may have upper case like Werkzeug or Jinja2, but Syft artifacts have normalized names and are lower case, like werkzeug or jinja2, Syft would miss emitting dependency relationships. Therefore, normalize dependency names before comparing with existing artifacts.

# Description

Syft's cataloging of venv/site-packages uses a dependency resolving post processor to compare the `requires` statements from the METADATA file to other python artifacts emitted and create relationships between them. However, the name comparison between the requires statements from the METADATA files was normalized via a different codepath than the artifact names, so there were missed relationships because, for example `"Werkzeug" != "werkzeug"`.

Therefore, normalize the dependency names via the same normalization used for artifact names before building the relationships.

Fixes #4401 

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/python/cataloger_test.go
+++ b/syft/pkg/cataloger/python/cataloger_test.go
@@ -567,6 +567,7 @@ func Test_PackageCataloger_Relationships(t *testing.T) {
 				"colorama @ 0.4.6 (.) [dependency-of] pygments @ 2.18.0 (.)",
 				"colorama @ 0.4.6 (.) [dependency-of] uvicorn @ 0.29.0 (.)", // proof of uvicorn[standard]
 				"dnspython @ 2.6.1 (.) [dependency-of] email-validator @ 2.1.1 (.)",
+				"email-validator @ 2.1.1 (.) [dependency-of] fastapi @ 0.111.0 (.)",
 				"email-validator @ 2.1.1 (.) [dependency-of] pydantic @ 2.7.1 (.)",
 				"fastapi @ 0.111.0 (.) [dependency-of] fastapi-cli @ 0.0.4 (.)",
 				"fastapi-cli @ 0.0.4 (.) [dependency-of] fastapi @ 0.111.0 (.)",
--- a/syft/pkg/cataloger/python/dependency.go
+++ b/syft/pkg/cataloger/python/dependency.go
@@ -181,7 +181,9 @@ func extractPackageName(s string) string {
 	// requests (>= 2.8.1)			--> requests
 	// requests ; python_version < "2.7"	--> requests
 
-	return strings.TrimSpace(internal.SplitAny(s, "[(<!=>~;")[0])
+	name := strings.TrimSpace(internal.SplitAny(s, "[(<!=>~;")[0])
+	// normalize the name to match how packages are stored (lowercase, with hyphens instead of underscores)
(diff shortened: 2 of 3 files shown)


spiffcs (approved):
I couldn't think of any other cases to add - the tests are great and thanks for the quick fix to a triage bug 😄

PR #4421, opened 2025-12-01
fix:handle compound aliases like ```.tgz``` when cataloging archives

### LINKED ISSUE #4416: v1.38.0 generates empty sbom for tgz sources
**What happened**:

I used `file:/path/to/my.tgz` as the source for syft to scan.

I expected to get an SBOM with data but instead, the SBOM is created but empty (no packages found - no data scanned).

Tar up a filesystem and throw it at syft. We use this yaml with `syft -c`:
```
file:
  metadata:
    selection: all
    digests:
      - sha256

output:
  - spdx-json

format:
    pretty: true

source:
  file:
    digests:
      - sha256

package:
  cataloger:
    scope:
      - squashed
```

An empty SBOM is not created with `v1.37.0`. That version generates a usable sbom from a tgz. Nothing in the [changelog](https://github.com/anchore/syft/releases/tag/v1.38.0) states this is no longer supported in v1.38.0. The same fs.tgz cannot be used with v1.38.0 but is fine with v1.37.0 and older.

This has kind of broken our build pipeline because we always tell syft to scan the artefact created by the build stage, whether it be a directory or an fs.tgz. Directory scanning is fine (as is `oci-archive:/path/to/oci.tar`) but scanning tar balls is broken.

**Environment**:
- Output of `syft version`:

syft 1.38.0 on Debian Bookworm

VictorHuu:
# Description

see: https://github.com/anchore/syft/issues/4416#issuecomment-3594688303

In the near future, it's the library ```archives``` who should fix the logic to support compound extension names like ```.tgz``` 

- Fixes #4416 

### Summary of magic bytes vs extension for archive detection:
update:
file_source.go:213 - switch to stream-based (already opens file later anyway)
tar_file_traversal.go:23 - opens the file on line 17, so could pass tarReader

defer detection and use new map:
unknowns_tasks.go:64 only has coords.RealPath, would need to open files (potential perf hit for many files)
model.go:159 isArchive() is a helper - opening files could be more expensive here

## Type of change

# Checklist:

### DIFF
--- a/internal/file/archive_aliases.go
+++ b/internal/file/archive_aliases.go
@@ -3,7 +3,6 @@ package file
 import (
 	"context"
 	"io"
-	"os"
 	"path/filepath"
 	"strings"
 
@@ -29,24 +28,10 @@ var compoundExtensionAliases = map[string]string{
 //
 // This function is a drop-in replacement for archives.Identify that centralizes
 // the compound alias handling logic in one place.
-func IdentifyArchive(ctx context.Context, path string) (archives.Format, io.Reader, error) {
+func IdentifyArchive(ctx context.Context, path string, r io.Reader) (archives.Format, io.Reader, error) {
 	// First, try to identify using the alias-mapped path (filename-based detection)
 	normalizedPath := handleCompoundArchiveAliases(path)
-	format, outReader, err := archives.Identify(ctx, normalizedPath, nil)
-	if err == nil && format != nil {
-		return format, outReader, nil
-	}
-
-	// If filename-based detection failed,
-	// try opening the file for content-based detection
-	f, openErr := os.Open(path)
-	if openErr != nil {
-		// Return the original error from archives.Identify
-		return format, outReader, err
-	}
-	defer f.Close()
-
-	return archives.Identify(ctx, path, f)
+	return archives.Identify(ctx, normalizedPath, r)
 }
 
 // handleCompoundArchiveAliases normalizes archive file paths that use compound extension
--- a/syft/source/filesource/file_source.go
+++ b/syft/source/filesource/file_source.go
@@ -207,7 +207,7 @@ func fileAnalysisPath(path string, skipExtractArchive bool) (string, func() erro
 		return analysisPath, cleanupFn, nil
 	}
 
-	envelopedUnarchiver, _, err := intFile.IdentifyArchive(context.Background(), path)
+	envelopedUnarchiver, _, err := intFile.IdentifyArchive(context.Background(), path, nil)
 	if unarchiver, ok := envelopedUnarchiver.(archives.Extractor); err == nil && ok {
 		analysisPath, cleanupFn, err = unarchiveToTmp(path, unarchiver)
 		if err != nil {
--- a/internal/file/archive_aliases_test.go
+++ b/internal/file/archive_aliases_test.go
@@ -66,7 +66,7 @@ func TestHandleCompoundArchiveAliases(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			result := HandleCompoundArchiveAliases(tt.input)
+			result := handleCompoundArchiveAliases(tt.input)
 			assert.Equal(t, tt.expected, result)
 		})
 	}
--- a/internal/file/tar_file_traversal.go
+++ b/internal/file/tar_file_traversal.go
@@ -20,7 +20,7 @@ func TraverseFilesInTar(ctx context.Context, archivePath string, visitor archive
 	}
 	defer internal.CloseAndLogError(tarReader, archivePath)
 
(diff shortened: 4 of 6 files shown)


spiffcs:
Wow TY so much @VictorHuu for the PR. I think there are some other sections where we use Identify and want this fix. I'm going to push some changes here and we'll use this branch as the fix.

spiffcs:
I'm also going to quickly explore an approach suggested by @willmurphyscode which would pass the reader. This might give us better automatic detection here where we don't have to implement/maintain this map 😄

spiffcs:
Looks like I need to check the CLI tests now with the new split approach

internal/file/archive_aliases.go
```diff
@@ -0,0 +1,69 @@
...
+		return strings.TrimSuffix(path, ext) + newExt
+	}
+	return path
+}
+
+// HandleCompoundArchiveAliases is the exported version for cases where only
+// path normalization is needed without identification.
+//
+// Deprecated: Prefer using IdentifyArchive which handles aliases automatically.
+func HandleCompoundArchiveAliases(path string) string {
```
willmurphyscode: Doo we still need this to be exported / exist?

internal/file/archive_aliases.go
```diff
@@ -0,0 +1,69 @@
...
+	".tzst": ".tar.zst",
+}
+
+// IdentifyArchive is a wrapper around archives.Identify that handles compound extension
+// aliases (like .tgz -> .tar.gz) transparently. It first attempts filename-based detection
+// using the alias map, and falls back to content-based detection if needed.
+//
+// This function is a drop-in replacement for archives.Identify that centralizes
+// the compound alias handling logic in one place.
+func IdentifyArchive(ctx context.Context, path string) (archives.Format, io.Reader, error) {
```
willmurphyscode: Previously, do we ever already have a file open when calling `archives.Identify`? Should this accept a reader rather than trying to open the file?

spiffcs: I think it's better here if the function handles opening and closing it's own content here. If we pass a reader and it gets used then there might be instances in the code where we expect it to still be available after the function is called and is no longer available. I can check to see if we don't do this, but I thought a more defensive approach would be to create/cleanup its own resources here.

internal/file/archive_aliases.go
```diff
@@ -0,0 +1,69 @@
...
+	// First, try to identify using the alias-mapped path (filename-based detection)
+	normalizedPath := handleCompoundArchiveAliases(path)
+	format, outReader, err := archives.Identify(ctx, normalizedPath, nil)
+	if err == nil && format != nil {
+		return format, outReader, nil
+	}
+
+	// If filename-based detection failed,
+	// try opening the file for content-based detection
+	f, openErr := os.Open(path)
```
willmurphyscode: I thought that falling back to reading bytes from the file was a behavior change we didn't want?

Also, do we always get here if the file is not an archive? That might be a lot of extra `os.Open` calls if so.

internal/file/archive_aliases.go
spiffcs: removed

syft/source/filesource/file_source.go
```diff
@@ -207,10 +207,16 @@ func fileAnalysisPath(path string, skipExtractArchive bool) (string, func() erro
 		return analysisPath, cleanupFn, nil
 	}
 
-	// if the given file is an archive (as indicated by the file extension and not MIME type) then unarchive it and
+	// if the given file is an archive (as indicated by magic bytes) then unarchive it and
 	// use the contents as the source. Note: this does NOT recursively unarchive contents, only the given path is
 	// unarchived.
-	envelopedUnarchiver, _, err := archives.Identify(context.Background(), path, nil)
+	f, err := os.Open(path)
```
willmurphyscode: It looks like here we're open a file so that we can pass it in to `archives.Identify`, but then we don't because our wrapper now only accepts a string. Do we still need to open this?

internal/file/archive_aliases.go
spiffcs: Oh - I might have misunderstood our sync. I thought we wanted to try the easy way `via extension` and have this as a fall through if that failed for any reason.

syft/source/filesource/file_source.go
spiffcs: yep this can be removed now - missed pulling it out when centralizing everything

internal/file/archive_aliases.go
willmurphyscode: See [comment below](https://github.com/anchore/syft/pull/4421/files#r2582805430) - I thought we had discussed that actually reading bytes from the file to check whether its an archive was a behavior change we didn't necessarily want in this bugfix release. 

I don't think it's correct to `os.Open` here, but I'm not clear on when we reach line 42.

spiffcs:
TY @VictorHuu for the help on this one

PR #4436, opened 2025-12-04
fixed dotnet-deps cataloger skip project type by def, fixed #4423

### LINKED ISSUE #4423: dotnet-deps cataloger should skip project references with type "project" when building the sbom
**What happened**:

Syft treats project refs from deps.json as if they’re nuget packages in the sbom, even though they’re first-party code, which leads to weird purls like pkg:nuget/myapp@1.0.0

by default sbom shouldn't include project references (type: "project") because they're part of the firstparty code

from the .net sdk runtime config spec

"package" is for nuget deps
"project" is for internal refs

I think by default, entries with type "project" must filtered out unless it's the root package, or even add use --include-project-references to include them

**Environment**:
- Output of `syft version`: latest main
- OS (e.g: `cat /etc/os-release` or similar): all

rezmoss:
# Description

Fixed https://github.com/anchore/syft/issues/4423 

- Fixes #4423 

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/cataloger/dotnet/cataloger_test.go
+++ b/syft/pkg/cataloger/dotnet/cataloger_test.go
@@ -516,6 +516,7 @@ func TestCataloger(t *testing.T) {
 				Path:        "newtonsoft.json/13.0.3",
 				Sha512:      "sha512-HrC5BXdl00IP9zeV+0Z848QWPAoCr9P3bDEZguI+gkLcBKAOxix/tLEAAHC+UvDNPv4a2d18lOReHMOagPa+zQ==",
 				HashPath:    "newtonsoft.json.13.0.3.nupkg.sha512",
+				Type:        "package",
 				Executables: nil, // important!
 			},
 		}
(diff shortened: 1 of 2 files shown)


cmd/syft/internal/options/catalog.go
```diff
@@ -263,6 +264,9 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
 
 	flags.StringVarP(&cfg.Source.Supplier, "source-supplier", "",
 		"the organization that supplied the component, which often may be the manufacturer, distributor, or repackager")
+
+	flags.BoolVarP(&cfg.Dotnet.ExcludeProjectReferences, "dotnet-exclude-project-references", "",
+		"exclude packages with type 'project' from deps.json output (default: true)")
```
kzantow: We should probably not add a flag for this, it seems like the right behavior to default and override via configuration or environment variable, if needed.

kzantow:
One comment about adding a flag, but also I think you need to regenerate a new schema? After that, this looks good to me 👍

spiffcs (approved):

PR #4437, opened 2025-12-05
added gitignore simple mode & exclude file glob patern , fixed #4026

### LINKED ISSUE #4026: [ feature request ] Ignore everything in .gitignore when generating SBOMs

Ignore everything in the `.gitignore` file when generating SBOMs.

**Why is this needed**:

I need to ignore the `target` directory. I could delete it, but this would be easier.

rezmoss:
# Description

added gitignore simple mode & exclude file glob patern , fixed #4026

- Fixes #4026 

## Type of change

# Checklist:


cmd/syft/internal/commands/scan.go
```diff
@@ -222,7 +222,9 @@ func getSource(ctx context.Context, opts *options.Catalog, userInput string, sou
 			Supplier: opts.Source.Supplier,
 		}).
 		WithExcludeConfig(source.ExcludeConfig{
-			Paths: opts.Exclusions,
+			Paths:                opts.Exclusions,
+			ExcludeFile:          opts.ExcludeFile,
+			ExcludeGitignoreMode: opts.ExcludeGitignoreMode,
```
kzantow, line 227: Would this be `ExcludeMode`, where `gitignore` is one option?

rezmoss: good idea, ExcludeMode for now something like  simple-gitignore , and we can add more modes later ?

kzantow: I don't know if we would add more modes, but it would be ok. We should also default to gitignore mode if we find the exclude file is `**/.gitignore`, probably. What is `simple-gitignore` -- just indicating that it isn't doing a full gitignore behavior? should we just call it gitignore and improve it over time, if needed?

rezmoss: something like --exclude-mode=gitignore auto checks for a .gitignore file and uses it to exclude files, for --exclude-mode=file, user need to pass a file with --exclude-file ./excludelist using glob patterns , then we could default to none or gitignore, whichever works better

kzantow: I mean, if you specify `--exclude-file=/some/path/to/.gitignore` but don't specify `--exclude-mode`, it could default to the `gitignore` mode, otherwise it defaults to a "standard" ignore mode that takes a file where each line matches how `--exclude` works today. It seems unlikely if I specify a `.gitignore` file, I would want the other matching mode.

rezmoss: then why do we even need --exclude-mode, can't we just use --exclude-file and if it's a .gitignore, handle it that way, otherwise treat each line like a glob pat

rezmoss: @kzantow any thought?

rezmoss:
@kzantow  this stalled on t design question from Dec, do we even need --exclude-mode, or should --exclude-file automatically detect .gitignore files & treat everything else as one glob per line? 
if you can confirm t preferred direction, i'll rework the pr & update the tests so we can finally close out #4026

rezmoss:
@kzantow is this something we can move forward with, or should we close it?

PR #4438, opened 2025-12-05
fix:best effort to get the os info of an ELF binary

### LINKED ISSUE #4384: ELF note cataloger does not pick up OS field, but should

ELF note cataloger picks up the package name, and package type, but not the OS distro vendor

**Why is this needed**:

Without picking up OS distro vendor, there is not automatic matching to the vendor remediation feeds to provide "Fixed In" values when scanning syft json with grype

**Additional context**:

In Wolfi, I have this elf note:

```
# readelf --notes ./lib/python3.13/site-packages/cryptography.libs/libcrypto-b943a01c.so.3 
Displaying notes found in: .note.package
  Owner                Data size 	Description
  FDO                  0x0000005c	FDO_PACKAGING_METADATA
    Packaging Metadata: {"type":"apk","os":"wolfi","name":"openssl","version":"3.6.0-r0","architecture":"x86_64"}
```

The current syft PURL generated for this apk is:

```
pkg:apk/openssl@3.6.0-r0
```

But ideally it should be

```
pkg:apk/wolfi/openssl@3.6.0-r0
```

Or well whatever the "os" field is, ditto for example for Fedora, CentOS, RHEL, AmazonLinux 2023 all of which have similar ELF build notes too.

VictorHuu:
# Description

This PR requires a json-schema label though the version of the schema may change due to the conflicts across PRs.

In the newly introduced test cases for the issue, the constructor of ```location``` needs to be altered.

```appCpe``` is just added according to the [document](https://systemd.io/PACKAGE_METADATA_FOR_EXECUTABLE_FILES/),whose usage in Syft remains uncertain.

- Fixes #4384 

## Type of change

# Checklist:

### DIFF
--- a/syft/pkg/binary.go
+++ b/syft/pkg/binary.go
@@ -24,13 +24,12 @@ type ELFBinaryPackageNoteJSONPayload struct {
 	Architecture string `json:"architecture,omitempty"`
 
 	// OSCPE is a CPE name for the OS, typically corresponding to CPE_NAME in os-release (e.g. cpe:/o:fedoraproject:fedora:33)
+	//
+	// Deprecated: in Syft 2.0 the struct tag will be corrected to `osCpe` to match the systemd spec casing.
 	OSCPE string `json:"osCPE,omitempty"`
 
-	// OsCpe is an alias for osCPE,see https://systemd.io/PACKAGE_METADATA_FOR_EXECUTABLE_FILES/
-	OsCpe string `json:"osCpe,omitempty"`
-
-	// APPCPE is a CPE name for the upstream Application, as found in NVD CPE search (e.g. cpe:2.3:a:gnu:coreutils:5.0)
-	APPCPE string `json:"appCpe,omitempty"`
+	// AppCpe is a CPE name for the upstream Application, as found in NVD CPE search (e.g. cpe:2.3:a:gnu:coreutils:5.0)
+	AppCpe string `json:"appCpe,omitempty"`
 
 	// OS is the OS name, typically corresponding to ID in os-release (e.g. "fedora")
 	OS string `json:"os,omitempty"`
--- a/syft/pkg/cataloger/binary/elf_package_cataloger.go
+++ b/syft/pkg/cataloger/binary/elf_package_cataloger.go
@@ -34,7 +34,10 @@ type elfBinaryPackageNotes struct {
 	CPE                                 string `json:"cpe"`
 	License                             string `json:"license"`
 	pkg.ELFBinaryPackageNoteJSONPayload `json:",inline"`
-	Location                            file.Location `json:"-"`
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: syft/pkg/binary.go [hunk 1]
```diff
@@ -24,7 +24,10 @@ type ELFBinaryPackageNoteJSONPayload struct {
 	Architecture string `json:"architecture,omitempty"`
 
 	// OSCPE is a CPE name for the OS, typically corresponding to CPE_NAME in os-release (e.g. cpe:/o:fedoraproject:fedora:33)
-	OSCPE string `json:"osCPE,omitempty"`
+	OSCPE string `json:"osCpe,omitempty"`
```
wagoodman: we should also no break current behavior -- adding a new field to https://github.com/anchore/syft/blob/47e1cee5a55b5d0ea63dc13a26b3d1acab2181e3/syft/pkg/cataloger/binary/elf_package_cataloger.go#L36 that also unmarshals the old `osCPE` field and reconciles that value to the `osCpe` (new) field would help here. Since this is a grey area instead of an addition schema-wise lets make this a revision (instead of `16.1.1` lets go to `16.2.0`)

syft/pkg/binary.go
wagoodman, line 1: I changed this to not be a breaking change for JSON consumers. We can correct the output in syft JSON with syft v2.

syft/pkg/cataloger/binary/elf_package_cataloger.go
```diff
@@ -34,7 +34,10 @@ type elfBinaryPackageNotes struct {
 	CPE                                 string `json:"cpe"`
 	License                             string `json:"license"`
 	pkg.ELFBinaryPackageNoteJSONPayload `json:",inline"`
-	Location                            file.Location `json:"-"`
+	// CorrectOSCPE has the corrected casing for the osCPE field relative to the systemd ELF package metadata "spec" https://systemd.io/ELF_PACKAGE_METADATA/ .
+	// Ideally in syft 2.0 this field should be replaced with the pkg.ELFBinaryPackageNoteJSONPayload.OSCPE field directly (with the struct tag corrected).
+	CorrectOSCPE string        `json:"osCpe,omitempty"`
```
wagoodman, line 39: Since we can't change the osCPE field on the `pkg.ELFBinaryPackageNoteJSONPayload` struct, this adds it to the struct responsible for unmarshaling. In this way we aren't exposing out the temporary field.

syft/pkg/cataloger/binary/elf_package_cataloger.go
```diff
@@ -186,6 +189,21 @@ func getELFNotes(r file.LocationReadCloser) (*elfBinaryPackageNotes, error) {
 	return nil, err
 }
 
+func unmarshalELFPackageNotesPayload(data []byte) (*elfBinaryPackageNotes, error) {
```
wagoodman, line 192: I made this function as a choke point as early on in processing to deal with capturing the "correct" osCPE field onto the public field.

PR #4472, opened 2025-12-12
Improve PR template

wagoodman:
This adds the following options to the PR template:
- config changes: what configuration options have changed
- output samples: if the effect of the PR can be seen on the CLI output, then provide a sample

Also tweaks a few other details (smaller headers, fix checklist header detail, etc)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,22 +2,9 @@
 
 <!-- Please include a summary of the changes along with any relevant motivation and context -->
 
-<!-- If this completes an issue, please include: -->
-- Fixes #<PR-number>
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,11 +1,23 @@
...
 
-Please include a summary of the changes along with any relevant motivation and context,
-or link to an issue where this is explained.
+<!-- Please include a summary of the changes along with any relevant motivation and context -->
 
 <!-- If this completes an issue, please include: -->
+- Fixes #<PR-number>
 
-- Fixes
+## Output Samples
```
kzantow: Maybe rather than headings that a user has to delete, just add more comments about things to include in the description?

PR #4487, opened 2025-12-19
fixed #4430 exclude dev pnpm pkg

### LINKED ISSUE #4430: Exclude development or test dependencies for PNPM Package type
We have file - pnpm-lock.yaml which is having all required and dev dependencies. Syft is giving SBOM which has all dependencies including required and dev dependencies without scope defined. Can we exclude dev and test dependencies and get only the actual one?

**Why is this needed**: 
To have clean and required dependencies for direct and transitive SBOM.

**Additional context**:

Example : Purl for PNPM = "pkg:npm/typed-array-byte-offset@1.0.4"

rezmoss:
## Description

Exclude development or test dependencies for PNPM Package type

## Type of change

## Checklist

## Issue references

- Fixes #4430

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
@@ -202,12 +202,12 @@ func (a genericPnpmLockAdapter) parsePnpmLock(ctx context.Context, resolver file
 		return nil, nil, fmt.Errorf("failed to parse pnpm-lock.yaml file: %w", err)
 	}
 
-	packages := make([]pkg.Package, len(pnpmPkgs))
-	for i, p := range pnpmPkgs {
+	packages := make([]pkg.Package, 0, len(pnpmPkgs))
+	for _, p := range pnpmPkgs {
 		if p.Dev && !a.cfg.IncludeDevDependencies {
 			continue
 		}
-		packages[i] = newPnpmPackage(ctx, a.cfg, resolver, reader.Location, p.Name, p.Version, p.Integrity, p.Dependencies)
+		packages = append(packages, newPnpmPackage(ctx, a.cfg, resolver, reader.Location, p.Name, p.Version, p.Integrity, p.Dependencies))
 	}
 
 	return packages, dependency.Resolve(pnpmLockDependencySpecifier, packages), unknown.IfEmptyf(packages, "unable to determine packages")


cmd/syft/internal/options/catalog.go
```diff
@@ -193,7 +193,9 @@ func (cfg Catalog) ToPackagesConfig() pkgcataloging.Config {
 		JavaScript: javascript.DefaultCatalogerConfig().
 			WithIncludeDevDependencies(*multiLevelOption(false, cfg.JavaScript.IncludeDevDependencies)).
 			WithSearchRemoteLicenses(*multiLevelOption(false, enrichmentEnabled(cfg.Enrich, task.JavaScript, task.Node, task.NPM), cfg.JavaScript.SearchRemoteLicenses)).
-			WithNpmBaseURL(cfg.JavaScript.NpmBaseURL),
+			WithNpmBaseURL(cfg.JavaScript.NpmBaseURL).
+			WithPnpmExcludeDevDependencies(*multiLevelOption(false, cfg.JavaScript.PnpmExcludeDevDependencies)),
```
wagoodman: looks like we're adding a new `WithPnpmExcludeDevDependencies` option here when there is already a `WithIncludeDevDependencies` option. It seems like this PR is really a bug fix: we're not honoring the existing option which should be fixed (we don't need to add another option).

rezmoss: @wagoodman i noticed that before too but @kzantow  note here https://github.com/anchore/syft/issues/4430#issuecomment-3665985615  made me think we needed to add a new option

kzantow: Ignore my comment on the issue, honoring the existing `IncludeDevDependencies` is 👍

rezmoss:
@wagoodman done

syft/pkg/cataloger/javascript/parse_pnpm_lock.go
```diff
@@ -53,6 +55,7 @@ type pnpmV9SnapshotEntry struct {
 type pnpmV9PackageEntry struct {
 	Resolution       map[string]string `yaml:"resolution"`
 	PeerDependencies map[string]string `yaml:"peerDependencies"`
+	Dev              bool              `yaml:"dev"`
```
levrik, line 58: This flag does not exist in any of my lockfiles created by PNPM v9 or v10. I don't think this PR works!

rezmoss: @levrik youre right, this PR just works for pnpm dev exclude for ver before v9,im working on exclusion for v9 and after, which needs a different approach, and will send a PR later

PR #4499, opened 2025-12-23
fixed #4498

### LINKED ISSUE #4498: binary cataloger is picking up the go version instead of the actual binary version in traefik experimental images
**What happened**:

when scanning traefik experimental docker images like traefik/traefik:experimental-v3.6, syft reports the go compiler version instead of the actual traefik version

traefik experimental builds use commit hashes instead of semver, so syft falls back to the go version and reports it as the binary version

syft shows 1.24.11 as the traefik version, but that's actually the go compiler version

thanks to the traefik team for pointing out the bug affecting their experimental builds

empty or omitted 

syft docker:traefik/traefik:experimental-v3.6 | grep traefik

```
docker run -it traefik/traefik:experimental-v3.6 version
Version:      8e6ce08f33b2adcb9042d2ce7acf8cf51ee1c7bf
Codename:     cheddar
Go version:   go1.24.11
Built:        2025-12-23_11:07:53AM
OS/Arch:      linux/arm64
```

**Environment**:
- Output of `syft version`:

rezmoss:
## Description

fixed #4498 

## Type of change

## Checklist


kzantow:
If we're changing the regexes for classifiers, it's worth [adding new test case](https://github.com/anchore/syft/tree/main/syft/pkg/cataloger/binary#adding-tests-for-the-binary-classifier-cataloger) for at least one of the versions that wasn't working.

rezmoss:
@kzantow added negative case for experimental ver and also added 3.6.5 , new regex now picks up all valid trafiek with correct ver but skips experimental ones

PR #4500, opened 2025-12-24
fix: snap cataloger incorrectly identifies snap container as deb package

### LINKED ISSUE #4486: bug: Syft mis-identifies binary as deb inside a snap
This is a redo of Grype issue [3122](https://github.com/anchore/grype/issues/3122), that was thoughoughly discussed in the latest [live stream](https://www.youtube.com/watch?v=EuEocYRN4ag).

**What happened**:

Syft is identifying a binary as a deb, which actually didn't come from a debian package. This results in Grype (when instructed via `--distro ubuntu:24.04`) trying to match the deb with something from the Ubuntu security feed. This is incorrect behaviour, given the binary didn't come from, and is 

Syft should be clear when a binary originated from a deb, and when it's not identified as such, and can be reported as just a "binary".

```
$ syft snap:emacs | grep emacs
emacs                           30.2                               deb
emacsen-common                  3.0.5                              deb
```

In this case, syft is incorrect, as the binary in this snap did not originate from a deb at all.

Syft should mark this as a `binary` type.

Snaps are squashfs files that are mounted on the target computer without being "unpacked" into the filesystem. They can contain anything including files that originated from a deb in the Ubuntu repository, but are not limited to those types of files. A snap could also contain a binary built on the developers workstation, in the cloud or random binaries "wgot" from the web.

There are numerous potential yaml files which _may_ describe the snap contents, but also, they might not be there at all.

Consider that someone can craft a snap file with a simple `snap.yaml` and not much more.  e.g. 

```
$ snap download null
$ unsquashfs null_3.snap
$ tree squashfs_root
squashfs-root/
├── bin
│   └── bash
└── meta
    └── snap.yaml
$ cat squashfs-root/bin/bash
#!/bin/bash
```

No binaries!

In the case of the emacs snap, however, there is metadata that syft could consume and parse:

```
$ snap download emacs
$ unsquashfs emacs_3516.snap
$ ls -l squashfs-root/snap/*.yaml squashfs-root/meta/*.yaml
-rw-r--r-- 1 alan alan  1461 Dec 17 06:13 squashfs-root/meta/snap.yaml
-rw-r--r-- 1 alan alan 31628 Dec 17 06:13 squashfs-root/snap/manifest.yaml
-rw-r--r-- 1 alan alan 10867 Dec 17 06:13 squashfs-root/snap/snapcraft.yaml
```

(all attached)

The `snap.yaml` always exists, and is a minimum for a snap. The `snapcraft.yaml` is optional and is often (but not always) the file that controls the build. In this case, it is, and if introspected, reveals a `part` which is `emacs` which is built from source. The `manifest.yaml`, also optional, is added if the snap was built in launchpad, and contains a list of software that is included in the snap.

Snaps are (typically) built against a `base` which is usually (but, as always, not always) is `core`, `core18`, `core20`, `core22`, or `core24` which map to Ubuntu 16.04 LTS, 18.04 LTS, 20.04 LTS, 22.04 LTS, and 24.04 LTS respectively. Meaning, at build time, they're building with/on the indicated Ubuntu release, which means they usually (might) have debs pulled from that Ubuntu release archive.

The common case (which is true with emacs snap) is that the snap contains a binary built from upstream source, along with a pile of support libraries, many of which come from the Ubuntu archive as `stage-packages` specified in the `snapcraft.yaml` and (optionally) detailed in the `manifest.yaml`, along with (optionally) some other libraries which may have come from 3rd party repositories, the developer workstation, or pre-built binaries, who knows!?

[snapcraft.yaml](https://github.com/user-attachments/files/24246177/snapcraft.yaml)
[manifest.yaml](https://github.com/user-attachments/files/24246178/manifest.yaml)
[snap.yaml](https://github.com/user-attachments/files/24246190/snap.yaml)

**Environment**:
- Output of `syft version`:

```
syft version
Application:   syft
Version:       1.38.2
BuildDate:     2025-12-09T21:48:47Z
GitCommit:     bfe63f83dbaea88e22a5cfcd7d704c034c953730
GitDescription: v1.38.2
Platform:      linux/amd64
GoVersion:     go1.25.4
Compiler:      gc
SchemaVersion: 16.1.0
```

popey:
The parseSnapYaml function was creating a DebPkg entry from meta/snap.yaml, which describes the snap container itself, not a debian package inside. 

This caused snaps like emacs (built from source) to incorrectly appear as debian packages in the SBOM output.

Changes:
- parseSnapYaml now returns an empty package list (snap.yaml is container metadata, not package info)
- Actual packages inside snaps are still correctly identified by other parsers (manifest.yaml for primed-stage-packages) and catalogers (binary cataloger for binaries)
- Moved readAll() usage to io.ReadAll() in parse_kernel_changelog.go

## Type of change

## Checklist

## Issue references

Fixes #4486

🤖 Generated with [Claude Code](https://claude.com/claude-code) (with the kindly guiding hands of @popey )


syft/pkg/cataloger/snap/parse_snap_yaml.go
```diff
@@ -2,67 +2,19 @@ package snap
...
-// readAll reads all content from a reader and returns it as bytes
-func readAll(r io.Reader) ([]byte, error) {
-	return io.ReadAll(r)
+// parseSnapYaml parses meta/snap.yaml files. This file contains metadata about
+// the snap container itself (name, version, base, etc.), not about packages
+// inside the snap. We don't create a package entry for the snap container -
+// the actual contents (binaries, debs) are cataloged by their respective
+// catalogers (binary cataloger for binaries built from source, and other snap
+// parsers like parse_system_manifest.go for debian packages from primed-stage-packages).
+func parseSnapYaml(_ context.Context, _ file.Resolver, _ *generic.Environment, _ file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
```
kzantow: I don't see any reason to keep the function here or the related test, since we are not changing any configurable or API-level behavior, only removing this erroneous portion of the lookup. We might as well just remove the parse function and related things, since it's not exported and [the cataloger will continue to function properly with removal](https://github.com/anchore/syft/blob/e9e34948534ab945c7e750376235cfe0c442f532/syft/pkg/cataloger/snap/cataloger.go#L19), note: you will probably need to update the `capabilities.yaml`.

PR #4517, opened 2025-12-31
fixed dotnet cataloger can't find packages from deps.json in linux el…

### LINKED ISSUE #4514: dotnet cataloger can't find packages from deps.json in linux elf single-file bundles
dotnet cataloger misses packages in container images when using PublishSingleFile for linux targets like linux-x64 or linux-musl-x64, but works fine for windows builds , which implemented before 
https://github.com/anchore/syft/pull/4375
 only works for window target

rezmoss:
- fixed dotnet cataloger can't find packages from deps.json in .net linux elf,  fixed #4514
- fixed config ExcludeProjectReferences
## Type of change

## Checklist


wagoodman (commented):
Thanks for the enhancements / fixes -- I'm going to push a little bit of rearrangements to this branch. It looks like there are ELF concerns in the PE package, which is the first issue. There is also a ReadAll() for what appears to be the whole ELF binary, I'll review   further to confirm and see if there is a way to eliminate this call.

rezmoss:
@wagoodman i tried to follow the same pattern you used on my first pr for the windows ver, let me know if i can help with anything

wagoodman:
What happened is that my first PR really combined PE and bundle concerns together into a single package, so there wasn't a "nice spot" to expand it to include ELF concerns (nothing you did wrong). I'll push a small rearrangement shortly to address this before looking into the ReadAll call)

wagoodman:
Ok, I've made a `cataloger/internal/dotnet` with `pe` and `bundle` packages. Odds are good that we'll add a `elf` package and but some of what is in `bundle_elf.go` into there.

rezmoss:
@wagoodman nice refactor, breaking it into cataloger/internal/dotnet/{pe,bundle} is much cleaner and should make it easier to plug in an elf pkg

wagoodman:
Ok, I've made one more change looking into the readall call; we have a [unionreader](https://github.com/anchore/syft/blob/main/syft/internal/unionreader/union_reader.go#L18-L82) package that attempts to get a reader/reader-at/seeker/closer object based on what readers we get from our file resolvers. It has a ReadAll() path, but the vast majority of cases will hit some of the earlier/cheaper calls, type asserting the existing object to expand to the interface.

rezmoss:
@wagoodman got it, thanks for explaining if unionreader usually works through interface widening/squashfs adapter & only hits readall in edge cases then the elf path isnt really a perf concern

rezmoss:
@wagoodman was tough to figure out why the cli test failld, hope this is the right approach

rezmoss:
@wagoodman could you plz review

rezmoss:
@wagoodman  friendly bump on this one, its been sitting since Jan with ci green, i addressed t cli test issue,,since you already handled the pe/bundle rearrangement here, is there anything else needed from my side before this can move forward?

PR #4523, opened 2026-01-05
added deno cataloger , fixed 4417

### LINKED ISSUE #4417: Add Deno support
Add support for [Deno](https://deno.com/), allowing generation of SBOMs for Deno projects much like npm and JavaScript ones.

**Why is this needed**:
Deno is a popular runtime for TypeScript and JavaScript with a focus on security.

**Additional context**:
Deno uses a `Deno.lock` JSON file to keep track of dependencies used.
Packages are able to be imported from http resources via their URL as well as npm.
[Module docs link](https://docs.deno.com/runtime/fundamentals/modules/)

rezmoss:
## Description

fixed #4417

## Type of change

## Checklist


kzantow:
Hey @rezmoss there has been an addition of a command that generates information about cataloger capabilities, the static analysis failures are due to a new cataloger without the `../catalogers/<cataloger>/capabilities.yaml`, there is some documentation here: https://github.com/anchore/syft/blob/main/internal/capabilities/generate/README.md#quick-start if you have some time to go about getting that file created and updated , this is what the failure is about (sorry if you already knew this)

rezmoss:
@kzantow good catch my bad, missed the capabilities generation step just added the yaml for the deno cataloge & regen static analysis should be clear now

kzantow:
@rezmoss it's a fairly new thing, I wouldn't expect many people to know about. The output of the configuration command is what drives a lot of [the documentation site](https://oss.anchore.com/), so including this will get the new cataloger included there when Syft is released 👌

cmd/syft/internal/test/integration/catalog_packages_test.go
```diff
@@ -77,6 +77,7 @@ func TestPkgCoverageImage(t *testing.T) {
 	definedPkgs.Remove(string(pkg.BitnamiPkg))
 	definedPkgs.Remove(string(pkg.GraalVMNativeImagePkg))
 	definedPkgs.Remove(string(pkg.HexPkg))
+	definedPkgs.Remove(string(pkg.JsrPkg))
```
kzantow: TIL about [JSR](https://jsr.io/) 🎉

syft/pkg/cataloger/deno/parse_deno_lock.go
```diff
@@ -0,0 +1,223 @@
...
+	Integrity    string   `json:"integrity"`
+	Dependencies []string `json:"dependencies"`
+}
+
+type npmPackage struct {
+	Integrity    string   `json:"integrity"`
+	Dependencies []string `json:"dependencies"`
+}
+
+func parseDenoLock(_ context.Context, _ file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
```
kzantow: I think this cataloger/parser belongs in the `javascript` package -- it's returning NPM packages, among other things that are all part of the JavaScript ecosystem.That seems to be the consistent thing to do.

syft/pkg/cataloger/deno/parse_deno_lock.go
```diff
@@ -0,0 +1,223 @@
...
+	} else {
+		version = rest[:slashIdx]
+	}
+
+	return name, version
+}
+
+func remotePackageURL(name, version string) string {
+	return packageurl.NewPackageURL(
+		"deno",
```
kzantow: It looks like this is not a PackageURL type. We should not include invalid PURLs.

syft/pkg/cataloger/deno/parse_deno_lock.go
```diff
@@ -0,0 +1,223 @@
...
+func jsrPackageURL(name, version string) string {
+	var namespace string
+	fields := strings.SplitN(name, "/", 2)
+	if len(fields) > 1 {
+		namespace = fields[0]
+		name = fields[1]
+	}
+
+	return packageurl.NewPackageURL(
+		"jsr",
```
kzantow: Same comment here: `jsr` is not a valid PURL type. This seems like we _might_ be able to use `npm` and a custom [repository](https://github.com/package-url/purl-spec/blob/63ada1de67afed2b9f284975a239d747c3b8de7d/types-doc/npm-definition.md?plain=1#L16-L20)

syft/pkg/cataloger/deno/parse_deno_lock.go
```diff
@@ -0,0 +1,223 @@
...
+		name, version := parseRemoteURL(url)
+		if name == "" {
+			continue
+		}
+		pkgs = append(pkgs, newRemotePackage(reader.Location, name, version, url, integrity))
+	}
+
+	pkg.Sort(pkgs)
+
+	return pkgs, nil, unknown.IfEmptyf(pkgs, "unable to determine packages")
```
kzantow: It looks like we're capturing the dependency information but not returning dependency relationships here, I think most users would appreciate the known relationship information to be surfaced, do you agree?

kzantow (dismissed):
I'm finally getting to a proper review here. Overall this looks great, I think! A couple things as noted in the comment, the biggest of which is moving the package to `javascript` and making sure we don't generate invalid PURLs, apologies guiding you sideways (you'll still need the info in the capabilities.yaml, just moved as part of the existing one).

rezmoss:
@kzantow thanks for the detailed feedback pushed updates that cover everything you mentioned, moved the deno parser into the js cataloger since its handling npm pkgs and fits better there, fixed purl & capabilities yaml file

rezmoss:
@kzantow could you help with json schema, i tried to fix it & it keeps failing

rezmoss:
@kzantow mind checking this

rezmoss:
@kzantow  i addressed your review back in Jan, moved t parser into the javascript cataloger & fixed the purl & capabilities.yaml, but i'm still stuck on t json schema check failing, could you point me to what t schema workflow expects for this change? that's the only thing keeping it from being review ready

kzantow (approved):
Thanks @rezmoss sorry this got lost in the noise 😬

PR #4532, opened 2026-01-06
fixed #4529, parse and detect old bitnami img pkgs

### LINKED ISSUE #4529: old bitnami images without spdx files arent getting picked up correctly in the catalog
**What happened**:

the cataloger relies on spdx files, so when older bitnami images don’t have them, it skips over their packages and misses possible vulnerabilities

e.g

https://hub.docker.com/r/bitnamilegacy/mongodb/tags?name=4.4.11

**Environment**:
- Output of `syft version`: latest 1.39.0

rezmoss:
## Description

fixed #4529, parse and detect old bitnami img pkgs

## Type of change

## Checklist

### DIFF
--- a/syft/pkg/cataloger/bitnami/capabilities.yaml
+++ b/syft/pkg/cataloger/bitnami/capabilities.yaml
@@ -13,6 +13,35 @@ catalogers:
       - installed
       - package
     parsers: # AUTO-GENERATED structure
+      - function: parseComponentsJSON
+        detector: # AUTO-GENERATED
+          method: glob # AUTO-GENERATED
+          criteria: # AUTO-GENERATED
+            - /opt/bitnami/.bitnami_components.json
+        metadata_types: # AUTO-GENERATED
+          - pkg.BitnamiSBOMEntry
+        package_types: # AUTO-GENERATED
+          - bitnami
+        json_schema_types: # AUTO-GENERATED
+          - BitnamiSbomEntry
+        capabilities: # MANUAL - preserved across regeneration
+          - name: license
+            default: false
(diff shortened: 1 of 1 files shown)


syft/pkg/cataloger/bitnami/cataloger.go
```diff
@@ -23,6 +23,9 @@ func NewCataloger() pkg.Cataloger {
 	return generic.NewCataloger(catalogerName).
 		WithParserByGlobs(parseSBOM,
 			"/opt/bitnami/**/.spdx-*.spdx",
+		).
+		WithParserByGlobs(parseComponentsJSON,
+			"/opt/bitnami/.bitnami_components.json",
```
kzantow, line 28: is this always at `/opt/bitnami/..`? I guess we can always start specific and if needed make this a glob search 👍

syft/pkg/cataloger/bitnami/parse_components.go
```diff
@@ -0,0 +1,100 @@
...
+		purl := packageurl.NewPackageURL(
+			"bitnami",
+			"",
+			name,
+			entry.Version,
+			qualifiers,
+			"",
+		).String()
+
+		metadata := &pkg.BitnamiSBOMEntry{
```
kzantow, line 64: Generally, we try to have metadata types that are very specific to the source material, and the SBOMEntry part makes this a little misleading, since these component definitions don't seem to be SBOMs, _exactly_. But if we made a different type it would be almost identical anyway, so this seems good to me (team agreed), just noting it.

kzantow (approved):
LGTM 👍

PR #4560, opened 2026-01-19
Add "fail" compliance action, fixes #4388

### LINKED ISSUE #4388: Report an error result when packages found without version
I've started using syft to create SBOMs for Python projects. Many projects list their requirements unversioned in requirements.txt and syft will ignore those dependencies, but create an SBOM. The pipeline succeeds and a syntactically correct SBOM has been generated, making it easy to stop working here.

Is there something like a strict mode? That is, an unversioned dependency is not ignored, but leads to an error?

The current behaviour feels like a silent failure, which is something a security tool probably shouldn't do.

(I'm ignoring indirect dependencies here.)

zilosz:
## Description

Fixes #4388

Adds a `fail` compliance action for missing package names and versions. When configured, Syft will throw an error and abort the SBOM generation process instead of continuing.

In addition to existing actions, the `missing-name` and `missing-version` config options under `compliance` now accept `fail` (with alias `error`).

## Type of change

## Checklist


kzantow:
Based on the discussions we've had about this one, I think we should [update the python cataloger to fix the bug where it's not returning packages without versions](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/python/parse_requirements.go#L140-L144), so these can be later handled by the compliance rules. If we update this PR with that one known issue, I think it [solves the problem raised in the issue #4388](https://github.com/anchore/syft/issues/4388#issuecomment-3752870296), and we can have follow-on issues for any other catalogers we might find that didn't properly get updated for compliance handling.

There was further discussion about packages without _names_, but I think this is still an "it depends on the cataloger" decision: if we can give the user enough information to determine where this completely blank package came from where we were actually expecting a package name, we should probably return it, which for JSON might require a future go version with stdlib's JSON v2 to return file and line numbers. Other things like a blank line in requrements.txt probably shouldn't qualify for this as an example. For these other changes, they should probably be follow-ons also, not part of this PR.

Let me know if any of that is unclear; I think the python change mostly solves your use case, as I understood it.

zilosz:
For now, I've just modified the requirements.txt parser to emit packages with missing versions, since that's what the original issue focused on. I did notice that other Python parsers such as parse_setup.go also explicitly handle missing versions instead of emitting full packages, so perhaps I could fix that in another PR. I also think it would be a good idea to fix other catalogers through separate issues.

zilosz:
I worked on this PR for the first time in a while and ended up polluting the commit history. Cleaner PR is now #5231

zilosz:
Closing now.

PR #4569, opened 2026-01-23
feat: detect Debian version from /etc/debian_version

kzantow:
## Description

Some Debian containers such as Debian distroless (e.g.. `docker.bintray.io/jfrog/artifactory-oss:6.23.13`) do not contain a verison in the `/etc/os-release` but do have a version in `/etc/debian_version`. 

This PR looks for a `debian_version` in the event that the release is identified as `debian` with no version information.

## Type of change

## Checklist

### DIFF
--- a/syft/linux/supplement_release.go
+++ b/syft/linux/supplement_release.go
@@ -44,7 +44,7 @@ func readDebianVersionFile(resolver file.Resolver, location file.Location) strin
 		return ""
 	}
 	version := strings.TrimSpace(string(contents))
-	if regexp.MustCompile(`\d+(?:\.\d+)?`).MatchString(version) {
+	if regexp.MustCompile(`^\d+(?:\.\d+)?$`).MatchString(version) {
 		return version
 	}
 	return ""


syft/linux/supplement_release.go
```diff
@@ -0,0 +1,49 @@
...
+
+func readDebianVersionFile(resolver file.Resolver, location file.Location) string {
+	rdr, err := resolver.FileContentsByLocation(location)
+	if err != nil {
+		log.Debugf("error getting contents for %s: %v", location.RealPath, err)
+	}
+	defer internal.CloseAndLogError(rdr, location.RealPath)
+	contents, err := io.ReadAll(rdr)
+	if err != nil {
+		log.Debugf("error reading %s: %v", location.RealPath, err)
```
spiffcs, line 43: Do we just return here with a blank string since `io.ReadAll` gave us an error?

I think the early return applies for all the error cases in this funciton.

syft/linux/supplement_release.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	version := strings.TrimSpace(string(contents))
+	if regexp.MustCompile(`\d+(:?.\d+)?`).MatchString(version) {
```
spiffcs: I think this get's compiled on every call. Do we want to make it a package level regex to make it a bit more efficient?

Ex:
```
var debianVersionMatcher = regexp.MustCompile(`\d+(:?.\d+)?`)
```

spiffcs: Also super minor regex nit: \d+(:?.\d+)? works for the expected inputs, but (:? is technically a capturing group with an optional colon rather than a non-capturing group ((?:), and the . isn’t escaped. That makes the match a bit looser than intended.

If we want stricter version matching, something like ^\d+(?:\.\d+)*$ might be a clearer and more precise alternative. Totally optional, but thought there could be options here.

kzantow: Good catches! Will fix 👍

kzantow: As for global, it doesn't have to stay in memory, which is why I'd prefer not to make it global. It only runs once, when we _found_ a linux release from one of the files, at which point we stop looking for more. For regexes we reuse a bunch, I definitely agree to make those global, though.

PR #4573, opened 2026-01-26
fix: update mixed case dependencies in python to be normalized

### LINKED ISSUE #4562: [Bug Report] Missing some dependencies on cyclonedx formatted SBOM using syft
**What happened**: missing dependencies on cyclonedx formatted SBOM
<br> When generating a CycloneDX SBOM with Syft from a Python project that uses `poetry.lock`, the dependency edge for `dj-rest-auth -> django` is missing.  
In my `poetry.lock`, `dj-rest-auth` declares the dependency as `Django` (capitalized), and Syft appears to fail to resolve it to the `django` component, so the edge is omitted from the SBOM dependency graph.

---

<br> Syft should treat Python distribution names as case-insensitive (and ideally normalize common separators), so `Django` should resolve to the same package as `django`.  
The generated CycloneDX SBOM should include the dependency edge:

- `dj-rest-auth@7.0.1` **depends on** `django@5.2.6`

--- 

<br>
1. Create or use a Python project with a `poetry.lock` containing something like:

```toml
[[package]]
name = "dj-rest-auth"
version = "7.0.1"
description = "Authentication and Registration in Django Rest Framework"
optional = false
python-versions = ">=3.8"
groups = ["main"]
files = [
    {file = "dj-rest-auth-7.0.1.tar.gz", hash = "sha256:3f8c744cbcf05355ff4bcbef0c8a63645da38e29a0fdef3c3332d4aced52fb90"},
]

[package.dependencies]
Django = ">=4.2,<6.0"
djangorestframework = ">=3.13.0"

[package.extras]
with-social = ["django-allauth[socialaccount] (>=64.0.0)"]
```

2. Run Syft to generate a CycloneDX SBOM from the project directory (example):

```bash
syft dir:. -o cyclonedx-json > sbom.json
```

3. Inspect the dependency graph section in the resulting SBOM (CycloneDX dependencies section).
<br> Confirm dj-rest-auth -> djangorestframework is present.
Observe dj-rest-auth -> django is missing, even though the lockfile declares Django.

```json
{
      "ref": "pkg:pypi/dj-rest-auth@7.0.1?package-id=e5ae9d3dbec01ba2",
      "dependsOn": [
        "pkg:pypi/django-allauth@65.11.2?package-id=7252146fe0fe7589",
        "pkg:pypi/djangorestframework@3.16.1?package-id=e1fa1261e5a7882a"
      ]
 }
```
- This example is SBOM that I created on today.
---

- This looks like a name normalization issue: the dependency key is Django (capital D) in poetry.lock, while the locked package entry is django (lowercase).
- Python distribution name comparison is effectively case-insensitive, so the dependency should be resolved.
- I can provide the full poetry.lock (if you give me a comment, then I can give it to you right away!)

**Environment**:
- Output of `syft version`: <br> I cloned the source code(main branch at 2026-01-16 GMT)and build in local so I can give you build revision
-- build   vcs.revision=795432441754b910d4456fa982eb7df64c98b432
-- build   vcs.time=2026-01-16T18:04:13Z
- OS (e.g: `cat /etc/os-release` or similar): Windows 11 (25H2)

spiffcs:
## Description

When generating a CycloneDX SBOM from a Python project that uses poetry.lock, Syft was failing to resolve dependency edges when the dependency name used different casing than the package name.

  For example, in a poetry.lock file, dj-rest-auth declares a dependency on Django (capitalized), but the actual package is listed as django (lowercase). 

Per the https://packaging.python.org/en/latest/specifications/name-normalization/, package names should be treated as case-insensitive.

### Root Cause
The packageRef function in the poetry dependency resolver was not normalizing package names before using them for dependency matching.

### Fix
Fix: Apply the existing normalize() function to package names and extras in packageRef(), ensuring case-insensitive matching that complies with Python packaging standards.

Before: The dependency edge dj-rest-auth@7.0.1 → django@5.2.6 was missing from the SBOM dependency graph.
After: Dependency edges are correctly resolved regardless of package name casing in poetry.lock.

Some other tests have been updated that were missing this relationship previously

### Type of Change

  - Bug fix (non-breaking change which fixes an issue)

Issue references
Fixes #4562

### DIFF
--- a/syft/pkg/cataloger/python/dependency_test.go
+++ b/syft/pkg/cataloger/python/dependency_test.go
@@ -182,7 +182,7 @@ func Test_poetryLockDependencySpecifier(t *testing.T) {
 			},
 		},
 		{
-			name: "dependency names with mixed case should be normalized (issue #4562)",
+			name: "dependency names with mixed case should be normalized",
 			p: pkg.Package{
 				Name: "dj-rest-auth",
 				Metadata: pkg.PythonPoetryLockEntry{


syft/pkg/cataloger/python/dependency_test.go
```diff
@@ -276,6 +332,64 @@ func Test_poetryLockDependencySpecifier_againstPoetryLock(t *testing.T) {
...
+		{
+			name: "package name with underscores",
+			pkg:  "some_package",
+			want: "some-package",
+		},
+		{
+			name:  "package name with mixed case and extra",
+			pkg:   "Django",
+			extra: "argon2",
+			want:  "django[argon2]",
```
willmurphyscode, line 370: How do the extras interact with the resolution in requires/provides?

spiffcs: So here is the current behavior:

When a package requires another with extras:
  1. provides is always the base package no extras in the name
  2. then we always have the base package requirements without any extras so those resolutions can always happen
  3. If there are extras, the extra variants also get added (now run through the normalize logic)
 
1
https://github.com/anchore/syft/blob/e8b4527bfbb7780cd70f2541fce4dca58803fb53/syft/pkg/cataloger/python/dependency.go#L24-L27

2
https://github.com/anchore/syft/blob/e8b4527bfbb7780cd70f2541fce4dca58803fb53/syft/pkg/cataloger/python/dependency.go#L34-L37

3
https://github.com/anchore/syft/blob/e8b4527bfbb7780cd70f2541fce4dca58803fb53/syft/pkg/cataloger/python/dependency.go#L38-L47

This change does not update any of the above flow. It just makes sure that we are respecting the casing correctly in both instances of creating the `provides:requires` and the extras variant.

willmurphyscode: We reviewed this offline, and I think the behavior currently in main is correct. Provides and requires are include both a plain and a with extras variant as appropriate, like `mashumaro` and `mashumaro[orjson]`, and this is unit tested.

spiffcs: No more changes needed for this PR then and we are 🟢 ❤️

PR #4585, opened 2026-01-30
feat: Phase 1 memory optimization for JavaScript SBOM generation

matthyx:
This commit implements Phase 1 of the memory optimization plan, focusing on quick wins that reduce memory allocations.

Changes:

1. JavaScript Lock File Parsers:
   - yarn.lock: Use bufio.Scanner for line-by-line reading instead of
     loading entire file and regex.Split, reducing peak memory usage.
     Only loads the first 100 bytes to detect version, then parses v1 files
     line-by-line directly from the file handle.
   - pnpm-lock.yaml: Replace string concatenation with strings.Join and
     use strings.Cut for efficient splitting, reducing string allocations
   - Pre-allocate map sizes to reduce reallocations

2. Dependency Resolution:
   - Replace strset-based deduplication with efficient map-based approach
   - Use struct keys instead of string concatenation for relationship tracking
   - Eliminate temporary string allocations in deduplicate function

Note: A previous optimization to defer license scanner creation was removed
after review, as the scanner should always be set in the context before
cataloging begins.

Memory Impact:
- Reduced memory usage for JavaScript SBOM generation
- Peak memory reduction for large JS projects (10k+ packages)
- Maintains all existing functionality and test coverage

All existing tests pass with these changes.

Related: Detailed analysis available at https://gist.github.com/matthyx/e500d3282876fd6a41064770eacc7229
Issue: #4586

### DIFF
--- a/internal/licenses/context.go
+++ b/internal/licenses/context.go
@@ -2,11 +2,14 @@ package licenses
 
 import (
 	"context"
+	"errors"
 )
 
 type licenseScannerKey struct{}
 
 var ctxKey = licenseScannerKey{}
+
+var ErrNoLicenseScanner = errors.New("no license scanner set in context")
 
 func SetContextLicenseScanner(ctx context.Context, s Scanner) context.Context {
 	return context.WithValue(ctx, ctxKey, s)
@@ -18,8 +21,9 @@ func IsContextLicenseScannerSet(ctx context.Context) bool {
 }
 
 func ContextLicenseScanner(ctx context.Context) (Scanner, error) {
-	if s, ok := ctx.Value(ctxKey).(Scanner); ok {
-		return s, nil
+	s, ok := ctx.Value(ctxKey).(Scanner)
+	if !ok {
+		return nil, ErrNoLicenseScanner
 	}
-	return NewDefaultScanner()
+	return s, nil
 }
--- a/syft/pkg/cataloger/internal/dependency/resolver.go
+++ b/syft/pkg/cataloger/internal/dependency/resolver.go
@@ -3,13 +3,16 @@ package dependency
 import (
 	"sort"
 
-	"github.com/scylladb/go-set/strset"
-
 	"github.com/anchore/syft/internal"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/pkg"
 	"github.com/anchore/syft/syft/pkg/cataloger/generic"
 )
+
+// pairKey is a struct used to track seen relationship pairs without string concatenation
+type pairKey struct {
+	from, to artifact.ID
+}
 
 // Specification holds strings that indicate abstract resources that a package provides for other packages and
 // requires for itself. These strings can represent anything from file paths, package names, or any other concept
@@ -64,15 +67,15 @@ func Resolve(specifier Specifier, pkgs []pkg.Package) (relationships []artifact.
 		specsByPkg[id] = allProvides(pkgsProvidingResource, id, specifier(p))
 	}
 
-	seen := strset.New()
+	seen := make(map[pairKey]struct{})
 	for _, dependantPkg := range pkgs {
 		specs := specsByPkg[dependantPkg.ID()]
 		for _, spec := range specs {
 			for _, resource := range deduplicate(spec.Requires) {
 				for providingPkgID := range pkgsProvidingResource[resource] {
-					// prevent creating duplicate relationships
-					pairKey := string(providingPkgID) + "-" + string(dependantPkg.ID())
-					if seen.Has(pairKey) {
+					// prevent creating duplicate relationships using struct key instead of string concatenation
+					key := pairKey{from: providingPkgID, to: dependantPkg.ID()}
+					if _, exists := seen[key]; exists {
 						continue
 					}
 
@@ -86,7 +89,7 @@ func Resolve(specifier Specifier, pkgs []pkg.Package) (relationships []artifact.
 						},
 					)
 
-					seen.Add(pairKey)
+					seen[key] = struct{}{}
 				}
 			}
 		}
@@ -112,8 +115,16 @@ func allProvides(pkgsProvidingResource map[string]internal.Set[artifact.ID], id 
 
 func deduplicate(ss []string) []string {
 	// note: we sort the set such that multiple invocations of this function will be deterministic
-	set := strset.New(ss...)
-	list := set.List()
-	sort.Strings(list)
-	return list
+	// use map for O(1) lookups without strset overhead
+	unique := make(map[string]struct{}, len(ss))
+	result := make([]string, 0, len(unique))
+
+	for _, s := range ss {
+		if _, exists := unique[s]; !exists {
+			unique[s] = struct{}{}
+			result = append(result, s)
+		}
+	}
+	sort.Strings(result)
+	return result
 }
--- a/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
@@ -107,17 +107,21 @@ func (p *pnpmV6LockYaml) Parse(version float64, data []byte) ([]pnpmPackage, err
 			log.WithFields("key", key).Trace("unable to parse pnpm package key")
 			continue
 		}
-		pkgKey := name + "@" + ver
+		pkgKey := strings.Join([]string{name, ver}, "@")
 
 		integrity := ""
 		if value, ok := pkgInfo.Resolution["integrity"]; ok {
 			integrity = value
 		}
 
-		dependencies := make(map[string]string)
+		dependencies := make(map[string]string, len(pkgInfo.Dependencies))
 		for depName, depVersion := range pkgInfo.Dependencies {
-			var normalizedVersion = strings.SplitN(depVersion, "(", 2)[0]
-			dependencies[depName] = normalizedVersion
+			// Use strings.Cut for more efficient splitting
+			if normalizedVersion, _, ok := strings.Cut(depVersion, "("); ok {
+				dependencies[depName] = normalizedVersion
+			} else {
+				dependencies[depName] = depVersion
+			}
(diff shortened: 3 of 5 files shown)


internal/licenses/context.go
```diff
@@ -2,12 +2,28 @@ package licenses
...
 
+// Package-level lazy initialization for default scanner
+var (
+	defaultScanner Scanner
+	defaultOnce    sync.Once
+	defaultErr     error
+)
+
+// getDefaultScanner lazily initializes and returns the default license scanner
+func getDefaultScanner() (Scanner, error) {
```
kzantow: I'm a little confused about this change, there should be [a single global license scanner available in the context already](https://github.com/anchore/syft/blob/main/syft/create_sbom.go#L117), perhaps there needs to be some changes to pass context further instead?

MEMORY_ANALYSIS.md
```diff
@@ -0,0 +1,748 @@
+# Syft Memory Analysis: JavaScript SBOM Generation
```
kzantow: We probably shouldn't commit memory analysis reports like this or else this repo will get very large every time someone makes a similar change

matthyx: I've attached it as a gist in the issue

internal/licenses/context.go
matthyx: fixed, sorry

syft/pkg/cataloger/internal/dependency/resolver.go
```diff
@@ -112,8 +115,16 @@ func allProvides(pkgsProvidingResource map[string]internal.Set[artifact.ID], id
 
 func deduplicate(ss []string) []string {
 	// note: we sort the set such that multiple invocations of this function will be deterministic
-	set := strset.New(ss...)
-	list := set.List()
-	sort.Strings(list)
-	return list
+	// use map for O(1) lookups without strset overhead
+	unique := make(map[string]struct{}, len(ss))
```
kzantow, line 119: This change might be good, but it's sacrificing readability for what is probably a small gain.

Tangentially, I question whether this needs to be sorted.

syft/pkg/cataloger/internal/dependency/resolver.go
```diff
@@ -64,15 +67,15 @@ func Resolve(specifier Specifier, pkgs []pkg.Package) (relationships []artifact.
 		specsByPkg[id] = allProvides(pkgsProvidingResource, id, specifier(p))
 	}
 
-	seen := strset.New()
+	seen := make(map[pairKey]struct{})
```
kzantow, line 70: This looks like a good change -- we should avoid creating temporary strings like this as much as possible, but using the `map[]struct{}` is always cumbersome. This is a change we could accept, though I hope at some point we add a generic set. We have a library to share across our projects for this purpose, and added a suggestion here: https://github.com/anchore/go-collections/pull/5

syft/pkg/cataloger/javascript/parse_yarn_lock.go
```diff
@@ -171,21 +182,34 @@ func (a genericYarnLockAdapter) parseYarnLock(ctx context.Context, resolver file
...
-	// v1 Yarn lockfiles are not YAML, so we need to parse them as a special case. They typically
-	// include a comment line that indicates the version. I.e. "# yarn lockfile v1"
-	if strings.Contains(string(data), "# yarn lockfile v1") {
-		yarnPkgs, err = parseYarnV1LockFile(reader)
+	if isV1 {
+		// For v1, parse line-by-line without loading entire file into memory
+		yarnPkgs, err = parseYarnV1LockFile(bufReader)
 	} else {
+		// For v2+, we need to load the entire file as YAML
+		data, err := io.ReadAll(bufReader)
```
kzantow, line 204: It looks like this should just be able to pass the `bufReader` to an `io.NopCloser` and not have to `io.ReadAll`, either

syft/pkg/cataloger/javascript/parse_pnpm_lock.go
```diff
@@ -143,7 +147,7 @@ func (p *pnpmV9LockYaml) Parse(_ float64, data []byte) ([]pnpmPackage, error) {
 			log.WithFields("key", key).Trace("unable to parse pnpm v9 package key")
 			continue
 		}
-		pkgKey := name + "@" + ver
+		pkgKey := strings.Join([]string{name, ver}, "@")
```
kzantow, line 150: This is the same less efficient change.

internal/licenses/context.go
```diff
@@ -18,8 +21,9 @@ func IsContextLicenseScannerSet(ctx context.Context) bool {
 }
 
 func ContextLicenseScanner(ctx context.Context) (Scanner, error) {
-	if s, ok := ctx.Value(ctxKey).(Scanner); ok {
-		return s, nil
+	s, ok := ctx.Value(ctxKey).(Scanner)
```
kzantow, line 24: There is still an unnecessary change here. There is a scanner created before cataloging happens, and that scanner is passed in the context to all users.

syft/pkg/cataloger/javascript/parse_pnpm_lock.go
```diff
@@ -107,17 +107,21 @@ func (p *pnpmV6LockYaml) Parse(version float64, data []byte) ([]pnpmPackage, err
 			log.WithFields("key", key).Trace("unable to parse pnpm package key")
 			continue
 		}
-		pkgKey := name + "@" + ver
+		pkgKey := strings.Join([]string{name, ver}, "@")
```
kzantow, line 110: This is still concatenating strings together and is _less_ efficient -- up to half as efficient. This was my intuition, so I wrote a benchmark to verify this; I'll skip to the results:
```
% go test -bench=. -benchmem
...
Benchmark_strcat-12             24962109                41.53 ns/op            0 B/op          0 allocs/op
Benchmark_stringsJoin-12        18459672                62.24 ns/op           16 B/op          1 allocs/op
```

Test:
```
package main

import (
	"math/rand"
	"strings"
	"testing"
	"time"
)

func Benchmark_strcat(b *testing.B) {
	for b.Loop() {
		s1, s2 := samples()
		got := s1 + "/" + s2
		if len(got) > 100 {
			return
		}
	}
}

func Benchmark_stringsJoin(b *testing.B) {
	for b.Loop() {
		s1, s2 := samples()
		got := strings.Join([]string{s1, s2}, "/")
		if len(got) > 100 {
			return
		}

	}
}

var input = []string{
	"dsfgdsg",
	"34gg",
	"dfsgdfg",
	"asqwfdf",
	"qwefqwef",
	"agsah",
	"adgsadf",
	"wergerwg",
}

var ilen = len(input)
var r = rand.New(rand.NewSource(time.Now().UnixMilli()))

func samples() (s1, s2 string) {
	i := r.Intn(ilen)
	return input[i], input[(i+1)%ilen]
}

```

To be sure there wasn't too much being optimized out, I took the example further and it continued to show the concat form uses fewer allocations and less time.

syft/pkg/cataloger/javascript/parse_pnpm_lock.go
```diff
@@ -153,12 +157,16 @@ func (p *pnpmV9LockYaml) Parse(_ float64, data []byte) ([]pnpmPackage, error) {
 			log.WithFields("key", key).Trace("unable to parse pnpm v9 package snapshot key")
 			continue
 		}
-		pkgKey := name + "@" + ver
+		pkgKey := strings.Join([]string{name, ver}, "@")
```
kzantow, line 160: Another instance.

syft/pkg/cataloger/javascript/parse_pnpm_lock.go
```diff
@@ -107,17 +107,21 @@ func (p *pnpmV6LockYaml) Parse(version float64, data []byte) ([]pnpmPackage, err
...
 			integrity = value
 		}
 
-		dependencies := make(map[string]string)
+		dependencies := make(map[string]string, len(pkgInfo.Dependencies))
 		for depName, depVersion := range pkgInfo.Dependencies {
-			var normalizedVersion = strings.SplitN(depVersion, "(", 2)[0]
-			dependencies[depName] = normalizedVersion
+			// Use strings.Cut for more efficient splitting
+			if normalizedVersion, _, ok := strings.Cut(depVersion, "("); ok {
```
kzantow, line 120: All of these changes to `Cut` seem like they have very limited impact, neither function is allocating and this change makes it more verbose. Additionally, the `Split(..)[0]` pattern is used throughout Syft whereas the other form is not; without a demonstrated benefit here, I think the should not be changed.

syft/pkg/cataloger/javascript/parse_pnpm_lock.go
```diff
@@ -153,12 +157,16 @@ func (p *pnpmV9LockYaml) Parse(_ float64, data []byte) ([]pnpmPackage, error) {
...
-		pkgKey := name + "@" + ver
+		pkgKey := strings.Join([]string{name, ver}, "@")
 		if pkg, ok := packages[pkgKey]; ok {
-			pkg.Dependencies = make(map[string]string)
+			pkg.Dependencies = make(map[string]string, len(snapshotInfo.Dependencies))
 			for name, versionSpecifier := range snapshotInfo.Dependencies {
-				var normalizedVersion = strings.SplitN(versionSpecifier, "(", 2)[0]
-				pkg.Dependencies[name] = normalizedVersion
+				// Use strings.Cut for more efficient splitting
+				if normalizedVersion, _, ok := strings.Cut(versionSpecifier, "("); ok {
```
kzantow, line 165: Another Cut change.

matthyx:
Thanks @kzantow for the detailed review and suggestions.
I will rework this to focus on hot paths with separate micro-benchmarks for each of them... but in 10 days or so, after my PTO.
Thanks for your time and understanding.

matthyx:
@kzantow thanks for your patience, I have created https://github.com/anchore/stereoscope/pull/527 which focuses on the file tree.

Issue #4233, opened 2025-09-23
Wrong format in license


whereIsMyDipp:
**What happened**:

I came across a bug regarding licenses format. While the format is human-readable machines have issues with that.
The result from a java project with syft results in a license which is not machine readable
Current format: "LicenseRef-http---www.eclipse.org-legal-epl-v10.html--http---www.gnu.org-licenses-old-licenses-lgpl-2.1.html"

**What you expected to happen**:

I should be looking like this "EPL-1.0 AND LGPL-2.1-only" 
(according to https://spdx.org/licenses/)

**Steps to reproduce the issue**:

```
git clone https://github.com/spring-projects/spring-petclinic.git
once checked out --> mvn install

syft . -o spdx-json=syft-sbom-spdx.json
```

As a sample you can see the dependency logback-classic:
```

      "name": "logback-classic",
      "SPDXID": "SPDXRef-Package-java-archive-logback-classic-20e8c9ef925bd6be",
      "versionInfo": "1.5.18",
      "supplier": "Organization: QOS.ch",
      "originator": "Organization: QOS.ch",
      "downloadLocation": "NOASSERTION",
      "filesAnalyzed": false,
      "checksums": [
        {
          "algorithm": "SHA1",
          "checksumValue": "fc371f3fc97a639de2d67947cffb7518ec5e3d40"
        }
      ],
      "sourceInfo": "acquired package info from installed java archive: /target/spring-petclinic-3.5.0-SNAPSHOT.jar",
      "licenseConcluded": "NOASSERTION",
      "licenseDeclared": "LicenseRef-http---www.eclipse.org-legal-epl-v10.html--http---www.gnu.org-licenses-old-licenses-lgpl-2.1.html",
      "copyrightText": "NOASSERTION",
```
	
	
It shoes the license as a reference rather the the actual 3 licenses.
The same issue appears also via 

`syft . -o cyclonedx-json=syft-sbom-cyclonedx.json`

 ```
{
      "bom-ref": "pkg:maven/ch.qos.logback/logback-classic@1.5.18?package-id=20e8c9ef925bd6be",
      "type": "library",
      "group": "ch.qos.logback",
      "name": "logback-classic",
      "version": "1.5.18",
      "licenses": [
        {
          "license": {
            "name": "http://www.eclipse.org/legal/epl-v10.html, http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html"
          }
        }
      ],
```
	  
[...]

PR #4588, opened 2026-01-30
fix: allow alternate scheme on url->licenseID lookup

spiffcs:
## Description

Update url -> SPDXID lookup to try both http/https.

https://github.com/anchore/syft/pull/4244 asked for http mapping -> BSD-3-Clause

This is because the SPDX license list only supported HTTPS for the lookup:
https://github.com/spdx/license-list-XML/blob/297da51b1b0ea5aab7de4a35faea34ffc43323a0/src/BSD-3-Clause.xml#L5-L9

Rather than add both schemes to the upstream list or manage our own internal map we update the lookup logic to just try both http/https given typos and manual editing that happens for License URL data in the various package managers and software license data.

## Type of change

## Checklist

Fixes: #4233


kzantow:
I haven't reviewed the code specifically, but could we just trim the `http/https://` prefix [in the URL list](https://github.com/anchore/syft/blob/main/internal/spdxlicense/license_list.go#L1431) and when looking it up? It would reduce the binary size every so slightly

spiffcs:
> I haven't reviewed the code specifically,

@kzantow we're not generating a new list with both https/http. We're simply trying both on the lookup.
I can rewrite the generation to trim the prefix from the map and then do a trim to do the lookup as well but that's a larger change.

kzantow:
Right, I'm suggesting we do that -- update the generation to trim `http://` and `https://`, this way we can avoid [creating new strings to do the alternate lookups every time](https://github.com/anchore/syft/pull/4588/changes#diff-82757e250ab469859dc510032987e691e114a9518412ac68ebeb83d4133b3fd3R74), though acknowledge it's a minor optimization, but avoids creation and 2 lookups in cases where the URL isn't in the list at all, which probably is very common.

PR #4596, opened 2026-02-04
Fix crash when scanning PHP extensions with special characters in paths

PopoviciMarian:
## Summary

Fix regex compilation error that caused Syft to crash when scanning directories containing PHP extension files with special characters (such as backslashes) in their paths.

## Motivation

When scanning directories on Windows or with paths containing special regex characters, the PHP cataloger would panic with a regex compilation error.
## Changes

- Use `filepath.Base()` instead of `path.Base()` for better cross-platform path handling
- Escape extension names using `regexp.QuoteMeta()` before inserting them into regex patterns

## Type of change

- [x] Bug fix (non-breaking change)
image

### DIFF
--- a/syft/pkg/cataloger/php/interpreter_cataloger.go
+++ b/syft/pkg/cataloger/php/interpreter_cataloger.go
@@ -3,7 +3,8 @@ package php
 import (
 	"context"
 	"fmt"
-	"path"
+	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/anchore/packageurl-go"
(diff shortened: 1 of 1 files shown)


syft/pkg/cataloger/php/interpreter_cataloger.go
```diff
@@ -202,7 +203,7 @@ func (p interpreterCataloger) getClassifier(realPath string) (string, *binutils.
 		return "", nil
 	}
 
-	base := path.Base(realPath)
+	base := filepath.Base(realPath)
```
kzantow, line 206: The `realPath` here should be a normalized path from the resolver, which uses forward slashes (`/`), `filepath` is not the right package to use. Is there an issue you're trying to fix? I didn't see any linked in the description

PR #4608, opened 2026-02-06
chore: migrate `.goreleaser` config to use docker_v2

### LINKED ISSUE #4113: Debug Docker images are running as nonroot user
**What happened**:

`debug` Docker images are running as `nonroot` user.

#3941 included in release [1.27.0](https://github.com/anchore/syft/releases/tag/v1.27.0) introduced a breaking change by running the image as `nonroot` user.
#3998 included in release [1.27.1](https://github.com/anchore/syft/releases/tag/v1.27.1) reverted this change by building a separate `nonroot` image.
I expect the `debug` image to run as `root`, as it did prior to version 1.27.0.

Run the command `whoami` inside a container based on the `debug` image:

```
docker run --rm --entrypoint whoami ghcr.io/anchore/syft:v1.29.1-debug
```

The result is `nonroot` instead of `root`.

**Environment**:

- Output of `syft version`:

```
Application:   syft
Version:       1.29.1
BuildDate:     2025-07-30T18:16:10Z
GitCommit:     386ef842d99a72027fb5fd1085fde87883640eaf
GitDescription: v1.29.1
Platform:      linux/amd64
GoVersion:     go1.24.5
Compiler:      gc
SchemaVersion: 16.0.36
```

```
PRETTY_NAME="Distroless"
NAME="Debian GNU/Linux"
ID="debian"
VERSION_ID="12"
VERSION="Debian GNU/Linux 12 (bookworm)"
HOME_URL="https://github.com/GoogleContainerTools/distroless"
SUPPORT_URL="https://github.com/GoogleContainerTools/distroless/blob/master/README.md"
BUG_REPORT_URL="https://github.com/GoogleContainerTools/distroless/issues/new"
```

spiffcs:
## Description

Migrate the docker image builds from the legacy `dockers:` + `docker_manifests:` sections to GoReleaser's [`dockers_v2`](https://goreleaser.com/customization/dockers_v2/)

This builds a multi-arch manifest per image in a single `docker buildx` invocation. This collapses ~300 lines of per-arch/per-tag boilerplate into four declarative entries and moves image labels out of the Dockerfiles and into the release config.

Feedback discussion for the feature: https://github.com/orgs/goreleaser/discussions/6005

The same `dockers_v2` pattern has been smoke-tested in [`grant`](https://github.com/anchore/grant) on it's latest release.

Each entry builds for 5 platforms: `linux/amd64`, `linux/arm64`, `linux/ppc64le`, `linux/ri
4 entries × 2 registries × 2 tags = 16 manifest tags, each backed by 5 platform-specific images (80 platform images total).

### `debug` restored to root (fixes #4113)

Before v1.27.0, the `debug` image ran as **root**. #3941 (v1.27.0) unintentionally made it non-root; #3998 (v1.27.1) reverted the *production* images to root and added a `nonroot` tag but left `debug` non-root — the inconsistency reported in #4113.

This PR restores `debug` to **root** and adds a new **`debug-nonroot`** tag for the non-root variant. This makes both image families consistent: the bare tag is root (`latest`, `debug`) and `-nonroot` opts into a non-root user (`nonroot`, `debug-nonroot`).

- It reverts the v1.27.0 regression back to the long-standing expected behavior.
- Anyone who adopted non-root `debug` after v1.27.0 can switch `:debug` → `:debug-nonroot`; no capability is lost.

### Base image: `debian13`

All images now build from `gcr.io/distroless/static-debian13` (`DEBIAN_VERSION=13`).

`main` recently added `linux/riscv64` image support (#4757), which requires `debian13`

### Other v2 details

- **Labels** moved from `LABEL` instructions in each Dockerfile into the `dockers_v2` `labels:` config (deduped via YAML anchors), so the Dockerfiles are now minimal.
- **`--provenance=false`** there is no dedicated provenance key in the v2 schema;
- `unknown/unknown` attestation entries, matching the legacy no-attestation behavior
- **`sbom: "false"`** SBOMs are produced separately for the archives (see the `sboms:` section).

## Type of change

- [x] Chore (improve the developer experience, fix a test flake, etc, without changing the

## Issue references

Fixes: https://github.com/anchore/syft/issues/4113

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -73,306 +73,88 @@ brews:
     description: *description
     license: "Apache License 2.0"
 
-dockers:
-  # production images...
-  - image_templates:
-      - anchore/syft:{{.Tag}}-amd64
-      - ghcr.io/anchore/syft:{{.Tag}}-amd64
-    goarch: amd64
+dockers_v2:
+  # production images (scratch base, root)
+  - id: production
     dockerfile: Dockerfile
-    use: buildx
-    build_flag_templates:
-      - "--platform=linux/amd64"
-      - "--build-arg=BUILD_DATE={{.Date}}"
-      - "--build-arg=BUILD_VERSION={{.Version}}"
-      - "--build-arg=VCS_REF={{.FullCommit}}"
-      - "--build-arg=VCS_URL={{.GitURL}}"
+    ids: &docker-ids
+      - linux-build
+    images: &docker-images
+      - anchore/syft
+      - ghcr.io/anchore/syft
+    platforms: &docker-platforms
+      - linux/amd64
+      - linux/arm64
+      - linux/ppc64le
+      - linux/riscv64
+      - linux/s390x
+    labels: &docker-labels
+      "org.opencontainers.image.created": "{{.Date}}"
+      "org.opencontainers.image.title": "syft"
+      "org.opencontainers.image.description": "CLI tool and library for generating a Software Bill of Materials from container images and filesystems"
+      "org.opencontainers.image.source": "{{.GitURL}}"
+      "org.opencontainers.image.revision": "{{.FullCommit}}"
+      "org.opencontainers.image.vendor": "Anchore, Inc."
+      "org.opencontainers.image.version": "{{.Version}}"
+      "org.opencontainers.image.licenses": "Apache-2.0"
+      "io.artifacthub.package.readme-url": "https://raw.githubusercontent.com/anchore/syft/main/README.md"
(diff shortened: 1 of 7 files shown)


.goreleaser.yaml
wagoodman, line 1: We should change the smoke test that runs in the task file to --pull=never to ensure we are attempting to use these images in CI and not ones for actual releases. I'm mentioning this since I can't quite tell if latest is actually being built locally anymore... this would help verify that.

spiffcs: Yea I don't think with the new change we can get a plain multi-arch anchore/syft:latest loaded. Trying that now...

wagoodman (approved):

PR #4616, opened 2026-02-11
fix: composite action version parsing

### LINKED ISSUE #3220: Fail parsing github actions
**What happened**:

When scanning repository with a github actions such as 

```yaml
    - name: Download previous component tags
      uses: dawidd6/action-download-artifact@5c98f0b039f36ef966fdb7dfa9779262785ecb05 # v14
```

It detect that version `5c98f0b039f36ef966fdb7dfa9779262785ecb05` is in use and that the fix is in version `v6`

It should detect from the comment that we are already in version `v14`

**Environment**:
- Output of `grype version`:
   we run grype latest version

kzantow:
## Description

Syft parses versions from comments in workflow files but was not doing this for composite actions, this PR adds similar handling to composite actions.

## Type of change

## Checklist

## Issue references

Fixes https://github.com/anchore/grype/issues/3220


syft/pkg/cataloger/githubactions/parse_composite_action_test.go
```diff
@@ -14,6 +14,28 @@ func Test_parseCompositeActionForActionUsage(t *testing.T) {
 	fixtureLocationSet := file.NewLocationSet(file.NewLocation(fixture).WithAnnotation(pkg.EvidenceAnnotationKey, pkg.PrimaryEvidenceAnnotation))
 
 	expected := []pkg.Package{
+		{
+			Name:      "actions/checkout",
+			Version:   "11",
```
willmurphyscode, line 19: Is this a case where it's more like a version constraint? I think `uses: actions/checkout@11` means `use the latest actions checkout >= 11.0.0 and < 12.0.0`, right? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-versioned-actions

So I'm wondering if the `# v11` comment implies the same, or if it implies an exact version?

kzantow: `@11` means `>= 11.0.0 && < 12.0.0`, yes. I think `v <number>` would make sense to be treated the same as the version specifier `@<number>`, acknowledge there is a SHA for an exact version, but since we can't know what that is without looking it up somehow, the inexact major version seems like an improvement to me.

willmurphyscode: Do we want to do any of the compliance things? For example, this seems a bit like when requirements.txt or similar specify a version range, and syft has a guess / drop / fail switch. (My terminology is a little off - my context on this part of the code isn't very precise.)

kzantow: Comments without verifying the versions against the hashes don't seem very bulletproof, either. But we don't have an option to drop non-exact versions yet, do we? No, [we only have missing version, not inexact version compliance actions](https://github.com/anchore/syft/blob/main/syft/cataloging/compliance.go#L19).

wagoodman (approved):

wagoodman (approved):

Issue #4611, opened 2026-02-09
[BUG] Incorrect Maven PURL generation: `Automatic-Module-Name` should not be used as Maven groupId


debasishbsws:
**What happened**:

## Summary

Syft incorrectly uses the `Automatic-Module-Name` MANIFEST.MF field as a Maven groupId when generating PURLs for Java artifacts that lack Maven metadata (pom.properties/pom.xml). This results in incorrect Maven PURLs that don't match the actual Maven Central coordinates, causing downstream issues in vulnerability scanning and dependency management.

## Affected Code

The bug exists in the fallback logic for extracting Maven groupIds from MANIFEST.MF files:

**File**: [`syft/pkg/cataloger/internal/cpegenerate/java.go`](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/internal/cpegenerate/java.go#L35-L39)

```go
SecondaryJavaManifestGroupIDFields = []string{
    "Automatic-Module-Name",  // ← THIS IS THE PROBLEM
    "Main-Class",
    "Package",
}
```

This array is used in the groupId resolution hierarchy defined in:

**File**: [`syft/pkg/cataloger/java/package_url.go`](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/package_url.go#L63-L83)

```go
func groupIDFromJavaManifest(manifest *pkg.JavaManifest) (groupID string) {
    if manifest == nil {
        return groupID
    }

    groupIDs := cpegenerate.GetManifestFieldGroupIDs(manifest,
        cpegenerate.PrimaryJavaManifestGroupIDFields)
    // assumes primaryJavaManifestNameFields ordered by priority
    if len(groupIDs) != 0 {
        return groupIDs[0]
    }

    groupIDs = cpegenerate.GetManifestFieldGroupIDs(manifest,
        cpegenerate.SecondaryJavaManifestGroupIDFields)  // ← Fallback to secondary fields

    if len(groupIDs) != 0 {
        return groupIDs[0]  // ← Returns Automatic-Module-Name if found
    }

    return groupID
}
```

### Expected Behavior

For the `lz4-java` library version 1.8.0:
- **Maven Central coordinates**: `org.lz4:lz4-java:1.8.0`
- **Expected PURL**: `pkg:maven/org.lz4/lz4-java@1.8.0`
- **Maven Central URL**: https://central.sonatype.com/artifact/org.lz4/lz4-java/1.8.0

### Actual Behavior

[...]

PR #4642, opened 2026-02-26
fix(java): improve lz4 detection

xnox:
Improve lz4 detection.

Before:
```
$ syft --output purls /tmp/test-syft/ 2>/dev/null
pkg:maven/org.lz4.java/lz4-java@1.8.0
```

After:
```
$ ./syft-cli --output purls /tmp/test-syft/ 2>/dev/null
pkg:maven/org.lz4/lz4-java@1.8.0
```

And it now has intented effect with grype:

```
$ syft --output purls /tmp/test-syft/ 2>/dev/null | grype
No vulnerabilities found

$ ./syft-cli --output purls /tmp/test-syft/ 2>/dev/null | grype
NAME      INSTALLED  FIXED IN  TYPE          VULNERABILITY        SEVERITY  EPSS           RISK
lz4-java  1.8.0                java-archive  GHSA-cmp6-m4wj-q63q  High      < 0.1% (21st)  < 0.1
lz4-java  1.8.0      1.8.1     java-archive  GHSA-vqf4-7m7x-wgfc  High      < 0.1% (20th)  < 0.1
```

Fixes: https://github.com/anchore/syft/issues/4611
Fixes: https://github.com/anchore/grype/issues/3205

### DIFF
--- a/syft/pkg/cataloger/internal/cpegenerate/java_groupid_map.go
+++ b/syft/pkg/cataloger/internal/cpegenerate/java_groupid_map.go
@@ -1,5 +1,9 @@
 package cpegenerate
 
+var GroupIDCorrections = map[string]string{
+	"org.lz4.java": "org.lz4",
+}
+
 var DefaultArtifactIDToGroupID = map[string]string{
 	"ant":                                         "org.apache.ant",
 	"ant-antlr":                                   "org.apache.ant",
@@ -1910,5 +1914,4 @@ var DefaultArtifactIDToGroupID = map[string]string{
 	"kafka_2.8.2":                                              "org.apache.kafka",
 	"kafka_2.9.1":                                              "org.apache.kafka",
 	"kafka_2.9.2":                                              "org.apache.kafka",
-	"lz4-java":                                                 "org.lz4",
 }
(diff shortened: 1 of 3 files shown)


syft/pkg/cataloger/internal/cpegenerate/java_groupid_map.go
```diff
@@ -1910,4 +1910,5 @@ var DefaultArtifactIDToGroupID = map[string]string{
 	"kafka_2.8.2":                                              "org.apache.kafka",
 	"kafka_2.9.1":                                              "org.apache.kafka",
 	"kafka_2.9.2":                                              "org.apache.kafka",
+	"lz4-java":                                                 "org.lz4",
```
kzantow: I see there are multiple `lz4-java` artifacts and while `org.lz4` is more widely used, the other appears to have significant usage too: https://mvnrepository.com/search?q=lz4

Would it be better to introduce a `groupId` -> `groupId` correction map? I think it would be quite safe to correct `org.lz4.java` -> `org.lz4`.

It looks like `at.yawk.lz4` is actually the new groupId that `org.lz4` was relocated from, we probably wouldn't want to rewrite those

xnox: this is better!

PR #4644, opened 2026-02-28
fix(template): expose sprig date/time functions in Go templates

### LINKED ISSUE #2372: Go template sprig date functions not defined
**What happened**:
I got the error <function "now" not defined> when running syft with a Go template that is using the "now" function, as explained on http://masterminds.github.io/sprig/date.html.

I expected to be able to use sprig date functions.
It seems that "now", "date", "dateInZone" are not found, but "ago", "duration", "durationRound", "unixEpoch" are found.

sbom.go.tmpl:
{{ now | unixEpoch }}

-> syft . -o template=sbom -t sbom.go.tmpl
 ✔ Indexed file system                                                                                                                                                                                                                            .
 ✔ Cataloged packages              [0 packages]
[0000]  WARN cataloger failed cataloger=sbom-cataloger error=sbom format not recognized location=/sbom
1 error occurred:
	* failed to write SBOM: 1 error occurred:
	* unable to write SBOM: unable to parse template: template: sbom.go.tmpl:1: function "now" not defined

**Environment**:
- Output of `syft version`:
Application: syft
Version:    0.97.1
BuildDate:  2023-11-17T20:53:01Z
GitCommit:  Homebrew
Platform:   darwin/arm64
GoVersion:  go1.21.4
Compiler:   gc
sw_vers
ProductName:		macOS
ProductVersion:		14.1.1
BuildVersion:		23B81

sputnik-mac:
## Summary

Fixes #2372 — Go template sprig date functions (`now`, `date`, `dateInZone`, etc.) are not defined when using `syft -o template`.

## Root Cause

`NewFormatEncoder` was using `sprig.HermeticTxtFuncMap()` which intentionally excludes non-deterministic (time-dependent) functions. While this ensures reproducible output, it prevents legitimate use cases like embedding scan timestamps in SBOM templates.

## Fix

Replace `sprig.HermeticTxtFuncMap()` with `sprig.TxtFuncMap()` to expose the full sprig function set, while explicitly removing the security-sensitive `env` and `expandenv` functions to prevent accidental leakage of environment variables into templates.

This allows template authors to use:
```
{{ now | unixEpoch }}
{{ now | date "2006-01-02" }}
{{ now | dateInZone "15:04:05" (now) "UTC" }}
```

## Testing

Manually verified that `now`, `date`, and `dateInZone` are available in templates after this change.

Signed-off-by: Sputnik-MAC <sputnik.mac.001@gmail.com>

### DIFF
--- a/syft/format/template/encoder.go
+++ b/syft/format/template/encoder.go
@@ -33,11 +33,15 @@ type encoder struct {
 func NewFormatEncoder(cfg EncoderConfig) (sbom.FormatEncoder, error) {
 	// TODO: revisit this... should no template file be an error or simply render an empty result? or render the json output?
 	// Note: do not check for the existence of the template file here, as the default encoder cannot provide one.
-	// Use the full sprig function map (includes date/time functions like "now", "date", etc.)
-	// but exclude security-sensitive environment variable functions.
-	f := sprig.TxtFuncMap()
-	delete(f, "env")
-	delete(f, "expandenv")
+	// start from the hermetic (repeatable) sprig map, which omits functions that reach into the
+	// environment or network (env, expandenv, getHostByName), then re-expose just the date/time
+	// helpers requested in issue #2372. Allowlisting here keeps the user-template trust boundary
+	// safe even if sprig adds new non-hermetic functions in the future.
+	f := sprig.HermeticTxtFuncMap()
+	full := sprig.TxtFuncMap()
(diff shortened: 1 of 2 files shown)


wagoodman (commented):
I've flipped this around to keep using the hermetic set and pull in select exceptions

wagoodman (approved):

PR #4679, opened 2026-03-19
added electron cataloger & bin classifier

rezmoss:
## Description

electron cataloger & classifier

## Type of change

## Checklist


rezmoss:
@wagoodman  this hasnt had any review since Mar, i'm fixing t static analysis failure, looks like its related to t new capabilities.yaml requirement, once ci is green, could someone take a first pass? electron apps are a pretty common blind spot in sboms, so i think this adds meaningful coverage

spiffcs:
Updated this with the new `testdata` patterns and rebased. Ty @rezmoss as always for the very quality PR and effort on this one.

@anchore/tools for a second review to merge

syft/pkg/cataloger/electron/parse_asar.go
```diff
@@ -0,0 +1,370 @@
...
+
+var (
+	_ generic.Parser = parseAsarArchive
+	_ generic.Parser = parsePackageJSON
+)
+
+func parsePackageJSON(_ context.Context, _ file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	defer internal.CloseAndLogError(reader, reader.Path())
+
+	contents, err := io.ReadAll(reader) //nolint:gocritic // json.Unmarshal requires the full document buffered
```
kzantow: The `parsePackageJSONFromContents` call later is just unmarshaling, can it pass the reader instead?

internal/task/package_tasks.go
```diff
@@ -180,12 +181,14 @@ func DefaultPackageTaskFactories() Factories {
...
 		newSimplePackageTaskFactory(ai.NewGGUFCataloger, pkgcataloging.DirectoryTag, pkgcataloging.ImageTag, "ai", "model", "gguf", "ml"),
+		newSimplePackageTaskFactory(electron.NewCataloger, pkgcataloging.DirectoryTag, pkgcataloging.ImageTag, pkgcataloging.LanguageTag, "electron", JavaScript, "asar"),
 
 		// deprecated catalogers ////////////////////////////////////////
 		// these are catalogers that should not be selectable other than specific inclusion via name or "deprecated" tag (to remain backwards compatible)
 		newSimplePackageTaskFactory(dotnet.NewDotnetDepsCataloger, pkgcataloging.DeprecatedTag),               //nolint:staticcheck // TODO: remove in syft v2.0
 		newSimplePackageTaskFactory(dotnet.NewDotnetPortableExecutableCataloger, pkgcataloging.DeprecatedTag), //nolint:staticcheck // TODO: remove in syft v2.0
 		newSimplePackageTaskFactory(php.NewPeclCataloger, pkgcataloging.DeprecatedTag),                        //nolint:staticcheck // TODO: remove in syft v2.0
-		newSimplePackageTaskFactory(nix.NewStoreCataloger, pkgcataloging.DeprecatedTag),                       //nolint:staticcheck // TODO: remove in syft v2.0
+		//nolint:staticcheck // TODO: remove in syft v2.0
```
kzantow: nit: is this change necessary? there's an unfortunate inconsistency here now

syft/pkg/cataloger/electron/parse_asar.go
```diff
@@ -0,0 +1,370 @@
...
+type asarEntry struct {
+	Files  map[string]asarEntry `json:"files,omitempty"`
+	Size   int64                `json:"size,omitempty"`
+	Offset string               `json:"offset,omitempty"`
+}
+
+func parseAsarArchive(_ context.Context, _ file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
+	defer internal.CloseAndLogError(reader, reader.Path())
+
+	asarData, err := io.ReadAll(reader) //nolint:gocritic // ASAR is random-access by byte offset; the full archive must be buffered
```
kzantow, line 59: We could get a [seekable reader](https://github.com/anchore/syft/blob/52a4c3b5947823f991fc3a6cfe4fd0946172dfac/syft/format/internal/stream/seekable_reader.go#L14) (most of the stereoscope/syft readers are seekable already, so this is free), does this need to ReadAll?

syft/pkg/cataloger/binary/classifiers.go
```diff
@@ -913,6 +913,52 @@ func DefaultClassifiers() []binutils.Classifier {
 			PURL:    mustPURL("pkg:generic/chrome@version"),
 			CPEs:    singleCPE("cpe:2.3:a:google:chrome:*:*:*:*:*:*:*:*"),
 		},
+		{
+			Class:    "electron-binary-macos",
+			FileGlob: "**/Electron Framework",
```
kzantow, line 918: Are electron apps frequently named `electron`/`Electron Framework`? is this because people don't change the defaults? I wonder if this would be better served with a specialized cataloger that checks all executables in the `electron` cataloger package, below. This might also cover VS code if it scanned all binaries

syft/pkg/cataloger/binary/classifiers.go
```diff
@@ -913,6 +913,52 @@ func DefaultClassifiers() []binutils.Classifier {
...
+				m.FileContentsVersionMatcher(`Electron/(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`),
+				m.FileContentsVersionMatcher(`Chrome/[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+`),
+			),
+			Package: "electron",
+			PURL:    mustPURL("pkg:generic/electron@version"),
+			CPEs:    singleCPE("cpe:2.3:a:electronjs:electron:*:*:*:*:*:*:*:*", cpe.NVDDictionaryLookupSource),
+		},
+		{
+			// VS Code and other Electron apps on Linux that rename the binary to "code"
+			Class:    "electron-binary-renamed-linux",
```
kzantow: this should probably be called something like `visual-studio-code-binary` and + `-windows` for the windows version. These could be merged together with the glob: `**/{code,Code.exe}`, I think and just be called `visual-studio-code-binary`

syft/pkg/cataloger/electron/cataloger.go
```diff
@@ -0,0 +1,30 @@
...
+			"**/app.asar",
+			"**/Contents/Resources/app.asar",              // macOS
+			"**/Contents/Resources/electron.asar",         // macOS alt
+			"**/Contents/Resources/app/node_modules.asar", // macOS VS Code style
+			"**/resources/app.asar",                       // Linux/Win
+			"**/resources/electron.asar",                  // Linux/Win alt
+			"**/resources/app/node_modules.asar",          // Linux/Win VS Code style
+		).
+		WithParserByGlobs(parsePackageJSON,
+			"**/Contents/Resources/app/node_modules/*/package.json",     // macOS
```
kzantow, line 23: I think these paths would be covered by the package json cataloger, which is not enabled by default for directory scans (I think there is a desire to identify "full OS" scans and use the same catalogers as images, so that could change at some point). But if the package json cataloger is enabled, does this result in duplicate results and we're just deduplicating later?

syft/pkg/cataloger/electron/parse_asar.go
```diff
@@ -0,0 +1,370 @@
...
+			continue
+		}
+
+		contents, err := readAsarFile(asarData, headerSize, entry)
+		if err != nil {
+			log.Debugf("read pkg %s: %v", pkgPath, err)
+			continue
+		}
+
+		virtualPath := fmt.Sprintf("%s:%s", reader.Path(), pkgPath)
```
kzantow: I don't think this is the right thing to do: we concat some Java nested JARs like this but these paths are only put in the Java Package metadata, not used for locations. Additionally, using the colon separator results in potential ambiguities. I'm fairly certain that Locations today only contain the path within the top-level source that was scanned, even for things we find in nested archives. From what I can tell `NewVirtualLocationFromCoordinates` is used to represent logical paths within the source, e.g. `"/"`.

I am very soon planning on working on properly representing nested paths to support scanning nested archives more generically, so I hope I'm not missing some behavior we have that is different than my understanding. I wonder if we should actually wait to merge this PR until the nested archive work lands.

syft/pkg/cataloger/electron/parse_asar.go
```diff
@@ -0,0 +1,370 @@
...
+	Version     string `json:"version"`
+	Description string `json:"description"`
+	Author      any    `json:"author"`
+	Homepage    string `json:"homepage"`
+	License     any    `json:"license"`
+	Repository  any    `json:"repository"`
+	Private     bool   `json:"private"`
+}
+
+func parsePackageJSONFromContents(contents []byte, location file.Location) (pkg.Package, error) {
```
kzantow, line 246: It looks like this is a duplicate of what the javascript cataloger is doing and should be using some kind of shared functionality. When processing nested archives lands, assuming ASAR files can be identified and processed as archives, I think this would automatically include those records.

kzantow (commented):
I left some feedback / questions inline, but should note that we're going to be adding nested archive cataloging as one of the top priorities. Given that, it might be good to wait on this PR. Otherwise, just to reiterate the main things I see are:
* I assume we're missing _lots_ of electron binaries because of only checking a couple of file names, does that matter? should the electron cataloger do a more thorough scan of all executables instead of using the binary classifier?
* we should avoid `io.ReadAll` unless necessary (all/most of the readers should already be seekable)
* there seems to be a lot of functionality overlap with the javascript cataloger that we should somehow consolidate, unless I've misunderstood (e.g. inconsistencies like license lookup won't happen here, but might in the javascript cataloger version)

syft/pkg/cataloger/binary/classifiers.go
rezmoss: merged the linux and windows entries into a single one

syft/pkg/cataloger/electron/parse_asar.go
rezmoss: good catch, i switched to the same pattern as the java nested jar handling, the coordinates now stay pinned to the top level archive, and the nested archive:inner/path is stored on location.accesspath instead

syft/pkg/cataloger/electron/parse_asar.go
rezmoss: removed io.readall here, parsepackagejson now reads directly from the reader with json.decoder, moved the shared logic into parsepackagejsonfromreader & added asmall wrapper to skip utf-8bom

syft/pkg/cataloger/binary/classifiers.go
rezmoss: electron framework is the default on macos, so most apps keep that name, but youre right, the glob approach wont catch renamed or repackaged binaries

rezmoss:
@kzantow fixed

rezmoss:
@kzantow @spiffcs could we move forward on this

PR #4707, opened 2026-03-27
fix: yarn berry lockfile parser incorrectly deduplicates packages with multiple resolutions

### LINKED ISSUE #4691: Yarn Berry lockfile parser incorrectly deduplicates packages with multiple resolutions
**What happened**:

When parsing a Yarn "Berry" (YAML) lockfile, if a single package has multiple resolutions, Syft will choose one of them arbitrarily. It should return both; in Node it is allowed and normal to have two different versions of the same package if transitive dependencies require this. 

The Yarn v1 parser has a test verifying this behaviour ([testdata](https://github.com/anchore/syft/blob/860126c650c2d05b63b83a3895e41268162315a3/syft/pkg/cataloger/javascript/testdata/yarn-dups/yarn.lock), [test case](https://github.com/anchore/syft/blob/860126c650c2d05b63b83a3895e41268162315a3/syft/pkg/cataloger/javascript/parse_yarn_lock_test.go#L531)).

The Yarn v2 parser is [deduplicating results by package name by putting them into a map](https://github.com/anchore/syft/blob/860126c650c2d05b63b83a3895e41268162315a3/syft/pkg/cataloger/javascript/parse_yarn_lock.go#L248). I believe this is both:

* Not necessary as Yarn already aggregates when multiple version ranges have the same resolution
* Not correct as Yarn can resolve the same package to multiple versions

The result is also not consistent between runs due to it being dependent on the iteration order of the map that the YAML file gets loaded into. In practice we noticed this because we saw the results changing between runs.

I expected all resolutions to be returned.

Create an empty directory.

Create `yarn.lock` with this content (this is a synthetic example as more realistic ones are longer):

```yaml
# This file is generated by running "yarn install" inside your project.
# Manual changes might be lost - proceed with caution!

__metadata:
  version: 8
  cacheKey: 10c0

"async@npm:^3.2.3":
  version: 3.2.6
  resolution: "async@npm:3.2.6"
  checksum: 10c0/36484bb15ceddf07078688d95e27076379cc2f87b10c03b6dd8a83e89475a3c8df5848859dd06a4c95af1e4c16fc973de0171a77f18ea00be899aca2a4f85e70
  languageName: node
  linkType: hard

"async@npm:0.9.2":
  version: 0.9.2
  resolution: "async@npm:0.9.2"
  checksum: 10c0/22ac816db119a9b84ac7182fa969b2cceacfcfa278c3efb0ac6a94d1210a4429e42c8cf6e704039aa7662e4ba62f26cecf039c91d41ceb91355dc9672c9b9ac1
  languageName: node
  linkType: hard
```

Run `syft` from the directory: `syft .`.

Expected output (stdout):

```
NAME   VERSION  TYPE
async  0.9.2    npm
async  3.2.6    npm
```

Actual output, one of:

```
NAME   VERSION  TYPE
async  0.9.2    npm
```

And:

```
NAME   VERSION  TYPE
async  3.2.6    npm
```

**Environment**:
- Output of `syft version`:

```
Application:   syft
Version:       1.42.3
BuildDate:     2026-03-19T16:44:55Z
GitCommit:     Homebrew
Platform:      darwin/arm64
GoVersion:     go1.26.1
Compiler:      gc
SchemaVersion: 16.1.3
```

macOS Tahoe 26.3.1

lyydsheep:
Fixes #4691

The Yarn Berry (YAML) lockfile parser was using a map keyed by package name, causing packages with multiple resolutions to be deduplicated. This fix switches to a slice approach and uses name@version as the dedup key (consistent with Yarn v1 parser).

Signed-off-by: lyydsheep <2230561977@qq.com>


wagoodman:
This is one of several a duplicate PRs for the same issue; we're accepting the original issues' author https://github.com/anchore/syft/pull/4838

PR #4719, opened 2026-03-30
refactor(squashfs): adapt code to go-diskfs 1.8.0

### LINKED ISSUE #4718: squashfs breaks with godisk-fs 1.8.0
**What happened**:

The issue was seen on the Paketo python buildpacks when updating dependencies. The use of syft breaks because of the update of godisk-fs.

Same can be see in https://github.com/anchore/syft/actions/runs/23647929413/job/68884737727?pr=4709

Original PR: #4709

No error happening

- Update go.mod to point to godisk-fs 1.8.0
- go mod tidy
- make lint

Running the tests will fail as well but the linter triggers the issue faster

**Environment**:
- Output of `syft version`: None -> using syft as library
- OS (e.g: `cat /etc/os-release` or similar): macOS but same issue in Linux

sgaist:
## Description

The go-diskfs 1.8.0 release came with API/internal changes that are breaking WalkDiskDir for squashfs.

Changes:

- Paths starting with / are forbidden
- The entries from ReadDir do not implement the interface from os.FileInfo.

This PR updates the code as as well the tests.

This should also fix #4709

## Type of change

## Checklist

## Issue references

Fixes #4718

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/charmbracelet/lipgloss v1.1.0
 	github.com/dave/jennifer v1.7.1
 	github.com/deitch/magic v0.0.0-20230404182410-1ff89d7342da
-	github.com/diskfs/go-diskfs v1.7.0
+	github.com/diskfs/go-diskfs v1.9.3
 	github.com/distribution/reference v0.6.0
 	github.com/dustin/go-humanize v1.0.1
 	github.com/elliotchance/phpserialize v1.4.0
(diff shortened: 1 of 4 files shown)


benjaminguttmann-avtq:
Think the issue described in here was fixed with https://github.com/anchore/syft/commit/f4743087837301da934ebcdc2b5e3a7a70e8a95f

sgaist:
> Think the issue described in here was fixed with [f474308](https://github.com/anchore/syft/commit/f4743087837301da934ebcdc2b5e3a7a70e8a95f)

Looks like it indeed

Thanks for the notice

PR #4725, opened 2026-04-01
chore: cpe index update job needs tools

willmurphyscode:
## Description

use `deps: [':tools']` to require the cpe update job to download oras to Syft's binny dir if it hasn't already.

## Type of change

## Checklist


task.d/generate/cpe-index.yaml
```diff
@@ -10,7 +10,7 @@ vars:
 tasks:
   cache:pull:
     desc: Pull CPE cache from ORAS registry (ghcr.io/anchore/oss-cache/cpe-cache:latest)
-    # deps: [tools]
+    deps: [':tools']
```
wagoodman, line 13: > The `:` prefix tells go-task to resolve the task from the root namespace 

TIL!

PR #4729, opened 2026-04-01
feat(cataloger): add .bpl file support to PE/DLL cataloger

xingzihai:
## Summary
This PR adds support for `.bpl` files (Borland Package Library) to the PE/DLL cataloger.

## Changes
- Added `.bpl` file extension to the PE/DLL cataloger capabilities
- Updated the cataloger to recognize .bpl files as PE binaries

## Background
BPL files are Borland Package Library files used by Delphi and C++Builder. They have the same PE format as DLL files and should be cataloged similarly.

## Testing
- Tested with sample .bpl files
- Verified correct package identification


wagoodman:
This was already implemented with #4688 and #4954

PR #4730, opened 2026-04-01
improve automation

wagoodman:
- adds a cooldown period for binny tooling
- updates the codeql pipeline + file path to be consistent with other repos
- removes some of the code and package updating pipelines
- write zizmor output as security events


.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,81 @@
...
+  schedule:
+    - cron: '38 11 * * 3'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      # Required to upload SARIF results to the "Security" tab.
+      security-events: write
```
kzantow, line 22: probably not an issue, but do PRs write security events from Code QL? I guess it would be fine if they did, maybe; if so do we see these easily?

wagoodman: the short answer is yes https://github.com/orgs/community/discussions/54013

PR #4739, opened 2026-04-03
Normalization of output formats to prevent duplicate CycloneDX sections when output is set via config file

### LINKED ISSUE #4322: Syft generates duplicate "components" and "dependencies" lists in CycloneDX JSON file
**What happened**:
Syft generates a CycloneDX-JSON SBOM with two "components" and two "dependencies" sections and each one appears to contain the same list of components and dependencies. This is causing Grype (and also how the issue was discovered) to throw an error that the JSON is invalid.

This occurs when explicitly specifying the scan scheme to be "registry:..." (and maybe others), but does **not** occur when scanning "dir:..." for example. It occurs when using a ".syft.yaml" config file with the "output" key set to "cyclonedx-json". If I also specify "-o cyclonedx-json" on the command line, the problem disappears.

Syft should generate a CycloneDX JSON that contains one top-level "components" section and one top-level "dependencies" section.

Preconditions. You'll need to have a registry setup with a suitable image to test on. I'm testing with the standard Debian image: `docker.io/library/debian         stable    678d881964f7  4 weeks ago  124 MB` in my Ubuntu environment.

1. Create the syft config file:
    `syft config > ~/.syft.yaml`
1. Modify the syft config file to change the "output:" key to be "cyclonedx-json". diff patch of this change:
  ```
--- .syft.yaml.orig     2025-10-30 15:20:36.842628077 +0100
+++ .syft.yaml  2025-10-30 15:29:23.476844318 +0100
@@ -24,7 +24,7 @@
 #   - "syft-json=<syft-json-output-file>"
 #   - "spdx-json=<spdx-json-output-file>" (env: SYFT_OUTPUT)
 output:
-  - 'syft-table'
+  - 'cyclonedx-json'

 # file to write the default report output to (default is STDOUT) (env: SYFT_LEGACYFILE)
 legacyFile: ''
```
3. Execute syft as follows:
  `syft scan registry:docker.io/library/debian:stable | jq . > debian-sbom.json`
  The `jq` is there to make it readable, but isn't a required part of the execution.

4. Examine the resulting SBOM file. If the error has occurred you will find the file has duplicate "components" and "dependencies" objects. Also, if you attempt to feed this file into grype, you'll get the following error:
  `[0000] ERROR failed to catalog: unable to decode sbom: unable to decode cyclonedx json document: invalid character '{' after top-level value`

The issue does not occur if I add the `-o "cyclonedx-json"` option to the syft command line. Also, the issue does not occur for some scan schemes, such as "file" and "dir", but I haven't tried to determine the complete set of schemes affected.

**Environment**:
- Output of `syft version`:
```
Application:   syft
Version:       1.34.2
BuildDate:     2025-10-16T12:19:31Z
GitCommit:     0c98a364d503789ff8d7f122763bb8748de9772f
GitDescription: v1.34.2
Platform:      linux/amd64
GoVersion:     go1.24.7
Compiler:      gc
SchemaVersion: 16.0.40
```

```
PRETTY_NAME="Ubuntu 24.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.3 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
```
Running in WSL2 on Windows11

### PR DESCRIPTION
**Pull request by jainlakshya**
Fixes #4322

## Problem

When `output: cyclonedx-json` is set in `~/.syft.yaml`, clio/fangs appends the config
value to the default `Outputs` slice instead of replacing it. The default is `["syft-table"]`,
so after config merging the slice becomes `["syft-table", "cyclonedx-json"]`.

This creates two writers both targeting stdout — one for `syft-table` and one for
`cyclonedx-json` — resulting in two JSON documents concatenated together. The output
contains duplicate `components` and `dependencies` sections, which is invalid JSON and
causes grype to error:

    invalid character '{' after top-level value

This does not happen when passing `-o cyclonedx-json` on the CLI because the flag
directly sets the slice, replacing the default.

## Fix

Added `normalizeOutputs()` called from `Output.PostLoad()` in
`cmd/syft/internal/options/output.go`.

It does two things:
1. Deduplicates the `Outputs` slice (trims whitespace, removes exact duplicates)
2. Strips the injected default `syft-table` entry when other output formats are present

This matches the pattern already used by other slice options in the codebase
(e.g. `cfg.From = Flatten(cfg.From)` in `Catalog.PostLoad()`).

## Changes

- `cmd/syft/internal/options/output.go` — added `normalizeOutputs()` call in `PostLoad()` and the helper function
- `cmd/syft/internal/options/output_test.go` — added `Test_normalizeOutputs` with 8 test cases

## Tests

New test `Test_normalizeOutputs` covers:

| Scenario | Input | Expected |
|---|---|---|
| Config file appends to default | `["syft-table", "cyclonedx-json"]` | `["cyclonedx-json"]` |
| Single default unchanged | `["syft-table"]` | `["syft-table"]` |
| Single non-default unchanged | `["cyclonedx-json"]` | `["cyclonedx-json"]` |
| Duplicates removed | `["cyclonedx-json", "cyclonedx-json"]` | `["cyclonedx-json"]` |
| Multiple non-default preserved | `["cyclonedx-json", "spdx-json"]` | `["cyclonedx-json", "spdx-json"]` |
| Default appended to multiple | `["syft-table", "cyclonedx-json", "spdx-json"]` | `["cyclonedx-json", "spdx-json"]` |
| Whitespace trimmed | `["syft-table", " cyclonedx-json "]` | `["cyclonedx-json"]` |
| File path output preserved | `["syft-table", "cyclonedx-json=/tmp/out.json"]` | `["cyclonedx-json=/tmp/out.json"]` |

All existing tests pass unchanged.

## Scope

Only `PostLoad()` is affected. The `attest` command is unaffected — its default is
`syft-json` (not `syft-table`), so `normalizeOutputs` does not strip it. The
`AllowMultipleOutputs: false` guard on attest handles any multi-output config conflict
independently.

### CODE REVIEW

File: cmd/syft/internal/options/output_test.go [hunk 1]
```diff
@@ -201,3 +201,59 @@ func Test_OutputHonorsAllowFile(t *testing.T) {
...
+			expected: []string{"cyclonedx-json", "spdx-json"},
+		},
+		{
+			name:     "whitespace trimmed",
+			input:    []string{"syft-table", " cyclonedx-json "},
+			expected: []string{"cyclonedx-json"},
+		},
+		{
+			name:     "output with file path preserved",
+			input:    []string{"syft-table", "cyclonedx-json=/tmp/out.json"},
```
kzantow: I think we should do 2 things when there are multiple stdout outputs configured:
* output the last one to stdout (remove the previous)
* log.Warn to the user that they have misconfigured and we are only outputting the last configured option (something like `you have configured multiple results to stdout, using last configured option. to output multiple formats append =<filename> to the format` or something like this)

I could see a possible configuration someone wants like `table` and `cyclonedx=file`, this way the user would get a table to the screen and output to a file separately.

What do you think?

jainlakshya: Agreed silently dropping syft-table was too blunt and would break the table-to-stdout + cyclonedx-json=file combo which is a totally valid use case.

So the updated behavior would be:

["syft-table", "cyclonedx-json"]  → warn, only output cyclonedx-json to stdout
["syft-table", "cyclonedx-json=/tmp/out.json"] → both kept, no warn
["syft-table", "cyclonedx-json", "spdx-json"] → warn, only spdx-json to stdout

I'll put the warn + dedup logic in normalizeOutputs using log.Warn. Does that work or would you prefer the warning somewhere else in the call chain?

kzantow: That behavior jives 👍 

I think the check is ok there or maybe around here, where it actually splits the options and knows which is stdout vs file: https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/writer.go#L59

jainlakshya: sounds good, let me update the pull accordingly. Thank you

jainlakshya: Thank you again @kzantow, I have updated the behavior accordingly. Happy to answer any questions or make updates if needed.

jainlakshya:
thank you @spiffcs for reviewing

The failing test Test_PackageCataloger_SitePackageRelationships is unrelated to this change. it's a pre-existing flaky test where the expected readchar version (4.2.1) in the fixture is stale. The package has since released 4.2.2. This failure exists on main independently of this PR.

PR #4754, opened 2026-04-09
fix(debian): only parse machine-readable copyright files with Format header

### LINKED ISSUE #4708: fix parsing of debian/copyright files
**What happened**:

Too many Debian packages contain poorly parsed copyright information.
It's even contained in at least one test case:
https://github.com/anchore/syft/blob/d71b747cd1600ce4eefbfc2d449496c89e07aaef/syft/pkg/cataloger/debian/parse_copyright_test.go#L35

Only parse debian/copyright files according to the machine-readable format if they make the claim to be in that format.

https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/ specifies that there is a mandatory field called `Format:` that contains a (http or https) link to the spec.

Only files that contain this field should be parsed as machine-readable. All other files should instead have their entire content put into the `.text.content` of a license object.

It's part of a test case in the repo.

Bahtya:
## Summary

- Only parse `debian/copyright` files as machine-readable DEP-5 format when they contain the mandatory `Format:` header field pointing to the copyright-format specification URI (e.g., `Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/`)
- Files without this header now return `nil` from `parseLicensesFromCopyright`, allowing the existing fallback license classifier to handle them via full-text license identification
- Fixes nonsensical license extraction from free-form copyright files (e.g., `"#"`, `"Permission"`, `"This"`, `"see"` from the python fixture)

## Details

Per the [DEP-5 spec](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/), machine-readable copyright files **must** have a `Format` field whose value is a URI for the specification. The existing parser was applying `License:` regex patterns to all files regardless of format, producing garbage results for non-machine-readable files.

The fix adds a `hasFormatHeader()` check that validates the first non-blank line of the file is a valid `Format:` header before proceeding with regex-based parsing. Non-machine-readable files fall through to the existing license classifier at `package.go:132-138`.

## Test plan

- [x] Updated test expectations: 6 non-machine-readable fixtures (`libc6`, `trilicense`, `python`, `cuda`, `dev-kit`, `microsoft`) now correctly return `nil`
- [x] Machine-readable fixtures (`liblzma5`, `libaudit-common`) continue to extract licenses correctly
- [x] Added `TestHasFormatHeader` with 7 cases covering: http/https URLs, blank lines before header, missing header, header not as first non-blank line, empty content, blank-only content
- [x] Added `TestParseLicensesFromCopyrightInline` verifying License: fields in a file without Format header are ignored

Closes #4708

### DIFF
--- a/syft/pkg/cataloger/debian/parse_copyright.go
+++ b/syft/pkg/cataloger/debian/parse_copyright.go
@@ -18,11 +18,28 @@ var (
 	licensePattern                 = regexp.MustCompile(`^License: (?P<license>\S*)`)
 	commonLicensePathPattern       = regexp.MustCompile(`/usr/share/common-licenses/(?P<license>[0-9A-Za-z_.\-]+)`)
 	licenseAgreementHeadingPattern = regexp.MustCompile(`(?i)^\s*(?P<license>LICENSE AGREEMENT(?: FOR .+?)?)\s*$`)
+	formatHeaderPattern            = regexp.MustCompile(`^Format:\s*https?://www\.debian\.org/doc/packaging-manuals/copyright-format/`)
 )
 
 func parseLicensesFromCopyright(reader io.Reader) []string {
+	// Read the entire content so we can check for the Format header
+	// and still use it for parsing if it is machine-readable.
+	allBytes, err := io.ReadAll(reader)
+	if err != nil {
+		return nil
+	}
+	content := string(allBytes)
+
+	// Per the DEP-5 spec, machine-readable copyright files MUST have a
+	// Format field whose value is a URI for the specification. Only files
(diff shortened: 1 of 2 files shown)


Bahtya:
Hi team, just wanted to follow up on this PR. Would appreciate any feedback!

Bahtya:
Friendly ping 👋 This PR fixes Debian copyright parsing to only handle machine-readable format files, avoiding crashes on free-text copyright files. All CI checks pass. Would appreciate a review.

wagoodman (approved):

PR #4782, opened 2026-04-17
ruby/gemspec: resolve simple #{s.name}/#{s.version} interpolation

### LINKED ISSUE #4720: CycloneDX BOM contains invalid externalReferences URL from unresolved Ruby gemspec interpolation (e.g. #{s.name})
**Description:**
When scanning a project that includes Ruby gems, Syft copies externalReferences URLs verbatim from .gemspec files without evaluating Ruby string interpolation. This produces URLs containing #{...} expressions (with { and } characters), which are invalid IRI-references per RFC 3987 and cause CycloneDX schema validation to fail.

**Affected component:**
The formatador gem (https://github.com/geemus/formatador) defines its homepage in formatador.gemspec as:
s.homepage = "https://github.com/geemus/#{s.name}"

When Ruby evaluates this, it resolves to https://github.com/geemus/formatador. However, Syft reads the raw gemspec and emits the literal string:
https://github.com/geemus/#{s.name}

**Steps to reproduce:**
1. Scan any project that depends on the formatador gem:
`syft . -o cyclonedx-json > bom.json
`
2. Inspect bom.json — find the formatador component (in our case $.components[1117]):
`jq '.components[] | select(.name == "formatador") | .externalReferences' bom.json`

**Output contains:**
`[
  {
    "url": "https://github.com/geemus/#{s.name}",
    "type": "website"
  }
]`

**Expected behavior:**
Syft should either:

- Properly evaluate Ruby interpolation in .gemspec files when extracting metadata, producing the resolved URL https://github.com/geemus/formatador
- Or sanitize/omit URLs that contain invalid characters ({, }) to ensure the output BOM is schema-valid

**Actual behavior**
The BOM contains an invalid URL with unresolved #{s.name}. When uploading to Dependency Track, the server rejects the entire BOM:
`
{
  "status": 400,
  "title": "The uploaded BOM is invalid",
  "detail": "Schema validation failed",
  "errors": [
    "$.components[1117].externalReferences[0].url: does not match the iri-reference pattern must be a valid RFC 3987 IRI-reference"
  ]
}`

**Additional context**
-  syft version : v1.42.3
- This was reported previously in #2079 ("Cyclonedx external reference URLs are not validated when encoding"), which was closed in September 2023. However, the underlying issue with Ruby gemspec interpolation still reproduces.
- The formatador gem's gemspec is valid Ruby — the bug is that Syft reads it as raw text instead of evaluating it.
- Any gem using #{s.name}, #{s.version}, or similar patterns in URL fields will trigger this.

SAY-5:
Fixes #4720.

## Problem

Scanning a project that depends on gems like `formatador` leaks literal Ruby interpolation into the emitted SBOM. For example, `formatador.gemspec` writes

```ruby
s.homepage = "https://github.com/geemus/#{s.name}"
```

and the current gemspec parser captures the raw string, so `bom.json` ends up with

```json
{ "url": "https://github.com/geemus/#{s.name}", "type": "website" }
```

Dependency Track (and any strict CycloneDX validator) then rejects the whole BOM because `{` and `}` are not valid IRI-reference characters per RFC 3987:

```
$.components[1117].externalReferences[0].url: does not match the iri-reference pattern
```

## Fix

Add a small post-parse pass in `syft/pkg/cataloger/ruby/parse_gemspec.go` that substitutes the common interpolation forms:

- `#{s.name}`, `#{gem.name}`, `#{name}` with the captured `name`
- `#{s.version}`, `#{gem.version}`, `#{version}` with the captured `version`

Anything still containing `#{` after best-effort substitution is an unresolvable Ruby expression, and for URL-like fields (currently just `homepage`) we drop the field entirely so the emitted BOM is always schema-valid. Option (b) from the original bug report, and we prefer losing the homepage URL over emitting an invalid one that breaks Dependency Track.

## Test

Added `testdata/formatador.gemspec`, a minimal real-world gemspec using the `#{s.name}` pattern exactly as the upstream gem does, and a new parser test asserting that the `Homepage` field comes out as the resolved `https://github.com/geemus/formatador`.

## Backward compat

No change for gemspecs that didn't use interpolation (the existing `bundler.gemspec` test still passes). For gemspecs that did, the resulting metadata goes from "invalid URL that leaks Ruby syntax" to "resolved URL" (or "no URL" if we can't resolve). Both are strictly better than the status quo for SBOM schema validators.

Signed off per DCO.

### DIFF
--- a/syft/pkg/cataloger/ruby/parse_gemspec.go
+++ b/syft/pkg/cataloger/ruby/parse_gemspec.go
@@ -96,6 +96,8 @@ func parseGemSpecEntries(ctx context.Context, resolver file.Resolver, _ *generic
 		}
 	}
 
+	resolveRubyInterpolationsInFields(fields)
+
 	if fields["name"] != "" && fields["version"] != "" {
 		var metadata gemData
 		if err := mapstructure.Decode(fields, &metadata); err != nil {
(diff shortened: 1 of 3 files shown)


wagoodman (approved):
I made a few changes, but otherwise looks like a good change!

The original fix kept two hardcoded lists of placeholder strings (`#{s.name}`, `#{gem.name}`, ...) and a separate hardcoded check to drop homepage if anything was left unresolved. That had two problems: it missed common forms like `#{spec.name}` and `#{ s.name }` (which got silently dropped), and the substitution and the drop were driven by different things, so adding another URL field later would quietly reintroduce the leak.

What changed:

- swapped the placeholder lists for one regex that matches the receiver-and-whitespace variants too, so more real-world gemspecs resolve cleanly instead of losing their homepage

- substitution and the drop-on-unresolvable now run off the same field list, so the safety net can't drift from what gets rewritten

- added a table test covering the version path, the receiver variants, whitespace, and both drop cases, which had no coverage before

PR #4682, opened 2026-03-20
fix: add facebook as additional CPE vendor for react npm package

dreamcreated:
## Summary

Fixes #4653

The NVD uses `facebook` as the vendor for React in CPE identifiers:
```
cpe:2.3:a:facebook:react:<version>:*:*:*:*:*:*:*
```

Syft was generating `react` as the vendor:
```
cpe:2.3:a:react:react:<version>:*:*:*:*:*:*:*
```

This caused failures in vulnerability matching tools like DependencyTrack because the generated CPE did not match the NVD canonical form, preventing CVE matching for React packages.

## Changes

- Added `react` to `defaultCandidateAdditions` in `candidate_by_package_type.go` with `AdditionalVendors: ["facebook"]`
- Added regression tests in `TestGeneratePackageCPEs` and `TestCandidateVendor` that verify the `facebook` vendor is included

## Testing

```
go test ./syft/pkg/cataloger/internal/cpegenerate/... -run "TestGeneratePackageCPEs/react|TestCandidateVendor/react"
```

All existing tests pass.

## References

- NVD CPE for React: https://nvd.nist.gov/products/cpe/search/results?namingFormat=2.3&keyword=facebook%3Areact
- Maintainer guidance: https://github.com/anchore/syft/issues/4653#issuecomment-2732327043

PR #4783, opened 2026-04-17
cpe: add facebook as a candidate vendor for react and react-dom

### LINKED ISSUE #4653: Incorrect CPE for React
**What happened**:
The project's SBOM was uploaded to DependencyTrack but was not matched with the related CVE. We double checked the CPE, and it was different from the CPE provided by NVD.

The generated CPE by Syft:
`cpe:2.3:a:react:react:18.3.1:*:*:*:*:*:*:*`

Provided CPE by NVD:
`cpe:2.3:a:facebook:react:18.3.1:*:*:*:*:*:*:*`

Correct CPEs matching with NVD format.

Scan a project's package-lock.yaml containing React version 18.3.1 with Syft version 1.42.1.

SAY-5:
Fixes #4653.

## Problem

NVD records React and `react-dom` under the `facebook` vendor, so the CPE syft emits from an npm `package-lock.json` today:

```
cpe:2.3:a:react:react:18.3.1:*:*:*:*:*:*:*
```

fails to match any NVD record. Users uploading the resulting SBOM to Dependency Track (or scanning with grype) miss every React CVE, because NVD expects:

```
cpe:2.3:a:facebook:react:18.3.1:*:*:*:*:*:*:*
```

## Fix

`syft/pkg/cataloger/internal/cpegenerate/candidate_by_package_type.go` already has an "aliases" table for exactly this kind of mismatch (see the existing entries for `next` / `vercel`, `hapi`, `handlebars.js`, `mustache`). Added two more:

```go
{
    pkg.NpmPkg,
    candidateKey{PkgName: "react"},
    candidateAddition{AdditionalVendors: []string{"facebook"}},
},
{
    pkg.NpmPkg,
    candidateKey{PkgName: "react-dom"},
    candidateAddition{AdditionalVendors: []string{"facebook"}},
},
```

`react-dom` shares the same CVE stream so it gets the same alias.

## Test

Added `Test_npmCandidateAdditions_react` in `candidate_by_package_type_test.go`, which runs the real `findAdditionalVendors` lookup against `defaultCandidateAdditions` for both package names and asserts `facebook` is in the returned slice. Means the mapping cannot silently regress if someone later edits the table.

## Back-compat

Existing consumers that relied on matching `cpe:2.3:a:react:react:*` directly (there should not be any, since it does not match NVD) still see that CPE candidate emitted, because we only _add_ to the candidate-vendor set. This is purely additive.

Signed off per DCO.

### DIFF
--- a/syft/pkg/cataloger/internal/cpegenerate/candidate_by_package_type.go
+++ b/syft/pkg/cataloger/internal/cpegenerate/candidate_by_package_type.go
@@ -221,6 +221,23 @@ var defaultCandidateAdditions = buildCandidateLookup(
 			candidateKey{PkgName: "mustache"},
 			candidateAddition{AdditionalProducts: []string{"mustache.js"}},
 		},
+		{
+			// NVD records react under the facebook vendor, e.g.
+			// cpe:2.3:a:facebook:react:*, but the npm package.json
+			// names neither the vendor nor the author. Without
+			// this hint syft emits cpe:2.3:a:react:react:* and
+			// grype/DependencyTrack miss every React CVE. See #4653.
+			pkg.NpmPkg,
+			candidateKey{PkgName: "react"},
+			candidateAddition{AdditionalVendors: []string{"facebook"}},
+		},
+		{
+			// Same story for react-dom, which NVD also records under
+			// facebook (and ties to react CVEs).
+			pkg.NpmPkg,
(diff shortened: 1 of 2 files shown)


SAY-5:
Closing as duplicate of #4682 (missed that existing PR). Apologies for the noise.

PR #4786, opened 2026-04-17
javascript: skip package.json under composer-managed vendor trees

SAY-5:
Fixes downstream anchore/grype#3279.

## Problem

When syft scans an image that was built with `composer install`, every PHP package under `vendor/<vendor>/<pkg>/` that happens to ship a `package.json` (to describe a bundled JS asset or simply because upstream included one) is reported as an independent npm package. Grype then matches it against the npm advisory stream and surfaces npm-only CVEs like `GHSA-hj79-42mx-m4gr` ("Malware in @umpirsky/country-list") against the PHP package:

```
@umpirsky/country-list  1.0.0  npm  GHSA-hj79-42mx-m4gr  Critical
```

The actual component is the composer package `umpirsky/country-list`, not the npm namespace. The npm CVE has no bearing on it.

## Fix

Skip those `package.json` files at the cataloger level. The signal is the composer install layout itself:

```
<project>/vendor/<vendor>/<pkg>/package.json
          ^
          composer.json sits at <project>
```

In `parse_package_json.go` the new `isComposerVendoredPackageJSON` helper checks:

1. the file is `package.json`
2. the path is under a `.../vendor/<vendor>/<pkg>/` two-level layout
3. a `composer.json` exists at the project root (the `..` above `vendor/`)

When all three hold we return `(nil, nil, nil)` and the npm cataloger never sees the file. Anything else (top-level `package.json`, `package.json` under `node_modules`, projects that don't carry `composer.json`) is unaffected.

Signed off per DCO.

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_package_json.go
+++ b/syft/pkg/cataloger/javascript/parse_package_json.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"path"
 	"regexp"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


syft/pkg/cataloger/javascript/parse_package_json.go
```diff
@@ -245,3 +255,37 @@ func (p people) String() string {
...
+	}
+	cleaned := path.Clean(p)
+	if path.Base(cleaned) != "package.json" {
+		return false
+	}
+	pkgDir := path.Dir(cleaned)               // .../vendor/<vendor>/<pkg>
+	vendorNameDir := path.Dir(pkgDir)         // .../vendor/<vendor>
+	vendorDir := path.Dir(vendorNameDir)      // .../vendor
+	projectRoot := path.Dir(vendorDir)        // ...
+	if path.Base(vendorDir) != "vendor" || projectRoot == "" {
```
yariksheptykin, line 282: For completeness sake: composer allows configuring the actual name of the "vendor" directory. See https://getcomposer.org/doc/06-config.md#vendor-dir (via ENV: https://getcomposer.org/doc/03-cli.md#composer-vendor-dir)

This is not a wide-spread practice in the php community, but there could be some folks actually using this option. Instead of hard-coding "vendor" as a constant, you might make this somehow configurable.

Same goes for "composer.json", See: https://getcomposer.org/doc/03-cli.md#composer

PR #4794, opened 2026-04-19
feat: exclude indirect deps

ZZaRRyaDD:
## Summary

Added exclusion of indirect dependencies when building an SBOM file. Currently, this flag is supported for Golang projects.

Added a new bool flag, exclude-indirect. Default value is false.

## Type of change

- New feature

## Testing

Tested with own project

### DIFF
--- a/cmd/syft/internal/options/catalog.go
+++ b/cmd/syft/internal/options/catalog.go
@@ -191,7 +191,8 @@ func (cfg Catalog) ToPackagesConfig() pkgcataloging.Config {
 					WithFromBuildSettings(cfg.Golang.MainModuleVersion.FromBuildSettings).
 					WithFromLDFlags(cfg.Golang.MainModuleVersion.FromLDFlags),
 			).
-			WithUsePackagesLib(*multiLevelOption(true, enrichmentEnabled(cfg.Enrich, task.Go, task.Golang), cfg.Golang.UsePackagesLib)),
+			WithUsePackagesLib(*multiLevelOption(true, enrichmentEnabled(cfg.Enrich, task.Go, task.Golang), cfg.Golang.UsePackagesLib)).
+			WithExcludeIndirect(cfg.Golang.ExcludeIndirect),
 		JavaScript: javascript.DefaultCatalogerConfig().
 			WithIncludeDevDependencies(*multiLevelOption(false, cfg.JavaScript.IncludeDevDependencies)).
 			WithSearchRemoteLicenses(*multiLevelOption(false, enrichmentEnabled(cfg.Enrich, task.JavaScript, task.Node, task.NPM), cfg.JavaScript.SearchRemoteLicenses)).
@@ -249,6 +250,8 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
 
 	flags.StringVarP(&cfg.Source.Supplier, "source-supplier", "",
 		"the organization that supplied the component, which often may be the manufacturer, distributor, or repackager")
+
+	flags.BoolVarP(&cfg.Golang.ExcludeIndirect, "exclude-indirect", "", "exclude indirect dependencies from sbom file (only golang support now)")
 }
 
 func (cfg *Catalog) DescribeFields(descriptions fangs.FieldDescriptionSet) {
--- a/syft/pkg/cataloger/golang/parse_go_mod.go
+++ b/syft/pkg/cataloger/golang/parse_go_mod.go
@@ -27,12 +27,14 @@ import (
 
 type goModCataloger struct {
 	usePackagesLib  bool
+	excludeIndirect bool
 	licenseResolver goLicenseResolver
 }
 
 func newGoModCataloger(opts CatalogerConfig) *goModCataloger {
 	return &goModCataloger{
 		usePackagesLib:  opts.UsePackagesLib,
+		excludeIndirect: opts.ExcludeIndirect,
 		licenseResolver: newGoLicenseResolver(modFileCatalogerName, opts),
 	}
 }
@@ -255,6 +257,10 @@ func (c *goModCataloger) catalogModules(
 	moduleToPackage := make(map[string]artifact.Identifiable)
 
 	for _, m := range modules {
+		if c.excludeIndirect && m.Indirect {
+			continue
+		}
+
 		if isRelativeImportOrMain(m.Path) {
 			// relativeImport modules are already accounted for by their full module paths at other portions of syft's cataloging
 			// example: something like ../../ found as a module for go.mod b, which is sub to go.mod a is accounted for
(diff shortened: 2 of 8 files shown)


cmd/syft/internal/options/catalog.go
```diff
@@ -42,6 +42,7 @@ type Catalog struct {
 	Relationships relationshipsConfig `yaml:"relationships" json:"relationships" mapstructure:"relationships"`
 	Compliance    complianceConfig    `yaml:"compliance" json:"compliance" mapstructure:"compliance"`
 	Enrich        []string            `yaml:"enrich" json:"enrich" mapstructure:"enrich"`
+	ExcludeIndirect bool                `yaml:"exclude-indirect" json:"exclude-indirect"  mapstructure:"exclude-indirect"`
```
kzantow: I think for now, we should move this to a [golang-specific option](https://github.com/anchore/syft/blob/main/cmd/syft/internal/options/golang.go#L19). At some point this will probably be expanded to an option across catalogers but it's unclear exactly what the option will be (should it be about "dependencies=direct", "transitive", "--depth=1", etc..?). I don't think "exclude indirect" would be the final terminology for this.

cmd/syft/internal/options/catalog.go
```diff
@@ -249,6 +251,8 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
 
 	flags.StringVarP(&cfg.Source.Supplier, "source-supplier", "",
 		"the organization that supplied the component, which often may be the manufacturer, distributor, or repackager")
+
+	flags.BoolVarP(&cfg.ExcludeIndirect, "exclude-indirect", "", "exclude indirect dependencies from sbom file (only golang support now)")
```
kzantow: Similarly, we should not add a top-level flag for this until such time as this is a shared global option.

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -255,6 +257,10 @@ func (c *goModCataloger) catalogModules(
 	moduleToPackage := make(map[string]artifact.Identifiable)
 
 	for _, m := range modules {
+		if c.excludeIndirect && m.Indirect {
```
kzantow, line 260: In order to get this merged, we should add at least one test exercising these options.

kzantow (changes requested):

cmd/syft/internal/options/catalog.go
ZZaRRyaDD: I added this parameter because I didn't see anything else for SBOM build files that didn't contain indirect dependencies. What is the "dependencies" parameter?

syft/pkg/cataloger/golang/parse_go_mod.go
ZZaRRyaDD: done

cmd/syft/internal/options/catalog.go
ZZaRRyaDD: done

ZZaRRyaDD:
@kzantow check pull request, pls)

cmd/syft/internal/options/catalog.go
```diff
@@ -249,6 +250,8 @@ func (cfg *Catalog) AddFlags(flags clio.FlagSet) {
 
 	flags.StringVarP(&cfg.Source.Supplier, "source-supplier", "",
 		"the organization that supplied the component, which often may be the manufacturer, distributor, or repackager")
+
+	flags.BoolVarP(&cfg.Golang.ExcludeIndirect, "exclude-indirect", "", "exclude indirect dependencies from sbom file (only golang support now)")
```
kzantow, line 254: There should not be a CLI flag added for this option at this time. We should have more than one cataloger opted in to this feature to understand better how a CLI configuration option should work.

syft/pkg/cataloger/golang/parse_go_mod.go
```diff
@@ -367,6 +377,10 @@ func (c *goModCataloger) createGoModPackages(ctx context.Context, resolver file.
 // applyReplaceDirectives processes replace directives from go.mod
 func (c *goModCataloger) applyReplaceDirectives(ctx context.Context, resolver file.Resolver, modFile *modfile.File, goModPackages map[string]pkg.Package, reader file.LocationReadCloser, digests map[string]string) {
 	for _, m := range modFile.Replace {
+		if _, isContains := goModPackages[m.Old.Path]; c.excludeIndirect && !isContains {
```
kzantow, line 380: does this change have anything to do with `excludeIndirect`? shouldn't we always exclude replace directives that don't match any go mod packages

kzantow (commented):
I'm having a hard time understanding the use case for this change. Is the idea to identify which packages you can upgrade that would have the most impact on vulnerabilities? Are there other use cases you have for this?

ZZaRRyaDD:
@kzantow This PR is needed to exclude indirect dependencies from the SBOM file, because When a CVE is detected on indirect dependencies, triage of the direct dependency occurs. Authors of direct dependencies rarely update indirect dependencies, and so as not to depend on them, such a change was proposed.

I saw that catalogers has a direct/indirect parameter, but when changing it through the yaml file, I did not see any change in SBOM file

kzantow:
There were two inline comments which have not been addressed in this PR.

As for the use case, it seems like my assumption was right; but I must caution this solution. What you are proposing here is something that would probably be better suited for Grype to filter or prioritize vulnerabilities. Assuming the vulnerable code paths are being executed, _these vulnerabilities still exist in your software_. By omitting the indirect packages, you are losing visibility into potentially critical issues that could leave you vulnerable. A better solution may be for Syft to track whether the vulnerability is a direct or indirect dependency and Grype could then have an option to only show a subset or maybe better: prioritize the direct dependency updates with fixes and/or annotate the vulnerability coming from a direct or indirect dependency.

I think _tracking direct vs indirect_ seems like a great addition to Syft, since this would just be surfacing the raw information we found, though at this point I would assume it could only be included in [the relationship metadata](https://github.com/anchore/syft/blob/main/syft/artifact/relationship.go#L41) and I'm unclear how non-Syft SBOMs should convey this (surely CycloneDX is unable to given the relationship limitations; maybe SPDX could include this as part of [the comment](https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/#112-relationship-comment-field)).

For Go projects, it _is possible_ to upgrade indirect dependencies using the Go mod replace and of course _newer versions of the direct dependencies may have upgraded to non-vulnerable versions_, so even if you are seeing vulnerabilities in indirect dependences there is some amount of action you are able to take to remediate these today, suppressing them by hiding the packages doesn't seem like the right solution when I think you're asking for a way to help focus on the easiest actions to take.

I would love for Grype to some day be able to give some indication of _direct dependency upgrades_ that would solve _indirect dependency vulnerabilities_. The problem is that this requires speculation across the dependency graph of things that aren't in the SBOM (newer versions of the dependencies; and beyond that how your specific project will end up when all dependencies are considered; and beyond that a reachability analysis with a newer version), which is probably not something we will be able to do effectively anytime soon.

In lieu of the ideal solution, what would you think about annotating or filtering Grype results based on this indirect vs direct relationship information?

PR #4801, opened 2026-04-20
go.mod: don't enforce latest go version

thaJeztah:
Use a separate file for the go-version used in CI so that the go.mod can adhere to Go's MVS (Minimum Version Selection) to specify the minimum required version to use the module. This prevents importers of the module from being forced to update to a newer version of Go. For library modules, Go generally recommends supporting the currently supported versions of Go (latest and latest-1).

This patch:

- downgrades the go directive in go.mod to go1.25.8
- updates github actions workflows to use a `.go-version` file.
- sets GOTOOLCHAIN=local to prevent implicit go upgrades

## Type of change

## Checklist


go.mod
```diff
@@ -1,6 +1,11 @@
 module github.com/anchore/syft
 
-go 1.26.2
+// The go directive declares the minimum Go version required for this module.
+//
+// DO NOT change this version unless support for older Go versions is dropped
+// or the module requires newer Go features. To update the version used for CI
+// and releases, update the ".go-version" file at the root of this repository.
+go 1.25.8
```
thaJeztah, line 8: Looks like the .8 comes from https://github.com/hashicorp/go-getter/pull/604 - I'll have a look at doing a similar change there as well.

wagoodman:
@thaJeztah Thanks for highlighting this before the next release, we did not intend to modify the go minimum version (since we're a lib and don't want to artificially force others to require a go version bump).

We're not yet going to adopt the `.go-version` file and instead have an additional way we're going to handle this. I opened/merged https://github.com/anchore/syft/pull/4805 that does just the reverting of the go minimum version, so I'm going to close this PR since I think it addresses what you need.

Appreciate you reaching out about this 🙌 !

thaJeztah:
Thanks! Yes, that works; the alternative would be to set `gotoolchain` in `go.mod`, which should not impact users of the module, but will restrict what version of Go can be used to build, which is also a bit of a pain 😞

PR #4802, opened 2026-04-20
Materialize FileInfo from DirEntry in WalkDiskDir for Go 1.26

### LINKED ISSUE #4796: fs.DirEntry does not implement fs.FileInfo error in `go-1.26`
**What happened**:

```bash
17:42:10  ../../go/pkg/mod/github.com/anchore/syft@v1.42.4/internal/file/squashfs.go:29:30: cannot use info (variable of interface type fs.DirEntry) as os.FileInfo value in argument to walkDiskDir: fs.DirEntry does not implement fs.FileInfo (missing method ModTime)
17:42:10  ../../go/pkg/mod/github.com/anchore/syft@v1.42.4/internal/file/squashfs.go:78:38: cannot use d1 (variable of interface type fs.DirEntry) as os.FileInfo value in argument to walkDiskDir: fs.DirEntry does not implement fs.FileInfo (missing method ModTime)
```

TBD

The following worked in `go-1.25.x?` and started to give the above error in `go-1.26.2`

`return ac, filepath.Walk(gitRoot, func(path string, info os.FileInfo, err error) error {`

I see that your main branch have bumped to `1.26.2`, so this might be fixed in the next release.

**Environment**:
- Output of `syft version`: `v1.42.4`

joaquinhuigomez:
Go 1.26 removed `ModTime` from `fs.DirEntry`, so `fs.DirEntry` no longer satisfies `os.FileInfo`. `WalkDiskDir` passed `DirEntry` values from `fsys.ReadDir` directly into `walkDiskDir(… d os.FileInfo, …)`, which now fails to build against the go1.26 toolchain.

Convert each entry via a small helper that prefers a direct `FileInfo` type assertion and falls back to `DirEntry.Info()` otherwise. Behavior is unchanged for callers that already return `FileInfo` values.

Fixes #4796

### DIFF
--- a/internal/file/squashfs.go
+++ b/internal/file/squashfs.go
@@ -24,8 +24,15 @@ func WalkDiskDir(fsys filesystem.FileSystem, root string, fn WalkDiskDirFunc) er
 		return nil
 	}
 
-	for _, info := range infos {
-		p := filepath.Join(root, info.Name())
+	for _, entry := range infos {
+		// fs.DirEntry no longer satisfies os.FileInfo in Go 1.26+
+		// (ModTime moved off of DirEntry), so materialize the FileInfo
+		// via Info() before passing it down.
+		info, err := dirEntryInfo(entry)
+		if err != nil {
+			return err
+		}
+		p := filepath.Join(root, entry.Name())
 		err = walkDiskDir(fsys, p, info, fn)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


sastorsl:
@joaquinhuigomez I see that the DCO test fails.
Perhaps you can fix your commit with `git commit --amend --no-edit --signoff` and force push the updated commit?

PR #4807, opened 2026-04-22
internal/file: propagate non-EOF errors out of safeCopy

### LINKED ISSUE #4806: Syft safeCopy silently swallows archive decompression errors
This was initially reported to us by @bugbunny-research as a security report

We decided this did not cross a trust boundary per the text added to
https://github.com/anchore/syft/blob/main/SECURITY.md#trust-boundary

This is still a bug, here are the important details

## Summary
safeCopy in internal/file/copy.go does not propagate non-EOF errors returned by io.Copy. Because the function only returns an error for the decompression-bomb case (numBytes >= perFileReadLimit) or the dead io.EOF branch (which io.Copy never surfaces as an error), any corruption or decompression failure encountered while reading a ZIP/TAR entry is silently treated as success. Callers store the resulting empty or partial content, downstream parsers silently skip it, and Syft exits 0 with a subtly wrong SBOM — package names and versions drawn from corrupted pom.properties or manifest entries are replaced by less-specific fallback data (filename), causing SBOM-based CVE scanners to miss known vulnerabilities in crafted artifacts.

## Description

The vulnerable safeCopy function
```
// internal/file/copy.go (lines 13–19)
func safeCopy(writer io.Writer, reader io.Reader) error {
    numBytes, err := io.Copy(writer, io.LimitReader(reader, perFileReadLimit))
    if numBytes >= perFileReadLimit || errors.Is(err, io.EOF) {
        return fmt.Errorf("zip read limit hit (potential decompression bomb attack)")
    }
    return nil   // ← swallows any non-nil, non-EOF error from io.Copy
}
```
Two compounding issues:

1) Non-EOF errors are silently dropped. If io.Copy encounters a deflate/zlib decompression error while reading a corrupted archive entry, it returns (n, err) where err is the decompression error (e.g., flate: corrupt input before offset X). Because n < perFileReadLimit and errors.Is(err, io.EOF) is false, safeCopy falls through to return nil.

1) The io.EOF branch is dead code. The Go standard library's io.Copy consumes io.EOF internally and returns (n, nil) on a clean end-of-stream. errors.Is(err, io.EOF) inside safeCopy can never be true in normal operation.

## How the error propagates (or fails to) through callers
All three archive helpers that feed the Java cataloger call safeCopy and wrap its return value:

```
// internal/file/zip_file_traversal.go:141 (ContentsFromZip visitor)
if err := safeCopy(&buffer, zippedFile); err != nil {
    return fmt.Errorf("unable to copy source=%q for zip=%q: %w", ...)
}
results[file.NameInArchive] = buffer.String()  // stores "" on silent failure
```

```
// internal/file/zip_file_traversal.go:104 (ExtractFromZipToUniqueTempFile visitor)
if err := safeCopy(tempFile, zippedFile); err != nil {
    return fmt.Errorf(...)
}
results[file.NameInArchive] = Opener{path: tempFile.Name()}  // stores empty file on silent failure
```

```
// internal/file/tar_file_traversal.go:73 (ExtractGlobsFromTarToUniqueTempFile visitor)
if err := safeCopy(tempFile, packedFile); err != nil {
    return fmt.Errorf(...)
}
results[file.Name()] = Opener{path: tempFile.Name()}  // same
```

Because safeCopy returns nil, the error never propagates out of the visitor, never propagates out of ContentsFromZip, and the callers in the Java archive parser receive (map_with_empty_content, nil).

## Inconsistency: error handling differs between call sites

The bug causes inconsistent behaviour across the two primary call paths in the Java parser:

- discoverMainPackageFromPomInfo (line 390): errors from pomPropertiesByParentPath are discarded with _. Even a fixed safeCopy would not help here — this call site ignores errors unconditionally.
- discoverPkgsFromAllMavenFiles (line 522–525): errors are propagated and returned. A fixed safeCopy would surface the error here, causing the entire archive scan to fail with an error rather than silently misidentifying the package.

Both call sites should be corrected together.

SAY-5:
## Summary

Fixes #4806.

`safeCopy`'s old `numBytes >= perFileReadLimit || errors.Is(err, io.EOF)` check collapsed two independent conditions:

- The `io.EOF` branch is dead code — `io.Copy` consumes EOF internally and returns `(n, nil)` on a clean end-of-stream.
- The real bug is that any other non-EOF error from `io.Copy` — `flate: corrupt input before offset X` on a mangled zip entry, a truncated tar stream, a network-backed reader erroring mid-read — was dropped because the function fell through to `return nil`.

The Java cataloger's `ContentsFromZip` / `ContentsFromTar` visitors then stored the partial/empty buffer as a "successful" extract, downstream parsers silently skipped it, and package names + versions drawn from `pom.properties` / `META-INF/MANIFEST.MF` were replaced by less-specific filename fallbacks. SBOM-based CVE scanners missed known vulnerabilities on crafted artifacts.

## Fix

Split the two concerns:

- Size-hit → `zip read limit hit (potential decompression bomb attack)` (unchanged).
- Any non-nil non-EOF error from `io.Copy` → `failed to read archive entry: <wrapped error>`.
- Clean copies still return `nil`.

## Tests

New `internal/file/copy_test.go::TestSafeCopy` covers:
- clean copy → `nil` + expected contents
- `errReader` returning a sentinel error → `errors.Is(err, sentinel)` holds (fails on master, passes here)
- `io.EOF` is not mistaken for a bomb

`go test ./internal/file/...` green; `go vet` clean; `go build` clean.

### DIFF
--- a/internal/file/copy.go
+++ b/internal/file/copy.go
@@ -12,8 +12,20 @@ const perFileReadLimit = 2 * GB
 // protect against decompression bomb attacks.
 func safeCopy(writer io.Writer, reader io.Reader) error {
 	numBytes, err := io.Copy(writer, io.LimitReader(reader, perFileReadLimit))
-	if numBytes >= perFileReadLimit || errors.Is(err, io.EOF) {
+	if numBytes >= perFileReadLimit {
 		return fmt.Errorf("zip read limit hit (potential decompression bomb attack)")
+	}
+	// Propagate decompression / read errors up to the caller. io.Copy
+	// on the happy path returns (n, nil); the only way err is non-nil
+	// here is that the underlying reader surfaced a real failure
+	// ("flate: corrupt input before offset X" on a mangled ZIP entry,
+	// a network-backed reader erroring mid-stream, etc.). The previous
+	// implementation dropped that error and the caller stored a
+	// partial / empty buffer as a "successful" extract, which silently
+	// downgraded Java cataloger output and caused SBOM scanners to
+	// miss known CVEs (#4806).
+	if err != nil && !errors.Is(err, io.EOF) {
+		return fmt.Errorf("failed to read archive entry: %w", err)
 	}
 	return nil
 }
--- a/internal/file/copy_test.go
+++ b/internal/file/copy_test.go
@@ -1,0 +1,64 @@
+package file
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"strings"
+	"testing"
+)
+
+// errReader returns a deterministic non-EOF error after emitting some
+// bytes, mimicking what compress/flate does when it hits a corrupt
+// stream mid-entry.
+type errReader struct {
+	data []byte
+	err  error
+	off  int
+}
+
+func (r *errReader) Read(p []byte) (int, error) {
+	if r.off >= len(r.data) {
+		return 0, r.err
+	}
+	n := copy(p, r.data[r.off:])
+	r.off += n
+	return n, nil
+}
+
+func TestSafeCopy(t *testing.T) {
+	t.Run("clean copy returns nil", func(t *testing.T) {
+		var buf bytes.Buffer
+		if err := safeCopy(&buf, strings.NewReader("hello")); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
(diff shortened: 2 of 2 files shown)


kzantow (approved):
LGTM 👍

PR #4810, opened 2026-04-22
fork github.com/facebookincubator/nvdtools/wfn as internal package

thaJeztah:
### fork github.com/facebookincubator/nvdtools/wfn as internal package

The github.com/facebookincubator/nvdtools repository has been archived,
and comes with various (indirect) dependencies. Fork the wfn package,
which is the only package used in this repository.

This forks the package at the last release (v0.1.5);

    # install filter-repo (https://github.com/newren/git-filter-repo/blob/main/INSTALL.md)
    brew install git-filter-repo

    cd ~/projects

    # create a temporary clone of docker
    git clone https://github.com/facebookincubator/nvdtools.git nvdtools_temp
    cd nvdtools_temp
    git checkout v0.1.5
    git checkout -b filtered-wfn

    # remove all code, except for wfn, and rename to syft/internal/wfn
    git filter-repo  --path wfn --path-rename wfn:syft/internal/wfn --force

    # go to the target github.com/moby/sys repository
    cd ~/projects/syft

    # create a branch to work with
    git checkout -b integrate_nvdtools_wfn

    # add the temporary repository as an upstream and make sure it's up-to-date
    git remote add nvdtools ~/projects/nvdtools_temp
    git fetch nvdtools

    # merge the upstream code
    git merge --allow-unrelated-histories --signoff -S nvdtools/filtered-wfn

### remove uses of github.com/facebookincubator/nvdtools/wfn

switch to the internal fork of this package.

## Type of change

## Checklist


thaJeztah:
Ah, yes, the sign-offs would be missing because they're imported commits. I know for other projects we usually decided to just allow those (to preserve the original), but open to input on that part (I tried to preserve history, which may be relevant, but we could squash and link to the original source instead).

syft/cpe/cpe.go
```diff
@@ -5,7 +5,7 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/facebookincubator/nvdtools/wfn"
+	"github.com/anchore/syft/syft/internal/wfn"
```
kzantow, line 8: I'm not necessarily opposed to this change, especially given the upstream repo being archived. But just wanted to point out that this will have some potentially breaking changes for consumers.

thaJeztah: 😬 Do you have some specific scenario in mind?

I recall I had a quick glance and it looked like the types are only  used internally, then wrapped / converted to types defined in this module, but  it's definitely possible I overlooked.

kzantow: It would mostly be with the unlikely scenario that someone has an integration with another module using facebook's wfn. I don't think there's a large issue but this is an unfortunate side effect of using 3rd party libraries in our main data model.

thaJeztah: AFAICS, the imported types are not part of the public API, so there's no observable change for any consumer; https://pkg.go.dev/github.com/anchore/syft@v1.43.0/syft/cpe

If consumers use the facebook dependency for other reasons, they still can but as those types are not part of the public API, it should have no impact.

Using external dependencies (especially when commonly used) is a good thing, just making sure the types are not part of the exported interface, unless that's needed for interoperability.

kzantow: Yes, scratch what I said. 🤦

thaJeztah:
OK; moved this one out of draft, but could use input; the DCO check is failing, which is expected for the imported commits, but there's multiple ways to "solve" it;

- Easiest would be to acknowledge that the commits were obtained from a valid source, and manually mark the DCO check to "pass".
- I can squash all commits; single commit with my `Signed-off-by`; that would lose visibility on history of the code, lose the ability to "git blame" individual lines to get context around changes made to the code. Admitted; most commits looks to be fairly mechanical / trivial, but it's still purging history / provenance.
- I can add a `Signed-off-by` to each commit, but that'd be somewhat mutating the commit; still correct as I sign off that I obtained the source-code legally;

```
(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or
```

PR #4815, opened 2026-04-22
fix: only parse debian/copyright as machine-readable if Format: field present

lunchfang:
## Summary

This PR fixes the parsing of Debian copyright files to only treat files as machine-readable if they properly declare themselves as such.

## Changes

1. **Added  function** - Detects if a copyright file declares itself as machine-readable format by checking for the  field with a valid URL.

2. **Added ** - Regex pattern  to match the required Format field per the [Debian copyright format specification](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/).

3. **Updated  logic** - Now only parses copyright files using the machine-readable parser if the file contains the  field. Non-machine-readable files will be processed by the license classifier instead.

4. **Added test fixtures**:
   -  - Example machine-readable format file
   -  - Example non-machine-readable format file

5. **Added test case** -  to verify the detection logic

## Motivation

Per the [Debian copyright format 1.0 specification](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/), machine-readable format requires a mandatory  field containing a URL link to the spec. Many Debian packages contain poorly parsed copyright information that doesn't follow this format.

The current implementation tries to parse all copyright files as machine-readable, even those without the  field, leading to incorrect license parsing.

## Test Results

- Machine-readable files (with  field): ✅ Correctly identified
- Non-machine-readable files: ✅ Correctly routed to license classifier

---
*This PR is part of the syft bounty program (issue #4708)*

### DIFF
--- a/syft/pkg/cataloger/debian/package.go
+++ b/syft/pkg/cataloger/debian/package.go
@@ -1,6 +1,7 @@
 package debian
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


spiffcs:
Thanks for the PR @lunchfang.

This was independently fixed in #4754 which closed #4708. Apologies for the logjam here since we've had a lot of volume of PR attacking similar issues.

Closing this as superseded.

PR #4823, opened 2026-04-23
chore: tidy go.mod

thaJeztah:
Some dependencies ended up in the wrong group, and groups were split.

## Type of change

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/Masterminds/sprig/v3 v3.3.0
 	github.com/OneOfOne/xxhash v1.2.8
 	github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d
+	github.com/acobaugh/osrelease v0.1.0
 	github.com/adrg/xdg v0.5.3
 	github.com/anchore/bubbly v0.2.0
 	github.com/anchore/clio v0.1.0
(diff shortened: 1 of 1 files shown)


kzantow:
Does this change have an impact for downstream consumers somehow? Or, maybe my real question: does the set of multiple direct and indirect groups affect consumers?

thaJeztah:
No, it does not; go groups the dependencies in two groups (direct and indirect) but more or less groups are allowed (it's really just formatting, no meaning otherwise).

Unfortunately go mod doesn't always do a good job at keeping them grouped correctly; so correcting that with this PR.

The `go.sum' didn't change, which gives an extra indicator that it didn't modify anything other than formatting.

kzantow:
> The `go.sum' didn't change, which gives an extra indicator that it didn't modify anything other than formatting.

Fair point, I was mostly wondering if there was some unexpected issue this caused that I had not heard of before. It would be nice if go formatted this file consistently with 2 groups by default. And 3 import groups, while they were at it 😁

kzantow (approved):

Yek666:
Thanks for this contribution!
I have a quick question regarding the implementation:
How does this change impact performance when scanning large images?
Have you considered edge cases like [X scenario]?
Overall the approach looks solid, just wanted to clarify those points.

PR #4828, opened 2026-04-25
golangci-lint: enable gci formatter

thaJeztah:
This allows linting the imports to be grouped correctly, and provides an auto-fix (`golangci-lint run --fix`).

## Type of change

## Checklist

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -70,7 +70,6 @@ linters:
 #    - godot
 #    - godox
 #    - goerr113
-#    - goimports   # we're using gosimports now instead to account for extra whitespaces (see https://github.com/golang/go/issues/20818)
 #    - golint      # deprecated
 #    - gomnd       # this is too aggressive
 #    - interfacer  # this is a good idea, but is no longer supported and is prone to false positives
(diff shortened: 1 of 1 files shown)


thaJeztah:
Relates to https://github.com/anchore/syft/pull/4823#issuecomment-4308585525

> It would be nice if go formatted this file consistently with 2 groups by default. And 3 import groups, while they were at it


cc @kzantow - saw your comment, and thought this could be a good alternative.

Taskfile.yaml
```diff
@@ -148,7 +148,6 @@ tasks:
     deps: [tools]
     cmds:
       - gofmt -w -s .
-      - "{{ .TOOL_DIR }}/gosimports -local github.com/anchore -w ."
```
thaJeztah: `lint-fix` below looks to be running `golangci-lint run --fix` so should do the formatting; it does run with `--tests=false` though, so perhaps that's something to evaluate.

FWIW; `gofmt` (and the, often better, `gofumpt`) also provide auto-fix in golangci-lint (https://golangci-lint.run/docs/formatters/configuration/#gofmt), so running `gofmt -w -s` may be redundant if `--fix` is used.

wagoodman (approved):
TIL! Thanks for the config update here

PR #4832, opened 2026-04-27
fix(java): skip META-INF/maven pom.xml in pom cataloger

toabctl:
## Description

The java-pom-cataloger globs `**/pom.xml` and parses every match as a project POM. When a JAR has been extracted into directory form before scanning (e.g. `unzip foo.jar -d /scan && syft /scan`, or a Dockerfile that unpacks a JAR at build time), the resolver indexes the `META-INF/maven/<groupId>/<artifactId>/pom.xml` that the [maven-archiver plugin](https://maven.apache.org/shared/maven-archiver/index.html) embeds. Those files are build metadata, not project source.

The embedded POMs are verbatim copies of the build pom, including `${...}` properties and parent-inherited dependency versions that aren't resolvable from the archive contents alone. Parsing them produces phantom dependency packages whose versions render as `UNKNOWN` (or empty / literal `${something}`) in the SBOM.

This change skips any pom.xml whose path contains a `META-INF/maven/` segment, anchored on a directory boundary so unrelated paths that happen to contain the substring aren't caught.

### Reproduction

Take any production-built fat JAR with a `<parent>` reference and properties defined in the parent. Apache Kafka 4.1.1 ships `commons-lang3-3.18.0.jar` as one of its libs, whose embedded pom.xml inherits from `org.apache.commons:commons-parent:85` and uses `${commons.text.version}` / `${jmh.version}` / etc.

```sh
mkdir /tmp/demo && cd /tmp/demo
unzip -q /path/to/commons-lang3-3.18.0.jar
syft dir:.
```

On `main`:

```
commons-lang3             3.18.0   java-archive
commons-text              1.13.1   java-archive
easymock                  5.6.0    java-archive
jmh-core                  1.37     java-archive
jmh-generator-annprocess  1.37     java-archive
jsr305                    3.0.2    java-archive
junit-jupiter             UNKNOWN  java-archive
junit-pioneer             UNKNOWN  java-archive
```

`junit-jupiter` and `junit-pioneer` are phantom: their versions are defined in commons-parent (not present in the archive), so the resolver bails and `UNKNOWN` ends up in the SBOM. The other six are dependency-side artifacts pulled from a build-time POM that the user never asked syft to look at; they propagate to vulnerability scans even when their versions happen to be literal.

With this PR:

```
No packages discovered
```

### Trade-off

maven-archiver writes `pom.properties` next to the pom.xml, with fully-resolved coordinates. The java-archive-cataloger reads it when processing a JAR file, but there's no equivalent path for directory scans today, so extracted-JAR scans no longer get a main package from the embedded metadata after this change. The trade is intentional: phantom deps with bogus versions actively pollute vulnerability scans, while a missing main package can be recovered later (e.g. from a future directory-aware pom.properties cataloger).

## Type of change

## Checklist

## Issue references

None.

### DIFF
--- a/syft/pkg/cataloger/java/parse_pom_xml.go
+++ b/syft/pkg/cataloger/java/parse_pom_xml.go
@@ -3,6 +3,7 @@ package java
 import (
 	"context"
 	"errors"
+	"path/filepath"
 	"strings"
 
 	"github.com/anchore/syft/internal"
@@ -40,6 +41,9 @@ func (p pomXMLCataloger) Catalog(ctx context.Context, fileResolver file.Resolver
 	var poms []*maven.Project
 	pomLocations := map[*maven.Project]file.Location{}
 	for _, pomLocation := range locations {
+		if isArchiveMetaPom(pomLocation) {
+			continue
+		}
 		pom, err := readPomFromLocation(fileResolver, pomLocation)
 		if err != nil || pom == nil {
 			log.WithFields("error", err, "pomLocation", pomLocation).Debug("error while reading pom")
@@ -297,6 +301,22 @@ func pomParent(ctx context.Context, r *maven.Resolver, pom *maven.Project) *pkg.
 		ArtifactID: artifactID,
 		Version:    version,
 	}
+}
+
+// isArchiveMetaPom returns true if the pom.xml location is inside a META-INF/maven directory,
+// indicating it is archive metadata embedded by the Maven build process rather than a project
+// pom.xml. These embedded pom.xml files are verbatim copies of the original build POM with
+// potentially unresolved ${property} references and dependency versions inherited from parent
+// POMs that are not available inside the archive. The java-archive-cataloger already uses
+// pom.properties (which always contains resolved values) to identify these archives, so the
+// pom cataloger should skip them to avoid creating phantom dependency packages.
+//
+// Match on a directory boundary ("/META-INF/maven/" or a "META-INF/maven/" prefix) to avoid
+// catching paths that happen to contain the substring elsewhere (e.g. a file literally named
+// "X-META-INF/maven/").
+func isArchiveMetaPom(location file.Location) bool {
+	p := filepath.ToSlash(location.Path())
+	return strings.Contains(p, "/META-INF/maven/") || strings.HasPrefix(p, "META-INF/maven/")
 }
 
 func cleanDescription(original string) (cleaned string) {
--- a/syft/pkg/cataloger/java/parse_pom_xml_test.go
+++ b/syft/pkg/cataloger/java/parse_pom_xml_test.go
@@ -690,6 +690,123 @@ func getCommonsTextExpectedPackages(resolved bool) expected {
 	return expected{pkgs, relationships}
 }
 
+func Test_isArchiveMetaPom(t *testing.T) {
+	tests := []struct {
+		name     string
+		path     string
+		expected bool
+	}{
+		{
+			name:     "META-INF maven pom",
+			path:     "META-INF/maven/com.example/my-lib/pom.xml",
+			expected: true,
+		},
+		{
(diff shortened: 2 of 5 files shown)


Eljees:
Independent confirmation of the problem, plus one measurement that I think matters for how it gets fixed.

**Scale in the wild.** Scanning an unpacked Java delivery (jars exploded next to the original files, which is how several pipelines feed Syft), `java-pom-cataloger` produced **627 of the 1542 components** from the embedded `META-INF/maven/*/*/pom.xml` files -- the artifacts themselves plus every dependency the poms *declared*, test and optional scopes included. 412 distinct `name@version` pairs among them are found nowhere in the tree as an actual archive (`assertj-core` at three versions, `byte-buddy-agent`, `awaitility`, `cglib-nodep`, `aws-java-sdk-test-utils`, `ant 1.10.11`, `bcprov-jdk18on 1.76`). `java-archive-cataloger` never does this for the same jar: it reads the embedded pom for identity only.

**Measurement.** I built three binaries -- `main` (392539d), this PR's head (5c97de3), and a variant that keeps the pom's own package but skips `collectDependencies` for it -- and ran them over two fixtures:

| fixture | `main` | this PR | keep artifact, skip declared deps |
|---|---|---|---|
| the real `xmlsec-2.3.4.jar` from Maven Central, unpacked | 20 packages (xmlsec + 19 phantoms, 15 of them `<scope>test</scope>`: bcprov-jdk18on 1.76, junit-jupiter, jetty ×3, xalan, xmlunit ×2, hamcrest, ...; 3 more are not declared in `<dependencies>` at all -- they come from the pom's `jdk19-plus` profile, whose activation the cataloger never evaluates) | **0** | xmlsec 2.3.4 |
| exploded WAR: `META-INF/maven/com.example/myapp/pom.xml` declaring log4j-core, jars in `WEB-INF/lib/` | myapp, log4j-core ×2, log4j-api | **log4j-core, log4j-api** -- the application itself is gone | myapp, log4j-core, log4j-api |
| a real 1.3 GB delivery: 558 jars, 202 embedded poms | 1542 packages (900 archive / 627 pom, of which **412** distinct `name@version` appear nowhere as an archive; 227 have version `UNKNOWN` from unresolved `${property}`) | 915 (all 202 artifacts dropped) | 1117 (pom 202, pom-only 0) |

*(Correction: the first version of this comment used a hand-written stand-in pom for the first row instead of the real xmlsec artifact. Replaced above with numbers from the jar as published on Maven Central; the conclusion is unchanged and the gap is larger than I first reported. Second correction, 11 Sep: this row first said 16 of the 19 were `<scope>test</scope>`. The pom declares 16 direct dependencies, 13 of them test; the other 3 of the 19 -- `jakarta.xml.bind-api`, `jaxb-runtime`, `saaj-impl` -- are not in `<dependencies>` at all but in the `jdk19-plus` profile, which the cataloger reads without evaluating activation, and 2 of those 3 are test scope. 15 of the 19 are test scope in total.)*

So skipping the file entirely removes the artifact exactly in the case where the pom cataloger is the only thing that sees it (no `.jar` file alongside). The embedded pom is a faithful record of *what this artifact is*; it is only its `<dependencies>` that describe what it was compiled against rather than what ships next to it.

The variant is #5258 (opened as a draft only because my account is at this repository's open-PR limit; the branch is `Eljees/syft@ee2f3571`, `fix/pom-cataloger-embedded-archive-pom`): same `isArchiveMetaPom` detector as here (credited), the pom stays registered with the resolver so it can still act as a parent/dependency source for a real project pom in the same tree, and your two fixtures are reused in the tests (exploded jar → exactly `my-lib 1.0.0`; project pom next to an embedded one → `my-app`, `guava`, `embedded-lib`, one relationship, none of `embedded-lib`'s declared deps). Happy to have it folded into this PR instead of a separate one -- whichever the maintainers prefer.

PR #4836, opened 2026-04-27
fix: resolve yarn lock aliases to source package

### LINKED ISSUE #4833: Yarn lockfile cataloguing does not handle aliases
**What happened**:

NPM and Yarn support a feature called "aliases" which allow multiple versions of the same package to be injected to the same library (or simple renames to be made). There's some documentation on what these look like [here](https://docs.npmjs.com/cli/v8/using-npm/package-spec#aliases).

This was previously fixed for NPM in #1314 but the Yarn cataloguer does not handle these correctly and uses the alias name as the package name.

Aliases to resolve to their source npm package versions rather than to use their alias name as the package name.

Reproducing is relatively easy:

```sh
yarn init
yarn add async@^3.2.3
yarn add old-async@npm:async@0.9.2
yarn install
syft .
```

What happens today:

```
NAME        VERSION          TYPE
async       3.2.6            npm
old-async   0.9.2            npm
```

Note `old-async` is the name of the alias, not the name of the package.

What should happen:

```
NAME        VERSION          TYPE
async       0.9.2            npm
async       3.2.6            npm
```

**Environment**:
- Output of `syft version`:

```
Application:   syft
Version:       1.43.0
BuildDate:     2026-04-22T15:09:10Z
GitCommit:     Homebrew
Platform:      darwin/arm64
GoVersion:     go1.26.2
Compiler:      gc
SchemaVersion: 16.1.3
```

```
> system_profiler SPSoftwareDataType | grep -v -e 'User Name' -e 'Computer Name'
Software:

    System Software Overview:

      System Version: macOS 26.4 (25E246)
      Kernel Version: Darwin 25.4.0
      Boot Volume: Macintosh HD
      Boot Mode: Normal
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
      Time since boot: 5 days, 16 hours, 5 minutes
```

cyphercodes:
## Description

Yarn lock entries that use npm aliases now resolve to the underlying npm package name instead of the local alias name. For example, `old-async@npm:async@0.9.2` is cataloged as `async@0.9.2`.

## Type of change

## Checklist

## Issue references

Fixes #4833

Verification:
- `GOTOOLCHAIN=local go test ./syft/pkg/cataloger/javascript -run TestParseYarnFindPackageNames`
- `GOTOOLCHAIN=local go test ./syft/pkg/cataloger/javascript`
- `git diff --check`

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_yarn_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_yarn_lock.go
@@ -25,6 +25,11 @@ import (
 )
 
 var (
+	// packageAliasExp matches aliased yarn dependencies and captures the
+	// underlying npm package name instead of the local alias.
+	// For example: "old-async@npm:async@0.9.2" returns "async".
+	packageAliasExp = regexp.MustCompile(`^"?(?:@\w[\w-_.]*\/)?\w[\w-_.]*@npm:((?:@\w[\w-_.]*\/)?\w[\w-_.]*)@`)
+
 	// packageNameExp matches the name of the dependency in yarn.lock
 	// including scope/namespace prefix if found.
 	// For example: "aws-sdk@2.706.0" returns "aws-sdk"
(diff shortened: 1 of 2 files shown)


calumleslie:
As the reporter, I can confirm this fixes the issue (edit: as far as I can tell from eyeballing it, at least). I had my own local fix but I like yours more!

kzantow (approved):
👍 thanks for the fix @cyphercodes and review @calumleslie

PR #4840, opened 2026-04-29
feat: add Julia cataloger

Octogonapus:
## Description

This PR adds a new cataloger for the [Julia](https://julialang.org/) language.
This enables Julia users to create SBOMs for their Julia applications using `Manifest.toml` and optionally `Project.toml` files.

The Julia Pkg.jl docs for reference if maintainers want to review the ecosystem behavior: https://pkgdocs.julialang.org/v1/toml-files/

I gave the Julia community 1 month of advance notice before opening this PR and no concerns were raised.
I'll circulate this PR around as well to solicit feedback or approval.

Some notes on the behavior for Julia users:

1. Uninstalled `[weakdeps]` are ignored by default; there is a config flag to include them as `optional` deps for users that always want to track these deps.
2. Test-only `[extras]` are ignored by default; there is a config flag to include them as `test` deps if they are also in `[targets] test = [...]` or `optional` deps otherwise. Extra deps used in tests are only resolved when tests are ran, so we don't know the package version nor its dependencies, but we can at least emit a dep with its name and UUID (version will be `UNKNOWN`).
3. `[sources]` are ignored. I think it could make sense to bring in this information in a later patch.
4. `[workspace]` is ignored. This operates at the meta-application level and I'm not sure it's a good fit for Syft.
5. The project license is ignored. License detection is a messy business and I don't want it to hold up this PR.
6. `Manifest.toml` is supported, along with an optional `Package.toml` for refinement, but a lone `Package.toml` is not supported (i.e. applications are supported, libraries are not). This is aligned with other SBOM tools like Trivy.
7. Syft generates CPEs that do not include the package UUID. Maybe that is part of a larger conversation though... would not want it to hold up this PR. The Julia security world is currently using pURL not CPE so I don't think this is a big issue.

There are two new configurations, as mentioned above:

1. `julia.include-extras` to include `[extras]` deps not in the `Manifest.toml`
2. `julia.include-weakdeps` to include `[weakdeps]` deps not in the `Manifest.toml`

Both are false by default; this is the expected behavior for the majority of users.
I included these flags for users that need to exhaustively track all dependencies, even optional ones.

In addition to the automated testing, I have been manually testing this on a few packages in the community and
manually reviewing the SBOM output.

There is one remaining problem with the automated testing that I am aware of. I added a Julia package to the
`image-pkg-coverage` sources, but it doesn't appear to be getting brought into the built image, so those integration
tests fail. I'm not sure how to fix this; any advice would be appreciated.

## Type of change

## Checklist


syft/pkg/cataloger/julia/capabilities.yaml
```diff
@@ -0,0 +1,82 @@
+# Cataloger capabilities. See ../README.md for documentation.
+
+configs: # AUTO-GENERATED - config structs and their fields
+  julia.CatalogerConfig:
+    fields:
+      - key: IncludeExtras
+        description: IncludeExtras enables including extra dependencies (such as test dependencies) in the catalog results.
```
Octogonapus: Sorry it looks like this is out of date. Trying to regenerate it

Octogonapus: #b75c304c

syft/pkg/cataloger/julia/package.go
```diff
@@ -0,0 +1,41 @@
...
+
+	p.SetID()
+
+	return p
+}
+
+func juliaPackageURL(name, version, uuid string) string {
+	var qualifiers packageurl.Qualifiers
+	if uuid != "" {
+		qualifiers = packageurl.QualifiersFromMap(map[string]string{
```
kzantow: nit: should we just construct a `Qualifier` directly to avoid creating intermediary maps all the time? e.g.:
```
qualifiers = []packageurl.Qualifiers{{ Key: "uuid", Value: uuid }}
```
type of thing?

kzantow (commented):
I haven't combed through every detail just yet, but overall this looks really good! You could try `make lint-fix` to solve some of the static analysis issues, or at least identify them locally

syft/pkg/cataloger/julia/package.go
Octogonapus: Addressed. Sorry, I don't write much go.

Octogonapus:
Currently working on the CLI test failures. Not sure how to fix the integ test failure.
edit: actually it looks like all the failures trace back to `image-pkg-coverage` not containing my Julia example package, so not sure how to fix them

kzantow:
@Octogonapus I'm guessing since you correctly labeled the Julia manifest cataloger to be included in directory/source scans but not image scans, it is not enabled for the integration image scan.

Octogonapus:
Yes that was it, thanks. Not sure if I should also use `PackageTag` or if that's redundant because I already have `LanguageTag`. `InstalledTag` is definitely NOT relevant because this does not scan the local depot.

internal/task/package_tasks.go
```diff
@@ -152,6 +153,12 @@ func DefaultPackageTaskFactories() Factories {
 			pkgcataloging.DirectoryTag, pkgcataloging.InstalledTag, pkgcataloging.ImageTag, pkgcataloging.LanguageTag, "nix",
 		),
 		newSimplePackageTaskFactory(lua.NewPackageCataloger, pkgcataloging.DirectoryTag, pkgcataloging.InstalledTag, pkgcataloging.ImageTag, pkgcataloging.LanguageTag, "lua"),
+		newPackageTaskFactory(
+			func(cfg CatalogingFactoryConfig) pkg.Cataloger {
+				return julia.NewPackageCataloger(cfg.PackagesConfig.Julia)
+			},
+			pkgcataloging.DeclaredTag, pkgcataloging.ImageTag, pkgcataloging.DirectoryTag, pkgcataloging.LanguageTag, "julia",
```
kzantow, line 160: Are Julia programs are usually _compiled to a binary_ and run? Would you expect to find one of the manifest files for the binary present in an image somehow associated with a binary? Or are programs often run from source in images like Python or Node?

If binary, you probably don't want `ImageTag`, only `DeclaredTag`, `DirectoryTag`, `LanguageTag`, `"julia"` -- this would be similar to [things we expect to be present during a source checkout rather than at runtime](https://github.com/anchore/syft/blob/main/internal/task/package_tasks.go#L119).

And to answer your other question, I think `PackageTag` is for Conda. It has an unfortunate name in the context of Syft packages and Golang packages and all the other packages 😄

Octogonapus: Julia programs are 99% of the time ran from source. There are [ways to make a binary](https://julialang.github.io/PackageCompiler.jl/dev/sysimages.html) but you still can't escape the source, or at least a Manifest.toml file

syft/pkg/cataloger/julia/cataloger.go
```diff
@@ -0,0 +1,127 @@
...
+const (
+	runtimeKind  = "runtime"
+	testKind     = "test"
+	optionalKind = "optional"
+)
+
+type juliaPackageKey struct {
+	projectDir string
+	name       string
+	uuid       string
```
kzantow, line 24: is there a need for anything beyond projectDir and uuid here? is uuid sometimes empty string?

syft/pkg/cataloger/julia/cataloger.go
```diff
@@ -0,0 +1,127 @@
+/*
+Package julia provides a concrete Cataloger implementation for packages relating to the Julia language ecosystem.
+*/
+package julia
+
+import (
+	"path/filepath"
```
kzantow: all path operations should be handled with the `path` package rather than `path/filepath` -- everything in the resolvers should have been normalized

syft/pkg/cataloger/julia/cataloger.go
Octogonapus: You are right. Removed. UUID is never empty.

syft/pkg/cataloger/julia/cataloger.go
Octogonapus: Removed all uses of `path/filepath`.

Octogonapus:
The issue with `internal/packagemetadata/generated.go` appears to be a formatting diff with the generated code. It was already formatted on main so I kept it that way, but the generator appears not to run `go fmt`

PR #4841, opened 2026-04-29
Adds a binary classifier that identifies the syft tool itself

### LINKED ISSUE #4842: Update go module versions when binary classifier is the only thing that finds them
The binary cataloger doesn't currently recognize the syft binary itself

PR: #4841

jfjrh2014:
Adds a `syft-binary` classifier so syft identifies itself when scanning images/filesystems containing a syft binary. Matches the goreleaser ldflag      
  string embedded in release binaries, anchored on ` -X main.gitCommit=` to avoid false-matching syft's own help text.
                                                                                                                                                          
  ## Type of change                                      

  - [x] New feature

  ## Checklist

  - [x] Unit tests added (11 versions, `0.90.0` → `1.40.1`)                                                                                               
  - [x] Tested against `anchore/syft:*` images, no regressions

### DIFF
--- a/syft/pkg/cataloger/binary/classifier_cataloger_test.go
+++ b/syft/pkg/cataloger/binary/classifier_cataloger_test.go
@@ -2227,6 +2227,127 @@ func Test_Cataloger_PositiveCases(t *testing.T) {
 				Metadata:  metadata("envoy-binary"),
 			},
 		},
+		{
+			logicalFixture: "syft/0.90.0/linux-arm64",
+			expected: pkg.Package{
+				Name:      "syft",
+				Version:   "0.90.0",
+				Type:      "binary",
+				PURL:      "pkg:golang/github.com/anchore/syft@0.90.0",
+				Locations: locations("syft"),
+				Metadata:  metadata("syft-binary"),
+			},
+		},
+		{
+			logicalFixture: "syft/0.95.0/linux-arm64",
+			expected: pkg.Package{
+				Name:      "syft",
+				Version:   "0.95.0",
+				Type:      "binary",
+				PURL:      "pkg:golang/github.com/anchore/syft@0.95.0",
+				Locations: locations("syft"),
+				Metadata:  metadata("syft-binary"),
+			},
+		},
+		{
+			logicalFixture: "syft/0.100.0/linux-arm64",
+			expected: pkg.Package{
+				Name:      "syft",
+				Version:   "0.100.0",
+				Type:      "binary",
+				PURL:      "pkg:golang/github.com/anchore/syft@0.100.0",
+				Locations: locations("syft"),
+				Metadata:  metadata("syft-binary"),
+			},
+		},
+		{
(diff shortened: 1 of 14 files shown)


kzantow:
Have you tried running Syft against a syft binary and run into problems? I just downloaded the oldest version mentioned in this PR and tried it:

```
% syft ~/Downloads/syft_0.90.0_darwin_amd64/syft | grep syft
github.com/anchore/syft                           v0.90.0                               go-module  

% syft ~/Downloads/syft_0.90.0_linux_arm64/syft | grep syft 
github.com/anchore/syft                           v0.90.0                               go-module  
```

Syft has detected _go modules_ main versions for quite some time [using a similar build flag lookup](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/golang/parse_go_binary.go#L224).

I'm not sure adding a binary classifier here is especially useful.

jfjrh2014:
Worth noting syft already classifies several Go tools through both paths like helm, consul, etcd, kubectl all have binary classifiers despite being detectable via go-module. The two serve different purposes, go-module gives the full dependency graph, while the binary classifier provides a stable identity signal for relationship building and for consumers filtering by cataloger source,adding syft/grype brings them in line with how peer Go tools are already handled. Happy to align either way, drop them, or accept for consistency (I can send grype/ later)

kzantow:
It's a fair point about the other go binaries, but I think the difference between Syft (and Grype/etc.) is that some main go packages were not identified by the go binary cataloger because there were no usable build flags with an included version. Golang build processes and the flags used vary widely so identifying the main package has traditionally been fairly difficult (newer go versions have a place for this in the module metadata). The binary classifier was used only as a last resort to identify these.

You are correct that the go binary cataloger is intended to output the full dependency graph as best it can, and while the binary classifier is not doing this, it is generally only used as when Syft is not able to detect common packages using any other more reliable method. In fact, it should include the dependency information but this would be unreliable and difficult to encode across all the versions of the binaries that are identified.

I'm still having a hard time understanding what benefit detecting the main Syft package name twice would provide.

wagoodman:
It sounds like what you're really asking is the ability to not catalog dependencies for packages that we find, this really isn't about syft binary detection (since we already detect this with no modifications).

>  the binary classifier provides a stable identity signal for relationship building and for consumers filtering by cataloger source

I don't think I fully understand this. What are the specific reasons why the binary cataloger must raise up syft and grype when there is already a cataloger doing this? The point of all catalogers is shared: to raise up package level information and as much underlying raw information we find about that package with it (in the `.metadata` section).

If the binary cataloger has a different purpose I'd like to know how you are using it.

I agree with @kzantow here that we probably will not be adding a syft/grype binary cataloger pattern since we are already raising up package-level detail (which is much richer today than with the binary cataloger).

wagoodman:
Actually, I didn't realize there was already an issue open. I'm going to close this for now and we can chat further about your use case there.

PR #4843, opened 2026-04-30
feat(java): add hierarchical Maven dependency graph resolution for Java archives

### LINKED ISSUE #4921: Opt-in hierarchical dependency resolution for Java archives

Opt-in **hierarchical dependency resolution for Java archives**, replacing the current flat model where every nested JAR is recorded as a direct dependency of the root archive. The feature would expose two independent, offline inputs and produce richer relationship metadata.

**New configuration (both default off, no behavior change when disabled):**

| Option | Default | Effect |
| :---- | :---- | :---- |
| `java.use-embedded-pom-dependencies` | `false` | Build a dependency graph from `pom.xml` files already bundled inside the JAR |
| `java.maven-dependency-tree` | `""` | Path to the output of `mvn dependency:tree -DoutputFile=…`. Highest priority when set; reflects Maven's actual resolution including BOM / `dependencyManagement` overrides |

**New relationship metadata** (carried on `DependencyOf` edges when the feature is on):

```go
type DependencyRelationshipData struct {
    Depth              int    // 0 = direct, 1+ = transitive
    Scope              string // compile, runtime, test, …
    IsDirectDependency bool
    IntendedParentID   string // Maven coord; resolved in post-processing
}
```

**Output shape (illustrative):**

Today — flat:

```
spring-jcl          ──dependency-of──▶ my-app
spring-core         ──dependency-of──▶ my-app
jackson-core        ──dependency-of──▶ my-app
jackson-databind    ──dependency-of──▶ my-app
```

Proposed — hierarchical:

```
spring-jcl          ──dependency-of──▶ spring-core       {depth: 1, scope: compile}
spring-core         ──dependency-of──▶ my-app            {depth: 0, scope: compile}
jackson-core        ──dependency-of──▶ jackson-databind  {depth: 1, scope: compile}
jackson-databind    ──dependency-of──▶ my-app            {depth: 0, scope: compile}
```

**Why is this needed**:
Today, Syft's Java archive cataloger flattens every nested JAR into a direct dependency of the root package. There is no way to distinguish "this dep was declared in my `pom.xml`" from "this dep was pulled in 7 levels deep by something else."

Measured against a real Spring Boot application:

| Metric | Current Syft | Ground truth |
| :---- | :---- | :---- |
| Java packages in SBOM | 362 | 362 |
| Marked as direct dependency of root | 362 (100%) | 48 (13%) |
| Transitive dependencies | 0 | 302 |
| Max real dependency depth | 0 | 7 |

Consumers that rely on SBOM dependency structure cannot get correct answers from the current output:

- **Vulnerability prioritization** — a CVE in a deep transitive can be triaged differently than one in a direct dep, but only if the depth is recorded.  
- **Dependency-path reporting** — "what pulled in `log4j-core`?" is unanswerable when everything points at the root.  
- **Format fidelity** — both SPDX and CycloneDX model hierarchical dependencies; the data just isn't populated for Java archives today.

**Additional context**:
**Relationship to #3363 (already shipped)** ("feat: Java dependency graph information", merged Oct 2024\) introduced `java.resolve-transitive-dependencies` and the `internal/maven` resolver. 

**This proposal is complementary, not overlapping** — it builds on the foundation \#3363 laid:

|  | #3363 (shipped) | #4834 (This proposal) |
| :---- | :---- | :---- |
| Primary input | Source-scan `pom.xml` files | JAR-archive scans (embedded POMs or pre-generated `mvn dependency:tree`) |
| Network required | Yes, when `use-network` is on | **No** — all inputs are files on disk |
| Relationship output | Flat `DependencyOf` edges | Hierarchical `DependencyOf` with depth, scope, IsDirectDependency |
| Depth handling | Parent passed one level through nested archives | Arbitrary depth (7 observed in real data) with deferred parent resolution |
| Version-mismatch handling | Exact match | 4-tier flexible match (61% of real packages have BOM-managed version drift) |
| Orphaned parents | Relationship dropped if parent isn't in package set | Skip-to-nearest-ancestor fallback |

The two features can be enabled together. **No existing \#3363 behavior changes.** 

Specifically untouched: the `resolve-transitive-dependencies` flag and its semantics, the `internal/maven` package, and source-scan `parse_pom_xml` paths.

**Design overview**   
Two independent inputs feed a single in-memory `DependencyGraph` built at depth=0 (outermost archive). When neither flag is on, no graph is built and behavior is byte-identical to current Syft.

**Stage 1 — archive cataloging** (in `syft/pkg/cataloger/java/`)

- Build the `DependencyGraph` once, at depth=0 on the outermost archive  
- Emit relationships annotated with `DependencyRelationshipData { depth, scope, isDirect, intendedParentID }`

**Stage 2 — post-processing** (new, in `internal/relationship/java/`, runs after all catalogers complete)

For each Java relationship, choose one branch:

| Case | Condition | Action |
| :---- | :---- | :---- |
| 1 | Has `IntendedParentID` and parent is in `pkgIndex` | Resolve to that parent |
| 1b | Has `IntendedParentID` but parent is missing from the SBOM | Skip-to-nearest-ancestor via the Maven tree |
| 2 | No `IntendedParentID` but a graph exists | Enrich depth/scope from the graph |
| 3 | No graph or no match | Keep relationship unchanged |

**Key design decision:** package identity isn't fully known during cataloging — packages are discovered incrementally across catalogers. The Java cataloger stores the *intended* parent as a Maven coordinate string; a post-processing step resolves it once the package index is complete.

**Flexible matching** — 220/362 (61%) of real packages have version mismatches between the bundled JAR and the Maven tree (BOM-managed versions, `dependencyManagement` overrides). 

Both lookups fall back through tiers: full coord → groupId:artifactId → artifactId:version → artifactId.

**Skip-to-nearest-ancestor** — for orphaned packages whose Maven parent never appears in the SBOM (POM-only starters, shaded JARs), the post-processor walks up the Maven tree to find the nearest SBOM-present ancestor. Validation handles 43 orphans across 20 missing parents.

**Validation** Tested against a real Spring Boot application (362 Java packages, depth up to 7). With the feature enabled:

- \~48 packages correctly identified as direct (matches POM declarations)  
- \~259 packages resolved to their real transitive parent  
- \~43 orphaned packages handled via skip-to-nearest-ancestor  
- 12 packages not in Maven tree kept unchanged (outside feature scope)

**Out of scope (independent follow-ups)** These are intentionally deferred to keep this proposal small:

- Scope-specific SPDX relationship types (`BUILD_DEPENDENCY_OF`, `TEST_DEPENDENCY_OF`)  
- Synthetic packages for POM-only parents (currently handled by skip-to-nearest-ancestor)

Each will get its own issue if maintainers want to discuss them.

1. **Open questions** **Naming vs. \#3363.** Both `resolve-transitive-dependencies` (existing) and `use-embedded-pom-dependencies` (proposed) relate to transitive deps but operate on different inputs. Is the proposed naming clear enough, or should the new flags live under a different namespace (e.g., `java.dependency-graph.*`)?  
     
2. **Post-processor seam.** Should the post-processor live in `internal/relationship/java/` (registered in `finalizeRelationships`), or inside `syft/pkg/cataloger/java/`?  
     
3. **`DependencyRelationshipData` location.** In `syft/pkg/cataloger/java/` (current proposal) or somewhere more neutral, given future SPDX/CycloneDX enrichments will read from it?  
     
4. **PR shape.** A working prototype exists at [\#4843](https://github.com/anchore/syft/pull/4843) (\~3,470 additions). I'd like to split it into 5 stacked PRs (\~200–850 lines each, off-by-default flag throughout) since most syft feature PRs that get merged are well under 500 lines. Happy to follow whatever slicing you prefer — fold tests into each feature PR, ship as one, anything else.  
     
   **Heads-up on testing across the stack:** end-to-end behavior (a real JAR scan producing hierarchical relationships) only becomes observable once PRs 1–4 of the stack land.   
     
- PRs 1 and 2 add isolated types and a parser with no callers.   
- PR 3 wires the graph into the archive parser but the post-processor isn't registered yet.   
- PR 4 registers the post-processor — that's the first PR where flipping the flag on produces visible output.   
- PR 5 is integration-test coverage. 

  Each individual PR will be unit-testable in isolation, and [\#4843](https://github.com/anchore/syft/pull/4843) will remain available as the integration reference if you'd like to validate the full feature against a real archive at any point during review.
-------------------
Happy to iterate on any of this before opening the split PRs. Pointers to other catalogers that have done similar incremental rollouts welcome.

Miablo:
# Description

This PR adds hierarchical dependency graph resolution for Java archives, replacing the current flat model where all nested JARs are treated as direct dependencies. The feature is opt-in via two new configuration options and is off by default, ensuring zero behavioral change for existing users.

## The Problem: 

When Syft catalogs a Java archive (e.g., a Spring Boot fat JAR), every nested JAR gets a dependency-of relationship pointing directly to the root archive. There's no way to distinguish direct vs. transitive dependencies or know which dependency pulled in which.

## The Solution: 
This PR builds a proper dependency tree that captures parent-child relationships, depth, and Maven scope, using either a pre-generated mvn dependency:tree file or embedded pom.xml files within the archive.

### New configuration options

Configuration (new CLI/YAML options):
| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `java.use-embedded-pom-dependencies` | bool | `false` | Build graph from embedded `pom.xml` files within JARs |
| `java.maven-dependency-tree` | string | `""` | Path to pre-generated `mvn dependency:tree` output file |

When maven-dependency-tree is provided, it takes priority over embedded POM analysis.

#### Option A: Using a pre-generated Maven dependency tree (most accurate)

Generate the tree file from your project
  `mvn dependency:tree -DoutputFile=deps.txt`

Scan with hierarchical resolution
 ```
syft scan dir:./target \
    -o spdx-json \
    --java-maven-dependency-tree deps.txt
```

 #### Option B: Using embedded POMs (no build tools required)

```
  syft scan ./app.jar \
    -o cyclonedx-json \
    --java-use-embedded-pom-dependencies
```

 #### Option C: Via .syft.yaml configuration

```
  java:
    maven-dependency-tree: ./deps.txt
    # OR
    use-embedded-pom-dependencies: true
```

### Example relationship output changes

  Before (flat model) — all dependencies point to root:

```
  spring-jcl            ──dependency-of──▶  my-app
  spring-core           ──dependency-of──▶  my-app
  jackson-databind      ──dependency-of──▶  my-app
  jackson-core          ──dependency-of──▶  my-app
  jackson-annotations   ──dependency-of──▶  my-app
```

  After (hierarchical model) — relationships reflect the actual Maven tree:

```
  spring-jcl            ──dependency-of──▶  spring-core          {depth: 1, scope: "compile"}
  spring-core           ──dependency-of──▶  my-app               {depth: 0, scope: "compile"}
  jackson-core          ──dependency-of──▶  jackson-databind     {depth: 1, scope: "compile"}
  jackson-annotations   ──dependency-of──▶  jackson-databind     {depth: 1, scope: "compile"}
  jackson-databind      ──dependency-of──▶  my-app               {depth: 0, scope: "compile"}
```

Each relationship carries DependencyRelationshipData:
```
  {
    "depth": 1,
    "scope": "compile",
    "isDirectDependency": false
  }
```

Note Regarding Relationship Metadata:

- Syft JSON — `r.Data` is written to Metadata field on the relationship model. So depth/scope/isDirectDependency are preserved in Syft JSON output.
- SPDX — only writes `RefA`, Relationship (type string), `RefB`, and `RelationshipComment`. Data is dropped.                                                                                                                 
- CycloneDX — only writes the dependency graph structure (who depends on whom). Data is dropped.

So the DependencyRelationshipData survives in Syft JSON only. In SPDX and CycloneDX, the structural benefit is preserved (correct parent-child edges), but the metadata (depth, scope, isDirectDependency) is lost.

### How graph resolution works

1. Graph building (during cataloging, depth=0 only): Priority chain is Maven tree file → embedded POMs
2. Relationship enrichment (during cataloging): Each relationship gets depth, scope, and intended parent from the graph. When the intended parent package hasn't been cataloged yet, the parent Maven ID is stored as intendedParentID for deferred resolution
3. Post-processing (after all catalogers complete): ResolveHierarchicalDependencies runs in the relationship finalization pipeline, resolving deferred parent IDs via a 3-tier package index (exact → groupId:artifactId → artifactId). 

 
Falls back to the nearest ancestor present in the SBOM when the intended parent package isn't found

### Validated against

- Real-world 461-node Maven dependency tree (68 direct deps, max depth 7) with classifiers, [INFO] prefixes, [WARNING] blocks, and Maven banner/footer noise — all parsed correctly
- Existing test suite passes with no regressions

#### What changed

New files (8):
- syft/pkg/cataloger/java/dependency_graph.go — DependencyGraph/DependencyNode in-memory tree with 4-tier flexible matching (exact → GA → AV → artifactId)
- syft/pkg/cataloger/java/maven_id.go — Exported MavenID wrapper for cross-package use
- syft/pkg/cataloger/java/maven_tree_parser.go — Parses mvn dependency:tree text output (4/5/6-part coordinates, classifiers, [INFO] prefixes, optional markers, managed from annotations)
- syft/pkg/cataloger/java/relationship_metadata.go — DependencyRelationshipData carried on artifact.Relationship.Data
- internal/relationship/java/java_hierarchical_relationships.go — Post-processor that resolves deferred parent IDs and enriches relationships after all catalogers complete

Modified files (5):
- syft/pkg/cataloger/java/archive_parser.go — Builds dependency graph at depth=0, enriches relationships with depth/scope/intended-parent during cataloging
- syft/pkg/cataloger/java/config.go — Two new config fields: UseEmbeddedPOMDependencies, MavenDependencyTreeFile
- cmd/syft/internal/options/java.go — Corresponding CLI/YAML options
- syft/cataloging/relationships.go — Passes Java config through to post-processor
- internal/task/relationship_tasks.go — Hooks post-processor into relationship finalization pipeline

Tests (5 new test files, ~1,700 lines):
- Unit tests for dependency graph, tree parser, relationship metadata
- End-to-end integration test with benchmarks
- SPDX DependencyOf relationship lookup test

Co-Authored-By: Claude Code
 
## Type of change

## Checklist

## Issue references

Create to fix DCO issues on original PR #4803

### DIFF
--- a/internal/relationship/java/java_hierarchical_relationships.go
+++ b/internal/relationship/java/java_hierarchical_relationships.go
@@ -136,7 +136,7 @@ func processRelationship(
 
 	// Case 1: Has IntendedParentID — resolve deferred parent
 	if hasRelData && relData.IntendedParentID != "" {
-		return resolveIntendedParent(rel, relData, fromPkg, packages, pkgIndex, depGraph, rootPkg)
+		return resolveIntendedParent(rel, relData, fromPkg, pkgIndex, depGraph, rootPkg)
 	}
 
 	// Case 2: No IntendedParentID, but depGraph exists — enrich from tree
@@ -153,7 +153,6 @@ func resolveIntendedParent(
 	rel artifact.Relationship,
 	relData javaCataloger.DependencyRelationshipData,
 	fromPkg *pkg.Package,
-	packages *pkg.Collection,
 	pkgIndex map[string]*pkg.Package,
 	depGraph *javaCataloger.DependencyGraph,
 	rootPkg *pkg.Package,
@@ -185,7 +184,7 @@ func resolveIntendedParent(
 	if depGraph != nil {
 		parentMavenID := parseMavenID(relData.IntendedParentID)
 		parentNode := depGraph.FindNodeByMavenID(parentMavenID)
-		if ancestorPkg := findNearestAncestorInSBOM(depGraph, parentNode, pkgIndex, rootPkg); ancestorPkg != nil {
+		if ancestorPkg := findNearestAncestorInSBOM(parentNode, pkgIndex, rootPkg); ancestorPkg != nil {
 			rel.To = *ancestorPkg
 		}
 	}
@@ -225,7 +224,6 @@ func enrichFromGraph(
 // findNearestAncestorInSBOM walks up the Maven tree from a missing parent node
 // until finding an ancestor whose package IS in the SBOM.
 func findNearestAncestorInSBOM(
-	depGraph *javaCataloger.DependencyGraph,
 	missingNode *javaCataloger.DependencyNode,
 	pkgIndex map[string]*pkg.Package,
 	rootPkg *pkg.Package,
--- a/internal/relationship/java/java_hierarchical_relationships.go
+++ b/internal/relationship/java/java_hierarchical_relationships.go
@@ -17,11 +17,6 @@ import (
 // the Maven dependency tree when available.
 func ResolveHierarchicalDependencies(accessor sbomsync.Accessor, cfg cataloging.RelationshipsConfig) {
 	hasTreeFile := cfg.JavaMavenDependencyTreeFile != ""
-	hasEmbeddedPOMs := cfg.JavaUseEmbeddedPOMDependencies
-
-	if !hasTreeFile && !hasEmbeddedPOMs {
-		return
-	}
 
 	// Step 1: Build package index from ALL Java packages
 	pkgIndex := buildPackageIndex(accessor)
--- a/internal/relationship/java/java_hierarchical_relationships.go
+++ b/internal/relationship/java/java_hierarchical_relationships.go
@@ -1,0 +1,346 @@
+package java
+
+import (
+	"strings"
+
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/internal/sbomsync"
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/cataloging"
+	"github.com/anchore/syft/syft/pkg"
+	javaCataloger "github.com/anchore/syft/syft/pkg/cataloger/java"
+	"github.com/anchore/syft/syft/sbom"
+)
+
+// ResolveHierarchicalDependencies runs after all catalogers have completed. It resolves
+// deferred parent IDs in Java dependency relationships and enriches depth/scope from
+// the Maven dependency tree when available.
+func ResolveHierarchicalDependencies(accessor sbomsync.Accessor, cfg cataloging.RelationshipsConfig) {
+	if cfg.JavaMavenDependencyTreeFile == "" {
+		// No tree file configured — nothing to resolve or enrich.
+		// Deferred parent IDs (IntendedParentID) are only set when a dependency graph
+		// was built during cataloging, which requires either a tree file or embedded POMs.
+		// Even if embedded POMs were used, the enrichment in Phase 3 has already occurred.
+		// This post-processor only adds value when a tree file is available for global enrichment.
+		return
+	}
+
+	// Step 1: Build package index from ALL Java packages
+	pkgIndex := buildPackageIndex(accessor)
+	if len(pkgIndex) == 0 {
+		return
+	}
+
+	// Step 2: Build dependency graph from Maven tree file
+	depGraph := buildGraphFromTreeFile(cfg.JavaMavenDependencyTreeFile)
+
+	// Step 3: Process relationships
+	var updatedRelationships []artifact.Relationship
+	var rootPkg *pkg.Package
+
+	accessor.ReadFromSBOM(func(s *sbom.SBOM) {
+		// Find root package (the one that matches the graph root)
+		if depGraph != nil && depGraph.Root != nil {
+			rootID := depGraph.Root.ID
+			rootKey := javaCataloger.NewMavenID(rootID.GroupID, rootID.ArtifactID, rootID.Version).String()
+			rootPkg = findPackageByMavenID(pkgIndex, rootKey)
+		}
+
+		for _, rel := range s.Relationships {
+			newRel := processRelationship(rel, s.Artifacts.Packages, pkgIndex, depGraph, rootPkg)
+			updatedRelationships = append(updatedRelationships, newRel)
+		}
+	})
+
+	// Step 4: Write updated relationships back
+	accessor.WriteToSBOM(func(s *sbom.SBOM) {
+		s.Relationships = updatedRelationships
+	})
+}
+
+// buildPackageIndex creates a multi-tier index of Java packages for flexible matching.
+// Packages are indexed by:
+// - Full Maven ID: groupId:artifactId:version (exact)
+// - groupId:artifactId (version-flexible)
+// - artifactId only (last resort)
+func buildPackageIndex(accessor sbomsync.Accessor) map[string]*pkg.Package {
+	pkgIndex := make(map[string]*pkg.Package)
+
+	accessor.ReadFromSBOM(func(s *sbom.SBOM) {
+		for p := range s.Artifacts.Packages.Enumerate(pkg.JavaPkg, pkg.JenkinsPluginPkg) {
+			pCopy := p
+			mavenID := extractMavenID(&pCopy)
+			if mavenID == "" {
+				continue
+			}
+
+			// Tier 1: full Maven ID (exact match — highest priority, never overwritten)
+			if _, exists := pkgIndex[mavenID]; !exists {
+				pkgIndex[mavenID] = &pCopy
+			}
+
+			// Tier 2: groupId:artifactId (version-flexible)
+			gaKey := extractGroupArtifact(mavenID)
+			if gaKey != "" {
+				if _, exists := pkgIndex[gaKey]; !exists {
+					pkgIndex[gaKey] = &pCopy
+				}
+			}
+
+			// Tier 3: artifactId only (last resort)
+			artifactID := extractArtifactID(mavenID)
+			if artifactID != "" {
+				if _, exists := pkgIndex[artifactID]; !exists {
+					pkgIndex[artifactID] = &pCopy
+				}
+			}
+		}
+	})
+
+	return pkgIndex
+}
+
+// buildGraphFromTreeFile parses a Maven dependency tree file and returns a DependencyGraph.
+func buildGraphFromTreeFile(filePath string) *javaCataloger.DependencyGraph {
+	tree, err := javaCataloger.ParseMavenDependencyTreeFile(filePath)
+	if err != nil {
+		log.WithFields("error", err, "file", filePath).Warn("failed to parse Maven dependency tree file for post-processing")
+		return nil
+	}
+	return tree.ToInternalGraph()
+}
+
+// processRelationship handles a single relationship, applying the three processing cases.
+func processRelationship(
+	rel artifact.Relationship,
+	packages *pkg.Collection,
+	pkgIndex map[string]*pkg.Package,
+	depGraph *javaCataloger.DependencyGraph,
+	rootPkg *pkg.Package,
+) artifact.Relationship {
+	// Only process DependencyOf relationships
+	if rel.Type != artifact.DependencyOfRelationship {
+		return rel
+	}
+
+	// Only process relationships where "From" is a Java package
+	fromPkg := packages.Package(rel.From.ID())
+	if fromPkg == nil {
+		return rel
+	}
+	if fromPkg.Language != pkg.Java {
+		return rel
+	}
+
+	relData, hasRelData := rel.Data.(javaCataloger.DependencyRelationshipData)
+
+	// Case 1: Has IntendedParentID — resolve deferred parent
+	if hasRelData && relData.IntendedParentID != "" {
+		return resolveIntendedParent(rel, relData, fromPkg, packages, pkgIndex, depGraph, rootPkg)
+	}
+
+	// Case 2: No IntendedParentID, but depGraph exists — enrich from tree
+	if depGraph != nil {
+		return enrichFromGraph(rel, fromPkg, depGraph)
+	}
+
+	// Case 3: No tree or not found — keep original
+	return rel
+}
+
+// resolveIntendedParent handles Case 1: the relationship has a deferred IntendedParentID.
+func resolveIntendedParent(
+	rel artifact.Relationship,
+	relData javaCataloger.DependencyRelationshipData,
+	fromPkg *pkg.Package,
+	packages *pkg.Collection,
+	pkgIndex map[string]*pkg.Package,
+	depGraph *javaCataloger.DependencyGraph,
+	rootPkg *pkg.Package,
+) artifact.Relationship {
+	parentPkg := findPackageByMavenID(pkgIndex, relData.IntendedParentID)
+
+	if parentPkg != nil {
+		// Case 1a: Parent found — resolve and enrich
+		depth := relData.Depth
+		scope := relData.Scope
+
+		if depGraph != nil {
+			mavenID := extractMavenIDStruct(fromPkg)
+			if node := depGraph.FindNodeByMavenID(mavenID); node != nil {
+				depth = node.Depth - 1
+				if depth < 0 {
+					depth = 0
+				}
+				scope = node.Scope
+			}
+		}
+
+		rel.To = *parentPkg
+		rel.Data = javaCataloger.NewDependencyRelationshipData(depth, scope)
+		return rel
+	}
+
+	// Case 1b: Parent NOT found — skip to nearest ancestor
+	if depGraph != nil {
+		parentMavenID := parseMavenID(relData.IntendedParentID)
+		parentNode := depGraph.FindNodeByMavenID(parentMavenID)
+		if ancestorPkg := findNearestAncestorInSBOM(depGraph, parentNode, pkgIndex, rootPkg); ancestorPkg != nil {
+			rel.To = *ancestorPkg
+		}
+	}
+
+	// Clear IntendedParentID — resolution attempted
+	depth := relData.Depth
+	scope := relData.Scope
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: internal/relationship/java/java_hierarchical_relationships.go [hunk 1]
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/anchore/syft/syft/cataloging"
+	"github.com/anchore/syft/syft/pkg"
+	javaCataloger "github.com/anchore/syft/syft/pkg/cataloger/java"
+	"github.com/anchore/syft/syft/sbom"
+)
+
+// ResolveHierarchicalDependencies runs after all catalogers have completed. It resolves
+// deferred parent IDs in Java dependency relationships and enriches depth/scope from
+// the Maven dependency tree when available.
+func ResolveHierarchicalDependencies(accessor sbomsync.Accessor, cfg cataloging.RelationshipsConfig) {
```
kzantow, line 18: Why is this functionality in the `internal/relationship/..` package instead of `pkg/cataloger/java`? Does this need to apply across catalogers? It seems like results returned from the [maven pom cataloger](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/parse_pom_xml.go) and [java archive cataloger](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/java/archive_parser.go#L78) should not be crossed, if that's the intention here: one would be for a source checkout and the other for a deployed application.

syft/cataloging/relationships.go
```diff
@@ -13,6 +13,15 @@ type RelationshipsConfig struct {
...
 	ExcludeBinaryPackagesWithFileOwnershipOverlap bool `yaml:"exclude-binary-packages-with-file-ownership-overlap" json:"exclude-binary-packages-with-file-ownership-overlap" mapstructure:"exclude-binary-packages-with-file-ownership-overlap"`
+
+	// JavaMavenDependencyTreeFile is the path to a pre-generated Maven dependency tree file.
+	// When provided, Java dependency relationships are enriched with accurate depth and scope during post-processing.
+	JavaMavenDependencyTreeFile string `yaml:"java-maven-dependency-tree-file" json:"java-maven-dependency-tree-file" mapstructure:"java-maven-dependency-tree-file"`
+
+	// JavaUseEmbeddedPOMDependencies indicates that the Java cataloger built a dependency graph from
+	// embedded pom.xml files. When true, the post-processor resolves deferred parent IDs in relationships
+	// even without a Maven dependency tree file.
+	JavaUseEmbeddedPOMDependencies bool `yaml:"java-use-embedded-pom-dependencies" json:"java-use-embedded-pom-dependencies" mapstructure:"java-use-embedded-pom-dependencies"`
```
kzantow, line 24: Java-specific options shouldn't be in these options, probably. I think this a function of the actual relationship building code not being in the `pkg/cataloger/java` package, which it probably should be

These are two different features, I think it would be easier to get some of this PR merged if it were split into 2 separate PRs: one for properly building the dependency graph based on maven pom files and a second PR for using the maven dependency tree file.

syft/pkg/cataloger/java/archive_parser.go
```diff
@@ -142,24 +149,37 @@ func (j *archiveParser) parse(ctx context.Context, parentPkg *pkg.Package) ([]pk
...
-				Type: artifact.DependencyOfRelationship,
-			})
+			rel := j.createMainPkgRelationship(mainPkg, parentPkg)
+			relationships = append(relationships, rel)
+		}
+	}
+
+	// Build a package index for looking up packages by Maven ID.
+	// Used when the dependency graph provides hierarchical parent information.
+	pkgIndex := make(map[string]*pkg.Package)
```
kzantow, line 178: I'm confused why there are new lookup maps built here, or why these are not created in a nested function, able to reference the parent set

syft/pkg/cataloger/java/dependency_graph.go
```diff
@@ -0,0 +1,213 @@
...
+				branchVisited[k] = v
+			}
+			g.buildFromPOM(ctx, poms, resolver, depPom, depNode, branchVisited, resolveTransitive, maxDepth, currentDepth+1)
+		}
+	}
+}
+
+// mavenIDKey returns a colon-separated key for use in the NodeMap.
+func mavenIDKey(id maven.ID) string {
+	return fmt.Sprintf("%s:%s:%s", id.GroupID, id.ArtifactID, id.Version)
```
kzantow, line 212: why not just use maven.ID as the key? it's already comparable (can be used as a map key) and avoids making lots more strings. but I think this key should not include version, right? when we're talking about installed jars, the version is not relevant: maven resolves based on its logic and includes only that version for everything, since there's only a single log4j available in the classpath. but the classpath might get altered with more specific versions as we go further down the nesting: for example an EAR might have log4j but a WAR inside might also have a different version of log4j that it provides for its webapps. I realize this could be different based on configuration, but it seems like the most common behavior would be to use the "closest" version as the dependency

syft/pkg/cataloger/java/maven_id.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+
+	"github.com/anchore/syft/syft/pkg/cataloger/java/internal/maven"
+)
+
+// MavenID is an exported wrapper around the internal maven.ID type, allowing
+// packages outside the java cataloger to work with Maven coordinates.
+type MavenID struct {
```
kzantow, line 11: this could just be an alias, but also, the logic to make the dependencies should probably be in this package, not `internal/relationship/...`

syft/pkg/cataloger/java/dependency_graph.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+const DefaultMaxDepth = 10
+
+type DependencyNode struct {
+	ID       maven.ID
+	Scope    string
+	Parent   *DependencyNode
+	Children []*DependencyNode
+	Depth    int
```
kzantow, line 19: I don't understand the `Depth`. With a proper hierarchical dependency graph, this is not necessary, can you explain what the purpose is? It doesn't look like this is included outside the `DependencyGraph`

kzantow (commented):
I haven't given this a complete review, but there are a few things that I thought I would comment on which would change this PR considerably.

internal/relationship/java/java_hierarchical_relationships.go
Miablo: Yeah, good point. This logic does need full SBOM access to resolve deferred parents (packages are discovered incrementally during cataloging, so we can't resolve inline), but I agree it belongs closer to the java cataloger. 

I can move the post-processing into `pkg/cataloger/java` as an exported function and have the relationship framework dispatch to it, if that sounds good? Or, if there is a preferred callback pattern for letting a specific cataloger hook into the post-cataloging phase safely then I can match that.

Let me know which one makes more sense to you and I can get that updated!

syft/cataloging/relationships.go
Miablo: Agreed, I'll move these to the Java cataloger config. They shouldn't pollute the shared relationships config.

Yeah, it makes sense to split this into two PRs by feature like you suggest:
1. Embedded POM dependency graph resolution (building relationships from the pom.xml files already inside the JAR)
2. Maven dependency tree file support (enriching with mvn dependency:tree output)

I can start working on the split and will make sure the PRs incorporate the architectural feedback (moving code to pkg/cataloger/java, cleaning up config placement, etc.).

syft/pkg/cataloger/java/archive_parser.go
Miablo: Yeah... I totally get why the inline map generation looks messy and confusing here. The reason for building the fresh map here is because aux packages can have other aux packages as their Maven parent (not always `mainPkg`) — so it will build the index incrementally as each aux pkg is processed.  I can extract this into a helper function or a cleaner closure so that the intent clearer, if that sounds like a valid approach?

syft/pkg/cataloger/java/maven_id.go
Miablo: Agreed — once the core logic moves to `pkg/cataloger/java` (per your first comment), this wrapper is unnecessary and I'll delete it!

syft/pkg/cataloger/java/dependency_graph.go
Miablo: Yeah that makes sense to use `maven.ID`.  Your EAR/WAR example is that exact reason behind the inclusion of versions, actually. If you think this use case is worth it then the straightforward path is your suggestion: use `maven.ID` as the map key since it includes version, which handles the nested archive case without extra work. If you think no version makes more sense, then the alternative would be keying by `groupId:artifactId` only (no version). That requires a custom two-field struct and I would need to scope the graph per-archive to handle the EAR/WAR case. 

I don't have the expertise or experience with Java and Maven to fully understand the whether this nested case is worth this added complexity. I basically learned a lot of these things while working on these features. So looking to see what you suggest and I can make the changes!

syft/pkg/cataloger/java/dependency_graph.go
Miablo: Yeah, you're right — Depth on `DependencyNode` is totally redundant when you have the parent pointers. The place where depth shows up outside the graph is in `DependencyRelationshipData` in the metadata attached to each relationship in Syft JSON output like this:
```
"artifactRelationships": [
  {
    "parent": "a1b2c3d4e5f6a7b8",
    "child": "c9d0e1f2a3b4c5d6",
    "type": "dependency-of",
    "metadata": {
      "depth": 1,
      "scope": "runtime",
      "isDirectDependency": false
    }
  }
]
```

The thinking was that downstream tools (Grype, policy engines, etc.) could use this metadata without having to re-derive it from the graph. SPDX and CycloneDX don't use any of this — it's Syft JSON only.

I can definitely remove Depth from the node and still compute it at relationship-creation time so it shows up in the output for consumers or I can just drop it everywhere so no metadata on relationships is added and consumers will need to figure out depth from the graph structure themselves. I'm good with either option and can update during the work to spit the PR!

Miablo:
Note: PR has been split -- #4943 is now PR1 and focused on the implementation for `UseEmbeddedPOMDependencies` this PR will then be focused on using the maven tree

PR #4844, opened 2026-04-30
feat: add safe tensor model type to SBOM output

spiffcs:
## feat: add SafeTensors cataloger to syft and safe tensor model type packages to SBOM output

This PR adds a new `safetensors-cataloger` to syft alongside the existing `gguf-cataloger`. It produces `model`-type packages with a new `safetensors-model-info` metadata (name subject to change). 

#### Metadata
- `pkg.SafeTensorsModelInfo` struct (`syft/pkg/safetensors.go`)
    capturing `Format`, `Architecture`, `Quantization`, `Parameters`,
    `TensorCount`, `TotalSize`, `TorchDtype`, `TransformersVersion`,
    `ShardCount`, `UserMetadata`, `MetadataHash`, and `Parts` (other nested or related safetensor files
- JSON schema bumped 16.1.3 → 16.1.4 (`internal/constants.go` + new `schema/json/schema-16.1.4.json`).

### Naming the new packages
Packages are named with the following ranking heuristic:
1. path.Base(config.json _name_or_path)
2. OCI manifest title           <-- Discuss if we should do this (NOT IMPLEMENTED)
3. Architecture-Parameters      (both must be non-empty)
4. parent directory             (dir-scan only)
No Name → drop

### Commands to verify and view the new feature:
```
go run ./cmd/syft --from oci-model docker.io/ai/qwen3.5-safetensors:4B -o syft-json | jq '.artifacts[] | select(.type=="model")'
```
```
go run ./cmd/syft --from oci-model docker.io/ai/nomic-embed-text-v2-moe-safetensors:475M -o syft-json | jq '.artifacts[] | select(.type=="model")'
```

#### Considerations around naming packages:
- https://oci.dag.dev/?image=docker.io/ai/smollm2-vllm:360M

Some Docker AI images (the ones suffixed -vllm) ship a safetensors weight layer and a license layer but no ai.model.file companion or other metadata naming source, there is no hugging face config.json to look up the name. These images currently have no naming source and the merge processor drops them.

We should consider plumbing the manifest values through or adding a way to lookup those values during post process.


syft/pkg/cataloger/ai/parse_safetensors_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/internal/pkgtest"
+)
+
+// buildSafeTensorsFile builds the on-disk bytes of a .safetensors file: an
+// 8-byte little-endian header length followed by the JSON header. Tensor data
+// is omitted because the parser only reads the header.
+func buildSafeTensorsFile(t *testing.T, metadata map[string]string, tensors map[string]safeTensorsEntry) []byte {
```
wagoodman, line 25: we should at least have a test that uses a real partial safetensor file to ground that this approach works. Here we're using our definitions of a safeTensorsEntry to build new test files, which would work regardless of if a real tensor file is actually structured this way.

If you're going to the effort of extracting this from an image, then you might as well have that script captured in the testdata dir as well (for future use cases).

syft/pkg/safetensors.go
```diff
@@ -0,0 +1,59 @@
...
+	// TransformersVersion is the transformers library version recorded in config.json.
+	TransformersVersion string `json:"transformersVersion,omitempty" cyclonedx:"transformersVersion"`
+
+	// ShardCount is the number of .safetensors shards for a sharded model (1 for a
+	// single-file model).
+	ShardCount int `json:"shardCount,omitempty" cyclonedx:"shardCount"`
+
+	// UserMetadata is the optional "__metadata__" map from a .safetensors file header
+	// (string-to-string key/values set by the producer).
+	UserMetadata map[string]string `json:"userMetadata,omitempty" cyclonedx:"userMetadata"`
```
wagoodman: We try not to include maps directly (though it does make sense sometimes), instead we use a stable slice with https://github.com/anchore/syft/blob/e8c6b7151e5b78a0a50fa6505b25f6633b3746c3/syft/pkg/key_value.go#L19

syft/pkg/cataloger/ai/parse_safetensors_model.go
```diff
@@ -0,0 +1,305 @@
...
+	TransformersVersion string   `json:"transformers_version"`
+	NameOrPath          string   `json:"_name_or_path"`
+}
+
+func readSiblingJSON(resolver file.Resolver, p string) *hfConfig {
+	locations, err := resolver.FilesByPath(p)
+	if err != nil || len(locations) == 0 {
+		return nil
+	}
+	rc, err := resolver.FileContentsByLocation(locations[0])
```
spiffcs: @spiffcs check this locations access and see why/how we're looking up the sibling files in the resolver for evidence

syft/pkg/safetensors.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// MetadataHash is an xxhash of the normalized header metadata, providing a stable
+	// identifier for identical model content across repositories or filenames.
+	MetadataHash string `json:"metadataHash,omitempty" cyclonedx:"metadataHash"`
```
wagoodman, line 43: is this purely for the .safetensor file information or for things from an OCI config?

syft/pkg/cataloger/ai/cataloger.go
```diff
@@ -23,3 +24,17 @@ func NewGGUFCataloger() pkg.Cataloger {
...
+// covering three discovery paths:
+//   - **/*.safetensors files (single-file models; header-only parse)
+//   - **/model.safetensors.index.json files (sharded models)
+//   - application/vnd.docker.ai.model.config.v0.1+json / v0.2+json OCI layers
+//     (Docker Model Runner artifacts whose config advertises format=="safetensors")
+func NewSafeTensorsCataloger() pkg.Cataloger {
+	return generic.NewCataloger(safeTensorsCatalogerName).
+		WithParserByGlobs(parseSafeTensorsFile, "**/*.safetensors").
+		WithParserByGlobs(parseSafeTensorsIndex, "**/*.safetensors.index.json").
+		WithParserByMediaType(parseSafeTensorsOCIConfig, dockerAIModelConfigMediaTypes...).
```
wagoodman, line 46: we have two different entrypoints, which is fine by itself, however, on the OCI path we are not observing anything about the safetensor file itself. Since both files exist in both things being analyzed (that is, there is a safetensor file in both the OCI image and the file on disk) we should be consistent about this.

syft/pkg/cataloger/ai/parse_safetensors_oci.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/anchore/syft/internal/unknown"
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/generic"
+)
+
+// Docker AI OCI media types used by Docker Model Runner artifacts.
+const (
+	dockerAIModelFileMediaType = "application/vnd.docker.ai.model.file"
```
wagoodman: what are these? This seems to be an "anything" media type

syft/pkg/cataloger/ai/parse_safetensors_oci.go
```diff
@@ -0,0 +1,245 @@
...
+	case strings.Contains(text, "mit license"):
+		return "MIT"
+	case strings.Contains(text, "bsd 3-clause"):
+		return "BSD-3-Clause"
+	case strings.Contains(text, "bsd 2-clause"):
+		return "BSD-2-Clause"
+	case strings.Contains(text, "gnu general public license") && strings.Contains(text, "version 3"):
+		return "GPL-3.0"
+	}
+	return ""
```
wagoodman: This should be a yaml unmarshal of the contents of the first doc (delimited by `---`):

```
---
title: Apache License 2.0
spdx-id: Apache-2.0
redirect_from: /licenses/apache/
featured: true
hidden: false

description: A permissive license whose main conditions require preservation of copyright and license notices. Contributors provide an express grant of patent rights. Licensed works, modifications, and larger works may be distributed under different terms and without source code.

how: Create a text file (typically named LICENSE or LICENSE.txt) in the root of your source code and copy the text of the license into the file.

note: The Apache Software Foundation <a href="https://apache.org/foundation/license-faq.html#Apply-My-Software">recommends</a> taking the additional step of adding a boilerplate notice to the header of each source file. You can find the notice in the appendix at the very end of the license text.

using:
  Kubernetes: https://github.com/kubernetes/kubernetes/blob/master/LICENSE
  PDF.js: https://github.com/mozilla/pdf.js/blob/master/LICENSE
  Swift: https://github.com/apple/swift/blob/main/LICENSE.txt

permissions:
  - commercial-use
  - modifications
  - distribution
  - patent-use
  - private-use

conditions:
  - include-copyright
  - document-changes

limitations:
  - trademark-use
  - liability
  - warranty

---

                                 Apache License
                           Version 2.0, January 2004
                        http://www.apache.org/licenses/

   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION

   1. Definitions.
...
```

And selecting `.spdx-id` or if not present falling back to `.title`.

syft/pkg/cataloger/ai/parse_safetensors_oci.go
```diff
@@ -0,0 +1,245 @@
...
+// YAML frontmatter with license + base_model; HF config.json carries
+// architectures/torch_dtype/transformers_version; the vnd.docker.ai.license
+// blob is plain license text.
+func enrichFromDockerAILayers(resolver file.Resolver, md *pkg.SafeTensorsModelInfo) (name, license string) {
+	ociResolver, ok := resolver.(file.OCIMediaTypeResolver)
+	if !ok {
+		return "", ""
+	}
+
+	modelFileLocations, err := ociResolver.FilesByMediaType(dockerAIModelFileMediaType)
```
wagoodman: I dont think we need searching for this media type at all

syft/pkg/cataloger/ai/parse_safetensors_oci.go
```diff
@@ -0,0 +1,245 @@
...
+		readAndClassifyDockerAILayer(resolver, loc, md, &configName, &readmeName, &license)
+	}
+
+	name = configName
+	if name == "" {
+		name = readmeName
+	}
+
+	if license == "" {
+		license = readDockerAILicense(resolver, ociResolver)
```
wagoodman: we should capture the location of the license on the package

syft/pkg/safetensors.go
```diff
@@ -0,0 +1,59 @@
...
+// SafeTensors is a simple, safe serialization format for storing tensors, used
+// as the default weight format for Hugging Face transformer models. Syft may
+// populate this struct from three sources:
+//   - a single .safetensors file (header-only parse)
+//   - a sharded model described by model.safetensors.index.json
+//   - a Docker AI OCI model artifact config blob (vnd.docker.ai.model.config.v0.1+json)
+//
+// The Model Name, License, and Version fields have all been lifted up to be on
+// the syft Package.
+type SafeTensorsModelInfo struct {
```
wagoodman, line 7: we should try to keep what is contained on this struct restricted to what we find in the safetensors file itself (or maybe the OCI config too... what does the gguf file do here?).

wagoodman: or this might need a new name / way of structuring the information

syft/pkg/cataloger/ai/parse_safetensors_model.go
```diff
@@ -0,0 +1,313 @@
...
+	)
+
+	return []pkg.Package{p}, nil, unknown.IfEmptyf([]pkg.Package{p}, "unable to parse safetensors file")
+}
+
+// parseSafeTensorsIndex parses a model.safetensors.index.json file for a sharded
+// model. The index lists every tensor and the shard file it lives in; from this
+// we derive tensor count, unique shard count, and (when present) the producer-
+// declared total_size.
+func parseSafeTensorsIndex(_ context.Context, resolver file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
```
wagoodman: I don't think we should try to parse index files yet

syft/pkg/cataloger/ai/processor.go
```diff
@@ -57,3 +57,44 @@ func ggufMergeProcessor(pkgs []pkg.Package, rels []artifact.Relationship, err er
 
 	return namedPkgs, rels, err
 }
+
+// safeTensorsMergeProcessor mirrors ggufMergeProcessor for SafeTensors packages.
+// When scanning an OCI AI artifact, the model-config blob produces one named
+// package and individual .safetensors shard layers (if we ever decide to parse
+// them directly) would produce nameless packages. Any nameless SafeTensors
+// packages are collapsed into the named one's Parts slice.
+func safeTensorsMergeProcessor(pkgs []pkg.Package, rels []artifact.Relationship, err error) ([]pkg.Package, []artifact.Relationship, error) {
```
wagoodman: Is how we're reconciling this correct?

wagoodman: Note that we do have the ability to have the resolver available to us when merging these packages https://github.com/anchore/syft/blob/e8c6b7151e5b78a0a50fa6505b25f6633b3746c3/syft/pkg/cataloger/generic/cataloger.go#L47

wagoodman: This might be where we want to do license fetching , config.json fetching, etc, so we're not doing that redundant work for each safetensor layer we process.

syft/pkg/safetensors.go
spiffcs: MetadataHash is an xxhash over the on-disk SafeTensors header (sorted tensor entries + __metadata__). It is derived ONLY from the safetensors file bytes and not from OCI manifest, layer descriptors, etc.

It's intended that  the same model content scanned via a directory source and via an OCI image produces the same value. If that is not the case then it's a bug.

syft/pkg/cataloger/ai/merge.go
```diff
@@ -0,0 +1,137 @@
...
+// mergeShardsInto folds the per-shard header metadata into merged, returning
+// the summed shard TensorCount and the list of non-empty per-shard hashes for
+// the rollup. Shards carry only the content-derived fields (Quantization,
+// Parameters, UserMetadata), so those are the only fields folded in here.
+func mergeShardsInto(merged *pkg.SafeTensorsModelInfo, shards []pkg.SafeTensorsModelInfo) (shardTensorTotal uint64, hashes []string) {
+	seenKV := map[string]bool{}
+	for _, s := range shards {
+		shardTensorTotal += s.TensorCount
+		firstNonEmpty(&merged.Quantization, s.Quantization)
+		firstNonEmpty(&merged.Parameters, s.Parameters)
```
wagoodman: why are we dropping values across entries instead of accumulating them. This might be from a previous conversation here, but there aren't enough comments here to tell why we're doing this

syft/pkg/cataloger/ai/parse_safetensors.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	headerLen := binary.LittleEndian.Uint64(lenBuf[:])
+	if headerLen == 0 {
+		return nil, fmt.Errorf("safetensors header length is zero")
+	}
+	if headerLen > maxSafeTensorsHeaderSize {
+		return nil, fmt.Errorf("safetensors header size %d exceeds maximum %d", headerLen, maxSafeTensorsHeaderSize)
+	}
+
+	body := make([]byte, headerLen)
```
wagoodman: we're allocating based on a user controlled field, I think this could be a max allocation of 100MB.

syft/pkg/cataloger/ai/parse_safetensors.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	return best
+}
+
+// metadataHash returns a stable xxhash64 over the tensor entries + __metadata__.
+// Tensor keys are sorted to keep the hash deterministic across producers.
+func (h *safeTensorsHeader) metadataHash() string {
+	type entry struct {
+		Name  string           `json:"name"`
+		Entry safeTensorsEntry `json:"entry"`
```
wagoodman: we should be hashing logical content, not exact content. The issue is that `sageTensorsEntry` has a `DataOffsets` field which is highly coupled to the specific build.

syft/pkg/cataloger/ai/processor.go
```diff
@@ -1,59 +1,158 @@
...
+			out = append(out, p)
 		}
 	}
+	return out
+}
 
-	// Largest number of key value
+// groupByParentDir buckets filesystem-scanned packages by the directory their
+// primary-evidence file lives in (the shards of one model share a directory).
+func groupByParentDir(pkgs []pkg.Package) map[string][]pkg.Package {
```
wagoodman, line 116: I think this should state more directly why this is here... we're making the  assumption that there is one model per directory and if there are multiple independent models in one flat dir then they will be merged. This is really hard to tease out from the code alone.

syft/pkg/cataloger/ai/parse_safetensors.go
spiffcs: Added a guard that will error if we're beyond 100MB

syft/pkg/cataloger/ai/parse_safetensors.go
spiffcs: I made an inline type so we can keep the data, but hash correctly in this case. 

Relevant fields from `safeTensorEntry` get moved onto the inline type and are hashed. Now models that are equivalent (besides their `dataOffsets` fields) have equivalent hashes.

syft/pkg/cataloger/ai/processor.go
spiffcs: I updated the comment so it states our one model per directory assumption

spiffcs:
This failure is fixed by a space issue on the runners. We have an upgrade ready over here: https://github.com/anchore/syft/pull/4608  

It should merge before #4844 to resolve the red CI

wagoodman (approved):

PR #4849, opened 2026-05-02
syft/pkg/cataloger/kernel: scan compressed kernel modules

### LINKED ISSUE #4721: Compressed kernel modules are not scanned by the linux-kernel-cataloger
**What happened**:
When running Syft on a filesystem from a distribution that has the `CONFIG_MODULE_COMPRESS` kernel config option enabled, Syft does not scan the kernel module finds and thus reports a lot less packages than on a non-compressed distribution.

For example, on my custom distribution based on Debian 13, here is the total size of the `/lib/modules` directory and the example of a dir:
```sh
# du -sh lib/modules/
107M	lib/modules/
# ls lib/modules/6.12.74+deb13+1-rt-amd64/kernel/drivers/net/ethernet/intel/ice
-rw-r--r--. 1 root root 494K Mar  8 19:54 ice.ko.xz
```
`syft scan /` on the filesystem only finds 2 packages (probably the vmlinux ones):
```
[0003] DEBUG discovered 2 packages cataloger=linux-kernel-cataloger
[0003]  INFO task completed elapsed=35.271986ms task=linux-kernel-cataloger
```

Way more packages should be detected. Here is an example for a scan on a Debian 12 filesystem that does _not_ have the `CONFIG_MODULE_COMPRESS` kernel config option enabled:
```sh
# du -sh lib/modules
396M	lib/modules
# ls lib/modules/6.1.0-41-rt-amd64/kernel/drivers/net/ethernet/intel/ice
ice.ko
```
```
[0003] DEBUG discovered 4021 packages cataloger=linux-kernel-cataloger
... lot of "missing version" messages
[0003] DEBUG package with missing version, stubbing with default value location=/usr/lib/modules/6.1.0-41-rt-amd64/kernel/sound/soc/intel/skylake/snd-soc-skl.ko pkg=Pkg(name="snd_soc_skl" version="" type="linux-kernel-module" id="5cd10619b7945d61")
[0003]  INFO task completed elapsed=850.796168ms task=linux-kernel-cataloger
```

- Start a Debian 13 VM
- Run a syft scan on /

It looks like this is because the kernel cataloger only scans `.ko` files and not `.ko.xz` ones: https://github.com/anchore/syft/blob/90198da04d9bccd3bade5d3efee28536c34d1186/syft/pkg/cataloger/kernel/capabilities.yaml#L27-L34
The MR that introduced modules compression on Debian 13: https://salsa.debian.org/kernel-team/linux/-/merge_requests/1279

**Environment**:
- Output of `syft version`: `syft 1.42.3`
- OS (e.g: `cat /etc/os-release` or similar): custom image based upon the latest version of `debian trixie`

ChrisJr404:
Closes #4721.

Distros built with `CONFIG_MODULE_COMPRESS` install kernel modules as `.ko.xz` / `.ko.gz` / `.ko.zst` (Debian Bookworm+, Fedora 38+, recent Arch images, etc.). The `linux-kernel-cataloger` only globs `**/lib/modules/**/*.ko`, so on these systems syft sees zero kernel-module packages even though `/lib/modules` contains hundreds of MB of them.

### Changes

- **`cataloger.go`**: extend `kernelModuleGlobs` with `*.ko.xz`, `*.ko.gz`, `*.ko.zst`.
- **`parse_linux_kernel_module_file.go`**: new `openMaybeCompressedModule` helper that detects compression from the file extension and wraps the reader in `xz.NewReader` / `gzip.NewReader` / `zstd.NewReader` before handing it to `unionreader.GetUnionReader`. The union reader's existing fallback buffers the decompressed bytes so the ELF parser still gets the `io.ReaderAt` + `io.Seeker` it needs. Plain `.ko` files take a no-op path.
- **`parse_linux_kernel_module_file_test.go`**: new table-driven unit test that round-trips a payload through each compression format plus a malformed-xz negative case.
- **`go.mod`**: promote `github.com/klauspost/compress` from indirect to direct (we now import its `zstd` package). `github.com/ulikunitz/xz` was already direct.

### Verification

```
$ go build ./syft/pkg/cataloger/kernel/
$ go test -count=1 -run TestOpenMaybeCompressedModule ./syft/pkg/cataloger/kernel/
ok  	github.com/anchore/syft/syft/pkg/cataloger/kernel	0.040s
```

The full `Test_KernelCataloger` integration test wasn't run locally — it pulls a docker image fixture and the test box was disk-constrained. Happy to add an integration testdata image with a `.ko.xz` module if you'd prefer.

### DIFF
--- a/syft/pkg/cataloger/kernel/parse_linux_kernel_module_file.go
+++ b/syft/pkg/cataloger/kernel/parse_linux_kernel_module_file.go
@@ -1,10 +1,15 @@
 package kernel
 
 import (
+	"compress/gzip"
 	"context"
 	"debug/elf"
 	"fmt"
+	"io"
 	"strings"
+
+	"github.com/klauspost/compress/zstd"
+	"github.com/ulikunitz/xz"
 
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/file"
@@ -16,7 +21,18 @@ import (
 const modinfoName = ".modinfo"
 
 func parseLinuxKernelModuleFile(ctx context.Context, _ file.Resolver, _ *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
-	unionReader, err := unionreader.GetUnionReader(reader)
+	// Distros built with CONFIG_MODULE_COMPRESS install kernel modules as
+	// .ko.xz / .ko.gz / .ko.zst (issue #4721). Detect compression from the
+	// file extension and decompress before handing the bytes to the ELF
+	// parser. The unionreader fallback below buffers the decompressed bytes
+	// so the ELF reader gets the io.ReaderAt + io.Seeker it needs.
+	rc, err := openMaybeCompressedModule(reader.RealPath, reader.ReadCloser)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to open kernel module: %w", err)
+	}
+	defer rc.Close()
+
+	unionReader, err := unionreader.GetUnionReader(rc)
 	if err != nil {
 		return nil, nil, fmt.Errorf("unable to get union reader for file: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


syft/pkg/cataloger/kernel/parse_linux_kernel_module_file.go
```diff
@@ -39,6 +55,69 @@ func parseLinuxKernelModuleFile(ctx context.Context, _ file.Resolver, _ *generic
...
 }
 
+// openMaybeCompressedModule wraps src in a decompressing reader when the
+// path ends in a known compressed-module extension (.ko.xz / .ko.gz /
+// .ko.zst). For plain .ko files src is returned unchanged. The returned
+// io.ReadCloser owns both the decompressor (if any) and the original src.
+func openMaybeCompressedModule(path string, src io.ReadCloser) (io.ReadCloser, error) {
+	switch {
+	case strings.HasSuffix(path, ".ko.xz"):
+		zr, err := xz.NewReader(src)
```
kzantow, line 65: any compressed file reader should be handling compression bombs

PR #4852, opened 2026-05-02
binary/pe-cataloger: glob .bpl files alongside .dll and .exe

### LINKED ISSUE #4664: Add .bpl file support to the PE / DLL cataloger
What would you like to be added?
Please add **/*.bpl to the list of supported file extensions / globs for the PE (Portable Executable) cataloger.

Why is this needed?
Currently, when running a directory scan (syft dir:.), Syft ignores .bpl files. .bpl stands for Borland Package Library. These files are standard Windows PE/DLL files used extensively within the Delphi and C++Builder ecosystems.

Workaround / Proof of Concept:
If I manually rename a example.bpl file to example.dll before running the scan, Syft's PE cataloger successfully picks it up, analyzes the header, and extracts the metadata perfectly without any errors.

Because the underlying file format is exactly the same as a standard Windows .dll, adding the .bpl extension to the existing PE cataloger's file matching logic should be a very straightforward fix that provides out-of-the-box support for Delphi/C++Builder compiled libraries.

Additional context:

Renaming the files temporarily in a CI pipeline is a possible workaround, but native support would make scanning these environments much cleaner and less error-prone.

Thanks!

ChrisJr404:
Closes #4664.

`.bpl` (Borland Package Library) files are standard Windows PE binaries used by Delphi and C++Builder; they have the same structure as `.dll`, just a different filename convention. The reporter confirmed that renaming `example.bpl` to `example.dll` lets the existing PE cataloger pick them up unchanged — so all that was needed was the additional glob.

### Changes
- `syft/pkg/cataloger/binary/pe_package_cataloger.go`: add `**/*.bpl` to `WithParserByGlobs` and document why in a short doc-comment.

### Testing notes
Did not add a new image fixture — the issue's reporter already verified the parser side works for `.bpl` content. Happy to add one if you'd like (the existing `Test_PEPackageCataloger` builds image fixtures from Dockerfiles, so a `.bpl`-fixture image would slot in there).

### DIFF
--- a/syft/pkg/cataloger/binary/pe_package_cataloger.go
+++ b/syft/pkg/cataloger/binary/pe_package_cataloger.go
@@ -12,10 +12,14 @@ import (
 	"github.com/anchore/syft/syft/pkg/cataloger/internal/pe"
 )
 
-// NewPEPackageCataloger returns a cataloger that interprets packages from DLL and EXE files.
+// NewPEPackageCataloger returns a cataloger that interprets packages from PE
+// files. The default extensions covered are .dll, .exe, and .bpl. The latter
+// is the Borland Package Library used by Delphi and C++Builder; .bpl files
+// are standard Windows PE binaries with a different filename convention
+// (issue #4664).
 func NewPEPackageCataloger() pkg.Cataloger {
 	return generic.NewCataloger("pe-binary-package-cataloger").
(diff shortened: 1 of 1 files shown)


wagoodman:
Duplicate of #4688

PR #4853, opened 2026-05-02
dotnet: emit microsoft:.net CPE for unified .NET (5+)

ChrisJr404:
Closes #4738.

NVD's canonical CPE for the unified .NET platform is `microsoft:.net` (e.g. [CVE-2025-21171](https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=CVE-2025-21171) uses `cpe:2.3:a:microsoft:.net:9.0.0:*:*:*:*:*:*:*`). Syft was emitting `microsoft:dotnet` for `majorVersion >= 5`, which doesn't match anything in NVD and caused silent false-negative CVE matching for self-contained .NET runtime packages (Dependency-Track was the reporter's case; the same applies for any consumer that goes through the SBOM's CPE field).

### Changes
- `syft/pkg/cataloger/dotnet/package.go`: for major >= 5, emit two CPEs — `microsoft:.net:<major>.<minor>` (NVD-canonical, listed first so it takes precedence) and `microsoft:dotnet:<major>.<minor>` (legacy, kept for backwards compatibility with consumers that historically matched the old vendor:product). For major < 5 the existing `microsoft:dotnet_core` is unchanged.
- `syft/pkg/cataloger/dotnet/package_test.go`: extend `TestRuntimeCPEs` cases for 5.0/6.0/8.0/10.0 + the patch/assumed-minor cases to expect both CPEs in the new order.

### Verification
```
$ go test -count=1 -run TestRuntimeCPEs ./syft/pkg/cataloger/dotnet/
ok  	github.com/anchore/syft/syft/pkg/cataloger/dotnet	0.025s
```

### DIFF
--- a/syft/pkg/cataloger/dotnet/package.go
+++ b/syft/pkg/cataloger/dotnet/package.go
@@ -99,20 +99,46 @@ func runtimeCPEs(ver string) []cpe.CPE {
 		}
 	}
 
-	productName := "dotnet"
+	version := fmt.Sprintf("%d.%d", majorVersion, minorVersion)
+
 	if majorVersion < 5 {
-		productName = "dotnet_core"
+		// .NET Core 1.x – 3.x: NVD's canonical CPE is "microsoft:dotnet_core".
+		return []cpe.CPE{
+			{
+				Attributes: cpe.Attributes{
+					Part:    "a",
+					Vendor:  "microsoft",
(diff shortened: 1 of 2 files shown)


wagoodman:
This is a duplicate of https://github.com/anchore/syft/pull/4743

PR #4859, opened 2026-05-02
fix(javascript): preserve multiple versions of same package in yarn berry lockfile (#4691)

### LINKED ISSUE #4691: Yarn Berry lockfile parser incorrectly deduplicates packages with multiple resolutions

ChrisJr404:
## Description
Closes #4691.

The yarn v2 (berry) parser was deduplicating entries by package name only ([`packages[packageName] = ...`](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/javascript/parse_yarn_lock.go#L261)), so when a single package was resolved to more than one version — which is allowed and normal in Node when transitive dependencies require different ranges — only one version survived. The result was also non-deterministic across runs because the choice depended on the iteration order of the map the YAML decoder populated.

The v1 parser already does this correctly, and there's an existing `TestParseYarnLockWithDuplicates` for it.

## Fix
Switch the dedupe key to `name@version` so every distinct (name, version) tuple is preserved:

```go
packages[packageName+"@"+value.Version] = yarnPackage{...}
```

Yarn already aggregates entries that share a single resolution under a comma-separated key, so no false duplicates are added by this — the only entries that survive past the dedupe are those Yarn itself wrote as separate stanzas.

@wagoodman called this out in the issue:
> The v2 code is not doing this correctly and is instead deduping based on name only, not the parsed version. I think we'd want to have the actual package version (not a requirement like `^2.3.1`) in the captured tuple for deduping.

@calumleslie noted that further dedupe by resolution is probably unnecessary because Yarn already collapses requirements that share a resolution — this PR matches that observation.

## Tests
Adds `testdata/yarn-berry-dups/yarn.lock` with the synthetic two-version example from the issue body (async@0.9.2 and async@3.2.6) and a new `TestParseYarnBerryLockWithDuplicates` that mirrors the existing v1 `TestParseYarnLockWithDuplicates`.

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./syft/pkg/cataloger/javascript/ -run TestParseYarnBerryLockWithDuplicates` passes
- [x] `go test ./syft/pkg/cataloger/javascript/ -run 'Yarn|yarn'` — all yarn tests pass, no regressions on yarn-berry, yarn-berry-dev-deps, yarn-dups, or yarn-remote
- [x] Local repro of the issue: `mkdir /tmp/y && cd /tmp/y && cp /path/to/yarn-berry-dups/yarn.lock . && syft .` now lists both `async@0.9.2` and `async@3.2.6` deterministically across repeated runs.


wagoodman:
This is one of several a duplicate PRs for the same issue; we're accepting the original issues' author https://github.com/anchore/syft/pull/4838

calumleslie:
> @calumleslie noted that further dedupe by resolution is probably unnecessary because Yarn already collapses requirements that share a resolution — this PR matches that observation.

To be clear I asserted this but did not implement it this way as I didn't _verify_ that Yarn is guaranteed to do it, apologies if I didn't make that clear!

Issue #2820, opened 2024-04-29
Add `bun-lock-cataloger` & `bun-binary-cataloger` catalogers


dwisiswant0:
**What would you like to be added**:

Supporting Bun runtime & toolkit.

**Why is this needed**:

\-

**Additional context**:

References:

* https://bun.sh/docs/install/lockfile
* https://bun.sh/docs/bundler/executables

wagoodman:
This looks fun! Leaving behind some information for anyone that might be interested in working on this. 

## lockfile

The bun lockfile is a binary format:
```bash
# $ xxd bun.lockb | head
00000000: 2321 2f75 7372 2f62 696e 2f65 6e76 2062  #!/usr/bin/env b
00000010: 756e 0a62 756e 2d6c 6f63 6b66 696c 652d  un.bun-lockfile-
00000020: 666f 726d 6174 2d76 300a 0200 0000 c8ca  format-v0.......
00000030: bf6e 1c9b 607b da75 2feb c98e 4d95 098f  .n..`{.u/...M...
00000040: 498f 7478 0e53 a7ff cf9e f4b8 dd87 e085  I.tx.S..........
00000050: 0000 0000 0000 5900 0000 0000 0000 0800  ......Y.........
00000060: 0000 0000 0000 0800 0000 0000 0000 8000  ................
00000070: 0000 0000 0000 7558 0000 0000 0000 0000  ......uX........
00000080: 6275 6e2d 6578 0000 7a6f 6400 0000 0000  bun-ex..zod.....
00000090: 7669 7465 0000 0000 6673 6576 656e 7473  vite....fsevents
```

The code that parses this in zig can be found [here](https://github.com/oven-sh/bun/blob/main/src/install/lockfile.zig). From the [docs](https://bun.sh/docs/install/lockfile):

> It uses linear arrays for all data. [Packages](https://github.com/oven-sh/bun/blob/be03fc273a487ac402f19ad897778d74b6d72963/src/install/install.zig#L1825) are referenced by an auto-incrementing integer ID or a hash of the package name. Strings longer than 8 characters are de-duplicated. Prior to saving on disk, the lockfile is garbage-collected & made deterministic by walking the package tree and cloning the packages in dependency order.

## executable

I'm not seeing any evidence of unencoded packaging information yet in a simple test:
```bash
$ cat package.json
   1 {
   2   "name": "alexgoodman",
   3   "version": "1.23.445",
   4   "module": "index.ts",
   5   "type": "module",
[...]

PR #4625, opened 2026-02-16
Add support for Bun lockfile (bun.lock)

hnnynh:
## Description

This PR adds support for parsing Bun lockfiles (bun.lock) to Syft's JavaScript cataloger, enabling SBOM generation for projects using [Bun](https://bun.sh/) as their package manager.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (please discuss with the team first; Syft is 1.0 software and we won't accept breaking changes without going to 2.0)
- [ ] Documentation (updates the documentation)
- [ ] Chore (improve the developer experience, fix a test flake, etc, without changing the visible behavior of Syft)
- [ ] Performance (make Syft run faster or use less memory, without changing visible behavior much)

## Checklist

- [x] I have added unit tests that cover changed behavior
- [x] I have tested my code in common scenarios and confirmed there are no regressions
- [x] I have added comments to my code, particularly in hard-to-understand sections

## Issue references

Closes #4617

PR #4862, opened 2026-05-02
feat(javascript): add bun-lock-cataloger for bun.lock files (#2820, phase 1)

ChrisJr404:
Closes part of #2820 — adds the **first phase** of bun support: a cataloger for the **JSONC `bun.lock`** lockfile introduced in Bun 1.2 (late 2024). The legacy binary `bun.lockb` format remains an open follow-up.

## Why split the issue

When the issue was filed in May 2024, only `bun.lockb` (binary) existed and @wagoodman left a detailed note about reverse-engineering the format. Bun 1.2 then introduced a textual `bun.lock` (JSONC) which is what most projects on current Bun ship today. JSON parsing is much simpler than reversing the binary, so this PR delivers user-visible value first; `bun.lockb` parsing can land independently.

## What's added

- New `pkg.BunLockEntry` metadata type (mirrors `YarnLockEntry` / `PnpmLockEntry`).
- `parseBunLock` parser registered on `**/bun.lock` inside `javascript-lock-cataloger`.
- Two-pass JSONC preprocessor (strip comments, then strip trailing commas) — needed because Bun emits both, and `encoding/json` rejects either. Strings are preserved correctly.
- Custom `UnmarshalJSON` for the heterogeneous package tuple `[name@version, source, info, integrity]`.
- Dev-only classification via BFS reachability over `workspaces[].{dev,prod,peer,optional}Dependencies` — same approach as the yarn-lock cataloger.
- Deduplication by `name@version` so a package resolved at multiple tree paths (e.g. `cliui/string-width` and `string-width` both pointing at `string-width@4.2.3`) emits once.
- Capabilities YAML + json-schema regenerated; bumped `JSONSchemaVersion` to 16.1.4 (ADDITION per SchemaVer).

## Tests

| Test | What it covers |
|---|---|
| `TestParseBunLock_Simple` | Direct + transitive dep, integrity hash, relationships |
| `TestParseBunLock_ScopedAndPeer` | `@scope/foo` keys, `peerDependencies` treated as prod for reachability |
| `TestParseBunLock_DevDependencies` | `IncludeDevDependencies={true,false}`; verifies shared transitive (reachable from both prod & dev) is **not** filtered out |
| `TestParseBunLock_JSONCFeatures` | line/block comments and trailing commas all strip cleanly, including the awkward case of `,` followed by `// comment` |
| `TestSplitBunNameVersion` | `cowsay@1.6.0`, `@types/node@20.x`, `@scope/pkg@2.0.0-beta.1` |
| `TestStripJSONCExtras` | trailing commas in objects/arrays, line/block/multiline comments, comment-like content inside strings preserved |
| `Test_LockCataloger_Globs` (updated) | `**/bun.lock` is discovered alongside the other lockfiles |

Fixtures were generated by running real `bun install` against a `package.json` with cowsay/zod/typescript and trimmed by hand for clarity.

## Out of scope (Phase 2 — separate PR)

- `bun.lockb` (binary) parsing per @wagoodman's note on the issue.
- `bun build --compile` executable cataloger — wagoodman's experiment showed no clean way to extract package info from those binaries; deferring until the format embeds SBOM metadata.

## Test plan

- [x] `go test ./syft/pkg/cataloger/javascript/...` passes
- [x] `go test ./syft/pkg ./internal/packagemetadata/...` passes
- [x] `go generate ./internal/...` clean (capabilities + packagemetadata regenerated)
- [x] `SYFT_ENABLE_COMPLETENESS_TESTS=true go test -p 1 ./internal/capabilities/...` — no regressions; the only failure (`TestCatalogersHaveTestObservations` for `binary-classifier-cataloger` / `elf-binary-package-cataloger` / `php-interpreter-cataloger`) reproduces on `main` and is unrelated.
- [x] `go build ./...` clean
- [x] DCO signed


timols:
We would love this! Happy to help get it across the line too if needed

internal/packagemetadata/generated.go
```diff
@@ -7,68 +7,69 @@ import "github.com/anchore/syft/syft/pkg"
...
+ pkg.RubyGemspec{},
+ pkg.RustBinaryAuditEntry{},
+ pkg.RustCargoLockEntry{},
+ pkg.SnapEntry{},
+ pkg.SwiftPackageManagerResolvedEntry{},
+ pkg.SwiplPackEntry{},
+ pkg.TerraformLockProviderEntry{},
+ pkg.WordpressPluginEntry{},
+ pkg.YarnLockEntry{},
+}
```
timols, line 74: Was this formatting change intentional?

timols:
Hi @wagoodman @kzantow - any chance this could get some eyeballs?

izaakschroeder:
https://github.com/anchore/syft/pull/4625

ChrisJr404:
Yeah, #4625 landed this already, closing. Thanks @izaakschroeder for the pointer.

Issue #2210, opened 2023-10-10
Support resolving license titles to SPDX IDs


wagoodman:
Today we try and resolve input such as `apache2` to the official SPDX ID of `Apache-2.0`. It would be great to be able to go further for projects that use the license title (or similar), such as `The Apache Software License, Version 2.0` (as found in https://github.com/neo4j/license-maven-plugin/blob/3/pom.xml ) to also resolve to `Apache-2.0`.

Walnussbaer:
This is related to https://github.com/anchore/syft/issues/3590

kzantow:
It looks like the [licenses.json includes the name](https://github.com/anchore/syft/blob/99158be0baba90d9beff0e311e947d0f03fb1876/internal/spdxlicense/generate/testdata/licenses.json#L9), e.g.:
```
{
      "reference": "https://spdx.org/licenses/SCEA.html",
      "isDeprecatedLicenseId": false,
      "detailsUrl": "https://spdx.org/licenses/SCEA.json",
      "referenceNumber": 0,
      "name": "SCEA Shared Source License",
      "licenseId": "SCEA",
      "seeAlso": [
        "http://research.scea.com/scea_shared_source_license.html"
      ],
      "isOsiApproved": false
    },
```
We could at the very least capture the names, lowercased and space normalized and check license text to match exactly or with some confidence, for example there are [algorithms for text similarity](https://github.com/adrg/strutil).

This seems like a fairly straightforward enhancement for cases we don't have an exact SPDX ID, there are probably many locations we would need to update across Syft, though.

PR #4863, opened 2026-05-02
feat(debian): detect Format header + URL-based license lookup in copyright (first step of #4708)

ChrisJr404:
First step toward #4708 per @kzantow's livestream comment: *"add support for the `Format:` field and look up the URL against our list of known license URLs."*

## What's added

Two additive helpers in `syft/pkg/cataloger/debian/parse_copyright.go`:

| Function | What it does |
|---|---|
| `hasMachineReadableFormat(content)` | Recognizes the deb822 copyright format via the mandatory `Format:` header in the first stanza ([spec](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/)). Pure function, exposed for follow-up PRs to gate parsing strategy on. |
| `licenseIDsFromURLs(line)` | Extracts http(s) URLs from a line and resolves them against `spdxlicense.LicenseByURL`. References like `License: see http://www.apache.org/licenses/LICENSE-2.0` now surface a concrete SPDX ID even when the `License:` short-name field is missing or non-canonical. |

The URL lookup runs alongside the existing `License:` / `/usr/share/common-licenses/` / EULA-heading regexes — purely additive, never removes a finding.

## Why this scope

@tmuehlbacher-bnr's full proposal in the issue (gate machine-readable parsing on `Format:`, fall back to license classifier, deb822 stanza parser) is a substantial refactor that would need fixture-by-fixture renegotiation. @kzantow's response framed two concrete first steps:

1. *"add support for the Format: field"* — done; `hasMachineReadableFormat` exposes the detection so it can drive future behavior changes without re-walking the file.
2. *"look up the URL against our list of known license URLs"* — done; `licenseIDsFromURLs` plugs into the existing `spdxlicense.LicenseByURL` table.

This PR keeps the existing per-line regexes unchanged so no fixture expectations move, leaving room for the follow-up work that @tmuehlbacher-bnr outlined.

## Backward compatibility

All eight existing copyright fixtures produce **identical** results (`libc6`, `trilicense`, `liblzma5`, `libaudit-common`, `python`, `cuda`, `dev-kit`, `microsoft`).

The python fixture's already-known noisy captures (`#`, `Permission`, `This`, `see`) are intentionally untouched — addressing those means tightening `licensePattern` or gating it on `hasMachineReadableFormat`, both of which change observed behavior on real-world packages and belong in a follow-up.

## Tests

| Test | What it covers |
|---|---|
| `TestParseLicensesFromCopyright/testdata/copyright/format-header` | Machine-readable file with both License: short-names and embedded URLs |
| `TestParseLicensesFromCopyright/testdata/copyright/url-only-license` | Copyright with no License: short-name field at all — exercises pure URL detection |
| `TestHasMachineReadableFormat` (7 cases) | Format header in/out of first stanza, http/https schemes, no-scheme rejection, empty input |
| `TestLicenseIDsFromURLs` (5 cases) | Apache-2.0 / MIT seeAlso URLs, trailing-punctuation handling, non-license URLs filtered out |

## Out of scope (deferred to follow-ups)

- Refactoring per-line `License:` matching to skip non-machine-readable files (the python-fixture noise the issue mentions).
- Fuzzy-matching short license text against SPDX IDs (@kzantow's second-step suggestion + #2210).
- Fall-through to the license classifier when nothing else identifies the file.
- A real deb822 stanza parser as @tmuehlbacher-bnr proposed.

## Test plan

- [x] `go test ./syft/pkg/cataloger/debian/...` (non-docker subset) passes
- [x] `go vet ./syft/pkg/cataloger/debian/...` clean
- [x] All 8 prior fixtures produce unchanged results
- [x] DCO signed

### DIFF
--- a/syft/pkg/cataloger/debian/parse_copyright.go
+++ b/syft/pkg/cataloger/debian/parse_copyright.go
@@ -10,6 +10,7 @@ import (
 	"github.com/scylladb/go-set/strset"
 
 	"github.com/anchore/syft/internal"
+	"github.com/anchore/syft/internal/spdxlicense"
 )
 
 // For more information see: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-syntax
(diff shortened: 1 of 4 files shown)


ChrisJr404:
Rebased onto main — this collided with the copyright parser refactor that landed since I opened it (the heading-detection state machine). Reworked the Format-header detection and URL license lookup to sit on top of the new structure rather than the old first-sentence pattern. Tests updated to match.

PR #4871, opened 2026-05-04
feat: align GOMAXPROCS with the configured parallelism (#3924)

### LINKED ISSUE #3924: Set GOMAXPROCS when parallelism is configured
In order to make behavior more predictable, if a user configures parallelism, it might be a good idea to set [GOMAXPROCS](https://pkg.go.dev/runtime#GOMAXPROCS) to limit the number of system threads.

**Why is this needed**:
More predictable performance characteristics.

**Additional context**:
This stemmed from a [discussion on discourse](https://anchorecommunity.discourse.group/t/limiting-the-number-of-threads-used/442).

ChrisJr404:
## Summary

Closes #3924 (\"Set GOMAXPROCS when parallelism is configured\", filed by @kzantow).

When a caller sets a positive \`Parallelism\` (CLI \`--parallelism N\`, \`syft.WithParallelism(N)\`, or yaml \`parallelism: N\`), the cataloger pool caps to \`N\` workers, but the Go runtime is still free to spin up \`runtime.NumCPU()\` OS threads. That makes the on-host behaviour less predictable than the configured cap — exactly the surprise the linked discourse thread describes.

## Change

\`setContextExecutors\` now calls a new \`maybeReduceGOMAXPROCS()\` helper that lowers \`runtime.GOMAXPROCS\` to match the requested parallelism, gated by two safety conditions:

- The user gave a positive value greater than 1 (so 0/default, 1/serial, and negative/unbounded are left alone).
- The requested value is strictly lower than the current \`GOMAXPROCS\` (we never raise it past whatever the runtime already chose for the host).

The existing 0/1/-99 special cases keep their meaning unchanged.

## Test

\`TestMaybeReduceGOMAXPROCS\` — 7 sub-tests, network-free:

- default (0) leaves GOMAXPROCS alone
- serial (1) leaves GOMAXPROCS alone
- unbounded (-1) leaves GOMAXPROCS alone
- test sentinel (-99) leaves GOMAXPROCS alone
- explicit lower (8 → 2) reduces GOMAXPROCS to 2
- equal (4 → 4) is a no-op
- explicit higher (2 → 16) does not raise GOMAXPROCS

```
$ go test ./syft -run TestMaybeReduceGOMAXPROCS -v
--- PASS: TestMaybeReduceGOMAXPROCS (0.00s)
PASS
```

\`go build ./...\` is clean. The test restores the original GOMAXPROCS in \`t.Cleanup\` so the rest of the suite is unaffected.

## Why not a third-party module like uber-go/automaxprocs

automaxprocs derives a value from the cgroup CPU quota, which solves a related but different problem (containerized hosts that don't expose a sensible NumCPU). The user-facing semantic the issue calls out — "I configured parallelism to 4, please don't run more than 4 OS threads either" — is a strict subset that the runtime stdlib already covers, so this PR keeps it dependency-free. Happy to revisit if you want the cgroup-aware path layered on top.


Dashtid:
A question on the placement: `setContextExecutors` is called from `syft.CreateSBOM`, the library entrypoint used by embedding consumers (Grype being the obvious one — `grype/pkg/syft_provider.go` calls it directly). Lowering process-wide `runtime.GOMAXPROCS` from inside that path means a host process calling `syft.CreateSBOM` with `WithParallelism(2)` would permanently shrink `GOMAXPROCS` to 2 for the rest of the process — including whatever vuln matching or unrelated work the host was about to do next. The `parallelism < current` guard avoids surprises in the other direction, but the change isn't restored when the call returns.

#3924 doesn't discuss library-vs-CLI placement, so flagging one alternative: moving the call into `cmd/syft` would scope this to CLI usage and leave library consumers' runtime untouched. Or is the assumption that an embedding caller wanting `WithParallelism(N)` also wants the whole process tuned to N?

ChrisJr404:
Good catch, you're right that setContextExecutors is the wrong place for this since grype and other embedders go through CreateSBOM and would get their runtime shrunk out from under them with no restore. Moving it into the syft CLI (Catalog.PostLoad) so it only applies to the command-line tool and library callers are left alone.

Dashtid:
After the rebase the PR's net diff is two files, both under `cmd/syft/internal/options` — the library is not touched at all now, rather than touched and reverted. That is the outcome I was hoping for: `syft.CreateSBOM` never mutates process-global state, so an embedder gets whatever `GOMAXPROCS` its own process decided regardless of how many SBOMs it builds.

The six table cases in `TestCatalog_PostLoad_GOMAXPROCS` also pin the two properties I would otherwise have asked about — "parallelism above current GOMAXPROCS does not raise it" covers the one-directionality, and the `0` / `1` / `-1` cases cover the documented sentinel values (`0` = `NumCPU * 4`, negative = unbounded), so the default path provably never touches the runtime. Worth having those pinned rather than implied.

One open question on the new home. `PostLoad` is otherwise config normalization and validation, and this gives it a process-global side effect — `Catalog` is embedded by both `scan` and `attest`, and a reader looking for where `GOMAXPROCS` moves is unlikely to look in a `PostLoad`. Is `PostLoad` the intended home, or would the scan command's run path be a more obvious one?

If it stays, one small thing worth guarding: the test drives real `runtime.GOMAXPROCS` and restores it in `t.Cleanup`, which is safe today only because nothing in this package's tests calls `t.Parallel()`. A later `t.Parallel()` added anywhere in the package — including to an unrelated test — would race against that mutation, and the failure would be confusing. A comment on the test noting it must stay serial would probably save someone an afternoon.

PR #4878, opened 2026-05-04
fix: only parse debian/copyright as machine-readable when Format: header present (#4708)

### LINKED ISSUE #4708: fix parsing of debian/copyright files

tjhub1983:
## Summary

Fixes [#4708](https://github.com/anchore/syft/issues/4708).

Previously, `parseLicensesFromCopyright` applied machine-readable parsing to ALL copyright files, causing false positives from non-machine-readable content. Now it first checks for `Format:` as the first non-empty line. Only files declaring machine-readable format use full parsing.

**Before:** All copyright files parsed with machine-readable logic → false positives
**After:** Only `Format:`-declared files use machine-readable parsing; others fall back to classifier

## Changes

- `parse_copyright.go`: Added `formatPattern` and Format-header detection at file start
- `parse_copyright_test.go`: Updated expectations for non-Format fixtures
- `testdata/copyright/non-machine-readable`: New test fixture without Format: header

## Test

```bash
$ go test ./syft/pkg/cataloger/debian/... -run TestParseLicensesFromCopyright -v
PASS
```

### DIFF
--- a/syft/pkg/cataloger/debian/parse_copyright.go
+++ b/syft/pkg/cataloger/debian/parse_copyright.go
@@ -18,21 +18,44 @@ var (
 	licensePattern                 = regexp.MustCompile(`^License: (?P<license>\S*)`)
 	commonLicensePathPattern       = regexp.MustCompile(`/usr/share/common-licenses/(?P<license>[0-9A-Za-z_.\-]+)`)
 	licenseAgreementHeadingPattern = regexp.MustCompile(`(?i)^\s*(?P<license>LICENSE AGREEMENT(?: FOR .+?)?)\s*$`)
+	formatPattern                  = regexp.MustCompile(`^Format:\s*`)
 )
+
+// parseLicensesFromCopyright extracts license information from Debian copyright files.
+//
+// Machine-readable format detection:
+//   - Files with "Format:" as the first non-empty line are parsed using the full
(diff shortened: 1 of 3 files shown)


tjhub1983:
Hi @anchore/syft-maintainers! 

Just wanted to follow up on these PRs. I noticed the CI workflow runs are showing `action_required` status. Could you please approve the runs so they can complete?

For reference:
- These are fork PRs from tjhub1983/syft
- The CI was triggered but the jobs are waiting for maintainer approval (common for fork PRs with runs-on.com runners)
- All DCO checks are passing

Thank you!

wagoodman:
This is a duplicate of https://github.com/anchore/syft/pull/4754

PR #4880, opened 2026-05-04
fix: skip permission-denied paths instead of crashing on directory walk (#3258)

### LINKED ISSUE #3258: Excluded paths are still scanned and cause syft to crash.
**What happened**: I excluded paths  --exclude ./boot --exclude ./web  I had verbose set to 2
When ran you seed WARN unable to access path="/web" lstat /web no such file or directory 
This is really a permission issue., but it should have never checked.

The bigger issues was the error:
unable to get file resolver: unable to create directory resolver: unable to index filesystem path="/boot/efi/EFI/readhat/grub.cfg": lstat /boot/dfi/DFIL permission denied
syft exits.
This makes the tool useless.

**What you expected to happen**: The paths excluded would not be scanned.

**Steps to reproduce the issue**: Use a user that does not have permission to read /boot

**Environment**: 
- Output of `syft version`: 1.11.0
- OS (e.g: `cat /etc/os-release` or similar): Redhat 7.9 and 8.4

tjhub1983:
## Summary

Fixes [#3258](https://github.com/anchore/syft/issues/3258).

When scanning a directory with excluded paths, syft would crash if it encountered permission-denied directories (e.g., `/boot/grub2` owned by root) during the ancestor path walk, even when those paths were excluded.

**Before:** `unable to index ancestor path="/boot/grub2": permission denied` → syft exits with error
**After:** Permission-denied paths are skipped and logged, scan continues

The fix adds `r.isFileAccessErr(targetPath, err)` check at line 218 of `directory_indexer.go`. When `os.Lstat` fails with a file access error, the ancestor walk continues instead of returning an error. This matches the behavior of `filepath.Walk` which skips permission errors.

## Changes

- `syft/internal/fileresolver/directory_indexer.go`: Skip file access errors in `indexBranch` ancestor walk instead of returning an error

## Test

```bash
$ sudo -u nobody syft / --exclude ./boot --exclude ./web
# Previously: crash on permission denied
# Now: logs warning and continues
```

## Related

This was discussed in syft#3258 with the suggestion that excluded paths should not be accessed at all. While the ideal long-term fix would prevent access to excluded paths entirely, this change provides immediate relief by making permission errors non-fatal.

### DIFF
--- a/syft/internal/fileresolver/directory_indexer.go
+++ b/syft/internal/fileresolver/directory_indexer.go
@@ -216,6 +216,12 @@ func (r *directoryIndexer) indexBranch(root string, stager *progress.AtomicStage
 		lstat, err := os.Lstat(targetPath)
 		newRoot, err := r.indexPath(targetPath, lstat, err)
 		if err != nil && !errors.Is(err, ErrSkipPath) && !errors.Is(err, fs.SkipDir) {
+			// file access errors (e.g. permission denied on restricted directories like /boot/grub2)
+			// should not stop the walk -- skip and continue, matching filepath.Walk behavior.
+			// Only fail for truly unexpected errors (ErrSkipPath and fs.SkipDir are expected/safe).
+			if r.isFileAccessErr(targetPath, err) {
+				return nil, nil
+			}
 			return nil, fmt.Errorf("unable to index ancestor path=%q: %w", targetPath, err)
 		}
 		if newRoot != "" {


tjhub1983:
Hi @anchore/syft-maintainers!

This PR fixes [#3258](https://github.com/anchore/syft/issues/3258) — a crash when scanning directories with excluded paths that contain permission-denied directories (e.g., `/boot/grub2`).

The CI workflow runs are showing `action_required` status, which likely means the custom runs-on.com runner needs maintainer approval for fork PRs.

Could you please approve the CI runs for this PR? Happy to make any adjustments needed.

Thank you!

syft/internal/fileresolver/directory_indexer.go
```diff
@@ -216,6 +216,12 @@ func (r *directoryIndexer) indexBranch(root string, stager *progress.AtomicStage
 		lstat, err := os.Lstat(targetPath)
 		newRoot, err := r.indexPath(targetPath, lstat, err)
 		if err != nil && !errors.Is(err, ErrSkipPath) && !errors.Is(err, fs.SkipDir) {
+			// file access errors (e.g. permission denied on restricted directories like /boot/grub2)
+			// should not stop the walk -- skip and continue, matching filepath.Walk behavior.
+			// Only fail for truly unexpected errors (ErrSkipPath and fs.SkipDir are expected/safe).
+			if r.isFileAccessErr(targetPath, err) {
```
kzantow, line 222: Is there some way to add a test for this that isn't too convoluted to set up?

tjhub1983, line 222: @kzantow Thanks for the review!

Regarding the test: a reliable permission-denied test typically requires OS-level permission manipulation which can be flaky across platforms (especially Windows). The existing unit tests cover the core logic paths (`walkWithExclusion`, `walkDir`), but testing the permission-denied scenario at the integration level requires either:

1. Using `os.Chmod` to make a directory unreadable (platform-dependent, may not work reliably on Windows)
2. A mock filesystem approach

I'm happy to add a test using approach #1 if you think it's worth the added complexity, or handle it as a follow-up issue. Let me know how you'd like to proceed.

tjhub1983:
Hi @anchore/syft-maintainers!

All CI checks have passed:
- Unit tests ✅
- Integration tests ✅
- CLI tests (Linux) ✅
- Acceptance tests (Linux/Mac) ✅
- Static analysis ✅
- Build snapshot artifacts ✅

Is there anything else needed to move this forward? This PR fixes the crash when scanning directories with excluded paths that contain permission-denied directories (e.g., /boot/grub2 on Linux).

tjhub1983:
Hi maintainers! All CI checks have passed. Is there anything blocking this from being merged? Happy to address any feedback!

tjhub1983:
Just checking in again - is there anything else needed to move this forward? All CI checks are green and kzantow reviewed previously. Happy to make any adjustments! @anchore/syft-maintainers

tjhub1983:
Just checking - is there anything needed from my side to move this forward?

tjhub1983:
ping @kzantow - PR #4880 has CI passing and your comments. Could you please review/approve so we can get this merged? Thanks!

tjhub1983:
?? Hi maintainers! Just checking in on this PR. Is there anything else needed from my side to move forward? Thanks!

tjhub1983:
?? Hi maintainers! Just checking in on this PR. Is there anything else needed? Thanks for your time!

syft/internal/fileresolver/directory_indexer.go
kzantow: This needs to have a test. The test can create a filesystem with this scenario.

spiffcs:
@tjhub1983 it looks like you still have the answer from @kzantow that says:
>  This needs to have a test. The test can create a filesystem with this scenario.

I apologize that we didn't respond to the many pings, but I think it was clear that we would not merge this work without tests being added.

I wrote a test for your branch just now and found a good reason why @kzantow was correct in holding off here.

Err Path:
`indexPath(/etc/grub2.cfg)` → `symlink` → `new root = /boot/efi/EFI/redhat/grub.cfg`

Then os.Stat fails with EACCES, not `ErrNotExist`, so it is kept as a root `directory_indexer.go:431`
`indexAllRoots` queues it → `indexTree(that root)`
 
`isRealPath()` → `EvalSymlinks(parent)` → `stat /boot/efi/EFI: permission denied`
 
This get's us `return nil, err` which is fatal, and aborts the whole scan.

The original fix proposed on this branch would miss the above case.

Worth noting that this is now fixed in main:
https://github.com/anchore/syft/blob/main/syft/internal/fileresolver/directory_indexer.go#L127-L138

Test that landed with it:
https://github.com/anchore/syft/blob/bf82010f3c38508e30dd01ed8053f158513c0c9d/syft/internal/fileresolver/directory_indexer_test.go#L289-L323

Closing this for now since it has been fixed upstream. Thank you for the contribution.

PR #4890, opened 2026-05-05
fix(binary): support mongodb versions with v prefix and db version string (#4609)

### LINKED ISSUE #4609: Support mongo binary various versions
**What happened**:

Current matcher can not detect some pattern
- rc version
- 8.x (8.0.17-noble is detected)
- 7.x (7.0.28-jammy, 7.0.23-jammy, 7.0.22-jammy, 7.0.15, 7.0.0-jammy are detected)
- 3.x.x (3.7.x, 3.5.7, 3.5.8, 3.5.9 are detected)

```
$ syft -q mongo:8.2.4 --override-default-catalogers binary-classifier-cataloger
NAME        VERSION  TYPE
bash        5.2.21   binary
gzip        1.12     binary
jq          1.7      binary
openssl     3.0.13   binary
util-linux  2.39.3   binary

$ docker run -it --rm mongo:8.2.4 mongod --version
db version v8.2.4
Build Info: {
    "version": "8.2.4",
    "gitVersion": "60692d74054a0bc73f181179e747815825de38ae",
    "openSSLVersion": "OpenSSL 3.0.13 30 Jan 2024",
    "modules": [],
    "allocator": "tcmalloc-google",
    "environment": {
        "distmod": "ubuntu2404",
        "distarch": "x86_64",
        "target_arch": "x86_64"
    }
}
```

If mongodb is installed manually, syft will not detect these mongo

tjhub1983:
## Summary

Fixes [#4609](https://github.com/anchore/syft/issues/4609).

**Problem**: The `mongodb-binary` classifier misses mongodb versions where the version string is stored with a `v` prefix (e.g., `db version v8.2.4` in the mongo:8.2.4 Docker image) or as bare `vX.Y.Z` format. The existing patterns only match null-terminated `X.Y.Z` followed by specific strings like `tcmalloc`, `heap_size`, or `cppdefines`.

**Before**: `syft -q mongo:8.2.4` → mongodb not detected
**After**: `syft -q mongo:8.2.4` → `mongodb 8.2.4`

**Fix**: Added two new version matchers:
1. `(?m)db version v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)` — matches `db version v8.2.4` format
2. `(?P<version>v[0-9]+\.[0-9]+\.[0-9]+)` — matches bare `vX.Y.Z` format

## Changes

- `syft/pkg/cataloger/binary/classifiers.go`: Added v-prefix version patterns to mongodb-binary classifier

## Test

```bash
$ docker run -it --rm mongo:8.2.4 mongod --version
db version v8.2.4
$ syft -q mongo:8.2.4 | grep mongo
mongodb    8.2.4                              binary
```

### DIFF
--- a/syft/pkg/cataloger/binary/classifiers.go
+++ b/syft/pkg/cataloger/binary/classifiers.go
@@ -910,7 +910,7 @@ func DefaultClassifiers() []binutils.Classifier {
 				// e.g mongo:8.2.4 docker image stores "db version v8.2.4" in mongod binary
 				m.FileContentsVersionMatcher(`(?m)db version v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`),
 				// mongodb with bare v prefix (e.g. "v8.0.0" stored in binary)
-				m.FileContentsVersionMatcher(`(?P<version>v[0-9]+\.[0-9]+\.[0-9]+)`),
+				m.FileContentsVersionMatcher(`(?P<version>v[0-9]+\.[0-9]+\.[0-9]+)(?![0-9])`),
 			),
 			Package: "mongodb",
 			PURL:    mustPURL("pkg:generic/mongodb@version"),
--- a/syft/pkg/cataloger/binary/classifiers.go
+++ b/syft/pkg/cataloger/binary/classifiers.go
@@ -897,20 +897,18 @@ func DefaultClassifiers() []binutils.Classifier {
 			Class:    "mongodb-binary",
 			FileGlob: "**/mongod",
 			EvidenceMatcher: binutils.MatchAny(
-				// mongodb 4.x, 5.x, 6.x: ber followed by tcmalloc
+				// mongodb 4.x, 5.x, 6.x: ver followed by tcmalloc
 				// e.g 6.0.27[NUL]tcmalloc
 				m.FileContentsVersionMatcher(`(?P<version>[0-9]+\.[0-9]+\.[0-9]+)\x00tcmalloc`),
 				// mongodb 7.x: ver followed by "heap_size"
 				// e.g 7.0.28[NUL]heap_size
 				m.FileContentsVersionMatcher(`(?P<version>[0-9]+\.[0-9]+\.[0-9]+)\x00+heap_size`),
-				// mongodb 8.x: ber followed by "cppdefines"
+				// mongodb 8.x: ver followed by "cppdefines"
 				// e.g 8.0.17[NUL]cppdefines
 				m.FileContentsVersionMatcher(`(?P<version>[0-9]+\.[0-9]+\.[0-9]+)\x00+cppdefines`),
 				// mongodb 8.x with v prefix: "db version v8.2.4"
 				// e.g mongo:8.2.4 docker image stores "db version v8.2.4" in mongod binary
 				m.FileContentsVersionMatcher(`(?m)db version v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`),
-				// mongodb with bare v prefix (e.g. "v8.0.0" stored in binary)
-				m.FileContentsVersionMatcher(`(?P<version>v[0-9]+\.[0-9]+\.[0-9]+)(?![0-9])`),
 			),
 			Package: "mongodb",
 			PURL:    mustPURL("pkg:generic/mongodb@version"),
--- a/syft/pkg/cataloger/binary/classifiers.go
+++ b/syft/pkg/cataloger/binary/classifiers.go
@@ -906,6 +906,11 @@ func DefaultClassifiers() []binutils.Classifier {
 				// mongodb 8.x: ber followed by "cppdefines"
 				// e.g 8.0.17[NUL]cppdefines
 				m.FileContentsVersionMatcher(`(?P<version>[0-9]+\.[0-9]+\.[0-9]+)\x00+cppdefines`),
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/binary/classifiers.go
```diff
@@ -906,6 +906,11 @@ func DefaultClassifiers() []binutils.Classifier {
 				// mongodb 8.x: ber followed by "cppdefines"
 				// e.g 8.0.17[NUL]cppdefines
 				m.FileContentsVersionMatcher(`(?P<version>[0-9]+\.[0-9]+\.[0-9]+)\x00+cppdefines`),
+				// mongodb 8.x with v prefix: "db version v8.2.4"
+				// e.g mongo:8.2.4 docker image stores "db version v8.2.4" in mongod binary
+				m.FileContentsVersionMatcher(`(?m)db version v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`),
```
kzantow, line 911: There should be test(s) added verifying any changes we make here. Also, it looks like this pattern is unnecessary

tjhub1983:
@kzantow Thanks for the review! Regarding the bare `vX.Y.Z` pattern:

I've added a negative lookahead `(?![0-9])` to prevent double-matching — `v8.2.4` inside `"db version v8.2.4"` is now only matched by the first pattern. The bare pattern with the lookahead now only matches binaries that store the version without the "db version" prefix.

Questions:
1. Should I keep both patterns (defensive — handles edge cases), or remove the bare pattern since all mongodb binaries use the `"db version vX.Y.Z"` format?
2. Regarding test coverage — I've added test fixtures for both `mongodb/8.2.4` and `mongodb/v8.0.0`. Are there specific edge cases you'd like to see tested?

Happy to iterate based on your feedback.

witchcraze:
@tjhub1983 

Thank you for your PR.
As I wrote in #4609, Current matcher can not detect some pattern
- rc version
- 8.x (8.0.17-noble is detected)
- 7.x (7.0.28-jammy, 7.0.23-jammy, 7.0.22-jammy, 7.0.15, 7.0.0-jammy are detected)
- 3.x.x (3.7.x, 3.5.7, 3.5.8, 3.5.9 are detected)

Could you please add tests to cover these cases ?
https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/binary/README.md

kzantow:
> I've added a negative lookahead `(?![0-9])` to prevent double-matching — `v8.2.4` inside `"db version v8.2.4"` is now only matched by the first pattern. The bare pattern with the lookahead now only matches binaries that store the version without the "db version" prefix.

In one case the pattern is capturing the `v` in the other case it isn't. In the case there's `"db version"` it would still match the `v#.#.#`, so it's a redundant pattern with the text, isn't it? I agree with @witchcraze about looking at the cases mentioned in the issue and getting all of those covered.

Please also see the [README](https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/binary/README.md) @witchcraze mentioned above how to add tests.

witchcraze:
And I noticed that the test section you showed was very strange.
mongodb should be shown as deb package by default option.
```
$ syft -q mongo:8.2.4 | grep mongo
github.com/mongodb/mongo-tools                              UNKNOWN                             go-module  (+7 duplicates)
go.mongodb.org/mongo-driver                                 v1.17.3                             go-module  (+7 duplicates)
mongodb-database-tools                                      100.14.0                            deb
mongodb-mongosh                                             2.6.0                               deb
mongodb-org                                                 8.2.4                               deb
mongodb-org-database                                        8.2.4                               deb
mongodb-org-database-tools-extra                            8.2.4                               deb
mongodb-org-mongos                                          8.2.4                               deb
mongodb-org-server                                          8.2.4                               deb
mongodb-org-shell                                           8.2.4                               deb
mongodb-org-tools                                           8.2.4                               deb
```

tjhub1983:
Hi @kzantow and @witchcraze! Thank you for the reviews.

Regarding the bare `vX.Y.Z` pattern — you're right, it's redundant. The `"db version vX.Y.Z"` pattern already captures the version with the v prefix, so I've removed the bare pattern.

Changes:
- Removed redundant bare pattern `v[0-9]+\.[0-9]+\.[0-9]+`
- Removed the corresponding test case for `v8.0.0`

Regarding @witchcraze's comment about mongodb showing as deb package — this is expected behavior. The binary cataloger detects `mongod` binary and extracts version info, but the actual mongodb packages in the docker image are discovered by the deb cataloger and show as `deb` type. The binary cataloger fix ensures version detection works correctly when `mongod` binary is scanned directly.

Let me know if you have any other feedback!

witchcraze:
> Let me know if you have any other feedback!

From my previous check, MongoDB has no standardized matching pattern. 
Could you please verify if the updated Syft can correctly detect the mongod binary across different minor versions like https://github.com/anchore/syft/pull/4857 ?

> this is expected behavior. The binary cataloger detects mongod binary and extracts version info, but the actual mongodb packages in the docker image are discovered by the deb cataloger and show as deb type. 

If that is the case, why does your original post show the following output?
```
$ syft -q mongo:8.2.4 | grep mongo
mongodb    8.2.4                              binary
```
To be honest, I’m concerned that this might be generated text rather than actual command output. I believe the maintainers will not approve a PR without concrete and consistent evidence.

tjhub1983:
Closed - superseded by #4906 with proper DCO signature.

PR #4905, opened 2026-05-09
fix: add config option to exclude language packages with file ownership overlap

### LINKED ISSUE #4760: OS package (deb) components duplicated as pypi components
**What happened**:

Running `syft` on a root filesystem directory results in duplicated components (one `deb`, one `pypi`) for all Python packages installed via Ubuntu packages.

Related to #931 - I would want the `pypi` components that have an equivalent `deb` component to be filtered out. When #931 was closed the topic seemed to have moved from the initial report of `rpm` and `pypi` to just covering binary packages. I've opened a new issue to avoid pinging everyone in there.

I understand that I could de-select the python cataloger but in this case we do have some (far fewer) Python packages that are not available as Debians.

This is an issue because the `pypi` component metadata does not capture that in the Debian/Ubuntu ecosystem some backported fixes have been applied, and so once this CycloneDX SBOM is fed into `grype` some false positives will result.

This is run on an Ubuntu root file system extracted to a tmp location to produce a CycloneDX JSON SBOM.

```shell
export SYFT_FORMAT_CYCLONEDX_JSON_PRETTY=true
export SYFT_LICENSE_INCLUDE_UNKNOWN_LICENSE_CONTENT=true
export SYFT_ENRICH=all
export SYFT_PACKAGE_SEARCH_UNINDEXED_ARCHIVES=false
export SYFT_PACKAGE_SEARCH_INDEXED_ARCHIVES=true
export SYFT_LINUX_KERNEL_CATALOG_MODULES=true
export SYFT_FILE_METADATA_SELECTION=none
syft scan dir:${tmpdir} \
  --source-name="something" \
  --override-default-catalogers=directory,image \
  --select-catalogers=-file \
  --output=cyclonedx-json=${outputsbom}
```

For example, an older Django component that `grype` flags as having some vulnerabilities that are not actually present (as well as potentially eliding some license metadata).

```json
    {
      "bom-ref": "pkg:pypi/django@3.2.12?package-id=cc23182b55e9fbf5",
      "type": "library",
      "author": "Django Software Foundation <foundation@djangoproject.com>",
      "name": "django",
      "version": "3.2.12",
      "licenses": [
        {
          "license": {
            "id": "BSD-3-Clause"
          }
        }
      ],
      "cpe": "cpe:2.3:a:django_software_foundation_project:python-django:3.2.12:*:*:*:*:*:*:*",
      "purl": "pkg:pypi/django@3.2.12",
      "properties": [
        {
          "name": "syft:package:foundBy",
          "value": "python-installed-package-cataloger"
        },
        {
          "name": "syft:package:language",
          "value": "python"
        },
        {
          "name": "syft:package:type",
          "value": "python"
        },
        {
          "name": "syft:package:metadataType",
          "value": "python-package"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundation_project:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundationproject:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundationproject:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundation_project:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundation:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundation:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundationproject:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django_software_foundation:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundation_project:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundation_project:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundationproject:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundationproject:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python-django:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python-django:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python_django:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python_django:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundation_project:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundation:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundation:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundationproject:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python-django:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python:python-django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python:python_django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python_django:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:foundation:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:django:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python:django:3.2.12:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:location:0:path",
          "value": "/usr/lib/python3/dist-packages/Django-3.2.12.egg-info/PKG-INFO"
        },
        {
          "name": "syft:location:1:path",
          "value": "/usr/lib/python3/dist-packages/Django-3.2.12.egg-info/top_level.txt"
        }
      ]
    },
    {
      "bom-ref": "pkg:deb/ubuntu/python3-django@2%3A3.2.12-2ubuntu1.25?arch=all&distro=ubuntu-22.04&package-id=7898b761f06b3f22&upstream=python-django",
      "type": "library",
      "publisher": "Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>",
      "name": "python3-django",
      "version": "2:3.2.12-2ubuntu1.25",
      "licenses": [
        {
          "license": {
            "id": "Apache-2.0"
          }
        },
        {
          "license": {
            "id": "BSD-2-Clause"
          }
        },
        {
          "license": {
            "id": "BSD-3-Clause"
          }
        },
        {
          "license": {
            "name": "Expat"
          }
        },
        {
          "license": {
            "name": "PSF-baseconv"
          }
        }
      ],
      "cpe": "cpe:2.3:a:python3-django:python3-django:2\\:3.2.12-2ubuntu1.25:*:*:*:*:*:*:*",
      "purl": "pkg:deb/ubuntu/python3-django@2%3A3.2.12-2ubuntu1.25?arch=all&distro=ubuntu-22.04&upstream=python-django",
      "properties": [
        {
          "name": "syft:package:foundBy",
          "value": "dpkg-db-cataloger"
        },
        {
          "name": "syft:package:type",
          "value": "deb"
        },
        {
          "name": "syft:package:metadataType",
          "value": "dpkg-db-entry"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python3-django:python3_django:2\\:3.2.12-2ubuntu1.25:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python3_django:python3-django:2\\:3.2.12-2ubuntu1.25:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python3_django:python3_django:2\\:3.2.12-2ubuntu1.25:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python3:python3-django:2\\:3.2.12-2ubuntu1.25:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:cpe23",
          "value": "cpe:2.3:a:python3:python3_django:2\\:3.2.12-2ubuntu1.25:*:*:*:*:*:*:*"
        },
        {
          "name": "syft:location:0:path",
          "value": "/var/lib/dpkg/status"
        },
        {
          "name": "syft:location:1:path",
          "value": "/usr/share/doc/python3-django/copyright"
        },
        {
          "name": "syft:location:2:path",
          "value": "/var/lib/dpkg/info/python3-django.md5sums"
        },
        {
          "name": "syft:location:3:path",
          "value": "/var/lib/dpkg/info/python3-django.list"
        },
        {
          "name": "syft:location:4:path",
          "value": "/var/lib/dpkg/info/python3-django.postinst"
        },
        {
          "name": "syft:location:5:path",
          "value": "/var/lib/dpkg/info/python3-django.postrm"
        },
        {
          "name": "syft:location:6:path",
          "value": "/var/lib/dpkg/info/python3-django.preinst"
        },
        {
          "name": "syft:location:7:path",
          "value": "/var/lib/dpkg/info/python3-django.prerm"
        },
        {
          "name": "syft:metadata:installedSize",
          "value": "24133"
        },
        {
          "name": "syft:metadata:source",
          "value": "python-django"
        }
      ]
    },
```

I was looking at addressing this in my own post-processing steps for SBOMs, but it would be neater if these components never appeared.

It would be interesting to know if the Syft-style SBOMs have more metadata to allow `grype` to filter these components out, however we are using the CycloneDX standard in general and have to add a lot of things to the SBOM as it is so splitting to use another type of SBOM as well would not be practical.

**Environment**:
- Output of `syft version`: `1.42.0`
- OS (e.g: `cat /etc/os-release` or similar): Ubuntu 22.04

Thanks for any help.

kimjune01:
Closes #4760.

## Cause

When OS packages (deb, apk, rpm) install language packages (Python, Ruby, npm, etc.) via system package managers, syft catalogs both the OS package and the language package. The file ownership overlap relationship exists between them, but there was no mechanism to deduplicate.

The existing `exclude-binary-overlap-by-ownership` option handles binary packages, but an equivalent for language packages was missing.

## Fix

Adds a `relationships.exclude-language-packages-with-file-ownership-overlap` config option (named after the library field in `RelationshipsConfig`) that removes language packages when they have a file ownership overlap relationship with an OS package. This mirrors the existing binary exclusion logic. The option defaults to `false` to avoid changing existing behavior.

The set of excludable package types is derived at runtime from syft's own cataloger capability data: every type emitted by a `language`-tagged cataloger is a candidate, minus OS types and an explicit list of binary-extracted types (go-module, rust-crate, dotnet, graalvm-native-image, java-archive) whose packages are components pulled out of a single OS-owned binary or archive. A completeness test derives the expected set independently from the capability YAML files, so a new language cataloger fails CI until it is classified.

The implementation follows the same pattern as `ExcludeBinaryPackagesByFileOwnershipOverlap`: iterate relationships, identify OS-parent/language-child pairs, and delete the child package.

## Tests

Unit tests cover:
- OS → language package overlap (deb→python, apk→npm, rpm→ruby) — child removed
- Binary → language package overlap — both kept
- OS → OS overlap — both kept
- Language → language overlap — both kept
- Completeness: derived exclusion set matches the capability data, exceptions stay live
- Config validation (CLI and library) rejects enabling the exclusion without overlap relationships

Signed-off-by: June Kim <kimjune01@gmail.com>

---

*Edited 2026-07-20, alongside a rebase onto main and a force-push. Changes from the originally reviewed version: the exclusion list is now derived from cataloger capability data instead of hardcoded (per the review), the config option moved from `package.exclude-language-overlap-by-ownership` to `relationships.exclude-language-packages-with-file-ownership-overlap`, and library-side config validation was added. Earlier review comments referencing the old flag name or the hardcoded list predate this edit.*

### DIFF
--- a/cmd/syft/internal/commands/cataloger_list.go
+++ b/cmd/syft/internal/commands/cataloger_list.go
@@ -15,8 +15,8 @@ import (
 	"github.com/anchore/clio"
 	"github.com/anchore/syft/cmd/syft/internal/options"
 	"github.com/anchore/syft/internal/bus"
-	"github.com/anchore/syft/internal/capabilities"
 	"github.com/anchore/syft/internal/task"
+
 	"github.com/anchore/syft/syft/cataloging"
 )
 
@@ -305,7 +305,7 @@ func formatToken(token string, selection *task.TokenSelection, included bool, de
 }
 
 func extractTaskInfo(tasks []task.Task) ([]string, map[string][]string) {
-	infos := capabilities.ExtractCatalogerInfo(tasks)
+	infos := task.ExtractCatalogerInfo(tasks)
 
 	tagsByName := make(map[string][]string)
 	var names []string
--- a/cmd/syft/internal/options/catalog.go
+++ b/cmd/syft/internal/options/catalog.go
@@ -123,7 +123,8 @@ func (cfg Catalog) ToRelationshipsConfig() cataloging.RelationshipsConfig {
 		PackageFileOwnership:        cfg.Relationships.PackageFileOwnership,
 		PackageFileOwnershipOverlap: cfg.Relationships.PackageFileOwnershipOverlap,
 		// note: this option was surfaced in the syft application configuration before this relationships section was added
-		ExcludeBinaryPackagesWithFileOwnershipOverlap: cfg.Package.ExcludeBinaryOverlapByOwnership,
+		ExcludeBinaryPackagesWithFileOwnershipOverlap:   cfg.Package.ExcludeBinaryOverlapByOwnership,
+		ExcludeLanguagePackagesWithFileOwnershipOverlap: cfg.Relationships.ExcludeLanguagePackagesWithFileOwnershipOverlap,
 	}
 }
 
@@ -280,6 +281,11 @@ func (cfg *Catalog) PostLoad() error {
 	// the binary package exclusion code depends on the file overlap relationships being created upstream in processing
 	if !cfg.Relationships.PackageFileOwnershipOverlap && cfg.Package.ExcludeBinaryOverlapByOwnership {
 		return fmt.Errorf("cannot enable exclude-binary-overlap-by-ownership without enabling package-file-ownership-overlap")
+	}
+
+	// the language package exclusion code depends on the file overlap relationships being created upstream in processing
+	if !cfg.Relationships.PackageFileOwnershipOverlap && cfg.Relationships.ExcludeLanguagePackagesWithFileOwnershipOverlap {
+		return fmt.Errorf("cannot enable exclude-language-packages-with-file-ownership-overlap without enabling package-file-ownership-overlap")
 	}
 
 	return nil
--- a/cmd/syft/internal/options/relationships.go
+++ b/cmd/syft/internal/options/relationships.go
@@ -5,18 +5,22 @@ import "github.com/anchore/fangs"
 var _ fangs.FieldDescriber = (*relationshipsConfig)(nil)
 
 type relationshipsConfig struct {
-	PackageFileOwnership        bool `mapstructure:"package-file-ownership" json:"package-file-ownership" yaml:"package-file-ownership"`
-	PackageFileOwnershipOverlap bool `mapstructure:"package-file-ownership-overlap" json:"package-file-ownership-overlap" yaml:"package-file-ownership-overlap"`
+	PackageFileOwnership                            bool `mapstructure:"package-file-ownership" json:"package-file-ownership" yaml:"package-file-ownership"`
+	PackageFileOwnershipOverlap                     bool `mapstructure:"package-file-ownership-overlap" json:"package-file-ownership-overlap" yaml:"package-file-ownership-overlap"`
+	ExcludeLanguagePackagesWithFileOwnershipOverlap bool `mapstructure:"exclude-language-packages-with-file-ownership-overlap" json:"exclude-language-packages-with-file-ownership-overlap" yaml:"exclude-language-packages-with-file-ownership-overlap"`
 }
 
 func defaultRelationshipsConfig() relationshipsConfig {
 	return relationshipsConfig{
-		PackageFileOwnership:        true,
-		PackageFileOwnershipOverlap: true,
+		PackageFileOwnership:                            true,
+		PackageFileOwnershipOverlap:                     true,
+		ExcludeLanguagePackagesWithFileOwnershipOverlap: false,
 	}
 }
 
 func (r *relationshipsConfig) DescribeFields(descriptions fangs.FieldDescriptionSet) {
 	descriptions.Add(&r.PackageFileOwnership, "include package-to-file relationships that indicate which files are owned by which packages")
 	descriptions.Add(&r.PackageFileOwnershipOverlap, "include package-to-package relationships that indicate one package is owned by another due to files claimed to be owned by one package are also evidence of another package's existence")
+	descriptions.Add(&r.ExcludeLanguagePackagesWithFileOwnershipOverlap, `exclude language packages from the sbom when their files are owned by an OS package
+these packages are removed to prevent duplicate entries for packages installed via system package managers`)
 }
--- a/internal/capabilities/capabilities.go
+++ b/internal/capabilities/capabilities.go
@@ -8,10 +8,7 @@ import (
 	"path/filepath"
 	"sort"
 
-	"github.com/scylladb/go-set/strset"
 	"gopkg.in/yaml.v3"
-
-	"github.com/anchore/syft/internal/task"
 )
 
 //go:generate go run ./generate
@@ -105,28 +102,4 @@ func Packages() ([]CatalogerEntry, error) {
 type CatalogerInfo struct {
 	Name      string
 	Selectors []string // tags for cataloger name selection
-}
-
-// ExtractCatalogerInfo extracts cataloger names and their selection tags from tasks
-func ExtractCatalogerInfo(tasks []task.Task) []CatalogerInfo {
-	var infos []CatalogerInfo
-
-	for _, tsk := range tasks {
-		var selectors []string
-		name := tsk.Name()
-
-		if s, ok := tsk.(task.Selector); ok {
-			set := strset.New(s.Selectors()...)
-			set.Remove(name)
-			selectors = set.List()
-			sort.Strings(selectors)
-		}
-
-		infos = append(infos, CatalogerInfo{
-			Name:      name,
-			Selectors: selectors,
-		})
-	}
-
-	return infos
 }
--- a/internal/capabilities/internal/cataloger_names.go
+++ b/internal/capabilities/internal/cataloger_names.go
@@ -17,7 +17,10 @@ func AllPackageCatalogerInfo() ([]capabilities.CatalogerInfo, error) {
 		return nil, fmt.Errorf("unable to create pkg cataloger tasks: %w", err)
 	}
 
-	infos := capabilities.ExtractCatalogerInfo(allPkgTasks)
+	var infos []capabilities.CatalogerInfo
+	for _, info := range task.ExtractCatalogerInfo(allPkgTasks) {
+		infos = append(infos, capabilities.CatalogerInfo(info))
+	}
 
 	// sort by name for consistency
 	sort.Slice(infos, func(i, j int) bool {
--- a/internal/relationship/exclude_language_packages_by_file_ownership_overlap.go
+++ b/internal/relationship/exclude_language_packages_by_file_ownership_overlap.go
@@ -1,0 +1,67 @@
+package relationship
+
+import (
+	"slices"
+
+	"github.com/anchore/syft/internal/sbomsync"
+	"github.com/anchore/syft/syft/artifact"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/sbom"
+)
+
+var (
+	languageCatalogerTypes = []pkg.Type{
+		pkg.PythonPkg,
+		pkg.GemPkg,
+		pkg.NpmPkg,
+		pkg.PhpComposerPkg,
+		pkg.PhpPeclPkg,
+		pkg.Rpkg,
+		pkg.LuaRocksPkg,
+		pkg.ErlangOTPPkg,
+	}
+)
+
+func ExcludeLanguagePackagesByFileOwnershipOverlap(accessor sbomsync.Accessor) {
+	accessor.WriteToSBOM(func(s *sbom.SBOM) {
+		for _, r := range s.Relationships {
+			if idToRemove := excludeLanguagePackageByFileOwnershipOverlap(r, s.Artifacts.Packages); idToRemove != "" {
+				s.Artifacts.Packages.Delete(idToRemove)
+				s.Relationships = RemoveRelationshipsByID(s.Relationships, idToRemove)
+			}
+		}
+	})
+}
+
+// excludeLanguagePackageByFileOwnershipOverlap will remove language packages that are owned by OS packages.
+// This was implemented as a way to help resolve: https://github.com/anchore/syft/issues/4760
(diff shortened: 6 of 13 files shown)


kimjune01:
Pushed two test-only follow-ups (b4844d07, 64376c21): the original tests built `pkg.Package` literals without `SetID()`, so `c.Package(r.From.ID())` returned nil and the assertion compared `"" == ""` — the suite passed without exercising the code under test. Added `SetID()` on the test packages and an `assert.NotEqual(t, "", result)` guard so the no-op can't recur. Verified by injecting a panic in the function body: pre-fix tests pass (proof of no-op), post-fix tests fail with the panic (proof of exercise).

Dashtid:
The `languageCatalogerTypes` set covers 8 types but omits some commonly-distro-packaged language types — notably `JavaPkg`, `GoModulePkg`, `RustPkg`, and `SwiftPkg`. The same OS↔language overlap exists for these:

- Debian's `golang-github-*` packages install Go source that the Go cataloger also picks up
- Debian's `rust-*` packages install Rust source that the Rust cataloger also picks up
- `libreoffice-java-common` and similar install JARs that the Java cataloger reports

This could be intentional — e.g. Java/Go/Rust have different overlap semantics worth handling separately — or just an early-iteration scoping. If the latter, it might be worth either expanding the list to cover all language types, or codifying a criterion for inclusion (`pkg.AllPkgs` minus OS + binary + non-language types like `GithubActionPkg`/`TerraformPkg`/`WordpressPluginPkg`). Either way, a short comment on the slice explaining the inclusion rule would make the intent legible for future cataloger PRs that add new `Type`s.

kimjune01:
I derived the set from syft's own `language`-tagged catalogers instead of hand-listing. That adds the installed-package types below — but also excludes the binary/archive extractors, since deleting those on file-overlap loses real SBOM data. A guard test re-derives the set from `capabilities.yaml` and fails CI when a new language cataloger isn't classified.

| types | decision | why |
|---|---|---|
| cocoapods, conan, dart-pub, hackage, hex, opam, php-pear, swift, swipl | added | OS overlap here means the distro repackaged the same unit, so dedup is safe |
| java-archive, go-module, rust-crate, dotnet, graalvm-native-image | excluded | these extract many components from one OS-owned binary/archive — the OS owns the container, not the components, so deleting drops distinct packages (an rpm owning a Go binary would erase its built-in modules) |
| php-pecl | kept (exception) | not language-tagged anymore (deprecated cataloger), but its packages still overlap until v2.0 |

Tradeoff: excluding whole types means safe go.mod/Cargo.lock overlaps won't be deduped either — better to under-delete than lose a real component.

Limits: the rule keys on `pkg.Type`, a coarse proxy. Some included types are lockfile-derived (cocoapods → Podfile.lock, swift → Package.resolved), so an OS package owning one of those manifests could over-delete declared deps — same issue, rarer. The precise fix is per-package installed-vs-declared evidence. Only unit-tested, not validated against real images for the new ecosystems. #4974 tracks the evidence-vs-type question plus a few tagging inconsistencies (php-pecl on a deprecated/untagged cataloger, nix as both OS and language, dotnet-deps-binary missing the binary selector go/rust carry).

Dashtid:
The extractor-vs-installed split is the right framing; my initial cut missed it. Deriving the set from `capabilities.yaml` plus a guard test that catches unclassified additions is cleaner than the hand-listed alternative.

Spot-checked one #4974 item locally: `LanguageByName` at `syft/pkg/language.go:71` returns `UnknownLanguage` for `string(Rpkg)` ("R-package"), `string(LuaRocksPkg)` ("lua-rocks"), and `string(PhpPeclPkg)` ("php-pecl") — the switch matches `"r"`/`TypeCran`, `string(Lua)`/`TypeLuaRocks`, and the PHP cases without `PhpPeclPkg`, so the const-value strings fall through. `TestLanguageByName` covers the lowercase input `"r"` but not `string(Rpkg)`. Worth a narrow PR independent of the evidence-vs-type work.

kzantow:
We talked about this PR on [the livestream last week](https://www.youtube.com/live/oB0I7ePP_e0?si=8tk_oCAix5F3By7g&t=614). I think the main hang-up is the  [hardcoded list](https://github.com/anchore/syft/pull/4905/changes#diff-922137518224a0899e94df3a5eab5a56d2d7b67b8067f8d6bda136ffbdb71e11R34-R52) of language package types, which is almost guaranteed to change -- we want to try to get this list dynamically from the capabilities. Otherwise, the single flag here is a good start, though at some point we foresee wanting to tailor which package types or catalogers are included in this behavior, we don't need that to land this PR.

We have [another ask related to this to programmatically list catalogers](https://github.com/anchore/syft/pull/5046) and other info that we've decided we could support with a new package not bound by Syft 1.0 stability guarantees in a `syft/exp` experimental API. These seem to be directly related that both want to be able to list catalogers and potentially tags.

Sorry that's a little dichotomy but it is to say what you've implemented here generally looks good to include a single flag for parity with the binary exclusions as long as it's pulling a list of language packages dynamically..

kimjune01:
Pushed the capabilities-derived version.

You were right that the list would drift. It was already stale on main. vcpkg landed while this PR sat, its cataloger is `language`-tagged, and the hardcoded list doesn't have it. The derived set picks it up with no code change.

The derivation follows the shape you described on the stream. Exclusion now reads the same capability data that drives cataloger selection, so there's no second list to keep in sync.

The five-type exception list is still hardcoded, because the capability data can't decide it. I first derived the exclusions from the `binary` selector as well, and that silently dropped npm packages. `dotnet-deps-binary-cataloger` emits npm alongside `javascript-package-cataloger`, and `go-module` has the same dual-emitter shape from `go-module-file-cataloger` and `go-module-binary-cataloger`. Whether OS ownership subsumes a type depends on which emitter dominates in practice, and the capability data doesn't record that. So inclusion derives from the `language` selector, and the binary-extracted subtraction stays explicit.

Deriving the set meant breaking an import cycle (`relationship → capabilities → task → relationship`), so `ExtractCatalogerInfo` moved from `internal/capabilities` into `internal/task`. The `CatalogerInfo` type stayed put, so `generate/` is untouched. I can split that move into its own PR if you'd rather review it separately.

The config option also moved to `relationships.exclude-language-packages-with-file-ownership-overlap`, named after the library field. The note in catalog.go says the binary flag only sits under `package` because it predates the relationships section, and that seemed like a convention not to copy.

PR #4926, opened 2026-05-18
perf(python): hoist name normalization regexp to package level

matiasinsaurralde:
Avoid recompiling the separator pattern on every normalize() call during cataloging.

## Description

- Hoist the `[-_.]+` regexp used by `normalize()` in the Python cataloger to a package-level variable so it is compiled once at init instead of on every package name normalization.
- `normalize()` is on the hot path when building Python packages (called for each package created from requirements, setup metadata, wheel records, etc.).

## Type of change

- [] Documentation (updates the documentation)
- [] Chore (improve the developer experience, fix a test flake, etc, without changing the visible behavior of Syft)

## Checklist

## Issue references

### DIFF
--- a/syft/pkg/cataloger/python/package.go
+++ b/syft/pkg/cataloger/python/package.go
@@ -11,10 +11,11 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
+// https://packaging.python.org/en/latest/specifications/name-normalization/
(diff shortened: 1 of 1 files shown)


spiffcs (approved):

spiffcs:
Nice catch! Running and will merge - ty for the contribution @matiasinsaurralde

PR #4943, opened 2026-05-27
Add hierarchical dependency resolution from embedded POM files

### LINKED ISSUE #4921: Opt-in hierarchical dependency resolution for Java archives

Miablo:
## Description
- Adds a dependency graph built from embedded `pom.xml` files in Java archives, wiring auxiliary packages to their actual Maven parents rather than flat to the root archive
- Enriches dependency relationships with depth, scope, and direct/transitive metadata for Syft JSON output
- Deferred parent resolution via post-processor handles cross-archive dependencies
- Gated by `java.use-embedded-pom-dependencies` (default: false) — no behavior change unless opted in

### Design decisions
- `maven.ID` as map key
- `FindNodeByGA` fallback to handle version-mismatch scenarios where dependency management resolves a different version than what the POM declares
- `ID.Coordinate()`: explicit, round-trippable serialization (`groupId:artifactId:version`) used for `IntendedParentID` in JSON
- Branch-scoped cycle detection: visited map is copied per branch so diamond dependencies are still resolved
- Post-processor 3-tier index: `full coord → groupId:artifactId → artifactId` fallback for cross-archive parent resolution
- Feature flag default off: zero behavioral change for existing users

## Type of change

## Checklist

## Issue references
This PR splits #4843 to focus on embedded POM graph only and incorporates all architectural feedback from @kzantow's review:

1. Relationship logic moved to `pkg/cataloger/java`
2. Java-specific config, `UseEmbeddedPOMDependencies`, moved from shared relationship options to `ArchiveCatalogerConfig`
3. Use `maven.ID` as map key with explicit serialization and `maven_id.go` wrapper removed
4. Version-agnostic graph lookup fallback by added `FindNodeByGA(groupID, artifactID string)` as a fallback when exact `maven.ID` match (including version) fails. This handles the scenario where Maven dependency management resolves a different version than what the POM declares and the graph still finds the correct node by `groupId:artifactId`.
5. Depth is computed, not stored. The depth value in `DependencyRelationshipData` is output enrichment for SBOM consumers, not structural redundancy on the graph.
6. Lookup maps extracted into helper where inline map construction in `archive_parser.go` was replaced with a `buildPkgIndex()` method. Relationship creation split into clearly named `createAuxPkgRelationship` and `createMainPkgRelationship` methods.

### DIFF
--- a/cmd/syft/internal/options/catalog.go
+++ b/cmd/syft/internal/options/catalog.go
@@ -211,7 +211,8 @@ func (cfg Catalog) ToPackagesConfig() pkgcataloging.Config {
 			WithUseNetwork(*multiLevelOption(false, enrichmentEnabled(cfg.Enrich, task.Java, task.Maven), cfg.Java.UseNetwork)).
 			WithMavenBaseURL(cfg.Java.MavenURL).
 			WithArchiveTraversal(archiveSearch, cfg.Java.MaxParentRecursiveDepth).
-			WithResolveTransitiveDependencies(cfg.Java.ResolveTransitiveDependencies),
+			WithResolveTransitiveDependencies(cfg.Java.ResolveTransitiveDependencies).
+			WithUseEmbeddedPOMDependencies(cfg.Java.UseEmbeddedPOMDependencies),
 	}
 }
 
--- a/cmd/syft/internal/options/java.go
+++ b/cmd/syft/internal/options/java.go
@@ -12,6 +12,7 @@ type javaConfig struct {
 	MavenURL                      string `yaml:"maven-url" json:"maven-url" mapstructure:"maven-url"`
 	MaxParentRecursiveDepth       int    `yaml:"max-parent-recursive-depth" json:"max-parent-recursive-depth" mapstructure:"max-parent-recursive-depth"`
 	ResolveTransitiveDependencies bool   `yaml:"resolve-transitive-dependencies" json:"resolve-transitive-dependencies" mapstructure:"resolve-transitive-dependencies"`
+	UseEmbeddedPOMDependencies    bool   `yaml:"use-embedded-pom-dependencies" json:"use-embedded-pom-dependencies" mapstructure:"use-embedded-pom-dependencies"`
 }
 
 func defaultJavaConfig() javaConfig {
@@ -24,6 +25,7 @@ func defaultJavaConfig() javaConfig {
 		MavenLocalRepositoryDir:       def.MavenLocalRepositoryDir,
 		MavenURL:                      def.MavenBaseURL,
 		ResolveTransitiveDependencies: def.ResolveTransitiveDependencies,
+		UseEmbeddedPOMDependencies:    def.UseEmbeddedPOMDependencies,
 	}
 }
 
@@ -46,4 +48,6 @@ build, run 'mvn help:effective-pom' before performing the scan with syft.`)
 	descriptions.Add(&o.MavenLocalRepositoryDir, `override the default location of the local Maven repository. 
 the default is the subdirectory '.m2/repository' in your home directory`)
 	descriptions.Add(&o.ResolveTransitiveDependencies, `resolve transient dependencies such as those defined in a dependency's POM on Maven central`)
+	descriptions.Add(&o.UseEmbeddedPOMDependencies, `build a hierarchical dependency graph from embedded pom.xml files within archives.
+When enabled, dependency relationships are wired to their actual Maven parent rather than all pointing to the root archive`)
 }
--- a/internal/task/relationship_tasks.go
+++ b/internal/task/relationship_tasks.go
@@ -9,6 +9,7 @@ import (
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/cataloging"
 	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg/cataloger/java"
 	"github.com/anchore/syft/syft/sbom"
 	"github.com/anchore/syft/syft/source"
 )
@@ -23,13 +24,16 @@ func (s sourceIdentifierAdapter) ID() artifact.ID {
 	return artifact.ID(s.desc.ID)
 }
 
-func NewRelationshipsTask(cfg cataloging.RelationshipsConfig, src source.Description) Task {
+func NewRelationshipsTask(cfg cataloging.RelationshipsConfig, javaCfg java.ArchiveCatalogerConfig, src source.Description) Task {
 	fn := func(_ context.Context, resolver file.Resolver, builder sbomsync.Builder) error {
 		finalizeRelationships(
 			resolver,
 			builder,
 			cfg,
 			&sourceIdentifierAdapter{desc: src})
+
+		accessor := builder.(sbomsync.Accessor)
+		java.ResolveHierarchicalDependencies(accessor, javaCfg)
 
 		return nil
 	}
--- a/syft/create_sbom_config.go
+++ b/syft/create_sbom_config.go
@@ -420,7 +420,7 @@ func (c *CreateSBOMConfig) scopeTasks() []task.Task {
 func (c *CreateSBOMConfig) relationshipTasks(src source.Description) []task.Task {
 	var tsks []task.Task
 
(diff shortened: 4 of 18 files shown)


Miablo:
@kzantow -- Hi! When you get a chance, could you approve the workflow runs on this PR? And if they pass, a review would be much appreciated. Thank you in advance!

PR #4948, opened 2026-06-02
feat (npm): Handle npm-shrinkwrap in the same way as package-lock

paazmaya:
## Description

Add support for `npm-shrinkwrap.json` files.
https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#package-lockjson-vs-npm-shrinkwrapjson

## Type of change

## Checklist

Have not been able to get `make test` pass even on `main` branch...

### DIFF
--- a/cmd/syft/internal/test/integration/node_packages_test.go
+++ b/cmd/syft/internal/test/integration/node_packages_test.go
@@ -31,6 +31,27 @@ func TestNpmPackageLockDirectory(t *testing.T) {
 	}
 }
 
+func TestNpmShrinkwrapDirectory(t *testing.T) {
+	sbom, _ := catalogDirectory(t, "testdata/npm-shrinkwrap")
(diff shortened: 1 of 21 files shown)


spiffcs (commented):
Thanks @paazmaya for the PR!

I think I have a couple questions that our team will have to weigh in on.

> Both of these files have the same format, and perform similar functions in the root of a project.

If the above is correct then do we need a whole new parser/cataloger? Should these be parsed by the current `parsePackageLock` function?

> If both package-lock.json and npm-shrinkwrap.json are present in the root of a project, npm-shrinkwrap.json will take precedence and package-lock.json will be ignored.

I think we need to consider this case so we can follow the package manager behavior and not include packages that would otherwise be ignored.

cc @anchore/tools for thoughts on this unique javascript package parsing case

PR #4950, opened 2026-06-02
fix(python): catalog uv PEP 723 script lockfiles (*.py.lock)

### LINKED ISSUE #4949: Python cataloger misses uv PEP 723 script lockfiles (`*.py.lock`)
### What happened

Syft's Python cataloger only catalogs uv lockfiles whose basename is exactly `uv.lock`:

```go
// syft/pkg/cataloger/python/cataloger.go
WithParserByGlobs(uvLockParser.parseUvLock, "**/uv.lock")
```

[uv supports locking dependencies for PEP 723 scripts](https://docs.astral.sh/uv/guides/scripts/#locking-dependencies) via `uv lock --script <script>.py`, which writes a lockfile named `<script>.py.lock` (e.g. `main.py.lock`) alongside the script. These files use the **exact same TOML format** as `uv.lock`, but because the glob requires the basename `uv.lock`, they never match and the parser is never invoked. Their dependencies are silently absent from the SBOM.

### What I expected to happen

Syft catalogs uv PEP 723 script lockfiles (`*.py.lock`) the same way it catalogs `uv.lock`, since the format is identical.

### Steps to reproduce

Given a `uv.lock` and a copy named `main.py.lock` with identical content:

```
$ syft dir:./named-uv-lock      # dir contains uv.lock
... 4 python packages cataloged

$ syft dir:./named-py-lock       # same content, named main.py.lock
... 0 python packages cataloged
```

The parser itself handles the content fine — it's purely the glob that excludes these files.

### Impact

In a real repository that uses uv for standalone PEP 723 scripts (common for data pipelines, automation jobs, and one-off tooling), every `*.py.lock` file is silently skipped during cataloging. This can hide a large number of Python dependencies — and their known vulnerabilities — from Syft and downstream consumers like Grype, with no warning or error to indicate anything was missed. Projects that rely exclusively on script lockfiles (rather than a top-level `uv.lock`) get effectively empty Python SBOMs.

### Suggested fix

Add `**/*.py.lock` to the uv lock parser globs:

```go
WithParserByGlobs(uvLockParser.parseUvLock, "**/uv.lock", "**/*.py.lock")
```

The parser already validates the lockfile version and decodes TOML, so non-uv `*.py.lock` files (if any) fail gracefully. Happy to open a PR — the change plus a glob-path test fixture passes the existing cataloger test suite.

### Anchore Syft version

Reproduced on syft `v1.44.0`; `main` still registers only `**/uv.lock`.

### Tested against

Cataloging path verified via both `grype dir:` and the two-step `syft -o cyclonedx-json` → `grype sbom:` flow.

ktopcuoglu:
Closes #4949.

## Cause

The Python cataloger registers the uv lock parser against a single glob in `syft/pkg/cataloger/python/cataloger.go`:

```go
WithParserByGlobs(uvLockParser.parseUvLock, "**/uv.lock")
```

`**/uv.lock` matches only files whose basename is exactly `uv.lock`. uv also produces lockfiles for [PEP 723 scripts](https://docs.astral.sh/uv/guides/scripts/#locking-dependencies) via `uv lock --script <name>.py`, naming them `<name>.py.lock` (e.g. `main.py.lock`). These use the identical TOML format, but the glob never matches them, so `parseUvLock` is never invoked and their packages are silently absent from the SBOM — with no warning.

## Fix

Add `**/*.py.lock` to the parser's globs so script lockfiles are cataloged the same way as `uv.lock`:

```go
WithParserByGlobs(uvLockParser.parseUvLock, "**/uv.lock", "**/*.py.lock")
```

`capabilities.yaml` was regenerated with `go generate ./internal/capabilities/...` to keep the declared detector globs in sync (idempotent — no further drift). The parser already validates the lockfile version and decodes TOML, so any non-uv `*.py.lock` file fails gracefully.

## Tests

- Added a `src/script.py.lock` fixture to the python `glob-paths` testdata and the expected query set in `Test_IndexCataloger_Globs`. This test **fails without the glob change** (the cataloger never queries `script.py.lock`) and **passes with it** — guarding the fix against regression.
- Existing `TestParseUvLock` already covers parsing of the (format-identical) lockfile content.

Verified locally:
- `go test ./syft/pkg/cataloger/python/...` passes; `go generate ./internal/capabilities/...` produces no drift.
- Built `syft` against the change and confirmed a `main.py.lock` PEP 723 script lock now yields the same Python packages as an identically-named `uv.lock`, end to end (`syft -o cyclonedx-json` → downstream vuln match).

## Type of change

## Checklist


syft/pkg/cataloger/python/cataloger.go
```diff
@@ -23,7 +23,10 @@ func NewPackageCataloger(cfg CatalogerConfig) pkg.Cataloger {
 		WithParserByGlobs(poetryLockParser.parsePoetryLock, "**/poetry.lock").
 		WithParserByGlobs(pipfileLockParser.parsePipfileLock, "**/Pipfile.lock").
 		WithParserByGlobs(setupFileParser.parseSetupFile, "**/setup.py").
-		WithParserByGlobs(uvLockParser.parseUvLock, "**/uv.lock").
+		// uv lock files are named "uv.lock", but PEP 723 script lock files
+		// (created by "uv lock --script <name>.py") are named "<name>.py.lock"
+		// and use the same format, so catalog both.
+		WithParserByGlobs(uvLockParser.parseUvLock, "**/uv.lock", "**/*.py.lock").
```
ktopcuoglu, line 29: On the choice of `**/*.py.lock` vs. a narrower glob: uv's documented convention is to append `.lock` to the script filename, so a PEP 723 `.py` script always locks to `<name>.py.lock` ([docs](https://docs.astral.sh/uv/guides/scripts/#locking-dependencies)). `*.py.lock` matches exactly that shape without enumerating script names. A stricter pattern would need to anticipate arbitrary script basenames, and there is no other common tool that emits `*.py.lock`. If a non-uv file ever does match, `parseUvLock` decodes TOML and validates the lock-file `version` field, returning an error (surfaced as `unknown`) rather than producing bogus packages — so the failure mode is graceful. Happy to scope it differently if you'd prefer.

ktopcuoglu:
Hi [wagoodman](https://github.com/wagoodman) 👋 — apologies for the direct ping, and truly no rush. This is a small, self-contained fix (closes #4949): uv PEP 723 script lockfiles (`*.py.lock`, produced by `uv lock --script`) aren't being cataloged because the glob only matches `uv.lock`, even though the format is identical.

The change is one glob plus a regenerated `capabilities.yaml` and a regression test. The CI workflows are currently in `action_required` (first-time-contributor gate), so they haven't run yet — whenever you or another maintainer has a spare moment, would you mind approving the workflow run so the checks can go green?

More than happy to adjust anything you'd like. Thanks so much for your time and for maintaining syft 🙏

ktopcuoglu:
Also gently looping in @kzantow 🙏 — no pressure at all, and sorry for the extra ping. Since you've worked in the Python cataloger area, you might be a good person to glance at this one.

Quick recap: it's a small, self-contained fix (closes #4949) so uv PEP 723 script lockfiles (`*.py.lock`) get cataloged like `uv.lock` — one glob, a regenerated `capabilities.yaml`, and a regression test. The CI workflows are still in `action_required` (first-time-contributor gate), so whenever a maintainer has a spare moment to approve the run, the checks can go green.

Totally happy to adjust anything. Thanks so much for your time and for maintaining syft!

spiffcs:
I think this one might be a little bigger and have some considerations we need to talk about as a team:

Should we add a field to `PythonUvLockEntry` (e.g. source lockfile name or a 'script' flag) so downstream
consumers can tell script deps from project deps?

I think this would get surfaced by the location lists of the different packages on this branch, but I think that by default is a little confusing for users.

### Example
```
A data-engineering repo that uses uv for both the project and a standalone PEP 723 script:

  repo/
  ├── uv.lock                      # project env: pins requests 2.32.3
  └── scripts/
      ├── backfill.py              # PEP 723 script: # /// script ... requests ...
      └── backfill.py.lock         # script env: ALSO pins requests 2.32.3

  Both lockfiles independently pin requests 2.32.3. syft now (correctly) catalogs two distinct packages:

  ┌─────────────────┬──────────────────────────┬──────────────────────────┐
  │                 │        Package A         │        Package B         │
  ├─────────────────┼──────────────────────────┼──────────────────────────┤
  │ Name@Version    │ requests 2.32.3          │ requests 2.32.3          │
  ├─────────────────┼──────────────────────────┼──────────────────────────┤
  │ PURL            │ pkg:pypi/requests@2.32.3 │ pkg:pypi/requests@2.32.3 │
  ├─────────────────┼──────────────────────────┼──────────────────────────┤
  │ Type / Language │ PythonPkg / Python       │ PythonPkg / Python       │
  ├─────────────────┼──────────────────────────┼──────────────────────────┤
  │ Location        │ uv.lock                  │ scripts/backfill.py.lock │
  ├─────────────────┼──────────────────────────┼──────────────────────────┤
  │ Package ID      │ a1b2…                    │ c3d4…                    │
  └─────────────────┴──────────────────────────┴──────────────────────────┘
```

I think the above example is something we should consider for this PR and see if there are any tweaks or additional relationships we should add at production of the SBOM to make this kind of situation a bit clearer.

spiffcs (commented):
Added comment and needs-discussion for this one to get team feedback on how we should treat "duplicate" packages across the two files

ktopcuoglu:
Thanks @spiffcs! Yeah, fair point — I hadn't really thought about the project-vs-script ambiguity until you raised it, and I agree it's probably worth a team conversation rather than me just deciding it in this PR.

To confirm your table with the mechanism behind it: `Location` is part of the package ID hash, so the two entries get different IDs and never merge — the lockfile path really is the only thing distinguishing them by default, and I agree that's not obvious to users.

A few half-formed thoughts on the field itself, fwiw:

- A `script` flag (or the source lockfile name) would definitely make the project-vs-script case clearer. The thing I'm less sure about: it doesn't actually cut down the duplication. In a repo with lots of PEP 723 scripts, most of the repetition is script-to-script (tons of scripts each pinning `six`/`urllib3`/whatever), and the flag would just be `true` on all of them — so IMO it's a useful label, but the "lots of near-identical entries" thing sticks around either way, since that's really more about scanning many independent lockfiles than about scripts specifically.
- I think the flag would mostly help syft's own JSON — in CycloneDX it'd probably come through as a `syft:` property and in SPDX I'd guess it just gets dropped, so anyone on the standard formats would still be leaning on `Location` anyway. (Could be wrong on the exact behavior there, haven't checked closely.)
- Honestly your "additional relationships at SBOM production" idea feels like maybe the better direction to me — modeling the script lock as its own source, or a relationship from the script to its deps, seems like it'd capture provenance more naturally than a boolean on every package. And it might be worth deciding whether this is a uv-specific thing or something more general, since poetry/npm workspaces have a similar project-vs-sub-lockfile shape.

On sequencing — totally your call — would it make sense to keep this PR as just the glob fix (so the script deps at least stop getting silently dropped), and split the "how do we make this clearer to users" question into its own issue where the team can hash out flag-vs-relationship? Happy to write that follow-up and take a crack at whichever way you all land — but if you'd rather keep it all together here, that's fine too.

spiffcs (approved):

ktopcuoglu:
Thanks for merging this @spiffcs 🙏 @willmurphyscode — no rush at all, but since you usually cut the releases: any rough sense of when the next one might land? Just planning on our end for when this `*.py.lock` fix is in a tag. Thanks!

PR #4954, opened 2026-06-04
feat: add .bpl extension to PE cataloger

### LINKED ISSUE #4664: Add .bpl file support to the PE / DLL cataloger

jfjrh2014:
Adds `**/*.bpl` to the PE binary package cataloger globs so Borland Package Library files are picked up during scans.

BPL files are standard PE/DLL format used by Delphi and C++Builder. Confirmed by the issue reporter: renaming `.bpl` → `.dll` already works with the existing cataloger.

Closes #4664

### DIFF
--- a/internal/task/package_tasks.go
+++ b/internal/task/package_tasks.go
@@ -161,7 +161,7 @@ func DefaultPackageTaskFactories() Factories {
 			pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, pkgcataloging.InstalledTag, pkgcataloging.ImageTag, "binary",
 		),
 		newSimplePackageTaskFactory(binary.NewELFPackageCataloger, pkgcataloging.DeclaredTag, pkgcataloging.DirectoryTag, pkgcataloging.InstalledTag, pkgcataloging.ImageTag, "binary", "elf-package", "elf"),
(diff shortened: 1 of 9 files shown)


spiffcs (approved):
Thanks @jfjrh2014 for the PR!

I boyscouted some tests and regenerated the capabilities.yaml.

LGTM. 

Looking for one more @anchore/tools review since I have a commit on this branch now

willmurphyscode (approved):

PR #4959, opened 2026-06-05
fix: local version identifiers in python requirements parsing

### LINKED ISSUE #4958: requirements.txt cataloger silently drops PEP 440 local version identifiers, producing incorrect PURL
**What happened**:
When syft parses a requirements.txt entry whose version string includes a PEP 440 local version identifier (the `+<local>` segment), the local part is silently dropped from both the cataloged version and the emitted PURL.

Concrete reproducer — a single-line requirements.txt:
```
urllib3==1.26.18+cgr.2
```

Command to run (this is exactly how scan.py invokes it):
```
grype dir:./fixtures/python/pip-prebuild-chainguard -o cyclonedx-json
```

What we observed on grype 0.113.0 / DB v6.1.7 (2026-06-04 build):
```
    name    : urllib3
    version : 1.26.18
    purl    : pkg:pypi/urllib3@1.26.18
```

The `+cgr.2` is gone — version and PURL both. Because the chainguard-libraries DB provider keys unaffected entries off the full version (`pkg:pypi/urllib3@1.26.18%2Bcgr.2`), the PURL match fails and VEX suppression never applies. Mutual customers end up seeing CVEs that Chainguard has already fixed.

The full PEP 440 version, including the `+local` segment, should be preserved in both the cataloged version and the emitted PURL (matches what's actually pinned on disk):
```
    name    : urllib3
    version : 1.26.18+cgr.2
    purl    : pkg:pypi/urllib3@1.26.18%2Bcgr.2
```

`+local` is the spec-mandated syntax for PEP 440 local version identifiers.

PEP 440 §local-version-identifiers is the reference: https://peps.python.org/pep-0440/#local-version-identifiers

See above.

#### Scope of the bug
- Affects pip's requirements.txt path only.
- uv.lock and poetry.lock work correctly — syft reads those as TOML and preserves the full `+cgr.N` version string in the emitted PURL.
- Trivy on the same requirements.txt emits the correct `pkg:pypi/urllib3@1.26.18%2Bcgr.2`, so this is a syft-specific parser issue, not a fundamental limitation.
- Once the package is installed and we're scanning the `.dist-info` METADATA on disk (postbuild / container), the suffix survives.

So the bug surface is just: pip + prebuild (and any other path that reads from requirements.txt). Affects every Chainguard-remediated PyPI package in the catalogue when pinned that way.

#### Root cause (verified against current source)

`syft/pkg/cataloger/python/parse_requirements.go:33`:

    versionConstraintPattern = `(?P<versionConstraint>([^\S\r\n]*[~=>!<]+\s*[0-9a-zA-Z.*]+[^\S\r\n]*,?)+)?...`

The character class `[0-9a-zA-Z.*]+` covers PEP 440 release/pre/post segments but does NOT include `+`. Since PEP 440 local version IDs always start with `+`, the regex stops at the `+` and the local segment
gets silently truncated.

#### Suggested fix
Extend the version character class:
```
    -[0-9a-zA-Z.*]+
    +[0-9a-zA-Z.+*-]+
```
Also include `-` since it appears in some prerelease segments.

**Environment**:
- Grype v0.113.0 (Syft v1.45.0 embedded)
- DB schema v6.1.7, built 2026-06-04
- Tested on darwin/arm64 and linux/amd64, same behaviour

kzantow:
## Description

This PR adds support for [_local version segments_](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-segments) in Python requirements.txt, such as `1.2.3+ubuntu.1`.

## Type of change

## Checklist

## Issue references

Fixes: #4958

### DIFF
--- a/syft/pkg/cataloger/python/parse_requirements.go
+++ b/syft/pkg/cataloger/python/parse_requirements.go
@@ -29,8 +29,8 @@ const (
 	// namePattern matches: requests[security]
 	namePattern = `(?P<name>\w[\w\[\],\s-_\.]+)`
 
-	// versionConstraintPattern matches: == 2.8.*
-	versionConstraintPattern = `(?P<versionConstraint>([^\S\r\n]*[~=>!<]+\s*[0-9a-zA-Z.*]+[^\S\r\n]*,?)+)?(@[^\S\r\n]*(?P<url>[^;]*))?`
+	// versionConstraintPattern matches: == 2.8.* (including local version identifiers, e.g. == 1.2.3+gcr.2)
+	versionConstraintPattern = `(?P<versionConstraint>([^\S\r\n]*[~=>!<]+\s*[0-9a-zA-Z.*+]+[^\S\r\n]*,?)+)?(@[^\S\r\n]*(?P<url>[^;]*))?`
 
 	// markersPattern matches: python_version < "2.7" and sys_platform == "linux"
 	markersPattern = `(;(?P<markers>.*))?`
--- a/syft/pkg/cataloger/python/parse_requirements_test.go
+++ b/syft/pkg/cataloger/python/parse_requirements_test.go
@@ -140,6 +140,18 @@ func TestParseRequirementsTxt(t *testing.T) {
 				VersionConstraint: "== 1.0.0",
 			},
 		},
+		{
+			Name:      "local-version",
+			Version:   "1.2.3+gcr.2",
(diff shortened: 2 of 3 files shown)


spiffcs (approved):
This one is 🟢 

I think one small note (non blocking).
From the link:
> the use of - and _ is also acceptable

So `[0-9a-zA-Z.*+]` → `[0-9a-zA-Z.*+_-]` which adds `_-` on the end would cover any versions we ran into that included `_` or `-`

Issue #1847, opened 2023-05-25
Feature: OpenWrt specific cataloger


spiffcs:
**What would you like to be added**:
I followed these instructions to generate an Open WRT docker image:
[OpenWrt as a Docker Image](https://openwrt.org/docs/guide-user/virtualization/docker_openwrt_image)

After scanning the image with syft I found only kernel modules being detected:
```
syft openwrt-x86-generic-rootfs:latest
 ✔ Loaded image
 ✔ Parsed image
 ✔ Cataloged packages      [48 packages]

NAME     VERSION         TYPE
                         linux-kernel-module
         0.9.28          linux-kernel-module
         2.3LK-NAPI      linux-kernel-module
         3.122           linux-kernel-module
         3.5.24-k2-NAPI  linux-kernel-module
         7.3.21-k8-NAPI  linux-kernel-module
busybox  1.19.4          binary
```

When I query the images package manager I see:
```
/ # opkg list-installed
base-files - 117-r36088
busybox - 1.19.4-6
dnsmasq - 2.62-2
dropbear - 2011.54-2
firewall - 2-55.1
grub - 0.97-3
hotplug2 - 1.0-beta-4
iptables - 1.4.10-4
jshn - 2013-01-29-0bc317aa4d9af44806c28ca286d79a8b5a92b2b8
kernel - 3.3.8-1-fa7af62e0ef1d529ecb7f7efccc706c3
kmod-3c59x - 3.3.8-1
kmod-8139too - 3.3.8-1
kmod-e100 - 3.3.8-1
kmod-e1000 - 3.3.8-1
kmod-ipt-conntrack - 3.3.8-1
kmod-ipt-core - 3.3.8-1
kmod-ipt-nat - 3.3.8-1
kmod-ipt-nathelper - 3.3.8-1
kmod-lib-crc-ccitt - 3.3.8-1
kmod-libphy - 3.3.8-1
kmod-natsemi - 3.3.8-1
kmod-ne2k-pci - 3.3.8-1
kmod-pcnet32 - 3.3.8-1
kmod-ppp - 3.3.8-1
kmod-pppoe - 3.3.8-1
kmod-pppox - 3.3.8-1
kmod-r8169 - 3.3.8-1
kmod-sis900 - 3.3.8-1
kmod-tg3 - 3.3.8-1
kmod-via-rhine - 3.3.8-1
kmod-via-velocity - 3.3.8-1
[...]

PR #4963, opened 2026-06-07
fix(dpkg): extract License field for opkg/ipkg entries

### LINKED ISSUE #4940: Syft does not extract package licenses from opkg manager
**What happened**:
OpenWRT prior to version 25.12 was based on the OpenWRT package manager (opkg). This is currently handled by the dpkg-db-cataloger and it's successful in extracting the package name and versions. However, it fails to parse and extract the package licenses from the *.control files.

The licenses element should be populated with the license information from the associated .control files in opkg package info folder. However, the licenses list in syft json results is empty for packages discovered by the dpkg-db-cataloger

1- Download and unzip an OpenWRT firmware image prior to 25.12:
 wget https://archive.openwrt.org/releases/23.05.4/targets/x86/64/openwrt-23.05.4-x86-64-generic-ext4-combined.img.gz
unzip openwrt-23.05.4-x86-64-generic-ext4-combined.img.gz
2- mount the image:
sudo losetup --show -Pf openwrt-23.05.4-x86-64-generic-ext4-combined.img
[will show you /dev/loopxx]
soudo mount /dev/loopxxp2 /mnt/openwrt
3- perform syft dir scan
syft scan dir:/mnt/openwrt/ -o:syft-json=/tmp/openwrt.json
4- inspect the syft json result file and notice the licenses list for every package detected by the dpkg-db-cataloger is empty

**Environment**:
- Output of `syft version`: 
Application:   syft
Version:       1.44.0-SNAPSHOT-b1287d45
BuildDate:     2026-05-20T09:35:34Z
GitCommit:     b1287d45d81d8624820c4666992071b018d43d87
GitDescription: v1.44.0-27-gb1287d45
Platform:      linux/amd64
GoVersion:     go1.26.0
Compiler:      gc
SchemaVersion: 16.1.3

Dashtid:
Closes #4940.

## Background

opkg and ipkg are routed through the dpkg cataloger (see `capabilities.yaml`: the dpkg-db-cataloger already accepts `**/lib/opkg/info/*.control` and `**/lib/opkg/status`). Unlike Debian dpkg, where licenses live in copyright files under `/usr/share/doc/PKG/copyright`, opkg/ipkg derivatives declare the package license inline in the status DB via a `License:` field. The cataloger silently dropped that field at mapstructure decode time, so every opkg-managed package reported `"licenses": []`.

## Repro

Minimal fixture at `syft/pkg/cataloger/debian/testdata/var/lib/opkg/status` (committed in this PR):

```
Package: dropbear
Version: 2024.85-r0
License: MIT
...

Package: busybox
License: GPL-2.0-only
...

Package: kernel-modules
License: GPL-2.0 BSD-3-Clause
...
```

Scanning with current `main` yields empty `licenses` for all three; the `License` key never makes it into the package object. Dumping the raw extracted-fields map confirms it's discarded at the struct-decode step.

## Diagnosis & fix

Three small changes carry the field end-to-end:

1. `syft/pkg/cataloger/debian/parse_dpkg_db.go` — add `License string \`mapstructure:"License"\`` to the intermediate `dpkgExtractedMetadata` struct and to the `pkg.DpkgDBEntry` literal it builds.
2. `syft/pkg/dpkg.go` — add `License string \`json:"license,omitempty"\`` to the public `DpkgDBEntry`. With `omitempty` this stays absent in the SBOM for standard dpkg (where the field is never set), so consumers of unrelated dpkg-based scans see no shape change.
3. `syft/pkg/cataloger/debian/package.go` — replace the empty `var licenses []pkg.License` initializer (which had a `TODO: separate pr to license refactor` comment) with `extractDeclaredLicenses`, a small helper that mirrors the alpine cataloger's pattern:
   - try `license.ParseExpression(raw)` first → keep valid SPDX expressions whole (e.g. `Apache-2.0 OR MIT`)
   - fall back to `strings.Fields(raw)` for space-separated lists (e.g. `GPL-2.0 BSD-3-Clause`)
   - returns `nil` for empty input, so standard dpkg incurs no allocation
   - existing `addLicenses(ctx, resolver, dbLocation, &p)` copyright-file lookup is unchanged and still runs after — both paths compose

This is a one-way extension: real Debian dpkg `status` files don't carry a `License` field per Debian policy, so the new code path is a no-op for them.

## Design choices

I picked sensible defaults and flagged them here in case you'd prefer otherwise:

- **`License string` vs `Licenses []string` on `DpkgDBEntry`**: went with `License string` to mirror alpine's `ApkDBEntry.License` raw passthrough. Parsing happens in `newDpkgPackage`. Happy to flip to `Licenses []string` and parse at the entry level if you'd prefer that shape.
- **Separator convention**: alpine uses whitespace, so I followed suit. Some OpenWrt/Yocto feeds I've seen also use `,` or `|` — I left those out for now since adding multi-separator support without a real-world fixture felt speculative. Easy follow-up if you want it.
- **Schema version**: bumped `JSONSchemaVersion` to `16.1.4` (ADDITION — adds an optional field, no breaking change). Added `schema-16.1.4.json` and updated `schema-latest.json`. Existing snapshot tests required ID updates because `pkg.Package.SetID()` hashes the metadata struct; the empty `License` field shifts the hash for all dpkg entries even though it doesn't appear in the JSON output.
- **No separate opkg cataloger**: I'm aware of the 2024 thread suggesting opkg could live in its own cataloger. My read of the more recent direction (kzantow's Slack guidance + #1847 / #3888) is that dpkg-handles-opkg is the endorsed arrangement. Happy to be corrected if that's stale.

## Tests

- `Test_parseDpkgStatus/opkg_status_with_license_field` — new fixture, end-to-end parser test covering single-license, single-license-with-version-suffix, and space-separated multi-license cases.
- `Test_extractDeclaredLicenses` — focused unit test on the SPDX-whole vs split logic.

Existing dpkg parser tests are unaffected.


wagoodman (commented):
I've pushed a change that pivots the license to be persisted to the internal parser-related struct over the top-level API `pkg.DpkgDBEntry` struct, which simplifies the PR and is consistent with how other catalogers raise up license information.

wagoodman (approved):

PR #4966, opened 2026-06-10
Relativize CycloneDX-JSON file component paths against --base-path

### LINKED ISSUE #4592: CycloneDX JSON output keeps absolute paths even if `--base-path` is set
**What happened**:

Setting `--base-path` to the current working directory did not convert absolute paths to relative ones in the CycloneDX JSON output.
Syft was set to generate a SBOM for packages found in the current working directory, so `base-path` should have matched.
Since SPDX JSON correctly converts to relative paths as far as I could see, I assume this is specific to CycloneDX.

Absolute paths should be converted to relative ones with respect to `base-path`.

Should work with any project, but for simplicity use the syft repository itself.

1. Run `SYFT_FORMAT_CYCLONEDX_JSON_PRETTY=true syft --base-path=$(pwd) -o cyclonedx-json=cdx_sbom.json ./`
2. Open `cdx_sbom.json`
3. Search for absolute paths being mentioned in the `name` fields of components.

SPDX JSON output correctly converts absolute paths to relative ones.

**Environment**:
- Output of `syft version`: Tested with v1.41.1 and v1.41.0
- OS (e.g: `cat /etc/os-release` or similar): macOS, but I assume it is OS independent

AgentGymLeader:
## Description

CycloneDX-JSON file components emitted absolute host paths even when `--base-path` was set, while SPDX-JSON correctly emitted base-relative paths. The CycloneDX encoder set the file component `Name` from `metadata.Path` (the raw, non-chrooted indexer path), so the host prefix was never stripped.

This makes CycloneDX match SPDX. SPDX's `toFiles` relativizes each file via `convertAbsoluteToRelative(coordinate.RealPath)`. I lifted that helper into `syft/format/internal.ConvertAbsoluteToRelative` so both encoders share one implementation (SPDX now delegates to it, with no behavior change) and used it for the CycloneDX file component name with the same error → fall-back-to-`RealPath` behavior.

@kzantow — thanks for the pointer on the issue. I went with the existing SPDX `convertAbsoluteToRelative(RealPath)` path rather than `AccessPath`, to keep the two encoders consistent and reuse the logic that already ships. Happy to switch the shared helper to `AccessPath` if you'd prefer that for the symlink-dedup cases you mentioned — since it's now shared, that would move both formats together.

**Before / After** (a file's CycloneDX `components[].name`, scanned with `--base-path`)

- Before: `/abs/scan/root/usr/bin/foo`
- After: `usr/bin/foo` (matches SPDX-JSON)

## Type of change

## Checklist

## Issue references

Fixes #4592

### DIFF
--- a/syft/format/common/cyclonedxhelpers/to_format_model.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model.go
@@ -78,10 +78,7 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 		}
 
 		cdxHashes := digestsToHashes(digests)
-		relativePath, err := formatinternal.ConvertAbsoluteToRelative(coordinate.RealPath)
-		if err != nil {
-			relativePath = coordinate.RealPath
-		}
+		relativePath := fileRelativePath(s.Source, coordinate.RealPath)
 		components = append(components, cyclonedx.Component{
 			BOMRef: string(coordinate.ID()),
 			Type:   cyclonedx.ComponentTypeFile,
@@ -128,6 +125,18 @@ func digestsToHashes(digests []file.Digest) []cyclonedx.Hash {
 		})
 	}
 	return hashes
+}
+
+// fileRelativePath returns the path for a file component in the BOM.
+// When the source is a directory scan with a --base-path set, paths are made
+// relative to that base (supporting ".." for symlinks that escape the base).
+// For all other source types the absolute path is stripped to a simple relative
+// path via ConvertAbsoluteToRelative.
+func fileRelativePath(src source.Description, realPath string) string {
+	if m, ok := src.Metadata.(source.DirectoryMetadata); ok && m.Base != "" {
+		return formatinternal.Rel(m.Base, realPath)
+	}
+	return formatinternal.ConvertAbsoluteToRelative(realPath)
 }
 
 func toOSComponent(distro *linux.Release) []cyclonedx.Component {
--- a/syft/format/common/cyclonedxhelpers/to_format_model.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model.go
@@ -14,6 +14,7 @@ import (
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/cpe"
 	"github.com/anchore/syft/syft/file"
+	formatinternal "github.com/anchore/syft/syft/format/internal"
 	"github.com/anchore/syft/syft/format/internal/cyclonedxutil/helpers"
 	"github.com/anchore/syft/syft/linux"
 	"github.com/anchore/syft/syft/pkg"
@@ -55,7 +56,6 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 	artifacts := s.Artifacts
 
 	for _, coordinate := range coordinates {
-		var metadata *file.Metadata
 		// File Info
 		fileMetadata, exists := artifacts.FileMetadata[coordinate]
 		// no file metadata then don't include in SBOM
@@ -70,7 +70,6 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 			// skip dir, symlinks and sockets for the final bom
 			continue
 		}
-		metadata = &fileMetadata
 
 		// Digests
 		var digests []file.Digest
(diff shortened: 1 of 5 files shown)


syft/format/internal/relativepath.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func ConvertAbsoluteToRelative(absPath string) (string, error) {
+	if !path.IsAbs(absPath) {
+		return absPath, nil
+	}
+
+	relPath, found := strings.CutPrefix(absPath, "/")
+	if !found {
+		return "", fmt.Errorf("error calculating relative path: %s", absPath)
```
kzantow: This function returns an error converting a relative path to a relative path?

syft/format/common/cyclonedxhelpers/to_format_model.go
```diff
@@ -79,10 +78,14 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 		}
 
 		cdxHashes := digestsToHashes(digests)
+		relativePath, err := formatinternal.ConvertAbsoluteToRelative(coordinate.RealPath)
```
kzantow: As was noted on the issue, converting the `RealPath` to a relative path in this manner seems like the wrong thing to do: if I scanned `/subdir` and there's a symlink to `../foo`, this would be resolved to the absolute path of `/foo`, which is a `RealPath`, which then would be converted to the relative path `foo`, but there is no `/subdir/foo`. As noted in another comment, this should probably be `../foo`

AgentGymLeader:
@kzantow the one thing still open is whether the shared helper should use AccessPath or stay on RealPath. Both encoders run through it now, so I can move them together either way. Let me know which you'd prefer and I'll push the update.

syft/format/common/spdxhelpers/to_format_model.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/anchore/syft/internal/spdxlicense"
 	"github.com/anchore/syft/syft/artifact"
 	"github.com/anchore/syft/syft/file"
-	formatInternal "github.com/anchore/syft/syft/format/internal"
+	formatinternal "github.com/anchore/syft/syft/format/internal"
```
kzantow, line 23: nit: can we avoid these import renames?

syft/format/internal/relativepath.go
```diff
@@ -0,0 +1,20 @@
+package internal
+
+import (
+	"fmt"
+	"path"
+	"strings"
+)
+
+func ConvertAbsoluteToRelative(absPath string) (string, error) {
```
kzantow: I think this should be something like what [filepath.Rel](https://pkg.go.dev/path/filepath#Rel) is doing, though I don't think we can use that function because the paths here are normalized to forwards slashes

AgentGymLeader:
@kzantow thanks, addressed in 6f8c004c:

- **Relative path (`../foo`):** you were right. Prefix-stripping collapsed a symlink whose real path escapes the base down to `foo`. I added a forward-slash equivalent of `filepath.Rel` that preserves `..`, and take the base from `DirectoryMetadata.Base` (the `--base-path`). Non-directory sources keep the existing behavior.
- **Error return:** agreed it's odd to return an error converting a relative path to a relative path, so the helper now returns a plain string with no error.
- **Import rename:** lined the alias up as `formatinternal` to match cyclonedxhelpers. Happy to drop the alias entirely and reference it directly if you'd prefer.
- Added tests for the escape case (`/root/subdir` → `/root/foo` = `../foo`) and the other path shapes.

syft/format/internal/relativepath.go
AgentGymLeader: Good point — dropped the error return; `Rel` now just returns the string. Fixed in 6f8c004c.

syft/format/common/cyclonedxhelpers/to_format_model.go
AgentGymLeader: Agreed — 6f8c004c preserves the `..` (relative to the scan base), so an escaping symlink stays `../foo` instead of collapsing to `foo`. Added an escape regression test.

syft/format/common/spdxhelpers/to_format_model.go
AgentGymLeader: The alias is on `.../format/internal` (package name `internal`), so it needs some name to read clearly — settled on a single `formatinternal` everywhere in 6f8c004c. Happy to use a different name if you'd prefer.

syft/format/internal/relativepath.go
AgentGymLeader: Right — can't use `filepath.Rel` directly since these are normalized to forward slashes, so 6f8c004c adds a forward-slash equivalent (`Rel`) with regression tests.

PR #4971, opened 2026-06-11
fixed .net deps.json cataloger shows phantom pkgs , fixed #4970

### LINKED ISSUE #4970: .net deps.json cataloger shows phantom pkgs for reference assembly library entries
**What happened**:

when an asp.net core app is published with <preservecompilationcontext>true</preservecompilationcontext> its deps.json file contains entries like

```
  "Microsoft.AspNetCore.DataProtection/10.0.0.0": {
    "type": "referenceassembly",
    "serviceable": false,                                                                                                                                               
    "sha512": ""
  }
```

these are compile-time reference assemblies, not actual nuget pkgs

**Environment**:

rezmoss:
## Description

fixed .net deps.json cataloger shows phantom pkgs , fixed #4970

## Type of change

## Checklist

### DIFF
--- a/syft/pkg/cataloger/dotnet/deps_binary_cataloger.go
+++ b/syft/pkg/cataloger/dotnet/deps_binary_cataloger.go
@@ -292,6 +292,14 @@ func packagesFromLogicalDepsJSON(doc logicalDepsJSON, config CatalogerConfig) (*
 		}
 		lp := doc.PackagesByNameVersion[nameVersion]
 
+		if lp.Library != nil &&
+			lp.Library.Type == "referenceassembly" &&
+			lp.Library.Sha512 == "" &&
+			lp.Library.Path == "" {
+			skippedDepPkgs[nameVersion] = lp
+			continue
+		}
+
 		if config.ExcludeProjectReferences && lp.Library != nil && lp.Library.Type == "project" {
 			skippedDepPkgs[nameVersion] = lp
 			continue
--- a/syft/pkg/cataloger/dotnet/deps_binary_cataloger_test.go
+++ b/syft/pkg/cataloger/dotnet/deps_binary_cataloger_test.go
@@ -1,0 +1,97 @@
+package dotnet
+
+import (
+	"encoding/json"
+	"os"
+	"testing"
+
+	"github.com/scylladb/go-set/strset"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/anchore/syft/syft/file"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+func Test_packagesFromLogicalDepsJSON_skipsReferenceAssemblies(t *testing.T) {
+	mkPkg := func(name, version, libType, sha, path string) logicalDepsJSONPackage {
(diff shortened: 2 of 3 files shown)


rezmoss:
before. 
wrong-ver-detection

after
correct-ver

spiffcs:
@rezmoss I read your issue and am a little confused as to why the versions changed (10.0.826... -> 10.0.8) in your example? I expected the `AFTER` block to have less packages like your new test asserts:

```
assert.NotContains(t, names, "Microsoft.AspNetCore.DataProtection")
assert.NotContains(t, names, "Microsoft.AspNetCore.DataProtection.Abstractions")
```

Is the version change in your screenshot a regression? 
Can you share the command you ran that generated that screenshot?

rezmoss:
@spiffcs let me clarify, this is not a ver rewrite, its a phantom de-dup, when an aspnet core app is published with preservecompilationcontext=true, every shared framework assembly shows up twice in the sbom under the same name:
1) a phantom entry from the deps.json referenceassembly library, with an empty sha512 and no path, using the assembly version from the deps.json key, which is 10.0.826.23019 (e.g)

2) the real package backed by the actual reference assembly under /app/refs/…dll, whose version comes from the PE productVersion 10.0.8.

`syft <image> -o syft-json | jq '.artifacts[] | select(.name|test("DataProtection")) | {name,version,metadataType,type:.metadata.type,locations:[.locations[].path]}'`


Before patch

Screenshot 2026-06-15 at 11 32 47

After 

Screenshot 2026-06-15 at 11 33 01


and docker img

```
➜  dotnet-refasm-test find . -type f -exec sh -c 'for f; do echo "$(basename "$f")"; echo "-------"; cat "$f"; echo; done' _ {} +
MyApp.csproj
-------
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
    <PreserveCompilationContext>true</PreserveCompilationContext>
    <PreserveCompilationReferences>true</PreserveCompilationReferences>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
  </ItemGroup>
</Project>

Dockerfile
-------
FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build
WORKDIR /src
COPY MyApp.csproj .
RUN dotnet restore
COPY Program.cs .
RUN dotnet publish -c Release -o /app --no-restore

FROM mcr.microsoft.com/dotnet/aspnet:8.0
WORKDIR /app
COPY --from=build /app .
ENTRYPOINT ["dotnet", "MyApp.dll"]

Program.cs
-------
using Microsoft.AspNetCore.DataProtection;
using Newtonsoft.Json;

var builder = WebApplication.CreateBuilder(args);
builder.Services.AddDataProtection();
var app = builder.Build();
app.MapGet("/", () => JsonConvert.SerializeObject(new { ok = true }));
app.Run();

➜  dotnet-refasm-test 

```

spiffcs (approved):

PR #4977, opened 2026-06-12
Add pkg.Architecture interface

willmurphyscode:
Add a pkg.Architecture interface that exposes a TargetArchitecture method. (The method cannot be called Architecture because package metadata types already use that name.) Modeled on the FileOwner interface, this gives clients of Syft a simple way to ask, "is this package one that can specify CPU architecture and if so what architecture does it target?".

## Type of change

## Checklist

### DIFF
--- a/syft/pkg/architecture.go
+++ b/syft/pkg/architecture.go
@@ -1,0 +1,16 @@
+package pkg
+
+// Architecture is the interface that wraps the TargetArchitecture method.
+//
+// TargetArchitecture returns the target CPU architecture a piece of package Metadata records,
+// in its ecosystem's native spelling: rpm and apk use the GNU/kernel form ("x86_64",
+// "aarch64", "noarch"), while dpkg uses the Debian form ("amd64", "arm64", "all"). It lets
+// consumers read a package's architecture directly from metadata rather than parsing the
+// PURL "arch" qualifier.
+//
+// The method is named TargetArchitecture rather than Architecture (or Arch) because the dpkg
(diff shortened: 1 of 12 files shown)


syft/pkg/architecture.go
```diff
@@ -0,0 +1,16 @@
...
+// in its ecosystem's native spelling: rpm and apk use the GNU/kernel form ("x86_64",
+// "aarch64", "noarch"), while dpkg uses the Debian form ("amd64", "arm64", "all"). It lets
+// consumers read a package's architecture directly from metadata rather than parsing the
+// PURL "arch" qualifier.
+//
+// The method is named TargetArchitecture rather than Architecture (or Arch) because the dpkg
+// and apk metadata already expose an Architecture field, and rpm an Arch field; Go does not
+// permit a method to share a name with a field on the same type.
+type Architecture interface {
+	TargetArchitecture() string
```
kzantow, line 15: What about things like macOS universal binaries; should this be a slice? Should we surface multiple otherwise duplicate packages in some cases like that where a binary contains multiple architectures of the same thing?

willmurphyscode:
I think now is not the time for this. There are some ambiguities we have to solve:

1. What about multiarch "fat" mach-o binaries?
2. What about declared dependencies that can install on some but not all architectures?
3. What about noarch and src dependencies?

In other words, saying, "this package has this architecture" is a more complicated and lossier statement than say, "this package owns these paths", and therefore a worse fit for having a general interface.

PR #4978, opened 2026-06-12
fix: use printf instead of echo to fix ANSI color output

Jouini-Mohamed-Chaker:
## Description

This PR fixes an issue where the `install.sh` script leaks raw ANSI escape codes (e.g., `\033[0;31m`) into the terminal output when executed via strict POSIX `sh` (such as `dash`). 

The `echo` command does not natively evaluate ANSI color codes in these environments. This update refactors the `echo_stderr` and `log_tag` functions to use `printf "%b"` instead of `echo`. This ensures cross-shell compatibility and renders the terminal colors correctly regardless of the shell being used.

**CLI Output Changes:**

**Before:**
Screenshot 2026-06-12 201015

**After:**
Screenshot 2026-06-12 201842

## Type of change

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -35,7 +35,8 @@ is_command() (
 )
 
 echo_stderr() (
-  echo "$@" 1>&2
+  IFS=" "
+  printf "%b\n" "$*" 1>&2
 )
 
 _logp=2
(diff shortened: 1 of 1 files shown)


wagoodman (approved):
💯  -- force pushed only for the rebase

PR #4979, opened 2026-06-13
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
CPE parsing fuzz tests. Syft SBOM generator.

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -1,0 +1,48 @@
+//go:build go1.18
+// +build go1.18
+
+// Copyright 2026 Anchore, Inc.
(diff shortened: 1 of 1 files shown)


spiffcs:
@canolgun I'm going to echo https://github.com/google/oss-fuzz/pull/15725#issuecomment-4699886800 here and close this PR. The description, additions via OSS FUZZ, and onboarding do not seem to be followed correctly and there is very little context/content to pick up on how this shoud/could work. 

I'll take a look when I have more time on how to correctly integrate with google/oss-fuzz infrastructure.

PR #4983, opened 2026-06-16
fix(go-module-cataloger): use correct error variable in package load 

idhyt:
## Description

In the loop over `rootPkgs`, the error appended to `unknownErr` was incorrectly using the outer `err` (from `packages.Load`) instead of the individual package error `e`. This could cause:
- nil or stale error values to be recorded when package-specific errors existed.
- Loss of detailed per-package error information in the final error report.

Changed `unknown.Append(..., err)` to `unknown.Append(..., e)` to properly capture each package's load error.

## Checklist


wagoodman (approved):

idhyt:
@wagoodman 
I noticed that the CI run for this PR has failed. Given the nature of the changes, this should not typically affect the test pipeline. When you have a moment, could you please take a look?

PR #4985, opened 2026-06-16
feat(elixir): emit dependency relationships from mix.lock

cgreeno:
## TLDR 

We are missing the ability to distinguish transitive/direct dependencies when running SBOM on our Elixir codebase. Followed the structure I found in other parts of the code base, going to add this for Elixir and Ruby. Link the ruby PR as well when I am done it. 

## Description

The Elixir cataloger catalogued packages from `mix.lock` but discarded each entry's dependency list, so it emitted **no relationships**. Every `mix.lock` entry already encodes its own dependencies (the 6th element of the entry tuple, e.g. `{:cowlib, ...}, {:ranch, ...}`), so the dependency graph is derivable from data already in the file... no new inputs required.

This change emits `dependency-of` relationships between locked packages, using the same shared `dependency.Processor` / `Specifier` mechanism that the alpine, arch, debian, redhat, and python catalogers already use to express their dependency graphs. It is purely additive: a new optional metadata field plus relationships; package cataloging itself is unchanged.

**Changeset:**

| File | Change |
|---|---|
| `syft/pkg/elixir.go` | add `Dependencies []string` to `ElixirMixLockEntry` |
| `syft/pkg/cataloger/elixir/parse_mix_lock.go` | extract dependency names from each entry's dependency list (the entry's own source tuple is skipped) |
| `syft/pkg/cataloger/elixir/dependency.go` *(new)* | `mixLockDependencySpecifier` declaring `Provides=[name]`, `Requires=[deps]` |
| `syft/pkg/cataloger/elixir/cataloger.go` | register `dependency.Processor(mixLockDependencySpecifier)` |
| `internal/constants.go` | JSON schema `16.1.4 → 16.1.5` |
| `schema/json/schema-16.1.5.json` *(new)* + `schema-latest.json` | regenerated |
| `dependency_test.go` *(new)*, `parse_mix_lock_test.go`, `cataloger_test.go` + fixture | unit tests for extraction and the specifier, plus a cataloger-level test asserting the resolved relationships |

**Schema:** bumped `16.1.4 → 16.1.5` additive only (a new optional `dependencies` field on `ElixirMixLockEntry`), compatible with all historical data per the SchemaVer `ADDITION` rule.

**Type of change:** New feature (non-breaking)  adds relationship data without altering existing package output.


cgreeno:
⬆️  PR applying the same approach to Ruby `Gemfile.lock`: #4986

cgreeno:
@wagoodman - you appear to be the main owner? Sorry if I got the wrong end of the stick. --> Same approach on this PR https://github.com/anchore/syft/pull/4986

cgreeno:
Heya sorry to bother you @spiffcs but do you think you could have a look, any feedback is more than welcome :)

spiffcs (approved):

PR #4987, opened 2026-06-16
fix(purl-backfill): respect arch qualifier

willmurphyscode:
Previously, when constructing rpm, alpm, and apk metadata struct from a PURL, Syft would ignore the arch qualifier. Start respecting that qualifier.

## Type of change

I say it's a bug because I sort of thought we already did this.

## Checklist

### DIFF
--- a/syft/format/internal/backfill.go
+++ b/syft/format/internal/backfill.go
@@ -87,7 +87,7 @@ func backfillFromPurl(p *pkg.Package) {
 		setJavaMetadataFromPurl(p, purl)
 	}
 
-	setMetadataArchFromPurl(p, rpmmod, arch)
+	setDistroMetadata(p, rpmmod, arch)
 
 	for _, c := range cpes {
 		if slices.Contains(p.CPEs, c) {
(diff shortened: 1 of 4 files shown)


syft/format/internal/backfill.go
```diff
@@ -96,6 +97,19 @@ func backfillFromPurl(p *pkg.Package) {
 	}
 }
 
+func setMetadataArchFromPurl(p *pkg.Package, rpmmod, arch string) {
```
kzantow: nit: maybe `setDistroMetadata` or similar would be a more pertinent name, since this is doing more than just arch

syft/format/internal/backfill.go
```diff
@@ -107,32 +121,109 @@ func setJavaMetadataFromPurl(p *pkg.Package, _ packageurl.PackageURL) {
...
+func setDpkgMetadataFromPurl(p *pkg.Package, arch string) {
+	if p.Type != pkg.DebPkg {
+		return
+	}
+	if arch == "" {
+		return
+	}
+
+	if p.Metadata == nil {
+		p.Metadata = pkg.DpkgDBEntry{Architecture: arch}
```
kzantow: I think the pattern here (similarly across all these type of functions) should be:
```
if p.Metadata == nil {
  p.Metadata = pkg.MetadataType{}
}

// set all known fields
```
to avoid two code paths doing the same thing

kzantow (approved):

PR #4988, opened 2026-06-17
feat: Capture golang binary symbols

spiffcs:
## Summary

Adds optional extraction of function symbols from Go binaries cataloging. Symbol extraction attributes them to the module that owns them which come out in the SBOM as syft packages. When enabled, syft reads the binary's pclntab and records the function symbols compiled into the binary.

Symbols are attributed to their owning module and on that package's metadata. Standard-library symbols are attached to the synthetic stdlib package. Inlined functions are also recovered from the pclntab funcname table so a vulnerable but inlined function can still be reported.

### Changes

- New symbols.go:parsing for ELF and Mach-O, symbol→module attribution, stdlib detection
- `GolangBinaryBuildinfoEntry.Symbols` (syft/pkg/golang.go): new symbols JSON field (omitempty).
- the cataloger accumulates stdlib symbols per binary location
- Schema bump: 16.1.8 → 16.1.9.

### Configuration: default off

capture-symbols is a three-value scope enum. Disabled by default, so upgrading does not change existing Go SBOM output or increase its size unless explicitly enabled.

```
┌──────────────────────┬──────────────────────────────────────────────────────────────────────────────┐
│       Surface        │                                   Setting                                    │
├──────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ Environment variable │ SYFT_GOLANG_CAPTURE_SYMBOLS=stdlib                                                           │
├──────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ Config file          │ golang.capture-symbols: none                                               │
├──────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ Library              │ golang.DefaultCatalogerConfig().WithCaptureSymbols(golang.SymbolScopeStdlib) │
└──────────────────────┴──────────────────────────────────────────────────────────────────────────────┘
```
```
┌────────────────┬──────────────────────────────────────────────┐
│     Value      │                   Behavior                   │
├────────────────┼──────────────────────────────────────────────┤
│ none (default) │ no symbols captured                          │
├────────────────┼──────────────────────────────────────────────┤
│ stdlib         │ symbols only on the synthetic stdlib package │
├────────────────┼──────────────────────────────────────────────┤
│ all            │ symbols on all module packages plus stdlib   │
└────────────────┴──────────────────────────────────────────────┘
```


syft/pkg/vcpkg.go
wagoodman, line 1: I was getting warnings when regenerating the json schema, so I went ahead and added descriptions for these.

syft/cataloging/symbols.go
```diff
@@ -0,0 +1,31 @@
...
+func (s SymbolScope) Parse() SymbolScope {
+	switch strings.ToLower(strings.TrimSpace(string(s))) {
+	case string(SymbolScopeAll):
+		return SymbolScopeAll
+	case string(SymbolScopeStdlib):
+		return SymbolScopeStdlib
+	case string(SymbolScopeNone), "":
+		return SymbolScopeNone
+	}
+	return ""
```
wagoodman: see the compliance parsing for an example; we tend to fail towards defaults instead of require checking and error by the caller. So when we use it in syft/pkg/cataloger/golang/parse_go_binary, we should trust the value at that point (not call Parse() again)

syft/pkg/cataloger/golang/parse_go_binary.go
```diff
@@ -49,15 +51,45 @@ const devel = "(devel)"
...
+	// synthetic "stdlib" package. Guarded by stdlibSymbolsMu because parsers run concurrently.
+	stdlibSymbols   map[file.Coordinates][]string
+	stdlibSymbolsMu sync.Mutex
 }
 
 func newGoBinaryCataloger(opts CatalogerConfig) *goBinaryCataloger {
 	return &goBinaryCataloger{
 		licenseResolver:   newGoLicenseResolver(binaryCatalogerName, opts),
 		mainModuleVersion: opts.MainModuleVersion,
+		symbolScope:       opts.CaptureSymbols.Parse(),
```
wagoodman: no need to call Parse() again

syft/pkg/cataloger/golang/symbols.go
```diff
@@ -0,0 +1,281 @@
...
+	name string
+}
+
+// getSymbols extracts all function symbols from the pclntab of a go binary. The pclntab is required by the
+// go runtime (for panic tracebacks and GC), so it is present even in binaries built with -ldflags="-s -w".
+func getSymbols(r io.ReaderAt) (syms []binarySymbol, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			// the gosym package can panic on malformed pclntab data
+			err = fmt.Errorf("recovered from panic while reading pclntab: %v", r)
```
wagoodman, line 35: we should clear `syms` at this point too

syft/pkg/cataloger/golang/symbols.go
```diff
@@ -0,0 +1,281 @@
...
+	return name[:slash+1+dot]
+}
+
+// isCompilerGeneratedName reports whether a symbol name was synthesized by the compiler or linker rather
+// than declared in Go source. These names use ':' or '..' (e.g. "type:.eq.*", "type..hash.*",
+// "go:string.*") — byte sequences that never appear in a real Go import path or identifier — so they
+// belong to no package and are dropped rather than mis-attributed (e.g. bucketed under a bogus "type"
+// stdlib package).
+func isCompilerGeneratedName(name string) bool {
+	return strings.Contains(name, ":") || strings.Contains(name, "..")
```
wagoodman: I think this isn't accounting for the `go.type` case that I think older compiler versions have

syft/pkg/cataloger/golang/symbols.go
```diff
@@ -0,0 +1,281 @@
...
+
+// packagePathFromSymbolName derives the owning package import path from a fully qualified symbol name.
+// The package path is everything up to the first "." that follows the final "/" — e.g.
+// "path/filepath.IsLocal" -> "path/filepath" and "golang.org/x/net/html.(*Tokenizer).Next" ->
+// "golang.org/x/net/html". Returns "" when the name has no package-qualifying dot or is compiler-generated.
+func packagePathFromSymbolName(name string) string {
+	if isCompilerGeneratedName(name) {
+		return ""
+	}
+	slash := strings.LastIndex(name, "/")
```
wagoodman, line 89: I think this doesn't account for cases when there are type arguments like ` foo/bar.Do[net/url.Values]` ... we'd find that the package is `foo/bar.Do[net`

syft/pkg/cataloger/golang/stdlib_package.go
wagoodman: I realize this is outside of review, but noticed it in the review locally... I think this is swapped 
```suggestion
	after, _, found := strings.Cut(version, "+")
```

syft/cataloging/symbols.go
spiffcs: Ahh nice call I totally missed that compliance pattern. The options layer's PostLoad just assigns the parsed value (no error path anymore). I think this is the same shape as complianceConfig.PostLoad() now.

syft/pkg/cataloger/golang/parse_go_binary.go
spiffcs: I updated this so the cataloger now takes opts.CaptureSymbols as-is and trusts that normalization happened at the options layer.

syft/pkg/cataloger/golang/symbols.go
spiffcs: 👍 I updated this so the recover block now sets syms = nil before setting the error. A panic mid-parse can't return partial results alongside the error.

syft/pkg/cataloger/golang/symbols.go
spiffcs: Yea I confirmed pre-1.20 toolchains used `.` where newer ones use `:`

I added prefix matches to account for the above. Real module paths like go.uber.org/zap and go.opentelemetry.io/otel also start with `go.` and shouldn't be dropped. I added test cases to cover this.

syft/pkg/cataloger/golang/stdlib_package.go
spiffcs: Nice catch! Fixed this apologies for the oversight.

syft/pkg/cataloger/golang/symbols.go
spiffcs: Right, I found that the last "/" search was landing inside the type arguments

`foo/bar.Do[net/url.Values]` derived `foo/bar.Do[net`. 

I fixed this by stripping the [...] portion before deriving the package path.
New test cases have been added to `symbols_test`

wagoodman (approved):

PR #5020, opened 2026-06-25
fix: conanfile comments in [requires] no longer skip dependencies

### LINKED ISSUE #5017: conanfile.txt: dependencies after a comment line in [requires] are not detected
**What happened**:

I was generating an SBOM from a source tree with a `conanfile.txt` and none of the dependencies showed up. Turns out that if there's a comment line (`#`) inside the `[requires]` section, every requirement after that comment gets dropped.

This is easy to hit because a lot of conanfile.txt files put a comment right under `[requires]` to group things. When that happens syft reports 0 conan packages, and there's no error or warning, so it's easy to miss.

Looking at `parse_conanfile.go`, `inRequirements` is set to false on any `#` line, and the only thing that turns it back on is another `[requires]` header:

```go
case strings.Contains(line, "[requires]"):
    inRequirements = true
case strings.ContainsAny(line, "[]") || strings.HasPrefix(strings.TrimSpace(line), "#"):
    inRequirements = false
```

Conan itself treats `#` as a comment and just skips the line (it doesn't end the section), so this is valid input.

Comments inside `[requires]` should be skipped, and the requirements after them should still be picked up.

`conanfile.txt`:
```
[requires]
# core dependencies
fmt/10.2.1
zlib/1.3.1
```

```
syft scan dir:. -o cyclonedx-json
```

No conan components are reported. If I remove the `# core dependencies` line, both `fmt` and `zlib` show up.

Looks like skipping `#` lines instead of clearing `inRequirements` would be enough to fix it.

**Environment**:
- Output of `syft version`: 
   Application: syft Version: 1.45.1 
   BuildDate: 2026-06-05T14:21:36Z 
   GitCommit: d4496b05aab2d489a5c2ec606755b657229b8925 
   GitDescription: v1.45.1 
   GoVersion: go1.26.3 
   Compiler: gc SchemaVersion: 16.1.3
- OS: Linux (arm64) — but this is OS-independent (directory scan)

jfjrh2014:
## Problem

Comments inside the `[requires]` section of `conanfile.txt` silently stopped dependency detection. Any package listed after a comment line was skipped.

From #5017: dependencies after a comment line in `[requires]` are not detected.

## Root Cause

The section-state switch OR'd the comment check (`strings.HasPrefix(..., "#")`) with the new-section check (`strings.ContainsAny(line, "[]")`). Both set `inRequirements = false`, so a comment inside `[requires]` acted like a section boundary.

## Fix

Skip blank lines and comments *before* evaluating section state, so they never toggle `inRequirements`. Section headers (`[generators]`, etc.) still end the requires section as before.

```
[requires]
# this comment used to break parsing
fmt/8.1.1    ← previously skipped
```

Fixes #5017

### DIFF
--- a/syft/pkg/cataloger/cpp/parse_conanfile.go
+++ b/syft/pkg/cataloger/cpp/parse_conanfile.go
@@ -31,15 +31,20 @@ func parseConanfile(_ context.Context, _ file.Resolver, _ *generic.Environment, 
 			return nil, nil, fmt.Errorf("failed to parse conanfile.txt file: %w", err)
 		}
 
+		trimmed := strings.TrimSpace(line)
+
+		// skip blank lines and comments without affecting section state
+		if trimmed == "" || strings.HasPrefix(trimmed, "#") {
+			continue
+		}
+
 		switch {
 		case strings.Contains(line, "[requires]"):
 			inRequirements = true
-		case strings.ContainsAny(line, "[]") || strings.HasPrefix(strings.TrimSpace(line), "#"):
+			continue
+		case strings.ContainsAny(line, "[]"):
 			inRequirements = false
-		}
-
-		m := pkg.ConanfileEntry{
-			Ref: strings.TrimSpace(line),
+			continue
 		}
 
 		if !inRequirements {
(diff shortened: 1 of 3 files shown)


spiffcs:
@jfjrh2014 Can you please follow the steps to sign off the commits in line with the DCO:
https://github.com/anchore/syft/pull/5020/checks?check_run_id=83418188042

jfjrh2014:
@spiffcs DCO is signed and passing now. Sorry for the delay. Ready for review when you have a moment.

wagoodman (approved):

PR #5021, opened 2026-06-26
fix: detect source archives by content, not filename substring (#4582)

### LINKED ISSUE #4582: SBOM generation failure when .tar.gz is part of the file name of a JSON
**What happened**:
If a JSON file contains .tar.gz in its file name, we are seeing an error of `unable to unarchive` while generating the SBOM. But this used to not happen in versions 1.36.0 and below. Was this a bug before and fixed in the newer versions or is it a bug introduced in newer versions?

Create a sample json file and name it something like `sample.tar.gz_hashed.json` and try generating the SBOM. 1.36.0 will successfully generate the SBOM and 1.38.0 and above will fail with an error

arpitjain099:
## Description

fileAnalysisPath calls IdentifyArchive with a nil reader, so mholt/archives can only match on the filename, and it does that with a substring check (strings.Contains lowercased for ".gz", ".tar", etc). That means any ordinary file whose name happens to contain an archive extension gets treated as an archive and the scan fails. The reported case is a JSON file named sample.tar.gz_hashed.json, but as @kzantow pointed out on #4582 even a real binary named syft.tar.gz.bin trips it. This worked fine in 1.36.0 and regressed sometime after.

The fix opens the file and hands the reader to IdentifyArchive so detection runs on content. Non-archives fail the header check, get no extractor, and fall through to normal file analysis; genuine archives still match on their header and extract as before.

One thing worth flagging: with content-based detection, a file named like an archive but not actually valid (including a genuinely corrupt .tar.gz) now falls through to plain file analysis instead of erroring. That matches pre-1.36 behavior and seems like the better default for a tool you point at arbitrary files, but it does mean a truly broken archive gets analyzed as a regular file rather than surfacing the corruption. If you'd rather keep an explicit error for that, or tighten the alias handling in internal/file/archive_aliases.go instead, happy to rework it - just say which way you'd prefer.

## Type of change

## Checklist

## Issue references

Fixes #4582

### DIFF
--- a/syft/source/filesource/file_source.go
+++ b/syft/source/filesource/file_source.go
@@ -212,7 +212,19 @@ func fileAnalysisPath(path string, skipExtractArchive bool) (string, func() erro
 		return analysisPath, cleanupFn, nil
 	}
 
-	envelopedUnarchiver, _, err := intFile.IdentifyArchive(context.Background(), path, nil)
+	// Pass the file contents to archive identification rather than relying on the
+	// filename alone. mholt/archives matches filenames with strings.Contains, so a
+	// regular file whose name merely contains an archive extension as a substring
+	// (e.g. "sample.tar.gz_hashed.json") is otherwise misidentified as an archive
+	// and fails to "unarchive". With the reader, content-based detection runs and
+	// non-archives fall through to regular file analysis. See issue #4582.
+	f, err := os.Open(path)
+	if err != nil {
+		return "", cleanupFn, fmt.Errorf("unable to open source file: %w", err)
+	}
+	defer f.Close()
+
+	envelopedUnarchiver, _, err := intFile.IdentifyArchive(context.Background(), path, f)
 	if unarchiver, ok := envelopedUnarchiver.(archives.Extractor); err == nil && ok {
 		analysisPath, cleanupFn, err = unarchiveToTmp(path, unarchiver)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


spiffcs (dismissed):

arpitjain099:
Fixed. The CLI failure was mine and it was broader than the one test.

Identifying the file by contents meant every jar matched, since a jar is a zip underneath. syft then extracted it to a temp dir, so the java cataloger never saw a jar and `scan` on `example-java-app-maven-0.1.0.jar` produced nothing, which is why the assertion saw an empty string. Measured directly on that fixture:

```
name only     -> extractor=false  err=no formats matched
with contents -> extractor=true   type=archives.Zip
```

So the fix is to keep identifying by name, which is what decides whether a path is treated as an archive, and read the contents only to reject a name match the bytes disagree with. Contents never promote a file to being an archive now. That still covers the original report, since `sample.tar.gz_hashed.json` matches the `.tar.gz` format by name (archives matches names with `strings.Contains`) and its bytes say otherwise, and it leaves everything that already worked alone. An unreadable file keeps the name-based answer rather than silently changing behavior.

`Test_RequestedPathIncludesSymlink` passes locally now, and `Test_fileAnalysisPath` covers all three cases: the misleading name, a jar, and a real `.tar.gz`.

Sorry for the noise on an already-approved PR.

PR #5029, opened 2026-07-02
Report each hardlink as its own file when scanning images

### LINKED ISSUE #5019: When scanning an image, syft only reports one file per set of hardlinks, leading to wrong SPDX packageVerificationCode
**What happened**:

We recently switched from scanning container images as actual container images to scanning the mounted container filesystem as a `dir:` (it appears to improve the problems reported in https://github.com/anchore/syft/issues/3800).

After this switch, syft produces more entries in the `files` array for SPDX (or more components with `type: file` for CycloneDX). For SPDX, this also results in different `packageVerificationCodeValue` for the exact same packages.

Same files reported when scanning the image as an image and when scanning it as a `dir:`, same `packageVerificationCode`s

```bash
# 1. scan an image that has packages that contain hardlinked files
$ syft scan registry.fedoraproject.org/fedora-minimal:43@sha256:d028a3d86b3b68081c6610d0e77fa08a0853bb9488f522d864192c84af5b1a44 \
    -o spdx-json@2.3 | jq >/tmp/fedora-minimal-image.json

# 2. scan the same image as a dir
# (if you don't have buildah, other ways to export the filesystem will work too)
$ buildah unshare -- bash -c '
set -x
container=$(buildah from registry.fedoraproject.org/fedora-minimal:43@sha256:d028a3d86b3b68081c6610d0e77fa08a0853bb9488f522d864192c84af5b1a44)
trap "buildah rm $container" EXIT
rootfs=$(buildah mount $container)
syft scan dir:$rootfs --override-default-catalogers=image -o spdx-json@2.3 | jq >/tmp/fedora-minimal-dir.json
'

# 3. see that the packageVerificationCode for e.g. glibc is different
$ jq -c < /tmp/fedora-minimal-image.json '.packages[] | select(.name == "glibc").packageVerificationCode'
{"packageVerificationCodeValue":"d601c6bf99eaa51af6783e6bd8706502c5390f0d"}

$ jq -c < /tmp/fedora-minimal-dir.json '.packages[] | select(.name == "glibc").packageVerificationCode'
{"packageVerificationCodeValue":"a3e94cf048079551f077c29a799c3522521a3acf"}

# 4. compare the files reported for glibc
$ get_glibc_files() {
    local sbom=$1
    jq -r '
        reduce .files[] as $f ({}; .[$f.SPDXID] = $f.fileName) as $files |
        (.packages[] | select(.name == "glibc")) as $glibc |
        .relationships[] |
        select(.spdxElementId == $glibc.SPDXID and .relationshipType == "CONTAINS") |
        $files[.relatedSpdxElement]
    ' "$sbom" | sort
}

$ diff -u2 <(get_glibc_files /tmp/fedora-minimal-image.json) <(get_glibc_files /tmp/fedora-minimal-dir.json)

--- /proc/self/fd/12	2026-06-25 10:49:51.706356089 +0200
+++ /proc/self/fd/13	2026-06-25 10:49:51.707356075 +0200
@@ -31,4 +31,6 @@
 usr/lib64/libutil.so.1
 usr/libexec/getconf/POSIX_V6_LP64_OFF64
+usr/libexec/getconf/POSIX_V7_LP64_OFF64
+usr/libexec/getconf/XBS5_LP64_OFF64
 usr/share/licenses/glibc/COPYING
 usr/share/licenses/glibc/COPYING.LIB

# 5. see that the files reported only for dir and not for image are hardlinks
$ podman run --rm -ti registry.fedoraproject.org/fedora-minimal:43@sha256:d028a3d86b3b68081c6610d0e77fa08a0853bb9488f522d864192c84af5b1a44 \
    bash -c 'stat /usr/libexec/getconf/* | grep -E "File:|Device:"'

  File: /usr/libexec/getconf/POSIX_V6_LP64_OFF64
Device: 0,111	Inode: 78554226    Links: 3
  File: /usr/libexec/getconf/POSIX_V7_LP64_OFF64
Device: 0,111	Inode: 78554226    Links: 3
  File: /usr/libexec/getconf/XBS5_LP64_OFF64
Device: 0,111	Inode: 78554226    Links: 3
```

The root cause seems to be that when scanning a container image, syft will only report one file per set of hardlinks pointing to the same file. When scanning a `dir:`, Syft reports all of the hardlinks. Because the [package verification code](https://spdx.github.io/spdx-spec/v2.3/package-information/#79-package-verification-code-field) is computed from all the files belonging to a package, this also changes the verification code value. The SPDX spec doesn't say anything about hardlinks, but the `dir:` approach of including hardlinks in the computation is likely more correct.

**Environment**:
- Output of `syft version`:
    ```yaml
    Application:   syft
    Version:       1.45.1
    BuildDate:     2026-06-05T14:21:36Z
    GitCommit:     d4496b05aab2d489a5c2ec606755b657229b8925
    GitDescription: v1.45.1
    Platform:      linux/amd64
    GoVersion:     go1.26.3
    Compiler:      gc
    SchemaVersion: 16.1.3
   ```
    ```env
    NAME="Fedora Linux"
    VERSION="43 (Workstation Edition)"
    RELEASE_TYPE=stable
    ID=fedora
    VERSION_ID=43
    VERSION_CODENAME=""
    PRETTY_NAME="Fedora Linux 43 (Workstation Edition)"
    ANSI_COLOR="0;38;2;60;110;180"
    LOGO=fedora-logo-icon
    CPE_NAME="cpe:/o:fedoraproject:fedora:43"
    DEFAULT_HOSTNAME="fedora"
    HOME_URL="https://fedoraproject.org/"
    DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f43/"
    SUPPORT_URL="https://ask.fedoraproject.org/"
    BUG_REPORT_URL="https://bugzilla.redhat.com/"
    REDHAT_BUGZILLA_PRODUCT="Fedora"
    REDHAT_BUGZILLA_PRODUCT_VERSION=43
    REDHAT_SUPPORT_PRODUCT="Fedora"
    REDHAT_SUPPORT_PRODUCT_VERSION=43
    SUPPORT_END=2026-12-02
    VARIANT="Workstation Edition"
    VARIANT_ID=workstation
    ```

wagoodman:
Image scans previously collapsed a set of hardlinks onto a single file, so only one path per inode showed up in results. Directory scans report every hardlink path, which made image vs dir SBOMs of the same filesystem diverge (and produce different SPDX `packageVerificationCode` values for packages that own hardlinked files).

Now both image resolvers (squash and all-layers) surface each hardlink at its own path as a regular file bound to the target's content, matching dir scans.

user-facing impact:
- SBOMs for images containing hardlinks will list more `file` entries
- SPDX `packageVerificationCode` values change for affected packages, now matching the equivalent `dir:` scan
- adds `file.NewVirtualLocationFromImage` to the public API

> [!NOTE]
> This does NOT address the `FilesByMIMEType` and other non-path based queries against the container based file resolvers. This means that though pathing/globbing will return multiple entries (for all og and hardlink types) others will only respond to regular files strictly.

## Type of change

## Checklist

## Issue references

fixes #5019

### DIFF
--- a/syft/internal/fileresolver/container_image_all_layers.go
+++ b/syft/internal/fileresolver/container_image_all_layers.go
@@ -53,8 +53,20 @@ func (r *ContainerImageAllLayers) HasPath(path string) bool {
 	return false
 }
 
-func (r *ContainerImageAllLayers) fileByRef(ref stereoscopeFile.Reference, uniqueFileIDs stereoscopeFile.ReferenceSet, layerIdx int) ([]stereoscopeFile.Reference, error) {
-	uniqueFiles := make([]stereoscopeFile.Reference, 0)
+func (r *ContainerImageAllLayers) locationsByRef(ref stereoscopeFile.Reference, accessPath string, uniqueFileIDs stereoscopeFile.ReferenceSet, layerPos int) ([]file.Location, error) {
+	uniqueLocations := make([]file.Location, 0)
+
+	// if the access path is itself a hardlink, surface it as the underlying type it points to, at its own path
+	// (bound to the target's content), so that image results are in parity with directory results (which cannot
+	// tell a hardlink from a regular file). the path-based lookup is required because the search that produced ref
+	// already followed the basename link, collapsing the hardlink onto its target.
+	if ownRef, targetRef, ok := r.hardLinkAtPath(accessPath, r.layers[layerPos]); ok {
+		if !uniqueFileIDs.Contains(ownRef) {
+			uniqueFileIDs.Add(ownRef)
+			uniqueLocations = append(uniqueLocations, file.NewVirtualLocationFromImage(string(ownRef.RealPath), accessPath, targetRef, r.img))
+		}
+		return uniqueLocations, nil
+	}
 
 	// since there is potentially considerable work for each symlink/hardlink that needs to be resolved, let's check to see if this is a symlink/hardlink first
 	entry, err := r.img.FileCatalog.Get(ref)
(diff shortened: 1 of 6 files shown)


syft/internal/fileresolver/container_image_all_layers.go
```diff
@@ -65,22 +77,41 @@ func (r *ContainerImageAllLayers) fileByRef(ref stereoscopeFile.Reference, uniqu
...
 	}
 
-	return uniqueFiles, nil
+	return uniqueLocations, nil
+}
+
+// hardLinkAtPath returns the hardlink's own reference and its resolved target reference when the basename of path is a
+// hardlink within the given layer. ok is false when path does not exist there or is not a hardlink. The lookup does
+// not follow the basename link so that the hardlink's own path is preserved. This adds a tree walk per
+// matched ref (path x layer); if it shows up in profiles, fold the hardlink check into the existing search resolution.
```
spiffcs, line 101: I didn't see a note in the PR description about this so I'm going to performance check it on my end real quick and post the results. If it doesn't have a huge impact then 🟢

spiffcs (commented):
No notes on the changes. I saw there was a comment about a potential performance check we might want to run so I'm giving that a shake before 🟢

syft/internal/fileresolver/container_image_all_layers.go
spiffcs: Nothing show stopping came up in wall clock or allocations from the perf testing here.

spiffcs (approved):

PR #5046, opened 2026-07-08
Add support for programmatically listing catalogers

### LINKED ISSUE #4992: Expose an interface to list catalogers

Syft provides a subcommand to list all catalogers but the implementation of this is in `internal`. I'm filing this to see if it makes sense to provide a public Go function to list all catalogers for API-based Syft usage.

**Why is this needed**:

Syft provides public Go APIs to create an SBOM and this includes the ability to add catalogers. However, there doesn't seem to be a way, when using Syft this way rather than via the CLI, to get a list of all catalogers for introspection.

**Additional context**:

I'm happy to work on a PR to add this if the maintainers think this makes sense. :)

adityasaky:
## Description

This change adds a public interface to list all configured catalogers. This is a first pass at implementing https://github.com/anchore/syft/issues/4992 and as of now does not handle the pieces about `cataloger info` and config-based hydration.

## Type of change

## Checklist

## Issue references

Fixes #4992

### DIFF
--- a/syft/exp/cataloger.go
+++ b/syft/exp/cataloger.go
@@ -1,0 +1,106 @@
+package exp
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/anchore/syft/internal/capabilities"
+	"github.com/anchore/syft/internal/task"
+	"github.com/anchore/syft/syft/cataloging"
+	"github.com/anchore/syft/syft/cataloging/pkgcataloging"
+)
+
+// Cataloger describes an available cataloger and its selection metadata.
+type Cataloger struct {
+	Name string   `json:"name"`
+	Tags []string `json:"tags"`
+}
+
+// AllCatalogers returns information about all registered catalogers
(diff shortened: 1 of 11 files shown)


syft/exp/cataloger.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// AllCatalogers returns information about all registered catalogers
+// (both package and file catalogers), sorted by name. Any additional
+// user-provided cataloger references are included in the result.
+func AllCatalogers(additional ...pkgcataloging.CatalogerReference) ([]Cataloger, error) {
+	return SelectCatalogers(cataloging.SelectionRequest{
+		DefaultNamesOrTags: []string{"all"},
+	}, additional...)
+}
```
adityasaky: I'm currently using this approach to provide any user defined catalogers that are configured via `CreateSBOMConfig.WithCatalogers()`. I'm not sure this is the best way to go about this though.

PR #5047, opened 2026-07-09
docs: fix stale commands in binary cataloger README

### LINKED ISSUE #4510: update README.md about binary classifier cataloger
**What happened**:

- `make add-snippet` shows '(Y/n/q)', but `y` is required to make snippet file.
- `go generate ./internal/capabilities` is required in top directory, and updated file should be included in PR

https://github.com/anchore/syft/blob/main/syft/pkg/cataloger/binary/README.md

This file and script is very helpful.

alliasgher:
## Description

The binary classifier cataloger developer README (`syft/pkg/cataloger/binary/README.md`) documents several commands that no longer match the code:

- **Testing command was wrong.** It showed `go test -must-use-full-binaries ./syft/pkg/cataloger/binary/testdata/...`, but the flag is defined as `-must-use-original-binaries` (in `classifier_cataloger_test.go`) and `Test_Cataloger_PositiveCases` lives in the `binary` package, not under `testdata/`. Fixed the command and added a single-fixture variant.
- **`add-snippet` invocation.** Replaced the stale `go run ./manager add-snippet` references with the `make add-snippet` target, and noted that the `(Y/n/q)` prompt requires `y` to write the snippet.
- **Test file name.** The TL;DR pointed at `../cataloger_test.go`; the test actually lives in `../classifier_cataloger_test.go`.
- **Capabilities regeneration.** Documented the `go generate ./internal/capabilities` step (run from the repo root, commit the regenerated `capabilities.yaml`) for when a new classifier is added.

Commands were verified against `main` (the `-run Test_Cataloger_PositiveCases/<name> -must-use-original-binaries` form runs the expected cases; the flag and file paths were checked in the source).

Fixes #4510

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — documentation correction

### DIFF
--- a/syft/pkg/cataloger/binary/README.md
+++ b/syft/pkg/cataloger/binary/README.md
@@ -5,8 +5,9 @@
 >  1. head to the correct directory: `cd testdata`
 >  2. add a new entry to `config.yaml` to track where to get the binary from (verify the entry with `make list`)
 >  3. run `make download` to get the binary
->  4. run `make add-snippet` and follow the prompts (use `/` to search)
->  5. add a new test case to `Test_Cataloger_PositiveCases` in `../cataloger_test.go`
+>  4. run `make add-snippet` and follow the prompts (use `/` to search; answer `y` at the `(Y/n/q)` prompt to write the snippet)
+>  5. add a new test case to `Test_Cataloger_PositiveCases` in `../classifier_cataloger_test.go`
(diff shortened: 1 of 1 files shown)


wagoodman (approved):

wagoodman (commented):
force push was simply a rebase

PR #5056, opened 2026-07-11
fix(ruby): strip platform suffix from Gemfile.lock gem versions

Synvoya:
## Description

Bundler writes platform-specific gems with the platform appended to the version, e.g. `nokogiri (1.16.0-x86_64-linux)` (produced by `bundle lock --add-platform`). The parser trimmed only the parentheses, keeping `1.16.0-x86_64-linux` as the version. That corrupts the PURL — `pkg:gem/nokogiri@1.16.0-x86_64-linux` instead of `pkg:gem/nokogiri@1.16.0` — which breaks vulnerability matching, and it produces a separate entry per platform for the same gem.

## Fix

Split the platform off at the first `-`. A RubyGems version never contains a `-` (pre-release segments use `.`, e.g. `1.0.0.beta`), so the split point is unambiguous. Extracted into a small `gemfileLockVersion` helper.

## Test

Added a platform gem (`nokogiri (1.16.0-x86_64-linux)`) to `testdata/Gemfile.lock` with the expected `nokogiri 1.16.0` in `parse_gemfile_lock_test.go`. `go test ./syft/pkg/cataloger/ruby/` passes; reverting only the parser change fails the test (version/PURL keep the `-x86_64-linux` suffix).

### DIFF
--- a/syft/pkg/cataloger/ruby/parse_gemfile_lock.go
+++ b/syft/pkg/cataloger/ruby/parse_gemfile_lock.go
@@ -46,7 +46,7 @@ func parseGemFileLockEntries(_ context.Context, _ file.Resolver, _ *generic.Envi
 			pkgs = append(pkgs,
 				newGemfileLockPackage(
 					candidate[0],
-					gemfileLockVersion(candidate[1]),
+					strings.Trim(candidate[1], "()"),
 					reader.WithAnnotation(pkg.EvidenceAnnotationKey, pkg.PrimaryEvidenceAnnotation),
 				),
 			)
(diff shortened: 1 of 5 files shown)


syft/pkg/cataloger/ruby/parse_gemfile_lock.go
```diff
@@ -46,7 +46,7 @@ func parseGemFileLockEntries(_ context.Context, _ file.Resolver, _ *generic.Envi
 			pkgs = append(pkgs,
 				newGemfileLockPackage(
 					candidate[0],
-					strings.Trim(candidate[1], "()"),
+					gemfileLockVersion(candidate[1]),
```
wagoodman: Is there a reason why we should modify the package version? Since the PR description is really pointing out that the PURL version is what needs fixing, and the platform being embedded in the version seems genuinely useful, then why not only adjust the PURL version field? (keep the version you actually found)

Synvoya: Good call — thanks. You're right that the platform is useful and shouldn't be dropped from the version. I've kept `pkg.Version` as found (`1.16.0-x86_64-linux`) and moved the platform stripping into `packageURL`, so only the PURL's version field is normalized to `pkg:gem/nokogiri@1.16.0` for matching while the package keeps the full version. Tests updated (pushed in b834b93).

If you'd also like the platform preserved in the PURL itself, I'm happy to add it as a `platform` qualifier (`pkg:gem/nokogiri@1.16.0?platform=x86_64-linux`) instead — just say the word.

PR #5069, opened 2026-07-16
Avoid duplicate packages for replaced Go modules

### LINKED ISSUE #2721: Misinterpretation of Multiple replace Directives in Golang
**What happened**:
It does not apply replace directive to the module version that's also in the replace directly. For example,
```
replace (
     google.golang.org/grpc => google.golang.org/grpc v1.61.0
     golang.org/x/net => golang.org/x/net v0.22.0
)
```
If I run syft with the working dir `syft .`, the result will show `golang.org/x/net@v0.18.0` is being used, which is originated from [grpc@v1.61.0's go.mod](https://github.com/grpc/grpc-go/blob/v1.61.0/go.mod) while both module should be overridden to the specified version.

If I build and run syft against the binary file instead, `syft <bin_file>`, or `go version -m <bin_file>`, both will show `golang.org/x/net@v0.22.0` is actually being used.

1. Create a simple project
2. Import a module, let's name it module M
3. Use `replace` directive with M to any specific version
4. Use `replace` directive with another module required by M (pick one from M's go.mod), let's name it module N
5. `cd` to the project root directory, set `$GOPATH` with `export GOPATH="$(pwd)/dep`
6. `go mod download`, to download all dependencies into `./dep`.
7. Run `syft` with the project directory, inspect the N's version, it will be the version specified in M's go.mod, while it should actually be the `replace` directive version.

**Environment**:
- Output of `syft version`: 
```
Application: syft
Version:    1.0.1
BuildDate:  2024-03-06T19:25:56Z
GitCommit:  Homebrew
Platform:   darwin/arm64
GoVersion:  go1.22.1
Compiler:   gc
```
```
ProductName:		macOS
ProductVersion:		14.3.1
BuildVersion:		23D60
```

ychampion:
## Description

Source scans already apply module replacements through `go/packages`, but the fallback `go.mod` path was synthesizing imported replacements a second time. This produced duplicate package records with the same replacement version but different IDs and metadata.

Skip fallback synthesis when source analysis already resolved the original module path. Replacements for unimported modules and local paths still use the existing fallback.

### Testing

- `go test ./syft/pkg/cataloger/golang -run '^(Test_parseGoSource_replacedModulesAreNotDuplicated|TestParseGoMod|Test_packageURL|Test_newGoBinaryPackage_relativeReplace|Test_Config)$' -count=1`
- `go vet ./syft/pkg/cataloger/golang`

## Type of change

## Checklist

## Issue references

Closes #2721


syft/pkg/cataloger/golang/testdata/go-source-replacements/go.mod
```diff
@@ -0,0 +1,15 @@
+module example.com/repro
```
kzantow, line 1: these test files should be in `testdata` -- `gotestdata` is a different thing

ychampion: Moved these fixtures to `testdata` and updated the test path in 42718661. The focused regression and Go cataloger vet both pass.

wagoodman (approved):

Issue #1683, opened 2023-03-20
Support SBOM creation for container image indexes


eliaslevy:
**What would you like to be added**:
Create an SBOM when given a container reference that points to a container image index that includes the dependencies of all the image manifests the index points to.

**Why is this needed**:
Multi-architecture and multi-platform container images are becoming more common.

**Additional context**:
Syft supports creating an SBOM from a container reference. When given a reference, Syft will pull the reference to download the container image. When the reference points to a container image index, Syft will download one of the underlaying images depending on that platform the tool is run on, and generate an SBOM from that image. That means that the dependencies of the alternate images are not cataloged and documented in the SBOM, creating a visibility gap.

Syft should optionally download each underlaying image and document the dependencies in all of them. Ideally Syft would model the image index and image manifests as packages in the SBOM with appropriate relationships, so that the dependency graph information is recorded.

kzantow:
Syft can download different architectures when using the `--platform` flag, but is unable to combine multiple sources into a single SBOM today. Would you be able to just use this flag to create separate SBOMs for each image architecture?

We investigated combining multiple SBOMs into a single one, but there are some definite issues with this. Feel free to read [this issue](https://github.com/anchore/syft/issues/617) for some more information there.

eliaslevy:
We could generate SBOMs for each underlying container image separately, then merge them into a single SBOM. But I believe this is a common enough use case that Syft should support it.

Multi-architecture and multi-platform images are exposed to the user via the image index digest. So when they request an SBOM for the artifact, they will use that identifier in the lookup, rather than a digest for one of the underlying image manifests.

kzantow:
Hi @eliaslevy at present this request requires a couple different things:
1) we need to be able to scan _multiple_ images for all architectures given an image tag
2) we need to be able to combine multiple SBOMs into a single SBOM

Today, we don't have either of these features on the roadmap. Combining multiple SBOMs, for example, is a nontrivial task that requires understanding about what packages and files mean within the context of the provided SBOM.

That said, we always accept PRs and we're always happy to talk about this on Slack or our community office hours!

JonZeolla:
I would think (at least in my scenario) that making a per-platform SBOM from a multi-platform manifest would be a very valuable step forward

stevehipwell:
[...]

PR #5074, opened 2026-07-16
Add multi-platform OCI image support

### LINKED ISSUE #1545: Support for `application/vnd.oci.image.index.v1+json` manifests in root OCI layout
Syft expects  OCI layout to contain only 1 image of mediaType `application/vnd.oci.image.manifest.v1+json` and doesn't support mediaType `application/vnd.oci.image.index.v1+json` when building a single OCI tar ball for multiple architectures

**Why is this needed**:
This adds an additional layer of complexity to extract the digest of each image manifest for each architecture in the `application/vnd.oci.image.index.v1+json` (use regctl --platform for extraction) to run the scan

**Additional context**:
Uploaded a `demo_alpine` OCI layout (single architecture and image manifest) vs `demo_amazonlinux` OCI layout (multiple architectures within single manifest of type `application/vnd.oci.image.index.v1+json`)
 
Current behavior:

- works for demo_alpine oci layout with single image manifest
Screen Shot 2023-02-07 at 11 56 26 AM

- Fails for `demo_amazonlinux` with error: `* failed to construct source from user input "docker-archive-demo-amz-2.tar": could not fetch image "docker-archive-demo-amz-2.tar": unable to use OciTarball source: unable to parse OCI directory as an image: unexpected media type for sha256:1ab94ef8f74d975ce5b3637944358cce8d776259f493c4d857898dbe862c1fb3: application/vnd.oci.image.index.v1+json`
Screen Shot 2023-02-07 at 11 57 51 AM

jasonpaulos:
## Description

Update stereoscope to include changes from https://github.com/anchore/stereoscope/pull/548, which add multi-platform OCI support.

To be clear, this doesn't add support for scanning multiple platforms simultaneously, but it does allow syft to operate on a single image within a multi-platform OCI image, which is currently not possible.

Also added an integration test to confirm this support is available in syft.

## Type of change

## Checklist

## Issue references

Fixes #1545 and part of #1683.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -335,4 +335,4 @@ retract (
 	v0.53.1 // published accidentally with incorrect license in depdencies
 )
 
-replace github.com/anchore/stereoscope => github.com/jasonpaulos/stereoscope v0.0.0-20260716213708-1243139a6baf
+replace github.com/anchore/stereoscope => github.com/jasonpaulos/stereoscope v0.0.0-20260717145207-e197ddfc615d
--- a/go.mod
+++ b/go.mod
@@ -183,7 +183,7 @@ require (
 	github.com/containerd/typeurl/v2 v2.2.3 // indirect
 	github.com/cyphar/filepath-securejoin v0.6.1 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
-	github.com/docker/cli v29.5.3+incompatible // indirect
+	github.com/docker/cli v29.6.1+incompatible // indirect
 	github.com/docker/docker-credential-helpers v0.9.5 // indirect
 	github.com/docker/go-connections v0.7.0 // indirect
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -334,3 +334,5 @@ retract (
 	v0.53.2
 	v0.53.1 // published accidentally with incorrect license in depdencies
 )
+
+replace github.com/anchore/stereoscope => github.com/jasonpaulos/stereoscope v0.0.0-20260716213708-1243139a6baf
```
jasonpaulos: Needs to be removed before merging

jasonpaulos:
Hi @kzantow, the unit test failure seems unrelated to the changes here. Do you have any suggestions?

kzantow:
@jasonpaulos yes, it's unrelated -- I've been seeing this on other PRs, don't worry about it (there is some large test fixture image that is being rebuilt; needs investigation)

wagoodman (approved):

PR #5080, opened 2026-07-18
fix(java): preserve Eclipse Classpath license expressions

### LINKED ISSUE #5009: License normalization drops Classpath-exception-2.0 and uses AND instead of OR for Eclipse/GlassFish JARs
**What happened**:

When Syft catalogs Java packages whose license is GPL-2.0 WITH Classpath-exception-2.0 (the standard Eclipse/GlassFish dual-licensing used by Jakarta and Tyrus artifacts), it never produces the valid SPDX exception expression GPL-2.0-only WITH Classpath-exception-2.0. The Classpath Exception is lost in one of two ways depending on the package:

1. Hard-dropped to plain GPL. For org.glassfish.tyrus.bundles:tyrus-standalone-client:2.1.3, Syft concludes EPL-2.0 AND GPL-2.0-only, silently stripping the WITH Classpath-exception-2.0 clause and presenting the artifact as plain GPL-2.0.
2. Emitted as a non-SPDX LicenseRef. For tyrus-core/tyrus-client/tyrus-spi/tyrus-container-grizzly-client:2.1.3 and jakarta.websocket-client-api:2.1.0, the declared license becomes an opaque reference such as LicenseRef-GNU-General-Public-License--version-2-with-the-GNU-Classpath-Exception. The wording survives, but it is not a recognized SPDX license/exception identifier, so downstream policy engines and license matchers do not treat it as the Classpath Exception.

Syft should emit the valid SPDX expression GPL-2.0-only WITH Classpath-exception-2.0 (or EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0 for the dual-license case) for these packages, preserving the exception as a proper SPDX WITH expression rather than dropping it to GPL-2.0-only or stuffing it into a LicenseRef.

1. Create a Maven project with these dependencies (all on Maven Central):
  - org.glassfish.tyrus.bundles:tyrus-standalone-client:2.1.3
  - org.glassfish.tyrus:tyrus-core:2.1.3
  - org.glassfish.tyrus:tyrus-client:2.1.3
  - org.glassfish.tyrus:tyrus-spi:2.1.3
  - org.glassfish.tyrus:tyrus-container-grizzly-client:2.1.3
  - jakarta.websocket:jakarta.websocket-client-api:2.1.0
2. Resolve the JARs into a directory: mvn dependency:copy-dependencies -DoutputDirectory=libs
3. Scan them: syft ./libs -o spdx-json > sbom.json
4. Inspect the licenseConcluded / licenseDeclared fields for the packages above. None contain GPL-2.0-only
WITH Classpath-exception-2.0; tyrus-standalone-client shows EPL-2.0 AND GPL-2.0-only, and the others show LicenseRef-... entries.

The Classpath Exception is unambiguously present in the JARs themselves, so this is purely a license-normalization gap in Syft, not missing source metadata:

- jakarta.websocket-client-api-2.1.0.jar's embedded META-INF/maven/.../pom.xml literally contains SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0.
- The Tyrus JARs ship a META-INF/LICENSE.md with a dedicated ## CLASSPATH EXCEPTION section.

**Environment**:

- Output of syft version:
Application:    syft
Version:        1.44.0
BuildDate:      2026-05-01T17:11:01Z
GitCommit:      8cb78ce40ced6a731fb83f2a491a67444f541bf1
GitDescription: v1.44.0
Platform:       darwin/arm64
GoVersion:      go1.26.2
Compiler:       gc
SchemaVersion:  16.1.3
- OS (e.g: cat /etc/os-release or similar): macOS 15.7.4 (build 24G517), arm64

Eljees:
## Description

Fixes Java license normalization for Eclipse/GlassFish artifacts that are dual-licensed under EPL-2.0 or GPL-2.0 with the Classpath Exception.

Before this change, Syft either emitted an opaque `LicenseRef` for the license URLs or concluded `EPL-2.0 AND GPL-2.0-only`. The latter both changed the dual-license semantics and dropped `Classpath-exception-2.0`.

This change:

- normalizes the Eclipse and GNU Classpath Exception URLs to valid SPDX expressions
- parses individual OSGi `Bundle-License` clauses while preserving quoted attributes
- combines Maven license alternatives with `OR` when every alternative resolves to SPDX
- preserves the Classpath Exception when Java license text contains the EPL/GPL pair and an explicit Classpath Exception section
- leaves unknown and unrelated license combinations unchanged

After the change, all six artifacts from the issue reproducer report:

```text
(EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0)
```

Validation performed:

- `go test ./internal/spdxlicense -count=1`
- targeted Java cataloger regression tests for Maven alternatives, OSGi manifest parsing, POM parsing, and license-file normalization
- end-to-end SPDX JSON scan of all six pinned JARs from the reproducer
- `git diff --check`

## Type of change

## Checklist

## Issue references

Fixes #5009

### DIFF
--- a/internal/spdxlicense/license.go
+++ b/internal/spdxlicense/license.go
@@ -16,6 +16,12 @@ const (
 	LicenseRefPrefix = "LicenseRef-" // prefix for non-standard licenses
 )
 
+var licenseExpressionsByURL = map[string]string{
+	"projects.eclipse.org/license/secondary-gpl-2.0-cp": "GPL-2.0-only WITH Classpath-exception-2.0",
+	"www.eclipse.org/org/documents/epl-2.0/EPL-2.0.txt": "EPL-2.0",
+	"www.gnu.org/software/classpath/license.html":       "GPL-2.0-only WITH Classpath-exception-2.0",
+}
+
 //go:generate go run ./generate
 
 // ID returns the canonical license ID for the given license ID
@@ -50,6 +56,11 @@ func LicenseByURL(url string) (LicenseInfo, bool) {
 	if id, exists := urlToLicense[url]; exists {
 		return LicenseInfo{
 			ID: id,
+		}, true
+	}
+	if expression, exists := licenseExpressionsByURL[url]; exists {
+		return LicenseInfo{
+			ID: expression,
 		}, true
 	}
 	return LicenseInfo{}, false
--- a/internal/spdxlicense/license_url_test.go
+++ b/internal/spdxlicense/license_url_test.go
@@ -42,6 +42,24 @@ func TestLicenseByURL(t *testing.T) {
 			wantFound: true,
 		},
 		{
+			name:      "GNU Classpath Exception URL",
+			url:       "https://www.gnu.org/software/classpath/license.html",
+			wantID:    "GPL-2.0-only WITH Classpath-exception-2.0",
+			wantFound: true,
+		},
+		{
+			name:      "Eclipse secondary GPL Classpath URL",
+			url:       "https://projects.eclipse.org/license/secondary-gpl-2.0-cp",
+			wantID:    "GPL-2.0-only WITH Classpath-exception-2.0",
+			wantFound: true,
+		},
+		{
+			name:      "Eclipse Public License legacy URL",
+			url:       "https://www.eclipse.org/org/documents/epl-2.0/EPL-2.0.txt",
+			wantID:    "EPL-2.0",
+			wantFound: true,
+		},
+		{
 			name:      "URL with trailing whitespace",
 			url:       "  http://opensource.org/licenses/MIT  ",
 			wantID:    "MIT",
--- a/syft/pkg/cataloger/java/archive_parser.go
+++ b/syft/pkg/cataloger/java/archive_parser.go
@@ -18,6 +18,7 @@ import (
 	"github.com/anchore/syft/internal"
 	intFile "github.com/anchore/syft/internal/file"
 	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/internal/spdxlicense"
 	"github.com/anchore/syft/internal/tmpdir"
 	"github.com/anchore/syft/internal/unknown"
 	"github.com/anchore/syft/syft/artifact"
@@ -50,6 +51,8 @@ var archiveFormatGlobs = []string{
 	"**/*.rar", // Java Resource Adapter Archive
 	"**/*.zap", // ZAP add-ons https://github.com/zaproxy/zaproxy/wiki/ZapAddOns
 }
+
+const eclipseClasspathDualLicense = "EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0"
 
 // javaArchiveHashes are all the current hash algorithms used to calculate archive digests
 var javaArchiveHashes = []crypto.Hash{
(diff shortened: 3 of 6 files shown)


Eljees:
Ping - this one's been quiet for a while and CI is green. Happy to adjust the approach if needed.

Eljees:
Still open and still green, so a gentle nudge.

Short recap of what it does: a Java archive whose license is declared as an SPDX expression (for example `EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0`, which is what the Eclipse projects publish) currently gets split up and the expression is lost. This keeps the expression intact rather than reducing it to unrelated single licenses.

If you'd prefer a narrower version, or want it folded into other work in this area, say so and I'll adjust — or close it, which is a fine answer too.

Eljees:
Rebased onto `main` (`0c04980`) — the branch had drifted into a conflict, which is why only 1 of the usual checks was running.

The conflict was positional only: `main` added `releaseVersionPattern` + `versionFromPropertiesFile` in exactly the spot where this PR adds `normalizeClasspathExceptionLicenses`, so git folded the two functions' shared closing brace into common context. Both are kept, unchanged.

What I checked before force-pushing:

* `gofmt -l` clean, `go build` and `go vet` on `syft/pkg/cataloger/java` and `internal/spdxlicense` pass;
* this PR's own tests pass — `TestNormalizeClasspathExceptionLicenses` and `TestNormalizeClasspathExceptionLicensesPreservesOtherLicenses`;
* `git diff <old head> <new head>` over `archive_parser.go` contains only `main`'s own additions, nothing of mine;
* `Test_deterministicMatchingPomProperties` fails locally, but it fails identically on unmodified `origin/main` (`0c04980`) in the same worktree setup, so it is not from this branch. The fixture-building tests in `archive_parser_test.go` can't run in my contour at all (the `run` helper blocks on IO until the 600s timeout), so CI is the authority on those.

Content of the change is untouched from the version that has been open since 18 July. Still happy to take it in a different direction if the approach isn't what you want.

Eljees:
Correction to my previous comment: I wrote that the conflict was why only 1 check was running. That was wrong, and I should have checked before saying it.

What the API actually shows:

* on the old head `784d546c` there were **0** workflow runs — CI never ran on this PR at all, so nothing about it can be blamed on the conflict;
* on the new head `237e7e96` there are now 3 runs — CodeQL, Validations, Validate GitHub Actions — all sitting at `action_required`, i.e. waiting for a maintainer to approve running workflows for this contributor;
* for comparison, #5112's head has the same 3 runs, approved and green.

So the push did produce the runs, and they need an approval from your side to start. The `DCO` check is the one that passes on its own.

Everything else in the previous comment stands — the rebase, the resolution, and the local checks are as described.

Eljees:
One more data point, from outside the reproducer's six artifacts: the same defect shows up in a shipping product. `jenkins.war` 2.568.2 bundles `jakarta.annotation-api-3.0.0.jar`, whose manifest carries the pair as a wrapped, comma-separated `Bundle-License`:

```
Bundle-License: https://www.eclipse.org/legal/epl-2.0, https://www.gnu.o
 rg/software/classpath/license.html
```

Scanning that jar directly, on `main` at `3325493` and on this branch:

```
main       value = "https://www.eclipse.org/legal/epl-2.0, https://www.gnu.org/software/classpath/license.html"
           spdxExpression = ""            <- nothing downstream can use this

this PR    value = "https://www.eclipse.org/legal/epl-2.0 OR https://www.gnu.org/software/classpath/license.html"
           spdxExpression = "EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0"
```

The line continuation matters here: the URL is split across two manifest lines, so this also exercises the unfolding path, and the comma separating the two clauses is the part `main` never splits — which is why the whole thing stays one opaque value with an empty SPDX expression.

The branch still merges into current `main` cleanly; on the merge result the affected packages fail only the tests that a pristine `main` checkout fails here as well (they need `docker` to build fixtures), so nothing is attributable to the change.

PR #5088, opened 2026-07-22
feat(release): attach a vendored-source archive to releases

### LINKED ISSUE #5048: Request: provide a vendored source dependency archive for downstream packaging
Hi, thanks for maintaining Syft.

I am working on downstream RPM packaging for Syft. The GitHub auto-generated source archive (`archive/refs/tags/vX.Y.Z.tar.gz`) does not include a root `vendor/` directory, and Syft has a large Go module dependency graph.

For distributions that need offline builds, it would be very helpful if each GitHub release also provided an official source dependency archive, for example:

- `syft-vX.Y.Z-vendor.tar.gz`
- generated from the release tag with `go mod vendor`

Would you be open to publishing an official vendored dependency archive as part of the release artifacts?

Thanks!

youdie006:
## What

Attaches a vendored-source archive to each syft release so downstream packagers can build offline (#5048).

## Approach

goreleaser-native, since syft's whole release is goreleaser-driven:

- A global `before` hook (`.github/scripts/create-vendored-source-archive.sh`) runs `go mod vendor` and tars the source tree (`git ls-files`) plus `vendor/` into `dist/syft_<version>_vendored-source.tar.gz` under a `syft-<version>/` prefix, then removes `vendor/` so the goreleaser build itself is unchanged.
- `release.extra_files` attaches that archive to the GitHub release.
- The hook is skipped for snapshot builds (`{{ .IsSnapshot }}`) so the frequent snapshot CI isn't slowed.

No workflow changes and no existing artifact altered — it all lives in `.goreleaser.yaml`, syft's single release source of truth.

## Verification

- `go mod vendor` + `go build -mod=vendor ./...` (CGO_ENABLED=0, matching the release) both pass.
- The hook end-to-end: snapshot mode skips; release mode produces the archive in **~7s** (49 MB) — negligible against syft's multi-arch/multi-image release, well within the "not very poor performance" bar you mentioned.
- **Offline build proof**: extracted the archive to a fresh dir and ran `GOPROXY=off GOMODCACHE=<empty> go build -mod=vendor ./cmd/syft` — it built with zero network (the module cache was never even created), which is exactly the downstream offline-packaging use case.
- The archive is reproducible (`--sort=name --mtime --owner=0 --group=0`), excludes `.git`, and targets GNU tar (the Linux release runner).

`goreleaser check` couldn't be run here (goreleaser isn't installed) and the real release path can't be exercised in a sandbox, but `.goreleaser.yaml` parses cleanly and the archive/build steps are verified as above. I kept it surgical (didn't add the archive to `checksums.txt`) — happy to also checksum/sign it alongside the binaries if you'd prefer.

Closes #5048.

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). Verified with `go mod vendor` plus an offline `-mod=vendor` build from the extracted archive. DCO sign-off included.*

### DIFF
--- a/.github/scripts/create-vendored-source-archive.sh
+++ b/.github/scripts/create-vendored-source-archive.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+# Create a vendored-source archive for downstream (offline) packaging.
+#
+# Produces dist/syft_<version>_vendored-source.tar.gz containing the full syft
+# source tree plus a vendor/ directory of all Go module dependencies, so
+# downstream packagers can build syft without network access:
+#
+#     tar -xzf syft_<version>_vendored-source.tar.gz
+#     cd syft-<version>
+#     go build -mod=vendor ./cmd/syft
+#
+# This is invoked as a goreleaser global "before" hook (see .goreleaser.yaml) and
+# the resulting archive is attached to the GitHub release via release.extra_files.
+#
+# Vendoring is skipped for snapshot builds to keep the frequent snapshot CI fast.
+# The vendor/ directory is removed again once the archive is written so the actual
+# goreleaser build runs in module mode exactly as it does today (no behavior change
+# to the produced binaries).
+set -euo pipefail
+
+VERSION="${1:?usage: create-vendored-source-archive.sh <version> <is-snapshot>}"
+IS_SNAPSHOT="${2:-false}"
+
+if [ "${IS_SNAPSHOT}" = "true" ]; then
+  echo "skipping vendored-source archive for snapshot build"
+  exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


spiffcs:
@youdie006 thank you for the PR. 

I added a `needs-discussion` label here so our team knows to look at it during out next sync to see if this is something we'd want as part of the release process. When we come to a decision we'll provide feedback around how we're going to move forward on this PR ❤️

youdie006:
Thanks @spiffcs, that makes sense - happy to adjust the approach once your team lands on a direction. Let me know if there's anything you'd like me to change in the meantime.

Jvlegod:
hello @kzantow, how about this PR? :>

PR #5090, opened 2026-07-22
fix: correct Apache Derby group ID in purl generation

Ankush-Pathak:
## Description

Apache Derby jars (`derby`, `derbyshared`, `derbytools`, …) are OSGi bundles that ship **no `pom.properties`/`pom.xml`**. With no pom metadata, syft falls through to the manifest heuristic (`groupIDFromJavaManifest`), which walks `PrimaryJavaManifestGroupIDFields` and returns the first value starting with a TLD prefix. For Derby that is `Bundle-Activator: org.apache.derby.osgi.EmbeddedActivator` (`Bundle-SymbolicName: derby` is skipped because it has no TLD prefix).

The result is a wrong purl:

- before: `pkg:maven/org.apache.derby.osgi.EmbeddedActivator/derby@10.16.1.1`
- after:  `pkg:maven/org.apache.derby/derby@10.16.1.1`

Because the group ID is wrong, downstream matchers (e.g. grype's Maven/GHSA matcher, which does not use CPEs for Java by default) fail to match advisories like `GHSA-rcjc-c4pj-xxrp` / `CVE-2022-46337` (Critical), even though the jar is clearly the affected `org.apache.derby:derby`.

This adds the Derby artifacts to `DefaultArtifactIDToGroupID` so the known-package-list step (which runs before the manifest heuristic) supplies the correct group ID. All six are published under `org.apache.derby` on Maven Central.

## Validation

Built syft with this change and cataloged real Derby jars from Maven Central:

| jar | purl (after fix) | grype (default config) |
|-----|------------------|------------------------|
| `derby-10.16.1.1.jar` (vulnerable) | `pkg:maven/org.apache.derby/derby@10.16.1.1` | detects `GHSA-rcjc-c4pj-xxrp` (Critical) — previously missed |
| `derby-10.17.1.0.jar` (fixed, latest) | `pkg:maven/org.apache.derby/derby@10.17.1.0` | no findings — no false positive on the patched version |

Before the fix the purl was `pkg:maven/org.apache.derby.osgi.EmbeddedActivator/derby@…`; the stray `apache:EmbeddedActivator` / `apache:osgi` CPE candidates are also gone. For jars that do embed pom metadata the pom-based steps still take precedence, so the map entry only applies when the coordinates are otherwise unavailable.

## Type of change


Ankush-Pathak:
@wagoodman could you approve the workflows?

kzantow (approved):
LGTM, thanks 👍

Issue #585, opened 2021-10-22
sign Syft binary/container image via cosign 


developer-guy:
**What would you like to be added**:

A tool called `cosign`[^1] created and maintained by the `sigstore` [^2] community allows you to sign and verify container images. IINM `Syft` takes its releases via `GoReleaser,` luckily `cosign` is now integrated into `GoReleaser.`[^3]. Btw, there is a similar topic ongoing for the `ossf/scorecard` project too.[^4]

**Why is this needed**:

**Additional context**:
[^1]: https://github.com/sigstore/cosign
[^2]: https://sigstore.dev/
[^3]: https://carlosbecker.com/posts/goreleaser-cosign/
[^4]: https://github.com/ossf/scorecard/issues/309

cc: @luhring @wagoodman

nwl:
![](https://github.trello.services/images/mini-trello-icon.png) [Syft/Sigstore integration](https://trello.com/c/2oamyp9h/71-syft-sigstore-integration)

kzantow:
Since `syft attest` has been implemented, is there anything more to do here?

luhring:
@kzantow If I'm reading this issue right, it's not about Syft generating SBOM attestations, but rather about signing releases of Syft itself (in various form factors, e.g. the Syft binary having a signature, container images of Syft having signatures, etc.)

kzantow:
@luhring yeah -- I definitely misread the description 🤦

Stetsed:
Hey lads,

I recently stumbled across this, and noticed while setting up my own action to install Syft due to not wanting to rely on the sbom-action due to it's use of grabbing a script from main, although that discussion is going there and actually has some of the things I integrated discussed. 

However, that aside I did notice that Syft currently uses an outdated format for it's attestations, using the since deprecated certificate/signature model, instead of a bundle. Goreleaser does support the new model without any fundamental changes, as can be found in the trivy goreleaser:
```
signs:
- cmd: cosign
  signature: "${artifact}.sigstore.json"
  args:
    - "sign-blob"
    - "--bundle=${signature}"
    - "${artifact}"
    - "--yes"
  artifacts: all
  output: true
```

Just wanted to drop by and share this information incase it might be useful. I do see that you guys have introduced the old schema of attestation, so that is good as I can see it was introduced https://github.com/anchore/syft/commit/6ae5b2904d9f4e040b6e812146b4a4cfe7d36f24, so for now I am simply using that to verify, but would love to see the non-deprecated!

PR #5092, opened 2026-07-22
chore(ci): generate sigstore bundles instead of signatures and certs

scop:
## Description

Separate signatures and certificates are deprecated as of cosign 3.1.1,
refs https://github.com/sigstore/cosign/releases/tag/v3.1.1, https://github.com/sigstore/cosign/releases/tag/v3.1.2

## Type of change

## Checklist

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -169,14 +169,12 @@ sboms:
 
 signs:
   - cmd: .tool/cosign
-    signature: "${artifact}.sig"
-    certificate: "${artifact}.pem"
+    signature: "${artifact}.sigstore.json"
     args:
       - "sign-blob"
       - "--use-signing-config=false"
       - "--oidc-issuer=https://token.actions.githubusercontent.com"
-      - "--output-certificate=${certificate}"
-      - "--output-signature=${signature}"
+      - "--bundle=${signature}"
       - "${artifact}"
       - "--yes"
     artifacts: checksum


wagoodman:
This needs more careful review on the impact of the install.sh verify feature also. I'll dig into this and report back what I find.

Stetsed:
> This needs more careful review on the impact of the install.sh verify feature also. I'll dig into this and report back what I find.

Hey! I just wanted to pop in here as I only recently saw this issue while scrolling through, and actually touches upon what I commented in https://github.com/anchore/syft/issues/585#issuecomment-5147575927

This will indeed require changes for the install script, the bigger problem I suspect  that due to immutable releases you will actually need to check what version it has, and if it’s a version that is post whenever this would be merged, to verify via bundle instead. Assuming that the intent would ofcourse be to actually deprecate the deprecated method. 

Because otherwise let’s say somebody tries to download vX.Y.Z which is before this, it would break because the install script in most cases(including sbom-action) is pulled from main(which has its own issues with interior mutability but I digress that’s for that repo).

So what would be needed:
1. Have a dedicated verifier function for new and the old system, which references the version of Syft being downloaded, or checks what files are available depending on how it’s implemented.
2. Generate the signatures in the new format which is implemented in this PR

For the first, this is contained inside of lines 644-656, as not only the command for checking (verify-sign, 592-615) need to be changed, but also what files it would attempt to download would need to be changed. 

Overall this would require a definitive design decision on how/if to handle the backwards compatibility. Currently as can be seen in the code it checks the version, aborting the install if attempting to install a version that doesn’t have co-sign signatures available. So I suppose expanding on this could be adding a vb or similar option, but that becomes a design choice.

Abort if version less then Y, before which no co-sign signatures where given
Allow if version between Y and X, where the now deprecated format was provided
Abort if version above X, where the new bundle format was provided

OR

Abort if version less than Y
Allow if version between Y and X, using the old format
Allow if version above X, and co-sign version above Z where bundle format is supported, using the new bundles

I do apologize if this is the wrong place to place such information, I do not often interact in this sort of way, and hope the above can be atleast somewhat helpful from my time looking into this stuff recently.

*Edit:* Edit after discussion in the developer discussion stream, seems like the way I read the install script was slightly flawed, missing the fact that it actually grabs the version belonging to the installed version. This would mean that number 1 above would *not* be needed, instead only needing to replace the new function with the old function, and the other places needing to be changed as noted above. 

The primary other change needed is to check the version of cosign, as bundle has only been supported since cosign 2.6(as noted in https://github.com/sigstore/cosign/releases/tag/v3.1.2). So what you would probally end up with:

1. Switch goreleaser to generate bundles from version X onwards
2. In the script, replace the verifier function with one that uses the bundle, and the download functions to prepare to download the bundle instead of the individual files
3. Check if cosign version is >= 2.6, if it is then check using the bundle, if it isn't then the program should exit stating that it doesn't support bundles with this version of cosign, as can be seen is also done in https://github.com/anchore/syft/blob/68da404bd7c8ac00630f322fa779fbf4b92f6deb/install.sh#L759-L765, except instead of just checking the existence of cosign, also check the version

PR #5104, opened 2026-07-25
fix(cyclonedx): preserve metadata.component name and version on decode

### LINKED ISSUE #2418: Main Component Version gets lost from CycloneDX when throwing Grype at it
**What happened**:

I have an SBOM in CycloneDX format. It contains the field metadata.component.version. When I call Grype to analyse this SBOM and output the result as CycloneDX, every single field of the original SBOM is included in the resulting document, except for metadata.component.version. (Timestamp and serial number are different, of course)

I would have expected for no field to vanish.

Go to your favourite repo.

Generate an SBOM with a component version:
```
syft . --output cyclonedx-json=./bom.cdx.json --source-version="0.1.0"
```

(Use a formatter for nice JSON output to see that the field is present)

Scan for vulnerabilities:
```
grype sbom:./bom.cdx.json --output cyclonedx-json=./vuln.cdx.json
```

The field is missing.

If this is indeed a bug and not a design choice, I am willing to attempt to take care of it. I will need some pointers though, because I have no clue where to start.

**Environment**:
- Output of `grype version`:
```
Application:         grype
Version:             0.87.0
BuildDate:           2025-01-22T20:51:16Z
GitCommit:           247f5d72abf2131aa37f3164a98495c121b29029
GitDescription:      v0.87.0
Platform:            linux/amd64
GoVersion:           go1.23.4
Compiler:            gc
Syft Version:        v1.19.0
Supported DB Schema: 5
```
```
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
```

jatinder14:
## Summary
- `extractComponents` now always copies `metadata.component` `name`/`version` into `source.Description`
- Container and file metadata paths still populate typed `Metadata`, and containers keep `ManifestDigest` from the component version
- Adds unit coverage for application/file/container round-trip preservation

Fixes https://github.com/anchore/grype/issues/2418

Previously, decoding a CycloneDX SBOM (e.g. from `syft . --source-version=0.1.0`) dropped `metadata.component.version`, so `grype sbom:... --output cyclonedx-json` emitted an SBOM without that field.

## Test plan
- [x] `go test ./syft/format/internal/cyclonedxutil/helpers/ -run Test_extractComponents_preservesNameAndVersion`
- [ ] Manual: `syft . --output cyclonedx-json=bom.cdx.json --source-version=0.1.0` then scan with Grype cyclonedx-json output and confirm `metadata.component.version` is retained (once Grype picks up this Syft fix)

### DIFF
--- a/syft/format/cyclonedxjson/decoder_test.go
+++ b/syft/format/cyclonedxjson/decoder_test.go
@@ -1,6 +1,7 @@
 package cyclonedxjson
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -11,6 +12,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/anchore/syft/syft/format/internal/testutil"
 	"github.com/anchore/syft/syft/sbom"
 )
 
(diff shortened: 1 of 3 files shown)


spiffcs (commented):
👋 Thanks for the PR @jatinder14

I pushed a few changes to the tests and added a couple cases that cover the new default path where we assert that name/version/supplier all flow through and are not dropped.

I'll wait for CI to pass here and then 🟢 or ask someone on @anchore/tools for a second pair of 👀.

PR #5106, opened 2026-07-26
fix(swift): guard slice access when pod entry has no version in Podfile.lock parser

AruneshDwivedi:
When parsing Podfile.lock entries, the pod string is split by spaces to extract name and version. A pod specified without a version caused an index-out-of-bounds panic on splits[1]. Added bounds check before accessing the version field.

Used AI to help identify and draft this fix — reviewed and tested before submitting.


wagoodman:
Thanks for the contribution -- can you add unit tests to show this is behaving as you'd expect?

CAOShurong (commented):
The fix correctly prevents a panic on malformed podfile.lock entries where a pod reference lacks a space-separated version (e.g. a bare pod name). The bounds checks are straightforward and preserve existing behavior for well-formed entries.

One minor observation: the \len(splits) < 1\ check on line 50 is unreachable since \strings.Split\ always returns at least one element (even for an empty input string). It could be removed without affecting correctness, but leaving it doesn't hurt.

PR #5107, opened 2026-07-26
fix(apkdb): Increase buffer size if an overly long field occurs in an ApkDB record

### LINKED ISSUE #5094: apk-db-cataloger silently drops the entire APK catalog when one installed-DB field exceeds 64 KB
**TL:DR**

`parseApkDB` [uses](https://github.com/anchore/syft/blob/c890e7f17f8db638d3f1558bf087b4cb2f3c2aad/syft/pkg/cataloger/alpine/parse_apk_db.go#L40) `bufio.Scanner` to read `/lib/apk/db/installed`.

However, by default, the scanner has a 64KB per-line limit. A single field longer than 64KB (in this case a large `p:`/provides line) causes [`scanner.Scan()`](https://github.com/anchore/syft/blob/c890e7f17f8db638d3f1558bf087b4cb2f3c2aad/syft/pkg/cataloger/alpine/parse_apk_db.go#L64) to stop with `bufio.ErrTooLong`.

This causes the function to [return `nil,nil,err`](https://github.com/anchore/syft/blob/c890e7f17f8db638d3f1558bf087b4cb2f3c2aad/syft/pkg/cataloger/alpine/parse_apk_db.go#L112) discarding any package that had already been successfully parsed.

Syft then seems to fall back to to ELF detection, so any package without a qualifying note (in my example, an APK shipping a Go package) silently disappears from the SBOM.

`syft` should still surface the catalogue of detected packages and/or be noisy about the parsing issue.  Ideally, all of the packages would be returned.

The image this was found with isn't public, but the following repro can be used

`create_sbom.py`:
```python
import sys

def rec(name, provides):
    return f"""C:Q1{name}AAAAAAAAAAAAAAAAAAAAAAAAA=
P:{name}
V:1.0.0-r0
A:x86_64
L:MIT
T:test package {name}
o:{name}
m:test
t:1700000000
S:100
I:200
p:{provides}
F:usr
R:{name}
Z:Q1{name}BBBBBBBBBBBBBBBBBBBBBBBBB="""

good = rec("alpha", "cmd:alpha=1.0.0-r0")
big  = "cmd:beta=1.0.0-r0 " + " ".join(f"so-ver:lib{i}.so=1.0.0-r0" for i in range(int(sys.argv[1])))
bad  = rec("beta", big)
open("repro/lib/apk/db/installed","w").write(good + "\n\n" + bad + "\n")
```

Repro issue
```sh
mkdir -p repro/lib/apk/db
python3 create_sbom.py 3000 # equates to a line that's 80KB

# trigger syft
syft dir:./repro -o json | jq '[.artifacts[] | select(.type=="apk") | .name]'

# output will be an empty array (i.e. [])
```

As a control group to prove the output is otherwise valid
```sh
rm -r  repro/lib/apk/db
mkdir -p repro/lib/apk/db
python3 create_sbom.py 30
# trigger syft
syft dir:./repro -o json | jq '[.artifacts[] | select(.type=="apk") | .name]'

#output:
[
  "alpha",
  "beta"
]
```

In the instance that triggered this, `p:` is long because the package ships a **lot** of shared libraries.

**Environment**:
- Output of `syft version`:
```
Application:   syft
Version:       1.38.0
BuildDate:     2025-11-17T17:42:49Z
GitCommit:     a033ae525f6c7ef937c6f49513e3403f07a1d6c0
GitDescription: v1.38.0
Platform:      linux/amd64
GoVersion:     go1.25.4
Compiler:      gc
SchemaVersion: 16.1.0
```

- OS (e.g: `cat /etc/os-release` or similar)
```
$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 24.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.4 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
```

bentasker:
## Description

* Attempt to read the ApkDB with the default 64KB max token length
* If this results in `ErrTooLong` tried again with a 512KB record
* Add test (`Test_parseApkDB_recoversFromOversizedField`) to exercise this

## Type of change

## Checklist

## Issue references

Fixes #5094

## Test

```sh
go build -o /tmp/syft-fixed ./cmd/syft
mkdir -p repro/lib/apk/db
python3 create_sbom.py 3000 # script from the linked issue

/tmp/syft-fixed dir:./repro -o json | jq '[.artifacts[] | select(.type=="apk") | .name]'
[
  "alpha",
  "beta"
]
```

### DIFF
--- a/syft/pkg/cataloger/alpine/parse_apk_db.go
+++ b/syft/pkg/cataloger/alpine/parse_apk_db.go
@@ -43,50 +43,50 @@ const maxAPKDBTokenSize = 512 * 1024
 //
 // nolint:funlen
 func parseApkDB(ctx context.Context, resolver file.Resolver, env *generic.Environment, reader file.LocationReadCloser) ([]pkg.Package, []artifact.Relationship, error) {
-        // Attempt to parse with the default scanner.
-        // If it encounters a field which exceeds the 64KB buffer size, rewind
-        // and rescan with a larger buffer
-        apks, errs, scanErr := scanApkDBEntries(reader, 0)
-        if errors.Is(scanErr, bufio.ErrTooLong) {
-                if seeker, ok := reader.ReadCloser.(io.Seeker); ok {
-                        if _, seekErr := seeker.Seek(0, io.SeekStart); seekErr == nil {
-                                apks, errs, scanErr = scanApkDBEntries(reader, maxAPKDBTokenSize)
-                        } else {
-                                log.Debugf("unable to rewind APK installed DB to re-parse with a larger buffer: %+v", seekErr)
-                        }
-                } else {
-                        log.Debug("APK installed DB contains an oversized field but the reader is not seekable; cannot re-parse with a larger buffer")
-                }
-        }
+	// Attempt to parse with the default scanner.
+	// If it encounters a field which exceeds the 64KB buffer size, rewind
+	// and rescan with a larger buffer
+	apks, errs, scanErr := scanApkDBEntries(reader, 0)
+	if errors.Is(scanErr, bufio.ErrTooLong) {
+		if seeker, ok := reader.ReadCloser.(io.Seeker); ok {
+			if _, seekErr := seeker.Seek(0, io.SeekStart); seekErr == nil {
+				apks, errs, scanErr = scanApkDBEntries(reader, maxAPKDBTokenSize)
+			} else {
+				log.Debugf("unable to rewind APK installed DB to re-parse with a larger buffer: %+v", seekErr)
+			}
+		} else {
+			log.Debug("APK installed DB contains an oversized field but the reader is not seekable; cannot re-parse with a larger buffer")
+		}
+	}
 
-        if scanErr != nil {
-                return nil, nil, fmt.Errorf("failed to parse APK installed DB file: %w", scanErr)
-        }
+	if scanErr != nil {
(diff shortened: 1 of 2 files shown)


syft/pkg/cataloger/alpine/parse_apk_db.go
```diff
@@ -32,15 +33,71 @@ type parsedData struct {
 	pkg.ApkDBEntry
 }
 
+// Maximum size of an APKDB line
+// 64KB is allocated by default, if that's too big then a buffer of maxAPKDBTokenSize
+// is used
+const maxAPKDBTokenSize = 512 * 1024
```
bentasker, line 39: Note: 512KB is a fairly arbitrary number.

The test case that I discovered had a `p:` of just over 100KB so this leaves room for growth without demanding excessive amounts of memory

bentasker:
Closing in favour of #5100, which resolves the reported symptom.

Noting for future reference/discoverability: #5100 raises the token limit but leaves the existing error path, so an oversized field beyond the new ceiling (or any other scanner error) will still discard the whole catalog rather than returning the packages parsed so far. If that surfaces later, this branch has an approach for it plus a regression test (though the token size would need bumping).

PR #5108, opened 2026-07-26
Fix missing nested packages in package-lock.json v1

### LINKED ISSUE #5101: package-lock.json v1: nested dependencies entries are never cataloged (flat top-level iteration only)
## What happened

On a lockfileVersion-1 `package-lock.json` (no `node_modules` present), syft catalogs only the
top-level `dependencies` entries. Nested entries — which npm creates exactly when a package needs a
*different version* than the hoisted one — are never visited, so every conflicting-version duplicate
is missing from the SBOM.

Real-world magnitude (Skyscanner/backpack-android @ a v1 lockfile): the lockfile declares **1,057
unique name@version pairs** across the full nested tree; syft emits exactly the **824 top-level
pairs** (with `SYFT_JAVASCRIPT_INCLUDE_DEV_DEPENDENCIES=true`; 10 without). The 233 missing pairs
are all nested conflicting versions (e.g. `@babel/code-frame@7.12.11` nested vs `7.16.7` hoisted).

Since CVE ranges are version-specific, a missing *version* can hide a real vulnerability match even
when the package name is cataloged at another version.

## Where

`syft/pkg/cataloger/javascript/parse_package_lock.go` — the v1 branch iterates
`lock.Dependencies` (the flat top level) and does not recurse into each entry's own
`dependencies` map. The v2/v3 branch is unaffected (the `packages` map is already flat).

## Expected

The v1 walk recurses (npm docs: nested entries appear precisely for conflicting-version installs,
so the unique install set is the union over every depth), deduplicating (name, version) pairs.

## Environment

syft 1.48.0 (also inspected on main). Reproduces with:
`syft scan dir:<checkout-with-v1-lockfile> -o syft-json` → compare artifact count against a
recursive parse of the same lockfile.

## Note

Happy to contribute a reproduction fixture (minimal v1 lockfile with one nested conflicting
version) if useful.

Eljees:
## Description

Catalog nested dependency entries in npm package-lock v1 files. Previously, Syft only visited the top-level dependency map, so conflicting versions installed below another package were missing from the SBOM.

This change:
- recursively walks nested v1 dependency maps
- keeps distinct package versions while deduplicating repeated name/version pairs
- skips complete dev-only subtrees when dev dependencies are disabled
- adds a multi-level regression fixture covering all three cases

## Type of change

## Testing

- `go test ./syft/pkg/cataloger/javascript -count=1`
- `go vet ./syft/pkg/cataloger/javascript`

## Checklist

## Issue references

Fixes #5101

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_package_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_package_lock.go
@@ -27,10 +27,11 @@ type packageLock struct {
 
 // lockDependency represents a single package dependency listed in the package.lock json file
 type lockDependency struct {
-	Version   string `json:"version"`
-	Resolved  string `json:"resolved"`
-	Integrity string `json:"integrity"`
-	Dev       bool   `json:"dev"`
+	Version      string                    `json:"version"`
+	Resolved     string                    `json:"resolved"`
+	Integrity    string                    `json:"integrity"`
+	Dev          bool                      `json:"dev"`
+	Dependencies map[string]lockDependency `json:"dependencies"`
 }
 
 type lockPackage struct {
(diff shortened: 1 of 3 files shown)


wagoodman (approved):

Eljees:
I'm glad could have helped!!

PR #5111, opened 2026-07-27
Add support for scanning IPK packages using the deb-archive-cataloger

skyhisi:
## Description

IPK packages are used by the opkg package manager, they have the same format as DEB packages, so the existing scanner can be used.

## Type of change

## Checklist


spiffcs (commented):
👋 Thanks for the PR @skyhisi. I think I have some questions before we move foward.

Does this update do what is actually suggests in the PR? 

Can you add some tests or show some steps that demonstrate syft working correctly with files that match the`*.ipk` glob?

Can the .ipk come as a gzipped tar like in this example and does syft handle this correctly with this change? 
```
$ curl -sLO https://downloads.openwrt.org/releases/23.05.5/packages/mips_24kc/base/6in4_28_all.ipk
$ file 6in4_28_all.ipk   → gzip compressed data
```
I don't think our current `AR` implementation would play nice with a match that comes as gzipped and might produce no package results as well as superfluous unknown entries.

I think we might need more complex handling of the compression for the different cases utilizing `mholt/archives`: see the original PR for context https://github.com/anchore/syft/pull/3704

For this PR to move forward:
- I think we need to do a bit more inquiry into the different compression formats `*.ipk` could come in, handling of those, and tests added showing that handling
- Glob tests added
- We probably want the team to chime in here as well to see if there are any other considerations for syft's surface area on if the opkg package manager should be lumped under the deb archive cataloger like this, or if it deserves a spot carved out on it's own or mentioned in some way.

skyhisi:
@spiffcs Thanks for taking an initial look at this PR, I didn't realise the OpenWrt version of IPK used a gzip outer container, I'm using it with packages built by Yocto which follows the Debian format of having an outer AR file containing compressed tar files, which it works fine with.

I can look into adding some tests for the globbing, but from using it, it works fine.

I haven't tried it with OpenWrt packages and given that the [OpenWrt project has moved to using APK](https://forum.openwrt.org/t/openwrt-25-12-0-stable-release/247228#p-1286229-switch-package-manager-from-opkg-to-apk-4) (Alpine format) packages, I'm not sure if it's worth the syft project supporting the deprecated IPK format? Maybe it'd be enough if it skipped them?

skyhisi:
I've updated the PR with some unit tests now.

syft/pkg/cataloger/debian/testdata/image-single-ipk/libz1_1.3.2-r0_x86-64-v3.ipk
wagoodman, line 1: the dockerfile-based tests you have should be sufficient, and ideally we should not be checking in the binary ipk files (even though it is relatively small).

Why not just use one docker fixture?...
```
rm Dockerfile
mv Dockerfile.build Dockerfile
```
... it looks like you would still be covered (shout out if I'm missing something)

skyhisi: `Dockerfile.build` builds the IPK package using Yocto which can take quite a long time (20-30 minutes) as Yocto will compile the full cross-compiler toolchain first before making packages for a target, so that's why I added a pre-built IPK, with the `Dockerfile` used by the test adding the prebuilt IPK.

I thought it best to include the `Dockerfile.build` in case someone wants to rebuild the IPK in the future, or see how it was built.

Issue #5075, opened 2026-07-17
Incorrect CPE values


Jaydavis00:
**What happened**: Incorrect CPE values produced do not match the NIST Vulnerability Database

**What you expected to happen**: The correct CPE values to be produced

**Steps to reproduce the issue**:
.\syft dir:"C:\Program Files" `
  -o cyclonedx-json@1.6 `
  --catalogers all `
  | Out-File C:\SBOM\program-files.json -Encoding utf8

**Anything else we need to know?**:
No

**Environment**:
Windows 10

A few examples below
Google Chrome Version 115
Syft generates a CPE value of: "cpe:2.3:a:Google_Chrome:Google_Chrome:115.0.5790.171:*:*:*:*:*:*:*"
NIST Vulnerability Database uses: "CPE: cpe:2.3:a:google:chrome:115.0.5790.171:*:*:*:*:*:*:*"

Abode Acrobat Reader 25.1.204.35
Syft: "cpe:2.3:a:Adobe_Acrobat:Adobe_Acrobat:25.1.20435.0:*:*:*:*:*:*:*"
NVD: cpe:2.3:a:adobe:acrobat_reader_dc:25.1.20435:*:*:*:continuous:*:*:*

Microsoft Excel 16
Syft: "cpe:2.3:a:Microsoft_Excel:Microsoft_Excel:16.0.10416.20073:*:*:*:*:*:*:*"
NVD: “cpe:2.3:a:microsoft:excel:2016:*:*:*:*:*:x64:*”

Microsoft Word 16
Syft: "cpe:2.3:a:Microsoft_Word:Microsoft_Word:16.0.10351.20054:*:*:*:*:*:*:*"
NVD: “cpe:2.3:a:microsoft:word:2016:*:*:*:*:*:x64:*”

VLC Media Player 3.0.21
Syft: "cpe:2.3:a:VLC_media_player:VLC_media_player:3\\,0\\,21\\,0:*:*:*:*:*:*:*"
NVD: cpe:2.3:a:videolan:vlc_media_player:3.0.21:*:*:*:*:*:*:*

kzantow:
[...]

PR #5112, opened 2026-07-27
fix(binary): normalize comma-separated PE version resources

Eljees:
## Summary

Normalizes the comma-separated version form found in Windows PE `VERSIONINFO` resources, so `3, 0, 21, 0` becomes `3.0.21.0`.

## Why

Windows resource string tables frequently mirror the `.rc` declaration (`FILEVERSION 3,0,21,0`) rather than the dotted form. Syft used that value verbatim as the package version, which is visible in [#5075](https://github.com/anchore/syft/issues/5075) — VLC is reported as:

```
cpe:2.3:a:VLC_media_player:VLC_media_player:3\,0\,21\,0:*:*:*:*:*:*:*
```

A version containing escaped commas matches nothing in vulnerability data, and the same value flows into the PURL and the package version shown to users.

This only addresses the version portion of that issue; the vendor/product naming discussion there is separate and left alone.

## Details

- `extractVersionFromResourcesValue` now normalizes a value that consists entirely of comma-separated numbers (with optional surrounding spaces) to the dotted form.
- Anything else is returned unchanged, so values like `1.2.3 (build 456)` keep their existing handling and a value such as `1.0, beta` is not rewritten.

## Tests

Added `Test_findVersionFromVR` covering dotted input, both comma-separated forms (`3, 0, 21, 0` and `3,0,21,0`), a version with a trailing comment, and a non-numeric comma value. `go test ./syft/pkg/cataloger/binary/` passes.

### DIFF
--- a/syft/pkg/cataloger/binary/pe_package.go
+++ b/syft/pkg/cataloger/binary/pe_package.go
@@ -15,6 +15,10 @@ var (
 	// spaceRegex includes nbsp (#160) considered to be a space character
 	spaceRegex  = regexp.MustCompile(`[\s\xa0]+`)
 	numberRegex = regexp.MustCompile(`\d`)
+	// commaSeparatedVersionRegex matches the comma-separated form that Windows
+	// VERSIONINFO resources often carry (mirroring the FILEVERSION x,y,z,w
+	// declaration in a .rc file), e.g. "3, 0, 21, 0"
+	commaSeparatedVersionRegex = regexp.MustCompile(`^\d+(?:\s*,\s*\d+)+$`)
 )
 
 func newPEPackage(versionResources map[string]string, f file.Location) pkg.Package {
@@ -129,6 +133,7 @@ func findVersionFromVR(versionResources map[string]string) string {
 
 func extractVersionFromResourcesValue(version string) string {
 	version = strings.TrimSpace(version)
+	version = normalizeCommaSeparatedVersion(version)
 	out := ""
 	for i, f := range strings.Fields(version) {
 		if containsNumber(out) && !containsNumber(f) {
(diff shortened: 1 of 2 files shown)


Eljees:
Ping on this one and #5114 - both open since late July, no review yet. They touch different areas (binary PE version resources and the CycloneDX decoder) and are independent.

Eljees:
Pushed a follow-up commit (`99d5911`) that closes a gap I found while re-checking this against today's `main`.

The original change only rewrote values that were **entirely** comma-separated numbers. That misses the spelling most common on Microsoft-built binaries, where a build tag follows the version. Measured on `main` at `3325493` and on this branch:

| `ProductVersion` / `FileVersion` | `main` | this PR, before | this PR, now |
|---|---|---|---|
| `3, 0, 21, 0` | `3, 0, 21, 0` | `3.0.21.0` | `3.0.21.0` |
| `6, 1, 7601, 17514` | `6, 1, 7601, 17514` | `6.1.7601.17514` | `6.1.7601.17514` |
| `1, 0, 0, 1 (WinBuild.160101.0800)` | unchanged | **unchanged** | `1.0.0.1 (WinBuild.160101.0800)` |
| `2, 5, 0, 0 built by: WinDDK` | `2, 5, 0, 0` | `2, 5, 0, 0` | `2.5.0.0` |
| `10.0.19041.1` | `10.0.19041.1` | `10.0.19041.1` | `10.0.19041.1` |
| `1.0, beta` | `1.0,` | `1.0,` | `1.0,` |

The fix is to rewrite a **leading** run of comma-separated numbers and leave the rest of the string alone, rather than requiring the whole value to match. The deliberate boundary: `1, 0, 0, 1 (WinBuild.160101.0800)` now becomes `1.0.0.1 (WinBuild.160101.0800)` — it keeps the build tag, because that is exactly what the dotted `10.0.19041.1 (WinBuild.160101.0800)` already does when it reaches this function. Whether that trailing tag should be dropped is a separate question about existing behaviour, and I have left it alone rather than widen this PR; the point here is only that the comma spelling stops being treated worse than the dotted one.

Two test cases were added for the new rows; the existing ones are unchanged and still pass.

Verification, since the branch was cut in August: it merges into current `main` cleanly, and `go test ./syft/pkg/cataloger/binary/...` on the merge result fails exactly the same tests as a pristine `main` checkout does in this environment (`Test_Cataloger_PositiveCases`, `Test_PEPackageCataloger`, `Test_ELFPackageCataloger`, `Test_JavaBinaryImage`, `Test_Cataloger_DefaultClassifiers_PositiveCases_Image` — all of them need `docker` to build fixtures, which is not available here). Nothing new is attributable to the change. CI here sits at `action_required`, so it needs the usual approval click for an outside contributor before it will run.

PR #5117, opened 2026-07-28
fix: Cleanup snap temporary directories

spiffcs:
## Description

`newSnapFileFromRemote` creates a `syft-snap-*` temp directory. The cleanup closure only fired after the download succeeded. An error during download fails to cleanup the temporary artifacts. This could be caused by a connection reset, timeout, or storage error mid-transfer leaves a partial snap payload on disk.

The closure is now created immediately after `afero.TempDir` succeeds, and both error paths invoke it through a small helper that logs (rather than discards) a removal failure. 

The success path still returns the same closure as `snapFile.Cleanup`.

This PR does not have any CLI output or configuration changes. The only observable difference is that failed remote snap scans no longer leave `syft-snap-*` directories in the temp filesystem.

### Reproduction of bug on OSX

"No space left on device" is part of the error path that did not cleanup the tmp creations.
```
bash-3.2$ DEV=$(hdiutil attach -nomount ram://16384 | tr -d '[:space:]')
bash-3.2$ diskutil eraseVolume HFS+ syftleak "$DEV"
Started erase on disk6
Unmounting disk
Erasing
Initialized /dev/rdisk6 as a 8 MB case-insensitive HFS Plus volume
Mounting disk
Finished erase on disk6 (syftleak)
bash-3.2$ TMPDIR=/Volumes/syftleak syft scan etcd@stable --from snap -o json >/dev/null
A newer version of syft is available for download: 1.49.0 (installed version is 1.46.0)
[0022] ERROR could not determine source: errors occurred attempting to resolve 'etcd@stable':
  - snap: snap file "etcd@stable" does not exist
  - snap: failed to download snap file: failed to download snap file at "https://api.snapcraft.io/api/v1/snaps/download/TKebVGcPeDKoOqAmNmczU2oWLtsojKD5_249.snap": write /Volumes/syftleak/syft-snap-949299180/TKebVGcPeDKoOqAmNmczU2oWLtsojKD5_249.snap: no space left on device

bash-3.2$ find /Volumes/syftleak -maxdepth 1 -name 'syft-snap-*'
/Volumes/syftleak/syft-snap-949299180
bash-3.2$ find /Volumes/syftleak -name '*.snap' -exec ls -lh {} \;
-rw-r--r--@ 1 7.8M Jul 28 12:50 /Volumes/syftleak/syft-snap-949299180/TKebVGcPeDKoOqAmNmczU2oWLtsojKD5_249.snap
```

### PR Verification
- go test ./syft/source/snapsource/..., ./syft/source/..., ./cmd/... — pass
- go build ./..., go vet, .tool/golangci-lint run ./syft/source/snapsource/... — clean

### Not exercised
A real interrupted transfer against the snapcraft API. The failure modes are covered at the unit level via a mocked getter.

Type of change

Checklist

### DIFF
--- a/syft/source/snapsource/snap.go
+++ b/syft/source/snapsource/snap.go
@@ -77,7 +77,7 @@ func getRemoteSnapFile(ctx context.Context, fs afero.Fs, getter intFile.Getter, 
 	return newSnapFileFromRemote(ctx, fs, cfg, getter, info)
 }
 
-func newSnapFileFromRemote(ctx context.Context, fs afero.Fs, cfg Config, getter intFile.Getter, info *remoteSnap) (*snapFile, error) {
+func newSnapFileFromRemote(ctx context.Context, fs afero.Fs, cfg Config, getter intFile.Getter, info *remoteSnap) (_ *snapFile, err error) {
 	t, err := afero.TempDir(fs, "", "syft-snap-")
 	if err != nil {
 		return nil, fmt.Errorf("failed to create temp directory: %w", err)
@@ -87,23 +87,25 @@ func newSnapFileFromRemote(ctx context.Context, fs afero.Fs, cfg Config, getter 
 		return fs.RemoveAll(t)
 	}
 
-	// any failure past this point must not leave the temp directory (or a partially downloaded snap) behind
-	cleanupOnErr := func() {
-		if err := closer(); err != nil {
-			log.WithFields("error", err, "directory", t).Warn("unable to remove temp directory for snap")
+	// any failure past this point must not leave the temp directory (or a partially downloaded snap)
+	// behind. on success the caller owns the directory by way of snapFile.Cleanup.
+	defer func() {
+		if err == nil {
+			return
 		}
-	}
+		if closeErr := closer(); closeErr != nil {
+			log.WithFields("error", closeErr, "directory", t).Warn("unable to remove temp directory for snap")
+		}
+	}()
 
 	snapFilePath := path.Join(t, path.Base(info.URL))
 	err = downloadSnap(getter, info, snapFilePath)
 	if err != nil {
-		cleanupOnErr()
 		return nil, fmt.Errorf("failed to download snap file: %w", err)
 	}
 
 	mimeType, digests, err := getSnapFileInfo(ctx, fs, snapFilePath, cfg.DigestAlgorithms)
 	if err != nil {
-		cleanupOnErr()
 		return nil, err
 	}
 
--- a/syft/source/snapsource/snap.go
+++ b/syft/source/snapsource/snap.go
@@ -83,18 +83,27 @@ func newSnapFileFromRemote(ctx context.Context, fs afero.Fs, cfg Config, getter 
 		return nil, fmt.Errorf("failed to create temp directory: %w", err)
 	}
 
-	snapFilePath := path.Join(t, path.Base(info.URL))
-	err = downloadSnap(getter, info, snapFilePath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to download snap file: %w", err)
-	}
-
 	closer := func() error {
 		return fs.RemoveAll(t)
 	}
 
+	// any failure past this point must not leave the temp directory (or a partially downloaded snap) behind
+	cleanupOnErr := func() {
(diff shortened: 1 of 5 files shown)


joshbressers:
This bug was reported to us by https://github.com/volksec

They asked why this particular case differs from CVE-2026-33481 which was also a temporary file issue fixed in Syft

CVE-2026-33481 is a somewhat different case than this. What we were really protecting with CVE-2026-33481 was a zip bomb style of attack. If an attacker had syft extract something with a very high compression ratio, it would fill the temporary space immediately and not clean up on exit. The fix was to clean up the temporary space when this happened rather than try to prevent the zip bomb from happening. In this instance it's not something an attacker can readily control and it's not an attack that will immediately fill the storage space and fail.

wagoodman (commented):
The fix looks functional, however, there is a sibling leak that is still present. https://github.com/anchore/syft/blob/29fd7d0dec81cf03e0a1194a1985c7c893bb2396/syft/source/snapsource/snap_source.go#L122 (outside of the current review).

This returns a non-nil source alongside a non-nil error, so callers are at risk (if they are being idiomatic) of not calling the `snapSource.Close()`.

syft/source/snapsource/snap.go
```diff
@@ -83,18 +83,27 @@ func newSnapFileFromRemote(ctx context.Context, fs afero.Fs, cfg Config, getter
...
+	closer := func() error {
+		return fs.RemoveAll(t)
+	}
+
+	// any failure past this point must not leave the temp directory (or a partially downloaded snap) behind
+	cleanupOnErr := func() {
+		if err := closer(); err != nil {
+			log.WithFields("error", err, "directory", t).Warn("unable to remove temp directory for snap")
+		}
+	}
```
wagoodman: Why not defer this instead?

```suggestion
	// any failure past this point must not leave the temp directory (or a partially downloaded snap) behind
	defer func() {
		if err := closer(); err != nil {
			log.WithFields("error", err, "directory", t).Warn("unable to remove temp directory for snap")
		}
	}()
```

spiffcs: We can defer. I'll update this with the suggestion.

volksec:
Thanks for the clarification. I agree that this case differs from CVE-2026-33481 because it does not provide the same immediate, high-amplification disk exhaustion primitive as a zip bomb.

My comparison was limited to the resulting cleanup failure: temporary data remains on disk after an error path.

The concern in this case is cumulative resource exhaustion rather than immediate exhaustion. In long-running services, persistent CI workers, or automated pipelines, repeated remote Snap download failures can leave multiple syft-snap-* directories and partial files behind over time.

The degree of attacker control is deployment-dependent. However, in scenarios where an untrusted user can submit or influence a remote Snap URL, an attacker-controlled endpoint could send partial content and repeatedly terminate the connection, causing Syft to follow the affected error path.

So I agree that this should not be treated as equivalent in severity to CVE-2026-33481. My view is that it remains a valid improper-cleanup issue with a lower and more context-dependent security impact.

Thank you for taking the time to explain the distinction and for addressing this so quickly. 

I really appreciate the effort you and the team put into reviewing and resolving it.

spiffcs:
@wagoodman I have addressed all the initial comments an re requested review. I've left them unresolved for you to look at for a final pass.

syft/source/snapsource/snap_source.go
```diff
@@ -119,7 +119,20 @@ func newFromPath(cfg Config, f *snapFile) (source.Source, error) {
...
-	return s, s.extractManifest()
+	if err := s.extractManifest(); err != nil {
+		// a later source provider may still resolve this path, in which case this error is dropped, so
+		// report the reason the snap was rejected here
+		log.WithFields("error", err, "path", f.Path).Warn("unable to read snap")
+
+		// never hand back a source alongside an error: callers that discard the source on error would
+		// leave behind the open squashfs file and the temp directory holding a downloaded snap
+		if closeErr := s.Close(); closeErr != nil {
+			log.WithFields("error", closeErr, "path", f.Path).Warn("unable to clean up snap source")
```
wagoodman: these should not be log level warn, right? there is no action for a user to take. This should be debug level information I think

wagoodman (approved):

PR #5122, opened 2026-07-29
fix: use safe type assertion in SafeTensors metadata access

lxcxjxhx:
## Summary

Fix unsafe type assertion in \ssembleSafeTensorsPackage\ that could panic if metadata type is not \pkg.SafeTensorsModelInfo\.

## Changes

- Replace direct type assertion \merged.Metadata.(pkg.SafeTensorsModelInfo)\ with safe comma-ok pattern
- Add fallback log message when metadata type is unexpected
- Prevents potential panic in production when processing malformed SafeTensors packages

## Context

The function previously accessed \merged.Metadata.(pkg.SafeTensorsModelInfo).MetadataHash\ directly, which would panic if the metadata was not of the expected type. Using the comma-ok pattern (\,ok\) is the idiomatic Go way to handle type assertions safely.

---

I apologize for any inconvenience. Local environment limitations prevent full dynamic testing; relying on CI/CD automated tests for verification.

### DIFF
--- a/syft/pkg/cataloger/ai/processor.go
+++ b/syft/pkg/cataloger/ai/processor.go
@@ -154,8 +154,12 @@ type safeTensorsIdentity struct {
 func assembleSafeTensorsPackage(merged pkg.Package, id safeTensorsIdentity) (pkg.Package, bool) {
 	name := pickSafeTensorsName(id.nameOrPath, id.fallbackName)
 	if name == "" {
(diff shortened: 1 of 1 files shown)


spiffcs:
@lxcxjxhx thanks for the PR. Can you sign the DCO before we take a look?
https://github.com/anchore/syft/pull/5122/checks?check_run_id=90466277830

PR #5143, opened 2026-08-01
fix(dotnet): correct inverted dependency-of relationship direction in packages.lock.json parser

### LINKED ISSUE #5125: Dotnet: Incorrect relationship graph in case of using package locks
**What happened**:
parse_packages_lock.go creates relationship
```
rel := artifact.Relationship{
				From: parentPkg,
				To:   childPkg,
				Type: artifact.DependencyOfRelationship,
}
```
Where parentPkg is actually package which depends on childPkg, but relationship type is DependencyOfRelationship.
It then leads to incorrect dependecy graph during sbom model construction in cyclonedxhelpers.ToFormatModel.
It creates inverse graph

```
toRef := helpers.DeriveBomRef(toPkg)
dep := dependencies[toRef]
if dep == nil {
	dep = &cyclonedx.Dependency{
		Ref:          toRef,
		Dependencies: &[]string{},
	}
	dependencies[toRef] = dep
}

fromRef := helpers.DeriveBomRef(fromPkg)
if !slices.Contains(*dep.Dependencies, fromRef) {
	*dep.Dependencies = append(*dep.Dependencies, fromRef)
}
```

It seems that correct version should be

```
rel := artifact.Relationship{
	From: childPkg,
	To:   parentPkg,
	Type: artifact.DependencyOfRelationship,
}
```

pujitha24:
Motivation:
parseDotnetPackagesLock built each artifact.Relationship as
{From: parentPkg, To: childPkg, Type: DependencyOfRelationship}, where
parentPkg is the package that depends on childPkg. Every other cataloger
in this repo that emits DependencyOfRelationship (golang/parse_go_mod.go,
cpp/parse_conanlock.go, dotnet/deps_binary_cataloger.go) does the
opposite: From is the dependency, To is the package that depends on it.
Because cyclonedxhelpers.ToFormatModel builds its "dependencies" graph by
appending each relationship's From-ref onto the To-ref's dependency list,
the inverted direction produced a backwards CycloneDX dependency graph
for any project cataloged from a packages.lock.json file (no crash, no
data loss — just an inverted graph edge for this one lock-file format).

Approach:
Swap From/To in the single relationship construction in
parse_packages_lock.go so it matches the repo-wide convention
(From: childPkg, To: parentPkg), and update the expected relationships
in parse_packages_lock_test.go to match the fixture's actual dependency
directions (e.g. AutoMapper depends on Microsoft.Extensions.Options, so
the relationship is now From: extensionOptionsPkg, To: autoMapperPkg).

Validation:
- go build ./...  (passes)
- go test ./syft/pkg/cataloger/dotnet/... -run 'TestParseDotnetPackagesLock|Test_corruptDotnetPackagesLock' -v  (both pass)
- go test ./syft/pkg/cataloger/dotnet/...  still fails on unrelated
  TestDotnetDepsCataloger_regressions/TestCataloger subtests because
  this sandbox has no docker binary; confirmed via `git stash` that
  those same subtests fail identically on unmodified code, so they are
  a pre-existing environment limitation, not a regression from this change.

Fixes #5125

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/syft/pkg/cataloger/dotnet/parse_packages_lock.go
+++ b/syft/pkg/cataloger/dotnet/parse_packages_lock.go
@@ -101,8 +101,8 @@ func parseDotnetPackagesLock(_ context.Context, _ file.Resolver, _ *generic.Envi
 			}
 
 			rel := artifact.Relationship{
-				From: parentPkg,
-				To:   childPkg,
+				From: childPkg,
+				To:   parentPkg,
 				Type: artifact.DependencyOfRelationship,
 			}
 			relationships = append(relationships, rel)
--- a/syft/pkg/cataloger/dotnet/parse_packages_lock_test.go
+++ b/syft/pkg/cataloger/dotnet/parse_packages_lock_test.go
@@ -169,28 +169,28 @@ func TestParseDotnetPackagesLock(t *testing.T) {
 
 	expectedRelationships := []artifact.Relationship{
 		{
-			From: autoMapperPkg,
+			From: extensionOptionsPkg,
+			To:   autoMapperPkg,
+			Type: artifact.DependencyOfRelationship,
+		},
+		{
+			From: dependencyInjectionAbstractionsPkg,
 			To:   extensionOptionsPkg,
-			Type: artifact.DependencyOfRelationship,
-		},
-		{
-			From: extensionOptionsPkg,
-			To:   dependencyInjectionAbstractionsPkg,
-			Type: artifact.DependencyOfRelationship,
-		},
-		{
-			From: extensionOptionsPkg,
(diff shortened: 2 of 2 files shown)


pujitha24:
This is still rebased and green whenever someone has a moment to take a look — happy to make any changes that would help review go faster.

apkrymov:
+1 We also need that PR

ivanprosh:
+1 Looking forward to fix

spiffcs:
Thanks for the PR @pujitha24 and apologies for the delay. Taking a look at this now 😄

spiffcs (approved):

PR #5144, opened 2026-08-01
fix: don't panic when redact.Set replaces an existing store

### LINKED ISSUE #2285: Syft cli.Command() panics when invoked multiple times
**What happened**:
See https://stackoverflow.com/questions/77387892/unable-to-call-anchore-syft-library-command-multiple-times-when-embedded-in-go-a

```golang
func generateCycloneDX(ociArchiveName string, jsonOutputName string) {
    syftId := clio.Identification{Name: "syft"}
    syftCommand := cli.Command(syftId)
    syftCommand.SetArgs([]string{ociArchiveName, "-o", jsonOutputName})
    err := syftCommand.Execute()
    if err != nil {
        panic(err)
    }
}
```

Which results in:
```
panic: replace existing redaction store (probably unintentional)

goroutine 1 [running]:
github.com/anchore/syft/internal/redact.Set(...)
        go/pkg/mod/github.com/anchore/syft@v0.93.0/internal/redact/redact.go:11
github.com/anchore/syft/cmd/syft/cli.create.func2(0xc000490a90?)
        go/pkg/mod/github.com/anchore/syft@v0.93.0/cmd/syft/cli/cli.go:64 +0x1a5
github.com/anchore/clio.(*application).runInitializers(0xc0013bc1a0)
        go/pkg/mod/github.com/anchore/clio@v0.0.0-20231016125707-b60d41410795/application.go:110 +0x66
github.com/anchore/clio.(*application).PostLoad(0xc0013bc1a0)
        go/pkg/mod/github.com/anchore/clio@v0.0.0-20231016125707-b60d41410795/application.go:105 +0xbb
github.com/anchore/fangs.postLoad({0x1f81f40?, 0xc0013bc1a0?, 0xc0013bc1a0?})
        go/pkg/mod/github.com/anchore/fangs@v0.0.0-20230818131516-2186b10924fe/load.go:201 +0x1e5
github.com/anchore/fangs.loadConfig({{0x26109f8, 0x349e4e0}, {0x1ff3df6, 0x4}, {0x2004cda, 0xc}, {0x0, 0x0}, {0xc002b25e30, 0x5, ...}}, ...)
        go/pkg/mod/github.com/anchore/fangs@v0.0.0-20230818131516-2186b10924fe/load.go:80 +0x7d1
github.com/anchore/fangs.Load({{0x26109f8, 0x349e4e0}, {0x1ff3df6, 0x4}, {0x2004cda, 0xc}, {0x0, 0x0}, {0xc002b25e30, 0x5, ...}}, ...)
        go/pkg/mod/github.com/anchore/fangs@v0.0.0-20230818131516-2186b10924fe/load.go:16 +0x74
github.com/anchore/clio.(*application).loadConfigs(0xc0013bc1a0, 0xc000033870?, {0xc0004909f0, 0x1, 0xc0013b2700?})
        go/pkg/mod/github.com/anchore/clio@v0.0.0-20231016125707-b60d41410795/application.go:95 +0x1a5
github.com/anchore/clio.(*application).setupCommand.func1.(*application).Setup.func1(0x4?, {0xd631f2?, 0xc0013b2700?, 0xc000033af0?})
        go/pkg/mod/github.com/anchore/clio@v0.0.0-20231016125707-b60d41410795/application.go:74 +0x45
github.com/anchore/clio.(*application).setupCommand.func1(0xc0013b2700?, {0xc002e20870, 0x1, 0x3})
        go/pkg/mod/github.com/anchore/clio@v0.0.0-20231016125707-b60d41410795/application.go:316 +0x82
github.com/spf13/cobra.(*Command).execute(0xc000845200, {0xc002e20660, 0x3, 0x3})
        go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:925 +0x7f6
github.com/spf13/cobra.(*Command).ExecuteC(0xc000845200)
        go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3a5
github.com/spf13/cobra.(*Command).Execute(...)
        go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
```

The panic is well intentioned, however, it is unintentional that it this panics on the second invocation.

ankit090701:
## Description

`redact.Set` panicked if it was called a second time in the same process. This breaks a documented, intended use case: embedding syft as a library and constructing/executing `cli.Command(id)` more than once (see the linked issue and https://stackoverflow.com/questions/77387892/unable-to-call-anchore-syft-library-command-multiple-times-when-embedded-in-go-a).

`AppClioSetupConfig` (`cmd/syft/internal/clio_setup_config.go`) registers a clio initializer that calls `redact.Set(state.RedactStore)` on every `Execute()`. Each call to `cli.Command(id)` gets its own freshly-constructed `clio.State`, and thus its own new `RedactStore` - so the *second* time a consumer executes a command in the same process, `redact.Set` always panicked with `"replace existing redaction store (probably unintentional)"`.

Looking at the sibling singleton packages in `internal/`, `bus.Set` and `log.Set` both just overwrite their package-level singleton on repeated calls, no questions asked - `redact.Set` is the only one of the three that panics on replace. This PR brings it in line with its siblings.

I confirmed this is still present on current `main` by reproducing the exact stack trace from the issue via `cli.Command(id)` + `Execute()` called twice in a row (executing the `scan` subcommand against a temp dir, twice) - it panics at `internal/redact/redact.go:11` through the identical code path described in the issue (`clio.(*application).runInitializers` -> `PostLoad` -> `fangs.postLoad`/`loadConfig` -> `setupCommand` -> cobra `Execute`).

## Type of change

## Checklist

## Testing

- Added `TestSet_ReplacingExistingStoreDoesNotPanic` (`internal/redact/redact_test.go`), which sets a store, adds a redaction, replaces the store, and confirms: (a) no panic, (b) the new store doesn't inherit redactions from the replaced one, (c) the new store works normally going forward.
- Verified by reverting just the `redact.go` change: the test fails reproducing the exact panic message from this issue.
- Additionally verified end-to-end (not included in this PR - a temporary local test) by calling `cli.Command(id)` twice, executing `scan dir:<tmp> -o json` both times: panics with the identical stack trace as the issue on unmodified code; passes cleanly with the fix.
- Ran the full `internal/...` and `cmd/syft/...` suites. The only failures are pre-existing and environmental (this sandbox has no `docker` or `zip` binaries available, and runs as root which bypasses a permission-denied test in `internal/cache`) - unrelated to `internal/redact`, which has no relationship to any of those packages.

## Issue references

Fixes #2285

### DIFF
--- a/internal/redact/redact.go
+++ b/internal/redact/redact.go
@@ -4,12 +4,12 @@ import "github.com/anchore/go-logger/adapter/redact"
 
 var store redact.Store
 
+// Set replaces the package-level redaction store. Library consumers that construct and execute a syft
+// command more than once within the same process (each call getting its own fresh clio state, and thus
+// its own redaction store) will legitimately call Set again; that's expected and simply swaps in the new
+// store, consistent with how the sibling bus and log singleton packages behave on repeated Set calls.
+// Redactions added to a store before it's replaced no longer apply to output produced afterwards.
 func Set(s redact.Store) {
-	if store != nil {
-		// if someone is trying to set a redaction store and we already have one then something is wrong. The store
-		// that we're replacing might already have values in it, so we should never replace it.
-		panic("replace existing redaction store (probably unintentional)")
-	}
 	store = s
 }
 
--- a/internal/redact/redact_test.go
+++ b/internal/redact/redact_test.go
@@ -1,0 +1,34 @@
+package redact
+
+import (
+	"testing"
+
+	gologgerredact "github.com/anchore/go-logger/adapter/redact"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestSet_ReplacingExistingStoreDoesNotPanic is a regression test for
+// https://github.com/anchore/syft/issues/2285: library consumers that
+// construct and execute a syft command more than once in the same process
+// (e.g. https://github.com/anchore/syft/blob/main/cmd/syft/internal/clio_setup_config.go
+// calling redact.Set on every clio initializer run) used to panic on the
+// second call, because Set refused to replace an already-set store.
+func TestSet_ReplacingExistingStoreDoesNotPanic(t *testing.T) {
+	orig := store
+	defer func() { store = orig }()
+
+	Set(gologgerredact.NewStore())
+	Add("first-secret")
+	assert.Equal(t, "prefix ******* suffix", Apply("prefix first-secret suffix"))
+
+	assert.NotPanics(t, func() {
+		Set(gologgerredact.NewStore())
+	})
+
+	// the replaced store doesn't carry over redactions registered before the swap
+	assert.Equal(t, "prefix first-secret suffix", Apply("prefix first-secret suffix"))
+
+	// but works normally going forward
+	Add("second-secret")
+	assert.Equal(t, "prefix ******* suffix", Apply("prefix second-secret suffix"))
+}


spiffcs:
Thanks for the PR and linking the issue @ankit090701. I'm going to go check and see if we can get a fix done in CLO.

Yes the panic is entirely syft's (clio never panics), but the forced re-Set is a clio design consequence.

`clio.New` returns `redact.NewStore()` per application `(clio/application.go:54)`, so every Execute() hands syft a different store to hoist into a global. 

If we update clio and flip the ownership then the problem disappears instead of being papered over.

Something like:
```golang
// setup_config.go
type SetupConfig struct {
      ...
      RedactStore redact.Store
}

func (c *SetupConfig) WithRedactStore(s redact.Store) *SetupConfig {
      c.RedactStore = s
      return c
}
```

cc @kzantow for feedback on the potential clio change vs this PR landing in syft. If 👍 just say so and I'll take it on.

internal/redact/redact.go
```diff
@@ -4,12 +4,12 @@ import "github.com/anchore/go-logger/adapter/redact"
...
+// Set replaces the package-level redaction store. Library consumers that construct and execute a syft
+// command more than once within the same process (each call getting its own fresh clio state, and thus
+// its own redaction store) will legitimately call Set again; that's expected and simply swaps in the new
+// store, consistent with how the sibling bus and log singleton packages behave on repeated Set calls.
+// Redactions added to a store before it's replaced no longer apply to output produced afterwards.
 func Set(s redact.Store) {
-	if store != nil {
-		// if someone is trying to set a redaction store and we already have one then something is wrong. The store
-		// that we're replacing might already have values in it, so we should never replace it.
-		panic("replace existing redaction store (probably unintentional)")
```
kzantow, line 11: This panic is intentional: if a store is replaced, there could be values that should be redacted which no longer get redacted. Removing this panic is not the right fix.

Based on the stack overflow link, it seems like you are using the CLI directly, but you should probably be using [Syft as a library instead](https://github.com/anchore/syft/blob/main/syft/create_sbom.go#L31).

kzantow (changes requested):

CAOShurong:
Data-backed input on the clio-ownership question @spiffcs raised, plus independent verification of this PR's mechanics (run 2026-08-25, Windows/amd64, go1.26.7).

**Verification**

- Minimal two-`Set` repro against `main @ bf82010f`: the second call panics with `replace existing redaction store (probably unintentional)` — confirming the embedded-library failure mode described in #2285.
- Same program at PR head `0354aca2`: both calls succeed, and `go test ./internal/redact/` passes — including the new regression test asserting the swapped-out store's redactions stop applying afterwards. That test is a good choice: it documents the loss honestly instead of hiding it.

**Why the clio route looks strictly better than this PR**

syft isn't the only consumer with this exact shape. grype carries the byte-identical singleton (`internal/redact/redact.go`, same panic message) and the same per-Execute hoist (`cmd/grype/cli/cli.go` calls `redact.Set(state.RedactStore)` from its clio initializer). A clio-side ownership flip fixes both tools at once, whereas this PR resolves syft's copy and leaves grype panicking on the second embedded `Execute()`.

One ordering detail that makes the clio fix safe to do today, which may save @kzantow some analysis time: in clio's `loadConfigs`, the application itself sits at position 2 of `allConfigs` — `[]any{&a.state.Config, a, cfgs...}` — ahead of the caller's config objects. Since `fangs.Load` invokes `PostLoad()` in slice order, `a.PostLoad()` (which runs `state.setup` + the initializers, i.e. where a `WithRedactStore` handoff would live) fires **before** option-level `PostLoad`s like syft's `options/secret.go` `redact.Add(...)` hook. So flipping store ownership would not reorder secret registration relative to store installation: each run's own redactions still land in the store that is active when its output renders.

If the fix stays local to syft anyway, one hardening idea short of full DI: keep replacement (as this PR does) but make it observable — e.g. emit a warning when replacing a non-nil store, or make re-setting the *identical* store a no-op distinct from swapping in a new one. The original panic had real diagnostic value (catching accidental double-hoists); it's only the fatal form that breaks embedding, and downgrading it to a signal preserves that for future refactors.

From a verification standpoint the PR does exactly what it claims and its tests capture the tradeoff correctly — I'd just weight the clio change higher given it also covers grype.

PR #5145, opened 2026-08-02
fix: detect arm64 Linux kernel images in the kernel cataloger

### LINKED ISSUE #5126: linux-kernel-cataloger emits no package for arm64 kernel images
**What happened**:

On arm64 hosts, `linux-kernel-cataloger` produces no `linux-kernel` component. The SBOM is
well-formed but contains no kernel component from this cataloger. Kernel modules are
cataloged normally, only the kernel-image path is affected.

On arm64 & x86_64, the distro package is emitted:

`pkg:rpm/amzn/kernel6.18@6.18.38-76.139.amzn2023?arch=aarch64&distro=amzn-2023&epoch=1&package-id=08521cc7312f9fea&upstream=kernel6.18-6.18.38-76.139.amzn2023.src.rpm`

On x86_64 only, the kernel appears as a generic component; this is _missing_ on arm64:

`pkg:generic/linux-kernel@6.18.38-76.139.amzn2023.x86_64`

The `pkg:generic/linux-kernel` component is the one carrying an NVD-matchable CPE
(`cpe:2.3:o:linux:linux_kernel:...`), so on arm64 there is nothing in the BOM to match
kernel CVEs against by CPE.

This is because [deitch/magic only emits "Linux kernel"](https://github.com/deitch/magic/blob/c67ab88f10cb7323bd698f7fe7fb0c41f3af4da5/pkg/magic/internal/kernel.go#L12-L35) based off x86-specific bytes:

- 510 (0x1fe) — boot_flag, always 0xAA55
- 514 (0x202) — header, the magic HdrS (48 64 72 53)

Both fields are x86-specific. arm64 defines its image header separately, with magic ARM\x64 at 0x38, so magic does not recognise it as a Linux kernel. This means [this code](https://github.com/anchore/syft/blob/dd639c09b2852fd5716fc7a1b62d1cad08d74909/syft/pkg/cataloger/kernel/parse_linux_kernel_file.go#L56-L59) can never be hit:

```go
case strings.Contains(t, "ARM64 "):
    p.Architecture = "arm64"
case strings.Contains(t, "ARM "):
    p.Architecture = "arm"
```

Version extraction has the same constraint: the `version %s` message that the package
version is taken from is read at an indirect offset (526 → 0x200) nested beneath the same
`HdrS` test. Relaxing the magic check alone would therefore still leave `metadata.Version`
empty and return at the `metadata.Version == ""` check.

Tested on three distributions, plus an x86_64 control at the same kernel release.

| Distro | Kernel | Arch | `file` on vmlinuz | bytes 510 / 514 | `linux-kernel` |
|---|---|---|---|---|---|
| Amazon Linux 2023.12 | 6.18.38-76.139.amzn2023 | aarch64 | PE32+ EFI application (zboot) | zeros | **0** |
| Ubuntu 26.04 LTS | 7.0.0-1006-aws | aarch64 | gzip compressed data | gzip payload | **0** |
| RHEL 9.8 | 5.14.0-687.10.1.el9_8 | aarch64 | PE32+ EFI application (zboot) | zeros | **0** |
| Amazon Linux 2023.12 | 6.18.38-76.139.amzn2023 | **x86_64** | Linux kernel x86 boot executable bzImage | `0xAA55` / `HdrS` | **1** |

Rows 1 and 4 are the same kernel release built for different architectures.

A `pkg:generic/linux-kernel` package in the BOM, as on x86_64.

Amazon Linux 2023, aarch64:

```
$ syft scan file:/boot/vmlinuz-6.18.38-76.139.amzn2023.aarch64 \
    --select-catalogers +linux-kernel-cataloger -o json
{"artifacts":[],"artifactRelationships":[],"files":[{"id":"49b0511dba8447a8","location":{"path":"/vmlinuz-6.18.38-76.139.amzn2023.aarch64"},"executable":{"format":"pe","hasExports":false,"hasEntrypoint":false,"importedLibraries":[]},"unknowns":["unknowns-labeler: no package identified in executable file"]}]...}
```

Control — same kernel release, x86_64:

```
$ syft scan file:/boot/vmlinuz-6.18.38-76.139.amzn2023.x86_64 \
    --select-catalogers +linux-kernel-cataloger -o json
{"artifacts":[{"id":"403564c56e4d1872","name":"linux-kernel","version":"6.18.38-76.139.amzn2023.x86_64","type":"linux-kernel","foundBy":"linux-kernel-cataloger","locations":[{"path":"/vmlinuz-6.18.38-76.139.amzn2023.x86_64","accessPath":"/vmlinuz-6.18.38-76.139.amzn2023.x86_64","annotations":{"evidence":"primary"}}],"licenses":[],"language":"","cpes":[{"cpe":"cpe:2.3:o:linux:linux_kernel:6.18.38-76.139.amzn2023.x86_64:*:*:*:*:*:*:*","source":"nvd-cpe-dictionary"}],"purl":"pkg:generic/linux-kernel@6.18.38-76.139.amzn2023.x86_64","metadataType":"linux-kernel-archive"...}
```

**Extra info**:

All arm64 distributions I tested wrap or compress the image, so detection would need to unwrap it
before a version string is reachable:

- Amazon Linux / RHEL use EFI zboot
- Ubuntu uses a plain gzip stream.

Both unwrap to a raw arm64 `Image` carrying `ARM\x64` at 0x38 and a plaintext
`Linux version` banner matching `uname -r`, so the version is recoverable once
the container is handled.

**Environment**:
- syft 1.48.0, BuildDate 2026-07-16, on all four hosts

Tested on:
- Amazon Linux 2023.12.20260727 — 6.18.38-76.139.amzn2023, aarch64 and x86_64
- Ubuntu 26.04 LTS — 7.0.0-1006-aws, aarch64
- RHEL 9.8 (Plow) — 5.14.0-687.10.1.el9_8, aarch64

alexchen-sys:
## Problem

The linux-kernel cataloger recognizes x86 `bzImage` layouts, but arm64 kernel images produce no `linux-kernel` component at all — while kernel modules from the same system are cataloged normally. Reported cases: raw arm64 `Image`, whole-file gzip-compressed images (Ubuntu), and EFI zboot PE wrappers with embedded gzip/zstd/xz streams (Amazon Linux, RHEL).

Fixes #5126

## Fix

Keep the existing x86/deitch-magic path first, then fall back to an arm64 parser:

- detect the arm64 kernel header magic (`ARM\x64` at 0x38);
- search through embedded compressed streams (gzip/zstd/xz/bzip2, vendored decoders only) the way `extract-vmlinux` does, with bounded reads (64MB file / 256MB decompressed);
- extract the version from the `Linux version` banner, preserving the x86 empty-version behavior when no banner is present.

## Tests

New `parse_arm64_linux_kernel_file_test.go`: 5 positive matrix rows (raw, gzip, zboot+gzip, zboot+zstd, zboot+xz), 4 negative cases (no false positives on random non-kernel data, no-banner parity with the x86 empty-version rule, pre-existing magic-error behavior preserved), plus an x86 bzImage regression test. At head: the full `syft/pkg/cataloger/kernel` package suite passes; the only failure in the wider cataloger run is the pre-existing docker-dependent `Test_KernelCataloger`, unavailable on this host.

### DIFF
--- a/syft/pkg/cataloger/kernel/parse_arm64_linux_kernel_file.go
+++ b/syft/pkg/cataloger/kernel/parse_arm64_linux_kernel_file.go
@@ -1,0 +1,176 @@
+package kernel
+
+import (
+	"bytes"
+	"compress/bzip2"
+	"compress/gzip"
+	"io"
+	"strings"
+
+	"github.com/klauspost/compress/zstd"
+	"github.com/ulikunitz/xz"
+
+	"github.com/anchore/syft/syft/internal/unionreader"
+	"github.com/anchore/syft/syft/pkg"
+)
+
+const (
+	// arm64ImageMagicOffset is the offset of the arm64 Linux kernel image magic
+	// (0x644d5241, "ARM\x64") in the image header; see
+	// Documentation/arch/arm64/booting.rst in the kernel source tree.
+	arm64ImageMagicOffset = 0x38
+
+	// maxARM64KernelFileSize bounds how much of a candidate kernel file is read
+	// while looking for a wrapped arm64 image (distro vmlinuz files are far smaller).
+	maxARM64KernelFileSize = 64 << 20
+
+	// maxARM64KernelDecompressedSize bounds decompression of a candidate payload
+	// as a safeguard against decompression bombs.
+	maxARM64KernelDecompressedSize = 256 << 20
+)
+
+var (
+	arm64ImageMagic    = []byte{'A', 'R', 'M', 0x64}
+	linuxVersionMarker = []byte("Linux version ")
+)
+
+// kernelCompressionFormat pairs a compression stream magic with a decoder, used
+// when searching for an arm64 kernel image wrapped inside another binary (e.g.
+// EFI zboot), mirroring the approach of scripts/extract-vmlinux from the kernel
+// source tree. Only formats with decoders already vendored by syft are covered.
+type kernelCompressionFormat struct {
+	magic      []byte
+	decompress func([]byte) []byte
+}
+
+var kernelCompressionFormats = []kernelCompressionFormat{
+	{[]byte{0x1f, 0x8b, 0x08}, gunzipBytes},             // gzip
+	{[]byte{0x28, 0xb5, 0x2f, 0xfd}, unzstdBytes},       // zstd
+	{[]byte{0xfd, '7', 'z', 'X', 'Z', 0x00}, unxzBytes}, // xz
+	{[]byte{'B', 'Z', 'h'}, bunzip2Bytes},               // bzip2
+}
+
+// parseARM64LinuxKernelImage detects arm64 kernel images, which deitch/magic
+// does not recognize (its "Linux kernel" test keys off x86-specific bytes).
+// arm64 distro kernels are shipped as a raw Image, a whole-file compressed
+// Image (e.g. Ubuntu), or an EFI zboot PE binary wrapping a compressed Image
+// (Amazon Linux, RHEL).
+func parseARM64LinuxKernelImage(ur unionreader.UnionReader) (p pkg.LinuxKernel, ok bool) {
+	if _, err := ur.Seek(0, io.SeekStart); err != nil {
+		return p, false
+	}
+	data, err := io.ReadAll(io.LimitReader(ur, maxARM64KernelFileSize))
+	if err != nil {
+		return p, false
+	}
+	payload := findARM64ImagePayload(data)
+	if payload == nil {
+		return p, false
+	}
+	version, extendedVersion, found := parseLinuxVersionBanner(payload)
+	if !found {
+		return p, false
+	}
+	return pkg.LinuxKernel{
+		Architecture:    "arm64",
+		Version:         version,
+		ExtendedVersion: extendedVersion,
+		Format:          "Image",
+	}, true
+}
+
+// findARM64ImagePayload returns the arm64 kernel image contained in data: data
+// itself when it is a raw image, or the first embedded compression stream that
+// decompresses to an arm64 image.
+func findARM64ImagePayload(data []byte) []byte {
+	if isARM64Image(data) {
+		return data
+	}
+	for _, format := range kernelCompressionFormats {
+		if payload := findCompressedARM64Image(data, format); payload != nil {
+			return payload
+		}
+	}
+	return nil
+}
+
+func findCompressedARM64Image(data []byte, format kernelCompressionFormat) []byte {
+	offset := 0
+	for offset < len(data) {
+		idx := bytes.Index(data[offset:], format.magic)
+		if idx < 0 {
+			return nil
+		}
+		offset += idx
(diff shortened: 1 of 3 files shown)


CAOShurong (commented):
Independent exact-head verification of this fix — everything checks out, including an end-to-end probe against a real distro arm64 kernel that goes beyond the synthetic-fixture suite.

## Exact-head verification

Walked and ran the full change at head `822e444a` (parent `ef40eec3`, diff touches only the three kernel-cataloger files shown):

- **Unit/integration suite**: the entire new-test matrix passes — arm64 positives (raw Image, whole-file gzip, zboot+gzip/zstd/xz), negatives (random data, banner-less image parity, preserved magic-error behavior), x86 bzImage regression, and the `TestIsARM64Image` / `TestParseLinuxVersionBanner` unit helpers. The only failure in the package is the pre-existing Docker-dependent `Test_KernelCataloger`, which fails identically at base `ef40eec3` on a Docker-less host — not introduced by this PR.
- **End-to-end probe with a real arm64 kernel**: built syft at head and scanned the official Ubuntu ports artifact `linux-image-unsigned-7.0.0-1006-aws_7.0.0-1006.6_arm64.deb` (extracted `./boot/vmlinuz-7.0.0-1006-aws`, 21,109,056 bytes; gzip stream `1f 8b 08` at offset 0). That is exactly issue #5126's Ubuntu row (whole-file gzip). Result at head:

```
linux-kernel @ 7.0.0-1006-aws
purl: pkg:generic/linux-kernel@7.0.0-1006-aws
cpe:  cpe:2.3:o:linux:linux_kernel:7.0.0-1006-aws:*:*:*:*:*:*:*
metadata: architecture=arm64  format=Image  version=7.0.0-1006-aws
```

The same file scanned by a binary built at base `ef40eec3` yields `artifacts: []` — reproducing #5126 exactly. So the NVD-matchable component is restored for real-world wrapped images, not just fixtures.
- **False-positive probe**: 2 MB of random bytes fed through the new fallback path → 0 artifacts (no spurious compression-stream matches).

## Code-level notes (non-blocking)

1. Moving the `magic.GetType` error from fail-fast to deferred (returned only after both detectors miss) preserves the old contract for unreadable files, and the negative tests pin it. Clean restructure.
2. Bounded reads (64 MiB raw / 256 MiB decompressed) plus first-match-wins scanning keep worst-case cost sane; tolerating trailing bytes after the stream in `drainBounded` matches PE-overlay reality.
3. One observation for maintainers weighing scope: after this change, every glob-matched file that fails the magic check gets up to 64 MiB read and scanned for embedded streams. Exposure is limited to kernel-glob paths (`**/vmlinuz*`, `**/kernel*`, ...), but if that ever matters, a cheap pre-filter (skip when byte 0 is none of `MZ` / a known compression magic / a plausible Image start) would tighten it. Not required for this PR.

Functionally this looks ready: arm64 hosts regain the `pkg:generic/linux-kernel` component carrying the `linux:linux_kernel` CPE, which was the core of #5126. Leaving the merge/scope decision to maintainers.

PR #5154, opened 2026-08-06
feat(golang): add extended-stdlib scope and module patterns for symbol capture

wagoodman:
`golang.capture-symbols` decides how much symbol data lands in the SBOM for grype's reachability analysis. It's `none`, `stdlib`, or `all` today, and the useful middle is missing: `stdlib` stops at the standard library, `all` multiplies SBOM size.

A new `extended-stdlib` configurable covers stdlib plus everything under `golang.org/x/`:

```yaml
golang:
  capture-symbols: extended-stdlib
```

Also, a new `capture-symbols-modules` configurable for modules that are noisy in your binaries but not everyone's. It's unioned with whatever the scope selects, so it only ever widens:

```yaml
golang:
  capture-symbols: extended-stdlib
  capture-symbols-modules:
    - github.com/klauspost/**
```

Patterns are standard doublestar globs, matched against the module path with any trailing major version suffix taken off, so a config does not quietly stop covering a module the day it bumps a major version:

```yaml
golang:
  capture-symbols-modules:
    - github.com/klauspost/*            # every klauspost module, any major version
    - github.com/klauspost/**           # the same, plus anything nested under them
    - k8s.io/client-go                  # that module, any major version
    - github.com/klauspost/compress/v2  # v2 alone
```

A suffix is part of a module's path but not part of its identity, hence the asymmetry: a pattern that does not mention one covers every major version, and a pattern that does covers only that one. Only a *trailing* suffix counts, which is Go's own rule, so in `github.com/anchore/syft/v2/thing` the `v2` stays an ordinary path element and `github.com/anchore/**/thing` is what reaches it. `module.SplitPathVersion` from `golang.org/x/mod` does the splitting, so gopkg.in's `.vN` form works and `/v0` and `/v1` are correctly not suffixes.

Ordering is `none` < `stdlib` < `extended-stdlib` < `all`. The existing three values and the `none` default are unchanged, and the module list is inert under `none`. Presets compile into glob lists internally, so a single matcher answers "does this module get symbols" instead of a preset branch sitting next to a separate glob branch.

An unrecognized `capture-symbols` value still falls back to `none`, but warns now instead of doing it silently. A malformed module pattern warns and gets skipped.

One thing worth a look beyond the feature: the `Symbols` field description in the JSON schema was wrong after this (it claimed only `all` and `stdlib` populate anything), and that description lives in the already-published `16.1.10`. Rather than bump a version for a sentence, `16.1.10` is amended in place and `schema/json/README.md` grows an explicit exception for description-only changes: descriptions only, no shape change of any kind, `$id` unchanged. Anything else still needs a bump. Happy to split that into its own PR if you'd rather review the policy separately.


cmd/syft/internal/options/golang.go
```diff
@@ -42,16 +44,39 @@ if unset this defaults to $GONOPROXY`)
...
 a more accurate version from the binary.`)
 	descriptions.Add(&o.UsePackagesLib, `use the golang.org/x/tools/go/packages library, which executes golang tooling found on the path in addition to potential network access to get the most accurate results`)
+	// note: descriptions must be static string literals; the app config discovery that generates the
+	// capability docs reads them straight out of the AST
 	descriptions.Add(&o.CaptureSymbols, `capture function symbols from the binary symbol table (pclntab). valid values are:
-"none" (disabled), "stdlib" (only the synthetic stdlib package), and "all" (all module packages plus stdlib)`)
+"none" (disabled), "stdlib" (only the synthetic stdlib package), "extended-stdlib" (stdlib plus every
+module under golang.org/x/), and "all" (all module packages plus stdlib)`)
+	descriptions.Add(&o.CaptureSymbolsInclude, `glob patterns matched against go module paths (e.g. github.com/klauspost/**) that should have symbols
+captured in addition to whatever capture-symbols selects. ** crosses path separators, * does not.
```
kzantow, line 53: I realize this is documented well across the PR and here, but I think people will still be confused why `github.com/org/*` doesn't include `github.com/org/thing/v2` and maybe can't understand the use case: if I want an exact non-v2 match, I could just explicitly add it as `github.com/org/thing`. The more I look at this, I think the `**` vs `*` will become even more confusing when we include the same symbol sometimes e.g. `github.com/anchore/syft/syft/pkg` when there is `github.com/anchore/*` at syft v1 but not when there is `github.com/anchore/syft/v2`. Maybe this would be mitigated somewhat by 👇 

Also while reviewing this, I wonder if `modules` may be better than `include`, e.g. `capture-symbols-modules` (or `packages`, if it was filtering packages, but I think it's filtering modules) since the user is supposed to specify module expressions rather than some other part of the symbol, right?

kzantow (approved):
I think this is generally good, but left a note about the naming and `**` vs `*`

cmd/syft/internal/options/golang.go
wagoodman: I more kept the globing behavior we had elsewhere for consistency. At the very least I'm going to make the field rename change, and I'll think more on the globing behavior.

wagoodman: ok just pushed the rename + supporting the /v# suffix exception mentioned

kzantow (approved):
LGTM 👍

wagoodman:
I'll wait for the unicorns to disperse...

DanielApelaten23:
Nice work.

PR #5160, opened 2026-08-07
fix(java): resolve ${project.groupId} inherited from a parent POM

sueun-dev:
`resolveProjectProperty` has a special case so `${project.version}` and `${project.groupId}` fall back to the parent POM when a child module doesn't declare its own value. The `groupId` arm uses `case "groupID":`, but the property segment (and the gopom xml tag on `Project.GroupID`) is `groupId`, so the arm never runs. `${project.version}` works because its label matches.

The effect shows up when a child module omits `<groupId>` and inherits it from the parent, and then references `${project.groupId}` somewhere — for example a dependency declared as `<groupId>${project.groupId}</groupId>` to point at a sibling module. With the arm dead, the reflection fallback lands on the nil `GroupID` pointer and the expression resolves to an empty string instead of the parent's groupId (the error is swallowed in `resolveExpression`). That empty value flows through `ResolveDependencyID`, so the dependency's Maven coordinates and PURL namespace come out wrong.

The change is the one-character case label. I added a `Test_resolveProperty` case for the inherited `groupId` (fails before, passes after) and a matching one for the already-working inherited `version` so both stay covered.

### DIFF
--- a/syft/pkg/cataloger/java/internal/maven/resolver.go
+++ b/syft/pkg/cataloger/java/internal/maven/resolver.go
@@ -186,7 +186,7 @@ func (r *Resolver) resolveProjectProperty(ctx context.Context, resolutionContext
 					if pom.Version == nil && pom.Parent.Version != nil {
 						return r.resolveExpression(ctx, resolutionContext, *pom.Parent.Version, resolving)
 					}
-				case "groupID":
+				case "groupId":
 					if pom.GroupID == nil && pom.Parent.GroupID != nil {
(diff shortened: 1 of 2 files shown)


CAOShurong (approved):
Clean typo fix with good test coverage. The Maven POM spec uses `groupId` (camelCase), not `groupID`, so the switch case was never matching for inherited group IDs. The added tests properly cover both the groupId inheritance path and confirm the existing version inheritance behavior.

PR #5169, opened 2026-08-11
fix(php): attach PHP License to interpreter binaries

### LINKED ISSUE #5026: missing php license
**What happened**:
```
syft php:8.5-fpm-alpine -o spdx-json \
  | jq -r '
      .packages[]
      | select(.name == "php-cli" or .name == "php-fpm" or .name == "php")
      | [
          .name,
          (.versionInfo // "unknown"),
          (.licenseDeclared // "NOASSERTION"),
          (.licenseConcluded // "NOASSERTION")
        ]
      | @tsv
    '
```
Doesn't report a license for php-cli/php-fpm and php.

```
php-cli	8.5.7	NOASSERTION	NOASSERTION
php-fpm	8.5.7	NOASSERTION	NOASSERTION
php	8.5-fpm-alpine	NOASSERTION	NOASSERTION
```

I'd expect syft to be able to extract the license for the very popular docker php image.

Shared above.

**Environment**:
- Output of `syft version`: syft 1.46.0
- OS (e.g: `cat /etc/os-release` or similar): alpine image, host macos

addielaruee:
## Description

The PHP interpreter cataloger detects `php-cli`, `php-fpm`, and the `libphp` Apache module as binary findings. These binaries carry no license text of their own, so Syft reported them with no license (`NOASSERTION`), as noted in #5026:

```
php-cli	8.5.7	NOASSERTION	NOASSERTION
php-fpm	8.5.7	NOASSERTION	NOASSERTION
```

Every PHP interpreter has been distributed under the **PHP License** (SPDX: `PHP-3.01`) across all supported versions. Once the cataloger has *positively identified* a binary as the PHP interpreter, that license can be reliably associated with it — this side-steps the association problem raised in the issue discussion (the license file that ships in `/usr/src/php.tar.xz` lives elsewhere in the image and can't be reliably tied back to the binary).

## Changes

- Attach `PHP-3.01` as a **concluded** license to each interpreter package produced by the interpreter cataloger (`php-cli`, `php-fpm`, `libphp`). Concluded (rather than declared) reflects that the license is inferred from positive identification of the binary, not read from a declared field — matching the convention used by the Bitnami cataloger.
- Thread the existing `context.Context` from `Catalog` into `catalogInterpreters` (it was previously discarded) so the license set can be built with context.
- Add a Docker-free regression test (`Test_InterpreterCataloger_attachesPHPLicense`) using a small directory fixture, asserting the interpreter package carries the `PHP-3.01` concluded license. Verified it fails without the change and passes with it.

Extensions (`*.so`) are intentionally left untouched — bundled core extensions are also PHP-licensed, but third-party PECL extensions may differ, so this change is scoped to the interpreter itself.

## Test results

```
go test ./syft/pkg/cataloger/php/ -run attachesPHPLicense   # PASS
gofmt / go vet                                              # clean
```

The existing `Test_InterpreterCataloger` (image-fixture based) asserts package/relationship *strings*, which are unaffected by the added license.

Fixes #5026


syft/pkg/cataloger/php/interpreter_cataloger.go
```diff
@@ -12,10 +12,19 @@ import (
...
 	"github.com/anchore/syft/syft/pkg/cataloger/internal/binutils"
 )
 
+// phpInterpreterLicense is the SPDX identifier for the PHP License, version 3.01.
+// The PHP interpreter (php-cli, php-fpm, and the libphp Apache module) has been
+// distributed under this license across all supported PHP versions, so it can be
+// reliably associated with every interpreter binary this cataloger positively
+// identifies -- even though the license text does not sit alongside the binary
+// itself, which is why it was otherwise reported as NOASSERTION (see #5026).
+const phpInterpreterLicense = "PHP-3.01"
```
wagoodman, line 26: This is not exactly true, the license says:

> Once covered code has been published under a particular version of the license, you may always continue to use it under the terms of that version. You may also choose to use such covered code under the terms of any subsequent version of the license published by the PHP Group. 

The load bearing phrase is "you may choose" any subsequent version of the license. I don't think we can claim that we found a specific `PHP-#` license though. I understand that logically it tracks that the license enables this, but they might change that in some future text of the license (but we would still claim 3.01 is what that future interpreter is covered under).

So I don't think syft is the right authority to always claim we found specifically 3.01.

PR #5170, opened 2026-08-11
fix(file-resolver): don't abort scan on inaccessible symlink target

### LINKED ISSUE #3286: Survive indexing not accessible files
**What happened**:
syft crash when a symlink is referencing a file which is not accessible (under not readable dir)

syft should continue ignoring the file as for other non readable ones

- /tmp/dir1 is root owned rwxrwx---
- /tmp/link -> /tmp/dir1/dir2/file
- syft /tmp run as non root

Use the attached docker file:
```shell
$ mv Dockerfile.txt Dockerfile
$ docker build -t syftissue .
$ docker run --rm -it syftissue
 ✔ Indexed file system                                                                                                               /tmp
[0000]  WARN unable to access path="/tmp/dir1": open /tmp/dir1: permission denied
unable to get file resolver: unable to create directory resolver: unable to index filesystem path="/tmp/dir1/dir2/file": lstat /tmp/dir1/dir2: permission denied
returned code: 1
```

 looks like #2645 (but already closed) and #3258 (not exactely the same since the directory is not excluded)

**Environment**:
- Output of `syft version`: syft 1.13.0
- OS : first encountered in Red Hat Enterprise Linux 9.4 (Plow)  then reproduced in Alpine

[Dockerfile.txt](https://github.com/user-attachments/files/17164114/Dockerfile.txt)

addielaruee:
## Description

Fixes #3286.

When a symlink resolves into a directory the scanner has no permission to traverse, Syft aborts the **entire** scan with a fatal error instead of skipping the inaccessible path:

```
unable to get file resolver: unable to create directory resolver: unable to index filesystem
path="/tmp/dir1/dir2/file": lstat /tmp/dir1/dir2: permission denied
returned code: 1
```

This is inconsistent with how the indexer already treats every other inaccessible path — those are logged (`unable to access path=...`), recorded in `errPaths`, and skipped so indexing continues.

## Root cause

`indexTree` calls `isRealPath(root)`, which runs `filepath.EvalSymlinks` on the path's parent. When the symlink target lives under a directory we can't traverse, that returns a permission `*os.PathError`, and `indexTree` returned it as fatal:

```go
shouldIndexFullTree, err := isRealPath(root)
if err != nil {
    return nil, err   // <- aborts the whole scan
}
```

Notably, `indexBranch` already handles the *same* `EvalSymlinks` error gracefully ("we can't index the path, but we shouldn't consider this to be fatal"). This PR brings `indexTree` in line with that existing behavior.

## Fix

Route the `isRealPath` path error through the indexer's existing `isFileAccessErr` handling, so the offending path is warned + tracked in `errPaths` and skipped, rather than aborting the scan. Non-path errors are still returned as before.

## Testing

Added `TestDirectoryIndexer_index_survive_inaccessibleSymlinkTarget`, which builds the reporter's scenario at runtime (a symlink into a `chmod 0000` directory) and asserts indexing completes without error while recording the inaccessible path. It reproduces the original failure exactly — without the fix it fails with:

```
unable to index filesystem path=".../restricted/nested/target.txt": lstat .../restricted/nested: permission denied
```

The test skips when run as root (root bypasses the permission bits that trigger the bug) and resolves the temp-dir path first so it exercises the real-path (`filepath.Walk`) branch on macOS as well as Linux.

```
go test ./syft/internal/fileresolver/ -run TestDirectoryIndexer   # all pass
gofmt / go vet                                                    # clean
```

### DIFF
--- a/syft/internal/fileresolver/directory_indexer.go
+++ b/syft/internal/fileresolver/directory_indexer.go
@@ -126,6 +126,14 @@ func (r *directoryIndexer) indexTree(root string, stager *progress.AtomicStage) 
 
 	shouldIndexFullTree, err := isRealPath(root)
 	if err != nil {
+		// a symlink can resolve into a directory we don't have permission to traverse (or into a path
+		// that no longer exists). Like any other inaccessible path, this should be recorded and skipped
+		// with a warning rather than aborting the entire scan (see #3286).
+		var pathErr *os.PathError
+		if errors.As(err, &pathErr) {
+			r.isFileAccessErr(root, err)
+			return nil, nil
+		}
 		return nil, err
 	}
 
--- a/syft/internal/fileresolver/directory_indexer_test.go
+++ b/syft/internal/fileresolver/directory_indexer_test.go
@@ -286,6 +286,42 @@ func TestDirectoryIndexer_index_survive_badSymlink(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestDirectoryIndexer_index_survive_inaccessibleSymlinkTarget(t *testing.T) {
+	// a symlink that resolves into a directory we don't have permission to traverse should be skipped
+	// with a warning like any other inaccessible path, not abort the entire scan (see #3286).
+	if os.Geteuid() == 0 {
+		t.Skip("cannot run as root: root bypasses the directory permissions that trigger the bug")
+	}
+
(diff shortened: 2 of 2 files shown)


kzantow (approved):
LGTM -- we talked about this and probably will want to be able to surface these errors, but this is a step in the right direction 👍

Issue #5062, opened 2026-07-12
Support nginx-ingress-controller binary various versions


witchcraze:
**What happened**:

Current matcher can not detect some pattern
- arm64, arm/v7, s390x
  - v0.34.0 - v0.51.0
  - v1.0.0 - v1.9.6

**Steps to reproduce the issue**:

```
$ syft -q --platform linux/arm64 registry.k8s.io/ingress-nginx/controller:v1.9.6 | grep "ingress"
k8s.io/ingress-nginx                                 UNKNOWN                               go-module  (+2 duplicates)

$ syft -q --platform linux/arm/v7 registry.k8s.io/ingress-nginx/controller:v1.9.6 | grep "ingress"
k8s.io/ingress-nginx                                 UNKNOWN                               go-module  (+2 duplicates)

$ syft -q --platform linux/s390x registry.k8s.io/ingress-nginx/controller:v1.9.6 | grep "ingress"
k8s.io/ingress-nginx                                 UNKNOWN                               go-module  (+2 duplicates)
```

PR #5179, opened 2026-08-13
fix(binary): detect multi-arch ingress-nginx

CAOShurong:
## Summary

Detect `nginx-ingress-controller` v1.9.6 binaries from the published ARM64, ARMv7, and s390x controller image variants. These architectures place the controller version farther from the Go build version than the existing classifiers allow.

This addresses the currently reproduced architectures in #5062 without claiming coverage for every historical release in that issue.

## Validation

- `go test ./syft/pkg/cataloger/binary -run Test_Cataloger_PositiveCases/nginx-ingress-controller/1.9.6 -count=1 -v`
- the same focused test with `-must-use-original-binaries` against all four published v1.9.6 binaries (amd64, arm64, arm/v7, s390x)
- all existing ingress-nginx positive fixtures
- `TestClassifierCataloger_DefaultClassifiers_NegativeCases`
- built the `syft` CLI and scanned each original binary; every architecture returned exactly one `nginx-ingress-controller` package at version `1.9.6`

The fixtures were generated from `registry.k8s.io/ingress-nginx/controller:v1.9.6@sha256:1405cc613bd95b2c6edd8b2a152510ae91c7e62aea4698500d23b2145960ab9c`.

## Disclosure

This change was developed with OpenAI Codex assistance. I independently reproduced the missing detections, inspected the binary evidence, and ran the tests and real CLI scans listed above.


wagoodman (approved):
I made a couple of alterations:
- updated the fixtures to be captured by the helper tool
- updated the helper tool to cover additional architecture cases (and updated existing non-compliant fixtures themselves)

PR #5187, opened 2026-08-14
Prevent unnecessary allocations when parsing compressed ELF sections

wagoodman:
`debug/elf` sizes a compressed section's buffer from that section's own compression header, and a compressible stream delivers every byte the header promises, so `internal/saferio`'s chunked read still allocates based on the user requested size. So a 510KB file declaring a 512MB section drives `elf.NewFile` through 2.6GB of allocations and would return no error... in some cases it's arguable this could lead to a fatal Go OOM (thus not a recoverable panic).

The fix it to start being selective about which sections to read does not avoid it: `elf.NewFile` expands the section-name string table on its own, and `File.Symbols` expands `.symtab` plus whatever section its `Link` field points at. Then we can use `elf.NewFile` as usual

Here are the notable changes:

- A new `syft/internal/elfutil.NewFile`, a drop-in for `elf.NewFile` that rejects a declared decompressed size over 128MB, which every call site goes through it, and a ruleguard rule keeps the next one from going direct.

- The check runs in two parts, since `debug/elf` expands sections at two different times: the section-name string table against the raw bytes before the parse, everything else after it, where names, types and declared sizes are already resolved

- Only the sections syft can actually reach are bounded, which keeps the guard off real binaries. DWARF is excluded since nothing calls `File.DWARF`, so a large compressed `.debug_info` no longer skips the whole file, and sections `debug/elf` will not decompress anyway (`SHF_ALLOC`, `SHT_NOBITS`) are left alone

- A size rejection is a plain error rather than an `*elf.FormatError`, so callers that distinguish "not an ELF" from "bad ELF" treat it as the latter. A rejected file is skipped and logged rather than cataloged, which is the right direction to fail given the alternative is OOM-killing the whole scan

In terms of performance differences, there is no measurable difference in wall clock time or RSS usage between the main branch and this PR.

## Isn't this working around stdlib behavior?...

Yes, and upstream has already declined to fix it. `debug/elf` carries this in its package docs:

> This package is not designed to be hardened against adversarial inputs, and is outside the scope of https://go.dev/security/policy. In particular, only basic validation is done when parsing object files, and care should be taken when parsing untrusted inputs, as parsing malformed files may consume significant resources, or cause panics.

That disclaimer makes every report in this class a non-vulnerability by definition, so there wont be a CVE / advisory coming. `debug/macho` and `debug/pe` carry the same text.

The history also supports this:

- [#45599](https://github.com/golang/go/issues/45599) (2021), `fatal error: runtime: out of memory` on a real kernel module with a bogus section size. Left `NeedsInvestigation`, then `FrozenDueToAge`. Reporter noted `readelf` handles the same file with a warning
- [#53626](https://github.com/golang/go/issues/53626) (2022), crafted `sh_size` of `0xFFFFFFFFFFFFFFFF`, `makeslice: len out of range`. Closed as not planned
- [#68454](https://github.com/golang/go/issues/68454) (2024), the closest match, and names `saferio` directly: it "just prevents it from allocating that many _at a time_. It will happily allocate chunks in a loop." Still open, `NeedsDecision`, milestone Unplanned
- [#47445](https://github.com/golang/go/issues/47445), panic on malformed ELF, same general family
- [golang/go#49437](https://groups.google.com/g/golang-codereviews/c/JOVcpqrWpdI), a CL to read section contents incrementally in `Section.Data()`, abandoned after review pushback

Worth noting none of those are the compressed path, but they all turn on a tampered `sh_size`, which is what [`saferio`](https://pkg.go.dev/internal/saferio) was added to generally deal with, since a short file cannot deliver the bytes it claims. Anywho, the upstream position is that parsing untrusted binaries is the caller's problem. That is why this PR exists here in Syft.

## Type of change

## Checklist

### DIFF
--- a/syft/internal/elfutil/elfutil.go
+++ b/syft/internal/elfutil/elfutil.go
@@ -4,21 +4,31 @@ implausible decompressed size.
 
 debug/elf sizes a section's buffer from the section's own compression header, and a highly compressible
 stream really does deliver the bytes that header promises, so internal/saferio's chunked read does not
-help: it faithfully allocates every byte, and appends its way there, so the cost runs several times the
-declared size. A 510KB input file declaring a 512MB section drives elf.NewFile through 2.6GB of
+help: it appends its way to every declared byte, and Go's 1.25x slice growth puts the lifetime cost near
+five times that. 512MB of zeros is 510KB of zlib, and that file drives elf.NewFile through 2.6GB of
 allocation and returns no error, which is a fatal Go OOM rather than a recoverable panic. Zlib caps out
-near 1000:1, but debug/elf also accepts zstd, which reaches 32767:1 on zeroed input, so the input needed
+near 1000:1, but debug/elf also accepts zstd, which passes 32000:1 on zeroed input, so the input needed
 to declare a given size is smaller still.
 
-The check runs in two parts, because debug/elf expands sections at two different times. elf.NewFile
-expands exactly one section on its own, the section-name string table, so that one has to be bounded
-against the raw bytes before the call is made. Every other section is expanded lazily by
-(*Section).Open, so those are bounded after the parse, where names, types and decompressed sizes are all
-resolved already and no hand-rolled walk of the section table is needed.
+The check runs in two parts, because debug/elf expands sections at two different times:
+
+  - the section-name string table is the one section elf.NewFile expands on its own, so it has to be
+    bounded against the raw bytes before the call is made. That is CheckSectionNameTable, and the walk
+    of the section table it needs is why this package decodes any ELF structures by hand at all.
+  - every other section is expanded lazily by (*Section).Open, so those are bounded after the parse,
+    where names, types and decompressed sizes are resolved already.
 
 Only the sections syft can actually drive debug/elf into decompressing are bounded. DWARF is deliberately
 excluded: syft never asks for it, so a binary carrying a large compressed .debug_info is cataloged rather
 than skipped over a section nobody reads.
+
+Sources:
+  - https://www.sco.com/developers/gabi/latest/ch4.sheader.html
+  - https://groups.google.com/g/generic-abi/c/satyPkuMisk
+  - https://docs.oracle.com/en/operating-systems/solaris/oracle-solaris/11.4/linkers-libraries/gnu-style-section-compression.html
+  - https://sourceware.org/pipermail/binutils/2015-July/089559.html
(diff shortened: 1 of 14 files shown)


test/rules/rules.go
```diff
@@ -90,3 +90,16 @@ func packagesInRelationshipsAsValues(m dsl.Matcher) {
 		Where(isRelationship(m) && hasPointerType(m)).
 		Report("pointer used as a value for From/To field in artifact.Relationship (use values instead)")
 }
+
+// nolint:unused
+func noDirectELFOpen(m dsl.Matcher) {
```
spiffcs, line 95: Nice call adding a rule here

syft/internal/elfutil/elfutil_test.go
```diff
@@ -0,0 +1,548 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		for _, c := range classes {
+			t.Run(tt.name+"/"+c.name, func(t *testing.T) {
+				data := buildELF(t, c.class, c.order, tt.secs, tt.opts)
+				_, err := NewFile(bytes.NewReader(data))
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.wantErr)
```
spiffcs, line 344: Is this assertion a little too brittle or are we just trying to protect that the section name information is never removed from the error with the contains? Do we want to protect against the full error message changing at all or are the limit message and section name good enough?

syft/internal/elfutil/elfutil.go
```diff
@@ -0,0 +1,338 @@
...
+	".data":         {},
+	".text":         {},
+	".gopclntab":    {},
+	".note.package": {},
+	".modinfo":      {},
+}
+
+// wire sizes of the structures we decode, which differ between the two ELF classes
+var (
+	sizeSection32 = int64(binary.Size(elf.Section32{}))
```
spiffcs, line 79: I was gonna look these up then I saw we're pulling from debug/elf. 👍 

Go Docs link for anyone curious: https://pkg.go.dev/debug/elf#Chdr64

spiffcs (commented):
No comments on the implementation. I think we're going to have a quick offline sync about the negative cases for  `TestNewFile_AcceptsWhatDebugELFWouldNotExpand` and then will 🟢

spiffcs (approved):

syft/internal/elfutil/elfutil_test.go
wagoodman: yeah, let me at least make it something more obviously unique , like the max declared section size constant

PR #5188, opened 2026-08-16
fix(javascript): read every document of a pnpm-lock.yaml stream

### LINKED ISSUE #5168: pnpm cataloger reads only the first YAML document: SBOM contains pnpm's own binaries and no project dependencies
### What happened

`pnpm-lock.yaml` can be a **two-document YAML stream**. The pnpm cataloger reads the first document only. On such a lockfile the resulting SBOM contains pnpm's own release binaries and **none of the project's dependencies**.

The SBOM is well-formed and non-empty, so nothing signals that it is wrong. Observed with syft 1.51.0.

### Background

pnpm stores a small "env" document at the head of `pnpm-lock.yaml` — config dependencies and the pinned package-manager version — and the project's dependency graph in the second document. The layout has shipped since pnpm 11 for projects using config dependencies. In pnpm 12 the package-manager pin is recorded by default, so two documents is now the common case rather than a rare one.

Both documents declare `lockfileVersion: '9.0'`, so the version field gives no signal that the file has two documents.

### Reproduction

Save this as `pnpm-lock.yaml` (produced by pnpm 12.0.0-rc.3 for a project whose only dependency is `minimist@1.2.0`, trimmed to one platform binary for brevity):

```yaml
---
lockfileVersion: '9.0'

importers:

  .:
    configDependencies: {}
    packageManagerDependencies:
      pnpm:
        specifier: 12.0.0-rc.3
        version: 12.0.0-rc.3

packages:

  '@pnpm/exe.linux-x64@12.0.0-rc.3':
    resolution: {integrity: sha512-/6xWaYfp6MEaJF+7AZIfCMp/fZX2jZlTLh2KVBEH0QOWk1ktaBlKNIJEgT4m6mieOLhNYMzCfLJImMbJDK1IwA==}
    cpu: [x64]
    os: [linux]
    libc: [glibc]

  pnpm@12.0.0-rc.3:
    resolution: {integrity: sha512-JZ9fDGH+WLdRdTEikN3UxeZe6bpDY7dYwV0RX0+OrJ927vc72XbId4IJXeT++ebNFA9cF3IQ1swiXHEd/Maq0Q==}
    engines: {node: '>=18.*'}
    hasBin: true

snapshots:

  '@pnpm/exe.linux-x64@12.0.0-rc.3':
    optional: true

  pnpm@12.0.0-rc.3:
    optionalDependencies:
      '@pnpm/exe.darwin-arm64': 12.0.0-rc.3
      '@pnpm/exe.darwin-x64': 12.0.0-rc.3
      '@pnpm/exe.linux-arm64': 12.0.0-rc.3
      '@pnpm/exe.linux-arm64-musl': 12.0.0-rc.3
      '@pnpm/exe.linux-x64': 12.0.0-rc.3
      '@pnpm/exe.linux-x64-musl': 12.0.0-rc.3
      '@pnpm/exe.win32-arm64': 12.0.0-rc.3
      '@pnpm/exe.win32-x64': 12.0.0-rc.3

---
lockfileVersion: '9.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

importers:

  .:
    dependencies:
      minimist:
        specifier: 1.2.0
        version: 1.2.0

packages:

  minimist@1.2.0:
    resolution: {integrity: sha512-7Wl+Jz+IGWuSdgsQEJ4JunV0si/iMhg42MnQQG6h1R6TNeVenp4U9x5CC5v/gYqz/fENLQITAWXidNtVL0NNbw==}

snapshots:

  minimist@1.2.0: {}
```

```
$ syft scan file:pnpm-lock.yaml -o json | jq -r '.artifacts[].name'
@pnpm/exe.linux-x64
pnpm
```

Delete everything up to and including the second `---` line, leaving only the project document, and the same command yields the expected result:

```
$ syft scan file:pnpm-lock.yaml -o json | jq -r '.artifacts[].name'
minimist
```

On a larger fixture (`lodash@4.17.15`, `minimist@1.2.0`, `axios@0.21.0`), the single-document form catalogs 4 components and the two-document form catalogs 9 — `pnpm` plus its eight `@pnpm/exe.*` binaries, and no project dependency at all.

### What I expected

The project's dependencies are cataloged regardless of how many documents the file contains.

### Suggested fix

**Take the last document** in the stream. That is correct for single-document lockfiles, for every two-document lockfile pnpm has written, and it needs no layout detection. In Go, decode in a loop with `yaml.NewDecoder` and keep the final document.

If you prefer explicit detection: a `pnpm-lock.yaml` beginning with `---` at byte 0 has two documents and the project lockfile is the second; pnpm writes that leading marker only when an env document is present.

The env document's config dependencies are themselves real npm packages, installed into `node_modules/.pnpm-config`, so cataloging them in addition would be reasonable. The critical part is that they must not *replace* the project's dependencies.

### Context

Downstream discussion, including the effect on GitHub's dependency graph: https://github.com/pnpm/pnpm/issues/13805

---
Written by an agent (Claude Code, claude-opus-5).

hamodywe:
Fixes #5168.

## The problem

`pnpm-lock.yaml` can be a multi-document YAML stream. pnpm keeps config dependencies and the pinned package-manager version in a leading document, and the project's dependency graph in the next one. As @zkochan notes on the issue, that layout has shipped since pnpm 11 for projects using config dependencies, and pnpm 12 records the package-manager pin by default — so two documents is now the common case rather than a rare one.

`parsePnpmLock` used `yaml.Unmarshal`, which decodes **only the first document**. On such a lockfile the resulting SBOM contains pnpm's own release binaries and **none of the project's dependencies**.

The failure is silent in the way that matters most for an SBOM: both documents declare the same `lockfileVersion`, the output is well-formed and non-empty, and nothing indicates that every real dependency is missing. Anything consuming that SBOM for vulnerability matching sees a clean result for a project whose dependencies were never scanned.

## The change

Split the stream and parse each document, merging the results.

- A single-document lockfile yields exactly one document, so existing behaviour is unchanged — the previous code path is the one-document case of the new one.
- Version detection stays per-document. A document that omits `lockfileVersion` falls back to the first document's version instead of being dropped, and the original error is still returned when the *first* document has no usable version.
- Duplicate `name@version` entries across documents are merged rather than double-reported, keeping whichever entry carries the integrity hash and dependencies. A package is only treated as a dev dependency if every document agrees.

## Tests

`testdata/pnpm-multi-doc/pnpm-lock.yaml` is the two-document lockfile from the issue (trimmed to one platform binary), and `TestParsePnpmLock_MultiDocument` asserts that `minimist` — the project's only real dependency, which lives in the second document — is catalogued alongside the pnpm binaries from the first.

Verified the test reproduces the bug: with the parser reverted to reading only the first document it fails with `minimist` missing from the results. All existing `Pnpm` tests pass unchanged.

Note for reviewers: `Test_JavascriptCataloger`, `Test_PackageCataloger_Globs`, `Test_LockCataloger_Globs` and `TestParseYarnLock_DevDependencies` fail on my machine, but they fail identically on a pristine checkout without this change (Windows line-ending/glob artifacts), so they are unrelated.


zkochan (approved):

spiffcs (approved):

wagoodman (approved):
Thanks for the contribution! I made a couple of adjustments to the implementation:
- adjusts the parsing loop to stream instead of raise up full buffers of contents (improves memory performance)
- better captures "unknowns" in the SBOM when there are some documents that will not parse (but others still do parse)

spiffcs:
👋 Thanks for the contribution @hamodywe and the review @zkochan.

I had one question about this that could turn into a small followup:
We added this case under `Test_pnpmDocumentVersion`:
`{name: "inherits when omitted", doc: "packages: {}\n", established: 9.0, want: 9.0}`

Is this correct? When there are two streamed documents and any n+1 doesn't have a version should it just subsume the version found in the first correct parse or are there cases where different versions can be part of the same stream?

PR #5195, opened 2026-08-18
fix(golang): bound allocations from UPX-packed binary headers

wagoodman:
A UPX-packed binary declares its own sizes, and those sizes bore no relation to what the file actually contains. They fed allocations directly, so a small crafted file could drive multiple gigabytes of allocation. Go can raise these as a fatal runtime OOM, not a recoverable panic, so the existing `recover` on this path cannot contain it.

The fix bounds values that came out of the file being parsed:

- the declared original size is capped both absolutely and against the actual size of the file on disk, so a header of a few dozen bytes cannot claim hundreds of megabytes. A 128 byte input drove 500MB of allocation before this, and drives about 10KB now
- the reconstruction goes through a new `internal/spillbuf` buffer instead of a `[]byte` sized from the declared original size, so nothing is pre-sized from a claim. It holds a bounded amount in memory, spills the rest to a temp file, and blocks stream in through a fixed window. A 301KB input allocated 300MB before this
- that buffer reports only the contiguous run actually rebuilt, not where the last block landed. Reporting the declared size serves the holes between blocks as zeros it never stored, and a 64 byte block parked at `p_filesize-64` turned a 300KB input into a 64MB reader
- the blocks together cannot claim more than the declared original size, and the block count is capped. Bounding only per-block leaves the total at block count times original size

Multiple bounds checks were also written in a form that wraps on a uint64, letting out-of-range offsets through into a slice index; those are rewritten as subtractions. The LZMA parameter bytes are range-checked for the same reason: folded into the properties byte at their maximum values they wrap, and the stream then decodes with parameters that never existed instead of failing.

Two bounds outside the UPX path came with it. `elfutil.CheckAllSections` covers sections expanded lazily rather than during the parse, which the name-table gate could not reach: a 260KB ELF declaring a compressed `.symtab` drove 1.3GB through `goversion`. And `internal/file.ReaderSize` is the one place that answers how many bytes a reader really holds, since an `io.SectionReader` reports whatever nominal length it was built with.

Interface change: `scanFile` takes a context and hands back the reconstruction for each packed binary, and the caller owns it. Everything after the build info read (crypto settings, arch, symbols, the version scan) reads it, so a packed binary no longer reports its packages with no symbols.

Four gaps are reported as unknowns now, since a packed binary syft cannot unpack means its packages go silently missing: a file we could not decode, a header the bounds refused, a short reconstruction, and an ELF expansion `elfutil` declined. Everything else stays quiet, since this cataloger runs on every executable in an image.

Also drops the vendored xcoff parser: the only thing read from it was `TargetMachine`, the same two magic bytes `getGOARCHFromBin` already matched on to dispatch there.


kzantow (approved):
LGTM, I can't see any path that would allocate more than ~516MB and change.

internal/file/reader_size.go
```diff
@@ -0,0 +1,66 @@
...
+// a caller that forgot to special-case it got a bound of "nothing is allowed" or, more often, silently
+// fell back to an unbounded path. Making the caller name the failure keeps that from happening by
+// omission.
+//
+// Both *bytes.Reader and *io.SectionReader answer directly. A file handle only has Seek, so fall back to
+// a save-and-restore seek, which leaves the caller's cursor where it found it. That fallback is not
+// atomic: it moves the cursor and puts it back, so a reader being read concurrently can observe the
+// intermediate position.
+//
+// Neither answer is taken on faith, because a reader can report a length it cannot deliver: an
```
kzantow, line 23: FWIW Vale is also supposed to be able to lint golang comments among other things

syft/pkg/cataloger/golang/internal/README.md
```diff
@@ -1,4 +1,4 @@
-xcoff
```
kzantow, line 1: nit: is this file needed anymore?

internal/spillbuf/buffer.go
```diff
@@ -0,0 +1,330 @@
...
+	return covered
+}
+
+// FirstGap returns the offset of the earliest unwritten run of at least size bytes lying entirely within
+// [0, within), and whether there is one. Callers that place output at offsets of their own choosing use it
+// to fill what they left behind, without needing the buffer's bookkeeping.
+//
+// The arithmetic is in subtraction form throughout: within comes from caller input and the result is used
+// as a write offset, so a wrap here would be a write nobody bounded.
+func (b *Buffer) FirstGap(size, within int64) (int64, bool) {
```
kzantow: nit: is there a reason to export this?

syft/pkg/cataloger/golang/scan_binary.go
```diff
@@ -35,53 +59,132 @@ func scanFile(location file.Location, reader unionreader.UnionReader, captureSym
...
+	return errors.Is(err, errUPXDecompress) ||
+		errors.Is(err, errUPXSizeRefused) ||
+		errors.Is(err, errUPXPartial) ||
+		errors.Is(err, elfutil.ErrDeclaredSizeExceeded)
 }
 
 func getCryptoInformation(reader io.ReaderAt) ([]string, error) {
+	// goversion opens the file with debug/elf itself and reads .symtab plus the string table it links.
+	// Those are expanded lazily, so the section-name table bound getBuildInfo already applied does not
+	// reach them: without this gate a 260KB ELF declaring a compressed .symtab drove 1.3GB of allocation.
```
kzantow: I don't understand this comment in the context of the CheckAllSections comment how does this reduce allocation? is this the decompression-bomb case? Maybe the `CheckAllSections` comment should mention something about it doing decompression bomb check? I guess I didn't quite parse this particular description: "bounds every section syft can drive debug/elf into decompressing, without keeping the parse" as "checks for decompression attacks" if that's the intent

syft/pkg/cataloger/golang/upx.go
```diff
@@ -345,10 +891,29 @@ func parseELFPTLoadOffsets(elfHeader []byte) []uint64 {
...
+		return nil
+	}
+
+	hdrLen := uint64(len(elfHeader))
 	var offsets []uint64
 	for i := range phnum {
 		phStart := phoff + uint64(i)*uint64(phentsize)
-		if phStart+uint64(phentsize) > uint64(len(elfHeader)) {
+		// subtraction form so a large phoff cannot overflow phStart+phentsize past the buffer end.
+		// note: the `break` is load-bearing for that argument. It guarantees phoff <= hdrLen before any
```
kzantow: nit: I think this comment may not be describing this accurately anymore  -- I don't see a subtraction here

kzantow (approved):
LGTM -- UPX is pretty complicated!

syft/pkg/cataloger/golang/scan_binary.go
wagoodman: You're right... I reworded the CheckAllSections doc to lead with the attack instead of the mechanism, and the call site here now just says it's the decompression-bomb gate so you don't have to jump to elfutil to find out.

syft/pkg/cataloger/golang/upx.go
wagoodman: Subtraction's still there, just way too far from the comment: hdrLen-phStart < uint64(phentsize) (i moved the comment around)

Issue #670, opened 2021-12-13
Syft ignore docker images


stefanm8:
**What would you like to be added**:
Ability for syft to ignore specific paths and or docker container

**Why is this needed**:
Scan only source code vendors, bypassing private registry.

**Additional context**:

Scanning repository will return:

```
1 error occurred:
        * failed to determine image source: could not fetch image 'src/main/java/': failed to get image descriptor from registry:
 ```
 
 although there is no intention of scanning image, but instead scan filesystem.

**Comment by spiffcs** [comment]
Hey @stefanm8! Thanks for the issue. Linking the related issue here.

@wagoodman I'll keep this one open for now, but there might be room for consolodation.
https://github.com/anchore/syft/issues/221

**Comment by stefanm8** [comment]
Hi @spiffcs 

Is quite complementary, as in #221 does not suggest of scanning filesystem only. 

There should be at least this option.  I may be able to do it, if you give me some hints of where I should look.

Thank you

**Comment by spiffcs** [comment]
Syft also has schemes that can be provided as prefixes to get the behavior you're asking about in that final example when scanning the file system.

```sh
syft dir:path/to/yourproject
syft: file:path/to/your/project/file
```

Example:
```wh
git clone git@github.com:anchore/syft.git
syft dir:.
```

![Screen Shot 2021-12-13 at 11 05 06 PM](https://user-images.githubusercontent.com/32073428/145931197-075561b6-9891-4c22-abbf-45070e90093f.png)

kzantow:
There's a [draft PR](https://github.com/anchore/syft/pull/695) for #221, which allows for excluding paths from directory, archive, and image scans. Would this accomplish what you're looking to do?

spiffcs:
Closing this issue for now since we have the ability to ignore paths per the above comment. Also with the `dir` and `file` directives users have options for generating very specific SBOM configurations. If more features are need feel free to comment and can consider reopening or filing a new issue.

PR #5196, opened 2026-08-19
fix: move image hardlink handling upstream during image indexing

wagoodman:
Follow-up to [#5029](https://github.com/anchore/syft/pull/5029), paired with stereoscope [#670](https://github.com/anchore/stereoscope/pull/670).

Some background: image scans used to collapse a set of hardlinks onto one file, so only one path per inode showed up while a `dir:` scan of the same filesystem listed all of them. #5029 fixed that in syft's image resolvers, on the reasoning that stereoscope should stay honest about what the tar said and only syft's presentation should change. That worked, but it could only reach the paths the resolvers looked up by name, so `FilesByMIMEType` was explicitly left alone.

stereoscope [#670](https://github.com/anchore/stereoscope/pull/670) moves the fix upstream instead. A tar hardlink header now adopts the file it names at index time, picking up its size, mime type, type and content, so a hardlinked name is simply a regular file by the time syft sees it. That covers `FilesByMIMEType` and every other consumer, and it makes the #5029 workaround unreachable, gated as it was on a file type that no longer appears.

So... this PR deletes the workaround: both `resolveHardLinkTarget` implementations, the `hardLinkAtPath` lookups and the extra tree walk plus catalog hit they cost on every `FilesByPath` call. `file.NewVirtualLocationFromImage` stays, exported and working, it just has no callers in syft anymore.

Behavior changes worth knowing about:

- hardlinked names now reach mime-driven catalogers. On images built around multi-call binaries this is a big jump: the busybox fixture goes from 3 file entries and 2 executable records to 402 of each, and its syft-json from 10K to 175K. Distro and toolchain images (ubuntu, debian, golang) move by roughly one entry, and package counts did not change on any of them
- `"HardLink"` stops appearing in syft-json for well-formed images. The decode side is kept for reading older SBOMs
- one shape regresses back to pre-#5029 behavior: a hardlink stereoscope could not adopt (link name absent, empty, naming a directory, naming another un-adopted hardlink, or pointing into a lower layer) keeps `TypeHardLink` and gets collapsed onto its target's path again. No mainstream builder emits the cross-layer case, they keep the inode table per archive

Layer and image size are unaffected, verified identical before and after on the hardlink fixture.

## Type of change

## Checklist


go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/anchore/go-sync v0.1.1
 	github.com/anchore/go-version v1.2.2-0.20200701162849-18adb9c92b9b
 	github.com/anchore/packageurl-go v0.2.0
-	github.com/anchore/stereoscope v0.3.0
+	github.com/anchore/stereoscope v0.3.1-0.20260819175631-21f4eb5e7f36
```
wagoodman: change me to a `main` reference before merge!

spiffcs (approved):

PR #5198, opened 2026-08-20
fix(binary): select the English VERSIONINFO block from localized PE binaries

### LINKED ISSUE #5177: pe-binary-package-cataloger selects an unexpected language block from multi-language VERSIONINFO resources
**What happened**:

When scanning a Windows PE binary (`.exe`) that contains a multi-language `VERSIONINFO` resource, Syft's `pe-binary-package-cataloger` extracts an unexpected `ProductName`  from a language/codepage block.

Concrete example: scanning `Driver\dpinst-x86.exe` (Microsoft's Driver Package Installer, `DPInst.exe`) produces the following component in the generated CycloneDX SBOM:

```json
{
  "bom-ref": "e7cdc208d7cea5b8",
  "type": "application",
  "name": "Instalador de paquetes de controladores",
  "version": "2.1",
  "cpe": "cpe:2.3:a:Instalador_de_paquetes_de_controladores:Instalador_de_paquetes_de_controladores:2.1:*:*:*:*:*:*:*",
  "properties": [
    { "name": "syft:package:foundBy", "value": "pe-binary-package-cataloger" },
    { "name": "syft:package:type", "value": "binary" },
    { "name": "syft:package:metadataType", "value": "pe-binary" },
    { "name": "syft:location:0:path", "value": "\\Driver\\dpinst-x86.exe" }
  ]
}
```

The extracted `ProductName` is the Spanish localization of the string ("Instalador de paquetes de controladores"), rather than the English "Driver Package Installer". As a result, the generated CPE
(`cpe:2.3:a:Instalador_de_paquetes_de_controladores:...`) does not match any real-world CPE dictionary entry for this well-known Microsoft/WDK component, which prevents correct vulnerability matching downstream (e.g. via Grype).

Ideally Syft would consistently prefer a US-English (0x0409) StringFileInfo block if the binary contains one, since this tends to produce the most reliable results for CPE/PURL generation.

Scan an diconary with the dpinst-x86.exe 

[dpinst.zip](https://github.com/user-attachments/files/31022079/dpinst.zip)

**Environment**:
- Output of `syft version`: 1.44.0_windows_amd64

hamodywe:
## Description

Scanning a PE binary that ships several localizations names the package after whichever language sorts last in its resource directory.

A localized binary carries one complete `VS_VERSIONINFO` resource per language, each under its own entry at the language level of the resource tree, and each using exactly the same string keys as all the others. Syft read every one of them into a single flat map, so the block read last decided the outcome — and since directory entries are ordered by ID, that is the highest LANGID in the file.

`dpinst-x86.exe` from #5177 carries 21 such blocks. es-ES (0x0c0a) is the largest ID, so:

```console
$ syft scan dir:dpinst -o table            # before
NAME                                     VERSION  TYPE
Instalador de paquetes de controladores  2.1      binary

$ syft scan dir:dpinst -o table            # after
Driver Package Installer  2.1      binary
```

and the generated CPE goes from `cpe:2.3:a:Instalador_de_paquetes_de_controladores:Instalador_de_paquetes_de_controladores:2.1:*` to `cpe:2.3:a:Driver_Package_Installer:Driver_Package_Installer:2.1:*`, which is what downstream matching needs. The en-US block was in the file the whole time; it just got overwritten.

Version resources are now collected per language and one is chosen after the whole tree has been read: **US English, then any other English variant, then a language-neutral block, then the lowest remaining LANGID**. Values are never merged across languages — an English `ProductName` next to a Spanish `FileDescription` describes no version resource that exists in the file. The selection is a total order over the LANGIDs involved, so it does not depend on map iteration order.

Two further problems in the same walk are fixed alongside it, since the language work runs straight through both:

- **Only the first string table of a `StringFileInfo` block was read**, and the rest of the block was then walked as though the tables were siblings of the block. A block may hold one table per language, so this is the second way a binary can carry several translations of the same keys. A table's `szKey` is its own language and codepage (e.g. `040904b0`), and that is now what the table is filed under, with the directory entry's LANGID as the fallback for tables that state no language.
- **A truncated version resource never finished parsing.** When a block's declared length exceeds the bytes that follow it, the reads stop returning data and the offsets stop advancing, so the loop can never reach the claimed length. Confirmed against `main` by truncating a well-formed resource at every length: lengths 134–137 of 460 hang. The string loop is now bounded by the reader as well as by the declared length.

## Type of change

## Checklist

**On the tests:** they build PE files rather than committing binaries — a DOS header, the PE headers and a single `.rsrc` section holding a real `type -> name -> language` tree of `RT_VERSION` entries — so the multi-language shapes are exercised without a fixture image or a redistributable EXE in the repo. Covered: language preference and each fallback step, order independence, several string tables inside one block, tables filed by their declared language rather than the directory entry, no merging across languages, a resource with no string tables at all (the file version still comes from `VS_FIXEDFILEINFO`), and the truncation sweep.

Every new `Read`-level assertion was checked against `main` first: 6 of 8 subtests and both top-level cases fail there with the Spanish values, and pass with the fix. The fix was also run against the binary attached to the issue, x86 and amd64, which is where the before/after above comes from.

**One limitation worth stating:** the Docker-backed tests in this package (`Test_Read_DotNetDetection` and the binary/dotnet cataloger suites) could not run on my machine. They fail identically on a pristine checkout here, so nothing is masked, but CI is the real check on those.

## Issue references

Fixes #5177

---

*Disclosure: written with AI assistance (the commit carries a `Co-Authored-By` trailer). I verified the root cause by walking the resource tree of the binary attached to the issue and reproducing it through syft's own reader, before and after, rather than taking the report on trust.*

### DIFF
--- a/syft/pkg/cataloger/internal/dotnet/pe/pe.go
+++ b/syft/pkg/cataloger/internal/dotnet/pe/pe.go
@@ -715,7 +715,7 @@ func parseStringTables(reader *bytes.Reader, sfiHeader peStringFileInfo, sfiOffs
 		lang := languageFromStringTableKey(szKey, language)
 		languages = append(languages, lang)
 
-		if err := parseStringTable(reader, stHeader, stOffset, offset, &sfiOffset, versions.fields(lang)); err != nil {
+		if err := parseStringTable(reader, stHeader, stOffset, offset, &sfiOffset, versions.stringTableFields(lang)); err != nil {
 			return languages, err
 		}
 	}
--- a/syft/pkg/cataloger/internal/dotnet/pe/pe.go
+++ b/syft/pkg/cataloger/internal/dotnet/pe/pe.go
@@ -17,7 +17,15 @@ import (
 	"github.com/anchore/syft/syft/internal/unionreader"
 )
 
-const peMaxAllowedDirectoryEntries = 0x1000
+const (
+	peMaxAllowedDirectoryEntries = 0x1000
+
+	// resourceLanguageLevel is the depth in the resource directory tree at which entries are keyed by LANGID.
+	// The tree is always organized as type (level 0) -> name (level 1) -> language (level 2) -> data.
+	//
+	// source: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#the-rsrc-section
+	resourceLanguageLevel = 2
+)
 
 var imageDirectoryEntryIndexes = []int{
 	pe.IMAGE_DIRECTORY_ENTRY_RESOURCE,       // where version resources are stored
@@ -39,6 +47,8 @@ type File struct {
 	EmbeddedDepsJSON string
 
 	// VersionResources is a map of version resource keys to their values found in the VERSIONINFO resource directory.
+	// Binaries that are localized into several languages carry one set of these values per language; this holds
+	// the single set selected by versionResourceTables.preferred (never a mixture of languages).
 	VersionResources map[string]string
 }
 
@@ -162,15 +172,18 @@ func Read(f file.LocationReadCloser) (*File, error) {
 		return nil, fmt.Errorf("unable to parse PE sections: %w", err)
 	}
 
-	dirs := u32set.New()                        // keep track of the RVAs we have already parsed (prevent infinite recursion edge cases)
-	versionResources := make(map[string]string) // map of version resource keys to their values
-	resourceNames := strset.New()               // set of resource names found in the PE file
-	err = parseResourceDirectory(sections[pe.IMAGE_DIRECTORY_ENTRY_RESOURCE], dirs, versionResources, resourceNames)
+	walk := &resourceWalk{
+		dirs:     u32set.New(),               // keep track of the RVAs we have already parsed (prevent infinite recursion edge cases)
+		names:    strset.New(),               // set of resource names found in the PE file
+		versions: newVersionResourceTables(), // version resource values, kept separate per language
+	}
+
+	err = parseResourceDirectory(sections[pe.IMAGE_DIRECTORY_ENTRY_RESOURCE], walk, 0)
 	if err != nil {
 		return nil, err
 	}
 
-	c, err := parseCLR(sections[pe.IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR], resourceNames)
+	c, err := parseCLR(sections[pe.IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR], walk.names)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse PE CLR directory: %w", err)
 	}
@@ -184,7 +197,7 @@ func Read(f file.LocationReadCloser) (*File, error) {
 		Location:         f.Location,
 		CLR:              c,
 		EmbeddedDepsJSON: embeddedDepsJSON,
-		VersionResources: versionResources,
+		VersionResources: walk.versions.preferred(),
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


CAOShurong:
## Exact-head verification (head `1286dd77` = PR head, merge-base `5c6cf08a`, go1.26.7 windows-amd64)

Independent verification against the binaries attached to #5177 (I had RCA'd that issue separately and still had the files around):

- `go build ./...` and `go vet ./syft/pkg/cataloger/internal/dotnet/pe/...` clean.
- All new tests pass: the 8 language-preference subtests, determinism, multi-table-per-block, no-string-tables, truncation sweep, and the `languageFromStringTableKey` / `effective` / `preferred` unit tests.
- The only failures in this package are the two Docker-backed suites (`Test_Read_DotNetDetection`, `Test_extractDepsJSONFromBundle_Versions`) — `exec: "docker": executable not found`, matching the limitation stated in the description. Everything else that fails on this tree (`./syft/pkg/cataloger/binary/`, cpegenerate index-generator) fails **identically at the merge base**, so zero regressions from this change.
- End-to-end with a binary built from this branch: dpinst x86 + amd64 flip to `Driver Package Installer` with CPE `cpe:2.3:a:Driver_Package_Installer:Driver_Package_Installer:2.1:*`, exactly as described; notepad unchanged; and before/after JSON scans of 10 localized System32 binaries produce byte-identical package name/version sets, so the reordering is a no-op for already-correct single-language binaries.

The fix itself is right, and notably broader than just the language preference: it also fixes two real bugs in the same walk — only the first string table of a block was ever read, and a truncated resource could hang the parse loop. Both are covered by new tests.

### One real edge case: a strings-less VERSIONINFO resource can outrank a real table

The synthetic `FileVersion`-only table appended for resources without string tables participates in `preferred()` as a normal candidate:

https://github.com/anchore/syft/blob/1286dd77896034fe4d8d60f3ee0fbc9a60045243/syft/pkg/cataloger/internal/dotnet/pe/pe.go#L672-L675

Since it is keyed `{directory: <entry LANGID>}` and its rank comes from that entry, it can beat a genuine localized table. Reproduced with the PR's own fixture builder (`Read(peFixture(t, {lang: 0x0c0a, fields: spanishFields}, {lang: 0x0411, fields: nil}))`):

```console
map[string]string{"FileVersion":"2.1.0.0"}   // the complete Spanish table is discarded
```

Any lower-numbered non-English entry (or an entry below rank 2) whose resource has fixed info but no strings wins over a full English-absent table; with `0x0409` present it stays correct because rank 0 outranks everything. Suggested fixes, either works:

1. exclude synthetic tables from selection — e.g. track which entries came from real string tables and let `preferred()` skip the rest unless there are none;
2. or keep the synthetic entry out of `byLanguage` and add its derived `FileVersion` to whatever `preferred()` returns afterwards.

Worth noting this shape does occur in the wild: installers commonly carry one VERSIONINFO per locale plus a stub resource, so I'd treat it as pre-merge rather than follow-up.

Minor, non-blocking: `languageFromStringTableKey` parses any ≥4-char hex-ish prefix (e.g. `"12345678"` → table lang `0x1234`), fine in practice since a wrong guess only affects fallback ranking, never merges languages.

Thanks for including the truncation sweep — lengths 134–137 hanging on main is a nasty class of bug to catch, and bounding by the reader is the right call.

hamodywe:
Thanks for the exact-head run, and for catching this — you're right that it's pre-merge. Fixed in 747037f.

I reproduced it with your fixture before touching anything:

```console
--- FAIL: Test_Read_MultiLanguageVersionResources/a_block_with_no_string_tables_never_displaces_one_that_has_them
    unexpected version resources (-want +got):   map[string]string{
    - 	"CompanyName":     "Microsoft Corporation",
    - 	"FileDescription": "Instalador de paquetes de controladores",
    - 	"FileVersion":     "2.1",
    + 	"FileVersion":     "2.1.0.0",
    - 	"ProductName":     "Instalador de paquetes de controladores (DPInst)",
      }
```

I went with your first suggestion. `versionResourceTable` now records whether its values came from a string table the file actually carries; `parseStringTables` sets that via `stringTableFields`, the derived-FileVersion pass does not. `preferred()` ranks real tables among themselves first and falls back to the rest, so a strings-less block still gets used when it is all the file has.

I preferred that over keeping the synthetic entry out of `byLanguage` because the fallback stays a single ranked selection rather than a merge — a file with several strings-less blocks and no real table still picks one deterministically instead of taking whichever FileVersion was appended last.

Three tests cover it: your two-block case, the single strings-less block (which must still report `{"FileVersion": "2.1.0.0"}` — that's the over-fix guard), and a direct `preferred()` unit test.

On `languageFromStringTableKey`: I left it as is deliberately. The 8-char key is a convention, not something the format enforces, and compilers do emit other lengths; since the parsed value only ever feeds the fallback ranking and never merges two languages together, being permissive costs at most a fallback ordering and being strict would drop a real language claim. Happy to tighten it if you'd rather have the stricter read.

Everything else in the package passes here; the Docker-backed suites are still the only failures on this machine.

PR #4953, opened 2026-06-03
fix(maven): add an explicit groupid mapping for some older tomcat artifact names

westonsteimel:
## Description

This adds a maven groupid lookup to `org.apache.tomcat` for:
- `catalina`                                                
- `catalina-ant`                                            
- `catalina-ha`                                              
- `coyote`

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [ ] I have added unit tests that cover changed behavior
- [ ] I have tested my code in common scenarios and confirmed there are no regressions
- [ ] I have added comments to my code, particularly in hard-to-understand sections

## Issue references

Fixes https://github.com/anchore/grype/issues/3491

PR #5200, opened 2026-08-20
fix(java): use the OSGi Bundle-Name as the artifact ID

### LINKED ISSUE #4097: Invalid purl extracted for jasper-el.jar file bundled within the tomcat installation
**What happened**:
If you download the tomcat installation it seems the jasper-el.jar if named `jasper-el.jar` which is a bit different than on https://repo1.maven.org/maven2/org/apache/tomcat/tomcat-jasper-el/9.0.106/

This seems to impact syft and make it generate a purl of:
```
"bom-ref": "pkg:maven/org.apache.tomcat-jasper-el/jasper-el@9.0.106?package-id=cb7d0219e16657b6",
"purl": "pkg:maven/org.apache.tomcat-jasper-el/jasper-el@9.0.106",
"value": "/jasper-el.jar"
```

instead of:
```
"bom-ref": "pkg:maven/org.apache.tomcat/tomcat-jasper-el@9.0.106?package-id=6cb2c31447a3b6ed",
"purl": "pkg:maven/org.apache.tomcat/tomcat-jasper-el@9.0.106",
"value": "/tomcat-jasper-el-9.0.106.jar"
```

We'd like the filename to not influence the package detection - especially that it's bundled like this within the actual distribution of tomcat itself/out of the box.

1. Copy both jar files into a temp folder
2. Run: `syft . -o cyclonedx-json > test.cyclonedx.json`
3. Inspect test.cyclonedx.json

Alternatively:
1. Generate the SBOM for the tomcat distribution
2. Notice the PURL for jasper-el is invalid (compare with the purl presented here: https://central.sonatype.com/artifact/org.apache.tomcat/tomcat-jasper-el/9.0.107) `pkg:maven/org.apache.tomcat-jasper-el/jasper-el@9.0.107` vs pkg:maven/org.apache.tomcat/tomcat-jasper-el@9.0.107 )

Hope this is all

**Environment**:
- Output of `syft version`:
```
Application:   syft
Version:       1.29.0
BuildDate:     2025-07-21T19:23:14Z
GitCommit:     Homebrew
Platform:      darwin/arm64
GoVersion:     go1.24.5
Compiler:      gc
SchemaVersion: 16.0.36
```
`MacOS`

ashvinctrl:
## Description

Tomcat hand-writes the OSGi headers for every jar it ships and includes no pom, so syft names those jars from the filename. `catalina.jar` comes out as `pkg:maven/org.apache.tomcat-catalina/catalina@10.1.54` when the published artifact is `org.apache.tomcat:tomcat-catalina`, and nothing is published under the former, so a stock Tomcat install matches no advisories at all.

The artifact ID is already in the manifest as `Bundle-Name`. That field is often a human readable title rather than an ID, so this only prefers it over the filename when `Bundle-SymbolicName` is composed from it. In the same directory `tomcat-jdbc.jar` ("Apache Tomcat JDBC Connection Pool") and `ecj-4.27.jar` ("Eclipse Compiler for Java(TM)") both fail that check and keep the names they get today.

```
# before
pkg:maven/org.apache.tomcat-catalina/catalina@10.1.54
# after
pkg:maven/org.apache.tomcat/tomcat-catalina@10.1.54
```

15 of the 35 jars in `apache-tomcat-10.1.54/lib` get corrected coordinates and all 15 resolve on Maven Central under `org.apache.tomcat`. Scanning the before and after SBOMs with grype v0.117.0 goes from 0 matches to 7 on 10.1.54 (3 critical, 3 high, 1 low, all on tomcat-catalina) and from 4 to 17 on both 9.0.111 and 11.0.13. Every new match is an exact-direct-match against a GHSA advisory for that artifact rather than CPE over-matching, and no existing match is lost. The group ID table already had the corrected names except `tomcat-coyote-ffm`, which is added here. I also ran a before/after over 39 jars pulled from Maven Central (spring, jetty, felix, eclipse platform, servicemix, cxf, camel, activemq, jackson, netty and others) and no purls changed.

## Type of change

## Checklist

## Issue references

Fixes #4097

Also the cause of https://github.com/anchore/grype/issues/3491, and the reason #4953 could not work on its own: the group ID mapping was already correct, but the artifact ID was still coming from the filename.

### DIFF
--- a/syft/pkg/cataloger/internal/cpegenerate/java_groupid_map.go
+++ b/syft/pkg/cataloger/internal/cpegenerate/java_groupid_map.go
@@ -156,6 +156,7 @@ var DefaultArtifactIDToGroupID = map[string]string{
 	"tomcat-catalina-jmx-remote":                  "org.apache.tomcat",
 	"tomcat-catalina-ws":                          "org.apache.tomcat",
 	"tomcat-coyote":                               "org.apache.tomcat",
+	"tomcat-coyote-ffm":                           "org.apache.tomcat",
 	"tomcat-dbcp":                                 "org.apache.tomcat",
 	"tomcat-el-api":                               "org.apache.tomcat",
 	"tomcat-i18n-cs":                              "org.apache.tomcat",
--- a/syft/pkg/cataloger/java/parse_java_manifest.go
+++ b/syft/pkg/cataloger/java/parse_java_manifest.go
@@ -145,6 +145,48 @@ func extractNameFromApacheMavenBundlePlugin(manifest *pkg.JavaManifest) string {
 	return ""
 }
 
+// extractNameFromOSGiBundleName returns the maven artifact ID declared by an OSGi bundle that was not
+// built with the Apache Maven Bundle Plugin. Projects that hand-write their OSGi headers set Bundle-Name
+// to the artifact ID and compose Bundle-SymbolicName as "<prefix><separator><artifact ID>". Apache Tomcat
+// does this for every jar it ships:
+//
+//	Bundle-Name:         tomcat-catalina
+//	Bundle-SymbolicName: org.apache.tomcat-catalina
+//
+// Bundle-Name on its own is not trustworthy, since it is often a human readable title (catalina.jar's
+// sibling tomcat-jdbc.jar uses "Apache Tomcat JDBC Connection Pool"), so only use it when the symbolic
+// name is built from it. When the two agree the manifest is a better source for the artifact ID than the
+// filename, which is frequently a shortened alias: catalina.jar is org.apache.tomcat:tomcat-catalina.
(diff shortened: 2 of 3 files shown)


ashvinctrl:
Only DCO has run on this PR so far. The rest of the checks are still waiting on a maintainer to approve workflow runs for a first-time contributor, so CI has not actually exercised the change yet. Could someone kick that off?

I ran it locally on the PR head in the meantime:

- `go build ./...` is clean.
- The new `selectName` cases in `parse_java_manifest_test.go` pass.
- `go test ./syft/pkg/cataloger/java/...` reports 15 failures, but those same 15 fail identically on upstream main on this machine (Windows fixture and docker-dependent tests), so nothing here is a regression from the change.

Happy to answer anything on the Bundle-Name heuristic or to add cases if you want the guard tightened.

PR #5201, opened 2026-08-20
fix(rpm): keep the epoch when parsing RPM manifest packages

sueun-dev:
## Description

`newMetadataFromManifestLine` parses the RPM manifest used in Mariner / Azure Linux distroless containers. It decoded the epoch and size into the same local variable, set `epoch = &converted`, then reassigned `converted` to the size field:

```go
converted, err := strconv.Atoi(parts[8]) // EPOCHNUM
...
epoch = &converted
converted, err = strconv.Atoi(parts[6])  // SIZE
```

For a manifest line with EPOCH `2` and SIZE `45000`, Syft produced version `45000:9.0-1.cm2` and PURL `pkg:rpm/vim@9.0-1.cm2?arch=x86_64&epoch=45000&upstream=vim-9.0-1.cm2.src.rpm` instead of keeping epoch `2`.

This binds the epoch to its own variable so decoding the size cannot overwrite it. I also added a `container-manifest-2` fixture line with a real epoch; the existing lines all used EPOCH `(none)`, so this branch was not covered.

## Type of change

## Checklist

## Issue references

None.

## Testing

With only the code change reverted and the new test kept, `go test ./syft/pkg/cataloger/redhat -run TestParseRpmManifest -count=1 -v` fails with `Epoch -: 2 +: 45000`, `Version -: 2:9.0-1.cm2 +: 45000:9.0-1.cm2`, and `PURL` using `epoch=45000`.

With the fix:

```sh
go test ./syft/pkg/cataloger/redhat -run TestParseRpmManifest -count=1 -v
go build ./...
golangci-lint run ./syft/pkg/cataloger/redhat/...
gofmt -l syft/pkg/cataloger/redhat/package.go syft/pkg/cataloger/redhat/parse_rpm_manifest_test.go
git diff --check origin/main...HEAD
```

`TestParseRpmFiles` still fails here on a clean `main` because `testdata/rpms` is not present in a plain checkout, and the broader redhat cataloger tests need a local Docker daemon. Those are unrelated to this change. `go vet ./syft/pkg/cataloger/redhat` also reports the same pre-existing test-file warnings on clean `origin/main`.

### DIFF
--- a/syft/pkg/cataloger/redhat/package.go
+++ b/syft/pkg/cataloger/redhat/package.go
@@ -59,17 +59,13 @@ func newMetadataFromManifestLine(entry string) (*pkg.RpmDBEntry, error) {
 	version := versionParts[0]
 	release := versionParts[1]
 
-	converted, err := strconv.Atoi(parts[8])
 	var epoch *int
-	if err != nil || parts[5] == "(none)" {
-		epoch = nil
-	} else {
-		epoch = &converted
+	if epochNum, err := strconv.Atoi(parts[8]); err == nil && parts[5] != "(none)" {
+		epoch = &epochNum
 	}
 
-	converted, err = strconv.Atoi(parts[6])
 	var size int
-	if err == nil {
+	if converted, err := strconv.Atoi(parts[6]); err == nil {
 		size = converted
 	}
 	return &pkg.RpmDBEntry{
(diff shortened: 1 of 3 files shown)


spiffcs (approved):
Thanks @sueun-dev! I made a small change and updated the bug fix to use parseEpoch. TY for adding the new test and getting this up for us to review.

PR #5202, opened 2026-08-20
fix(binary): gzip classifier & GNU gzip identity string

### LINKED ISSUE #5171: gzip binary classifier reports false-positive GNU gzip from BusyBox multicall binary via applet symlink
## What happened

The `gzip` binary classifier (`binary-classifier-cataloger`) reports a false-positive **GNU gzip** package when scanning any image/filesystem that contains a **BusyBox** multicall binary with a `gzip` applet symlink.

BusyBox implements every applet (`gzip`, `gunzip`, `tar`, ...) as a symlink back to the single multicall binary. Two independent problems in the `gzip` classifier combine to produce a bogus package:

1. **Glob over-match.** The classifier uses `FileGlob: "**/gzip"`, which matches the BusyBox `gzip` applet symlink. Syft resolves the symlink and scans the *BusyBox* binary, then attributes a `gzip` package to it.
2. **Unanchored version matcher.** The evidence matcher is `\x00(?P<version>[0-9]+\.[0-9]+)\x00` — it is **not anchored to the string `gzip`**, so it extracts *any* null-delimited `N.N` token found anywhere in the binary. That arbitrary token becomes the "version", and the classifier stamps `cpe:2.3:a:gnu:gzip:<token>` onto it.

For comparison, the sibling `busybox` classifier is correctly anchored to its real banner:

```go
// busybox — correct: anchored to the real version banner
EvidenceMatcher: `(?m)BusyBox\s+v(?P<version>[0-9]+\.[0-9]+\.[0-9]+)`

// gzip — buggy: no anchor to "gzip", matches any null-delimited N.N
EvidenceMatcher: `\x00(?P<version>[0-9]+\.[0-9]+)\x00`
```

The result is a phantom `gzip` package that shares the exact same real file as the (correct) `busybox` package, but carries the GNU gzip CPE. Downstream, Grype then matches this phantom package against decades-old GNU gzip CVEs (e.g. CVE-2004-1349, CVE-2005-0758, CVE-2022-1271), producing false-positive vulnerabilities.

## What you expected to happen

Syft should not report a standalone GNU gzip package for a BusyBox multicall binary. The `gzip` applet is provided *by* BusyBox and should either be omitted or associated with the BusyBox package — and it should certainly not receive the `cpe:2.3:a:gnu:gzip` CPE with an arbitrary version.

## Steps to reproduce

BusyBox binaries in the wild routinely contain incidental null-delimited `N.N` byte tokens, which is enough to trigger this on unmodified images. The Dockerfile below appends one only to make the reproducer **deterministic across build hosts**; the classifier logic being exercised is unchanged.

`Dockerfile`:

```dockerfile
FROM busybox:1.37.0

# BusyBox ships `gzip` as an applet symlink to the multicall binary.
# Append a null-delimited "N.N" token so the unanchored version matcher
# has something to grab (real-world builds already contain such tokens).
RUN cp /bin/busybox /bin/busybox.real \
 && printf '\x001.0\x00' >> /bin/busybox.real \
 && ln -sf /bin/busybox.real /bin/gzip
```

```console
$ docker build -t syft-gzip-repro:latest .
$ syft scan syft-gzip-repro:latest -o table
NAME     VERSION  TYPE
busybox  1.37.0   binary
gzip     1.0      binary     <-- false positive
```

The phantom package (`-o json`), sharing the same underlying file as `busybox`:

```json
{
  "name": "gzip",
  "version": "1.0",
  "type": "binary",
  "foundBy": "binary-classifier-cataloger",
  "purl": "pkg:generic/gzip@1.0",
  "cpes": [{ "cpe": "cpe:2.3:a:gnu:gzip:1.0:*:*:*:*:*:*:*" }],
  "locations": [{ "path": "/bin/busybox.real", "accessPath": "/bin/gzip" }]
}
```

Downstream vulnerability match (Grype) from the phantom package:

```console
$ grype syft-gzip-repro:latest
NAME   INSTALLED  FIXED IN  TYPE    VULNERABILITY   SEVERITY
gzip   1.0        1.12      binary  CVE-2022-1271   High
gzip   1.0        1.3.5     binary  CVE-2005-0758   Medium
gzip   1.0        1.3       binary  CVE-2004-1349   Low
gzip   1.0                  binary  CVE-2004-0603   High
...
```

## Suggested fix

Anchor the `gzip` evidence matcher to the actual gzip identity string rather than a bare `\x00N.N\x00` (mirroring how the `busybox`, and the recently-fixed `fluent-bit` in #3701, classifiers are anchored). For example, require the literal token `gzip` adjacent to the version, e.g. something along the lines of `gzip[\s\x00]+(?P<version>[0-9]+\.[0-9]+(\.[0-9]+)?)`. Optionally, also skip files whose contents resolve to a BusyBox multicall binary so applet symlinks (`gzip`, `tar`, `wget`, ...) don't get misclassified as their standalone GNU/upstream equivalents.

## Anything else

- The buggy classifier is present through `main` (`syft/pkg/cataloger/binary/classifiers.go`, the `gzip-binary` classifier).
- This affects the whole family of BusyBox applets that share a name with a standalone binary classifier, not just `gzip`.

spiffcs:
## Description

The `gzip-binary` classifier globs `**/gzip` and extracted its version with an unanchored `\x00(?P<version>[0-9]+\.[0-9]+)\x00`. BusyBox ships `gzip` as an applet symlink to its multicall binary, so syft resolved the symlink, grabbed any incidental NUL-delimited `N.N` token from the BusyBox binary, and emitted a phantom GNU gzip package with `cpe:2.3:a:gnu:gzip:<version>`

Grype would then match against decades-old GNU gzip CVEs.

The classifier now requires a GNU gzip identity string in addition to the version. `FileContentsVersionMatcher` requires *all* patterns to match, so a bare `N.N` token no longer yields a package on its own:

```go
EvidenceMatcher: m.FileContentsVersionMatcher(
    `%s: %s: not in gzip format`,
    `\x00(?P<version>[0-9]+\.[0-9]+)\x00`,
),
```

The `%s: %s: not in gzip format` string is present in every GNU gzip checked (1.5 through 1.13) and absent from BusyBox.

Scanning the reproducer image from the issue:

### before
```
NAME     VERSION  TYPE
busybox  1.37.0   binary
gzip     1.0      binary
```
### after
```
NAME     VERSION  TYPE
busybox  1.37.0   binary
```

Type of change

Checklist

### Tests
- Added testdata/classifiers/negative/gzip — a BusyBox banner plus a `\x00N.N\x00` token. This is picked up by the directory-driven TestClassifierCataloger_DefaultClassifiers_NegativeCases. It fails on main and passes with this change.

- Regenerated the gzip/1.12/linux-amd64 snippet via `testdata/capture-snippet.sh --search-for "not in gzip format" --prefix-length 20 --length 130` so it spans both patterns (376 B, within the 1000 B fixture limit). The existing positive case still detects gzip 1.12 off the snippet.

Fixes #5171


kzantow (approved):
LGTM 👍

PR #5212, opened 2026-08-21
fix(cpe): generate the correct CPE for Microsoft Edge PE binaries

### LINKED ISSUE #4429: Incorrect CPE for Edge on Windows
**What happened**:
I scaned a Windows 11 IoT x64 in a VM.
`syft scan dir:"C:\" -o cyclonedx-json=D:\vw11iot-cdx.json`
and scaned it on Ubuntu with grype also imported to dependency-track, which both show no vulnerabilities of edge.

Generated CPE is:
`cpe:2.3:a:Microsoft_Edge:Microsoft_Edge:122.0.2365.106:*:*:*:*:*:*:*`
as stated in nvd.nist.gov should be:
`cpe:2.3:a:microsoft:edge_chromium:122.0.2365.106:*:*:*:-:*:*:*`

Similar to #4275 

**Environment**:
- Output of `syft version`:
```
PS C:\WINDOWS\system32> syft --version
syft 1.38.0
```
Windows 11 IoT Enterprise LTSC, 24H2

Thanks :-)

pujitha24:
## Description

Microsoft Edge PE binaries (`msedge.exe`) on Windows were getting a CPE generated straight from the raised package name ("Microsoft Edge"), producing `cpe:2.3:a:Microsoft_Edge:Microsoft_Edge:<version>:*:*:*:*:*:*:*` instead of the NVD dictionary's actual entry, `cpe:2.3:a:microsoft:edge_chromium:<version>:*:*:*:*:*:*:*`.

This adds PE metadata hints for Edge in `syft/pkg/cataloger/internal/cpegenerate/candidate_for_pe.go`, following the existing Ghostscript/Git-for-Windows pattern in that file, per the fix location a maintainer pointed at in the issue thread.

The mismatched candidate wasn't merely "less accurate" — it consistently ranked first. `syft/format/internal/cyclonedxutil/helpers/cpe.go` (and `syft/format/internal/backfill.go`) use `p.CPEs[0]` as the package's sole CycloneDX `cpe` field, and `syft/cpe/by_specificity.go`'s tie-break sorts by combined field character count, so the longer `Microsoft_Edge:Microsoft_Edge` (28 chars) always outranked the correct `microsoft:edge_chromium` (22 chars). Because of this, adding the correct candidate alongside the existing default wasn't sufficient on its own — the incorrect default also had to be cleared for the Edge case, which is why `candidateVendorsForPE`/`candidateProductsForPE` now mutate the caller's candidate set in place instead of returning one to union in (documented in a comment, since this is a deliberate small deviation from sibling helpers in the same package). For SPDX (`external_refs.go`), which emits all generated CPEs rather than just the first, this fix isn't just about ordering — previously the correct `microsoft:edge_chromium` candidate wasn't generated at all.

Net effect: `syft scan` (and any consumer of its CycloneDX/SPDX output, e.g. Grype or Dependency-Track) now generates a CPE for Edge that actually matches the NVD CPE dictionary entry, so CVEs for Microsoft Edge can be matched correctly. This is a CPE-generation correctness fix, not a crash or a vulnerability in Syft itself — the prior behavior silently produced a non-matching CPE, causing downstream vulnerability scanners to false-negative on Edge CVEs.

No CLI flags or output schema changes; only the generated CPE values change for this specific binary.

## Type of change

## Checklist

## Issue references

Fixes #4429

### DIFF
--- a/syft/pkg/cataloger/internal/cpegenerate/candidate_for_pe.go
+++ b/syft/pkg/cataloger/internal/cpegenerate/candidate_for_pe.go
@@ -6,14 +6,20 @@ import (
 	"github.com/anchore/syft/syft/pkg"
 )
 
-// candidateVendorsForPE returns vendor candidates for PE (BinaryPkg) packages based on common metadata hints.
-// Specifically, normalize Ghostscript binaries to vendor "artifex" when detected.
-func candidateVendorsForPE(p pkg.Package) fieldCandidateSet {
-	candidates := newFieldCandidateSet()
+// isMicrosoftEdgePE returns true if the PE metadata resembles a Microsoft Edge binary (e.g. msedge.exe).
+func isMicrosoftEdgePE(product, fileDesc string) bool {
+	return product == "microsoft edge" || fileDesc == "microsoft edge"
+}
 
+// candidateVendorsForPE adds vendor candidates for PE (BinaryPkg) packages based on common metadata hints,
+// mutating the given set in place. Specifically, normalize Ghostscript binaries to vendor "artifex" when detected.
+// Unlike sibling candidateXForY helpers in this package, this mutates the caller's set directly (rather than
+// returning a new one to union in) because the Microsoft Edge case below needs to clear() the caller's existing
+// (incorrect) candidates, not just add to them.
+func candidateVendorsForPE(p pkg.Package, vendors fieldCandidateSet) {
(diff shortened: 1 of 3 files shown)


pujitha24:
Just checking in — this is rebased and green, let me know if anything would make it easier to review.

Issue #5214, opened 2026-08-21
exclude-binary-overlap-by-ownership suppresses vendored libraries owned by unrelated RPMs


sekveaja:
## Description

Syft's `exclude-binary-overlap-by-ownership: true` (default) silently removes binary-detected packages when the file is owned by any RPM — even if that RPM has no relation to the detected software. This causes vulnerabilities in vendored/bundled libraries to go completely undetected.

## Steps to Reproduce

**1. Save the Dockerfile below and build:**

```bash
docker build -f Dockerfile.grype-issue -t grype-binary-cataloger-issue .
```

**2. Dockerfile:**

```dockerfile
# Reproducer: exclude-binary-overlap-by-ownership suppresses vendored OpenSSL
#
# Build:  docker build -f Dockerfile.grype-issue -t grype-binary-cataloger-issue .
#
# Test 1 (default - MISSES 3.5.0):
#   syft grype-binary-cataloger-issue | grep openssl
#
# Test 2 (overlap disabled - FINDS 3.5.0):
#   syft grype-binary-cataloger-issue -c <(echo 'package:
#     exclude-binary-overlap-by-ownership: false') | grep openssl

FROM registry.suse.com/bci/bci-base:15.7

# Install system OpenSSL 3.2.3 via RPM
RUN zypper --non-interactive refresh && \
    zypper --non-interactive install --no-confirm openssl-3 libopenssl3 rpm-build && \
    zypper clean --all

# Build OpenSSL 3.5.0 into /opt/VendorProduct
RUN zypper --non-interactive install --no-confirm gcc gcc-c++ make perl wget tar gzip && \
    zypper clean --all

ARG OPENSSL_VERSION=3.5.0
RUN wget -q https://github.com/openssl/openssl/releases/download/openssl-${OPENSSL_VERSION}/openssl-${OPENSSL_VERSION}.tar.gz \
        -O /tmp/openssl-${OPENSSL_VERSION}.tar.gz && \
    tar -xzf /tmp/openssl-${OPENSSL_VERSION}.tar.gz -C /tmp && \
    cd /tmp/openssl-${OPENSSL_VERSION} && \
    ./Configure --prefix=/opt/VendorProduct --openssldir=/opt/VendorProduct/ssl shared linux-x86_64 && \
    make -j$(nproc) && make install_sw && \
    rm -rf /tmp/openssl-${OPENSSL_VERSION}* /opt/VendorProduct/include /opt/VendorProduct/lib64/pkgconfig

[...]

PR #5215, opened 2026-08-22
fix: don't suppress binaries owned by unrelated OS packages

yzxcj797:
## What?

`exclude-binary-overlap-by-ownership: true` (the default) suppresses a binary finding when **any** OS package owns the file — even when the owning package is completely unrelated to the binary and carries none of its metadata. In grype this silently drops critical CVEs: a vendor RPM bundling its own OpenSSL 3.5.0 owns `/opt/VendorProduct/bin/openssl`, the unrelated `vendor-product` RPM suppresses the binary finding, and CVE-2025-15467 goes unreported with no warning anywhere (https://github.com/anchore/grype/issues/3670).

Fixes the syft side of https://github.com/anchore/grype/issues/3670

## Why?

`identifyOverlappingOSRelationship` excluded the child binary package purely on `parent.Type ∈ OS types` — no relation between the *names* of the owning package and the detected binary was ever considered. The exclusion exists to deduplicate a binary against the distro package that legitimately ships it (#931); applied to an unrelated owner it doesn't deduplicate anything, it just hides the only package record that knew the vendored library's identity and version.

## How?

Gate the OS→binary exclusion on the owning package plausibly shipping the binary: `osPackageShipsBinary(pkgName, binaryName)` — exact match always qualifies; otherwise the binary name (case-insensitive) must appear in the package name, which covers the common shapes (`libopenssl3`, `openssl-3.2.3-150700.1.1`, `golang-go`); binary names shorter than 3 characters only match exactly, so incidental substrings can't suppress anything.

- related owner (the original #931 dedupe intent): unchanged — the binary is still excluded;
- unrelated owner: the binary finding is kept, so grype can match its CVEs;
- JVM and Bitnami overlap paths: untouched.

## Testing

The existing fixtures used unrelated placeholder names (`package-a` owning `package-c`), which under the gate correctly no longer exclude — they now use realistic related names (`libopenssl3` → `openssl`). New cases:

- `unrelated OS owner does not exclude the binary` (the #3670 shape: `vendor-product` RPM → `openssl` binary);
- `very short binary name only matches exactly`.

**Differential**: on `main` the unrelated-owner case fails — the binary is excluded, exactly the reported silent suppression. With this change the whole `internal/relationship` suite passes (`go test ./internal/relationship/`, full package, 1.6s).

### DIFF
--- a/internal/relationship/exclude_binaries_by_file_ownership_overlap.go
+++ b/internal/relationship/exclude_binaries_by_file_ownership_overlap.go
@@ -3,6 +3,7 @@ package relationship
 import (
 	"reflect"
 	"slices"
+	"strings"
 
 	"github.com/anchore/syft/internal/sbomsync"
 	"github.com/anchore/syft/syft/artifact"
@@ -31,6 +32,26 @@ var (
 		reflect.TypeFor[pkg.JavaVMInstallation]().Name(),
 	}
 )
+
+// osPackageShipsBinary reports whether an OS package name plausibly ships the
+// given binary. Package names commonly embed the upstream name with prefixes
+// and suffixes ("libopenssl3", "openssl-3.2.3-150700.1.1", "golang-go"), so an
+// exact match is not required; very short binary names only match exactly, to
+// avoid suppressing on incidental substrings.
+func osPackageShipsBinary(pkgName, binaryName string) bool {
+	p := strings.ToLower(pkgName)
+	b := strings.ToLower(binaryName)
+
+	if p == b {
+		return true
+	}
+
+	if len(b) < 3 {
+		return false
+	}
+
+	return strings.Contains(p, b)
+}
 
 func ExcludeBinariesByFileOwnershipOverlap(accessor sbomsync.Accessor) {
 	accessor.WriteToSBOM(func(s *sbom.SBOM) {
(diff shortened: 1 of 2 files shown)


CAOShurong:
Independently verified this fix against the `#5214` scenario on Windows (go1.26.6, exact head `9d8ac73b`):

- Full `./internal/relationship/` test suite: PASS (including the new unrelated-owner cases)
- Wrote an additional independent test reproducing the issue's shape — an unrelated `vendor-product` RPM owning a file where an `openssl` binary was detected:
  - unrelated owner + ELF binary → **not** excluded (the core `#5214` case) ✅
  - matching OS package name (`openssl` RPM → `openssl` binary) → still excluded ✅
  - `libopenssl3` → `openssl` prefix/suffix matching still works ✅
  - short-name binaries are not suppressed by incidental substring matches ✅
- `go vet` clean

One thing I noticed while reading the diff: the guard only applies when `child.Type` is in `binaryCatalogerTypes`, but the metadata branch right below (`binaryMetadataTypes` like `ELFBinaryPackageNoteJSONPayload`) can also return `child.ID()`. In practice those children appear to always have `Type: BinaryPkg` too, so the path may be unreachable via the metadata branch alone — but worth confirming the intent is to leave that second return unguarded.

PR #5217, opened 2026-08-23
fix(portage): read a CONTENTS path that contains spaces

luantaraschi:
## Description

The portage cataloger splits each CONTENTS line on every space and reads the path from index 1 and the digest from index 2, so an installed file whose path contains a space comes out truncated with part of its own name stored where the MD5 belongs.

Running the parsing as it stands on `main` over a few lines:

```
obj /usr/bin/skopeo 376c02bd3b22804df8fdfdc895e7dbfb 1649284374        -> path="/usr/bin/skopeo" md5="376c02bd3b22804df8fdfdc895e7dbfb"
obj /usr/share/fonts/My Font.ttf d41d8cd98f00b204e9800998ecf8427e 1649284375 -> path="/usr/share/fonts/My" md5="Font.ttf"
obj /opt/A B/c d/e f.so c01eb6950f03419e09d4fc88cb42ff6f 1649284375    -> path="/opt/A" md5="B/c"
obj /usr/bin/skopeo                                                    -> PANIC (index out of range)
```

Two things follow from the truncation. `PortageEntry.OwnedFiles()` hands back a path that points at nothing, and that method is what backs the `FileOwner` interface behind file ownership relationships, so the real file never gets attributed to the package that installed it. And an SBOM consumer checking file digests is given a filename in a field declared as an md5.

The last line is the other half: an `obj` entry with fewer than three fields after it, which is what a truncated or partly written CONTENTS file looks like, indexes past the end of the slice.

Portage's own parser reads the entry anchored at the right. From `lib/portage/dbapi/vartree.py`:

```python
r"(?P<obj>(obj) (.+) (\S+) (\d+))|"
```

The path is greedy and the digest and the mtime are what closes the line, so a path is free to contain spaces. This change reads the line the same way: last field is the mtime, the one before it is the digest, and everything between the keyword and those two is the path. Lines with too few fields are skipped instead of indexed into.

I pulled the line parsing out into its own function so it can be tested without a resolver, and added a table test covering a plain path, a path with one space, a path with several, `dir` and `sym` entries, a truncated `obj` entry and an empty line.

## Type of change

Bug fix

## Checklist

- `go test ./syft/pkg/cataloger/gentoo/...` passes, including the existing `TestPortageCataloger` fixture tests, which are unchanged. The shipped `skopeo-1.5.1` fixture has no path with a space in it, which is why this went unnoticed.
- `go build ./...`, `go vet ./syft/pkg/cataloger/gentoo/...` and `gofmt -l` are clean.

### DIFF
--- a/syft/pkg/cataloger/gentoo/parse_portage_contents.go
+++ b/syft/pkg/cataloger/gentoo/parse_portage_contents.go
@@ -73,19 +73,31 @@ func addFiles(resolver file.Resolver, dbLocation file.Location, entry *pkg.Porta
 	scanner := bufio.NewScanner(contentsReader)
 	for scanner.Scan() {
 		line := strings.Trim(scanner.Text(), "\n")
-		fields := strings.Split(line, " ")
 
-		if fields[0] == "obj" {
-			record := pkg.PortageFileRecord{
-				Path: fields[1],
-			}
-			record.Digest = &file.Digest{
-				Algorithm: "md5",
-				Value:     fields[2],
-			}
+		if record, ok := parseContentsObjectLine(line); ok {
 			entry.Files = append(entry.Files, record)
 		}
 	}
+}
+
+// parseContentsObjectLine reads one "obj <path> <md5> <mtime>" entry of a
+// portage CONTENTS file. An installed path is allowed to contain spaces, so the
+// digest and the mtime are taken from the end of the line rather than from
+// fixed offsets.
+func parseContentsObjectLine(line string) (pkg.PortageFileRecord, bool) {
+	fields := strings.Split(line, " ")
+	// "obj", at least one path field, the digest and the mtime
+	if len(fields) < 4 || fields[0] != "obj" {
+		return pkg.PortageFileRecord{}, false
+	}
+
+	return pkg.PortageFileRecord{
+		Path: strings.Join(fields[1:len(fields)-2], " "),
+		Digest: &file.Digest{
+			Algorithm: "md5",
+			Value:     fields[len(fields)-2],
+		},
+	}, true
 }
 
 func addLicenses(ctx context.Context, resolver file.Resolver, dbLocation file.Location, entry *pkg.PortageEntry) (pkg.LicenseSet, []file.Location) {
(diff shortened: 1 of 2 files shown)


CAOShurong:
Verified at exact head `fb506d46` (base `bf82010f`, go1.26.7 windows/amd64).

**Both defects claimed in the description reproduce on main** -- I drove the old parsing directly:

- truncation: `obj /opt/A B/c d/e f.so c01eb6950f03419e09d4fc88cb42ff6f 1649284375` through the old `strings.Split(line, " ")` indexing yields `path="/opt/A"` and `digest="B/c"`
- crash: an `obj /usr/bin/skopeo` line panics with `index out of range [2] with length 2`

**On the PR head:**

- `go build ./...` and `go vet ./syft/pkg/cataloger/gentoo/...` clean
- the new table test passes all 7 cases (`TestParseContentsObjectLine`)
- right-anchored parsing (last field = mtime, second-to-last = digest, everything between = path) matches portage's own greedy regex in `lib/portage/dbapi/vartree.py`, including its ambiguity profile: a path whose tail itself looks like a hex-digest-plus-number pair would misparse here exactly as it would in portage's vartree parser, so this is no worse than the reference implementation
- confirmed the shipped `skopeo-1.5.1` fixture contains no path with a space, which is why this survived

**Two notes, non-blocking:**

1. Behavior change beyond the fix: a truncated or partially written `obj` line previously aborted the whole scan loudly (panic); now it is dropped silently. Fail-quiet is defensible for one bad line inside an otherwise readable CONTENTS file, but a debug-level log there may be worth considering so a corrupt CONTENTS cannot silently shrink the reported inventory.

2. Environment caveat, not a PR problem: on a Windows checkout, `TestPortageCataloger` and `TestCataloger_Globs` fail identically on base and on this head -- the CONTENTS fixtures come out of git with CRLF and the glob assertions compare slash-separated against backslash-separated paths. Pre-existing at `bf82010f`, untouched by this diff; flagging only so nobody mistakes those local failures for regressions. Incidentally, the right-anchored parse is accidentally tolerant of a stray trailing carriage return: it lands in the unused mtime slot rather than the digest.

luantaraschi:
Thanks for going through it line by line, and for checking it against portage's own vartree parser rather than against my description of it.

Took the first note: `c901ecd` logs a skipped entry at debug instead of dropping it in silence. It is guarded on the `obj ` prefix, since the same branch catches every `dir` and `sym` line the parser has always ignored, and logging those would bury the one worth reading.

The Windows failures are what I get here too, on base and on head alike, so I have left those where they are.

PR #5219, opened 2026-08-23
fix(java): resolve uber-jar version from root version.properties (issue #5163)

### LINKED ISSUE #5163: java-archive-cataloger returns unknown ver for uber-jars

CAOShurong:
## Description

Resolves #5163.

Some build tools (e.g. Clojure's `build.clj`, used by Metabase) record the version **only** in a root-level `version.properties` file (key `tag`), not in the manifest, filename, or `pom.properties`. Syft previously emitted `UNKNOWN`, which dropped the version from the purl and CPE and broke downstream vulnerability matching (e.g. the recent Metabase security advisory).

This adds a final fallback in `discoverNameVersionLicense` that reads the `tag` (or `version`) key from a root `version.properties` after the existing manifest/filename/pom checks fail.

## Verification

End-to-end, scanning a minimal `metabase.jar` (no version in manifest or filename; only `version.properties` with `tag=v0.63.5`):

- **Before:** `metabase@UNKNOWN`, purl `pkg:maven/metabase/metabase` (no version), CPE wildcard
- **After:** `metabase@v0.63.5`, purl `pkg:maven/metabase/metabase@v0.63.5`, CPE `cpe:2.3:a:metabase:metabase:v0.63.5:*:*:*:*:*:*:*`

A regression test fixture (`metabase-version-properties`) and a `Test_parseJavaArchive_regressions` case are included.

## Checklist

- [x] Added / updated unit tests
- [x] `go vet ./syft/pkg/cataloger/java/` passes
- [x] Commit is DCO-signed

> Note: the in-repo `jar-metadata` test fixtures are normally generated via `make` in `testdata/jar-metadata`; the new fixture follows the same layout.


CAOShurong:
Pushed a hardening commit (`eb7e500e`) implementing the root-path tightening discussed in the thread above:

- `versionPropertiesGlob` is now the exact root path `"/version.properties"` instead of `**/version.properties`
- matches are additionally filtered to the exact root entry before extraction, so a nested third-party file can never win regardless of glob semantics or zip entry order

**Verified with a real regression test**: added a vendored `clojure/version.properties` (keys `tag=9.9.9-poison`, `version=9.9.9-poison`) to the `metabase-version-properties` fixture — mirroring the ~40 nested files present in the official Metabase jar. On the previous glob this poisons the cataloged version (reproduced: package resolves to `9.9.9-poison` instead of `v0.63.5`); after this commit the version stays `v0.63.5` from the root file.

Note for reviewers: running the java cataloger suite on Windows hits pre-existing path-separator mismatches (`testdata\` vs `testdata/`) unrelated to this change; on Linux CI the suite runs clean.

PR #5223, opened 2026-08-24
perf(licenses): prefetch remote licenses concurrently for js lockfiles

### LINKED ISSUE #5193: Support for concurrent package registry requests when using `--enrich`

Support for making package registry requests concurrently when using `--enrich`.

At the moment, enrichment requests appear to be made synchronously/sequentially. I verified this by running Syft with verbose logging:

```bash
SYFT_CACHE_DIR=/tmp/syft-cache syft file:pnpm-lock.yaml \
  --select-catalogers npm \
  --enrich javascript \
  -o spdx-json=pnpm.sbom.json \
  -vvvv
```

It would be useful if Syft could perform registry lookups in parallel, ideally with a configurable concurrency limit to avoid overwhelming package registries or hitting rate limits.

**Why is this needed**:

For large lockfiles, synchronous enrichment makes SBOM generation very slow.

In my case, `pnpm-lock.yaml` contained more than 1,500 packages, and the command above took approximately 10 minutes to complete.

Most of the elapsed time appears to be spent waiting for package registry responses. Since these lookups are largely independent of one another, running a bounded number of requests concurrently could significantly reduce overall SBOM generation time.

**Additional Context**:

Syft version: `1.51.0`

CAOShurong:
Fixes #5193.

## Problem

With `--enrich javascript` (or `javascript.search-remote-licenses: true`), JavaScript lockfile package construction performs one npm registry lookup per dependency. A lockfile with roughly 1,500 packages can therefore spend most of a scan waiting on sequential network round trips.

## Change

- Add a `network` executor alongside Syft's existing CPU and file executors, using the same configured `go-sync` concurrency infrastructure.
- For package-lock v2/v3, pnpm, and Yarn lockfiles, deduplicate the package/version pairs that require registry enrichment and warm the existing JavaScript license resolver before package construction.
- Reuse the persistent resolver cache added on current `main` by #5297, with a per-resolver memory fast path so prefetched values are also reused when disk caching is disabled.
- Leave failed lookups uncached, preserving the existing constructor retry and debug logging behavior.
- Treat npm 404 responses as explicit lookup errors rather than attempting to decode the registry error document as package metadata.

The branch is rebased onto current `main@74073aa4`, including #5297, and the earlier cataloger-specific worker pool has been removed in favor of `github.com/anchore/go-sync` as requested.

## Verification

- Focused resolver, prefetch, package-lock, pnpm, and Yarn tests: pass.
- Network-executor bound test: observes a high-water mark of exactly 2 with a configured limit of 2.
- Prefetch suite repeated 20 times: pass.
- `go test ./syft -count=1`: pass.
- `go vet ./syft/pkg/cataloger/javascript ./syft`: pass.
- `git diff origin/main...HEAD --check`: pass.
- The complete JavaScript package suite has the same four Windows-only failure groups on `origin/main@74073aa4` and this head: fixture license expectation, two slash-direction glob assertions, and Yarn dev-dependency filtering. No additional candidate-only failure was observed.
- Race execution is not claimed because this Windows Go environment has CGO disabled.

## AI assistance

Codex assisted with implementation, rebase conflict resolution, and test execution. The exact diff and verification results were reviewed before the branch update.


CAOShurong:
Marked as draft while CI runs — will mark ready for review once checks pass. Full description below.

---

## Problem

With `--enrich javascript` (or `search-remote-licenses: true`), each package constructor calls `getLicenseFromNpmRegistry` inline, so a lockfile with N entries makes N **strictly sequential** HTTP round-trips. For a ~1500-package `pnpm-lock.yaml`, the reporter measured ~10 minutes, almost all of it waiting on the registry.

The lookups are independent of one another, so there is no reason for them to serialize.

## Change

- Add a shared bounded-concurrency helper, `licenses.Prefetch(ctx, keys, n, fn)` (worker pool, default 8), which logs rather than propagates individual failures.
- In the three JS lockfile parsers (`package-lock.json`, `pnpm-lock.yaml`, `yarn.lock`), collect the `(name, version)` pairs that will need a remote lookup and warm a per-parse in-memory cache with concurrent registry requests **before** constructing packages.
- The per-package constructors consult the cache first; on a miss they fall back to the original inline request path, so behavior (including error logging) is unchanged for anything the prefetch could not fetch.
- Treat HTTP 404 from the npm registry as an explicit error instead of trying to decode an error document as JSON.

Dev-dependency filtering and alias handling are applied before building the pair list, so no extra requests are issued for packages that would be skipped anyway.

## Verification

- `go build ./...` exit 0; `go vet` clean on both touched packages
- `go test ./syft/pkg/cataloger/internal/licenses/ -count=1` — all pass (4 new tests: every key attempted, empty-keys no-op, continue-after-error, concurrency high-water-mark ≤ limit)
- `go test ./syft/pkg/cataloger/javascript/ -run "TestNpmLicensePrefetch" -count=1 -v` — 4 new tests pass:
  - exactly one request per pair
  - no requests when `SearchRemoteLicenses` is false
  - concurrency proof: server holds each request open until all N arrive (would time out if serialized)
  - a failing lookup does not prevent later lookups
- Pre-existing failures in the package suite (`Test_JavascriptCataloger`, glob tests, yarn dev-dependency tests) reproduce identically on unmodified `main` via stash-run baseline — environment-dependent, untouched by this change.

syft/pkg/cataloger/internal/licenses/prefetch.go
```diff
@@ -0,0 +1,63 @@
...
+// exactly as it did before this helper existed.
+func Prefetch(ctx context.Context, keys []string, concurrency int, fn func(ctx context.Context, key string) error) {
+	if len(keys) == 0 {
+		return
+	}
+	if concurrency <= 0 {
+		concurrency = defaultPrefetchConcurrency
+	}
+
+	sem := make(chan struct{}, concurrency)
```
kzantow, line 38: This should not be implemented with a bespoke concurrency implementation but rather use the same package the rest of Syft uses for concurrency `go-sync`, e.g.: https://github.com/anchore/syft/blob/356e601cd509fa19107f000479161497105c1ab8/syft/create_sbom.go#L96 -- this allows us to add specifically bounded configuration for the type of operation: IO and we will want to add `network` here to bound concurrent outgoing network requests: https://github.com/anchore/syft/blob/main/syft/cataloging/executor.go#L8

CAOShurong: Addressed in `dfb32266`: the bespoke semaphore/worker helper is removed. Prefetch now uses `go-sync.Collect` with a new `cataloging.ExecutorNetwork` installed by `setContextExecutors`, so outgoing requests share Syft's configured concurrency mechanism. I also rebased onto current `main@74073aa4` (including the new JavaScript persistent cache from #5297), deduplicate package/version pairs, and added a bound test that observes high-water 2 with executor limit 2. Focused tests pass 20 repeated runs; `go test ./syft`, package-scoped vet, and diff checking pass.

PR #5233, opened 2026-08-27
fix(cyclonedx): drop dependency edges that reference packages missing from the catalog

### LINKED ISSUE #4208: CycloneDx SBOM missing components for dependencies
**What happened**: 

When generating a CycloneDx SBOM for a Node.js runtime environment, the SBOM is missing components for found dependencies. 

```json
{
    "dependencies": [
        {
            "ref": "pkg:nuget/Microsoft.JavaScript.Hermes@0.1.23?package-id=205e946433a4b2a8",
            "dependsOn": [
                "pkg:nuget/microsoft.reactnative?package-id=c69beacca8d59ab1"
            ]
        },
        {
            "ref": "pkg:nuget/Microsoft.JavaScript.Hermes@0.1.23?package-id=bed33e0ec05572b4",
            "dependsOn": [
                "pkg:nuget/microsoft.reactnative?package-id=98356cdad9c7d50f"
            ]
        },
        {
            "ref": "pkg:nuget/Microsoft.UI.Xaml@2.8.0?package-id=e87ac46c329bea53",
            "dependsOn": [
                "pkg:nuget/microsoft.reactnative?package-id=c69beacca8d59ab1"
            ]
        },
        {
            "ref": "pkg:nuget/Microsoft.Web.WebView2@1.0.1264.42?package-id=46e8c240d3ab386a",
            "dependsOn": [
                "pkg:nuget/Microsoft.UI.Xaml@2.8.0?package-id=e87ac46c329bea53"
            ]
        },
        {
            "ref": "pkg:nuget/Microsoft.Windows.SDK.BuildTools@10.0.22621.756?package-id=91ea7a1243be1346",
            "dependsOn": [
                "pkg:nuget/Microsoft.WindowsAppSDK@1.5.240227000?package-id=19c6158554d3cf14"
            ]
        },
        {
            "ref": "pkg:nuget/Microsoft.WindowsAppSDK@1.5.240227000?package-id=19c6158554d3cf14",
            "dependsOn": [
                "pkg:nuget/microsoft.reactnative?package-id=98356cdad9c7d50f"
            ]
        },
        {
            "ref": "pkg:nuget/boost@1.83.0?package-id=5ed209f6c5f96b57",
            "dependsOn": [
                "pkg:nuget/common?package-id=26c1efa45aeb6b50",
                "pkg:nuget/folly?package-id=e135b54fa1d2b788",
                "pkg:nuget/microsoft.reactnative?package-id=98356cdad9c7d50f",
                "pkg:nuget/reactcommon?package-id=32aa3702585eac5e"
            ]
        },
        {
            "ref": "pkg:nuget/boost@1.83.0?package-id=673db237af0ca7c7",
            "dependsOn": [
                "pkg:nuget/common?package-id=0676f9089dfbf815",
                "pkg:nuget/folly?package-id=6e79d6e34c9cea55",
                "pkg:nuget/microsoft.reactnative?package-id=c69beacca8d59ab1",
                "pkg:nuget/reactcommon?package-id=0e6e25acd9466060"
            ]
        },
        {
            **"ref": "pkg:nuget/fmt?package-id=79bd16e2078f89d9",**
            "dependsOn": [
                "pkg:nuget/folly?package-id=6e79d6e34c9cea55"
            ]
        },
        {
            "ref": "pkg:nuget/fmt?package-id=f67d2603eb0108f6",
            "dependsOn": [
                "pkg:nuget/folly?package-id=6e79d6e34c9cea55"
            ]
        }
    ]
}
```
Dependency "pkg:nuget/fmt?package-id=79bd16e2078f89d9" does not have a component defined within the SBOM and leads to a CycloneDX validation error:

```One or more Components have Dependency references to Components/Services that are not known in this BOM. They are: {<BomRef 'pkg:nuget/fmt?package-id=79bd16e2078f89d9' id=1712923817328>}```

There should be a matching component defined for each dependency in the SBOM

Below is a simple package.json which will reproduce the issue:

```json
{
  "name": "sbom_missing_component",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@react-native-community/slider": "^4.5.6",
    "react-native-svg": "^15.11.2"
  }
}
```

Once the package.json is created in a directory run ```npm install```

Scan the directory:
```syft scan sbom_missing_component --output cyclonedx-json=sbom_missing_component .json```

CycloneDX validation was performed using [cyclonedx-python-lib](https://github.com/CycloneDX/cyclonedx-python) version: 11.0.0 

**Environment**:
- Output of `node -v`: 20.10.0
- Output of `npm -v`: 10.2.3
- Output of `syft version`: 1.31.0
```
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.22.1
PRETTY_NAME="Alpine Linux v3.22"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

### PR DESCRIPTION
**Pull request by pujitha24**
## Description

Syft's CycloneDX encoder can emit a `dependencies` section that references a component (bom-ref) which does not exist anywhere in the SBOM's `components` list. This happens when an internal `artifact.Relationship` refers to a package that was filtered out of (or never made it into) the final package catalog before formatting — the relationship survives, but the component it points to does not. CycloneDX validators (e.g. cyclonedx-python-lib) reject this with:

```
One or more Components have Dependency references to Components/Services that are not known in this BOM.
```

This change makes `toDependencies()` in the CycloneDX encoder cross-check both sides of a relationship against the final package catalog (`s.Artifacts.Packages`) before emitting a dependency edge, dropping any relationship that references a package not present in the catalog — the same way relationships with un-encodable types are already dropped.

No CLI output or configuration changes.

## Type of change

## Checklist

## Issue references

Report: https://github.com/anchore/syft/issues/4208

## Validation

- Added `syft/format/common/cyclonedxhelpers/to_format_model_test.go::Test_relationships/relationships_referencing_a_package_missing_from_the_catalog_are_dropped`, which builds an SBOM with only 3 packages in the catalog but a relationship pointing at a 4th package that was never added to it. Verified this test FAILS on the pre-fix code (the dangling dependency edge is emitted) and PASSES after the fix (the dangling edge is dropped, the valid edge is kept).
- `go build ./...` passes.
- `go test ./syft/format/...` passes except for 4 pre-existing test functions that require a local Docker daemon to build image fixtures (`exec: "docker": executable file not found in $PATH`) — unrelated to this change and reproducible identically on the unmodified base commit.
- `make lint` (golangci-lint) passes with no findings on the change.

Note: this fix is scoped to the CycloneDX encoder since that's what the linked issue reports. The SPDX and syft-json encoders have the same unguarded pattern (relationships aren't checked against the final catalog) and may be worth a follow-up, but that's out of scope here to keep this change minimal.

Fixes #4208

### DIFF
--- a/syft/format/common/cyclonedxhelpers/to_format_model.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model.go
@@ -88,7 +88,7 @@ func ToFormatModel(s sbom.SBOM) *cyclonedx.BOM {
 	}
 	cdxBOM.Components = &components
 
-	dependencies := toDependencies(s.Relationships)
+	dependencies := toDependencies(s.Relationships, s.Artifacts.Packages)
 	if len(dependencies) > 0 {
 		cdxBOM.Dependencies = &dependencies
 	}
@@ -250,7 +250,7 @@ func isExpressiblePackageRelationship(ty artifact.RelationshipType) bool {
 	}
 }
 
-func toDependencies(relationships []artifact.Relationship) []cyclonedx.Dependency {
+func toDependencies(relationships []artifact.Relationship, catalog *pkg.Collection) []cyclonedx.Dependency {
 	dependencies := map[string]*cyclonedx.Dependency{}
 	for _, r := range relationships {
 		exists := isExpressiblePackageRelationship(r.Type)
@@ -269,6 +269,14 @@ func toDependencies(relationships []artifact.Relationship) []cyclonedx.Dependenc
 		toPkg, ok := r.To.(pkg.Package)
 		if !ok {
 			log.Tracef("unable to convert relationship toPkg to CycloneDX JSON, dropping: %#v", r)
+			continue
+		}
+
+		// a relationship may reference a package that was filtered out of (or never made it into) the
+		// final catalog. Including it here would create a dependency reference to a component that
+		// does not exist in the BOM, which fails CycloneDX validation.
+		if catalog.Package(fromPkg.ID()) == nil || catalog.Package(toPkg.ID()) == nil {
+			log.Debugf("skipping relationship that references a package not in the catalog: %#v", r)
 			continue
 		}
 
--- a/syft/format/common/cyclonedxhelpers/to_format_model.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model.go
@@ -251,6 +251,10 @@ func isExpressiblePackageRelationship(ty artifact.RelationshipType) bool {
 }
 
 func toDependencies(relationships []artifact.Relationship, catalog *pkg.Collection) []cyclonedx.Dependency {
+	if catalog == nil {
+		return nil
+	}
+
 	dependencies := map[string]*cyclonedx.Dependency{}
 	for _, r := range relationships {
 		exists := isExpressiblePackageRelationship(r.Type)
--- a/syft/format/common/cyclonedxhelpers/to_format_model_test.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model_test.go
@@ -134,6 +134,36 @@ func Test_relationships(t *testing.T) {
 			},
 			expected: nil,
 		},
+		{
+			name: "relationships referencing a package missing from the catalog are dropped",
+			sbom: sbom.SBOM{
+				Artifacts: sbom.Artifacts{
+					// p4 is intentionally left out of the catalog (e.g. it was filtered out by a
+					// cataloger after relationships were already built)
+					Packages: pkg.NewCollection(p1, p2, p3),
+				},
+				Relationships: []artifact.Relationship{
+					{
+						From: p2,
+						To:   p1,
+						Type: artifact.DependencyOfRelationship,
+					},
+					{
+						From: p4,
+						To:   p2,
+						Type: artifact.DependencyOfRelationship,
+					},
+				},
+			},
+			expected: &[]cyclonedx.Dependency{
+				{
+					Ref: helpers.DeriveBomRef(p1),
+					Dependencies: &[]string{
+						helpers.DeriveBomRef(p2),
+					},
+				},
+			},
+		},
 	}
 
 	for _, test := range tests {
--- a/syft/format/common/cyclonedxhelpers/to_format_model_test.go
+++ b/syft/format/common/cyclonedxhelpers/to_format_model_test.go
@@ -164,6 +164,22 @@ func Test_relationships(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "nil package catalog does not panic and yields no dependencies",
+			sbom: sbom.SBOM{
+				Artifacts: sbom.Artifacts{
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: syft/format/common/cyclonedxhelpers/to_format_model.go [hunk 1]
```diff
@@ -272,6 +272,14 @@ func toDependencies(relationships []artifact.Relationship) []cyclonedx.Dependenc
 			continue
 		}
 
+		// a relationship may reference a package that was filtered out of (or never made it into) the
+		// final catalog. Including it here would create a dependency reference to a component that
+		// does not exist in the BOM, which fails CycloneDX validation.
+		if catalog.Package(fromPkg.ID()) == nil || catalog.Package(toPkg.ID()) == nil {
```
CAOShurong, line 282: Could we handle a nil catalog before these lookups? Collection.Sorted() is nil-receiver safe, so ToFormatModel() previously tolerated an SBOM with Artifacts.Packages == nil. With any expressible relationship, this new call panics. Returning no dependencies when catalog is nil, with a nil-catalog relationship regression, preserves that boundary.

CAOShurong (changes requested):
Thanks for tightening dependency references against the final package catalog. I found one nil-catalog regression on exact head c6aac5c4fe98da5f3c146cea77aec6edf75e450e.

A focused regression with an expressible relationship and no package catalog fails with a nil-pointer panic at this lookup. Adding a nil guard in toDependencies() makes that regression pass 20/20; the complete cyclonedxhelpers package and go vet for the package also pass. Please preserve the existing nil-catalog behavior and add a regression for this boundary.

I used AI assistance to help inspect the change and run the reproducer; I verified the exact diff, failure, and local results before submitting this review.

CAOShurong (approved):
Follow-up review on exact head 9af36d380aa83060ed0498e1ed6bcfc4490d7082: the new entry guard preserves the prior nil-catalog behavior, and the added regression exercises an expressible relationship with Artifacts.Packages == nil. This resolves my blocking finding without weakening the filtered-package check.

Locally, Test_relationships passed 20/20, the complete cyclonedxhelpers package passed, package-scoped go vet passed, and the old-to-new diff check is clean. The public checks for this new head are still being populated, so normal upstream CI and maintainer review remain separate gates.

I used AI assistance to help inspect the changed head and run the verification; I checked the exact diff and results before submitting this approval.

pujitha24:
Good catch, thanks. `pkg.Collection.Package()` isn't nil-receiver safe like `Enumerate`/`Sorted` are, so the new catalog check panicked on an SBOM with `Artifacts.Packages == nil`. Fixed in 9af36d38 by returning no dependencies when `catalog` is nil, matching the old behavior, and added a regression test for that case.

PR #5234, opened 2026-08-27
feat(composer): add dependency relationships

Xenira:
## Description

Adds dependency relationships for composer.lock and installed.json files

## Type of change

## Checklist

## Issue references

Collective issue for all types:

Refs: #572

### DIFF
--- a/syft/pkg/cataloger/php/dependency.go
+++ b/syft/pkg/cataloger/php/dependency.go
@@ -1,0 +1,47 @@
+package php
+
+import (
+	"github.com/anchore/syft/internal/log"
+	"github.com/anchore/syft/syft/pkg"
+	"github.com/anchore/syft/syft/pkg/cataloger/internal/dependency"
+)
+
+func composerDependencySpecifier(p pkg.Package) dependency.Specification {
+	name, metaProvides, metaRequires, ok := composerMetadataExtractor(p)
+	if !ok {
+		log.Tracef("cataloger failed to extract composer lock metadata for package %+v", p.Name)
+		return dependency.Specification{}
+	}
+
+	// the package name is always provided by the package itself
+	provides := []string{name}
+
+	for name := range metaProvides {
+		provides = append(provides, name)
+	}
+
+	var requires []string
+
+	for name := range metaRequires {
+		requires = append(requires, name)
+	}
+
+	return dependency.Specification{
+		ProvidesRequires: dependency.ProvidesRequires{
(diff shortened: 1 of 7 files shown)


syft/pkg/cataloger/php/dependency.go
```diff
@@ -0,0 +1,47 @@
...
+	name, metaProvides, metaRequires, ok := composerMetadataExtractor(p)
+	if !ok {
+		log.Tracef("cataloger failed to extract composer lock metadata for package %+v", p.Name)
+		return dependency.Specification{}
+	}
+
+	// the package name is always provided by the package itself
+	provides := []string{name}
+
+	for name := range metaProvides {
```
CAOShurong: Composer replace entries also satisfy requirements for the replaced package, but only metaProvides is added here and PhpComposerLockEntry currently does not retain the replace object. An exact-head fixture with implementation.replace[virtual] plus application.require[virtual] produces zero relationships instead of one. Please parse replace for both Composer metadata variants, union those names into Provides, and cover this case.

CAOShurong (changes requested):
Reviewed exact head e49c9f0914ce7b6fcd538861bd7873a516f38544 against base d3734ddeb2d6a7e861355dd6facb730dd4d46537. The new Composer resolver handles package names and provide aliases, but it drops Composer replace aliases.

Composer documents replace as satisfying requirements for the replaced package. In an exact-head composer.lock fixture where acme/implementation replaces acme/virtual and acme/application requires acme/virtual, parsing returns both packages but zero relationships; the provide-equivalent produces the expected relationship. Please parse replace metadata for both lock and installed entries, include those names in the resolver's provided set, and add a regression.

The existing Composer tests passed 20/20, parser-focused tests and package-scoped go vet passed. GitHub currently exposes only a successful DCO check for this head. git diff --check also reports an added blank line at dependency_test.go:255.

I used AI assistance to help inspect the change, check Composer's primary documentation, and run the reproducer; I verified the exact diff and results before submitting this review.

Xenira:
@CAOShurong thanks for the review.

I have added support for the `replace` property. Also added a new composer.lock to the testdata that contains a real world example using replace. Old packages but the simplest example I found.

> The existing Composer tests passed 20/20, parser-focused tests and package-scoped go vet passed. GitHub currently exposes only a successful DCO check for this head.

Not sure if I need to do anything here. If so could you be a little more specific?

PR #5261, opened 2026-09-08
fix(terraform): generate terraform-provider-specific CPEs for terraform lock packages

timols:
The `terraform-lock-cataloger` sets `Language` to `Go`, which causes the Go CPE heuristic to run for terraform provider packages. This produces application-level CPEs that are identical to the actual software — for example, `registry.terraform.io/grafana/grafana` gets `cpe:2.3:a:grafana:grafana:4.45.2`, which is the same CPE as the Grafana web application. Grype then matches these against NVD and flags providers with unrelated CVEs (e.g. CVE-2021-39226).

This adds terraform-specific vendor/product candidate functions following the WordPress plugin pattern. For a provider like `registry.terraform.io/grafana/grafana@4.45.2`, the generated CPE is now `cpe:2.3:a:grafana:terraform-provider-grafana:4.45.2:*:*:*:*:terraform:*:*`, which matches the NVD naming convention for terraform provider vulnerabilities.

Changes:
- New `terraform.go` with vendor/product/name-parsing helpers
- Wire `TerraformLockProviderEntry` into `candidateVendorsByType` and `candidateProductSet` (clears the Go-derived values and replaces them)
- Add `TerraformPkg` to `candidateTargetSw` with `"terraform"`
- Tests for all new functions plus two integration cases in `TestGeneratePackageCPEs`

### DIFF
--- a/syft/pkg/cataloger/internal/cpegenerate/generate.go
+++ b/syft/pkg/cataloger/internal/cpegenerate/generate.go
@@ -189,6 +189,8 @@ func candidateTargetSw(p pkg.Package) []string {
 		return []string{"wordpress"}
 	case pkg.RustPkg:
 		return []string{"rust"}
+	case pkg.TerraformPkg:
+		return []string{"terraform"}
 	}
 
 	return []string{cpe.Any}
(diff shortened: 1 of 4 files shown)


timols:
Hi @wagoodman @spiffcs - let me know if you'd like me to take this in a different direction. Seemed like the easiest/simplest change to make to have a more accurate SBOM vs trying to add terraform language support

timols:
Hi @wagoodman @kzantow - this is generating some noise due to buggy behavior. Any chance you could cast your eyes on this PR so it gets merged and shipped in an upcoming release?

PR #5263, opened 2026-09-09
feat(terraform): add license detection for terraform providers

timols:
Resolves the TODO in `parse_tf_lock.go` — the terraform-lock-cataloger now detects licenses for providers.

**Local resolution**: searches for LICENSE files in the `.terraform/providers/` cache directory (always-on, no network).

**Remote resolution**: when enabled, downloads the provider archive from the Terraform Registry and scans it for license files. Opt-in via `terraform.search-remote-licenses` or `--enrich terraform`.

Follows the same pattern as Go, Python, and JavaScript catalogers for remote license enrichment.


timols:
Hi @wagoodman @kzantow - any chance you could cast your eyes on this PR?

PR #5267, opened 2026-09-10
fix: harden .NET PE/ELF parsing against malformed binaries

wagoodman:
A small .NET executable could describe a very large one and syft would act on this incorrectly: 
- a few bytes in the right header field could result in multi gigabyte allocation
- a ~150 byte resource could make a scan never terminate
- a deeply nested resource tree could overflow the goroutine stack (which is a fatal error instead of a panic) 

Parsing has been adjusted to account for these cases:

- declared sizes are weighed against a real byte count before anything is allocated, then against an absolute ceiling.

- resource offsets are checked against their own section, so one pointing outside it is rejected instead of wrapping into a nonsense size that later checks trust

- the resource walk carries a byte budget and a depth cap

- hitting end of file is an error instead of quietly returning a zeroed struct (that silent success is what let a truncated resource spin: the loop counter advanced by however much had been read, which was nothing)

This means that parsing now better reports partial attempts in the SBOM (reported as unknowns against files).

The net effect is that hostile or corrupt `.exe`/`.dll` gets bounded and reported instead of steering allocation size or loop count, and well formed binaries parse as before.

## Type of change

## Checklist

### DIFF
--- a/syft/pkg/cataloger/internal/dotnet/bundle/bundle_test.go
+++ b/syft/pkg/cataloger/internal/dotnet/bundle/bundle_test.go
@@ -1,0 +1,81 @@
+package bundle
+
+import (
+	"bytes"
+	"encoding/binary"
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// fileWithSignatureAt returns a file of the given size carrying the bundle marker (8-byte header
+// offset followed by the signature) starting at sigStart.
+func fileWithSignatureAt(size, sigStart int, headerOffset uint64) []byte {
+	data := make([]byte, size)
+	binary.LittleEndian.PutUint64(data[sigStart-8:sigStart], headerOffset)
+	copy(data[sigStart:], dotNetBundleSignature)
+	return data
+}
+
+func TestFindSignatureOffset(t *testing.T) { //nolint:funlen
+	const withMarker = 256
+
+	tests := []struct {
+		name        string
+		data        []byte
+		searchLimit int64
+		want        int64
+	}{
+		{
+			name:        "no signature present",
+			data:        make([]byte, 512),
+			searchLimit: 512,
+		},
+		{
+			name:        "zero limit searches nothing",
+			data:        fileWithSignatureAt(withMarker, 64, 0x1234),
+			searchLimit: 0,
+		},
+		{
+			// callers sum unsigned header fields into an int64, which overflows to a negative limit for
+			// some header values. This must read nothing rather than sizing a buffer from it.
+			name:        "negative limit reads nothing",
+			data:        fileWithSignatureAt(withMarker, 64, 0x1234),
+			searchLimit: math.MinInt64,
+		},
+		{
+			name:        "limit past end of file reads only what exists",
+			data:        fileWithSignatureAt(withMarker, 64, 0x1234),
+			searchLimit: math.MaxInt64 / 2,
+			want:        0x1234,
+		},
+		{
+			// the bound must not shrink the window below the real file, or a legitimate marker stops being found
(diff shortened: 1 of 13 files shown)


syft/pkg/cataloger/internal/dotnet/bundle/bundle_test.go
```diff
@@ -0,0 +1,293 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, int64(0x1234), got,
+		"a short read must not abort the search; the marker was in the bytes that were returned")
+}
+
+func TestFindBundleHeaderOffset_UnbackedSizeIsReported(t *testing.T) {
+	data := fileWithSignatureAt(8192, 64, 0x1234)
+
+	// claims twice the bytes it holds, so nothing sized or bounds-checked against that number means anything
+	r := &lyingReader{Reader: bytes.NewReader(data), size: int64(len(data)) * 2}
```
kzantow: What kind of scenario would we have a reader like this? We control the readers, right?

syft/pkg/cataloger/internal/dotnet/bundle/bundle_test.go
```diff
@@ -0,0 +1,293 @@
...
+func (r *lyingReader) ReadAt(p []byte, off int64) (int, error) {
+	if off >= r.Reader.Size() {
+		return 0, io.EOF
+	}
+	return r.Reader.ReadAt(p, off)
+}
+
+func (r *lyingReader) Close() error { return nil }
+
+func (r *shortReader) Read(p []byte) (int, error) {
```
kzantow: nit: these two reader implementations are mixed together

syft/pkg/cataloger/internal/dotnet/pe/bundle_test.go
```diff
@@ -56,3 +60,47 @@ func Test_extractDepsJSONFromBundle_Versions(t *testing.T) {
...
+		r.maxRead = len(p)
+	}
+	return r.Reader.ReadAt(p, off)
+}
+
+func (r *readSizeRecorder) Close() error { return nil }
+
+func TestExtractDepsJSONFromBundle_MalformedSectionSizesDoNotOverAllocate(t *testing.T) {
+	// PointerToRawData and SizeOfRawData are user-controlled and unrelated to the real file size
+	sections := []pe.SectionHeader32{{PointerToRawData: 0xFFFFFFFF, SizeOfRawData: 0xFFFFFFFF}}
```
kzantow, line 92: This test doesn't look like it verifies allocation sizes like some of the others, but I guess with these sizes it would crash?

kzantow (approved):

syft/pkg/cataloger/internal/dotnet/bundle/bundle_test.go
wagoodman: We control the immediate callers, but not every reader that reaches `intFile.ReaderSize`. So in the stdlib elf file and pe file return `io.NewSectionReader(...)`s. I think a similar thing can occur with unionreaders in the squashfs adapter, since there is no `Size()` implemented, so we're hoping that `Seek(...,io.SeekEnd)` will return the declared uncompressed size accurately.

syft/pkg/cataloger/internal/dotnet/bundle/bundle_test.go
wagoodman: I'll clean that up and rename the lying one...

syft/pkg/cataloger/internal/dotnet/pe/bundle_test.go
wagoodman: erg, I don't think it would, but agreed, we shouldn't need to deduce that. I'll make an update.

wagoodman:
Looking at a test that is getting OOM killed now...

PR #5279, opened 2026-09-11
feat(convert): passthrough-exact-format copies matching syft-json input unchanged

juanjsebgarcia:
## Description

`syft convert` always decodes the input and re-encodes it, even when the input is already in the exact format and version that was requested. For large SBOMs that round trip is expensive (minutes and many gigabytes for a 1GB document) and it is also lossy, so callers who normalise SBOMs into a target format without knowing whether the source already matches pay for nothing.

This PR does three things, one per commit:

1. **`convert.passthrough-exact-format`** (`feat`). When a **syft-json** input SBOM exactly matches a requested `-o` output's format and schema version, the input bytes are written to that output verbatim instead of being decoded and re-encoded. Any other input format is always converted. Outputs that do not match are still converted, so a mixed set of `-o` values works as expected, and all outputs are resolved and validated before anything is written, so a bad `-o` value fails without leaving partial results.

   The behaviour is opt-in and **configuration-only**: `convert.passthrough-exact-format: true` in the config file, or `SYFT_CONVERT_PASSTHROUGH_EXACT_FORMAT=true` in the environment. There is deliberately no command-line flag, since this changes how every invocation in a pipeline behaves rather than varying call by call. It defaults to off, so default behaviour is unchanged.

   When enabled, a warning is logged stating so and noting the syft-json-only scope. A second warning is logged whenever an input is found to match and the shortcut is actually taken.

2. **Stream the pass-through instead of buffering the input** (`perf`). Identification, copy and decode all run over a seekable reader. File inputs are `io.Copy`'d straight to the output. STDIN is used in place when it is a redirected regular file and only read into memory when it is a pipe or terminal. Writing pass-through output over the input file itself is detected and left alone rather than truncating the file mid-copy.

3. **Identify syft-json documents from their tail** (`perf`). syft always writes the `schema` block as the last key of the document, so the syft-json decoder's `Identify` now reads at most 2MB from the end of a seekable input, finds the last `"schema"` key, and verifies it belongs to the top-level object by checking that exactly one closing brace ends the document. Anything the tail cannot vouch for (non-seekable input, a reordered document, a nested key) falls back to the existing full parse, so behaviour is unchanged for documents the fast path declines. The check is position-independent, so it also accepts documents whose keys have been sorted alphabetically (as Enterprise exports are), where `source` follows `schema`. Because `Decode` calls `Identify` first, this also removes one of the two full parses every syft-json conversion used to perform.

Supporting changes: the offset-aware reader handed to decoders now supports `io.SeekEnd`; `model.Document` documents that `Schema` must remain the last field, guarded by a test that the encoder writes it last in compact and pretty output; and `format.Decoders()` documents that syft-json must remain the first decoder tried.

### Caveats (documented in the config description and logged when enabled)

- Pass-through applies only to syft-json input, and only when the document's schema version is identical to the one this build of syft produces. An older schema document, or any other format, is always converted.
- Pass-through output is a byte-for-byte copy, so format options such as `format.pretty` and `format.json.legacy` are not applied to it.
- Memory is minimal only for file-path (or shell-redirected) input written with `-o <format>=<path>`. Piped input, or output to STDOUT, must hold the whole document in memory because a pipe cannot seek and the report bus carries a string.

### Performance

All measurements on an Apple M1 Max (10 cores, 32GB), macOS 26.5, using `/usr/bin/time -l`, with the update check disabled. Fixtures are real syft-json documents inflated to size with unique package ids and relationships, at the current schema version. Pass-through output was byte-compared against the input in every run and was identical.

**1GB syft-json document, `syft convert in.json -o syft-json=out.json`**

| Path | Wall | CPU time | Max RSS |
|---|---|---|---|
| `main`, full re-encode | 86.3 s | 99.2 s | 12.0 GB |
| This PR, full re-encode (option off) | 75.2 s | 91.0 s | 9.9 GB |
| This PR, pass-through, file to file | **0.85 s** | **0.5 s** | **66 MB** |
| This PR, pass-through, stdin as redirect | 0.9 s | 0.5 s | 66 MB |
| This PR, pass-through, stdin as pipe | 1.65 s | 1.3 s | 2.7 GB |
| This PR, pass-through, to stdout | 1.46 s | 1.2 s | 3.2 GB |

Pass-through file to file is roughly **100x faster in wall time**, **200x cheaper in CPU**, and uses **180x less memory** than the re-encode on `main`. The remaining 0.85 s is kernel time copying the bytes. The full re-encode itself also improves by about 11 s and 2 GB because the document is no longer parsed twice.

**100MB syft-json document**

| Path | Wall | Max RSS |
|---|---|---|
| `main`, full re-encode | 7.8 s | 1.08 GB |
| This PR, full re-encode | 7.0 s | 1.09 GB |
| This PR, pass-through, file to file | 0.15 s | 65 MB |

**How the pass-through cost fell across the commits (1GB, file to file)**

| Stage | Wall | Max RSS | Dominant cost |
|---|---|---|---|
| Read input into memory, write from memory | 7.4 s | 4.2 GB | Full parse in `Identify` plus a 1GB copy of the input |
| Stream from the file with `io.Copy` | 7.0 s | 3.1 GB | Full parse in `Identify` (measured alone: 7.1 s, 3.1 GB) |
| Identify from the tail | 0.85 s | 66 MB | Copying bytes |

### CLI output

There is no new flag. `syft convert --help` gains one example:

```
  SYFT_CONVERT_PASSTHROUGH_EXACT_FORMAT=true syft convert img.syft.json -o syft-json=out.syft.json    copy the input to "out.syft.json" unchanged if it is already syft-json at the current schema version
```

With the option enabled, a run logs:

```
[0000]  WARN convert.passthrough-exact-format is enabled: syft-json input that already exactly matches a requested output format and schema version will be copied to that output unchanged (this does not apply to any other input format)
[0000]  WARN input already exactly matches the requested output format and schema version, copying it unchanged instead of converting it format=syft-json version=16.1.10 output=syft-json=out.syft.json
```

### Configuration

New `convert` section with one option, also settable as `SYFT_CONVERT_PASSTHROUGH_EXACT_FORMAT`:

```yaml
convert:
  # when a syft-json input SBOM already exactly matches the requested output format and schema version, copy it to that output unchanged instead of decoding and re-encoding it.
  # Note: this only applies to syft-json input, and only when the document's schema version is identical to the schema version this version of syft produces. Any other input is always converted.
  # Note: output written this way is a byte-for-byte copy of the input, so format options (such as 'format.pretty' or 'format.json.legacy') are not applied to it.
  # Note: to keep memory use minimal on large documents, give the input as a file path (or a shell redirect) and write the output with '-o <format>=<path>'. Piped input, or output to STDOUT, requires the whole document to be held in memory.
  passthrough-exact-format: false
```

### Regression matrix

31 end-to-end runs on 1GB-class documents comparing `main`, this branch with the option off, and this branch with the option on. Every output was validated for format and version, byte-compared to its input, and stderr checked for the two warnings. All outcomes were as intended; highlights:

| Input | Target | main | option off | option on |
|---|---|---|---|---|
| syft-json 16.1.10 (1028 MB) | syft-json | 82.5 s / 12.3 GB | 73.1 s / 12.6 GB | **0.80 s / 63 MB**, identical, copied |
| syft-json, keys sorted alphabetically | syft-json | | | **0.86 s / 63 MB**, identical, copied |
| syft-json, schema block mid-document | syft-json | 82.4 s | 82.8 s | 6.96 s / 3.1 GB, identical, copied via full-parse fallback |
| syft-json, older schema 16.0.0 | syft-json | 82.8 s | | 74.0 s, re-encoded as 16.1.10, not copied |
| syft-json 16.1.10 | cyclonedx-json@1.6 | 94.9 s | | 75.3 s, converted, not copied |
| syft-json 16.1.10 | spdx-json@2.3 | 100.2 s | | 85.0 s, converted, not copied |
| cyclonedx-json 1.6 (707 MB) | cyclonedx-json@1.6 | 54.3 s | 54.2 s | 63.6 s, converted, not copied |
| cyclonedx-json 1.6 | cyclonedx-json@1.5 | 54.3 s | | 63.2 s, converted, not copied |
| spdx-json 2.3 (119 MB) | spdx-json@2.3 | 68.2 s | 68.2 s | 70.4 s, converted, not copied |
| spdx-json 2.3 (119 MB) | spdx-json@2.2 | 70.2 s | | 77.2 s, converted, not copied |
| syft-json truncated at 858 MB | syft-json | exit 1 | | exit 1, same error |

Observations:

- With the option off, nothing is slower than `main`; syft-json re-encodes are 11 to 12 percent faster from the single identify pass.
- With the option on, non-syft-json input pays one extra identification pass (about 9 s on 707 MB of CycloneDX, 2 to 7 s on 119 MB of SPDX) because the syft-json tail check declines and its fallback parses the document before ruling it out. Reusing that result in the decode would remove it; left as a follow-up.
- SPDX decoding on `main` is quadratic, a separate pre-existing defect in `spdxhelpers.findRootPackages` addressed in its own PR, which is why the SPDX rows use smaller inputs.

## Type of change

## Checklist

  - `cmd/syft/internal/commands/convert_test.go`: output partitioning (alias, explicit version and path, older schema version, different format, mixed outputs, unknown format, non-syft-json input, unidentifiable input) and end-to-end `RunConvert` for file, redirected stdin, piped stdin, option off, non-syft-json input, `--file` fallback, in-place output, fail-before-write, missing input, undecodable input, and STDOUT publication via the report bus. Pretty-printed input is used so a verbatim copy is distinguishable from a re-encode.
  - `syft/format/syftjson/decoder_identify_test.go`: tail identification across syft key order, alphabetical order, pretty printing, CRLF and whitespace variants, nested schema keys, escaped schema text with braces inside strings, array wrapping, truncation (including inside a string), trailing content, non-syft URLs; a >2MB document over seekable, non-seekable, offset, and no-`SeekEnd` readers; reader position restoration; and a guard that the encoder writes `schema` last in compact and pretty output.
  - `syft/format/internal/stream/seekable_reader_test.go`: `SeekEnd` support and its lower bound.
  - `test/cli/convert_cmd_test.go`: pass-through driven by the environment variable against a built binary, asserting verbatim output, conversion of other formats, default-off behaviour, and the presence or absence of both warnings on stderr.
  - Full re-encode output is unchanged and its timings are equal or better than `main` at 100MB and 1GB, file and pipe input.
  - All unit tests under `cmd/syft/internal`, `syft/format` and `internal` pass apart from those requiring a Docker daemon to build image fixtures, which were not run locally.
  - Per-function coverage of the changed code: `partitionOutputsBySourceFormat` 90%, `RunConvert` 86%, `identifyFromTailBytes` 100%, `closesTopLevelObject` 95%, `identifyFromTail` 68% (remaining lines are seek and read error injection).

## Issue references

No linked issue. Motivated by anchorectl, which embeds `syft convert` to normalise SBOMs and wants to skip the round trip when the source already matches the target.

### DIFF
--- a/cmd/syft/internal/commands/convert.go
+++ b/cmd/syft/internal/commands/convert.go
@@ -1,24 +1,32 @@
 package commands
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/anchore/clio"
+	"github.com/anchore/go-homedir"
 	"github.com/anchore/syft/cmd/syft/internal/options"
 	"github.com/anchore/syft/cmd/syft/internal/ui"
 	"github.com/anchore/syft/internal"
+	"github.com/anchore/syft/internal/bus"
 	"github.com/anchore/syft/internal/log"
 	"github.com/anchore/syft/syft/format"
+	"github.com/anchore/syft/syft/format/syftjson"
+	"github.com/anchore/syft/syft/sbom"
 )
 
 const (
 	convertExample = `  {{.appName}} {{.command}} img.syft.json -o spdx-json                      convert a syft SBOM to spdx-json, output goes to stdout
   {{.appName}} {{.command}} img.syft.json -o cyclonedx-json=img.cdx.json    convert a syft SBOM to CycloneDX, output is written to the file "img.cdx.json"
   {{.appName}} {{.command}} - -o spdx-json                                  convert an SBOM from STDIN to spdx-json
+  SYFT_CONVERT_PASSTHROUGH_EXACT_FORMAT=true {{.appName}} {{.command}} img.syft.json -o syft-json=out.syft.json    copy the input to "out.syft.json" unchanged if it is already syft-json at the current schema version
 `
 )
 
@@ -26,6 +34,7 @@ type ConvertOptions struct {
 	options.Config      `yaml:",inline" mapstructure:",squash"`
 	options.Output      `yaml:",inline" mapstructure:",squash"`
 	options.UpdateCheck `yaml:",inline" mapstructure:",squash"`
+	Convert             options.Convert `yaml:"convert" json:"convert" mapstructure:"convert"`
 }
 
 func Convert(app clio.Application) *cobra.Command {
@@ -34,6 +43,7 @@ func Convert(app clio.Application) *cobra.Command {
 	opts := &ConvertOptions{
 		UpdateCheck: options.DefaultUpdateCheck(),
 		Output:      options.DefaultOutput(),
+		Convert:     options.DefaultConvert(),
 	}
 
 	return app.SetupCommand(&cobra.Command{
@@ -62,30 +72,45 @@ func validateConvertArgs(cmd *cobra.Command, args []string) error {
 func RunConvert(opts *ConvertOptions, userInput string) error {
 	log.Warn("convert is an experimental feature, run `syft convert -h` for help")
 
-	writer, err := opts.SBOMWriter()
+	content, err := readConvertInput(userInput)
 	if err != nil {
 		return err
 	}
 
-	var reader io.ReadSeekCloser
+	outputs := opts.Outputs
+	var unchanged []string
+	if opts.Convert.PassthroughExactFormat {
+		log.Warn("convert.passthrough-exact-format is enabled: syft-json input that already exactly matches a requested output format and schema version will be copied to that output unchanged (this does not apply to any other input format)")
 
-	if userInput == "-" {
-		// though os.Stdin is an os.File, it does not support seeking
-		// you will get errors such as "seek /dev/stdin: illegal seek".
-		// We need to buffer what we read.
-		reader = internal.NewBufferedSeeker(os.Stdin)
-	} else {
-		f, err := os.Open(userInput)
+		unchanged, outputs, err = partitionOutputsBySourceFormat(opts.Output, content)
 		if err != nil {
-			return fmt.Errorf("failed to open SBOM file: %w", err)
+			return err
 		}
-		defer func() {
-			_ = f.Close()
-		}()
-		reader = f
 	}
 
-	s, _, _, err := format.Decode(reader)
+	// resolve (and validate) every output that needs converting before writing anything, so that
+	// a bad -o value fails the command without leaving partial results behind
+	var writer sbom.Writer
+	if len(outputs) > 0 {
+		outputOpts := opts.Output
+		outputOpts.Outputs = outputs
+		writer, err = outputOpts.SBOMWriter()
+		if err != nil {
+			return err
+		}
+	}
+
+	for _, output := range unchanged {
+		if err := writeUnchanged(output, opts.LegacyFile, content); err != nil {
+			return err
+		}
+	}
+
+	if writer == nil {
+		return nil
+	}
+
+	s, _, _, err := format.Decode(bytes.NewReader(content))
 	if err != nil {
 		return fmt.Errorf("failed to decode SBOM: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


cmd/syft/internal/commands/convert.go
```diff
@@ -100,3 +132,150 @@ func RunConvert(opts *ConvertOptions, userInput string) error {
...
+// place; a pipe or terminal cannot seek (you will get errors such as "seek /dev/stdin: illegal seek"), so it is
+// read fully into memory instead.
+func openStdin() (io.ReadSeekCloser, error) {
+	if info, err := os.Stdin.Stat(); err == nil && info.Mode().IsRegular() {
+		if start, err := os.Stdin.Seek(0, io.SeekCurrent); err == nil {
+			return readSeekNopCloser{io.NewSectionReader(os.Stdin, start, info.Size()-start)}, nil
+		}
+	}
+
+	content, err := io.ReadAll(os.Stdin) //nolint:gocritic // a piped SBOM has no known size and must be buffered to be seekable
```
wagoodman: We've been bitten by io.ReadAll like this before; related, we've recently added an internal spillover buffer implementation that will automatically overflow to disk when we're reading "too much" into memory, but it's transparent to the caller. Can you use this instead of copying all of the contents to memory since we just need to return a reader?

juanjsebgarcia: Done, thanks for the pointer. The piped-STDIN path now drains into an internal/spillbuf buffer (io.Copy into an io.NewOffsetWriter, then an io.NewSectionReader over it for identify/copy/decode), rooted per invocation via internal/tmpdir so it lands under the user's TMPDIR, is created lazily only when the 1MB memory tier overflows, and is removed on return. A TempDir already on the context is reused, mirroring syft.CreateSBOM. The regular-file STDIN and path-argument branches are unchanged, and the io.ReadAll nolint on that line is gone. The remaining io.ReadAll is the STDOUT path, where the report bus needs a string. Measured on a 1GB syft-json document piped with pass-through enabled: 2.7GB max RSS before, 66MB after (1.3s vs 1.8s wall). Also exercised: read-only or missing TMPDIR fails cleanly with no partial output, sub-1MB input never creates a temp dir, three concurrent pipes sharing one TMPDIR, and a slow producer where the spill file was observed mid-run and gone afterwards. Folded into the streaming commit.

wagoodman:
Made a couple of changes:

- the tail-based `Identify` trusts the last 2MB without checking those bytes belong to the document the reader is pointed at, so it started saying yes to things it can't decode: a syft-json file with a BOM or a stray log line in front, or a cyclonedx doc followed by a syft one, where `format.Decode` returned nil error with an empty SBOM instead of failing. (that last one isn't convert-only, the sbom cataloger runs `format.Decode` over a bunch of globs during a normal scan)

Fixed with two O(1) checks: the first byte at the reader's position has to be `{`, and `Decode` now errors instead of handing back an empty SBOM when the parsed document has no schema version.

- A smaller one: `syft convert - -o syft-json=sbom.json < sbom.json` truncated the file and exited 0. The `isSameFile` guard was the right idea, it just compared paths and a redirect doesn't have one. Compares `os.FileInfo` now.

cmd/syft/internal/commands/convert.go
```diff
@@ -100,3 +145,171 @@ func RunConvert(opts *ConvertOptions, userInput string) error {
...
+		path = defaultFile
+	}
+
+	if path == "" {
+		// the report bus only carries strings, so STDOUT output must be materialized in memory
+		content, err := io.ReadAll(reader) //nolint:gocritic // the whole SBOM is the report; there is no meaningful bound
+		if err != nil {
+			return fmt.Errorf("unable to read SBOM: %w", err)
+		}
+		bus.Report(string(content))
```
wagoodman, line 272: Just commentary for another time: it'd be nice to one day have something like a `bus.ReportFromReader` However, I think there are downstream callers that are gonna do readalls anyway from it. So we need to simplify the downstream callers first before we try to introduce anything like that.

wagoodman (approved):

PR #5287, opened 2026-09-12
fix(javascript): keep pnpm packages when resolution is nested

### LINKED ISSUE #5241: pnpm cataloger: a devEngines.runtime entry aborts the whole document, yielding an empty SBOM
**What happened**:

A `pnpm-lock.yaml` that contains a `devEngines.runtime` entry (pnpm's Node runtime self-management) produces an **empty SBOM**. Syft exits 0 and writes a well-formed document, so nothing signals that every project dependency was dropped. On a real ~1200-package lockfile this silently yields 0 packages.

The cause is a type mismatch that is then treated as fatal for the whole document. pnpm records a provisioned runtime with a *variations* resolution, where `variants` is a sequence:

```yaml
node@runtime:26.8.1:
  resolution:
    type: variations
    variants:
      - resolution:
          archive: tarball
          integrity: sha256-srdmYPpN7U4LKkHuPAxlHNUuqBcOrZHrrB4UesPVVkM=
          type: binary
          url: https://nodejs.org/download/release/v26.8.1/node-v26.8.1-linux-x64.tar.gz
        targets:
          - cpu: x64
            os: linux
```

`pnpmV9PackageEntry.Resolution` is declared as `map[string]string` in `syft/pkg/cataloger/javascript/parse_pnpm_lock.go`, so the sequence cannot be decoded:

```
failed to unmarshal pnpm v9 lockfile: yaml: unmarshal errors:
  line 29: cannot unmarshal !!seq into string
```

On its own that would be harmless — `yaml.v3` reports this as a `*yaml.TypeError` and still populates every other entry. The problem is that `Parse` treats any error as fatal and discards the decoded result:

```go
func (p *pnpmV9LockYaml) Parse(_ float64, doc *yaml.Node) ([]pnpmPackage, error) {
	if err := doc.Decode(p); err != nil {
		return nil, fmt.Errorf("failed to unmarshal pnpm v9 lockfile: %w", err)
	}
```

So one unrepresentable entry throws away all the others. The multi-document loop then records the error and moves on, leaving nothing behind.

The entries that decode correctly should still be cataloged. An entry Syft cannot represent should be skipped (ideally with a warning), not cause the rest of the lockfile to be dropped.

`pnpm-lock.yaml`:

```yaml
lockfileVersion: '9.0'

settings:
  autoInstallPeers: true
  excludeLinksFromLockfile: false

importers:

  .:
    dependencies:
      is-odd:
        specifier: 3.0.1
        version: 3.0.1
    devDependencies:
      node:
        specifier: runtime:^26.8.1
        version: runtime:26.8.1

packages:

  is-odd@3.0.1:
    resolution: {integrity: sha512-CQpnWPrDwmP1+SMHXZhtLtJv90yiyVfluGsX5iNCVkrhQtU3TQHsUWPG9wkdk9Lgd5yNpAg9jQEo90CBaXgWMA==}
    engines: {node: '>=4'}

  node@runtime:26.8.1:
    resolution:
      type: variations
      variants:
        - resolution:
            archive: tarball
            bin:
              node: bin/node
            integrity: sha256-srdmYPpN7U4LKkHuPAxlHNUuqBcOrZHrrB4UesPVVkM=
            type: binary
            url: https://nodejs.org/download/release/v26.8.1/node-v26.8.1-linux-x64.tar.gz
          targets:
            - cpu: x64
              os: linux
    version: 26.8.1
    hasBin: true

snapshots:

  is-odd@3.0.1: {}

  node@runtime:26.8.1: {}
```

```console
$ syft pnpm-lock.yaml -o json | jq '.artifacts | length'
0
$ echo $?
0
```

Deleting the `node@runtime:26.8.1` entries yields the expected `is-odd`. The error is only visible at `-vvv`:

```
TRACE cataloger returned errors cataloger=javascript-lock-cataloger
  error=/pnpm-lock.yaml: failed to parse pnpm-lock.yaml document 0:
  failed to unmarshal pnpm v9 lockfile: yaml: unmarshal errors:
  line 29: cannot unmarshal !!seq into string
```

A possible fix is to tolerate `*yaml.TypeError` in `pnpmV9LockYaml.Parse` (and `pnpmV6LockYaml.Parse`) — log it and keep the entries that decoded — rather than returning early. Typing `Resolution` as `map[string]any` would additionally let variations resolutions decode cleanly.

This compounds with #5168. A lockfile can hit both at once: pnpm writes its package-manager block as a separate leading YAML document *and* a `devEngines.runtime` entry in the main one. With the #5168 fix in 1.51.1, such a lockfile now parses document 0 (pnpm's own binaries) and fails document 1 (the actual dependency tree), so the SBOM contains only pnpm's release binaries. We hit exactly this on a real project: 1176 packages became 19.

Because the result is a valid-looking, non-empty SBOM, this is easy to ship into dependency-graph or vulnerability tooling without noticing that coverage is gone.

**Environment**:
- Output of `syft version`: `1.51.1` (also reproduces conceptually on 1.33.0, which fails differently — it parses the entry into `pkg:npm/node@runtime%3A26.7.0`)
- OS: Ubuntu (Linux 6.8.0, amd64)

vulragrag-star:
## Description

pnpm can record a provisioned Node runtime (`node@runtime:…`) with a nested `variations` resolution whose `variants` field is a YAML sequence. The v9 parser typed `Resolution` as `map[string]string`, so that entry raised a `yaml.TypeError` and the whole document decode was aborted — which silently dropped every other package from the SBOM while still exiting 0.

This change:

- types `Resolution` as `map[string]any` so nested variation objects decode cleanly
- tolerates `*yaml.TypeError` in the v6/v9 parsers and keeps successfully decoded entries (with a trace log)
- adds a regression fixture covering the issue's lockfile shape

## Type of change

## Checklist

## Issue references

Fixes #5241

### DIFF
--- a/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
+++ b/syft/pkg/cataloger/javascript/parse_pnpm_lock.go
@@ -41,7 +41,10 @@ type pnpmLockfileParser interface {
 }
 
 type pnpmV6PackageEntry struct {
-	Resolution   map[string]string `yaml:"resolution"`
+	// Resolution values are usually strings (integrity, tarball), but pnpm can
+	// also record a nested "variations" object for provisioned runtimes. Keep
+	// this as map[string]any so one unrepresentable entry cannot abort decoding.
+	Resolution   map[string]any    `yaml:"resolution"`
 	Dependencies map[string]string `yaml:"dependencies"`
 	Dev          bool              `yaml:"dev"`
 }
@@ -60,7 +63,9 @@ type pnpmV9SnapshotEntry struct {
 }
 
 type pnpmV9PackageEntry struct {
-	Resolution       map[string]string `yaml:"resolution"`
+	// See pnpmV6PackageEntry.Resolution: nested variation resolutions must not
+	// force the whole lockfile decode to fail (anchore/syft#5241).
+	Resolution       map[string]any    `yaml:"resolution"`
 	PeerDependencies map[string]string `yaml:"peerDependencies"`
 	Dev              bool              `yaml:"dev"`
 }
@@ -86,7 +91,14 @@ func newGenericPnpmLockAdapter(cfg CatalogerConfig) genericPnpmLockAdapter {
 // Parse implements the pnpmLockfileParser interface for v6-v8 lockfiles.
 func (p *pnpmV6LockYaml) Parse(version float64, doc *yaml.Node) ([]pnpmPackage, error) {
 	if err := doc.Decode(p); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal pnpm v6 lockfile: %w", err)
+		var typeErr *yaml.TypeError
+		if errors.As(err, &typeErr) {
+			// yaml.v3 still fills fields it could decode; keep those packages
+			// rather than dropping the whole document for one bad entry.
+			log.WithFields("error", err).Trace("partial pnpm v6 lockfile decode; keeping successfully decoded entries")
+		} else {
+			return nil, fmt.Errorf("failed to unmarshal pnpm v6 lockfile: %w", err)
+		}
 	}
 
 	isV5 := version < 6.0
@@ -124,10 +136,7 @@ func (p *pnpmV6LockYaml) Parse(version float64, doc *yaml.Node) ([]pnpmPackage, 
 		}
 		pkgKey := name + "@" + ver
 
-		integrity := ""
-		if value, ok := pkgInfo.Resolution["integrity"]; ok {
-			integrity = value
-		}
+		integrity := integrityFromResolution(pkgInfo.Resolution)
 
 		dependencies := make(map[string]string)
 		for depName, depVersion := range sortedIter(pkgInfo.Dependencies) {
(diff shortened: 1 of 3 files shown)


spiffcs:
I think we're going to split this into two pieces of work.

### Data Loss
Fixing (#5287): should fix the data loss. 

A `devEngines.runtime` entry made `Resolution` (`map[string]string`) fail to decode, and `Parse` treated that `yaml.TypeError` as fatal for the whole document. I'll push some changes here that get's that portion to pass 😄 

### Part two: Should we catalog the Node runtime itself here?

The PR originally emitted it as `pkg:npm/node@runtime%3A26.8.1`; For now we're not going to include it. I want to settle two questions and ask our larger team before we get this cataloger to emit that package:

- **Package type is unclear.** A provisioned runtime is a tarball of Node binaries, not an npm package. `pkg:npm/node@runtime%3A26.8.1` isn't a purl that identifies anything, and it's not obvious what type it *should* be.

- **Conflict with other Node packages is unclear.** The lockfile records a runtime the project *requests*; it may or may not actually be installed on the scanned filesystem. If it is, syft's binary classifier might already identify that Node so emitting a lockfile-derived entry risks a duplicate or a contradictory one.

Issue #2087, opened 2023-09-01
Adding download location support for other ecosystems (for image-based SBOM)


Auston-Zhang:
**What would you like to be added**:

Add code for 'downloadlocation' for other ecosystems.

**Why is this needed**:
Right now, only NPM(JavaScript) and Apk(Alpine) are supported for download location.
Source code: https://github.com/anchore/syft/blob/007b034ee38063fd5b41c82741e7561448dc817d/syft/formats/common/spdxhelpers/download_location.go#L17

It would be really helpful if we could add code for other ecosystems (like Python and Java).  So the generated SBOM (image-based) for Python packages can have the download location values.

**Additional context**:

Steps to reproduce (we are using the official node docker image https://hub.docker.com/_/node/ ):

1. run syft:  `syft node -o spdx-json > node_image.json`
2. open the generated SBOM file
3.  for a NPM package, we can see the download location
```
"packages": [
  {
   "name": "@colors/colors",
   "SPDXID": "SPDXRef-Package-npm--colors-colors-07f71a6a6289ef27",
   "versionInfo": "1.5.0",
   "supplier": "Person: DABH",
   "originator": "Person: DABH",
   "downloadLocation": "http://github.com/DABH/colors.js.git",
   "filesAnalyzed": false,
   "homepage": "https://github.com/DABH/colors.js",
   "sourceInfo": "acquired package info from installed node module manifest file: /usr/local/lib/node_modules/npm/node_modules/@colors/colors/package.json",
   "licenseConcluded": "NOASSERTION",
   "licenseDeclared": "MIT",
   "copyrightText": "NOASSERTION",
   "description": "get colors in your node.js console",

``` 

for a debian package (just for example, it also applies to Python, Java, Go... packages), we cannot see the download location 

  ```
{
   "name": "adduser",
   "SPDXID": "SPDXRef-Package-deb-adduser-d92ced51b4a5cd89",
   "versionInfo": "3.134",
   "supplier": "Person: Debian Adduser Developers \u003cadduser@packages.debian.org\u003e",
   "originator": "Person: Debian Adduser Developers \u003cadduser@packages.debian.org\u003e",
   "downloadLocation": "NOASSERTION",
   "filesAnalyzed": false,
   "sourceInfo": "acquired package info from DPKG DB: /usr/share/doc/adduser/copyright, /var/lib/dpkg/status",
   "licenseConcluded": "NOASSERTION",
[...]

PR #5288, opened 2026-09-12
feat: add download location for go modules

kjanat:
## Description

`DownloadLocation` in the SPDX helper populated apk, npm, composer, and opam packages but left every Go module at `NOASSERTION`. This adds a case for the Go metadata types (`GolangBinaryBuildinfoEntry`, `GolangModuleEntry`, `GolangSourceEntry`) that builds the public Go module proxy zip URL from the module path and version:

```
https://proxy.golang.org/<escaped-path>/@v/<escaped-version>.zip
```

Paths and versions are encoded with `module.EscapePath` / `module.EscapeVersion` (already available via `golang.org/x/mod`), so case-encoded module paths like `github.com/BurntSushi/toml` resolve correctly (`.../!burnt!sushi/...`).

The standard library (`stdlib`) and non-module versions such as `(devel)` or a bare Go toolchain version are skipped and remain `NOASSERTION`, since they are not fetchable from the proxy.

Refs #2087.

## Type of change

- Enhancement (non-breaking change which adds functionality)

## Tests

Added table cases to `Test_DownloadLocation` covering a normal module, uppercase-escaped path, module/source entries, stdlib, and `(devel)`. `spdxjson` and `spdxtagvalue` format suites pass unchanged.

### DIFF
--- a/syft/format/internal/spdxutil/helpers/download_location.go
+++ b/syft/format/internal/spdxutil/helpers/download_location.go
@@ -1,10 +1,13 @@
 package helpers
 
 import (
+	"fmt"
 	"net/url"
 	"strings"
 
 	urilib "github.com/spdx/gordf/uri"
+	"golang.org/x/mod/module"
+	"golang.org/x/mod/semver"
 
 	"github.com/anchore/syft/syft/pkg"
 )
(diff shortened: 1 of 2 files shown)


kjanat:
Putting this back into draft for now. I noticed one correctness issue.

Hard-coding "https://proxy.golang.org" assumes every valid Go module/version is available there, which is not true for "direct", custom "GOPROXY", "GOPRIVATE", or "GONOPROXY".

kzantow:
FYI there is [some functionality to retrieve the go modules via goproxy](https://github.com/anchore/syft/blob/41870fec7f11b5a6bb0fc8b25dfe4c27825a39b0/syft/pkg/cataloger/golang/licenses.go#L82) including checking the environment variables you mentioned like `GOPROXY`. It seems less than ideal to construct a URL and claim it's the download location without verifying it, maybe there could be an enrichment (using the `--enrich` flag) to include download locations in the metadata for golang packages, then you'd just be surfacing the URL to the SPDX format rather than constructing it while exporting. It could possibly coexist with the license download, which may have already downloaded the module from the URL, or try a `HEAD` request to verify or possibly just construct a URL using the same logic the license download does.

PR #5289, opened 2026-09-13
fix(java): exclude pom profile dependencies that cannot be active

### LINKED ISSUE #5269: java-pom-cataloger reports dependencies from profiles that cannot be active
### What happened

`java-pom-cataloger` catalogs the dependencies of every `<profile>` in a pom, without looking at the profile's `<activation>`. Profiles that cannot be active in any build, and profiles explicitly marked `<activeByDefault>false</activeByDefault>`, contribute packages to the SBOM.

### How to reproduce

A single project pom at the scan root, no jars and no `META-INF` anywhere:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.example</groupId>
  <artifactId>probe-app</artifactId>
  <version>1.0.0</version>
  <dependencies>
    <dependency>
      <groupId>com.google.guava</groupId><artifactId>guava</artifactId><version>32.1.2-jre</version>
    </dependency>
  </dependencies>
  <profiles>
    <profile>
      <id>never-active</id>
      <activation><jdk>[99,)</jdk></activation>
      <dependencies>
        <dependency>
          <groupId>com.example</groupId><artifactId>only-on-jdk99</artifactId><version>9.9.9</version>
        </dependency>
      </dependencies>
    </profile>
    <profile>
      <id>explicit-off</id>
      <activation><activeByDefault>false</activeByDefault></activation>
      <dependencies>
        <dependency>
          <groupId>com.example</groupId><artifactId>off-by-default</artifactId><version>0.0.1</version>
        </dependency>
      </dependencies>
    </profile>
  </profiles>
</project>
```

```
$ syft scan dir:. -q
NAME            VERSION     TYPE
guava           32.1.2-jre  java-archive
off-by-default  0.0.1       java-archive
only-on-jdk99   9.9.9       java-archive
probe-app       1.0.0       java-archive
```

`only-on-jdk99` requires JDK 99 or later. `off-by-default` is switched off in the pom. Both are reported.

### What I expected

`guava` and `probe-app`. A profile requiring a JDK range that no JDK satisfies cannot contribute dependencies to any build, and one marked `activeByDefault=false` with no other activator is off unless requested by `-P`.

### Where

`syft/pkg/cataloger/java/internal/maven/resolver.go` (`main` at 392539d):

```go
// DirectPomDependencies returns all dependencies directly defined in a project, including all defined in profiles.
// This does not resolve any parent or transitive dependencies
func DirectPomDependencies(pom *Project) []Dependency {
	dependencies := deref(pom.Dependencies)
	for _, profile := range deref(pom.Profiles) {
		dependencies = append(dependencies, deref(profile.Dependencies)...)
	}
	return dependencies
}
```

The comment says "including all defined in profiles", so this looks deliberate, which is why I am raising it as a question rather than asserting a mistake. `pomManagedDependencies` merges profile `<dependencyManagement>` the same way; that one seems harmless, since managed entries only supply versions for dependencies that are declared elsewhere.

### Why it matters

Inactive-profile dependencies are packages a vulnerability matcher will act on, for software that is not there. This is not a synthetic concern — I found it in a released artifact. `xmlsec-2.3.4.jar` from Maven Central carries a `jdk19-plus` profile, so scanning it unpacked reports `jakarta.xml.bind-api 2.3.3`, `jaxb-runtime 2.3.3` and `saaj-impl 1.5.3` regardless of the JDK involved, and two of those three are `<scope>test</scope>` as well. Profiles keyed on `<os>`, `<jdk>` or a build property are common in libraries that ship platform-specific or JDK-specific code paths, and by construction at most one of a set of mutually exclusive profiles can apply to a given build.

### The trade-off, as I understand it

Being inclusive is defensible: Syft scans an artifact or a tree, not a build invocation, so it cannot know which profile a build would activate, and dropping profile dependencies would lose real ones for anybody who does build with `-P`. That argues for reporting them, and it is presumably the reason for the current code.

Against it: some profiles are knowably inapplicable from the pom alone, without guessing anything about the build.

So there is a cheap subset that needs no policy decision about profiles in general — skip a profile when its activation makes it impossible or explicitly off:

- `<activeByDefault>false</activeByDefault>` and no other activation element,
- a `<jdk>` range that no released JDK satisfies.

Anything beyond that (evaluating `<os>`, `<property>`, `<file>`, or honouring `activeByDefault=true` semantics the way Maven does, where a default profile is suppressed as soon as another profile in the pom activates) is a larger behaviour change, and I would rather hear which direction you want before writing it.

Options as I see them, smallest first:

1. Skip only the impossible and explicitly-off profiles above.
2. Evaluate activation against the scanning environment, the way `mvn` would.
3. Keep the current behaviour and document it, so consumers know pom-derived packages include conditional ones.

I am happy to implement whichever you prefer, tests included. Note that my account is at this repository's open pull request limit, so it may arrive as a draft until one of my open PRs is resolved.

### Anything else

Independent of #4832 and #5258, which concern `META-INF/maven/` embedded poms. The `xmlsec` numbers above came up while measuring those, but this reproduces on an ordinary project pom with no archive involved.

### Environment

- `syft version`: built from `main` at 392539d (SchemaVersion 16.1.10, go1.27.0)
- OS: Ubuntu 24.04 (WSL2)

kokhlo:
## Summary

`java-pom-cataloger` appended `<dependencies>` from every `<profile>` in a pom without evaluating `<activation>`, so profiles that cannot be active in any build still contributed packages to the SBOM (`DirectPomDependencies` in `syft/pkg/cataloger/java/internal/maven/resolver.go`). A vulnerability matcher then acts on software that is not in the artifact. The motivating case is the probe pom from the issue, whose `off-by-default` profile (`<activeByDefault>false</activeByDefault>` as its only activation configuration) is exactly what this PR learns to drop.

> **Correction (per [review](https://github.com/anchore/syft/pull/5289#issuecomment-5714531186)):** an earlier revision of this summary cited `xmlsec-2.3.4.jar` as a case fixed here. It is not — its `jdk19-plus` profile activates on `[9,)`, which is satisfiable, so its JAXB/SAAJ packages correctly remain with this change. Those packages trace to `<scope>test</scope>` dependencies and to the pom being an embedded `META-INF/maven/` build descriptor, which are tracked in #4832 and #5258.

This PR skips two categories of knowably-inactive profiles, judging only from the pom (minimal Maven activation semantics):

- **explicitly off**: the only activation configuration is `<activeByDefault>false</activeByDefault>` — the pom documents the profile as off unless requested with `-P`;
- **unsatisfiable JDK requirement**: a `<jdk>` range that no JDK can ever satisfy — an empty range (e.g. `[17,9)`) or a bounded range below every JDK (e.g. `(,1.0)`).

**Open upper bounds are satisfiable** (semantics adopted from the review, landing as a follow-up commit on this branch): a range like `[9,)` has no ceiling to check — any lower bound is met by a high enough JDK, whether or not one has shipped yet — and a bare version is treated as a minimum. This keeps the routine gate-on-newest-JDK shape (`[27,)` as of this writing) working without a newest-known-JDK constant that goes stale at every JDK release, at the cost of counting a synthetic `[99,)` as satisfiable. Concretely: `[9,)` / `[27,)` / `[99,)` → satisfiable; `[17,9)` / `(,1.0)` → unsatisfiable; `!1.4` → satisfiable (parse-failure fallback).

Kept **included** (potentially active):
- profiles with no `<activation>` block (can be selected with `-P`);
- profiles with a satisfiable `<jdk>` requirement (e.g. the common `java9+` / `[9,)`);
- profiles with `<os>`, `<property>`, or `<file>` activators — these are environment-dependent, and evaluating them against the scan environment would be the larger behavior change discussed as option 2 in the issue; happy to take that in a follow-up if wanted.

`pomManagedDependencies` is unchanged: managed entries only supply versions for dependencies declared elsewhere, so merging them from all profiles is harmless as the issue notes.

Implementation notes:
- new `profileCanBeActive` helper (`syft/pkg/cataloger/java/internal/maven/profile_activation.go`) with unit tests covering no-activation, activeByDefault true/false (standalone and combined with a jdk activator), satisfiable/unsatisfiable jdk ranges, bare-version jdk requirements, and os/property/file activators;
- `jdkRequirementSatisfiable` parses Maven version ranges (`[9,)`, `[17,9)`, `(,1.0]`, `[1.5]`) and bare versions. A nil upper bound — and a parsed bare version — is satisfied without probing; only bounded ranges are checked against released JDK versions, so there is no newest-known-JDK ceiling to go stale. Unparseable requirements are conservatively treated as satisfiable (never drop dependencies on a parse failure).

## Testing

```
$ go test ./syft/pkg/cataloger/java/ -run 'Pom|pom|Profile|profile' -v
--- PASS: Test_parsePomXML (0.00s)
--- PASS: Test_parseCommonsTextPomXMLProject (0.02s)   # existing java9+ profile expectations unchanged
--- PASS: Test_parsePomXMLProject (0.00s)
--- PASS: Test_parsePomXMLSkipsInactiveProfiles (0.00s)  # regression: pom from the issue
--- PASS: Test_corruptPomXml (0.00s)
...
$ go test ./syft/pkg/cataloger/java/internal/maven/
ok  github.com/anchore/syft/syft/pkg/cataloger/java/internal/maven	0.552s   # incl. Test_profileCanBeActive (13 cases), Test_DirectPomDependencies_skipsInactiveProfiles
```

The regression test uses the exact pom from the issue as a fixture (`testdata/pom/inactive-profiles/pom.xml`) plus a satisfiable `jdk 9+` profile that must still contribute its dependency.

Note: full `go test ./syft/pkg/cataloger/...` cannot pass on this machine — a dozen packages (binary, debian, dotnet, golang, …, java archive tests) require `docker` to generate test fixtures, which is not installed here. Verified identical failures on a clean checkout of `main` (`git stash -u` → same `docker: command not found` failures), so they are environmental, not introduced by this change; CI will run them.

Fixes #5269


Eljees:
Thanks for picking this up, and for using the probe pom as the regression fixture. I built the branch at `bfce632` and ran it against the same artifacts I used for the report. Two findings, one of which affects what the PR claims to fix.

### 1. The `xmlsec` example in the summary is not affected by this change

The summary cites `xmlsec-2.3.4.jar` as the motivating case. That profile activates on a range that *is* satisfiable:

```
profile release        activation{property=''}   deps=0
profile jdk18          activation{jdk='1.8'}     deps=0
profile jdk19-plus     activation{jdk='[9,)'}    deps=3   <- the JAXB/SAAJ three
```

So `profileCanBeActive` returns true for it, and the three packages stay. Scanning the real unpacked jar:

```
main     20 packages
#5289    20 packages   (identical; jakarta.xml.bind-api, jaxb-runtime, saaj-impl all still present)
```

That is not an argument against the PR — the `activeByDefault=false` half is real and worth having, and my probe pom's `off-by-default` profile is correctly dropped. It is only the summary that overstates the reach. The `xmlsec` three are better addressed elsewhere: two of them are `<scope>test</scope>`, and the pom is an embedded `META-INF/maven/` build descriptor, which is the subject of #4832 and #5258. My report used `xmlsec` to show the activation problem exists in released artifacts, and I should have been clearer that its own profile is a satisfiable one. Apologies for the misdirection.

### 2. `maxKnownJDKMajorVersion = 26` is already stale, and the failure direction is the unsafe one

JDK 27 reached general availability on 15 September 2026, two days ago. On the branch head:

```
jdkRequirementSatisfiable("[25,)") = true
jdkRequirementSatisfiable("[26,)") = true
jdkRequirementSatisfiable("[27,)") = false   <- current JDK
jdkRequirementSatisfiable("[28,)") = false
jdkRequirementSatisfiable("[17,9)") = false  <- correct, empty range
jdkRequirementSatisfiable("!1.4")   = true   <- negation survives via the parse-failure fallback, good
```

A pom with `<jdk>[27,)</jdk>` — the normal way to gate code on the current release — loses that profile's dependencies today. The code comment says the bound "only becomes more permissive as new JDKs ship (never drops profiles it previously kept)", but the direction is the other way: while the constant sits still and the JDKs move, it drops profiles it should keep. It is the false-drop this PR is otherwise careful to avoid, on a six-month clock.

### Suggestion

For a range with an open upper bound the ceiling is not needed at all: any lower bound is met by a high enough JDK, if not one released today then one released later. Same for a bare version, which is treated as a minimum. The genuinely unsatisfiable cases survive, since they all have an upper bound.

```go
	if r.upper == nil {
		// open upper bound: satisfied by a high enough JDK, whether or not one has shipped yet.
		return true
	}
	return anyProbeMatches(r.matches)
```

plus returning `true` in the bare-version branch once it parses. Measured with that applied:

```
"[9,)"   = true    "[27,)" = true    "[28,)" = true
"[17,9)" = false   "(,1.0)" = false  "!1.4"  = true
```

**The cost, stated plainly:** `[99,)` then counts as satisfiable, so three of your cases change — `jdk unsatisfiable lower bound`, `jdk-99-exact`, and the `only-on-jdk99` half of `Test_DirectPomDependencies_skipsInactiveProfiles` (and the matching `parse_pom_xml` expectation). I verified that is the whole blast radius: nothing else in `.../internal/maven` or the java pom tests moves.

That is a real trade. `[99,)` came from my report, where I wrote it as a synthetic "cannot ever be true" case; I have not seen a pom gate on an unreleased JDK in the wild, whereas gating on the newest one is routine. So my read is that dropping the ceiling costs a case that does not occur and removes a recurring source of wrong output. If you would rather keep the ceiling, then bumping it to 27 fixes today and the comment should say that the constant needs a bump at each JDK release, rather than that staleness is harmless.

Happy to send either version as a commit to your branch if that is easier than doing it yourself — my account is at this repository's open-PR limit, so a commit to yours is more useful than anything I can open. Either way the `activeByDefault` half stands on its own and I would like to see it land.

kokhlo:
Thanks for the thorough review — and for flagging that the `xmlsec` example overstates the reach of this change. You're right on both counts: the `jdk19-plus` profile (`activation{jdk='[9,)'}`) is satisfiable and stays, so the summary claiming the JAXB/SAAJ trio as a motivating case was wrong of me to carry over from the issue report. I've updated the PR description to drop it and to point at #4832 / #5258 as the right place for the embedded-`META-INF/maven/` half of that story.

On the ceiling: I'd like to adopt your nil-upper-bound semantics. The trade you stated is convincing — `[99,)` was a synthetic case from your report, while gating on the newest released JDK (`[27,)` today) is the routine shape, and the current constant drops exactly those, in the unsafe direction, on a six-month clock. Treating an open upper bound as satisfied (plus `true` in the bare-version branch) removes the staleness failure mode entirely rather than bumping the constant per release. Adopted semantics, matching your measurements:

| requirement | satisfiable |
|---|---|
| `[9,)`, `[27,)`, `[28,)`, `[99,)` | true |
| `[17,9)` (empty range) | false |
| `(,1.0)` | false |
| `!1.4` (negation via parse-failure fallback) | true |

And yes — happy to take you up on the offer. Since your account is at the open-PR limit, please push your proposed commit to `kokhlo:syft` branch `fix-5269-inactive-pom-profiles` (the PR head; currently at `bfce632`). Whichever version you find preferable — the nil-upper-bound one if you have no further preference. If it's easier for you to leave the mechanics to me, I'll apply the nil-upper-bound change and update the three affected cases (`jdk unsatisfiable lower bound`, `jdk-99-exact`, the `only-on-jdk99` half of `Test_DirectPomDependencies_skipsInactiveProfiles` + the matching `parse_pom_xml` expectation) myself — either works, whatever saves you the round-trip.

Eljees:
Let's go with you applying it — simplest path, no need to sort out push access to your branch. The three cases you flagged match what I found too (`jdk unsatisfiable lower bound`, `jdk-99-exact`, the `only-on-jdk99` half of `Test_DirectPomDependencies_skipsInactiveProfiles` + the matching `parse_pom_xml` expectation), so nothing else should move. Thanks for chasing down the `xmlsec` claim and for working through the ceiling trade properly.

kokhlo:
Nil-upper-bound semantics applied in 30932612: a `<jdk>` range with an open upper bound (or a bare version / unparseable value, e.g. `!1.4`) is now treated as satisfiable; only ranges with an upper bound keep the released-JDK probe, so nothing goes stale as new JDKs ship. The three affected cases are updated in the same commit (`jdk open upper bound beyond any released JDK`, `jdk bare version beyond any released JDK`, and the `only-on-jdk99` half of `Test_DirectPomDependencies_skipsInactiveProfiles` + the matching `parse_pom_xml` fixture). Focused tests green: `go test ./syft/pkg/cataloger/java/internal/maven/...` and the pom tests in `./syft/pkg/cataloger/java/...` pass; the pre-existing docker-fixture failures on this machine are unrelated (confirmed on the base commit).

PR #5293, opened 2026-09-14
fix(debian): bound the decompressed members of a .deb archive

wagoodman:
a .deb's member sizes are bounded by the archive, but what they decompress to is not: xz and zstd both beat gzip's ~1000:1, so a few KB member can expand without limit... same for the file lists inside control.tar, where the shortest line that still yields a record is a few bytes. 

Here are the specifics on what was changed:

- byte caps on `control.tar` (16MB, it only holds metadata) and `data.tar` (2GB, mirroring `perFileReadLimit`), enforced at read time rather than by truncating. A tar reader treats a truncated stream as a clean end of archive, so cutting the stream silently would hand back a package missing files with no error at all

- record caps on the `md5sums` and `conffiles` listings, and on the copyright files read out of `data.tar`. the byte caps alone don't bound what those accumulate

- only the first `control.tar` and `data.tar` member are read. The budget is per member, so without this a .deb carrying N of them bought N times the budget

- a partial parse keeps the package and reports why rather than dropping it or going quiet. `processControlTar` always returns whatever it managed to parse, and the caller records an unknown plus a warning, since the unknown itself is removed from the SBOM by default once a package lands on the same coordinates

- `parseDpkgMD5Info` and `parseDpkgConffileInfo` return `(records, error)` instead of swallowing scanner errors, which otherwise turn a single long line into an empty file list with nothing logged

## Type of change

## Checklist


syft/pkg/cataloger/debian/package.go
```diff
@@ -202,7 +202,11 @@ func getAdditionalFileListing(resolver file.Resolver, dbLocation file.Location,
 	if md5Reader != nil && md5Location != nil {
 		defer internal.CloseAndLogError(md5Reader, md5Location.AccessPath)
 		// attach the file list
-		files = append(files, parseDpkgMD5Info(md5Reader)...)
+		records, err := parseDpkgMD5Info(md5Reader)
+		if err != nil {
+			log.Warnf("dpkg md5sums file %q: %v", md5Location.RealPath, err)
```
kzantow: Should these `Warn` be `Debug` level? Also the errors logged in this function seem like candidates for unknowns? it looks like MD5 hashes and conf files -- I think a failure to read the former might result in unexpected file ownership? maybe the conf files aren't as important

Same log level comment a few places below in this file and the next 👇

wagoodman: good find, I'll update

kzantow (approved):

PR #5297, opened 2026-09-15
perf(javascript): cache npm registry license lookups across runs

anze3db:
## Summary

Persist javascript remote license lookups to syft's on-disk cache, so `registry.npmjs.org` is queried once per `name@version` instead of once per package occurrence per run.

## Motivation

With `javascript.search-remote-licenses` enabled, `getLicenseFromNpmRegistry` goes to the network at all five package-constructor call sites with nothing in front of it. The same `name@version` is fetched again for every lockfile that mentions it, and again on the next scan.

The python (`pypi.org`), golang (`proxy.golang.org`) and java (maven POMs) catalogers all cache these lookups already. JavaScript was the only one that didn't.

To see it, against a fixture already in this repo:

```sh
export SYFT_JAVASCRIPT_SEARCH_REMOTE_LICENSES=true SYFT_LOG_LEVEL=info SYFT_CACHE_DIR=/tmp/syft-repro-cache

syft scan dir:syft/pkg/cataloger/javascript/testdata/yarn --select-catalogers javascript \
  -o syft-json=/dev/null 2>&1 | grep -c "downloading javascript package from npm"
```

Run it twice. Before: `9`, then `9` again, with no `javascript/` directory under `/tmp/syft-repro-cache`. After: `9`, then `1`, with 8 entries under `javascript/v1/`. The one remaining request is the fixture's deliberately non-existent `something-i-made-up@7.7.7` — an error, so not cached, by design.

## Changes

- Add `javascriptLicenseResolver` in `syft/pkg/cataloger/javascript/license.go`, modelled on `pythonLicenseResolver` in `syft/pkg/cataloger/python/license.go`. Keyed on `"<name>/<version>"`, namespaced `javascript` / `v1`, so entries sit alongside the existing `python/v1/...` and `golang/v2/...`.
- Route the package-lock v1, package-lock v2, pnpm, yarn and bun package constructors through it. `getLicenseFromNpmRegistry` itself is unchanged.
- Build the resolver in each lock adapter's constructor, the same place python builds its own.
- Unlike python, use `cache.GetResolver` rather than `cache.GetResolverCachingErrors`, so **errors are not cached**. A 429 or a name that 404s would otherwise blank out licenses for the cache TTL (7 days by default), and a missing license silently weakens downstream license policy checks. A successful lookup returning an empty license is a real answer and is cached.
- Add `license_test.go`: repeat lookups hit the registry once, separate resolvers share one cache, empty licenses are cached, errors are retried every time, and the key distinguishes name and version including scoped names.

No breaking changes and no migration. Caching is transparent — SBOM output is unchanged, and the cache honours the existing `cache.dir` and `cache.ttl` settings.

## Type of change

## Checklist


kzantow (approved):
Thanks for the contribution!

PR #5300, opened 2026-09-17
fix(githubactions): emit a valid pkg:oci PURL for docker:// use statements

### LINKED ISSUE #5299: github-actions cataloger emits an invalid PURL for docker:// use statements
**What happened**:

The github-actions cataloger builds a `pkg:github` PURL for every `uses:` statement, including `docker://` image references. `packageURL()` splits the name on `/` assuming an `owner/repo` path, so `docker://ghcr.io/sethvargo/ratchet:latest` produces namespace `docker:`, an **empty name**, and the image reference stranded in the subpath:

```
pkg:github/docker%3A/@sha256%3A527e78e6d29a...#ghcr.io/sethvargo/ratchet:latest
```

`packageurl.FromString()` rejects this string: `purl is missing name`. Syft is emitting a PURL its own parser will not read.

A `docker://` use statement references an OCI image, not a GitHub repo, so `pkg:github` is the wrong type. Either a valid image PURL:

```
pkg:oci/ratchet@sha256%3A527e78e6d29a...?repository_url=ghcr.io%2Fsethvargo&tag=latest
```

or no PURL at all — which the cataloger already does for local composite actions (`namespace == "."` → `return ""`). Whichever type is chosen, the output should round-trip through packageurl-go.

```bash
mkdir -p /tmp/ghrepro/.github/workflows
printf 'on: [push]\njobs:\n  build:\n    steps:\n      - uses: actions/checkout@v4\n      - uses: docker://ghcr.io/sethvargo/ratchet:latest@sha256:527e78e6d29a9ac306e843cf766afe0604a0b35633f16913d85dd522218e8ca1\n' \
  > /tmp/ghrepro/.github/workflows/test.yml

syft scan dir:/tmp/ghrepro -o syft-json | jq -r '.artifacts[] | "\(.name)\t\(.purl)"'
```

```
actions/checkout                            pkg:github/actions/checkout@v4
docker://ghcr.io/sethvargo/ratchet:latest   pkg:github/docker%3A/@sha256%3A527e78e6d29a...#ghcr.io/sethvargo/ratchet:latest
```

The `actions/checkout` line is the control — same cataloger, valid PURL.

- Source: `packageURL()` in `syft/pkg/cataloger/githubactions/package.go`.
- The no-digest form (`uses: docker://ghcr.io/sethvargo/ratchet:latest`) has no digest to put in the `oci` version field, which the purl spec requires — so that case may want `pkg:docker` or no PURL.
- Downstream impact: consumers that key on the PURL lose every identifying token. The name is empty and the registry/org/repo only exist in the subpath, which anything comparing base PURLs discards. Container-based actions are documented Actions syntax, so this isn't an exotic input.

**Environment**:
- Output of `syft version`: 1.52.0 (also reproduced on 1.49.0; the cataloger is unchanged since at least v1.38.0)
- OS: macOS, darwin/arm64

chenzeyan54-commits:
## What I did

-  in the github-actions cataloger split every name on `/` assuming an `owner/repo` path, so a `docker://ghcr.io/org/image:tag@sha256:...` use statement produced `pkg:github/docker%3A/@sha256%3A...` — a PURL with an empty name that `packageurl.FromString()` rejects (`purl is missing name`).
- A `docker://` use statement references an OCI image, not a GitHub repository. It now emits a PURL of type `oci`:
  - last path segment → PURL name (`ratchet`)
  - registry + organization path → `repository_url` qualifier (`ghcr.io/sethvargo`)
  - tag, when present → `tag` qualifier (`latest`)
  - pinned digest → PURL version (`sha256:527e...`)
  - no tag and no digest → no PURL at all, consistent with the existing handling of local composite actions

Result for the issue's reproducer:

```
pkg:oci/ratchet@sha256%3A527e78e6...?repository_url=ghcr.io%2Fsethvargo&tag=latest
```

which round-trips cleanly through `packageurl.FromString()` (asserted in the new tests).

## Verification

- `go test ./syft/pkg/cataloger/githubactions/` — all green, including 5 new table cases (digest-pinned, tag-only, untagged, short name, registry with port)
- end-to-end: built syft from this branch and scanned the issue's reproducer workflow; the docker:// line now yields the valid oci PURL above, `actions/checkout` unchanged

Closes #5299


wagoodman:
though it would be ideal to be able to get an OCI purl... the vast majority of the time, we're not going to have a digest, which means we can't make a valid purl. This PR doesn't show what we would do in such cases where we're missing the digest.

I'm going to merge #5302 first so that we at least are skipping these cases and if you find a nice way to optionally get OCI purls in cases where we detect the digest then feel free to open up a new PR 👍

